Issue #230, opened 2026-07-15
yaml/simd/x86: signed compare misreads UTF-8 bytes as control chars, panicking yq on non-ASCII


newhoggy:
## Summary

On x86_64, `succinctly yq '.field'` **panics** on any YAML whose string values contain
non-ASCII characters and are long enough to reach the SIMD path:

```console
$ printf -- '---\nwanted: love ♥ and peace ☮\n' > h.yaml
$ succinctly yq -o json -I 0 '.wanted' h.yaml
thread 'main' panicked at src/jq/stream.rs:145:29:
start byte index 7 is not a char boundary; it is inside '♥' (bytes 5..8 of string)
```

Field access is the most basic yq operation, and non-ASCII in YAML config is routine.
ARM64 is unaffected.

## Cause

`src/yaml/simd/x86.rs`, in `find_json_escape_avx2` / `find_json_escape_sse2`:

```rust
let control_threshold = _mm256_set1_epi8(0x20);
let is_control = _mm256_cmpgt_epi8(control_threshold, chunk);
```

`_mm256_cmpgt_epi8` is a **signed** compare. The predicate wanted is `byte < 0x20`
*unsigned*. For any byte >= 0x80 — i.e. every UTF-8 lead and continuation byte — the value
reads as negative, so `0x20 > byte` is true and the byte is reported as a control character
needing escaping.

`find_json_escape` therefore returns an index in the middle of a multi-byte character. Both
callers then slice on it:

- `src/yaml/light.rs:1907` — `write_json_string`, used by `YamlCursor::to_json`
- `src/jq/stream.rs:141` — `write_json_escaped`, used by the jq streaming path

`&s[i..escape_pos]` on a non-boundary index panics. (When it doesn't panic outright, the
byte falls through the caller's `match` into the `_ => {}` "shouldn't happen" arm, which
advances `i` by one *byte* into the middle of a character — so the alternative to the panic
is silent UTF-8 corruption.)

The existing comment shows the reasoning that led here — it identifies the problem and
then dismisses it:

```rust
// (bytes 0x80-0xFF would be negative in signed, but we're checking < 0x20)
// Actually need: unsigned byte < 0x20
// Trick: use _mm256_cmpgt_epi8(threshold, chunk) which gives 0xFF where threshold > chunk
// This works for bytes 0x00-0x1F (all less than 0x20 in both signed and unsigned)
```

NEON is correct because `vcltq_u8` is unsigned; the scalar fallback is correct because
[...]

Issue #192, opened 2026-07-10
test: land the green test scaffolding (CI arch-matrix, skip-visibility, differential harnesses, refactor snapshots)


newhoggy:
The "green, landable-now" half of #191 — test scaffolding that passes on `main` today, carries **no product-code change**, and de-risks the fix issues by making them verifiable and smaller. Split out from #191 so it can be merged independently and ahead of the correctness fixes.

Scope boundary: this issue is **only** the parts that stay green on current `main`. The correctness/differential tests that assert not-yet-true behavior (`1==1.0`, DSV offset-63, non-ASCII `to_json`, overflow boundaries, eval parity, `-e` fast path, `yq -I 0` types, NEON boundary) are **red until their fix lands** and stay in #191 / land with each fix. Don't add those here.

## 1. Make SIMD test coverage trustworthy (meta-hazard)

Today the x86 SIMD tests self-skip on non-matching hardware, e.g. `src/dsv/simd/bmi2.rs:218` `if !has_bmi2() { return; }` (same for AVX2 and SVE2). On aarch64 they pass **without asserting anything**, so a green `cargo test` on ARM proves nothing about the x86 BMI2/AVX2 paths. CI (`.github/workflows/ci.yml`) appears to run a single arch (the PR template only has manual "Tested on x86_64 / aarch64" checkboxes).

- [ ] Make a fully-skipped SIMD suite visible — emit a `SKIPPED (no bmi2)` marker and/or a summary count, so an all-skipped run doesn't read as "passed".
- [ ] Add a test-job **arch matrix** to `ci.yml`: x86_64 (with BMI2 + AVX2 available) **and** aarch64 (NEON). Document that SVE2 (`toggle64_sve2`) has no routine CI runner and how it's validated.

## 2. Reusable differential harnesses (land green with currently-passing cases)

Build the scaffolding now; wire the bug-exposing assertions later (with their fixes).

- [ ] DSV: a generic "build index via backend X vs the scalar parser → assert identical markers/rows/fields" runner, parameterized over {scalar, SSE2, AVX2, BMI2, NEON, SVE2} and a corpus. Land it with inputs the backends **already** agree on. (The offset-0..63 sweep that exposes #149 is red — it lands with #149/#182, not here.)
- [ ] jq: an "`eval_generic` (CLI path) vs `eval.rs` (full evaluator) produce the same result" runner over a builtin corpus, seeded with builtins that already agree. (The diverging builtins land with #157/#161/#162.)

[...]

Issue #191, opened 2026-07-10
test: add missing coverage exposed by the code review (SIMD differential, numeric equality, overflow, CLI fast paths)


newhoggy:
The full code-review pass surfaced correctness bugs whose fixes **cannot be verified, or could regress silently, because the relevant behavior is untested**. This issue tracks closing those gaps. It is a prerequisite/companion to the fix issues it references — several of those fixes should land *with* the corresponding test here.

> **Non-native SIMD coverage split out.** This issue covers the gaps that are developable and verifiable on the Apple Silicon dev host — **scalar + NEON** and the non-SIMD work. The backends that don't run natively here are split out: **x86 (SSE2/AVX2/BMI2)** → #193, and **ARM SVE2** (Apple Silicon is NEON-only; M4 adds SME but not general SVE2) → #194.

## Meta-hazard (fix this first)

The x86 SIMD tests self-skip on non-matching hardware, e.g. `src/dsv/simd/bmi2.rs:218` `if !has_bmi2() { return; }` (same for AVX2 and SVE2). On an ARM dev host / ARM-only CI, **all of these silently pass without asserting**, so a green `cargo test` is false confidence for the x86 BMI2/AVX2 and SVE2 paths.

- [ ] Make hardware-skips visible (print a `SKIPPED` line, or count skips) so a fully-skipped SIMD suite doesn't read as "passed". *(Shared mechanism; the x86 skip sites + CI leg are tracked in #193, the SVE2 skip site + emulation validation in #194.)*
- [ ] Ensure CI runs the SIMD suites on **aarch64 (NEON)**. *(x86_64 CI leg with BMI2 + AVX2 → #193; SVE2 validation under `SUCCINCTLY_SVE2` / emulation → #194.)*

## DSV quote-mask / carry (blocks #149, belongs with #182)

- [ ] Add a cross-backend differential test: for random 64-bit inputs **and a quote at every offset 0..64**, assert each backend (scalar, SSE2, AVX2, BMI2, NEON, SVE2) produces an identical index. The existing `test_quoted_spanning_chunks` only places a quote at offset 0, which is exactly why the bit-63 bug (#149) slipped through. *(Test authored here for scalar + NEON; SSE2/AVX2/BMI2 execution confirmed in #193, SVE2 in #194.)*
[...]

Issue #150, opened 2026-07-09
YAML to_json() drops multibyte UTF-8 on x86 (signed compare in find_json_escape)


newhoggy:
**Severity: High** — x86-only silent data corruption (code-traced; this reviewer is on ARM).

`src/yaml/simd/x86.rs:862,887,931` `find_json_escape` uses **signed** `_mm256_cmpgt_epi8(0x20, chunk)` / `_mm_cmpgt_epi8`, which is true for every byte >= 0x80 (negative as `i8`). The NEON path (`vcltq_u8`, unsigned) and all scalar fallbacks use `b < 0x20` and do not flag those bytes.

The caller `write_json_string` (`src/yaml/light.rs:1898`) treats the flagged byte as an escape; its match falls through `_ => {}` at `src/yaml/light.rs:1935` (comment "shouldn't happen") and **silently drops** each UTF-8 lead/continuation byte.

Reachable via public `to_json()` / `to_json_document()` for any string value >= 16 bytes containing non-ASCII inside a full SIMD chunk (e.g. `café...`). The CLI streaming path uses a correct scalar loop, which is why this is untested.

**Fix:** use an unsigned control-char test (e.g. `_mm256_min_epu8` + `cmpeq`, or saturating-sub) so only bytes < 0x20 match; add a non-ASCII SIMD-vs-scalar test. Also fix the false "shouldn't happen" comment at light.rs:1935.

---
_Filed from an automated full-code-review pass. Severity and verification status noted above._

newhoggy:
Already fixed — duplicate of #230 (same root cause: signed `_mm256_cmpgt_epi8` in `find_json_escape` misreading UTF-8 bytes ≥ 0x80 as control chars).

Fixed in #229 by commit `214ada17` ("fix(simd): use an unsigned compare for JSON control-char detection on x86"), already on `main`. All three x86 sites (AVX2 chunk at `x86.rs:861`, its SSE2 tail at `:888`, and `find_json_escape_sse2` at `:933`) now use the exact unsigned test this issue recommended:

    let is_control = _mm256_cmpeq_epi8(_mm256_subs_epu8(chunk, control_max), zero);

`subs_epu8(b, 0x1F)` saturates to 0 precisely when `b <= 0x1F`, so only bytes < 0x20 match.

[...]

PR #245, opened 2026-07-19
test(simd): implement comprehensive x86 SIMD test coverage and hard feature expectations for CI

### LINKED ISSUE #193: test: x86 SIMD coverage — CI leg + differential execution, find_json_escape signed-compare (split from #191)
Split out from #191 — collects the test-coverage work that specifically targets the **x86_64 SIMD backends** (SSE2 / AVX2 / BMI2) and therefore **requires x86 hardware / CI to develop or verify**. The primary dev host is aarch64, so these paths are `cfg`-gated out or self-skip locally, and a green `cargo test` on ARM says nothing about them.

The companion aarch64 / SVE2 / scalar and non-SIMD coverage stays in #191.

## x86 CI leg (from #191 meta-hazard)

- [ ] Add/enable an x86_64 CI job that runs the SIMD suites with **BMI2 + AVX2** available (and ideally a second config with AVX2 disabled to exercise the SSE2 fallback in `src/dsv/simd/sse2.rs` and `src/yaml/simd/x86.rs`). Without this leg the differential/fuzz tests authored in #191 never execute their SSE2/AVX2/BMI2 branches.
- [ ] Make the x86 hardware self-skips visible — `if !has_bmi2() / has_avx2() / has_sse2() { return; }` in `src/dsv/simd/{bmi2,avx2,sse2}.rs` and `src/util/simd/x86.rs` — print a `SKIPPED` line / count skips so a fully-skipped x86 run doesn't read as "passed". (Shares the visibility mechanism tracked in #191.)

## x86 backend execution of the DSV differential/fuzz tests (blocks #149, with #182)

The cross-backend differential + property/fuzz tests are authored in #191. This issue tracks confirming, on x86 CI, that they actually exercise and pass:

- [ ] SSE2, AVX2 and BMI2 each produce an index identical to scalar for a quote at **every** offset `0..64` (the bit-63 / #149 regression), and for randomized chunk-spanning CSV (quotes, embedded delimiters/newlines).

## YAML → JSON x86 escape scanning (blocks #150)

- [ ] Regression test for the x86 `find_json_escape` signed-compare bug: run the **library** `to_json()` / `to_json_document()` path on string values ≥ 16 bytes containing non-ASCII (`café`, CJK, emoji) and assert the multibyte bytes survive. The bug is in `find_json_escape_avx2` / `find_json_escape_sse2` (`src/yaml/simd/x86.rs:824`, `:911`), so the assertion only fails on x86 and needs the x86 CI leg above to be meaningful. (The NEON over-match counterpart, #186, stays in #191.)

---
_Split from #191. Cross-refs: #149 #150 #182 #186._

newhoggy:
## Description

This PR addresses issue #193 by implementing comprehensive SIMD test coverage infrastructure for x86 architectures. The work ensures that x86 SIMD test suites (AVX2, BMI2, SSE4.2, POPCNT) provide visible skip indicators when CPU features are unavailable, and adds a hard contract test that prevents silent feature degradation on CI runners.

### Problem Solved

Previously, x86 SIMD tests would silently self-skip on unsupported hardware without any visible indication, causing test suites to appear fully passing when they actually ran zero assertions. On CI, a runner-fleet change that removed a required CPU feature would go undetected. Additionally, per-kernel differential tests were missing for the YAML JSON escape scanners, and an SSE2 chunk-width accounting bug in the parser was lurking undetected on non-AVX2 hardware.

### Solution Overview

This PR implements a multi-layered approach:

1. **Visible Skip Indicators**: Routes all x86 SIMD guards through `note_simd_skip()` helper to emit visible `SKIPPED` lines instead of silent returns
2. **Hard Feature Expectations**: Adds `SUCCINCTLY_EXPECT_SIMD` environment variable test that fails CI if expected features become unavailable
3. **Per-Kernel Differential Tests**: Adds direct kernel calls to exercise SSE2 and AVX2 independently, catching issues regardless of dispatcher choices
4. **Bug Fix**: Corrects SSE2 chunk-width accounting in the unquoted-scalar fast path that was swallowing structural characters

## Type of Change
- [x] Test coverage improvement
- [x] CI/CD changes
- [x] Documentation update

## Related Issue
Fixes #193
Relates to #191, #150, #230

## Changes Made

**Test Infrastructure & Skip Visibility:**
- Created `note_simd_skip()` helper in `src/util/simd/mod.rs` to emit `SKIPPED` lines for unavailable CPU features
- Added `has_avx2()`, `has_bmi2()`, and `has_popcnt()` detection guards in `src/dsv/simd/avx2.rs`, `src/dsv/simd/bmi2.rs`, and `src/util/simd/x86.rs` that route through `note_simd_skip()`
- Removed `#[allow(dead_code)]` from `note_simd_skip()` now that x86 sites use it
- Updated `tests/simd_level_tests.rs` to detect backends once per test instead of per-assertion (eliminating 3200+ duplicate skip lines in differential fuzz tests)
- Updated `tests/dsv_simd_differential_tests.rs` to cache backend detection, reducing redundant feature checks

**Hard Expectation Test:**
- Added `tests/simd_expectation_tests.rs` with `test_expected_simd_features_are_detected()` that validates CPU features listed in `SUCCINCTLY_EXPECT_SIMD` environment variable
- Supports x86_64 features: `sse2`, `sse4.2`, `avx2`, `bmi2`, `popcnt`
- Supports aarch64 features: `neon`, `sve2`, `sve2-bitperm`
- Unknown feature names fail the test (prevents silent typos)

**Per-Kernel Differential Tests:**
- Added exhaustive per-kernel tests in `src/yaml/simd/x86.rs` that call `find_json_escape_sse2()` and `find_json_escape_avx2()` directly
- Added byte-by-byte exhaustive sweep tests for both SSE2 (40-byte buffer) and AVX2 (56-byte buffer) kernels
- Added multibyte UTF-8 regression tests in `src/yaml/light.rs` for `to_json()`, `to_json_document()`, and streaming paths
- Added yq CLI regression test in `tests/yq_cli_tests.rs` for the original #230 repro case

**Parser Bug Fix:**
- Modified `YamlCharClass` struct in `src/yaml/simd/x86.rs` to include `width` field reporting classified bytes (32 for AVX2, 16 for SSE2)
- Updated `classify_yaml_chars_avx2()` and `classify_yaml_chars_sse2()` to set the width field
- Fixed `skip_unquoted_simd()` in `src/yaml/parser.rs` to skip exactly the classified width instead of assuming 32 bytes
- Added regression test in `src/yaml/light.rs` for the plain-scalar folding bug

**Documentation:**
- Documented `SUCCINCTLY_EXPECT_SIMD` in `docs/reference/environment-variables.md` with per-target feature names and CI usage example
- Added comprehensive comments explaining the skip visibility and expectation contract

**CI Configuration:**
- Added `SUCCINCTLY_EXPECT_SIMD=sse2,sse4.2,avx2,bmi2,popcnt` environment variable to x86_64 test jobs in `.github/workflows/ci.yml`
- Added CPU feature detection step that prints available features at job start
- Applied same expectation to both stable and beta/nightly test jobs

## Testing

**Automated Testing:**
- [x] All existing tests pass
- [x] New skip visibility tests in all SIMD modules
- [x] New hard expectation test validates CI runner feature contract
- [x] Per-kernel differential tests for YAML escape scanners (exhaustive byte sweeps + multibyte UTF-8)
- [x] Regression tests for SSE2 parser chunk-width bug
- [x] yq CLI regression test for multibyte content
- [x] Verified on x86_64 (stable), aarch64 (neon/sve2 feature detection), and Rosetta 2

**Manual Testing:**
- [x] Tested on x86_64 with AVX2
- [x] Tested on aarch64 with NEON
- [x] Tested on Rosetta 2 (x86 without AVX2) — exposed the parser SSE2 bug
- [x] Verified skip lines appear in test output when features unavailable
- [x] Verified `SUCCINCTLY_EXPECT_SIMD` fails when features are missing
- [x] Verified `SUCCINCTLY_EXPECT_SIMD` fails when unknown feature names are used

### Test Commands
```bash
# Run all tests
cargo test

# Run SIMD-specific tests
cargo test simd

# Run parser tests
cargo test yaml

# Test feature expectation (with AVX2)
SUCCINCTLY_EXPECT_SIMD=sse2,sse4.2,avx2,bmi2,popcnt cargo test simd_expectation

# Verify no warnings

# Check formatting
cargo fmt --check
```

## Performance Impact
- [x] No performance impact
  - Skip detection happens only during test initialization, not in hot paths
  - Backend caching in differential tests reduces redundant feature checks
  - Parser bug fix improves correctness with no overhead change

## Checklist
- [x] CI configuration updated with feature expectations

### Why This Matters for CI

GitHub's x86_64 runners always have AVX2 available, so the dispatcher never reaches the SSE2 kernel in CI. Without per-kernel differential tests, the SSE2 code path remained untested by CI and the chunk-width bug went undetected for a long time. By adding exhaustive per-kernel tests (byte-by-byte sweeps), we now catch regressions in both code paths regardless of CPU capabilities. The hard expectation test adds a second line of defense: if a runner-fleet change removes a required feature, the leg fails instead of silently skipping.

### Architecture-Specific Notes

- **x86_64**: Tests exercise SSE2 (baseline), SSE4.2, AVX2, BMI2, and POPCNT paths
- **aarch64**: Tests exercise NEON (baseline) and SVE2 paths (if available)
- **Other targets**: SIMD tests are no-ops (scalar paths only)

The expectation test confirms the CI environment matches what the code was compiled and tested for, preventing silent regressions from infrastructure changes.

---

### Verified on genuine x86 hardware

Ran the branch (Rust content identical to PR head `5b5c8b49`; local `89e09b34` differs only by the #246 links.yml/.gitignore base commits) on an **AMD Ryzen 9 7950X** (Zen 4 — real AVX2/BMI2/AVX-512), in a dedicated worktree with the SHA asserted before testing:

- **27/27 test suites green** on default features, `--features simd`, and `--features portable-popcount`; cli-gated `yq_cli_tests` + `yq_golden_tests` green.
- **Zero `SKIPPED` lines** with `SUCCINCTLY_EXPECT_SIMD=sse2,sse4.2,avx2,bmi2,popcnt` — every SIMD test asserted rather than skipped, and the expectation test reported all five features detected.

### New finding: #149 confirmed on genuine BMI2 hardware

Running the `#[ignore]`d repro on the 7950X:

```
cargo test --test dsv_simd_differential_tests -- --ignored
```

fails at exactly **`bmi2/open=63`** (“index differs from scalar”) — the first hardware confirmation that the toggle64 PDEP bit-63 carry bug (#149) is real and that this PR’s differential detects it. Un-ignoring that test after #149’s fix will therefore be a meaningful gate, not a formality.

### Rosetta 2 as the missing “AVX2-disabled” configuration

Rosetta 2 on this dev host exposes SSE2/SSE4.2/POPCNT but **not** AVX2/BMI2 — the only environment available that exercises the x86 SSE2 *dispatch* branches (CI runners all have AVX2). That is how the `skip_unquoted_simd` chunk-width bug in this PR’s fix commit was found: the multibyte `to_json()` tests failed there on **pure-ASCII** long scalars, isolating the fault to chunk-width accounting rather than UTF-8 handling. After the fix, the affected suites pass under Rosetta with visible `SKIPPED … [avx2]/[bmi2]` lines for the guarded kernels.

### Review note: overlapping CI steps after the #192 rebase

The rebase onto #192’s scaffolding leaves `test-x86` with two CPU-feature steps: this PR’s log-only **“Check CPU features (x86 SIMD)”** and #192’s enforcing **“Verify SIMD CPU features (BMI2 + AVX2)”**. The log-only step is now redundant and can be dropped in a follow-up (or on this branch before merge). The `SUCCINCTLY_EXPECT_SIMD` test is **not** redundant with #192’s shell check: it enforces through the same `is_x86_feature_detected!` path the tests actually use (a `/proc/cpuinfo` grep can disagree with Rust’s detection), covers all five listed features rather than just AVX2/BMI2, and also works on the ARM legs and in local runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,9 +56,30 @@ jobs:
     name: Test (x86_64)
     needs: gate
     runs-on: ubuntu-latest
+    env:
+      # x86 SIMD CI leg (#193): the SIMD suites self-skip (with a visible
+      # `SKIPPED` line) when a CPU feature is missing, so this pins the runner
+      # contract — tests/simd_expectation_tests.rs fails the leg if any listed
+      # feature stops being detected, instead of the suites silently skipping.
+      # The SSE2 kernels don't need an AVX2-disabled config: the differential
+      # tests (tests/simd_level_tests.rs, tests/dsv_simd_differential_tests.rs,
+      # src/yaml/simd/x86.rs) call each kernel directly, so every level is
+      # exercised on this AVX2 runner. There is deliberately no x86 dispatch
+      # override (docs/reference/environment-variables.md).
+      SUCCINCTLY_EXPECT_SIMD: sse2,sse4.2,avx2,bmi2,popcnt
     steps:
       - uses: actions/checkout@v4
         if: needs.gate.outputs.code == 'true'
+
+      - name: Check CPU features (x86 SIMD)
+        if: needs.gate.outputs.code == 'true'
+        run: |
+          echo "=== CPU Info ==="
+          lscpu | grep -E "Model name|Architecture|CPU" || true
+          echo "=== x86 SIMD Detection (expecting $SUCCINCTLY_EXPECT_SIMD) ==="
+          for feature in sse2 sse4_2 avx2 bmi2 popcnt; do
+            if grep -qw "$feature" /proc/cpuinfo; then echo "$feature: supported"; else echo "$feature: not detected"; fi
+          done
 
       - name: Install Rust
         if: needs.gate.outputs.code == 'true'
@@ -252,6 +273,9 @@ jobs:
       matrix:
         rust: [beta, nightly]
     continue-on-error: ${{ matrix.rust == 'nightly' }}   # beta gates CI; nightly is advisory
+    env:
+      # Same x86 SIMD runner contract as test-x86 (#193).
+      SUCCINCTLY_EXPECT_SIMD: sse2,sse4.2,avx2,bmi2,popcnt
     steps:
       - uses: actions/checkout@v4
 
--- a/docs/reference/environment-variables.md
+++ b/docs/reference/environment-variables.md
@@ -15,6 +15,7 @@ Every environment variable succinctly reads, what it accepts, and what it change
 | [`JQ_LIBRARY_PATH`](#jq_library_path)                     | `succinctly jq`                  | `:`-separated directories      | Only `-L` paths and `~/.jq`   |
 | [`HOME`](#home)                                           | `succinctly jq`                  | A directory path               | No `~/.jq` auto-loading       |
 | [`TZ`](#tz)                                               | Library (jq date builtins)       | POSIX `STDoffset[DST]`         | UTC                           |
+| [`SUCCINCTLY_EXPECT_SIMD`](#succinctly_expect_simd)       | Test suite (`cargo test`)        | Comma-separated CPU features   | Expectation check skipped     |
 
 Queries can also read **any** variable through the [`env` builtins](#reading-the-environment-from-a-query).
 
@@ -86,6 +87,34 @@ TZ=EST5EDT succinctly jq -nc '1700000000 | localtime' # [2023,10,14,17,13,20,2,3
 
 For anything requiring true local time, set an explicit numeric offset rather than a zone name, and
 change it yourself across DST boundaries.
+
+## Test-Only Variables
+
+### `SUCCINCTLY_EXPECT_SIMD`
+
+Read only by the test suite ([`tests/simd_expectation_tests.rs`](../../tests/simd_expectation_tests.rs));
+the library and CLI never look at it.
+
+Pins the set of CPU features a `cargo test` run is expected to detect. The SIMD test suites self-skip
+when the running CPU lacks a feature (printing a `SKIPPED` line), which keeps local runs green on any
+hardware — but on CI it would let a runner-fleet change silently skip entire suites. Setting this
+variable turns those soft skips into a hard failure: every listed feature must be runtime-detected or
+`test_expected_simd_features_are_detected` fails the run (#193).
+
+The value is a comma-separated list of runtime feature names:
+
+| Target    | Recognized names                          |
+|-----------|-------------------------------------------|
+| `x86_64`  | `sse2`, `sse4.2`, `avx2`, `bmi2`, `popcnt` |
+| `aarch64` | `neon`, `sve2`, `sve2-bitperm`             |
+
+Unknown names — including names for the *other* architecture — fail the test, so a typo cannot
+silently satisfy the expectation. Unset, the check is skipped entirely.
+
+```bash
+# What CI sets on the x86_64 test legs (.github/workflows/ci.yml)
+SUCCINCTLY_EXPECT_SIMD=sse2,sse4.2,avx2,bmi2,popcnt cargo test --test simd_expectation_tests
+```
 
 ## CLI Variables
 
--- a/src/dsv/simd/avx2.rs
+++ b/src/dsv/simd/avx2.rs
@@ -170,11 +170,7 @@ mod tests {
     /// Detection guard for the AVX2 backend; emits a visible `SKIPPED` line
     /// when unavailable so a fully-skipped run doesn't read as green (#193).
     fn has_avx2() -> bool {
-        let available = is_x86_feature_detected!("avx2");
-        if !available {
-            crate::util::simd::note_simd_skip("avx2");
-        }
-        available
+        crate::util::simd::note_simd_skip_unless(is_x86_feature_detected!("avx2"), "avx2")
     }
 
     #[test]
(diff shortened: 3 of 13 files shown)


newhoggy:
Addressed the patch-coverage gaps in ba99ad9f. The 9 uncovered new lines fell into two categories:

**Skip-note branches (6 lines — now covered).** The `note_simd_skip(...)` call sites inside the `has_avx2()`/`has_bmi2()`/`has_popcnt()` guards only execute when the CPU *lacks* the feature, so the AVX2+BMI2 CI runners can never reach them — only the manual Rosetta runs do, where coverage isn't collected. The branch now lives in one shared `note_simd_skip_unless(available, feature)` helper with a test exercising both arms, so each guard is a single always-executed line. Skip behavior is unchanged (re-verified under Rosetta on a non-AVX2 host: same `SKIPPED` lines).

**Inherently negative paths (the remainder — intentionally uncovered).** `src/yaml/simd/x86.rs` 1436/1444/1446 are the `assert_eq!` failure-message arms of the differential tests (covered only if a kernel diverges from scalar), and 1470 is the early `return;` in the AVX2 exhaustive test's guard (covered only on non-AVX2 hardware). These are the lines whose *non*-execution is the passing state.

Issue #106, opened 2026-02-01
perf: investigate seq_item detection optimizations after bitvector elimination


newhoggy:
## Context

PR #[TBD] eliminated the `seq_items` bitvector from `YamlIndex`, replacing O(1) bitvector lookups with runtime text analysis. This reduces memory but introduced a 7-15% performance regression in CLI benchmarks.

**Commit**: `b99d4e4` on `eliminate-seq-items-bitvector` branch

## Current State

After implementing optimization #1 (branchless whitespace check), we recovered 2-7% performance:

| Pattern | Before Opt | After Opt |
|---------|------------|-----------|
| sequences/10 | +2.1% regression | **-6.1% improvement** |
| simple_kv/10 | +1.7% regression | **-7.1% improvement** |
| yq comprehensive | +7-15% regression | ~neutral to -5% |

### Option 1: Branchless Whitespace Check ✅ IMPLEMENTED

**Status**: Done in current branch

Simplified the seq_item text check from nested branches to single condition:

```rust
// Before: 3 branches
if self.text[text_pos] == b'-' {
    let is_seq_item = if text_pos + 1 >= self.text.len() {
        true
    } else {
        let next = self.text[text_pos + 1];
        next == b' ' || next == b'\t' || next == b'\n' || next == b'\r'
    };
}

// After: 1 branch
if self.text[text_pos] == b'-'
    && matches!(
        self.text.get(text_pos + 1).copied().unwrap_or(b' '),
        b' ' | b'\t' | b'\n' | b'\r'
    )
```

**Result**: 2-7% improvement across benchmarks.

---

### Option 2: Encode seq_item in High Bit of Position

**Status**: Not implemented

Since YAML files are typically < 2GB, text positions fit in 31 bits. We could steal the high bit to encode seq_item status:

```rust
// In parser: encode seq_item flag in position
fn mark_position(text_pos: u32, is_seq_item: bool) -> u32 {
    if is_seq_item { text_pos | 0x8000_0000 } else { text_pos }
}

// In lookup: O(1) extraction without text access
fn is_seq_item_from_pos(encoded: u32) -> bool {
    encoded & 0x8000_0000 != 0
}

fn text_pos_from_encoded(encoded: u32) -> u32 {
    encoded & 0x7FFF_FFFF
}
```

**Pros**:
- O(1) lookup like original bitvector
- Zero extra memory
- No text access needed

**Cons**:
[...]

Issue #133, opened 2026-02-06
perf(text): Broadword (SWAR) optimizations for scalar UTF-8 validation


newhoggy:
## Summary

Add broadword (SWAR - SIMD Within A Register) optimizations to the scalar UTF-8 validator to improve throughput while remaining portable and not using SIMD intrinsics.

## Current Performance (AMD Ryzen 9 7950X)

| Pattern | Throughput (MiB/s) |
|---------|-------------------|
| ASCII | 2,261 |
| log_file | 2,137 |
| source_code | 1,979 |
| mixed | 1,819 |
| CJK | 1,359 |
| emoji | 1,036 |

### 1. 8-byte ASCII Fast Path (Highest Impact)

Current implementation processes byte-by-byte. With broadword, we can check 8 bytes at once:

```rust
let word = u64::from_le_bytes(chunk);
if (word & 0x8080_8080_8080_8080) == 0 {
    // All 8 bytes are ASCII, skip ahead
    pos += 8;
    continue;
}
```

This reduces loop iterations by 8x for ASCII runs.

### 2. Broadword Newline Counting

For line tracking in error reporting, count newlines in parallel:

```rust
const LO: u64 = 0x0101_0101_0101_0101;
const HI: u64 = 0x8080_8080_8080_8080;
const NEWLINE: u64 = 0x0a0a_0a0a_0a0a_0a0a;

fn count_newlines_word(word: u64) -> u32 {
    let xored = word ^ NEWLINE;
    let mask = xored.wrapping_sub(LO) & !xored & HI;
    mask.count_ones() / 8
}
```

### 3. Non-ASCII Position Detection

Find first non-ASCII byte in a word:

```rust
fn find_first_non_ascii(word: u64) -> Option<usize> {
    let high_bits = word & 0x8080_8080_8080_8080;
    if high_bits == 0 {
        None
    } else {
        Some((high_bits.trailing_zeros() / 8) as usize)
    }
}
```

### 4. Continuation Byte Validation

Check if bytes are valid continuation bytes (`0x80..=0xBF`) in parallel:

```rust
fn all_continuations(word: u64) -> bool {
    let high_bits = word & 0x8080_8080_8080_8080;
    let second_bits = word & 0x4040_4040_4040_4040;
    high_bits == 0x8080_8080_8080_8080 && second_bits == 0
}
```

## Projected Impact

| Pattern | Current (MiB/s) | Projected (MiB/s) | Improvement |
[...]

PR #322, opened 2026-07-25
perf(yaml): consolidate seq-item detection and measure alternatives for O5/O6

newhoggy:
# perf(yaml): consolidate seq-item detection and measure alternatives

Consolidates sequence-item wrapper detection across five sites with three different predicates into a single definition, and provides measured analysis of two rejected alternatives.

## Summary

Issue #106 asked whether O(1) seq-item lookup should be restored after the removal of the `seq_items` bitvector in O4. Two mechanisms were proposed; analysis measures both and rejects them, but the investigation uncovers a much larger win by consolidating the detection predicate that diverged across three `YamlElements` methods.

## Changes Made

**Core Optimization (O6 - Seq-Item Predicate Consolidation):**
- Extracted `is_seq_item_at(text, text_pos)` as the single definition of "does this `-` open a sequence item?" in `src/yaml/index.rs`
- Widened the predicate to accept `\n`, `\r` and end-of-input, matching the two branchless forms in `YamlIndex::is_seq_item` and `YamlCursor::value`
- Routed all five detection sites through the single definition:
  - `YamlIndex::is_seq_item` — reduced to a wrapper
  - `YamlCursor::value()` — uses shared predicate
  - `YamlElements::uncons`, `uncons_cursor`, `get` — all call `is_seq_item_at`
- Added comprehensive tests verifying all three access paths (`uncons`, `get`, `uncons_cursor`) agree on the same sequence across diverse shapes, preventing future divergence
- Root cause: the three narrower predicates rejected `\n`/`\r`/EOI, so `-\n` items fell through to `value()`, which re-read the same open index. That backwards jump into `get_random` resets the `AdvancePositions` IB scan cursor to word 0, forcing the next read to rescan the bitmap from the start — **O(N·L/64)** per document

**Benchmark Coverage (Added `seqwrap` pattern):**
- Every existing sequence-bearing generator emitted `- ` (dash-space), leaving bare-dash items unmeasured
- Added `YamlPattern::SeqWrap` and `generate_seq_wrap()` in `yaml_generators.rs`, structurally identical to `users` except `-\n` instead of `- `
- Registered in `main.rs` and `yq_bench.rs` pattern tables
- This shape had no benchmark coverage, which is why the divergence remained hidden

**Analysis of Alternatives (O5 - Rejected):**
- Created `seq_item_stats` module with `succinctly dev bench seq-item-stats` command
- Measured seq-item density across real-workload corpus (6 files, 1296 nodes) and YAML Test Suite (279 valid cases)
- **Option 2 (high-bit tag)**: requires `OpenPositions::Dense` path, which is used by **0 of 6 real files** (only 5 of 279 suite cases). `Compact` representation has no per-open `u32` to tag. **Rejected.**
- **Option 3 (sparse index)**: The issue's stated 12.5% break-even is wrong by 4× (correct: 3.125% vs measured 8.87% aggregate, 21.43% max). Baseline is 0 bytes (O4 deleted the structure), so this is pure memory addition costing 68–148% of what O4 saved. **Rejected.**
- Verified two correctness invariants: predicate never false-positives and never misses wrappers with content; both are zero across entire corpus
- Documented context-dependent wrapper emission: wrapper-ness cannot be cached per sequence

**Documentation Updates:**
- Updated `CLAUDE.md` with O5 rejection and O6 acceptance in optimization summary
- Corrected `yaml-semi-indexing` skill documentation: `seq_items` bitvector no longer exists; wrapper emission is context-dependent
- Added O5 and O6 sections to `docs/optimizations/history.md` with measured numbers and decision rationale
- Added comprehensive O5/O6 analysis to `docs/parsing/yaml.md` with break-even arithmetic, density measurements, and key learnings
- Created `docs/benchmarks/seq-item-density.md` with measured density tables and verdict on both options
- Updated `docs/benchmarks/README.md` to link seq-item density report
- Re-measured O6 results on both designated benchmark machines (Apple M4 Pro, AMD Ryzen 9 7950X) with interleaved A/B and thermal idle verification

## Performance Impact

**O6 Results** (min of 15 alternating reps, both machines idle-verified):

Apple M4 Pro:
- seqwrap/1mb: **1.62×** faster
- bare-dash 1MB: **2.19×** faster
- bare-dash 4MB: **6.12×** faster
- users/sequences/nested: 1.00×–1.00× (neutral)

AMD Ryzen 9 7950X:
- seqwrap/1mb: **2.97×** faster
- bare-dash 1MB: **4.77×** faster
- bare-dash 4MB: **16.37×** faster
- users/sequences/nested: 0.99×–1.00× (neutral)

Speedup grows with file size (signature of removing an O(N·L) term). Zen 4 was hurt ~3× more pre-fix (2239 ms vs 758 ms at 4MB) because the pathology is a cache-thrashing IB rescan; post-fix times are comparable (136.8 ms vs 123.9 ms).

## Testing

**Automated Testing:**
- [x] All existing tests pass
- [x] Added `test_is_seq_item_at_predicate` for the single definition
- [x] Added `test_dash_newline_items_agree_across_paths` — the shape that triggered the pathology
- [x] Added 13 agreement tests ensuring `uncons`/`get`/`uncons_cursor` always agree
- [x] Added characterization tests for context-dependent wrapper emission and pre-existing CRLF defect
- [x] Verified O5 measurements with predicate correctness invariants (false positives: 0, undetected wrappers: 0)
- [x] Seq-item stats verifies two correctness invariants across entire corpus

**Manual Testing:**
- [x] Tested on both M4 Pro and Ryzen 9 7950X
- [x] Output byte-identical before/after on 48 configurations per machine (12 inputs × 4 queries)
- [x] All configurations still match system `yq`
- [x] Neutral performance on shapes that don't hit the divergent path

### Test Commands
```bash
cargo test --all
cargo test --bin succinctly -- --include-ignored
cargo fmt --check
succinctly dev bench seq-item-stats --data-dir data/bench/corpus/yaml
succinctly yq -- '.[]' <seqwrap pattern file>
```

## Checklist
- [x] My code follows the project's style guidelines (rustfmt applied)
- [x] I have performed a self-review of my own code
- [x] I have added tests that prove the fix is effective and the feature works
- [x] New and existing tests pass locally with my changes
- [x] I have updated documentation (YAML optimization guide, skill docs, benchmark report)
- [x] My changes generate no new warnings (no clippy warnings, applied rustfmt)
- [x] Outputs are byte-identical before/after on all test configurations

## Key Insights

1. **Duplicated predicates diverge silently** — three independent implementations, three behaviours, one quadratic. The fix is one definition plus tests that access paths agree.
2. **Stateful sequential cursors change redundant reads into complexity changes** — re-reading the same index with `AdvancePositions` resets the IB scan state, causing O(L/64) overhead per reread.
3. **An outdated comment can hide a shape from every benchmark** — the "parser requires `- `" claim kept bare-dash items out of the entire generated suite despite YAML specs and parser support.
4. **Measure where the data physically lives before proposing to tag it** — Option 2 was framed as zero memory, but 100% of real files use `Compact`, which has no per-open `u32`.
5. **Re-derive break-evens before trusting them** — the 12.5% break-even drops a bits-to-bytes conversion; the actual gate is 3.125%, and measured density exceeds it by 2.8×.

## Additional Notes

Refs #106. Both O5 rejection and O6 acceptance follow the documented analysis-stage process used for P5–P8. O6 was never proposed in the issue; it emerged from investigation of why O4's branchless form was applied to only two sites while three others diverged.


newhoggy:
Closing as stale/superseded: its issue #106 closed 2026-07-26, and the seq-item detection work this branch contains already landed on main under different branches (PR #104, PR #240) before this branch was last updated. This PR is now conflicting with main and duplicates already-merged work — same pattern as the #133/#134 duplicate-worktree case earlier rounds flagged. No action needed; closing without merging.

Issue #372, opened 2026-07-26
yaml: an alias to an unknown anchor silently yields null instead of erroring


newhoggy:
Split out of #171, which records this as a "related low-severity alias bug" alongside the merge-key
work:

> forward-reference aliases resolve to `null` because an anchor is recorded only if already seen
> (`parser.rs:~3239`): `a: *x` before `b: &x 5` yields `.a == null` rather than an error.
> **Fix:** […] decide forward-reference handling (error vs defer).

#171 stays open for `<<:` merge-key resolution, which is the larger, riskier half. This issue is
only the alias-resolution failure, and it is broader than the forward-reference framing suggests:
**any** alias naming an anchor that is not in scope silently becomes `null`.

## Current behaviour

Checked at `2a3a1cae` against `yq` v4.53.3 (the goldens' pinned version). All six positions where
an alias can appear diverge; `yq` errors on every one with
`unknown anchor 'x' referenced`:

| position | input | succinctly | yq |
|----------|-------|-----------|-----|
| value, forward ref | `a: *x`⏎`b: &x 5` | `{"a":null,"b":5}` | error |
| value, never defined | `a: *nope` | `{"a":null}` | error |
| mapping key | `*nope: v` | `{"":"v"}` | error |
| flow sequence | `a: [*nope]` | `{"a":[null]}` | error |
| flow mapping value | `a: {k: *nope}` | `{"a":{"k":null}}` | error |
| block sequence item | `- *nope` | `[null]` | error |

Valid aliases are unaffected and already agree with `yq`:

```console
$ printf 'b: &x 5\na: *x\n'   | syq -o json -I0 '.'   # {"b":5,"a":5}   (same as yq)
$ printf '&x k: 1\n*x: 2\n'   | syq -o json -I0 '.'   # {"k":1,"k":2}   (same as yq)
```

## Cause

Two resolution sites both treat a lookup miss as "no alias", and the node is then left with
nothing to resolve to, which renders as `null`:

`src/yaml/parser.rs:3430` (`parse_alias`, covering values, flow and sequence items):

```rust
if let Some(&target_bp_pos) = self.anchors.get(&name) {
    self.aliases.insert(alias_bp_pos, target_bp_pos);
}
// Note: If anchor not found, we don't record it.
// Forward references (alias before anchor) are not supported.
```

`src/yaml/parser.rs:1682` (alias-as-key in a block mapping) has the same `if let` with no `else`.

## Expected

[...]

Issue #328, opened 2026-07-25
yaml: anchor on a sequence item with a collection value silently corrupts the document


newhoggy:
An anchor attached to a block sequence item whose value is a collection is mis-parsed: the mapping is dropped and its key leaks out as a sibling. No error is raised.

## Repro

```yaml
list:
  - &m
    k: v
  - *m
```

```console
$ succinctly yq -o=json -I=0 . repro.yaml
{"list":["k"],"v":["k"]}

$ yq -o=json -I=0 . repro.yaml          # mikefarah/yq v4.53.3
{"list":[{"k":"v"},{"k":"v"}]}
```

The flow form is corrupted differently:

```yaml
items:
  - &first {id: 1}
  - *first
```

```console
$ succinctly yq -o=json -I=0 . repro2.yaml
{"items":[{"":"1}"},null]}

$ yq -o=json -I=0 . repro2.yaml
{"items":[{"id":1},{"id":1}]}
```

## What works

Anchors are fine everywhere else tested:

- on a mapping value — `key: &a` then an indented mapping
- on a scalar sequence item — `- &s value`
- aliases in any position, including inside flow collections

So the defect is specific to `- &name` where the anchored node is a collection.

## Severity

Silent structural corruption, not a parse error: `"v"` becomes a top-level key and the alias resolves to the wrong node. Callers get a well-formed but wrong document.

## Context

Found while adding an `anchors` pattern to the generated benchmark suite (#327). The generator deliberately avoids this form and documents why, so the benchmark data does not bake in a wrong shape — see the "Deliberately absent" list in `src/bin/succinctly/yaml_generators.rs`. `tests/yaml_bench_suite_coverage.rs` will need an entry moved from out-of-scope to required once this is fixed.

newhoggy:
**Real-world blast radius confirmed by the #326 survey** (merged via #331).

This silent-corruption bug triggers on anchored sequence items, and anchors are **not** an edge case: `docs/benchmarks/yaml-shape-survey.md` measured anchors/aliases in **0.494%** of 33,575 real YAML files, rising to **4–11%** in Home Assistant, Salt, Concourse, and GitLab CI. So "anchor on a sequence item with a collection value" is a mainstream shape in the ecosystems that use anchors, and silently dropping the mapping (with its key leaking out as a sibling) will corrupt real documents with no error.

[...]

Issue #153, opened 2026-07-09
YAML alias resolution: no cycle guard causes infinite recursion abort


newhoggy:
**Severity: High (DoS)** — confirmed (exit 134).

`src/yaml/light.rs:1880,2550` — alias materialization follows `target_cursor.write_json_to()` / `.stream_json_value()` with no visited-set / cycle guard, so a self-referential anchor recurses forever.

**Failing input:** `a: &anchor\n  self: *anchor` then query `.a.self.self` → stack-overflow abort.

**Fix:** detect alias cycles (anchor references an ancestor) at resolve time, or bound recursion depth.

---
_Filed from an automated full-code-review pass. Severity and verification status noted above._

PR #408, opened 2026-07-27
docs(docs): record the unknown-anchor rejection, and when build fails

### LINKED ISSUE #403: docs(yaml): limitations.md still says a cyclic alias is the loader's only structural check
Found while reviewing the #372 branch.

[`docs/compliance/yaml/limitations.md:112-116`](https://github.com/rust-works/succinctly/blob/main/docs/compliance/yaml/limitations.md#L112-L116) reads:

> ### One exception: cyclic aliases are rejected
>
> An alias that would make an anchored value contain itself (`a: &x {self: *x}`) is
> rejected at index build with `YamlError::AliasCycle` ("anchor value contains itself").
> **This is the one structural check the non-validating loader performs**, because a cyclic
> document cannot be materialized at all […]

Since #372 there are **two**: `AliasCycle` and the new `YamlError::UnknownAnchor`. The #372
CHANGELOG entry even draws the parallel explicitly ("it is now refused at build time, as a
cyclic alias always has been") without updating the section that states the parallel is unique.

The whole section frames the loader as lenient-by-design with exactly one carve-out, so a
reader currently comes away with the wrong model of when `YamlIndex::build` can fail.

## Fix

Retitle the section to cover both (e.g. "Two exceptions: unresolvable aliases are rejected")
and add the unknown-anchor case with its rationale — an alias must name a *previous* anchor
(YAML 1.2 §7.1), so a miss is invalid input rather than a value, and `yq` rejects the same
input.

## Not affected

The document counts in that file are unchanged. I scanned all 402 YAML Test Suite cases: none
contains an alias to an out-of-scope anchor, so the `82` accepted-but-invalid / `58`
validator-rejected figures and the `lax:anchors` rows still hold.

newhoggy:
Closes #403. Follows #380 (`0f3fae16`), which added `YamlError::UnknownAnchor` without updating this page.

`docs/compliance/yaml/limitations.md` said `AliasCycle` was "the one structural check the non-validating loader performs". Since #372 there are two, and #380's CHANGELOG entry drew the parallel explicitly — "it is now refused at build time, as a cyclic alias always has been" — while leaving the sentence that asserts the parallel is unique.

## What changed

**Retitled to cover both**, presented as one failure class (an alias with nothing to resolve to), with verbatim output for each and a resolvable alias alongside so the boundary is visible rather than implied.

**A new subsection, "The other ways `YamlIndex::build` fails"**, since the old framing left readers with the wrong model of when `build` can fail at all — not just which structural check exists. Eleven rows: absent feature, bytes the parser cannot tokenize, and the index's own ceilings, cross-linked to `docs/reference/limits.md` for the 4 GiB one.

Every row was checked against the built binary rather than read off the variant names, which was worth doing — two plausible-looking candidates (`a: [1 2]`, a bare `a` line) turned out to parse fine, and the real `UnexpectedCharacter` trigger makes the page's own point better than a made-up one would:

```
$ printf 'a: [[1] 2]\n' | succinctly yq '.'
Error: YAML parse error: unexpected character '2' at offset 8: expected ',' or ']' in flow sequence
$ printf 'a: [1 2]\n' | succinctly yq -o json -I0 '.'
{"a":["1 2"]}             # same missing comma, absorbed
```

The loader stops where it cannot continue, not where a rule is broken — which is why 11 `lax:flow` cases survive it.

## Numbers

Unchanged, and the page now says why: no corpus case contains an alias to an out-of-scope anchor, so 12/94, the 82 accepted-but-invalid documents and the `lax:anchors` row are unmoved by #372 — matching #380's own "no manifest movement" finding.

`cargo test --test yaml_test_suite -- --nocapture` reports 215/279 = 77.1%, 70/94 = 74.5%, 27/29 = 93.1%, 90 known failures: exactly what the page states.

## Scope

Docs only — one file, one hunk. No code, no CHANGELOG (#380 carries the behaviour entry), no manifest change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -515,8 +515,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `yq -R -s` now yields the entire input as a single string instead of an array of per-line strings, matching jq and `jq -R -s` (#271)
 - YAML alias cycles (`a: &anchor {self: *anchor}`) are rejected at index build with the
   new `YamlError::AliasCycle` variant instead of aborting with a stack overflow when the
-  value is materialized (#153). Matches `yq`, which fails at decode time on the same
-  input. Note: exhaustive `match`es on `YamlError` need a new arm.
+  value is materialized (#153). A deliberate divergence from `yq`, which accepts the same
+  input and emits a depth-limited expansion. Note: exhaustive `match`es on `YamlError`
+  need a new arm.
 - **BalancedParens L2 excess overflow** (#188): the per-L2-block excess
   counters were `i16` and overflowed at nesting depth > 32,767 (debug panic /
   silent wrap in release). Widened to `i32` across the scalar, NEON, and
--- a/docs/compliance/yaml/limitations.md
+++ b/docs/compliance/yaml/limitations.md
@@ -116,10 +116,11 @@ alias with nothing to resolve to:
 
 - **Cyclic.** An alias that would make an anchored value contain itself
   (`a: &x {self: *x}`) fails with `YamlError::AliasCycle` ("anchor value contains itself").
-  A cyclic document cannot be materialized at all; before
+  A cyclic document has no finite tree form; before
   [#153](https://github.com/rust-works/succinctly/issues/153) it aborted the process with a
-  stack overflow. YAML 1.2's representation graph technically permits cycles, but `yq`
-  rejects the same input at decode time, so rejecting is also the compatible behavior.
+  stack overflow. YAML 1.2's representation graph technically permits cycles. Unlike the
+  unknown-anchor case this is a deliberate divergence from `yq`, which accepts the input
+  and emits a depth-limited expansion (`{"a":{"self":{"self":{}}}}`) rather than an error.
 - **Unknown.** An alias naming an anchor that is not in scope — never defined, or defined
   *later* — fails with `YamlError::UnknownAnchor` since
   [#372](https://github.com/rust-works/succinctly/issues/372). YAML 1.2 §7.1 requires an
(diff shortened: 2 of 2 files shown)


newhoggy:
Added a second commit ([`3a207fb`](https://github.com/rust-works/succinctly/commit/3a207fb175579b74489ad804bf5c4625ee59a91a)) that goes slightly beyond #403, flagging it so the extra hunk is not a surprise.

While verifying this PR I checked the surviving half of the rewritten section against the actual binary. The cyclic bullet claimed:

> `yq` rejects the same input at decode time, so rejecting is also the compatible behavior.

That is false against `yq` v4.53.3 — the version this same page pins as the compatibility target. `yq` accepts every cyclic shape tried (`a: &x {self: *x}`, `a: &x [*x]`, `&x {a: *x}`, and a two-level cycle), exits 0, and emits a depth-limited expansion:

```
$ printf 'a: &x {self: *x}\n' | yq -o json -I0 '.'
{"a":{"self":{"self":{}}}}
```

So rejecting is a deliberate *divergence* from `yq`, not agreement with it. The repo already knew this: the #328 CHANGELOG entry says `yq` "instead emits a depth-limited expansion" and cites this page as the documented policy, while the older #153 entry ~20 lines above still said `yq` "fails at decode time". Both sit under `[Unreleased]`, so two contradictory statements were queued for the same release notes. The commit corrects both, and softens "cannot be materialized at all" to "has no finite tree form" — `yq` does materialize a truncated approximation.

The sentence predates this PR (it is verbatim from `main`), but this PR rewrote that paragraph into a symmetric two-bullet form where both bullets end in a `yq`-agrees-with-us clause, which lent the false half borrowed credibility.

The neighbouring unknown-anchor claim is unaffected and verified correct: `yq` really does report `unknown anchor 'x' referenced`.

Everything else in the PR checks out — all five CLI examples reproduce byte-exact, every row of the new `YamlError` table reproduces (including the exact 128/129 nesting boundary), the table is complete with respect to live variants (`DuplicateAnchor` and the two deprecated variants are never constructed), and an independent scan of all 402 corpus cases confirms the loader alone rejects exactly 12/94 with zero `UnknownAnchor` hits — so the "unmoved by #372" claim holds.

Issue #386, opened 2026-07-26
jq: the generic evaluator drops ? when it falls back to the full evaluator


newhoggy:
**Severity: Low** (latent — currently masked by #367) — found by inspection and pinned by a test during #358.

`eval_generic`'s builtin dispatch handles a handful of builtins itself and sends the rest to the full evaluator. That fallback rebuilds the expression from the bare `Builtin`, dropping the `optional` flag it was called with:

```rust
// src/jq/eval_generic.rs, eval_builtin_generic
// For other builtins, fall back to full evaluator via JSON
_ => {
    let owned = to_owned(&value);
    eval_on_owned::<S, _>(&Expr::Builtin(builtin.clone()), owned)
}
```

`optional` is in scope and is simply not threaded through, so an `Expr::Optional` wrapping any builtin the generic evaluator does not implement itself loses its optionality: the error is raised instead of suppressed. The `Expr::Optional` arm above it does the right thing (`eval_single::<S, _>(inner, value, true, cursor)`); the loss happens one level down.

This matters because the CLI (`sjq`, `syq`) runs on the generic evaluator while the library entry point runs on the full one — so the two evaluators answer differently for the same expression.

## Reproducing

Not reachable from the command line today: succinctly's parser rejects a postfix `?` after any function call (#367), so the expression cannot be spelled. Building it programmatically shows the split:

```rust
let expr = parse(r#"contains("a")"#).unwrap().optional();
let index = JsonIndex::build(b"1");

// full evaluator: suppressed, yields nothing (correct, matches jq)
eval::<Vec<u64>, JqSemantics>(&expr, index.root(b"1"));

// generic evaluator: still raises
eval_generic::eval_with_cursor(&expr, index.root(b"1"));
```

jq prints nothing for `1 | contains("a")?`.

Pinned as-is by `optional_suppresses_the_error` in `tests/jq_containment_tests.rs`, which asserts the generic evaluator *does* error and tells the reader to tighten the test once the flag is threaded — so a fix is forced to update it rather than silently passing.

## Why file it separately from #367

#367 is the parser half: `?` is only accepted after a path expression. Fixing #367 alone would expose this one immediately — every `builtin?` the generic evaluator does not implement would still raise through the CLI, which is the path most users are on. The two want fixing together, or at least in that order.

## Fix

[...]

Issue #489, opened 2026-07-30
jq: path() renders "no paths at all" as the root path, and keeps `?`-pruned components


newhoggy:
**Severity: Medium** — `path()` reports a path that does not exist, silently, and the value it reports (`[]`, the root) is the one path that always *does* resolve, so a caller feeding it to `getpath`/`setpath`/`delpaths` writes to the document root. Reproduced against the built CLI (release, `717466e2`, the #413 branch), jq oracle 1.7.1.

Found while reviewing the fix for #413. Two distinct defects in `builtin_path`, both visible without any computed key.

**1. "No paths at all" is rendered as the root path.** `path(empty)` is the minimal case — the tracker understands `empty` perfectly well, there is simply nothing to report:

```
$ echo '{"a":1}' | sjq -c '[path(empty)]'
[[]]                          # jq: []
```

**2. A `?`-pruned step still contributes its component.** The step never resolved, but the tracker emits it anyway:

```
$ echo '"s"' | sjq -c '[path(.a?)]'
[["a"]]                       # jq: []

$ echo '"s"' | sjq -c '[path(.[0]?)]'
[[0]]                         # jq: []

$ echo '"s"' | sjq -c '[path(.a?, .b?)]'
[[]]                          # jq: []
```

The third line shows the two defects composing: both branches prune, the result is empty, and defect 1 then turns that emptiness into the root path.

Correct behaviour is already in place when a path *does* resolve, which is why this has stayed hidden:

```
$ echo '{"a":1}' | sjq -c '[path(.a?)]'
[["a"]]                       # jq: same    ✓
$ echo '{"a":1}' | sjq -c '[paths]'
[["a"]]                       # jq: same    ✓
```

### Where

`builtin_path` (`src/jq/eval.rs`, ~line 9150). After the tracker runs, an empty `paths` vector is turned into the root path rather than into "no output":

```rust
if paths.is_empty() {
    if optional {
        QueryResult::None
    } else {
        QueryResult::Owned(OwnedValue::Array(vec![]))   // <- the root path
    }
}
```

`Owned(Array(vec![]))` is indistinguishable from a genuine `path(.)`, whose answer really is `[]`. "No paths" needs `QueryResult::None` regardless of `optional`; jq emits nothing for `path(empty)` and one output for `path(.)`.

[...]

Issue #413, opened 2026-07-27
jq: `?` on an assignment path prunes an error raised by the key itself


newhoggy:
Split out of the review of #390 (#360, computed keys in index brackets).

```
$ echo '"str"' | sjq -c '.[.k]? = 5'
"str"
                    # jq: jq: error (at <stdin>:0): Cannot index string with string "k"   (exit 5)
```

`resolve_node`'s `Expr::Optional` arm in `src/jq/eval.rs` prunes the branch on
*any* failure underneath it:

```rust
Expr::Optional(inner) => {
    // A failure under `?` prunes the branch instead of propagating.
    let Ok(branches) = resolve_node::<S>(inner, value) else {
        return Ok(Vec::new());
    };
```

That is right for a failure to *index* — `{"a":1} | .[null]? = 5` correctly
leaves the document alone. It is wrong for a failure raised while evaluating the
**key**, which `?` does not cover: on a string input, `.k` itself is the error,
and jq propagates it.

The branch already enforces exactly this rule in value position — `eval_index_expr`
deliberately passes `optional = false` into the key evaluation, so
`.[error("boom")]?` still raises `boom` and
`{"k":"a","a":1} | [.. | .[.k]?]` still fails as jq does. The path-context
resolver is the half that does not.

Low severity — it silently succeeds where jq errors, on a filter that is
already asking to be forgiving.

newhoggy:
This branch (#412's fan-out) widens where this bug bites, and changes its shape from a loud refusal into a silent partial write — worth a heads-up before anyone reasons about severity from the original repro alone.

Before #412's fix, `path(.. | objects | .[.k]?)`-shaped assignments were refused outright (`Cannot use a computed index after a multi-output path component`), so this bug was unreachable through them. Now that `resolve_node` fans `..`/`recurse`/the typeof filters out into one path per branch, every one of those branches can carry this same `?`-swallowing defect, and a multi-branch assignment applies each resolved path in sequence against the *same* mutating document:

```
$ echo '{"k":"a","a":{"k":"a","a":1}}' | sjq -c '(.. | objects | .[.k]?) = 7'
{"k":"a","a":7}
                    # jq: error (at <stdin>:1): Cannot index number with string "a"
```

[...]

PR #497, opened 2026-07-30
fix(jq): accept postfix ? after any expression, not just path expressions

### LINKED ISSUE #367: jq: postfix ? is only accepted after a path expression
Found while fixing #359: `setpath(["a"]; 1)?` does not parse, so the optional path through `builtin_setpath` had to be tested via `try … catch` instead.

jq accepts `?` after any postfix-able expression. succinctly's parser only takes it after a path expression — `.a?`, `.[0]?`, `.a.b?` — and rejects it everywhere else at the lexer.

| filter (input `{"a":1}`) | succinctly | jq |
|---|---|---|
| `.a?` | `1` | `1` |
| `.[0]?` / `.a.b?` | *(empty)* | *(empty)* |
| `length?` | `parse error at position 6: unexpected character '?'` | `1` |
| `keys?` | `parse error at position 4: …` | `["a"]` |
| `tonumber?` | `parse error at position 8: …` | *(empty)* |
| `(.a)?` | `parse error at position 4: …` | `1` |
| `(1)?` | `parse error at position 3: …` | `1` |
| `first(.[])?` | `parse error at position 10: …` | `1` |
| `getpath(["a"])?` | `parse error at position 14: …` | `1` |
| `setpath(["a"];1)?` | `parse error at position 16: …` | `{"a":1}` |

This fails loudly rather than silently — but `f?` is the idiomatic shorthand for `try f`, and it is common enough on builtins (`tonumber?`, `keys?`) to bite in ordinary filters.

The evaluator side is already in place: `Expr::Optional` exists and every builtin threads an `optional` flag, so this looks like moving `?` out of the path-expression parser into the general postfix loop. The sites consuming it today are around `src/jq/parser.rs:685`, `:887` and `:3119-3134`.

Note this is a parse failure, so it cannot be tracked in the error-message probe corpus the way a wording or behaviour divergence is — `tests/data/jq-error-probes.tsv` only holds filters that reach the evaluator.

Related: #360 tracks a different parser gap (`.[null]`, `.[true]`, `.[{}]` rejected).

Every cell above was measured against the pinned jq-1.7.1, not recalled.

newhoggy:
## Description

This PR fixes issue #367 where jq's postfix `?` operator (shorthand for `try`) was only accepted after path expressions like `.foo?` and `.[0]?`, but not after literals, parenthesized expressions, builtins, function calls, variables, or bare `.`. The fix involves both parser changes to accept `?` after any Term, and evaluator changes to properly thread the optional flag through all code paths that were previously unreachable.

### Root Cause

jq's grammar allows `?` after any Term, but succinctly's parser only checked for trailing `?` in two specific locations:
- The dot-field branch of `parse_primary` (`.foo?`)
- `parse_index_bracket_with_optional` (`.[0]?`)

Everywhere else — `length?`, `keys?`, `tonumber?`, `(1)?`, `(.a)?`, `first(.[])?`, `getpath(["a"])?`, `$x?`, `.?` — the trailing `?` was left unconsumed, causing the top-level parser to raise "unexpected character '?'" error.

Additionally, several evaluation paths never checked the `optional` flag they received, causing suppressed expressions to leak errors instead of silently failing.

## Type of Change
- [x] Test coverage improvement
- [x] Documentation update

## Related Issue
Fixes #367

## Changes Made

**Parser Changes (src/jq/parser.rs):**
- Refactored `parse_primary` by extracting the main dispatch logic into `parse_primary_inner`
- Wrapped the entire dispatch in an outer `parse_primary` that checks for trailing `?` after any Term
- Added `?` to `is_expr_terminator` so bare `.?` is recognized as identity rather than an attempted field name
- This single wrapping eliminates the need to patch ~13 individual branches

**Evaluator Changes (src/jq/eval.rs):**
- Updated `builtin_first`, `builtin_last`, and `builtin_reverse` to check `optional` flag before erroring
- Updated `builtin_tonumber` to check `optional` flag before erroring
- Modified `eval_arithmetic` to suppress errors when optional is true
- Modified `eval_error` to return `QueryResult::None` instead of error when optional is true
- Modified `eval_reduce` to propagate suppressed input and init sub-results instead of converting to errors
- Updated `builtin_setpath` to propagate suppressed sub-results and check optional on path validation

**Generic Evaluator Changes (src/jq/eval_generic.rs):**
- Updated `builtin_keys`, `builtin_keys_unsorted`, `builtin_length`, `builtin_first`, `builtin_last`, and `builtin_reverse` to check `optional` flag
- Updated `builtin_tonumber` to check `optional` flag with proper error handling
- Added optional-aware re-wrapping in the fallback bridge that delegates to the full evaluator, ensuring compound expressions properly suppress errors

**Test Coverage (tests/jq_tests.rs, tests/jq_containment_tests.rs, tests/jq_evaluator_parity_tests.rs):**
- Added `test_optional_after_builtin` covering `length?`, `keys?`, `tonumber?`
- Added `test_optional_after_parenthesized_expr` covering `(.a)?`, `(1)?`
- Added `test_optional_after_function_call` covering `first(.[])?`, `getpath(["a"])?`, `setpath(["a"];1)?`
- Added `test_optional_after_variable_and_identity` covering `$x?`, `.?`
- Added integration tests for actual query evaluation: `keys?` suppressing errors, `length?` on objects, `(1)?`, `.?`
- Simplified existing workaround tests to use direct parsing of `?` syntax instead of builder methods

**Documentation (CHANGELOG.md):**
- Documented the parser fix explaining the old limitation and the wrapping solution
- Documented which builtins were fixed to respect the optional flag
- Noted pre-existing gaps deliberately left unfixed (e.g., `null | reverse` without `?`)
- Documented evaluator fixes for arithmetic, error, reduce, and setpath suppression
- Documented the generic evaluator's fallback bridge fix
- Noted intentionally preserved behavior: key expression errors in `.[.k]?` still propagate
- Documented known limitations with `reduce` and `eval_owned_expr` shared across multiple flow control constructs

## Testing

**Automated Testing:**
- [x] All existing tests pass
- [x] New parser unit tests verify `?` parses correctly after builtins, parens, calls, variables, and bare `.`
- [x] New integration tests exercise the full evaluation pipeline for previously-failing syntax
- [x] Simplified existing workaround tests now use real syntax instead of builder methods
- [x] Verified row-by-row against jq 1.7.1 for builtin behavior

**Manual Testing:**
The following expressions now parse and evaluate correctly (previously failed with "unexpected character '?'"):
```bash
# Builtins
cargo test -- --nocapture | grep -E 'length\?|keys\?|tonumber\?'

# Parenthesized expressions
cargo test -- --nocapture | grep -E '\(\.a\)\?|\(1\)\?'

# Function calls
cargo test -- --nocapture | grep -E 'first\(.*\)\?|getpath|setpath'

# Variables and identity
cargo test -- --nocapture | grep -E '\$x\?|\.\?'
```

### Test Commands
```bash
cargo test
cargo fmt --check
```

## Performance Impact
- [x] No performance impact

The parser change wraps the dispatch once rather than patching multiple branches, slightly improving code locality. Evaluator changes only add conditional checks that were missing, with no additional overhead for the non-optional path.

## Checklist
- [x] I have added tests that prove my fix works

## Additional Notes

**Deliberate Non-Fixes:**
- `.[.k]?` and `.[error("boom")]?` still propagate errors raised while evaluating the key expression — this is correct behavior matching jq, where `?` only guards the indexing operation itself, not key computation
- `null | reverse` (without `?`) still errors in the generic evaluator instead of yielding `[]` — this is a pre-existing gap (missing `is_null()` arm) unrelated to this fix
- `reduce` built from a bare erroring generator or with type-erroring elements has suppression gaps that go through the shared `eval_owned_expr` helper and need a separate, broader fix alongside other stream-builtin error-swallowing issues

**Future Enhancements:**
The `eval_owned_expr` helper (used by `reduce`, `foreach`, `while`, `until`, `repeat`) collapses suppressed sub-results to `Ok(null)` before callers can distinguish them from real nulls. This should be addressed in a follow-up PR alongside other tracked stream-builtin error-swallowing issues (#386 class of problems).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1372,6 +1372,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **SelectIndex sample overflow** (#188): `SampleEntry` counters were `u32`
   and wrapped past 2^32 set bits (~512 MB of ones); widened to `u64`.
 
+- **jq postfix `?` was only accepted after a path expression** (#367): jq's
+  grammar allows `?` (shorthand for `try f`) after any Term — `length?`,
+  `keys?`, `(1)?`, `(.a)?`, `first(.[])?`, `getpath(["a"])?`,
+  `setpath(["a"];1)?`, `$x?`, `.?` — but succinctly's parser only checked for
+  a trailing `?` in two spots: the dot-field branch of `parse_primary`
+  (`.foo?`) and `parse_index_bracket_with_optional` (`.[0]?`). Everywhere
+  else a trailing `?` was left unconsumed, tripping the top-level "unexpected
+  character '?'" check. Rather than patch each of the ~13 branches that
+  didn't check for it individually, `parse_primary`'s whole dispatch is now
+  wrapped once (renamed to `parse_primary_inner`, with a thin outer
+  `parse_primary` checking for a trailing `?` after any Term); `?` was also
+  added to `is_expr_terminator` so bare `.?` is recognized as identity rather
+  than an attempted (invalid) field name. `Expr::Optional`'s evaluation was
+  already fully generic (`src/jq/eval.rs`/`eval_generic.rs`), so the parsing
+  gap itself needed no evaluator changes — but unlocking `?` after arbitrary
+  builtins surfaced latent divergences that were previously unreachable with
+  `optional` set through real syntax: `tonumber?`, `length?`, `keys?`,
+  `keys_unsorted?`, `first?`, `last?`, and `reverse?` raised a hard error
+  instead of suppressing it on a value they can't operate on, because their
+  `eval_generic.rs` arms (and, for `first`/`last`/`reverse`, their `eval.rs`
+  counterparts too) never checked the `optional` flag they were already
+  passed. All seven are now threaded through, verified row-by-row against jq
+  1.7.1. Not fixed here: `null | reverse` (no `?` involved) still errors
+  instead of yielding `[]` in the generic evaluator — a pre-existing,
+  unrelated gap (missing `is_null()` arm) that this change didn't expose or
+  touch.
+
 ### Changed
 
 - **A tag on an anchored sequence item is now rejected** (#328): `- &a !!str x`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1384,9 +1384,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   wrapped once (renamed to `parse_primary_inner`, with a thin outer
   `parse_primary` checking for a trailing `?` after any Term); `?` was also
   added to `is_expr_terminator` so bare `.?` is recognized as identity rather
-  than an attempted (invalid) field name. `Expr::Optional`'s evaluation was
-  already fully generic (`src/jq/eval.rs`/`eval_generic.rs`), so the parsing
-  gap itself needed no evaluator changes — but unlocking `?` after arbitrary
+  than an attempted (invalid) field name. Unlocking `?` after arbitrary
   builtins surfaced latent divergences that were previously unreachable with
   `optional` set through real syntax: `tonumber?`, `length?`, `keys?`,
   `keys_unsorted?`, `first?`, `last?`, and `reverse?` raised a hard error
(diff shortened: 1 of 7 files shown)


newhoggy:
Reviewed this branch's head against jq 1.7.1 while checking whether it interacts with #413/#482 (release build of `f3302df7`). The `?` half composes correctly — but the newly-parseable `(...)` path expressions reach four walkers that do not know `Expr::Paren`, and one of the four failure modes is silent.

**The good news first**, since it was what I came to check: `(.[.k])?` behaves as jq does, and it does *not* collide with #482's guard.

```
$ echo '"str"' | sjq -c '(.[.k])? = 5'
"str"                                # jq: same — (f)? is `try f`, so it catches the key error

$ echo '{"a":{"b":1},"k":"b"}' | sjq -c '(.a[.k])? = 5'
{"a":{"b":5},"k":"b"}                # jq: same
```

That is the right split: `(.[.k])?` catches the key evaluation, `.[.k]?` does not (#482 makes the bare form raise). The two land in different arms of `resolve_node` and both come out matching jq.

**Silent no-op** — `=` through a parenthesised path writes nothing and exits 0:

```
$ echo '{"a":{"b":1}}' | sjq -c '(.a.b)? = 5'
{"a":{"b":1}}                        # jq: {"a":{"b":5}}
```

The unparenthesised `.a.b? = 5` is correct, so it is the parens. Without the `?` the same path fails loudly, which is what makes the `?` form silent rather than merely wrong:

```
$ echo '{"a":{"b":1}}' | sjq -c '(.a.b) = 5'
jq: error: cannot use expression as assignment target       # jq: {"a":{"b":5}}
```

**Loud refusal** — `|=` and `del` reject the same shape:

```
$ echo '{"a":{"b":1}}' | sjq -c '(.a.b)? |= 5'
jq: error: cannot use expression as update target           # jq: {"a":{"b":5}}

$ echo '{"a":{"b":1}}' | sjq -c 'del((.a.b)?)'
jq: error: cannot use expression as delete target           # jq: {"a":{}}
```

### Cause

`set_path`, `get_path_mut`, `update_path` and `delete_at_path` have no `Expr::Paren` arm — `grep -c 'Expr::Paren'` is 0 in all four — so a parenthesised path falls to their catch-alls. `set_path`'s catch-all error is then swallowed by the `Expr::Optional` arm above it, which is why `=` is the silent one. The resolver side already looks through parens (`resolve_node`'s `Expr::Paren(inner) => resolve_node(inner)`, and `push_path_components`), as does `eval_with_path_tracking` — which is why `path((.a.b)?)` finds the path that the writers then cannot walk.

One arm each, mirroring what those two already do, should cover it.

### Two more rows, both pre-existing rather than yours

```
$ echo '{"a":{"b":1}}' | sjq -c '(..)? = 5'      # and 'recurse? = 5'
{"a":{"b":1}}            # jq: error: Cannot index number with string "a"
```

That is #498 (a `?` on a write path swallowing the failure to *apply* it) reached through a spelling this PR makes available; the same document errors correctly via `(.. | objects | .[.k]) = 7`.

```
$ echo '"str"' | sjq -c '[path((.a.b)?)]'
[[]]                     # jq: []
```

That is #489's root-path rendering, likewise not introduced here.

Worth a case or two in this PR's tests for the writer side specifically — `path()` reading a parenthesised path correctly is exactly what hid the walker gap that #493 fixed for `Optional`, and the shape is the same one.

Issue #485, opened 2026-07-30
yaml: an out-dented block sequence continuation silently drops the rest of the sequence and corrupts the next entry into an empty key


newhoggy:
## Summary

A block sequence continuation line indented *between* its parent key's indent and
the sequence's own indent silently drops that item **and everything after it**,
and corrupts the next mapping entry into a phantom empty key.

The input is invalid YAML — `yq` rejects it, and succinctly's opt-in strict
validator correctly rejects it too. But the default (non-validating) loader emits
well-formed JSON with data missing and a key that was never in the document, which
is the failure mode #325 was filed about: *silently losing data is worse than
either erroring or parsing the obvious extension*. Same class as #324/#328/#332/#371.

Found while reviewing #325.

## Reproduction

```console
$ printf 'b:\n    - x\n   - y\nc: 2\n' | succinctly yq -o json -I0 '.'
{"b":["x"],"":"c"}
```

Two separate corruptions in one line of output:

- `y` is **gone** — no error, no trace.
- `c: 2` became `"":"c"` — a key that does not appear in the document, and the
  value `2` is gone too.

`yq` rejects the same input outright:

```console
$ printf 'b:\n    - x\n   - y\nc: 2\n' | yq -o json -I0 '.'
Error: bad file '-': yaml: while parsing a block mapping at <unknown position>:
line 2, column 4: did not find expected key
```

The strict validator already agrees with `yq`, so only the lenient loader is at issue:

```console
$ printf 'b:\n    - x\n   - y\nc: 2\n' > /tmp/t.yaml && succinctly yaml validate /tmp/t.yaml; echo $?
1
```

### Everything after the misaligned item is lost, not just that item

```console
$ printf 'b:\n    - x\n   - y\n    - z\n' | succinctly yq -o json -I0 '.'
{"b":["x"]}                   # y AND z both gone
```

### Minimal form (no trailing entry, no outer nesting)

```console
$ printf 'b:\n  - x\n - y\n' | succinctly yq -o json -I0 '.'
{"b":["x"]}
```

The trigger is the continuation `-` sitting at an indent strictly between the
parent mapping key's indent and the established sequence's indent.

## Cause (starting point, not verified)

`close_deeper_indents` presumably closes the sequence when it sees the shallower
indent, but nothing then re-establishes a container for the out-dented `- y`, so
the item is parsed into a context that is discarded. The `"":"c"` phantom suggests
the subsequent `c: 2` line is consumed as the *value* of a node whose key was
[...]

Issue #151, opened 2026-07-09
JSON validator: unbounded recursion causes stack-overflow abort on deeply nested input


newhoggy:
**Severity: High (DoS)** — confirmed (SIGABRT).

`src/json/validate.rs:192,211,279` — `validate_value` -> `validate_array`/`validate_object` recurse per bracket with no depth limit. `validate(b"[[[…×500000…]]]")` aborts the process with `fatal runtime error: stack overflow` (uncatchable SIGABRT). Reachable from public `validate()` and the `--validate` / `json validate` CLI on attacker-controlled input.

**Fix:** thread a depth counter through `validate_value` and return an error past a bound (~128).

---
_Filed from an automated full-code-review pass. Severity and verification status noted above._

newhoggy:
Reopening: this was auto-closed by the merge of #242, whose description said the guard tests "fixes #151" — a GitHub closing keyword. #242 only added the depth-100 guard test (`tests/deep_nesting_valid_tests.rs`); the actual depth cap in `src/json/validate.rs` has not landed, and `validate_value`/`validate_object`/`validate_array` still recurse unbounded on `main`.

The YAML twin #152 was just fixed in #257 (`nesting_depth` counter + cap of 128, new error variant); the same pattern transfers directly here via the existing `self.error(kind)` helper.

Issue #130, opened 2026-02-06
State Machine JSON Validator: Implementation Complete, Optimization Opportunities


newhoggy:
## Summary

Implemented a state machine JSON validator (`validate_sm`) that processes 64-byte SIMD chunks continuously without stopping at string boundaries. This solves the UTF-8 carry state synchronization issue where the recursive descent parser would stop mid-chunk when finding a closing quote.

### Two-Phase Processing

```
Phase 1: SIMD Scanner (never stops early)
├── UTF-8 validation (Keiser-Lemire)
├── Quote mask (64-bit bitmap)
├── Backslash mask (64-bit bitmap)
├── Structural char mask ({ } [ ] : ,)
└── Control char detection

Phase 2: State Machine (processes significant positions from masks)
├── Explicit state enum (15 states)
├── Explicit stack for nesting
├── Process interesting positions from SIMD masks
└── Validate escapes, numbers, keywords byte-by-byte
```

### Key Data Structures

- `State` enum: 15 states covering structural, string, escape, number, and keyword parsing
- `NestingStack`: Replaces recursive call stack for object/array nesting
- `ChunkScan`: Contains UTF-8 validity, quote/backslash/structural masks
- `ChunkScanner`: Produces ChunkScan for each 64-byte chunk

### Files Modified

- `src/json/validate_simd.rs`: Added `ChunkScan`, `ChunkScanner`, structural character detection
- `src/json/validate.rs`: Added `State`, `NestingStack`, `StateMachine`, `validate_sm()`

## Current Performance

| Content Type | Size | Original | State Machine | Ratio |
|-------------|------|----------|---------------|-------|
| ASCII-heavy (objects) | 1KB | 528 MiB/s | 455 MiB/s | 0.86x |
| | 10KB | 525 MiB/s | 478 MiB/s | 0.91x |
| | 100KB | 532 MiB/s | 475 MiB/s | 0.89x |
| UTF-8 heavy (strings) | ~300B | 6.1 GiB/s | 8.0 GiB/s | **1.32x** |
| | ~1KB | 7.0 GiB/s | 8.1 GiB/s | **1.16x** |
| | ~7KB | 7.6 GiB/s | 7.5 GiB/s | 0.99x |

**Current state**: ~10-14% slower for typical structured JSON, faster for UTF-8 heavy content.

### O1: Carry Addition for Quote Region Masking (from DSV parser)

[...]

PR #127, opened 2026-02-05
feat(simd): implement AVX2-accelerated JSON string validation for x86_64

newhoggy:
## Description

This PR implements high-performance SIMD-accelerated JSON string validation for x86_64 platforms using AVX2 instructions. The implementation uses the industry-standard Keiser-Lemire UTF-8 validation algorithm combined with efficient terminator detection, achieving 2-10x throughput improvements over the scalar implementation.

Key achievements:
- **Nested JSON**: 1.8 GiB/s → 17.3 GiB/s (9.6x improvement)
- **String-heavy content**: 1.5 GiB/s → 9.6 GiB/s (6.4x improvement)
- **Unicode patterns**: 900 MiB/s → 1.9 GiB/s (2.1x improvement)

The implementation processes 64-byte chunks with combined UTF-8 validation and terminator scanning (quote, backslash, control characters) in a single pass, minimizing memory access overhead while maintaining strict RFC 8259 compliance.

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Performance improvement
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [x] Test coverage improvement
- [ ] CI/CD changes

## Related Issue
N/A - Performance enhancement initiative

## Changes Made

**Core SIMD Implementation (`src/json/validate_simd.rs`):**
- Created new SIMD module for x86_64 with AVX2 operations
- Implemented Keiser-Lemire UTF-8 validation algorithm with lookup tables
- Added 64-byte chunk processing with combined UTF-8 + terminator detection
- Implemented cross-chunk UTF-8 state tracking via `Utf8Carry` structure
- Added `SimdConstants` structure for pre-computed SIMD constant vectors
- Implemented proper 256-bit register shifting for cross-chunk boundary handling
- Added SSE2 fallback for 16-63 byte ASCII-only chunks
- Added scalar fallback for non-x86_64 platforms and chunks < 16 bytes

**Validator Integration (`src/json/validate.rs`):**
- Integrated SIMD module with conditional compilation for x86_64
- Added `SimdConstants` initialization at document start for reuse
- Updated `validate_string()` to use SIMD chunked processing with bitmask operations
- Implemented efficient backslash bitmap processing for batch escape validation
- Added comprehensive x86_64-specific code paths with `#[cfg(target_arch)]`
[...]

PR #521, opened 2026-07-31
Reject the SIMD chunk-scanner JSON validator (ADR-0012), and give the validator a safety net

newhoggy:
Closes the analysis for #130.

## What this is

#130 proposed replacing the scalar recursive-descent JSON validator with a
simdjson-style 64-byte SIMD chunk scanner feeding an explicit state machine.
**This PR rejects that proposal** at the analysis stage, per the ADR-0007
precedent, and ships the tooling and safety net the analysis exposed as missing.

`docs/adrs/adr-0012.md` has the full argument. In short: the deciding metric is
not run length but **interesting density** — state-machine visits per 64-byte
chunk. Every file in the real-workload corpus is on the wrong side of the
8–12 visits/chunk break-even, in one of two ways:

| corpus file | visits / 64 B | throughput |
|---|---|---|
| pretty/3d-ribbon.json | 4.0 | **1.79 GiB/s** (already at the bandwidth ceiling) |
| tabular/carshare-data | 8.0 | 1.21 GiB/s |
| geojson/us-election | 10.9 | 1.19 GiB/s |
| charts/bullet-data | 20.0 | 856 MiB/s (above break-even) |
| graph/miserables.json | 24.0 | 789 MiB/s (above break-even) |

Nothing is both slow enough to be worth optimising and sparse enough for the
scanner to help. Three of #130's premises had also gone stale — the UTF-8
validator and the quote/backslash scanners it planned to salvage from PR #127
already exist in `src/text/utf8.rs` and `src/util/simd/escape.rs`, and the
`NestingStack` was justified by #151, which `MAX_NESTING_DEPTH = 128` already
fixed.

Whitespace fraction, the intuitive proxy, points the opposite way from the
truth and nearly reversed the conclusion — 3d-ribbon.json is 85% whitespace,
but 0.00% of its whitespace runs reach 32 bytes.

## What it ships instead

**The validator had no external conformance check, no fuzzing and no
differential oracle**, despite being reachable from `succinctly json validate`
and `sjq --validate` on attacker-controlled input. It now has all three:

- **JSONTestSuite conformance** (`tests/json_test_suite.rs`), vendored at a
  pinned SHA with a CI drift detector. Scores **95/95 accept, 188/188 reject**,
  with all 35 implementation-defined cases pinned to recorded verdicts — which
  doubles as the only written description of which JSON dialect this crate
  accepts.
- **Property + `serde_json` differential tests** (13 tests), including
  exhaustive sweeps: every truncation point, every byte value in 9 embedding
  contexts, and every construct at every offset across three 64-byte boundaries.
- **A cargo-fuzz harness** (3 targets) sharing one oracle with the property
  tests so the divergence classifier cannot drift between them.

Plus:

- `perf(json)`: **-9.0% on `numbers`** from scanning digit runs in one step.
  (A keyword literal-compare change was measured at **+6.0% on `literals`** —
  slower on the pattern it targeted — and reverted rather than tuned, per
  ADR-0005/0006.)
- `fix(bench)`: `cargo bench --bench json_validate_bench` previously reported
  success while running **zero** benchmarks, because its corpus dir is
  gitignored and every loop `continue`d past the missing files. It now fails
  loudly and has a real-workload leg.

## Review fixes included

Found while reviewing the above:

- **Tightened the divergence classifier.** It excused any disagreement at depth
  `>= 127` (only 128 can legitimately diverge) and treated `0e0` as an
  underflow (excusing every divergence in any document containing a
  zero-with-exponent literal). This is the fuzzer's oracle — anything it
  excuses is invisible forever. `classifier_does_not_excuse_near_misses` pins
  both; each half was confirmed to fail against the previous predicates.
- **`corpus-shape.md` had silently drifted** and is refreshed here. Three YAML
  rows were stale from the #485 fix, in `prometheus-ci.yml` — the one YAML file
  in the fetched tier, i.e. exactly the region CI's seed-based golden cannot
  observe. Added `corpus-shape-full-drift` so it cannot recur. (The JSON
  section was unaffected, so ADR-0012's evidence stands.)
- **Added `fuzz-build`**: nothing in CI compiled `fuzz/`, so the harness could
  rot and be found broken exactly when needed.
- **Fixed the documented fuzz invocation**, which failed on a fresh checkout
  for 2 of 3 targets (`ERROR: The required directory ... does not exist`).
- **Corrected ADR-0012's reproduce instructions** — the stated bench command
  aborts without the synthetic ladder, and 4 of the 5 corpus files are fetched
  rather than vendored.
- **The real-corpus bench leg now reports when it is running seed-only** (1 of
  5 files, 0.06% of corpus bytes) rather than looking like a normal run.

## Verification

- `cargo clippy --all-targets --all-features -- -D warnings` clean
- Full validator suites green; JSONTestSuite 95/95 and 188/188
- Fuzzing campaign running on ARM at 4 CPU-hours/target; **x86 leg still
  outstanding** (`terminus` offline) and should land before merge


newhoggy:
## Fuzzing campaign — clean

Per [`fuzz/README.md`](../blob/issue-130-state-machine-json-validator-implementation/fuzz/README.md#when-to-run-it), which asks for ≥4 CPU-hours per target on both benchmark architectures before any change to the validator's control flow. Run at `bb331eb5`, seeded from `./fuzz/seed-corpus.sh` (320 seeds: the 318 vendored JSONTestSuite cases + the real-workload seed + proptest regressions).

| target | ARM runs | x86_64 runs | crashes |
| --- | ---: | ---: | ---: |
| `validate_never_panics` | 3,455,791,014 | 6,296,243,004 | 0 |
| `validate_position_invariant` | 3,119,124,980 | 4,999,242,414 | 0 |
| `validate_vs_serde_json` | 481,675,142 | 199,107,446 | 0 |

**~18.5 billion executions, zero crashes, zero unclassified `serde_json` divergences, zero position-invariant violations.** Both legs exited 0 on all three targets.

- ARM: Apple silicon, 12 workers (4 per target), 1 h wall
- x86_64: Zen 4 (Ryzen 9 7950X), 12 workers, 1 h wall

`validate_vs_serde_json` runs ~20× fewer iterations than the other two because every execution also parses the input with `serde_json`; its coverage is correspondingly the richest of the three (ft: 7004 vs ~1670).

### The one artifact was not a crash

A `slow-unit-` file predating the campaign (from an earlier smoke run) — 63 KB of huge-exponent float literals. Both implementations *reject* it, and outside ASAN+coverage instrumentation both do so in microseconds:

```
succinctly: 2µs      serde_json: 5.458µs
```

So it is an artifact of instrumented execution, not a slow path in the validator. No action.

### Caveat

This campaign validates the validator **as it stands on this branch**. Its value is as the baseline that makes the *next* change to `validate.rs` checkable — which is the point of landing the harness.

newhoggy:
## x86 fuzz leg — clean (closes the gap from #130's last comment)

Ran the campaign `fuzz/README.md` asks for (4 CPU-hours/target) on the x86_64 benchmark machine (`terminus`, 7950X, 32 threads), synced to this PR's head commit (`8f212622`).

| target | execs | artifacts |
|---|---|---|
| `validate_never_panics` | 1,558,447,510 | 0 |
| `validate_vs_serde_json` | 271,876,797 | 0 |
| `validate_position_invariant` | 1,335,577,506 | 0 |

Clean: no crashes, no panics, no oracle disagreements across 3.1B+ total executions. Combined with the ARM run already on record, both benchmark architectures `fuzz/README.md`'s "before any change to the validator's control flow" guidance asks for are now covered.

Issue #483, opened 2026-07-30
jq: path(..)/path(recurse)/path(select) silently return [] instead of the real paths


newhoggy:
**Severity: High** — `path(..)` is one of the most common jq path idioms and returns a wrong answer *silently*. Reproduced against the built CLI (release, `15583dad`), jq oracle 1.7.1.

Split out of the work on #412, which fixed the *computed-key* half of this. Input for every case below is `{"a":{"b":1},"c":2}`.

`path()` emits `[]` (the root path) instead of the real paths — no error, just a wrong answer:

```
$ echo '{"a":{"b":1},"c":2}' | sjq -c '[path(..)]'
[[]]                          # jq: [[],["a"],["a","b"],["c"]]

$ echo '{"a":{"b":1},"c":2}' | sjq -c '[path(recurse)]'
[[]]                          # jq: [[],["a"],["a","b"],["c"]]

$ echo '{"a":{"b":1},"c":2}' | sjq -c '[path(.. | numbers)]'
[[]]                          # jq: [["a","b"],["c"]]

$ echo '{"a":{"b":1},"c":2}' | sjq -c '[path(.[] | select(type=="number"))]'
[[]]                          # jq: [["c"]]
```

`=`, `|=` and `del()` at least fail loudly on the same filters:

```
$ echo '{"a":{"b":1},"c":2}' | sjq -c '(.. | numbers) = 9'
jq: error: cannot use expression as assignment target    # jq: {"a":{"b":9},"c":9}

$ echo '{"a":{"b":1},"c":2}' | sjq -c '(.. | numbers) |= .+1'
jq: error: cannot use expression as update target        # jq: {"a":{"b":2},"c":3}

$ echo '{"a":{"b":1},"c":2}' | sjq -c 'del(.. | numbers)'
jq: error: cannot use expression as delete target        # jq: {"a":{}}
```

### Where

`eval_with_path_tracking` and `collect_intermediate_with_paths` (`src/jq/eval.rs`) understand only `Identity`, `Field`, `Index`, `Iterate`, `Slice`, `Pipe`, `Optional`, `Paren` (plus a `debug_assert!`-only `IndexExpr` arm). `RecursiveDescent`, `Builtin::Recurse*`, `Builtin::Select`, the typeof filters and `Comma` all fall to the catch-all:

```rust
// For complex expressions that don't represent simple paths, we can't track
_ => {
    // For non-path expressions, we don't produce path output
}
```

[...]

PR #621, opened 2026-08-03
fix(jq): path() now refuses a filter that is not a path expression

### LINKED ISSUE #530: jq: path() silently accepts a non-path filter where jq raises "Invalid path expression"
**Severity: Medium** — `path()` accepts a filter that is not a path expression at all and answers with *nothing*, where jq refuses it by name. Reproduced against the built CLI (release, on top of `5d56243a` with #489 applied), jq oracle 1.7.1.

Found while sweeping for residual divergences after fixing #489. Distinct from #483: that issue is about expressions that **are** path expressions but have no arm in the walker (`..`, `recurse`, `select`, and — see the comment I left there — `first`, `if`, `//`, `limit`, `try`, `as $x |`, `label`). This one is the opposite half: expressions that are **not** path expressions, which jq is required to reject.

```
$ echo '{"a":1}' | sjq -c '[path(1)]'
[]                    # jq: error: Invalid path expression with result 1

$ echo '{"a":1}' | sjq -c '[path(length)]'
[]                    # jq: error: Invalid path expression with result 1

$ echo '{"a":1}' | sjq -c '[path(keys)]'
[]                    # jq: error: Invalid path expression with result ["a"]

$ echo '{"a":1}' | sjq -c '[path(.a + 1)]'
[]                    # jq: error: Invalid path expression with result 2

$ echo '{"a":1}' | sjq -c '[path(.a | tostring)]'
[]                    # jq: error: Invalid path expression with result "1"

$ echo '{"a":1}' | sjq -c '[path({a:1})]'
[]                    # jq: error: Invalid path expression with result {"a":1}
```

Before #489 these read `[[]]` — the root path — so the wrong answer was a path that resolved. #489 turned that into no output, which is safer but still not a diagnostic, and it is what makes this class worth naming separately.

### The write side is already loud, with its own wording

`=`, `|=` and `del()` refuse the same filters, so the machinery to reject them exists; only `path()` is silent, and the sentence differs:

| Filter          | succinctly                                     | jq                                       |
|-----------------|------------------------------------------------|------------------------------------------|
| `(1) = 5`       | `cannot use expression as assignment target`   | `Invalid path expression with result 1`  |
| `del(1)`        | `cannot use expression as delete target`       | `Invalid path expression with result 1`  |
| `(length) = 5`  | `cannot use expression as assignment target`   | `Invalid path expression with result 1`  |
| `[path(1)]`     | *(no output)*                                  | `Invalid path expression with result 1`  |

Whether to align the write-side wording with jq's is a judgement call that belongs with this issue rather than a separate one: jq uses the single sentence everywhere, and the value in it (`result 1`) is what tells a user *which* branch of their filter was not a path.

### `?` does not suppress it

Unlike an indexing error, this one survives `?` — it is a statement about the filter, not about the value:

```
$ echo '{"a":1}' | jq -c '[path(("a")?)]'
jq: error (at <stdin>:1): Invalid path expression with result "a"
```

So a fix cannot route it through the same `optional` check that #489 added for `Cannot index …`.

### Where

`walk_path` (`src/jq/eval.rs`) — the `_` catch-all that #489 left in place:

```rust
// Expressions with no path-tracking arm — `..`, `recurse`, `select`,
// arithmetic — name no path (#483). `builtin_path` renders that as no
// output, which is still the wrong answer but no longer a path that
// resolves.
_ => {}
```

That arm currently conflates the two classes. Splitting it needs the walker to know which expressions jq considers path-capable: everything path-capable but unimplemented is #483, and everything else should raise the new error carrying the value the filter produced — which means evaluating it, as jq does.

`EvalError` has no constructor for this sentence yet (`grep -rn "Invalid path expression" src/` is empty); it would sit alongside `cannot_index` and friends in `src/jq/error.rs`. Note the message embeds a *rendered value*, so it needs the same truncation rules as `cannot_iterate` — see `describe` and the `iterate_string_*_limit` probes.

### Testing

The error cases cannot be jq-golden fixtures (`scripts/sync-jq-golden.sh` fails on a non-zero jq exit), so they belong in `tests/data/jq-error-probes.tsv` alongside the eight `path_*` probes #489 added. Ordering: fixing this closes the last silent-wrong-answer path in `path()` that is not #483.

newhoggy:
## Summary

Fixes #530.

`path(expr)` silently answered with no output when `expr` was not a path expression at all (`path(1)`, `path(length)`, `path(keys)`, `path(.a + 1)`, `path({a:1})`, ...), where jq raises `Invalid path expression with result <v>` by name. `walk_path`'s catch-all conflated two things jq distinguishes: expressions that are genuinely not path expressions (this issue) and expressions that are, but have no tracking arm implemented yet (#483, untouched here).

- Split the catch-all into an explicit allowlist of #483's path-capable-but-unimplemented constructs (unchanged behavior) and a fallback that evaluates the expression for real and raises the new `EvalError::invalid_path_expression` on its first output.
- The refusal is not suppressed by a `?` nested *inside* the offending expression, only by an outer `?` around the whole `path(...)` call — verified against jq-1.7.1 directly, along with ~110 other live probes that pinned the exact classification of every `Expr`/`Builtin` variant.
- **Deferred, tracked as known divergences**: the write side (`=`, `|=`, `del()`) still uses its own wording for the same class of filter, since `set_path`/`delete_at_path` aren't generic over the evaluator yet (unlike `update_path`) — aligning them is a bigger refactor left to a follow-up issue.

See `docs/compliance/jq/limitations.md` for the full residual-divergence list this fix knowingly leaves open (partial `reduce`/`foreach` path-capability, two other jq "Invalid path expression" sentence shapes, and a handful of `jv_identical` edge cases).

## Commits

- `fix(jq)`: the `walk_path` split, new `EvalError` constructor, in-crate unit tests
- `test(jq)`: error-probe/known-divergence bookkeeping (resolves 3, tracks 4 deferred write-side ones)
- `test(test)`: 2 new golden fixture cases
- `docs(docs)`: `limitations.md` + `CHANGELOG.md`

## Test plan

- [x] `cargo build --features cli` + manual repro of all cases from the issue
- [x] `cargo test --features cli,regex` (52/52 test binaries pass)
- [x] Spot-checked #483 constructs (`path(..)`, `path(recurse)`, `path(if ... end)`, etc.) are unaffected
- [x] Re-verified after an automatic rebase onto a newer `main` (which added a computed-slice-bounds feature touching the same file) — confirmed no semantic collision, including `path()` over a computed slice bound


newhoggy:
Superseded by #618, which merged into `main` while this PR was in progress and fixes the same issue (#530) more comprehensively — it also resolves #483's path-tracking gaps that this PR deliberately left as an explicit allowlist, plus a bonus partial-output bug (`path(.a, 1)`) found along the way. Closing without merging; no further action needed on this branch.

Issue #666, opened 2026-08-08
jq: spike the smallest lazy-keys slice to validate #140's QueryResult/GenericResult invasiveness


newhoggy:
**Severity: Low** — de-risk step for #140, split out so the invasiveness question gets answered by a small throwaway prototype instead of discovered midway through the full implementation.

## Why

#140 proposes a new `JqValue::LazyKeysArray` variant for O(1)-memory `keys`/`keys_unsorted` streaming. Confirmed (2026-08-07 re-triage) that `JqValue` (`src/jq/lazy.rs:43`) is narrower in this codebase than the issue assumes — it's only used at the CLI output boundary (`lazy.rs`/`value.rs`/`jq_runner.rs`), not threaded through the core evaluator. The eager collection #140 wants to eliminate actually happens upstream, inside `builtin_keys` (`src/jq/eval.rs:2358`) and `eval_generic.rs`'s `Keys` arm (`:1721`), before any `JqValue` exists — so true laziness needs new `QueryResult`/`GenericResult` variants too, both matched exhaustively in multiple places in the evaluator. That's a materially larger, multi-file change than the issue's plan implies, and no prototype or memory measurement exists yet confirming the payoff is worth it.

## Scope

- Build the smallest possible slice — e.g. `keys_unsorted | length` returning a cached count without materializing the key array — far enough to touch every layer the real change would (`QueryResult`/`GenericResult` variant, the runner boundary, `JqValue`) and surface how invasive the exhaustive-match sites actually are.
- Measure actual memory savings on a realistic large-object workload (10K-100K fields, matching #140's own stated motivation) to confirm the premise is worth the multi-layer change before committing to it.
- Throwaway/spike — does not need to be production-quality or merged as-is; the deliverable is the finding, not the code.

## Acceptance

- [x] A minimal prototype demonstrates (or disproves) that `keys_unsorted | length` can avoid materializing the key array without a disproportionate `QueryResult`/`GenericResult` surface change.
- [x] Measured memory savings on a realistic large-object workload are recorded, confirming or revising #140's O(n)→O(1) motivation.
- [x] A clear go/no-go plus revised scope estimate for #140's full plan, based on what the spike found.

## Related

Precursor to #140 (M3.1: lazy streaming keys).

newhoggy:
## Spike findings

[...]

Issue #140, opened 2026-04-04
M3.1: Lazy streaming keys via JqValue::LazyKeysArray


newhoggy:
## Summary

`keys` and `keys_unsorted` currently materialize all keys into a `Vec<String>` then wrap them in `Vec<OwnedValue>`, producing O(n) heap allocations for an object with n fields. For large objects (10K–100K+ fields), this is wasteful when the result is either streamed to output or piped through `.[]`.

This issue proposes a `JqValue::LazyKeysArray` variant that wraps a cursor and yields keys lazily — O(1) memory for the keys operation itself.

## Current Code

Both eval paths collect eagerly:

**`eval.rs:1737–1756`** (JSON path):
```rust
let mut keys: Vec<String> = Vec::new();
for field in fields {
    if let StandardJson::String(k) = field.key() {
        if let Ok(cow) = k.as_str() {
            keys.push(cow.into_owned());
        }
    }
}
if sorted { keys.sort(); }
let arr: Vec<OwnedValue> = keys.into_iter().map(OwnedValue::String).collect();
QueryResult::Owned(OwnedValue::Array(arr))
```

**`eval_generic.rs:940–956`** (generic/YAML path):
```rust
let keys = fields.keys();
let owned_keys: Vec<OwnedValue> = keys.into_iter().map(OwnedValue::String).collect();
GenericResult::Owned(OwnedValue::Array(owned_keys))
```

## Why `JqValue` not `OwnedValue`

`OwnedValue` has no lifetime parameter — it can't hold cursor references. `JqValue<'a, W>` already bridges lazy cursors and owned values (`JqValue::Cursor`), so it's the natural home for a lazy keys variant.

## Key enabler: `JsonFields` and `YamlFields` are `Copy`

Both field iterators are `Copy` (they wrap an `Option<Cursor>`), so a `LazyKeysArray` can be iterated multiple times without consuming the source — critical for operations like `length` followed by `.[]`.

### Step 1: Add `JqValue::LazyKeysArray` variant

```rust
pub enum JqValue<'a, W = Vec<u64>> {
    // ... existing variants ...

    /// Lazy array of object keys, backed by a cursor field iterator.
    /// Avoids materializing all keys upfront for large objects.
    LazyKeysArray {
        /// Field iterator (Copy — can be iterated multiple times)
        fields: JsonFields<'a, W>,
        /// Number of fields (cached from count)
        len: usize,
        /// Whether keys should be sorted (keys vs keys_unsorted)
[...]

PR #673, opened 2026-08-08
perf(jq): spike lazy keys_unsorted to de-risk #140 (#666)

newhoggy:
## Summary

Throwaway spike for #666, precursor to #140. Prototypes `GenericResult::LazyKeysUnsorted` + `JqValue::LazyKeysArray` so `keys_unsorted | length` and bare `keys_unsorted` answer from the object's field iterator directly instead of decoding every key into an owned `String` first — avoiding the reserialize + `JsonIndex`-rebuild round trip `keys_unsorted | length` previously went through via `eval_on_owned`.

**Not intended to merge as-is** — per #666's scope, the deliverable is the finding, not the code. Opened as a draft PR for reference alongside the findings.

Full write-up: https://github.com/rust-works/succinctly/issues/666#issuecomment-5224198817

## Measured results

Interleaved A/B (`/usr/bin/time -l`, median of 7 reps), 100K-field synthetic object, output byte-identical to baseline on every query/edge case tested:

| Query | Before RSS | After RSS | Reduction | Time |
|---|---|---|---|---|
| `keys_unsorted \| length` | 29.29 MB | 10.90 MB | **63% (2.7x)** | 20ms → 10ms |
| `keys_unsorted` | 24.59 MB | 10.80 MB | **56% (2.3x)** | 20ms → 10ms |

At 100K fields the after-RSS for `keys_unsorted | length` lands within noise of plain `length` alone (10.86 MB) — key materialization overhead is eliminated down to the process floor.

## Invasiveness findings (the actual point of #666)

- `GenericResult<V>`: 14 exhaustive match sites needed a new arm (compiler-forced) — more than the ~11 statically estimated. Most are 1-3 lines via one shared materialize-fallback helper; 2 forward unchanged for free (`first`/`last`).
- `QueryResult` (`eval.rs`): **zero** sites touched — confirmed the live CLI path never routes this query shape through it, contra #140's original assumption.
- `JqValue`: 1 new variant, 3 compile-forced sites + 1 wildcard-covered site that's silently wrong unless updated, plus genuine new streaming logic in `write_json`/`print_json`.
- Runner boundary: `jq_runner.rs` (2 sites + `print_json`), `yq_runner.rs` (1 site — YAML has no lazy-output concept today).

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (lib + integration suites, run in isolation to avoid known cargo-lock contention flakiness)
- [x] Output byte-identical vs. pre-change binary across `keys_unsorted`, `keys_unsorted | length`, `keys_unsorted | .[0]`, `last`, `first`, `map`, `sort`, comparisons, nested `.[] | keys_unsorted`, empty object, and the YAML `yq` side
- [x] Interleaved A/B memory/time measurement at 10K and 100K fields

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -10,6 +10,7 @@ use std::io::{BufWriter, IsTerminal, Read, Write};
 use std::path::{Path, PathBuf};
 
 use succinctly::dsv::{build_index as build_dsv_index, DsvConfig, DsvRows};
+use succinctly::jq::document::DocumentFields;
 use succinctly::jq::eval_generic::{eval_with_cursor, to_owned as generic_to_owned, GenericResult};
 use succinctly::jq::{self, format_number_jq_compat, Expr, JqValue, OwnedValue, Program};
 use succinctly::json::light::{JsonCursor, StandardJson};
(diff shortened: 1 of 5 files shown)


newhoggy:
Closing without merging — this was a throwaway spike per #666's scope. The deliverable was the finding, not the code; full write-up is on #666 and the revised scope has been folded into #140.

Issue #725, opened 2026-08-09
jq: native Builtin::Map arm for plain array/object map/select (Slice 2 of #700)


newhoggy:
## Summary

Slice 2 of #700's design (`docs/plan/jq-lazy-map-select.md`): give `Builtin::Map` its first-ever native arm in `eval_generic.rs`'s `eval_builtin`, covering plain `arr | map(f)` / `obj | map(f)` (and `select` composed after it) on containers that never touch `keys_unsorted`/`keys`. Today `Builtin::Map` has zero arms anywhere in `eval_builtin` and unconditionally falls through the `to_owned` → reserialize → reindex → re-evaluate round trip.

This is the higher-value half of #700's scope decision: #686's measurement found that **75-95% of the fallback's measured cost (up to 4.8x time, 13.5x memory at 100MB) hits plain `map`/`select`**, not just `keys_unsorted`-derived ones. It is sequenced after Slice 1 (#724) because it's structurally larger — first real semantics for a previously-absent builtin arm — but both should land as part of the same follow-up effort from #700, not "Slice 1 now, Slice 2 someday."

## Scope

- Reuse the `LazySeq`/`LazySource`/`Instruction` machinery introduced by Slice 1 (#724) — this issue only needs two new `LazySource` variants (`Elements(V::Elements)`, `Values(V::Fields)`) and one new production site.
- Add a `Builtin::Map(f)` arm in `eval_builtin`, next to the existing `Builtin::Select` arm: on `value.as_array()`/`value.as_object()`, build `LazySeq::new(LazySource::Elements(..)|Values(..)).push_map(f, S::tag())` and return `GenericResult::LazySeq(..)`.
- Everything downstream (the `Pipe` fold's `GenericResult::LazySeq` composability arm, `stream_json`/`stream_yaml`, the shared `materialize_lazy` fallback) is already in place from Slice 1 — this issue should not need to touch those again.

## Non-goals (see design doc for full list)

- Elementwise `.[] | select(g)`/`.[] | map(f)` — the existing `Expr::Iterate` eager fan-out (`collect_cursors()`) is a related-but-distinct eagerness, out of scope here.
- `map_values`, recursive laziness (`map(map(f))` with the inner `map` also lazy), YAML merge-key mappings, backward/lookahead access on YAML.
- CLI streaming-output integration (`yq_runner.rs`'s `can_use_m2_streaming`) — separate follow-up (Slice 3, not yet filed).

## Measurement required before merge

[...]

Issue #700, opened 2026-08-09
jq: design a general lazy map/select primitive for eval_generic.rs


newhoggy:
## Summary

#686's measurement spike found that `keys_unsorted | map/select`'s eager fallback in `eval_generic.rs` costs **4.6-4.8x time and up to 13.5x memory** at 100MB scale, growing with input size, and that **75-95% of that cost is a pre-existing structural gap affecting *any* container's `map`/`select`**, not something specific to `keys_unsorted`. This issue is the "go" follow-up #686's own acceptance criteria calls for: design (not yet implement) a general lazy `map`/`select` primitive in `eval_generic.rs`.

## Background

- PR #678 (#140/M3.1) added `GenericResult::LazyKeys`/`LazyIndexRange` so `keys_unsorted | length/.[]/.[n]/first/last` avoid materializing keys eagerly, but explicitly could not extend that to `map`/`select` because **no container type has a native lazy `map`/`select` path today** — `Builtin::Map` has no arm in `eval_builtin` at all and always falls through the wildcard `to_owned` + `eval_on_owned` round trip ([eval_generic.rs:2537-2540](https://github.com/rust-works/succinctly/blob/main/src/jq/eval_generic.rs#L2537-L2540)); `Builtin::Select` is a native arm but only tests/republishes a single value, it doesn't itself iterate a container.
- #686 measured that fallback cost on a flat many-key object (`wide` pattern) and confirmed it's real, algorithmic (grows with size), and mostly structural rather than `keys_unsorted`-specific: see [#686's findings comment](https://github.com/rust-works/succinctly/issues/686#issuecomment-5231966440) for the full tables and methodology.
- The only existing "laziness" mechanism in `eval_generic.rs` is ad hoc: the `Expr::Pipe` dispatch pattern-matches the *literal next AST node* against `GenericResult::LazyKeys`/`LazyIndexRange` and special-cases exactly `length`/`.[]`/`.[n]`/`first`/`last` ([eval_generic.rs:1162-1320](https://github.com/rust-works/succinctly/blob/main/src/jq/eval_generic.rs#L1162-L1320)). There's no general lazy-fold/iterator primitive to plug a `map`/`select` arm into — building one is the actual design question here.

## Scope

This issue is for a **design**, informed by the #686 measurements — not an implementation PR. Questions to resolve:

[...]

Issue #686, opened 2026-08-08
jq: eval_generic.rs has no lazy map/select for any container type (blocks lazy keys_unsorted | map/select)


newhoggy:
## Summary

#140's original plan assumed `keys_unsorted | map(f)` and `keys_unsorted | select(f)` could stay lazy (iterate the field cursor, apply `f`, without materializing first). PR #678 found that's not actually available to opt into: **`eval_generic.rs` has no lazy `map`/`select` for any container type today, not just for `LazyKeysUnsorted`.** `map` is standard jq sugar (`def map(f): [.[] | f];`) and `select`'s filtering happens through the same general `.[]`-then-collect machinery — there is no dedicated lazy-iteration arm to plug a `LazyKeysUnsorted` fast path into.

## Scope

This is **not** a `keys`-specific follow-up like the other three deferred pieces — it's a standalone question of whether `eval_generic.rs` should grow a general lazy iteration/fold primitive that arrays, objects-via-`.[]`, and lazy-keys results could all share. That's a materially larger, unrelated design project:

- Would touch how `.[]`, `map`, `select`, and comma-fanout compose today, not just the `keys_unsorted` fast paths PR #678 added.
- No spike or measurement exists yet on whether real workloads (`keys_unsorted | map(...)`, `keys_unsorted | select(...)`, or general large-array `map`/`select`) are memory/time-bound enough to justify it — unlike `length`/`.[]`/`.[n]`/`first`/`last`, which #140's own spike (#666) measured a concrete 2.3-2.7x win for.

## What to do (before committing to any implementation)

Per this repo's own established practice (#661, #663, #664, #665, #666, #675 — measure before implementing), the right next step is a small measurement precursor, not a design doc:

- Benchmark `keys_unsorted | map(f)` / `keys_unsorted | select(f)` (and, for comparison, plain large-array `map`/`select`) to see whether the eager materialization these currently fall back to is actually a measurable bottleneck at realistic sizes, the same way #666 did for `length`/`.[]` before #140's real implementation.
- If the payoff is real, scope a general lazy-iteration design as its own issue; if not, close this as a documented non-goal the way P5-P8 in `docs/parsing/yaml.md` record rejected optimizations.

## Acceptance

- [ ] A measurement exists showing whether `keys_unsorted | map/select` (and general large-container `map`/`select`) have a real, worthwhile laziness payoff.
- [ ] Go/no-go recorded here; if "go," a properly scoped design issue is filed separately rather than folded into this one.

[...]

Issue #724, opened 2026-08-09
jq: lazy map/select fast path for keys_unsorted/keys (Slice 1 of #700)


newhoggy:
## Summary

Slice 1 of #700's design (`docs/plan/jq-lazy-map-select.md`): wire a new `GenericResult::LazySeq` primitive into the *existing* `GenericResult::LazyKeys`/`LazyIndexRange` fast path in `eval_generic.rs`'s `Expr::Pipe` fold, so `keys_unsorted|keys | map(f)/select(g)` (and chains like `keys_unsorted | map(f) | select(g)`) stay lazy instead of falling back to the full `to_owned` → reserialize → reindex → re-evaluate round trip.

This is the smaller, lower-risk half of #700's scope decision ("general primitive, staged delivery") — see the design doc for the full mechanism, the composability rationale, and why `select` itself gets no new code path. Slice 2 (the dominant cost share: plain `.foo | map/select` / `arr | map/select` on containers that never touch `keys_unsorted`) is tracked separately and should land alongside this, not instead of it — see the linked follow-up issue.

## Scope

- Add `LazyElem`/`LazySource`/`Instruction`/`LazySeq` types and the `GenericResult::LazySeq` variant in `src/jq/eval_generic.rs`, per the design doc.
- Extend the `Expr::Pipe` fold's existing `GenericResult::LazyKeys { .. }`/`GenericResult::LazyIndexRange(_)` arms with a `Builtin::Map` case each (same `!sorted` guard the existing `.[]`/`first`/`last` arms use for `LazyKeys`).
- Add the composability arm: `GenericResult::LazySeq(seq) => match unwrap_paren(expr) { ... }`, handling `Builtin::Map` (push another instruction), `Builtin::Length`/`Expr::Iterate`/`first`/`.[0]` (single-pass native), and a materializing fallback for everything else (`select`, `last`, nonzero `.[n]`, comparisons).
- Extend `GenericResult::stream_json`/`stream_yaml` with a `LazySeq` arm (new `stream_lazy_seq_json`/`_yaml` in `src/jq/stream.rs`, generalizing the existing `stream_lazy_keys_json`/`_yaml` idiom from #685).
[...]

PR #742, opened 2026-08-10
jq: lazy map fast path for keys_unsorted/keys (Slice 1 of #700)

newhoggy:
## Summary

Slice 1 of #700's design (`docs/plan/jq-lazy-map-select.md`): wires a new `GenericResult::LazySeq` primitive into the existing `LazyKeys`/`LazyIndexRange` fast path in `eval_generic.rs`'s `Pipe` fold, so `keys_unsorted|keys | map(f)/select(g)` (and chains like `keys_unsorted | map(f) | select(g)`) stay lazy instead of falling back to the full `to_owned` → reserialize → reindex → re-evaluate round trip (#686 measured that fallback at 4.6-4.8x time / up to 13.5x memory at 100MB).

- `EvalTag`/`EvalSemantics::tag()` (`src/jq/eval.rs`) let a deferred `map(f)` re-dispatch to the concrete `JqSemantics`/`YqSemantics` that built it from contexts where the generic evaluator's `S` type parameter has already been erased (e.g. `GenericResult`'s inherent `stream_json`/`into_owned`).
- `LazyElem`/`LazySource`/`Instruction`/`LazySeq` + the new `GenericResult::LazySeq` variant (`src/jq/eval_generic.rs`), wired into the `LazyKeys`/`LazyIndexRange` `Pipe`-fold arms via a new `Builtin::Map` case each (same `!sorted` guard the existing `.[]`/`first`/`last` arms use), plus a composability arm so `map(f) | map(g) | ...` composes onto the same `LazySeq` — one value, not one type per depth.
- `select` gets no new code path (`Builtin::Select` is untouched) — composed after a `map`, it reaches its native arm only via `LazySeq`'s own single-forward-pass materializing fallback, never the four-pass round trip.
- Every element's `map(f)` is still evaluated even for `first`/`.[0]`: jq's `map(f)` is `[.[] | f]`, an eager, non-short-circuiting array construction, so a later element's error must still surface even when only the first element is wanted. `materialize_atomic` mirrors `eval::map_over`'s atomicity (a mid-map error discards the whole array, never a partial prefix) — including through the new `stream_lazy_seq_json`/`_yaml` (`src/jq/stream.rs`), which buffer locally and write nothing to `out` on failure, since a partial array is never valid jq output for a failed `map`.
- `LazySource::Elements`/`Values` (plain-container `map`, the dominant cost share per #686) are defined now but constructed only by #725 (Slice 2), per `docs/STYLE_GUIDE.md`'s STYLE-0005.
- Retired the three fallback-pinning tests, replaced with tests asserting the lazy path is taken, plus new coverage for chains, atomicity, the `first`/`.[0]` error-visibility fix, YAML (`YamlFields`'s `Clone`-not-`Copy` cons-list), yq semantics, and the `.[0]`/`.[2]` fast-path boundary.

### Post-review fixes

A review pass found 6 issues in the initial implementation, fixed in 4 follow-up commits (all with new regression tests, full suite + all 3 CI clippy feature-set legs green after each):

- `Expr::Optional` (`?`/`try`) didn't catch an error raised inside a deferred `map(f)` chain — it matched the immediate `GenericResult`, but a `LazySeq` comes back unevaluated, so `(keys_unsorted | map(f))?` let the error escape instead of yielding empty.
- `stream_lazy_seq_json`/`_yaml` rendered a mapped element's own nested containers at the wrong depth, and the YAML block-style path wrote nothing instead of `[]` for an empty sequence.
- `push_generic_truthiness`'s `LazySeq` arm fully materialized the array (deep-cloning every element) just to check truthiness — an array is always truthy, so only error detection was needed (`drain_for_error`).
- ~10 duplicated `Err(Control::Error)/Err(Control::Break)` match arms collapsed into shared `control_into_generic`/`control_into_generic_optional` helpers.

One review finding was deliberately **not** fixed here: `yq_runner.rs`'s `can_use_m2_streaming` still doesn't admit `Builtin::Map`, so the new `stream_lazy_seq_json`/`_yaml` functions aren't yet reachable from the `syq`/yq CLI. That's explicitly out of scope for this issue (see "Non-goals" above and Slice 3 in the design doc), not a bug.

## Measurement

Interleaved A/B, `wide` pattern, 100kb-100mb, output byte-identical to system `jq -c` at every size (gated: 0 differences across all configurations on both platforms).

**Apple ARM (M-series):**

| Query | 100kb | 1mb | 10mb | 100mb |
|---|---|---|---|---|
| `keys_unsorted \| map(.)` | 1.13x | 1.38x | 1.45x | 1.40x |
| `keys_unsorted \| select(true)` | 0.96x | 0.98x | 0.98x | 0.97x |
| `keys_unsorted \| map(.) \| select(true)` | 1.03x | 1.07x | 1.09x | 1.11x |

**x86_64 (AMD Ryzen 9 7950X, Zen 4):**

| Query | 100kb | 1mb | 10mb | 100mb |
|---|---|---|---|---|
| `keys_unsorted \| map(.)` | 1.44x | 1.72x | 1.79x | 1.74x |
| `keys_unsorted \| select(true)` | 0.98x | 1.01x | 1.00x | 1.00x |
| `keys_unsorted \| map(.) \| select(true)` | 1.24x | 1.26x | 1.19x | 1.23x |

`map` shows a real, growing win on both platforms — notably larger on x86_64 (up to 1.79x) than ARM (up to 1.45x). Bare `select` (no `map`) is correctly unaffected on both — it has no new code path in this slice. `map | select` lands in between, since `select` still materializes once via `LazySeq`'s fallback. This slice is not expected to move plain `Map`/`Select` (that's #725); it didn't.

x86_64 run included a `--control` pass first (before-binary vs. a copy of itself) to establish this machine's noise floor: median -0.08%, range -5.2%..+1.5% across the same 12 configurations — every real result above is well outside that range in a consistent direction, so this isn't noise.

## Test plan

- [x] `cargo test` (default features, and `cli,simd,regex,serde`) — full suite green
- [x] `cargo clippy --all-targets -- -D warnings` for all three CI feature combinations (including `broadword-yaml`)
- [x] Output identity vs system `jq` (byte-identical)
- [x] x86_64/Zen 4 A/B leg

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -11,7 +11,9 @@ use std::path::{Path, PathBuf};
 
 use succinctly::dsv::{build_index as build_dsv_index, DsvConfig, DsvRows};
 use succinctly::jq::document::DocumentFields;
-use succinctly::jq::eval_generic::{eval_with_cursor, to_owned as generic_to_owned, GenericResult};
+use succinctly::jq::eval_generic::{
+    eval_with_cursor, materialize_atomic, to_owned as generic_to_owned, GenericResult, LazyElem,
+};
 use succinctly::jq::{self, format_number_jq_compat, Expr, JqValue, OwnedValue, Program};
 use succinctly::json::light::{JsonCursor, StandardJson};
 use succinctly::json::validate::{self, ValidationError};
@@ -1613,6 +1615,21 @@ fn evaluate_input(
         GenericResult::LazyIndexRange(len) => Ok(vec![OwnedValue::Array(
             (0..len).map(|i| OwnedValue::Int(i as i64)).collect(),
         )]),
+        // `keys_unsorted | map(f)` (#724) reaches this runner boundary as a
+        // not-yet-materialized `LazySeq` the same way `LazyKeys`/
+        // `LazyIndexRange` above do — one forward pass here, same
+        // sink-reporting shape as the `Error`/`Break` arms below.
+        GenericResult::LazySeq(seq) => match materialize_atomic(seq) {
+            Ok(o) => Ok(vec![o]),
+            Err(jq::Control::Error(e)) => {
+                sink.report(DiagStyle::Jq, &e, at);
+                Ok(vec![])
+            }
+            Err(jq::Control::Break(label)) => {
+                sink.report_break(DiagStyle::Jq, &label, at);
+                Ok(vec![])
+            }
+        },
         GenericResult::None => Ok(vec![]),
         GenericResult::Error(e) => {
             sink.report(DiagStyle::Jq, &e, at);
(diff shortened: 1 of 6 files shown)


newhoggy:
Closing as superseded: PR #740 (merged 2026-08-10) independently implemented the same `GenericResult::LazySeq` primitive and explicitly closed both #724 and #725 together, including two correctness fixes this branch's own follow-up commits were separately chasing (`map(f)?` error suppression, `.[]`-after-`map` atomicity). Rebasing this branch onto current `main` produces ~20 conflicts that are all "two names for the same thing" (`Rc<Vec<Instruction>>` vs `Vec<Instruction>`, `.materialize_atomic()` method vs free fn, `S::TAG` vs `S::tag()`, etc.), not genuine functional differences — not worth reconciling.

One independent, still-needed fix from this branch is *not* covered by #740: `d18dd4b1` (`fix(json): decode JSON strings before YAML-quoting in stream_json_as_yaml`) — a pre-existing bug in `src/json/light.rs` unrelated to the `LazySeq` work, confirmed still present on current `main`. Will file that separately.

PR #774, opened 2026-08-10
jq: add @base32/@base32d format functions

### LINKED ISSUE #719: jq: @base32/@base32d format is not implemented
## Summary

`@base32`/`@base32d` are not implemented. Confirmed at the source level:
`FormatType` (`src/jq/expr.rs`, ~line 459) has exactly 13 variants — `Text`,
`Json`, `Uri`, `Csv`, `Tsv`, `Dsv(String)`, `Base64`, `Base64d`, `Html`,
`Sh`, `Urid`, `Yaml`, `Props` — no `Base32`/`Base32d`. Live-verified:

```
$ echo '"abc"' | succinctly jq '@base32'
compile error: parse error at position 1: unknown format '@base'
```

Real jq (1.6+) has supported `@base32`/`@base32d` as standard format
builtins for a long time, alongside `@base64`/`@base64d`. This is a gap
against jq's supported format surface, not merely a coverage gap — there is
no fallback or workaround for a filter that uses it.

## Suggested implementation

Add `FormatType::Base32`/`Base32d` variants and encode/decode functions
alongside the existing `@base64`/`@base64d` implementation (RFC 4648 base32
alphabet, `=`-padded). Update `CLAUDE.md`'s jq format-functions table once
implemented.

## Discovery context

Found via a coverage audit cross-referencing succinctly's jq test suite
against jq-1.7.1's own test suite (`tests/base64.test` and jq's builtin
format list).

newhoggy:
## Summary
- Adds `@base32`/`@base32d` jq format functions (RFC 4648, `=`-padded), closing the gap against jq's standard format builtins.
- Fixes a latent parser bug found along the way: `parse_format_string`'s format-name scanner hardcoded `'6'`/`'4'` as the only digits allowed in a format identifier (just enough to admit `base64`), so `@base32` silently truncated to `@base` before the fix.
- Updates CLAUDE.md, docs/reference/jq-language.md, docs/reference/jq-evaluator.md, and CHANGELOG.md.

Closes #719

## Test plan
- [x] `cargo test --features cli,simd,regex,serde --workspace` (all 54 test binaries pass)
- [x] `cargo fmt --check`
- [x] Manually verified encode/decode against all six official RFC 4648 test vectors (`""`, `"f"`, `"fo"`, `"foo"`, `"foob"`, `"fooba"`, `"foobar"`) via the built CLI
- [x] Manually verified `isvalid(@base32)`/`isvalid(@base32d)` and `?`-suppression on non-string input

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **jq `@base32`/`@base32d` format functions** (#719): RFC 4648 base32
+  encoding/decoding, alongside the existing `@base64`/`@base64d`. Output
+  matches jq 1.6+ and the RFC's own test vectors (`"f"` → `"MY======"`,
+  `"foobar"` → `"MZXW6YTBOI======"`, etc.). Fixed a latent parser bug found
+  along the way: `parse_format_string`'s format-name scanner hardcoded `'6'`
+  and `'4'` as the only digits allowed in a format identifier (to admit
+  `base64`), so `@base32` truncated to `@base` before this fix — generalized
+  to `is_ascii_alphanumeric()`.
+
 - **jq `@csv`/`@tsv`/`@dsv`/`@sh` allocation overhead investigated: no
   measurable end-to-end effect** (#647, follow-up to #124's real win for
   `@uri`/`@html`): a byte-scanning rewrite of the four format functions was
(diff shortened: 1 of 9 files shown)


newhoggy:
Closing without merging — this doesn't actually close a jq compatibility gap.

Verified directly against jq 1.7.1 (`jq-1.7.1-apple`) and jq 1.8.2 (Homebrew):

```
$ echo '"hello"' | jq -r '@base32'
jq: error (at <stdin>:1): base32 is not a valid format
```

jq's own manual only documents `@text`, `@json`, `@html`, `@uri`, `@csv`, `@tsv`, `@sh`, `@base64`, `@base64d` — no `@base32` entry — and jqlang/jq's upstream `src/builtin.c` (master) has zero occurrences of "base32", only base64. Real jq has never implemented `@base32`/`@base32d`.

See #719 for the closing rationale.

PR #825, opened 2026-08-13
fix(jq): honor sub()'s "g" flag for a global replace

### LINKED ISSUE #728: sub(re; replacement; "g") ignores the g flag, never performs a global replace
## Summary

In real jq, `sub(re; str; flags)` treats a `"g"` flag exactly like `gsub`:
per jq's own `builtin.jq`, `gsub($re; str; $flags)` is literally defined as
`sub($re; str; $flags + "g")`. So `sub(re; str; "g")` and `gsub(re; str)`
are equivalent — both should replace **every** match, not just the first.

`builtin_sub_with_flags` in
[`src/jq/eval.rs`](../blob/main/src/jq/eval.rs) never inspects the `g`
character at all — it unconditionally calls `re.replace()` (single
replacement), regardless of what's in `flags`:

```rust
// Replace first match
let result = re.replace(&input, replacement.as_str());
QueryResult::Owned(OwnedValue::String(result.into_owned()))
```

```console
$ echo '"aaa"' | jq 'sub("a";"b";"g")'
"bbb"
$ echo '"aaa"' | succinctly jq 'sub("a";"b";"g")'
"baa"     # should be "bbb"
```

## Suggested fix

In `builtin_sub_with_flags` (`src/jq/eval.rs`), check `flags.contains('g')`
and dispatch to `re.replace_all()` instead of `re.replace()` when set —
essentially delegating to the same code path `builtin_gsub_with_flags`
already uses.

## Test coverage

Pinned as a known failure in `tests/data/jq-golden-known-failures.txt`:
`regex_sub_g_global_replace` (see `tests/data/jq-golden/cases/`).

Discovered while scoping #703 (regex flag test coverage).

newhoggy:
## Summary

`sub(re; replacement; "g")` silently only replaced the first match. jq
defines `gsub(re; str; flags)` as `sub(re; str; flags + "g")`, so
`sub(re; str; "g")` and `gsub(re; str)` should be equivalent — both
replace every match.

`builtin_sub_with_flags` (`src/jq/eval.rs`) never inspected the `g`
character in `flags` at all, unconditionally performing a single
replacement. Now checks `flags.contains('g')` and, when set, dispatches
to the same `global_captures`/`stitch_replacements` path
`builtin_gsub_with_flags` already uses.

Fixes #728

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 2350+ tests, 0 failed)
- [x] Diffed output against pinned real `jq` for `sub("a";"b";"g")` on `"aaa"` — matches (`"bbb"`)
- [x] Un-pinned the `regex_sub_g_global_replace` known-failure golden case, now passing
- [x] `omni-dev coverage diff` — 100% patch coverage (5/5 new lines)


newhoggy:
- Extracted the global-flag check into a `let global = ...` binding, matching the identical convention already used at `builtin_match`/`builtin_capture_with_flags`.
- Noted the known-failure removal inline, matching this file's established fix-trail convention.
- Broadened golden coverage: added a `"gi"` combined-flag case and a `"g"` + named-capture-interpolation case.

The capture-group case surfaced a real, pre-existing bug unrelated to this PR's scope: `sub`/`gsub`'s replacement expression is evaluated once against the original input instead of per-match with `.` bound to the match's captures, so `\(.name)` referencing a named capture errors. Reproduces on plain `sub`/`gsub` with no flags at all — filed as #826 and pinned as a known failure rather than fixed here.

PR #827, opened 2026-08-13
fix(jq): stop implode wrapping codepoints > u32::MAX

### LINKED ISSUE #738: bug(jq): implode wraps codepoints > u32::MAX instead of substituting U+FFFD
## Summary

`builtin_implode` (`src/jq/eval.rs:5266`) casts a codepoint to `u32` with `as`,
which **wraps** on truncation instead of range-checking. jq substitutes
U+FFFD for any codepoint outside `0..=0x10FFFF` (or in the surrogate range);
succinctly does that correctly for codepoints that overflow `u32` cleanly —
but when the low 32 bits of such a codepoint happen to land on a *valid*
codepoint, the wrap silently produces that codepoint instead of U+FFFD.

## Reproduction

```console
$ echo '[4294967297]' | jq -c implode          # 4294967297 = 2^32 + 1
"(U+FFFD replacement character)"
$ echo '[4294967297]' | ./target/debug/succinctly jq -c implode
""
```

`4294967297 as u32` wraps to `1`, and `char::from_u32(1)` succeeds, so the
out-of-range codepoint is silently reinterpreted as U+0001 rather than
replaced with U+FFFD. Same effect for any `codepoint > u32::MAX` whose
`codepoint % 2^32` lands in a valid, non-surrogate range.

## Fix sketch

Range-check before truncating instead of relying on the `as u32` wrap, e.g.:

```rust
let c = u32::try_from(codepoint)
    .ok()
    .and_then(char::from_u32)
    .unwrap_or('\u{FFFD}');
result.push(c);
```

`u32::try_from` fails (rather than wrapping) for any `i64` outside
`0..=u32::MAX`, which then falls through to the U+FFFD substitution —
matching jq's `c < 0 || c > 0x10FFFF` check without the wraparound gap.

## Discovery context

Found while reviewing #706's `implode` boundary-table tests. That issue's
boundary table tops out at `1114112` (one past the max valid codepoint),
which doesn't exercise `u32` wraparound — this needs codepoints past
`u32::MAX` (~4.3 billion), a separate, deeper edge case. The cast predates
#706's fix and isn't a regression from it.

newhoggy:
## Summary

`builtin_implode` (`src/jq/eval.rs`) cast a codepoint to `u32` with `as`,
which wraps on truncation instead of range-checking. jq substitutes
U+FFFD for any codepoint outside `0..=0x10FFFF` (or in the surrogate
range); succinctly did too, except when a codepoint's low 32 bits
happened to land on a *valid* codepoint after wrapping — e.g.
`4294967297` (2^32 + 1) wrapped to `1`, silently producing U+0001
instead of the correct U+FFFD.

Switches to `u32::try_from`, which fails (rather than wraps) for any
`i64` outside `0..=u32::MAX`, falling through to the existing U+FFFD
substitution — matching jq's range check without the wraparound gap.

Fixes #738

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 54 binaries, 0 failed)
- [x] Diffed output against pinned real `jq` for the issue's repro (`[4294967297]`) and several boundary cases (max valid codepoint, one past max, negative, plain ASCII) — all match
- [x] Added `implode_u32_wraparound` golden case (`tests/data/jq-golden/cases/`)
- [x] `omni-dev coverage diff` — 100% patch coverage (5/5 new lines)


newhoggy:
Addressed code-review findings (3 independent agents, no correctness bugs found) in a follow-up commit, then rebased onto \`main\` (now includes #728, merged in the meantime — clean rebase, no conflicts, unrelated function in the same file):

- Added a large-magnitude negative codepoint (\`-4294967196\`, which the old \`as u32\` cast wrapped to a spuriously "valid" \`U+0064\`) to the boundary table — the only negative case previously covered (\`-1\`) happened to already be correct under both old and new code, so it didn't actually exercise the fix.
- Folded the intermediate \`let c = ...;\` binding back into a single \`result.push(...)\`, matching the surrounding code's existing style.

Full verification (build/test/clippy x2/coverage) re-run post-rebase — all green, 100% patch coverage (9/9 new lines).

Issue #676, opened 2026-08-08
test(jq): seed a golden case pinning recurse(f)'s DFS traversal order, ahead of #635


newhoggy:
## Summary

#635 reports that `recurse(f)`/`recurse(f;cond)` use breadth-first queue order
instead of jq's depth-first definition. The only existing unit test exercising
`recurse(f)`, `test_recurse_with_filter` (`src/jq/eval.rs:26780`), cannot
detect this bug in either direction — it only asserts the output array's
*length*, never its order.

## The gap, worked through concretely

`test_recurse_with_filter` runs:

```
{"name": "root", "children": [{"name": "a", "children": [{"name": "b"}]}, {"name": "c"}]}
[recurse(.children[]?) | .name]
```

and asserts `arr.len() == 4`. Tracing both orderings by hand:

- **DFS (jq's real semantics)**: yield `root`; recurse fully into `a`'s subtree
  before moving to the next sibling → `["root", "a", "b", "c"]`.
- **BFS (this codebase's `resolve_recurse`, per #635)**: yield `root`; enqueue
  all of `root`'s children, dequeue each before descending further →
  `["root", "a", "c", "b"]`.

Both are length 4. The existing test passes today, and would keep passing
after either a correct DFS fix or no fix at all — it provides zero regression
signal for #635. There's also no entry for `recurse` in
`tests/data/jq-golden-known-failures.txt`.

## What to do

Add an order-sensitive assertion — either strengthen
`test_recurse_with_filter` to assert full array equality
(`["root", "a", "b", "c"]`, not just length), or add a golden case under the
existing manifest — that currently **fails** against the BFS implementation.
This gives #635 a real scoreboard: a red test to turn green, the same role
`tests/data/jq-golden-known-failures.txt` plays for #534.

## Precedent

Same shape as #572, which seeded golden coverage for #534's shapes before
that fix was attempted, so the fix has something concrete to make pass rather
than relying on manual verification.

Found during the 2026-08-08 backlog risk re-triage (#235).

PR #828, opened 2026-08-13
fix(jq): fix path(recurse(f)) dropping elements on a trailing iterate

### LINKED ISSUE #682: jq: path(recurse(f)) drops elements when f ends in a bare .foo[]/.foo[]? over 2+ items
## Summary

`resolve_recurse` (`src/jq/eval.rs:8671`) — the path-tracking evaluator behind
`path(recurse(f))`, `del(recurse(f) | ...)`, and `recurse(f) |= ...` — silently
drops all but the first result whenever `f`'s trailing component is a bare
`.foo[]` / `.foo[]?` (array/object iterate) that yields 2+ elements. Value-level
`recurse(f)` (`builtin_recurse_f`) is unaffected; only the path-context form is.

Confirmed present on `main` at 73a0864c, independent of #635/#676's BFS→DFS
fix — both the old queue-based and new stack-based `resolve_recurse` have it.

```
$ echo '{"children":[{"name":"a"},{"name":"c"}]}' | jq -c '[path(recurse(.children[]?))]'
[[],["children",0],["children",1]]

$ echo '{"children":[{"name":"a"},{"name":"c"}]}' | ./target/release/succinctly jq -c '[path(recurse(.children[]?))]'
[[]]
```

A single-element array does not trigger it (`{"children":[{"name":"a"}]}` →
correct `[[],["children",0]]`), which is what let the existing test suite miss
this — none of the current `path(recurse(f)...)` tests
(`tests/jq_computed_key_tests.rs`, `tests/jq_recurse_depth_tests.rs`, the
`path_recurse_cond_*` goldens) exercise an `f` whose *trailing* component is a
bare iterate over 2+ elements; they all either use `.[]?` directly on the
already-multi-key root object (which does work — see below) or nest a
computed key/typeof filter after the iterate, which takes a different,
correctly-fanning-out code path.

Confirmed value-level recurse is fine on the same input:

```
$ echo '{"children":[{"name":"a"},{"name":"c"}]}' | ./target/release/succinctly jq -c '[recurse(.children[]?)]'
[{"children":[{"name":"a"},{"name":"c"}]},{"name":"a"},{"name":"c"}]   # correct, matches jq
```

And a *standalone* (non-recurse) `path(.children[]?)` on the same input is
also fine:

```
$ echo '{"children":[{"name":"a"},{"name":"c"}]}' | ./target/release/succinctly jq -c '[path(.children[]?)]'
[["children",0],["children",1]]   # correct, matches jq
```

So the bug is specific to `resolve_recurse`'s use of `resolve_node`'s raw
output, not to path resolution of `.foo[]?` in general.

## Root cause

`.children[]?` parses as `Pipe([Field("children"), Optional(Iterate)])` —
the trailing `?` binds to `[]` alone, not the whole chain (confirmed via
`succinctly::jq::parse`). In `resolve_seq` (`src/jq/eval.rs:8955`):

```rust
let Some(last_dynamic) = flat.iter().rposition(needs_path_prepass) else {
    let end = value_after_components::<S>(&flat, value).map_err(|e| (Vec::new(), e))?;
    return Ok(vec![(flat, end)]);   // <-- single branch, no fan-out
};
```

`needs_path_prepass` (`src/jq/eval.rs:8066`) returns `false` for both
`Expr::Iterate` and `Expr::Optional(Iterate)`, so when the pipe's *only*
dynamic-looking element is a trailing bare iterate, `resolve_seq` never enters
its per-element fan-out loop and instead returns exactly one `(flat, end)`
branch — collapsing what should be N branches (one per array element) into
one. `resolve_recurse` (`src/jq/eval.rs:8671`) consumes this single branch
directly as `children`, so only one candidate ever reaches the stack;
`resolve_recurse`'s own null-pruning (`if matches!(child_value,
OwnedValue::Null) { continue; }`) then silently drops it entirely whenever
`value_after_components` can't collapse a multi-element array down to one
scalar (observed returning `Null` for 2+ elements, and apparently the sole
element itself for exactly one — explaining why the 1-element case looks
correct by coincidence).

Whatever code path handles a **top-level** `path(.children[]?)` call (not
inside `recurse`) evidently expands a trailing `Iterate`/`Optional(Iterate)`
path component into concrete per-element branches itself, downstream of
`resolve_node`. `resolve_recurse` bypasses that expansion by using
`resolve_node`'s output directly as `(path, value)` pairs to push onto its
stack, so it never benefits from it.

## Impact

Any `path(recurse(f))`, `del(recurse(f) | ...)`, or `(recurse(f)) |= ...`
where `f` ends in a bare `.foo[]`/`.foo[]?` over a container with 2+ elements
returns a wrong (too-short) result set — silently, not as an error. This is a
very common way to write `f` (it's exactly the shape used in #676's own
`recurse_f_dfs_sibling_order` golden, just at the value level, which is why
that golden didn't catch this).

## Suggested fix direction

Either make `resolve_recurse` route `f`'s resolution through whatever
top-level mechanism already expands a trailing bare `Iterate` component for
non-recurse `path(...)` calls, or make `needs_path_prepass` account for a
*trailing* `Iterate`/`Optional(Iterate)` requiring fan-out when the caller
(like `resolve_recurse`) needs concrete per-element branches rather than a
literal path-component list. Needs a regression test using the same
sibling-branching shape #676 used, but wrapped in `path(...)` /`del(...)`
instead of a bare value collection, since that's exactly the shape the
current test suite is missing for this code path.

## Found while

Reviewing #676/#635's DFS fix (`38c9a131`) and probing `resolve_recurse` for
the same order-sensitive coverage the value-level functions got, to check
whether the fix's own new goldens exercised the path-tracking twin too. They
don't, and probing turned this up as a separate, pre-existing bug.

newhoggy:
## Summary

`resolve_recurse` — the path-tracking evaluator behind `path(recurse(f))`, `del(recurse(f) | ...)`, and `recurse(f) |= ...` — silently dropped all but the first result whenever `f`'s trailing component was a bare `.foo[]`/`.foo[]?` (array/object iterate) yielding 2+ elements.

## Root cause

`needs_path_prepass` excluded `Expr::Iterate` from the "needs `resolve_node`'s help" set, on the theory that `walk_path` and friends already understand it natively — true for `walk_path` itself, but `resolve_seq`'s own "nothing needs prepass" fast path assumes something stronger: that its whole static tail is single-valued, and hands it to `value_after_components`, which silently collapses any component producing != 1 output to `Null`. `resolve_recurse` then treats that `Null` as "no children" and prunes it, so the fan-out never happens. A single-element container looked correct only by coincidence — `value_after_components`' 1-output case happens to be the right answer there.

## Fix

Route `Expr::Iterate` through the same per-element fan-out loop `resolve_seq` already uses for a genuine computed key, instead of excluding it from that decision. The static tail this fast path handles can, by `needs_path_prepass`'s own contract, never contain `Comma` (that would already force the loop), so this doesn't touch anything Comma-related, and `eval_generic.rs` has no parallel copy of this logic to keep in sync.

Fixes #682

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 54 binaries, 0 failed)
- [x] Diffed output against pinned real `jq` for: the issue's own repro, `del(recurse(f)|...)`-style usage, the #676 sibling-order shape wrapped in `path()`, and `Comma`-in-`f` (all match)
- [x] Confirmed no regression on value-level `recurse(f)` or standalone `path(.children[]?)`
- [x] Added `recurse_path_trailing_iterate_sibling_order` and `recurse_del_trailing_iterate_multi_element` golden cases
- [x] `omni-dev coverage diff` — 100% patch coverage (2/2 new lines)


newhoggy:
Code review surfaced a real, well-confirmed performance regression in the first version of this fix: making `needs_path_prepass` itself return `true` for `Expr::Iterate` also broadened `resolve_dynamic_indexes`'s outer guard — shared by `=`, `|=`, `del()`, and `path()`, not just `resolve_recurse`. That silently routed every bare-iterate path expression through the multi-branch resolver instead of the fast native-walker path, exposing latent O(n²) algorithms in `del()`'s path-dedup and `path()`'s per-step cloning. Measured by the review: `del(.foo[])` on a 200K-element array went from 0.04s to ~11.5s (~290x); `path(.foo[])` at the same size didn't finish in 4+ minutes (previously <0.01s).

Fixed in a follow-up commit: reverted `needs_path_prepass` to its original form, and introduced a separate `needs_fanout_pass` used only by `resolve_seq`'s own internal decision. `resolve_dynamic_indexes`'s outer guard (using the unchanged `needs_path_prepass`) still short-circuits before ever reaching `resolve_node` for a plain `.foo[]` with no computed key — the common case `del`/`path`/`=`/`|=` all hit — so their fast path is completely unaffected. `needs_fanout_pass` only changes behavior for calls that already reach `resolve_seq` for another reason: `resolve_recurse`'s unconditional call (the actual #682 fix), or the rare case of a genuine computed key coexisting with a trailing bare iterate in the same pipe.

Re-verified against the review's exact regression scenario (200K-element array): `del()` back to ~0.04s, `path()` ~0.17s (was 4+ min timeout). #682's original repro and both new goldens still pass. Also added a `debug_assert!` to `value_after_components` guarding the invariant this bug violated, and filed #829 for the review's other finding (no benchmark coverage exists for `=`/`|=`/`del()`/`path()` shapes, so a regression like this wouldn't be caught by CI).

Full verification (build/test/clippy x2/coverage) re-run after the narrowing and again after rebasing onto current `main` — all green, 90% patch coverage (9/10 new lines; the one uncovered line is the `debug_assert!`'s failure-message argument, which only evaluates if the invariant it guards is violated).

Issue #837, opened 2026-08-13
yq -C: compact sequence item's first key is missing its cyan color (colorize_yaml has no lookahead past '- ')


newhoggy:
**Severity: Low**

## Summary

`succinctly yq -C`'s simple single-pass text colorizer (`colorize_yaml` in `src/bin/succinctly/yq_runner.rs`) never colors the first key of a block-sequence item rendered in compact form (`- key: value`) — every other key in the document is colored cyan, but this specific token isn't. Purely cosmetic (terminal color only); the underlying YAML structure/data is unaffected.

Discovered while implementing #785, which fixed the compact-rendering bug itself; #785 didn't fix this pre-existing colorizer gap since it was previously unreachable (a compact-form sequence item is exactly what #785 added — see below), and fixing it properly needs lookahead the current colorizer design doesn't have. Not oracle-relevant: succinctly's `-C` uses its own simplified color scheme, not a byte-match of real yq's (which uses entirely different codes — magenta for values, no color at all for the list-marker dash, vs. succinctly's yellow dash / cyan keys / no value color).

## Repro

```
$ succinctly yq -C '.' <<< $'- a: 1\n  b: 2\n' | cat -v
-^[[36m a^[[0m: 1
  ^[[36mb^[[0m: 2^[[0m
```

(Manual color-code reading: the first `a` gets no `\x1b[36m` cyan-open at all; the second-line `b` does.)

## Root cause

`colorize_yaml`'s `at_key_start` flag drives whether the next non-whitespace character opens a cyan key-color span. The `'-' if at_key_start` arm (list-marker handling) unconditionally sets `at_key_start = false` after coloring the dash yellow — correct for every source shape *before* #785, where `- ` was always immediately followed by either a value (never colored) or a newline (which resets `at_key_start = true` on its own). #785's compact-rendering fix means `- ` can now be immediately followed by a *key* on the same line with no intervening newline, and the colorizer has no way to tell "key starts right after this dash" from "scalar value starts right after this dash" without lookahead past the current character — naively leaving `at_key_start = true` after the dash would instead miscolor plain scalar sequence items (`- 2`) as keys.

## Why this is separate from #785

#785 is entirely a structural re-serialization fix (byte-for-byte text output, no color codes involved) — verified by comparing ANSI-stripped output against real `yq`. This is a distinct, purely cosmetic gap in the (non-oracled) `-C` colorizer, newly *reachable* because of #785's fix but not part of what #785 was asked to fix.

[...]

Issue #836, opened 2026-08-13
yq: block scalar style (| / >) is lost on YAML re-serialization via the streaming/cursor path, not just OwnedValue


newhoggy:
**Severity: Medium**

## Summary

A YAML block scalar (`|` literal or `>` folded) loses its block style entirely on re-serialization, coming back out as a double-quoted string with embedded `\n` escapes instead. Discovered while verifying #785's fix didn't regress plain scalar sequence items; confirmed pre-existing on unmodified `main`, unrelated to and unaffected by #785's changes.

Unlike the already-tracked #739 ("`OwnedValue` has no room for YAML style metadata"), this reproduces on the **M2 streaming/identity path** (`YamlCursor::stream_yaml_value`, cursor-based, no `OwnedValue` involved at all) — so it's a distinct root cause, not simply another `OwnedValue` blind spot.

## Repro

```
$ printf -- '- |\n  line1\n  line2\n- next\n' | yq '.'
- |
  line1
  line2
- next
$ printf -- '- |\n  line1\n  line2\n- next\n' | succinctly yq '.'
- "line1
line2
"
- next
```

```
$ printf -- '- >\n  line1\n  line2\n- next\n' | yq '.'
- >
  line1 line2

- next
$ printf -- '- >\n  line1\n  line2\n- next\n' | succinctly yq '.'
- "line1 line2
"
- next
```

Reproduces identically as a plain top-level value, a mapping field's value, and a sequence item's value — not specific to sequences.

## Root cause

Not yet investigated in depth. `YamlCursor::stream_yaml_value`'s `YamlValue::String` arm (`src/yaml/light.rs`) has no branch for "was this scalar written as a block literal/folded scalar in the source" — it always calls `stream_yaml_string_value`, which picks a quoting style (plain/single/double) based on the *content*, with no path back to `|`/`>` at all. The cursor presumably has *some* way to know the scalar's original style (block vs flow, quoted vs plain) since parsing must distinguish them to interpret escaping/folding rules correctly — but that information evidently isn't threaded through to the re-serialization write path.

## Why this is separate from #739

[...]

Issue #835, opened 2026-08-13
yq: is_container() returns false for a sequence item's mapping value when the source uses dash-alone-then-indented style, truncating re-serialization


newhoggy:
**Severity: High**

## Summary

A block-sequence item whose value is a non-empty mapping, when the *source* YAML writes the item as a bare `-` on its own line followed by the indented mapping (rather than the compact `- key: value` form), re-serializes as just `-` with the entire mapping silently dropped — not merely misformatted, but genuinely truncated. Discovered while fixing #785 (a pure rendering/formatting issue); this is a distinct, more severe bug at the cursor-classification level, confirmed present on unmodified `main` (before and independent of the #785 fix).

## Repro

```
$ printf -- '-\n  a: 1\n  b: 2\n' | yq '.'
- a: 1
  b: 2
$ printf -- '-\n  a: 1\n  b: 2\n' | succinctly yq '.'
-
```

(Both the M2 streaming/identity path and the `-C`-forced DOM path reproduce this identically.) The parsed *data* is correct — `-o json` on the same input returns `[{"a":1,"b":2}]` — so this is purely a re-serialization bug, but unlike #785 it drops real content instead of just misplacing it.

## Root cause

`src/yaml/light.rs`'s `is_container()` (called by the private `is_yaml_cursor_container` helper used throughout `stream_yaml_value`'s Sequence arm) returns `false` for the sequence item's cursor in this specific shape, even though `cursor.value()` correctly resolves to `YamlValue::Mapping(..)` and `cursor.first_child().is_some()` is `true`. Confirmed via a temporary diagnostic test:

```
item_cursor.value()=Mapping(...) is_container()=false first_child().is_some()=true style()="" is_yaml_cursor_container=false
```

Since `is_yaml_cursor_container` short-circuits to `false`, `stream_yaml_value`'s Sequence arm takes the non-container ("- " + inline value) branch instead of the mapping-rendering branch, and the mapping content is never written.

The bug is specific to the "dash-alone-then-newline" (already non-compact) source style for a sequence item's mapping value — the equivalent compact-source form (`- a: 1\n  b: 2\n`) parses and renders correctly (`is_container()` returns `true` there). This suggests `is_container()`'s classification (likely reading a `ty`/type-tag array position keyed by `bp_pos`) is populated incorrectly, or reads the wrong `bp_pos`, specifically for this parse shape — a parser/index bug, not confined to rendering.

## Why this is separate from #785

[...]

Issue #765, opened 2026-08-10
yq: mapping-key trailing comment silently dropped from identity output when value is deferred to next line


newhoggy:
## Summary

Follow-up from #710's line-comment preservation work. When a mapping key's trailing `#` comment sits on the key's own line but the value is deferred to the next line (nested mapping/sequence), `succinctly yq` currently drops the comment entirely from identity output — with no warning, silent data loss.

```
$ printf 'a: # comment on key\n  b: 1\n' | succinctly yq '.'
a:
  b: 1
```

Real `yq` (v4.53.3) keeps it:

```
$ printf 'a: # comment on key\n  b: 1\n' | yq '.'
a: # comment on key
  b: 1
```

## Investigation (during #710 code review follow-up)

A code review of #710 initially flagged this as "the `.a | line_comment` getter should return `'comment on key'`, matching the existing trailing-same-line-comment scope." That expectation turned out to be **wrong** when checked against the pinned real `yq` binary:

```
$ printf 'a: # comment on key\n  b: 1\n' | yq '.a | line_comment'
(blank)
$ printf 'a: # comment on key\n  b: 1\n' | yq '.a | head_comment'
(blank)
$ printf 'a: # comment on key\n  b: 1\n' | yq '.a | foot_comment'
(blank)
$ printf 'a: # comment on key\n  b: 1\n' | yq '.a.b | line_comment'
(blank)
```

None of real `yq`'s comment getters expose this comment at any queryable path. It only survives identity output because `yq` re-serializes its full internal node tree, which stores the comment on the **key** node itself (not the value node that `.a` navigates to). Succinctly's `line_comment`/`CommentTree` design from #710 is entirely value-node-oriented (mirrors the anchor/style cursor pattern) — there's no concept of a comment owned by a *key* node today.

So succinctly's current `.a | line_comment` → `''` is already correct (matches real `yq`). The actual gap is narrower and different in kind: **identity/output preservation** for a comment that belongs to the key, not any getter.

## What a real fix needs

- A second, key-scoped comment capture (distinct from the existing value-scoped `line_comments`/`CommentTree`, since `.a | line_comment` etc. must keep returning blank for this case, matching real `yq`)
- Key cursors threaded into the write paths (`emit_yaml_value` in `src/bin/succinctly/yq_runner.rs`, `stream_yaml_value` in `src/yaml/light.rs`) so a mapping field's own key-comment can be appended right after `key:` when the value is deferred to the next line
[...]

PR #838, opened 2026-08-13
fix(yaml): compact-render mapping/sequence values inside block sequence items

### LINKED ISSUE #785: yq: a compact mapping/sequence value inside a block sequence item re-serializes as bare `-` + indented block, not `- key: value`
## Summary

A block sequence item whose value is a non-empty mapping or sequence always re-serializes as a bare `-` on its own line, followed by the value fully indented on the next line — even when the source YAML has it in real `yq`'s "compact" form, with the value's first line sharing the dash's line.

```
$ printf -- '- a: 1\n  b: 2\n' | yq '.'
- a: 1
  b: 2
$ printf -- '- a: 1\n  b: 2\n' | succinctly yq '.'
-
  a: 1
  b: 2
```

This is purely a formatting/re-serialization divergence — the parsed data is correct (`.` piped to `-o json` matches) — but it fires on a very common real-world shape (a list of records), e.g.:

```
$ printf 'containers:\n  - name: app\n    image: nginx\n  - name: sidecar\n    image: envoy\n' | yq '.'
containers:
  - name: app
    image: nginx
  - name: sidecar
    image: envoy
$ printf 'containers:\n  - name: app\n    image: nginx\n  - name: sidecar\n    image: envoy\n' | succinctly yq '.'
containers:
  -
    name: app
    image: nginx
  -
    name: sidecar
    image: envoy
```

Reproduces identically on both the identity/streaming path and any DOM-forcing filter (e.g. `select(true)`).

## Compounding effect on #765's key-scoped comment fix

Because this shape routes through different rendering code than a plain mapping field, #765's key-scoped comment preservation doesn't apply to it — the comment is dropped outright here, not just reformatted:

```
$ printf -- '- a: # comment on key\n    b: 1\n' | yq '.'
- a: # comment on key
    b: 1
$ printf -- '- a: # comment on key\n    b: 1\n' | succinctly yq '.'
-
  a:
    b: 1
```

## Code pointers

- Streaming path: `src/yaml/light.rs`, `YamlCursor::stream_yaml_value`'s block-style `YamlValue::Sequence` arm (around line 1543) — `is_yaml_cursor_container(&cursor)` is true for any non-empty mapping/sequence element and always takes the bare-`-`-then-newline branch, with no check for whether the element could instead render in the compact form (dash + first key/item on the same line).
- DOM path: `src/bin/succinctly/yq_runner.rs`, `emit_yaml_value`'s `OwnedValue::Array` block-style arm (around line 895) — same shape, `format!("{indent}-\n{item}{comment_suffix}")` unconditionally for any non-empty `Object`/`Array` element.

Fixing this will likely also need to thread the same key-scoped `key_comment` lookup (#765's `CommentTree::Object`'s third `IndexMap<String, String>` field) into whichever new compact-rendering branch replaces these, since the deferred-key-comment case above needs the same "comment right after `key:`, value starts on the next line" handling #765 already does for a *non*-sequence-item mapping field.

## Scope note

Filing this to make the gap trackable; found while reviewing #765. Not blocking #765, but likely a higher-impact gap than #765 itself given how common "list of mappings" YAML is (Kubernetes manifests, CI/CD configs, etc.) — worth prioritizing accordingly.

newhoggy:
## Summary

A block sequence item whose value is a non-empty mapping or sequence always re-serialized as a bare `-` on its own line, followed by the value fully indented on the next line, instead of real yq's "compact" form (`- key: value`, with subsequent fields/elements aligned under the first line's content). Fires on very common shapes — a list of records (Kubernetes manifests, CI/CD configs), and any nested-sequence-in-sequence-item.

```
$ printf -- '- a: 1\n  b: 2\n' | yq '.'
- a: 1
  b: 2
$ printf -- '- a: 1\n  b: 2\n' | succinctly yq '.'   # before this PR
-
  a: 1
  b: 2
```

Also fixed a compounding bug this issue's own report flagged: because the shape routed through different rendering code, #765's key-scoped comment preservation didn't apply, and the comment was dropped outright rather than just reformatted. Root-caused this to a genuine **parser** bug (not just a rendering gap): `parse_compact_mapping_entry` — used only for a block-sequence item's first field — never got #765's `maybe_capture_line_comment` call that its sibling `parse_mapping_entry` has, so the comment was never captured in the first place.

- **Streaming path**: `YamlCursor::stream_yaml_value`'s Sequence block-style arm (`src/yaml/light.rs`)
- **`--slurp` fast path**: `stream_yaml_sequence` (`src/yaml/light.rs`) — a third affected call site beyond the two named in the issue, found during investigation; also picked up a missing `.style() != "flow"` guard its own doc comment already promised but never had
- **DOM path**: `emit_yaml_value`'s `Array` arm (`src/bin/succinctly/yq_runner.rs`) — refactored from `depth: usize` to an explicit `indent: &str` parameter, since the compact continuation's `+2`-literal-character offset can't be expressed as a repetition count once `indent_str`'s own width can differ (`-I4`, `--tab`)
- **Parser**: `parse_compact_mapping_entry` (`src/yaml/parser.rs`) — one-line fix mirroring #765's existing capture in `parse_mapping_entry`

The compact continuation's alignment is always `+2` literal characters (the width of `- `), **never** `+ indent_spaces` — verified against real `yq` v4.53.3 across `-I1` through `-I6`: nested block indentation scales with `-I`, but compact alignment never moves.

An anchor written on the item's own line (`- &x\n  ...`) still occupies the compact slot and keeps the value deferred, matching real yq — `cursor.anchor()` only returns `Some` for that source shape; an anchor sharing the first key's own line (`- &x a: 1`) isn't captured by the cursor at all, a separate pre-existing gap.

## Follow-ups filed (out of scope for this PR)

- #835 (High) — `is_container()` returns `false` for a sequence item's mapping value when the source uses dash-alone-then-indented style, causing the *value to be dropped entirely* on re-serialization (not just misformatted). Confirmed pre-existing on unmodified `main`, unrelated to and unaffected by this fix.
- #836 (Medium) — block scalar (`|`/`>`) style is lost on re-serialization via the streaming/cursor path (distinct root cause from #739, which is `OwnedValue`-specific). Confirmed pre-existing on `main`.
- #837 (Low) — `yq -C`'s simplified, non-oracled colorizer doesn't color a compact item's first key (cosmetic only, newly *reachable* by this fix but not caused by it).

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 2356+ lib tests plus all integration suites green; one `jq_cli_tests` flake reproduced the known concurrent-cargo-contention pattern and passed clean in isolation)
- [x] Diffed output against pinned real `yq` v4.53.3 (matches `tests/data/yq-golden/YQ_VERSION`) for: basic compact mapping, list-of-records, comment-on-key, compact sequence-in-sequence-item, `-I4`/`-I6`, anchor-stays-deferred, plain scalar items (unaffected), flow-style items (unaffected) — on both the identity/streaming path and the DOM path (forced via `-C`, ANSI-stripped for structural comparison)
- [x] `./scripts/sync-yq-golden.sh --check` — no drift (no golden fixture asserts YAML-format output)
- [x] `cargo llvm-cov` + `omni-dev coverage diff` — 89.81% patch coverage; remaining gaps are a test-helper's defensive `panic!` branches (matching an existing sibling helper's established pattern), one diff-context artifact pointing at an unrelated pre-existing test, and `?`-operator error-propagation branches on `core::fmt::Write` calls that succeed unconditionally for a `String` sink
- [x] Updated 3 `light.rs` unit tests, 6 `yq_cli_tests.rs` integration tests, and 1 insta snapshot that pinned the old buggy output; added 4 new regression tests (2 for the parser comment-capture fix, 2 for the anchor-stays-deferred branches)

Fixes #785

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -1493,6 +1493,49 @@ fn yaml_quote_key(s: &str) -> String {
     }
 }
 
+/// Whether a compact block-sequence item's remaining source (the text
+/// right after `- `, as returned by `Chars::as_str()` at that point) opens
+/// with a mapping key rather than a scalar value — i.e. whether an
+/// unquoted `:` appears before the line ends.
+///
+/// `colorize_yaml`'s state machine colors a token *as it's written*, one
+/// `char` at a time, with no ability to go back and re-color something it
+/// already emitted — so at the position right after `- `, it has to decide
+/// up front whether what follows is a key (color it) or a value (don't),
+/// and the only way to tell is to look ahead. Before #785, `- ` was always
+/// followed by either a scalar value or a newline (a container value's
+/// mapping always started on its own line), so this ambiguity never arose.
+/// A nested compact marker (`- - 1`) recurses naturally: the caller
+/// re-invokes this same lookahead for the *inner* `-` too, so a mapping
+/// arbitrarily many `- ` markers deep (`- - a: 1`) still finds its `:` and
+/// colors `a`, while a purely scalar nested sequence (`- - 1\n  - 2`) does
+/// not color the inner marker - real, but narrow and, like the rest of
+/// this colorizer, not oracle-matched against real yq's own (differently
+/// coded) `-C` output, so left as a known residual gap rather than chased
+/// further (#785).
+fn compact_item_opens_with_key(rest_of_line: &str) -> bool {
+    let mut quote: Option<char> = None;
+    let mut chars = rest_of_line.chars();
+    while let Some(c) = chars.next() {
+        match quote {
+            Some(q) => {
+                if c == '\\' {
+                    chars.next(); // Skip the escaped character.
+                } else if c == q {
+                    quote = None;
+                }
+            }
+            None => match c {
+                '\n' => return false,
+                '"' | '\'' => quote = Some(c),
+                ':' => return true,
+                _ => {}
+            },
+        }
+    }
+    false
+}
+
 /// Colorize YAML output (basic ANSI colors).
 fn colorize_yaml(yaml: &str) -> String {
     let mut result = String::with_capacity(yaml.len() * 2);
@@ -1501,7 +1544,8 @@ fn colorize_yaml(yaml: &str) -> String {
     let mut at_key_start = true;
     let mut in_key = false;
 
-    for c in yaml.chars() {
+    let mut chars = yaml.chars();
+    while let Some(c) = chars.next() {
         if escape_next {
             result.push(c);
             escape_next = false;
@@ -1528,7 +1572,15 @@ fn colorize_yaml(yaml: &str) -> String {
                 at_key_start = false;
             }
             ':' if !in_string => {
-                result.push_str("\x1b[0m");
+                // Only close a color span that's actually open (`in_key`)
+                // - otherwise this `:` belongs to a value that was never
+                // colored (e.g. a quoted-key's colon after the closing
+                // quote already reset color, or a `:` inside an uncolored
+                // token), and unconditionally emitting a reset here would
+                // write an orphaned `\x1b[0m` with no matching open.
+                if in_key {
+                    result.push_str("\x1b[0m");
+                }
                 result.push(c);
                 in_key = false;
                 at_key_start = false;
@@ -1542,7 +1594,12 @@ fn colorize_yaml(yaml: &str) -> String {
                 result.push_str("\x1b[33m"); // Yellow for list markers
                 result.push(c);
                 result.push_str("\x1b[0m");
-                at_key_start = false;
+                // A real-yq "compact" block-sequence item (`- key: value`,
+                // #785) puts its mapping's first key directly after `- `
+                // on the same line, instead of on a fresh line (which
+                // already re-triggers `at_key_start` via the `\n` arm
+                // above) - see `compact_item_opens_with_key`'s doc comment.
+                at_key_start = compact_item_opens_with_key(chars.as_str());
             }
             _ if at_key_start && !c.is_whitespace() && !in_string => {
                 result.push_str("\x1b[36m"); // Cyan for keys
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -1083,7 +1083,7 @@ fn output_value<W: Write>(
     // For YAML output format (default)
     if config.output_format == OutputFormat::Yaml {
         // For YAML, scalars are printed without quotes by default (like -r in yq)
-        let body = emit_yaml_value(value, comments, config, 0, false);
+        let body = emit_yaml_value(value, comments, config, "", false);
         // Every non-root node's own trailing comment is appended by its
         // *parent* during `emit_yaml_value`'s recursion (see its Array/Object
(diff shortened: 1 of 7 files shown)


newhoggy:
## Review round addressed

The mandatory review (5 agents, findings relayed via the orchestrating session) found a real, confirmed-by-building blocker plus several related cleanup items. Summary of what changed:

**Blocking (fixed):**
- **`--tab` byte-inconsistency**: the streaming path and DOM path disagreed on actual output bytes for the exact compact-form shape this issue is about. Root cause (in two layers): (1) the compact alignment offset was being fed through `unit`-repetition instead of written as literal ASCII spaces, and (2) even after fixing that, a `(current_indent: usize, extra_spaces: usize)` two-counter representation collapses to a fixed unit-then-extra write order regardless of which kind of step chronologically happened first — wrong whenever an ordinary indent step follows a compact one (a very common shape, invisible under the default space unit since the two commute, only visible once `unit != ' '`). Fixed by switching `stream_yaml_value`/`stream_yaml_sequence` (`light.rs`) and `stream_owned_value_yaml` (`stream.rs`) to an explicit indent *string*, mirroring the DOM path's (`emit_yaml_value`) already-correct plain-concatenation design.
- **4th render path missed by the original fix**: `first(f)`/`last(f)`'s cursor-loss fallback (`stream_owned_value_yaml` in `src/jq/stream.rs`) had the same bare-`-`-then-newline bug, undiscovered because it's not one of the 3 sites the issue named. Fixed with the same compact-rendering + string-indent approach.
- **Colorizer regression**: `yq -C` never colored a compact item's first key (and emitted an orphaned ANSI reset). Fixed with a bounded, quote-aware lookahead (`compact_item_opens_with_key`) that decides key-vs-value before coloring, since the single-pass colorizer can't backtrack.

**Also fixed while in the same code:**
- Eliminated a double `cursor.style()` call per compact-eligible element (once in the gating check, once again in the callee's own dispatch) via a `known_not_flow` parameter — real cost on the flagship streaming path.
- `stream_yaml_sequence` (`--slurp`) never called `write_line_comment` for its own items at all, silently dropping a sequence item's trailing comment via `--slurp` specifically — pre-existing, but the code's own doc comment falsely claimed exact parity with `stream_yaml_value`. Now actually true.

**Verification after the fix:**
- Full test suite green (2358 lib tests + all integration suites, 483 `yq_cli_tests`)
- Both required clippy invocations clean
- `--tab` self-consistency tests added (single-level, nested, and the compact-then-normal ordering that exposed the deeper bug) — real yq has no `--tab` flag at all, so there's no external oracle; the bar is the streaming and DOM paths agreeing byte-for-byte
- Colorizer coverage tests for the lookahead's quote-handling branches
- Patch coverage: 95.69% (100% on the two files the review flagged hardest — `yq_runner.rs`, `stream.rs`); remaining gaps are `?`-operator closing-brace artifacts (unreachable without a failing `Write` sink) and pre-existing test-helper defensive panics, both already the accepted pattern elsewhere in this codebase
- `./scripts/sync-yq-golden.sh --check` — no drift
- Rebased clean on current `origin/main`, force-pushed

Given the scope of the fix (a real correctness bug, not just cosmetic), happy to take another review pass on the corrected diff if useful before merge.

PR #839, opened 2026-08-13
fix(jq): evaluate sub()/gsub() replacement per match, not once against the input

### LINKED ISSUE #826: jq: sub()/gsub() replacement's \(.name) capture-group interpolation errors instead of binding to the match's captures
**Severity: Medium**

## Summary

jq's `sub(re; replacement)` and `gsub(re; replacement)` evaluate `replacement` **once per match**, with `.` bound to an object of that match's named captures — this is how jq lets a replacement reference what it just matched, e.g. `sub("(?<x>[aeiou])"; "[\(.x)]")`. succinctly instead evaluates `replacement` once, up front, against the *original input value* (before any matching happens), so any `\(.x)` interpolation referencing a named capture fails outright unless `.x` happens to also be a valid field on the original input.

```console
$ echo '"hello world"' | jq 'sub("(?<x>[aeiou])"; "[\(.x)]")'
"h[e]llo world"
$ echo '"hello world"' | succinctly jq 'sub("(?<x>[aeiou])"; "[\(.x)]")'
jq: error (at <stdin>:1): Cannot index string with string "x"
```

Same failure for `gsub`:

```console
$ echo '"hello world"' | jq 'gsub("(?<x>[aeiou])"; "[\(.x)]")'
"h[e]ll[o] w[o]rld"
$ echo '"hello world"' | succinctly jq 'gsub("(?<x>[aeiou])"; "[\(.x)]")'
jq: error (at <stdin>:1): Cannot index string with string "x"
```

Not specific to the `g` flag — reproduces on plain `sub`/`gsub` with no flags at all.

## Root cause

`builtin_sub_with_flags` (`src/jq/eval.rs:7212-7222`) and the equivalent block in `builtin_gsub_with_flags` (`src/jq/eval.rs:~7300-7310`) both evaluate `replacement_expr` via `eval_single::<W, S>(replacement_expr, value.clone(), optional)` — i.e. against `value`, the **original input**, before any regex matching happens. When `replacement_expr` is a jq string literal containing `\(.x)`, that interpolation is evaluated as an ordinary field access on `value`, which fails (or silently succeeds with the wrong data) whenever `.x` isn't a valid path on the original input.

There *is* a `convert_jq_replacement` helper (`src/jq/eval.rs:7556`) that textually rewrites `\(.name)` → `${name}` for the regex crate's `caps.expand`, and a per-match `caps.expand(replacement.as_str(), &mut out)` call further down — but both are unreachable for this shape, since `eval_single` already errors out on `.x` before `replacement` even reaches `convert_jq_replacement`. The rewrite-to-`${name}` mechanism only works for interpolations that also happen to survive being evaluated as an ordinary expression against the original input (e.g. `\(.some_field_on_the_actual_input)`), which is a different, narrower feature than what it looks designed for.

## Impact

Any `sub`/`gsub` replacement that references a named capture group via `\(.name)` — the documented jq idiom for "use what I just matched" — fails with a misleading type error (or, worse, silently uses the wrong value if `.name` happens to also resolve on the original input). This is a common, useful pattern (e.g. rewriting/wrapping matched substrings using their own captured pieces) that's currently unusable.

## Suggested fix direction

`replacement_expr` needs to be evaluated **per match**, with `.` bound to that match's captures object (`capture_object(&re, caps)`, already used elsewhere in this file for `capture(re)`), not once against the original input. This likely means restructuring both `builtin_sub_with_flags` and `builtin_gsub_with_flags` to defer replacement evaluation until inside the match loop — closer to how `capture_object`/`build_match_object` already build a captures object per match — rather than pre-computing a single string via `eval_single` + textual `${name}` substitution. `convert_jq_replacement` and its `${name}`-based `caps.expand` path likely become dead code once replacement is evaluated as a real per-match jq expression instead of a regex-crate template string.

## Why this is separate from #728

#728 was specifically about `sub(...;...;"g")` ignoring the `g` flag (single- vs global-replace). This bug is unrelated to flags entirely — found while adding broader golden coverage for #728's fix (a `"g"` + capture-group-interpolation test case), but reproduces identically on plain `sub`/`gsub` with no flags.

## Test coverage

Pinned as a known failure: `regex_sub_g_capture_group` in `tests/data/jq-golden-known-failures.txt` / `tests/data/jq-golden/cases/regex_sub_g_capture_group/` (added alongside #728's PR).

newhoggy:
## Summary

- `sub(re; replacement)`/`gsub(re; replacement)` (and their `; flags` forms) pre-evaluated `replacement` **once**, up front, against the *original input value* — before any regex matching happened. Any `\(.name)` interpolation referencing a named capture group therefore errored (`Cannot index string with string "x"`) instead of binding to that match's captures, since jq evaluates the replacement filter **once per match**, with `.` bound to an object of that match's named captures.
- Add `eval_sub_replacement`, which builds the same captures object `capture(re)` already uses (`capture_object`) and evaluates `replacement_expr` against it per match, and `stitch_replacements_evaluated`, which drives that across a global match list the way the old `stitch_replacements`/`Captures::expand` template-substitution helper did.
- `convert_jq_replacement` (the `\(.name)` → `$name` textual rewrite feeding the regex crate's template expansion) is now dead code and removed along with `stitch_replacements`, since replacement text now comes from real jq evaluation instead of a regex-crate template string.
- The 2-arg `sub`/`gsub` forms (`builtin_sub`/`builtin_gsub`) and `builtin_gsub_with_flags` now delegate to `builtin_sub_with_flags` (`sub($re; str; "")` / `sub($re; str; flags + "g")`, matching jq's own definition) instead of duplicating the same pattern/input/regex-build/per-match logic a second and third time.

```console
$ echo '"hello world"' | jq 'sub("(?<x>[aeiou])"; "[\(.x)]")'
"h[e]llo world"
$ echo '"hello world"' | succinctly jq 'sub("(?<x>[aeiou])"; "[\(.x)]")'   # before: error
"h[e]llo world"                                                            # after: matches jq
```

Fixes #826

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, several times across the review-fix cycle — the documented concurrent-cargo-lock flake surfaced on unrelated tests twice, clean on retry both times)
- [x] Diffed output against the pinned real `jq` (jq-1.7.1) directly for the issue's repro plus a broad sweep of extra cases (multiple named captures, unicode, case-insensitive flag, no-match, replacement-type-error, multi/zero-output replacement filters, `halt_error` in the replacement position) — all match, including re-verification after the review-driven fixes below
- [x] Unpinned the `regex_sub_g_capture_group` known-failure golden case (now passes) and added 4 new golden cases (`regex_sub_capture_group`, `regex_gsub_capture_group`, `regex_sub_capture_group_multi`, `regex_sub_no_match`) generated from the pinned jq oracle via `./scripts/sync-jq-golden.sh`
- [x] Added self-contained CLI tests for divergent-wording/no-oracle branches (replacement type-error, `?`-swallowed, `isvalid`-forced `optional:true`, explicit-`"g"`-flag error propagation, multi-value-replacement stopgap for both `sub` and `gsub`)
- [x] `cargo llvm-cov --features cli,simd,regex,serde --workspace --json` + `omni-dev coverage diff` → 100% patch coverage (54/54 new lines)

## Review-driven fixes

Two review rounds surfaced one real regression and one real duplication, both fixed as a follow-up commit:

- **Multi-value replacement regression, fixed.** A replacement filter yielding more than one output per match (`sub("a"; "x","y")`) went from the pre-#826 code's non-fatal "take the first value" behavior to a hard type error (or, under `?`, silently dropping all output) once routed through `eval_owned_expr`'s array-collapse policy. `eval_sub_replacement` now uses `result_to_owned` over `eval_owned_input` instead, restoring the old take-first stopgap without touching the shared `eval_owned_expr`/`eval_owned_expr_ctrl` functions other callers rely on.
- **`builtin_gsub_with_flags` duplication, fixed.** It now delegates into `builtin_sub_with_flags` with a forced `"g"` flag instead of duplicating ~25 lines of pattern/input/regex-build boilerplate — `'g'` is a no-op in `build_regex`'s own flag parsing, so this reliably selects `sub`'s global-replace arm, matching gsub's own unconditional "replace every match" behavior.
- Minor: `builtin_sub_with_flags`'s single-match arm now calls `re.captures(&input)` instead of the equivalent inline `dv_captures_at(...)`, matching the existing pattern used at the same call shape elsewhere in the file.

## Note on scope

Full jq-correct handling of a replacement filter producing zero or multiple outputs per match — jq's own `reduce`-based `sub`/`gsub` definition forks the whole call across every combination the matches produce, and has a genuinely surprising cascading-null-accumulator effect on `empty` (verified live against jq 1.7.1: `gsub(...; empty)` on one match silently drops surrounding *unrelated* text too, not just that match) — is out of scope for #826, which is specifically about named-capture binding for the single-string-per-match case. This PR's stopgap (take the first output, matching what the pre-#826 code already did) avoids making that gap any worse than it already was; full fan-out support is tracked as a separate follow-up: #840.


newhoggy:
Filed #840 for a separate, pre-existing edge case found while verifying this fix: jq's `reduce`-based `sub`/`gsub` definition has surprising cascading behavior when the replacement filter itself emits zero or multiple outputs per match. Neither the old code nor this PR's fix replicates that (both diverge, in different ways) — out of scope for #826, which is specifically about `\(.name)` capture-group binding.

newhoggy:
Addressed both review rounds:

**Real regression (multi-value replacement filter)** — fixed. \`eval_sub_replacement\` now uses \`result_to_owned\` over \`eval_owned_input\` (take-the-first-output policy) instead of \`eval_owned_expr\` (array-collapse policy), restoring the pre-#826 code's non-fatal "take first value" behavior for a replacement filter that yields more than one output per match (\`sub("a"; "x","y")\`). Verified: no longer hard-errors, no longer silently drops all output under \`?\`. Full jq-correct multi-value fan-out (forking the whole call per combination, like real jq does) remains out of scope — filed as #840 alongside the earlier cascading-null-accumulator finding, since both are the same underlying gap (replacement filter producing something other than exactly one string per match). Added two new self-contained tests (\`test_sub_replacement_multi_value_takes_first_value\`, \`test_gsub_replacement_multi_value_takes_first_value\`) guarding the stopgap specifically, since there's no single jq output to pin as a golden fixture here.

**Duplication + misleading doc comment** — fixed together. \`builtin_gsub_with_flags\` now delegates to \`builtin_sub_with_flags\` with a forced \`"g"\` flag instead of duplicating ~25 lines of pattern/input/regex-build boilerplate; this also makes \`builtin_gsub\`'s existing doc comment (which already claimed the \`sub($re; str; "g")\` equivalence) accurate, since there's now a real delegation chain backing it up.

**Minor cleanup** — done. \`builtin_sub_with_flags\`'s single-match arm now calls \`re.captures(&input)\` instead of the equivalent inline \`dv_captures_at(...)\`, matching the existing pattern elsewhere in the file.

Re-verified after these fixes: full test suite green (one confirmed transient cargo-lock-flake retry, clean on retry), both required clippy invocations clean, oracle diff re-run across the original ~30 cases plus the multi-value scenarios (all match either real jq or the documented stopgap), 100% patch coverage (54/54 new lines).

The "per-match reindex cost" item (informational, not a bug) is left as-is per the review's own conclusion — correct and required by jq's per-match evaluation semantics, consistent with the existing `builtin_paths_filter` precedent for this trade-off.

Issue #739, opened 2026-08-10
yq: OwnedValue has no room for YAML style metadata — mutating queries drop style on touched AND untouched nodes


newhoggy:
**Severity: Medium** — output silently discards information present in the input, with no flag to opt out and no diagnostic. Split out from #707 (which covers the pure-navigation/identity-query case — see #707's fix).

## Summary

Any `yq` query that requires constructing a new/modified value — assignment (`=`, `+=`, ...), `del()`, object/array construction (`{...}`, `[...]`), `map`, arithmetic, and generally anything `can_use_m2_streaming()` rejects — discards YAML style (flow vs. block, quote style) for **every** node in the output, including sibling nodes the query never touched. This is strictly broader than #707's identity-query gap: it also drops scalar quote style, not just container flow style.

## Repro (verified against installed `yq` v4.53.3, pinned version used by `tests/data/yq-golden/YQ_VERSION`)

An edit to one field reformats an unrelated sibling's quote style:

```
$ printf "a: 'single'\nb: 1\n" | yq '.b = 2'
a: 'single'
b: 2
$ printf "a: 'single'\nb: 1\n" | succinctly yq '.b = 2'
a: single
b: 2
```

Same loss on plain object construction (no assignment involved at all):

```
$ printf "a: 'single'\nb: \"double\"\n" | yq '{a, b}'
a: 'single'
b: "double"
$ printf "a: 'single'\nb: \"double\"\n" | succinctly yq '{a, b}'
a: single
b: double
```

## Root cause

Traced while investigating #707. Any expression outside `can_use_m2_streaming()`'s allowlist (`src/bin/succinctly/yq_runner.rs:1007`) routes through `to_owned()` (`src/jq/eval_generic.rs:37`), which converts a `YamlCursor` into `OwnedValue` (`src/jq/value.rs:202`) — a JSON-shaped enum (`Null`/`Bool`/`Int`/`Float`/`NumberLiteral`/`String`/`Array`/`Object`) shared verbatim with the JSON evaluator. `OwnedValue` has **no field capable of carrying YAML style metadata** at all, so the information is gone structurally, not just unused, by the time `emit_yaml_value` (`src/bin/succinctly/yq_runner.rs:612`) serializes the result. `emit_yaml_value` then:

- always renders containers as block style unless already inside a flow-styled ancestor (`in_flow` starts `false` at the top level and is only ever propagated downward, never set from source style — `src/bin/succinctly/yq_runner.rs:561,652-694`)
[...]

PR #851, opened 2026-08-13
fix(yaml): preserve block scalar style (|/>) on YAML re-serialization

### LINKED ISSUE #836: yq: block scalar style (| / >) is lost on YAML re-serialization via the streaming/cursor path, not just OwnedValue

newhoggy:
## Summary

Fixes #836: a YAML block scalar (`|` literal or `>` folded style) lost its
block style entirely on re-serialization via `succinctly yq`, coming back
out as a double-quoted string with embedded `\n` escapes instead. This
reproduced on the M2 streaming/cursor identity path
(`YamlCursor::stream_yaml_value` in `src/yaml/light.rs`), not the
`OwnedValue`/DOM path — a distinct root cause from the already-tracked
#739 ("OwnedValue has no room for YAML style metadata").

### Root cause

`stream_yaml_value`'s `String` arm always called
`stream_yaml_string_value`, which only sees the quoting-relevant
`YamlString` variant (`DoubleQuoted`/`SingleQuoted`/`Unquoted`/block), not
the cursor's own source position, and falls back to smart-quoting a
decoded block scalar. A block scalar's decoded value almost always
contains a raw `\n`, which smart-quoting always double-quotes.

### Fix

`self.style()` already resolves the source style (`"literal"`/`"folded"`)
from the cursor's own text position. The `String` arm now consults it and
re-emits `|`/`>` with:

- A minimal chomping indicator (`chomping_indicator`) derived from the
  decoded value's own trailing-newline count, not copied from the
  source's original indicator.
- Content derived empirically against the pinned real `yq` oracle
  (v4.53.3) across the full clip/strip/keep × embedded/trailing
  newline-run space (`widen_folded_breaks`) — folded style specifically
  needs every *embedded* run of `\n` widened by one so the fold survives
  re-parsing (a single break between two equally-indented lines folds
  back to a space per YAML 1.2 §8.1.3; only a blank line preserves it).

Falls back to the pre-existing smart-quoting path in three cases, all
verified against the oracle:
- Flow/compact context (`indent_spaces == 0`) — YAML has no `|`/`>` syntax
  inside `[...]`/`{...}`.
- Empty decoded content — real yq drops block style here too.
- Any content line with a trailing space — real yq avoids block style
  here (trailing whitespace is invisible/fragile in block form; a
  trailing tab is fine, only a trailing space disqualifies).

### Other call sites checked

- `stream_yaml_sequence` (the `--slurp` fast path) delegates each element
  to the same `stream_yaml_value`, so it inherits the fix with no changes
  of its own (added a regression test).
- `write_yaml_child_inline` (flow-context children) needs no change — YAML
  has no way to write a block scalar inside a flow collection anyway.
- `stream_owned_value_yaml` in `src/jq/stream.rs` (the `OwnedValue`/DOM
  path, reached by queries that force materialization, e.g. `map(.)`) is
  confirmed **still affected** and deliberately left alone — `OwnedValue::
  String` has no field to carry style metadata at all, which is exactly
  #739's tracked scope, not this issue's.

### Follow-up filed

A separate, pre-existing, broader gap surfaced during verification: a
*bare* top-level or query-navigated scalar document root (e.g. `yq '.'`
on a bare `"hello"` document, or `yq '.a'` navigating to any styled
scalar) doesn't drop its styling the way real yq does — real yq strips
quotes/block-indicators entirely for a scalar that's the whole output,
regardless of style (quoted, plain, or block). This affects *all* scalar
styles, not just block scalars, so it's out of scope here; filing as a
separate issue.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde --lib` (unit tests, incl.
      updated pre-existing tests that had pinned the old buggy output)
- [x] `cargo test --features cli --test yq_golden_tests` (new
      `block_scalar_style_reserialize` golden case, captured from the
      pinned real yq v4.53.3 via `./scripts/sync-yq-golden.sh`)
- [x] `cargo test --features cli --test yq_cli_tests` (new direct unit
      tests for literal/folded × clip/strip/keep × sequence-item/mapping-
      field/`--slurp`/embedded-blank-line/empty/trailing-space shapes)
- [x] Manual diff against the pinned real `yq` oracle for the exact
      repros in #836, plus ~20 additional hand-derived edge cases (clip/
      strip/keep chomping, embedded vs. trailing newline runs, trailing
      space/tab, empty content, nested depth) — all byte-for-byte
      identical
- [x] `cargo llvm-cov --features cli,simd,regex,serde --workspace
      --html` — 100% line coverage on every new function/branch this PR
      adds


newhoggy:
## Review-driven fixes (2nd commit)

Independent review surfaced several issues in the original fix, all now addressed in `8980f4a4`:

**Two confirmed data-corruption/data-loss bugs (highest priority):**
1. **Explicit indent indicator (`|2`/`>2`) was silently dropped.** Re-parsing the resulting output would auto-detect the wrong content indentation and truncate/misparse the rest of the document. Fixed by deriving and re-emitting the indicator digit whenever the decoded content's first line has literal leading whitespace of its own.
2. **A bare top-level/navigated scalar result (e.g. `.a` field projection) lost its content entirely** — `stream_yaml_value` receives an empty `indent` at the document root (no parent loop has computed a deeper one), and block-style content there produced invalid zero-indent output that silently decoded back to `""`. Fixed by falling back to quoting (lossless) at that position, matching pre-#836 behavior there.

**Other findings addressed:**
- Astral characters (U+10000+, most emoji) now fall back to quoting, matching real yq's own rule.
- Avoids double-decoding a disqualified block scalar; matches on `s`'s own enum discriminant instead of the separately-derived `self.style()`; `widen_folded_breaks` now borrows instead of allocating when there's nothing to widen.
- `chomping_indicator` now returns the file's existing `ChompingIndicator` enum instead of ad hoc strings; content computation uses `Cow` instead of a hand-rolled borrowed/owned split.
- Restored `needs_yaml_quoting`/`looks_like_yaml_number` test coverage that the original fix's test rewrite incidentally dropped.

**Two more corruption bugs found via my own systematic verification** (a style × chomping × context sweep against the pinned oracle, run precisely because the review findings suggested the original fix's edge-case coverage was thin):
- A redundant guard in `widen_folded_breaks` skipped embedded-run widening entirely for any strip-chomped folded scalar with an embedded blank line.
- The auto-detect trailing-run "+1" quirk turned out to have a second trigger condition (whether the *last* content line is itself more-indented), not just "no explicit indent digit" — missing it reintroduced a spurious blank line.

**One deliberate, documented divergence remains**: real yq's own encoder has a confirmed non-idempotent bug — it asymmetrically widens an embedded run between a plain line and a more-indented one only when the plain line comes first (verified: round-tripping through real yq `.` twice changes the decoded value). This is not replicated; this crate's own output stays lossless there instead, verified via round-trip for every sweep-found case, not just spot-checked.

All fixes re-verified against the pinned real yq oracle (v4.53.3): the original targeted repros, two systematic sweeps (144 structured + 900 randomized cases), and explicit round-trip losslessness checks for every case where this PR's output deliberately diverges from the oracle.

newhoggy:
## Coverage note

Patch coverage: **98.65%** (146/148 new lines). The 2 uncovered lines (`src/yaml/light.rs:1558` and `:6257`) are both syntactic closing braces/parens on control-flow paths where every branch already returns early:

- `1558` closes the `if let Ok(decoded) = s.as_str() { ... }` block in `stream_yaml_value`'s `String` arm - every code path inside it ends in an explicit `return` (either `stream_yaml_block_scalar(...)` or the unconditional `stream_yaml_block_scalar_quoted(...)` fallthrough), so control can never reach this specific brace as a "fall to the bottom" - it's dead by construction, not by a missing test.
- `6257` closes a `match` expression passed directly as the argument to `out.write_str(...)` inside `stream_yaml_block_scalar`. All three match arms (`ChompingIndicator::Clip`/`Strip`/`Keep`) are independently confirmed covered (verified via the HTML coverage report's per-region hit counts).

Both are a known llvm-cov line-attribution artifact for this shape of code (also present on several pre-existing, unrelated lines elsewhere in this same file before this PR), not a real behavioral gap.

Issue #832, opened 2026-08-13
jq: path()'s try/catch loses the resolved prefix when the catch handler itself fails as a path expression


newhoggy:
**Severity: Low**

## Summary

`resolve_node`'s `Expr::Try` arm (`src/jq/eval.rs`) loses whatever path branches it had already resolved before a `try`'s body failed, if the `catch` handler *itself* then fails to resolve as a path expression (or raises any other error/halt). Real jq keeps the prefix; succinctly discards it.

## Repro

```
$ echo '{"a":1,"x":5}' | jq -c 'path(try (.a, .x[0]) catch "x")'
["a"]
jq: error (at <stdin>:1): Invalid path expression with result "x"
$ echo '{"a":1,"x":5}' | succinctly jq -c 'path(try (.a, .x[0]) catch "x")'
jq: error (at <stdin>:1): Invalid path expression with result "x"
```

Real jq: `["a"]` on stdout (the `.a` branch, resolved before `.x[0]` failed), then the catch handler's own `"x"` fails as a path expression (#530's "Invalid path expression" wording) and the process exits 5. succinctly: same error and exit code, but the `["a"]` prefix never reaches stdout at all.

Also reproduces via `break` instead of a plain resolution failure (found while adding coverage for #824):

```
$ echo '{"a":1}' | jq -c 'label $out | path(try (.a, break $out) catch error("y"))'
["a"]
jq: error (at <stdin>:1): y
$ echo '{"a":1}' | succinctly jq -c 'label $out | path(try (.a, break $out) catch error("y"))'
jq: error (at <stdin>:1): y
```

Same shape: the prefix `["a"]` is missing from succinctly's stdout; the diagnostic and exit code otherwise agree.

## Root cause

`resolve_node`'s `Expr::Try` arm (`src/jq/eval.rs`), in both its `Error` and `Break` catch-with-handler branches, builds `let mut out = prefix;` and then does:

```rust
out.extend(resolve_against_cow::<S>(catch_expr, ...)?);
Ok(out)
```

The trailing `?` returns early with `Err(...)` the moment `resolve_against_cow` itself fails, which discards the local `out` (and the `prefix` it already holds) entirely — instead of returning `Err((out, escape))` so the caller can still stream the prefix ahead of the failure, the same way every other failure site in this resolver already does (e.g. the sibling `Expr::Comma`/`Expr::As` arms explicitly thread their accumulated prefix through the `Err` side).

## Why this is separate from #824

[...]

Issue #495, opened 2026-07-30
jq: repeat(f) silently swallows an error raised inside f — no error, no output, exit 0


newhoggy:
**Severity: High** — total silent failure: not even a partial-output/error-message tradeoff, nothing happens at all.

Found alongside #494 while investigating #464 (`nth(n; expr)` swallowing errors raised inside its argument).

```
$ echo 5 | jq -c 'repeat(if . > 3 then error("boom") else .+1 end)'
(stdout empty)
jq: error (at <stdin>:1): boom
# exit code 5

$ echo 5 | succinctly jq -c 'repeat(if . > 3 then error("boom") else .+1 end)'
(stdout empty)
(stderr empty)
# exit code 0
```

succinctly produces no output, no error message, and exits 0 — indistinguishable from a filter that legitimately produced nothing.

### Where

`eval_repeat` (`src/jq/eval.rs:8351-8375`) drives its loop with something like `Err(_) => break` on failure from the per-iteration evaluation — this discards the error/break signal outright rather than returning `QueryResult::Error`/`QueryResult::Break`, and unlike the `eval_limit`/`eval_while`/`eval_foreach` bug in #494 (which loses already-*produced* output), here the error is dropped before ever being surfaced at all.

### Note

This is a distinct failure mode from #494 (silent value loss vs. total silent swallow) even though both live in the same family of stream-consuming builtins, so filing separately. Fix likely needs `eval_repeat` to propagate `Error`/`Break` from its per-iteration evaluation instead of treating any `Err` as plain loop termination.

PR #853, opened 2026-08-13
fix(jq): keep recurse(f)'s own partial fan-out before a later error

### LINKED ISSUE #842: jq: recurse(f) drops f's own partial fan-out when a later f output errors (value + path evaluators)
**Severity: Low**

## Summary

`recurse(f)`/`recurse(f; cond)` drops `f`'s own partial fan-out output when a *later* output of that same `f` call errors, in both the value-position evaluator (`builtin_recurse_f`, `src/jq/eval.rs`) and the path-position evaluator (`resolve_recurse`, `src/jq/eval.rs`). Real jq's generator semantics stream every output `f` already produced at a given node before the error that stops the whole traversal propagates; succinctly discards that node's own partial `f` output entirely and only keeps output from nodes visited *before* the one where `f` errored.

## Repro

Value position, streamed (not collected into `[...]`, which is atomic in both jq and succinctly and not part of this bug):

```
$ echo '{"a":1,"b":2}' | jq -c 'recurse(if . == {"a":1,"b":2} then (.a, .b[0]) else empty end)'
{"a":1,"b":2}
1
$ echo '{"a":1,"b":2}' | succinctly jq -c 'recurse(if . == {"a":1,"b":2} then (.a, .b[0]) else empty end)'
{"a":1,"b":2}
```

Real jq prints the root (`{"a":1,"b":2}`) *and* `1` (the `.a` branch of `f`'s fan-out, produced before `.b[0]` errors) before the error; succinctly prints only the root, dropping the `1`.

Path position (what `resolve_recurse` implements), same shape:

```
$ echo '{"a":1,"b":2}' | jq -c 'path(recurse(if . == {"a":1,"b":2} then (.a, .b[0]) else empty end))'
[]
["a"]
$ echo '{"a":1,"b":2}' | succinctly jq -c 'path(recurse(if . == {"a":1,"b":2} then (.a, .b[0]) else empty end))'
[]
```

Real jq prints `[]` (root) *and* `["a"]` (the `.a` branch, resolved before `.b[0]` failed) before erroring; succinctly prints only `[]`.

## Root cause

`resolve_recurse` (`src/jq/eval.rs`), inside its main loop:

```rust
let children = match resolve_against_cow::<S>(f, current) {
    Ok(children) => children,
    Err((_prefix, e)) => return Err((outputs, e)),
};
```

`_prefix` here is `f`'s own partially-resolved branches (relative to `current`, not absolute from the document root) before `f` itself failed — and it's discarded rather than folded into `outputs`. The doc comment directly above this site says this is deliberate, "matching `eval_owned_multi`'s all-or-nothing contract for `builtin_recurse_f`/`builtin_recurse_cond`" — and indeed `builtin_recurse_f` (value position, same file) has the exact same behavior via its own `eval_owned_multi`-based call to `f`, so the two evaluators are consistently *mirroring* each other. But live testing against jq 1.7.1 (above) shows this mirrored behavior itself diverges from real jq's actual semantics: jq's `recurse` is defined as `def r: ., (f | r); r;`, and `f | r`'s `|` streams each output of `f` as it's produced (real jq generator/backtracking semantics), so a later output of `f` erroring does not retroactively un-emit an earlier output — this is the same principle already fixed elsewhere in this file for other "later sibling errors" cases (#530, #636, #824), just not yet applied to `f`'s own within-node fan-out inside `recurse`.

## Why this is separate from #832

#832 (and this comment thread's investigation of it) is about `resolve_node`'s `Expr::Try` arm's `resolve_catch` helper losing the path prefix when a `catch` handler itself fails — a completely different function with no shared call graph with `resolve_recurse`/`builtin_recurse_f`. This was found by a code-review pass on #832's PR (#841) that went looking for the same *class* of bug (an accumulated prefix/output dropped on an escaping `Err`) elsewhere in the file, not by working on `recurse` directly.

It's also **not a simple one-line fix** the way #832's was: `resolve_recurse`'s `_prefix` values are relative to `current` (they need `prefix.clone(); path.extend(child_components)` treatment, the same as the adjacent `Ok` arm already does at lines ~10320-10321) before they can be folded into `outputs`, which holds absolute, document-root-relative paths — naively doing `outputs.extend(_prefix)` would silently produce corrupted (incomplete) paths, a new bug rather than a fix. And because `builtin_recurse_f`/`builtin_recurse_cond` and `resolve_recurse` are deliberately kept in parity (per the existing doc comment), a real fix needs to preserve that parity rather than fixing only the path-position side.

## Suggested fix direction

Fix both evaluators together to keep their documented parity:
- `resolve_recurse`: fold `f`'s own partial fan-out (rebased onto the current node's absolute `prefix`, the same way the `Ok` arm already rebases) into `outputs` before returning `Err`, instead of discarding it.
- `builtin_recurse_f`/`builtin_recurse_cond` (value position): the analogous fix using whatever partial-output-preserving primitive the value evaluator already has for this pattern (see #400/#494's `Partial`/`Control` handling elsewhere in this file).
- Add regression coverage in both `src/jq/eval.rs`'s unit tests and `tests/jq_cli_tests.rs` mirroring the repros above, diffed against real jq 1.7.1.

Found during review of PR #841 (#832).

newhoggy:
## Summary

`recurse(f)`/`recurse(f; cond)` dropped `f`'s own partial fan-out output
when a *later* output of that same `f` call errored (or broke/halted), in
both the value-position evaluator (`builtin_recurse_f`/`builtin_recurse_cond`)
and the path-position evaluator (`resolve_recurse`), `src/jq/eval.rs`.

Confirmed against real jq 1.7.1: `f`'s fan-out is a generator like any other
comma expression, so real jq streams every output `f` already produced at a
node — and that value's own recursive subtree — before the error that stops
the whole traversal propagates. succinctly instead discarded that node's own
partial `f` output entirely.

```
$ echo '{"a":1,"b":2}' | jq -c 'recurse(if . == {"a":1,"b":2} then (.a, .b[0]) else empty end)'
{"a":1,"b":2}
1
$ echo '{"a":1,"b":2}' | succinctly jq -c 'recurse(if . == {"a":1,"b":2} then (.a, .b[0]) else empty end)'
{"a":1,"b":2}     # (before this fix — dropped the `1`)
```

Both repros in the issue (value position and `path(recurse(...))`) now
match real jq exactly.

### Root cause and fix

`resolve_recurse`/`builtin_recurse_f`/`builtin_recurse_cond` treated `f`'s
own multi-output evaluation as all-or-nothing via `eval_owned_multi`'s
`Partial(_, Control::Error(e)) => Err(e.into())` arm, which is the *correct*
contract for every other caller of that shared helper (~15 call sites) but
wrong for `recurse`'s own children-evaluation step.

Rather than changing `eval_owned_multi`'s shared contract (used by ~15 other
call sites — this repo has been burned before by a shared-function change
with unintended blast radius, issue #682), this adds
`eval_owned_multi_keep_partial`: a new, recurse-scoped sibling that preserves
the prefix instead of discarding it. `eval_owned_multi` itself is untouched.

Empirically, real jq's behavior is more subtle than "just fold the partial
prefix into outputs": a successfully-produced child from `f`'s own fan-out
gets its *entire* recursive subtree visited (not just the bare value) before
the deferred error propagates. Confirmed live:

```
$ echo '{"child":{"child":"leaf","leafflag":true},"bad":3}' | jq -c \
    'recurse(if (.|type)=="object" then (.child, .bad[0]) else empty end)'
{"child":{"child":"leaf","leafflag":true},"bad":3}
{"child":"leaf","leafflag":true}
"leaf"
null
```

Both evaluators now use a `pending_error` that, once set, replaces the outer
stack (nothing queued before that point may run once an escape is pending —
jq's error/break/halt aborts the *entire* remaining traversal, not just the
current node) and keeps draining until the stack empties, deferring the
raise until then. The same treatment applies uniformly to `error`, `break`,
and `halt` (verified against jq 1.7.1 for all three).

### Scope

- `cond`'s own error-handling arm (in both evaluators) is left untouched —
  it's a structurally similar but separate, unreported class of bug (an
  eager per-node batch over `cond`-checked children that can itself drop an
  already-approved prefix on a *later* `cond` error), and #842 is
  specifically about `f`'s own fan-out. Noted inline at both sites for a
  future issue if it's worth filing separately.
- `eval_owned_multi`/`resolve_against_cow`'s existing contracts for their
  other ~15 call sites are unchanged.

## Test plan

- [x] Diffed output against pinned real `jq` 1.7.1 directly for: the issue's
      two repros, a deeper nested-subtree case, a `break` case, and a `halt`
      case — all match
- [x] New unit tests in `src/jq/eval.rs` (`*_842`): value position, path
      position, `cond` present, nested subtree, `break`
- [x] New CLI/golden tests in `tests/jq_cli_tests.rs` (`*_842`) mirroring the
      issue's repros end-to-end through the CLI binary

Fixes #842

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -10416,9 +10416,19 @@ fn resolve_recurse<'a, S: EvalSemantics>(
         queue_recurse_children(&mut stack, next, deferred_error, &mut pending_error);
     }
 
-    match pending_error {
-        Some(e) => Err((outputs, e)),
-        None => Ok(outputs),
+    // A `pending_error` only surfaces once `stack` drains naturally — if
+    // `MAX_ITEMS` cut the loop short instead (`stack` still non-empty), this
+    // matches the pre-existing MAX_ITEMS convention everywhere else in this
+    // file (silent truncation, no error) rather than surfacing an error
+    // whose presence would otherwise depend on where the arbitrary cap
+    // happened to land relative to it (#842 review).
+    if stack.is_empty() {
+        match pending_error {
+            Some(e) => Err((outputs, e)),
+            None => Ok(outputs),
+        }
+    } else {
+        Ok(outputs)
     }
 }
 
@@ -12762,8 +12772,13 @@ fn builtin_recurse_f<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         queue_recurse_children(&mut stack, children, deferred_error, &mut pending_error);
     }
 
-    if let Some(e) = pending_error {
-        return partial(outputs, e.into());
+    // See `resolve_recurse`'s matching check for the full rationale (#842
+    // review): a `pending_error` only surfaces once `stack` drains
+    // naturally, not when `MAX_ITEMS` cuts the loop short instead.
+    if stack.is_empty() {
+        if let Some(e) = pending_error {
+            return partial(outputs, e.into());
+        }
     }
 
     if outputs.is_empty() {
@@ -12859,8 +12874,13 @@ fn builtin_recurse_cond<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         queue_recurse_children(&mut stack, next, deferred_error, &mut pending_error);
     }
 
-    if let Some(e) = pending_error {
-        return partial(outputs, e.into());
+    // See `resolve_recurse`'s matching check for the full rationale (#842
+    // review): a `pending_error` only surfaces once `stack` drains
+    // naturally, not when `MAX_ITEMS` cuts the loop short instead.
+    if stack.is_empty() {
+        if let Some(e) = pending_error {
+            return partial(outputs, e.into());
+        }
     }
 
     if outputs.is_empty() {
@@ -34264,6 +34284,35 @@ mod tests {
                 r#"{"child":"leaf","leafflag":true}"#,
                 r#""leaf""#,
             ]
+        );
+    }
+
+    #[test]
+    fn test_recurse_f_max_items_truncation_suppresses_a_pending_error_842() {
+        // Review of #842's fix: a `pending_error` deferred past `MAX_ITEMS`
+        // items must not surface just because of where the (internal,
+        // non-jq-semantic) 10000-item safety cap happened to land relative
+        // to it — that would make two structurally identical large fan-outs
+        // (this one, vs. the same shape with no trailing error at all)
+        // diverge on whether an error appears, purely as an accident of the
+        // cap. `builtin_recurse_f` already silently truncates at
+        // `MAX_ITEMS` with no error for a non-erroring fan-out (unrelated
+        // to #842, pre-existing convention); this pins that a *deferred*
+        // error is suppressed the same way once the cap truncates the
+        // drain, rather than resurfacing as `QueryResult::Partial`.
+        let large_doc = format!(
+            r#"{{"items":[{}],"bad":5}}"#,
+            (0..15000)
+                .map(|i| i.to_string())
+                .collect::<Vec<_>>()
+                .join(",")
+        );
+        query!(
+            large_doc.as_bytes(),
+            r#"recurse(if type=="object" then (.items[], .bad[0]) else empty end)"#,
+            QueryResult::ManyOwned(vs) => {
+                assert_eq!(vs.len(), 10000);
+            }
         );
     }
 
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -9412,6 +9412,27 @@ fn push_path_components(out: &mut Vec<Expr>, expr: &Expr) {
 /// first (#694): an error/break mid-stream aborts the whole expression here,
 /// it does not silently degrade into "however many outputs came before it."
 ///
+/// A `break $label` aborts this expression exactly like `Error` — whatever a
+/// resolve-node caller was accumulating before this sub-expression is a
+/// separate concern it already handles via its own `PathResolveResult`
+/// prefix, not something this helper tries to preserve. Propagated as
+/// [`EvalEscape::Break`] rather than collapsed into a synthetic "break
+/// $label not in label" `Error`, so a `label` sitting outside the caller
+/// (e.g. across a `path(...)` call boundary) still catches it instead of
+/// seeing a bogus error (#824). A halt must abort this expression exactly
+/// like `Error`/`Break` rather than falling into `collect_owned()`'s "no
+/// outputs" treatment — that would make `del(.[(halt_error(3))])` or a
+/// computed key `.[(halt)]` a silent no-op instead of halting (#791). It
+/// travels as [`EvalEscape::Halt`], its own variant, so no caller can catch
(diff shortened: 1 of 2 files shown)


newhoggy:
Review surfaced two related, pre-existing (not introduced by this PR) bugs in the same "drop the escape after keeping partial output" bug class, filed as separate follow-ups per this repo's convention (same pattern as #842 itself being split out of #832):

- #854 — `recurse(f; cond)`'s `cond`-evaluation loop drops already-`cond`-approved siblings on a later `cond` error (distinct from this PR's fix, which is scoped to `f`'s own fan-out, not `cond`'s). Confirmed pre-existing via A/B against the pre-#853 base commit.
- #855 — `walk(f)`/`repeat(f)` drop a trailing error/break after a multi-output `f`'s partial values, via `eval_owned_expr_ctrl`'s `Partial(vs, _control)` arm. Distinct from #495 (closed — that was a total swallow with zero partial output, confirmed still fixed).

One claim from an earlier review pass that I independently re-verified and found to be **incorrect**, so not filed: `repeat(exp)` was reported as not threading iteration state forward (`repeat(.+1)` allegedly giving `1,2,3` in real jq vs `1,1,1` in succinctly). Direct testing against real jq 1.7.1 shows real jq's actual `repeat` builtin does *not* thread state either — `echo 5 | jq -c '[limit(5; repeat(.+1))]'` gives `[6,6,6,6,6]` in both real jq and succinctly, a well-known jq quirk (`repeat`'s definition re-evaluates `exp` against the original input each time, not the previous output). succinctly already matches this correctly.

Also noting an accepted tradeoff from review: this fix changes the worst-case work on `recurse`'s error path from ~immediate to up to `MAX_ITEMS` (10000) — draining the already-produced/approved subtree before raising is required for correctness (matching real jq's actual output), so an adversarially-crafted fast-erroring `f` can now do bounded (10000-unit) extra work before the error surfaces, where before it aborted almost immediately. Bounded performance-shape change, not a correctness concern; not worth engineering around for this fix.

newhoggy:
Further review found a real correctness inconsistency in the deferred-error mechanism itself, now fixed (pushed as c51ed8dc): whether a `pending_error` past a large fan-out actually surfaced depended on where the pre-existing `MAX_ITEMS=10000` safety cap happened to land relative to it — a structurally identical fan-out with no error silently truncates with no error (the existing convention everywhere else this cap is hit), so the erroring case needed the same treatment for consistency. All three functions now only honor `pending_error` once `stack` drains to empty naturally; hitting `MAX_ITEMS` first suppresses it, matching the pre-existing silent-truncation convention. Added a regression test (`test_recurse_f_max_items_truncation_suppresses_a_pending_error_842`) confirming a 15000-item fan-out with a trailing error truncates to 10000 with no error, same as the identical shape with no error at all.

One more pre-existing (not introduced here), path-position-only bug surfaced during this same review pass, filed as #856: `resolve_recurse`'s null-child guard discards a null child's own path entirely (not just recursion into it), unlike the value-position evaluator which correctly still emits it as a leaf. No error involved — confirmed independent of this PR's fix.

Issue #843, opened 2026-08-13
jq: path()'s catch handler treats the caught error/break value as path-trackable when real jq never does


newhoggy:
**Severity: Low**

## Summary

Inside `path(try ... catch HANDLER)`, real jq never treats the value bound to `.` inside `HANDLER` (the caught error's payload, or `null` for a caught `break`/an equivalent for `halt`) as path-trackable — *any* attempt to navigate into it (even a single `.field`) immediately raises `#530`'s "Invalid path expression", regardless of which trigger (`error(...)`, `break`, or presumably a would-be-caught `halt`) reached the handler. succinctly's `resolve_against_cow`/`resolve_catch` (`src/jq/eval.rs`) instead treats the caught payload as an ordinary, freshly path-trackable value, and lets navigation into it succeed.

## Repro

```
$ echo '{"a":10}' | jq -c 'path(try (.a, error({b:1,c:2})) catch (.b, "z"))'
["a"]
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "b" of {"b":1,"c":2}
$ echo '{"a":10}' | succinctly jq -c 'path(try (.a, error({b:1,c:2})) catch (.b, "z"))'
["a"]
["b"]
jq: error (at <stdin>:1): Invalid path expression with result "z"
```

Real jq stops at the very first navigation attempt into the caught payload (`.b`) and never reaches `"z"`; succinctly happily resolves `.b` as path `["b"]`, prints it, and only fails on the second, unrelated branch (`"z"`, a literal — correctly caught by the pre-existing #530 check, just for the wrong reason).

The same divergence reaches path-context through a caught `break` too (not just a caught `error`), since both share `resolve_catch`:

```
$ echo '{"a":1}' | jq -c 'label $out | path(try (.a, break $out) catch (.b, "z"))'
["a"]
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "b" of null
$ echo '{"a":1}' | succinctly jq -c 'label $out | path(try (.a, break $out) catch (.b, "z"))'
["a"]
jq: error (at <stdin>:1): Invalid path expression with result "z"
```

(succinctly's `null`-for-break stand-in, documented in `resolve_catch`'s doc comment, means `.b` on `null` resolves as an ordinary null-index instead of erroring — that's the same underlying gap, not a second one; no need to special-case a `break`-marker value separately from the general fix below.)

## Root cause

[...]

PR #857, opened 2026-08-14
fix(jq): path()'s // arm no longer requires a falsy left operand to be path-shaped

### LINKED ISSUE #845: jq: path()'s Alternative (//) arm requires the left operand to be path-shaped even when it's simply falsy and filtered out
**Severity: Low**

## Summary

`resolve_node`'s `Expr::Alternative(left, right)` arm (`src/jq/eval.rs`) resolves `left` as a full path expression *before* checking whether any of its outputs are truthy. Real jq only requires `//`'s left side to be path-shaped for the outputs that actually survive the truthy filter — a left side whose only output is filtered out (falsy) never needs to have been path-trackable in the first place, because it's never emitted or used as part of the result.

## Repro

```
$ echo '{"a":10}' | jq -c 'path(false // .b)'
["b"]
$ echo '{"a":10}' | succinctly jq -c 'path(false // .b)'
jq: error (at <stdin>:1): Invalid path expression with result false
```

Real jq falls through to `.b` because `false` (the left side's only output) is falsy and gets filtered by `//`; succinctly instead raises `#530`'s "Invalid path expression with result false" immediately, because `resolve_node::<S>(left, value, trackable)?` propagates the error via `?` before the `.filter(|(_, v)| v.is_truthy())` step ever runs.

This reproduces with no `catch`/`try` involved at all — it's unrelated to #843's catch-handler trackability fix (which is where this was found, while writing coverage for the `Expr::Alternative` arm's `trackable`-threading; `catch (false // .b)` surfaces the same underlying bug, not a second one).

## Root cause

`src/jq/eval.rs`, `resolve_node`'s `Expr::Alternative` arm:

```rust
Expr::Alternative(left, right) => {
    let branches: Vec<PathBranch<'a>> = resolve_node::<S>(left, value, trackable)?
        .into_iter()
        .filter(|(_, v)| v.is_truthy())
        .collect();
    if branches.is_empty() {
        resolve_node::<S>(right, value, trackable)
    } else {
        Ok(branches)
    }
}
```

The `?` after `resolve_node::<S>(left, ...)` means *any* error resolving `left` — including `#530`'s "not a path expression" complaint for a literal/computed left side — aborts the whole `Alternative` immediately, never reaching the truthy filter or the right-side fallback. The existing doc comment on this arm ("An error while resolving `a` propagates rather than falling back... the `?` here already gives us that for free") treats this as intentional, matching `eval_alternative`'s value-position semantics — but live testing against jq 1.7.1 shows `path()`'s own `//` is more permissive than that comment assumes: a left side that isn't path-shaped is fine as long as none of its outputs are truthy.

## Why this is separate from #843

#843 is about the value bound to `.` inside a `catch` handler being wrongly treated as path-trackable. This issue is about `//`'s left operand being required to be path-shaped even when the resulting value is filtered out by falsiness — a completely different code path (`Expr::Alternative`, not `resolve_catch`) that misbehaves the same way whether or not any `catch` is involved.

## Suggested fix direction

Only require path-shape for `left`'s outputs that actually survive the truthy filter — e.g. resolve `left` in a mode that treats "not path-shaped" as equivalent to "falsy" for values that are simply being tested for truthiness, or restructure the arm so `#530`'s check happens after filtering rather than during the initial resolution's `?`. Worth re-checking how `#530`'s check composes with jq's own `//` bytecode (FORK on error too, not just on falsy?) before picking an approach.

newhoggy:
## Summary

`resolve_node`'s `Expr::Alternative` arm (`src/jq/eval.rs`) resolved `//`'s
left operand as a full path expression *before* checking whether any of its
outputs were truthy. Real jq only requires `//`'s left side to be
path-shaped for the outputs that actually survive the truthy filter — a
falsy output that gets filtered out never needs to have been path-trackable
in the first place.

```
$ echo '{"a":10}' | jq -c 'path(false // .b)'
["b"]
$ echo '{"a":10}' | succinctly jq -c 'path(false // .b)'   # before this fix
jq: error (at <stdin>:1): Invalid path expression with result false
```

The fix adds `resolve_alternative_left`, a dedicated resolver for `//`'s
left operand. It behaves exactly like `resolve_node` except on `#530`'s
"Invalid path expression" error specifically: on that error it re-checks
whether the offending piece's own output was truthy, falling through
("contributes nothing") when it wasn't. Critically, it has its own `Comma`
element loop rather than delegating to the shared one, so each element's
recheck is scoped to *that element alone* — this is what keeps jq's
left-to-right, fail-fast-at-the-first-real-problem ordering intact for
multi-output left operands (a falsy, non-path-shaped element never blocks a
later valid one; a *truthy* non-path-shaped element still stops everything
immediately, before any later element — including one that would otherwise
halt or break — is ever reached).

**Revision history on this PR:** the first commit used a simpler
"catch-the-error-then-re-evaluate-the-whole-left-operand" strategy. Code
review (see below) independently reproduced three real bugs in it — a
multi-output left with both a falsy-invalid and a later truthy-valid
element raised instead of resolving the valid one; a `halt`/`break` reached
only after skipping a falsy element was silently swallowed into an ordinary
catchable error, violating `EvalEscape`'s explicit "never discard a
Halt/Break" contract; and a genuine later error was masked by an earlier,
stale one. The second commit replaces that strategy with the per-element
`resolve_alternative_left` described above, which fixes all three. See that
commit's message for details and the exact repros.

Kept intentionally scoped to `Alternative`'s own `left` operand rather than
threading a new "trackable"-style parameter through `resolve_node`'s other
combinator arms (the pattern `#843` uses for a related-but-separate
problem) — see `resolve_alternative_left`'s doc comment for why a left
operand built from `If`/`Try`/`As`/... instead of `Comma` still gets only
the coarser, whole-subexpression recheck (a narrower residual gap than the
`Comma` case this PR fixes). A left operand with genuine side effects that
also happens not to be path-shaped (e.g. `(stderr | false) // .b`) still
performs them twice on that one non-path-shaped piece — a narrow, documented
trade-off of the re-check strategy itself, not something either commit
eliminates.

Fixes #845

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full workspace suite, 2373+ tests, 0 failed)
- [x] `cargo test --features cli,simd,regex,serde --lib jq::eval::` (targeted)
- [x] `cargo test --features cli,simd,regex,serde --test jq_golden_tests --test jq_evaluator_parity_tests`
- [x] New unit tests: falsy-fallthrough, truthy-still-raises, genuine-error-propagates, mixed truthy/falsy `Comma` element (both orderings), later-genuine-error-wins, fail-fast-before-a-later-escape, halt/break-survives-a-preceding-falsy-element
- [x] New jq golden fixtures (`path_alternative_falsy_left_not_path_shaped`, `path_alternative_truthy_left_not_path_shaped`) captured from the pinned real jq 1.7.1 oracle via `scripts/sync-jq-golden.sh`
- [x] Manually diffed output (stdout + exit code) against the pinned real `jq` for 13+ cases including the reported repro, multi-output `Comma` orderings, and `halt_error`/`break` interactions — all match byte-for-byte


newhoggy:
Automated code review (via the `code-review` skill, run incidentally by a `grinding-backlog` session working on an unrelated issue — surfacing this here since the findings are real and this PR is open). Built the branch (`9396be20`) and empirically confirmed all three against real `jq` 1.7.1 before reporting.

## Findings

**1. `resolve_alternative_left`'s fallback for non-`Comma` left operands re-checks truthiness against the whole re-evaluated subexpression instead of the already-resolved prefix** (`src/jq/eval.rs:10083`)

A legitimate truthy, path-shaped branch that appears *before* a later non-path-shaped element in the same subexpression gets thrown away, and the stale #530 error is raised instead of the correct result:

```
$ echo '{"a":10}' | succinctly jq -c '((if true then (.a, false) else null end) // .b) = 99'
jq: error: Invalid path expression with result false
```

Real jq 1.7.1 outputs `{"a":99}` — `.a` (truthy, path-shaped) is the kept branch, `false` is simply filtered by `//`. This isn't `path()`-only: it breaks ordinary `=`/`|=`/`del()` writes using the `(a // b) = value` idiom whenever `a` is a non-`Comma` combinator (`if`/`try`/`as`/`label`/pipe) wrapping a `Comma`.

**2. The same re-check can turn a catchable #530 error into an uncatchable process halt** (`src/jq/eval.rs:10084`)

```
$ echo '{"a":10}' | succinctly jq -c 'path((if true then (1, halt_error(3)) else null end) // .b)'
```
halts the process with exit code 3 (printing the input value first). Real jq 1.7.1 raises `Invalid path expression with result 1` and never reaches `halt_error` — jq's fail-fast generator semantics stop at the first bad output, but `resolve_alternative_left`'s re-run via `eval_owned_multi` re-evaluates the whole expression and lets the later `halt` win.

**3. The fallback double-evaluates the left operand when the first pass raises #530** (`src/jq/eval.rs:10084`)

Unlike `eval_alternative` (the value-position `//`, line 2242), which evaluates its left operand exactly once, `resolve_alternative_left` re-runs it a second time via `eval_owned_multi` on error:

```
$ echo '5' | succinctly jq -c 'path(stderr // .b)'
```
writes `55` to stderr (the value `5` printed twice) before raising the error, instead of a single `5`. Any impure/expensive left operand (e.g. involving `input`/`stderr`) runs twice for one logical evaluation — with `input`, this would desync a multi-document program.

**Root cause (all three):** the fallback path for any non-`Comma` left operand re-evaluates the *entire* subexpression from scratch via `eval_owned_multi` rather than reusing/refining the partial result `resolve_node`'s first pass already computed. This is unsound whenever the subexpression can produce a legitimate prefix output before a later bad one (nested `Comma` inside `If`/`Try`/etc.), or has an observable side effect.

The doc comment above the fallback (~`eval.rs:10054`) frames this gap as merely "coarser"/"narrower" than the mode-flag alternative that was rejected — worth tightening, since the actual cost is outright wrong results and wrong control flow, not just reduced precision on a narrow input class.

Issue #856, opened 2026-08-14
yq/jq: resolve_recurse's null-child guard discards the null child's own path entirely


newhoggy:
**Severity: Low**

## Summary

`resolve_recurse`'s null-child guard (`src/jq/eval.rs`, the
`if matches!(child_value.as_ref(), OwnedValue::Null) { continue; }` line in
its main loop) is documented as stopping recursion *into* a null child (to
bound path-prefix growth against a non-terminating `f`, e.g.
`recurse(.a?)` over `{"a":null}` — see the function's own doc comment). But
it does more than that: it discards the null child's own path output
entirely, rather than still emitting it as a leaf the way
`builtin_recurse_f`/`builtin_recurse_cond` (value position) correctly do.

Confirmed independent of #842/#853/#854: no error involved, path-position
only, predates all three.

## Repro

```
$ echo '{"a":null}' | jq -c 'path(recurse(if . == {"a":null} then .a else empty end))'
[]
["a"]
$ echo '{"a":null}' | succinctly jq -c 'path(recurse(if . == {"a":null} then .a else empty end))'
[]
```

Real jq prints `[]` (root) *and* `["a"]` (the null child, emitted as a leaf
— its own recursion into `f` on `null` correctly produces nothing further,
since `null != {"a":null}`); succinctly prints only `[]`, dropping `["a"]`
entirely.

For comparison, the value-position evaluator gets this right already:

```
$ echo '{"a":null}' | succinctly jq -c 'recurse(if . == {"a":null} then .a else empty end)'
{"a":null}
null
```

## Root cause

`resolve_recurse`'s guard is a bare `continue` in the `for (child_components,
child_value) in children` loop, which skips the child entirely — not just
recursion into it:

```rust
if matches!(child_value.as_ref(), OwnedValue::Null) {
    continue;
}
```

The intent (per the function's own doc comment) is only to stop the *next*
level of recursion for a null child, not to suppress the child's own
emission — `builtin_recurse_f`/`builtin_recurse_cond` push a null child
onto their stack like any other value (it gets emitted when popped, and
its own `f` re-evaluation is what naturally terminates if `f` doesn't
navigate through null further); `resolve_recurse` needs the analogous
"emit but don't further recurse" treatment instead of "skip entirely."

## Why this is separate from #842/#853/#854

- #842/#853 (merged) fixed `f`'s own partial fan-out being dropped on a
  *later* error — an error-path bug. This is a *no-error* case; the
[...]

PR #859, opened 2026-08-14
fix(jq): keep recurse(f; cond)'s already-approved siblings before a later cond error

### LINKED ISSUE #854: jq: recurse(f; cond) drops cond's already-approved siblings on a later cond error
**Severity: Low**

## Summary

`recurse(f; cond)`'s `cond`-evaluation loop (both the value-position
`builtin_recurse_cond` and the path-position `resolve_recurse`, in
`src/jq/eval.rs`) drops the *already-`cond`-approved* siblings of a node's
children when `cond` itself errors on a *later* child in the same batch —
distinct from #842 (which was about `f`'s own fan-out, not `cond`'s).

Real jq's generator semantics mean each already-approved child gets its own
full recursive descent *before* `cond` is ever asked about the next
candidate child — same "never un-emit an already-produced output" principle
as #530/#636/#694/#824/#842, just not yet applied to this specific loop.

## Repro

Value position:

```
$ echo '[1,2,3]' | jq -c 'recurse(if type=="array" then .[] else empty end; if . == 2 then error("boom") else true end)'
[1,2,3]
1
$ echo '[1,2,3]' | succinctly jq -c 'recurse(if type=="array" then .[] else empty end; if . == 2 then error("boom") else true end)'
[1,2,3]
```

Real jq prints the root (`[1,2,3]`) *and* `1` (the first child, already
`cond`-approved and fully recursed before `cond` errors on the second
child, `2`) before erroring; succinctly prints only the root, dropping `1`.

Path position (`resolve_recurse`), same shape:

```
$ echo '[1,2,3]' | jq -c 'path(recurse(if type=="array" then .[] else empty end; if . == 2 then error("boom") else true end))'
[]
[0]
$ echo '[1,2,3]' | succinctly jq -c 'path(recurse(if type=="array" then .[] else empty end; if . == 2 then error("boom") else true end))'
[]
```

## Root cause

`builtin_recurse_cond`'s and `resolve_recurse`'s `cond`-evaluation loop
(`for child in children { ... match eval_owned_multi::<S>(cond, &child) {
Ok(...) => {...} Err(e) => return partial(outputs, e.into()) } }`) returns
immediately on a `cond` error, discarding `next` — the children that
already passed `cond` earlier in the same iteration of the loop but haven't
had their own recursive descent yet (that only happens on a later pass
through the outer `while` loop, once `next` is pushed onto `stack`).

## Why this is separate from #842

#842 (fixed in PR #853) was specifically about `f`'s *own* multi-output
fan-out erroring partway through (e.g. `f = (.a, .b[0])` where `.b[0]`
errors after `.a` already produced a value) — an escape from evaluating `f`
itself. This issue is about `cond`'s error, evaluated in a completely
separate loop over `f`'s *already-successfully-produced* children — a
different code path in both functions. #853's own review confirmed this is
pre-existing (present on the pre-#853 base commit too, unrelated to that
PR's change) and deliberately left untouched there to keep that fix
narrowly scoped; both call sites carry an inline comment noting the gap and
pointing here.

## Suggested fix direction

Likely wants the same `pending_error`/`queue_recurse_children` deferred-error
mechanism #853 added for `f`'s fan-out, applied to the `cond`-evaluation loop
too: on a `cond` error, keep whatever children already passed `cond` in this
batch, give them their own full recursive descent (isolated from anything
else queued), and only then raise the error — instead of returning
immediately. Add regression coverage in `src/jq/eval.rs`'s unit tests and
`tests/jq_cli_tests.rs`, diffed against real jq 1.7.1, mirroring the repros
above for both evaluators.

newhoggy:
## Summary

`recurse(f; cond)`'s `cond`-evaluation loop (`resolve_recurse` path
position, `builtin_recurse_cond` value position, both in `src/jq/eval.rs`)
returned immediately on a `cond` error, discarding whatever siblings of a
node's children had *already passed* `cond` earlier in the same loop —
this is the sibling of #842 (fixed in #853), which was about `f`'s own
fan-out erroring, not `cond`'s.

Extends #842's `queue_recurse_children`/`pending_error` deferred-error
mechanism to `cond`'s loop: on a `cond` error, the loop now stops gating
further children (matching real jq, which never asks `cond` about them
either) and defers the error via the same helper, giving the
already-`cond`-approved siblings their own full recursive descent before
the error finally surfaces. If `f` itself had also deferred an error at
the same node, `cond`'s error correctly supersedes it — in jq's real
lazy-generator interleaving, `f` is never even asked for its next output
until the current candidate has fully cleared `select(cond) | r`, so a
`cond` failure on an already-produced child is always chronologically
earlier.

A second, deeper layer surfaced during verification: `cond` itself can be
a multi-output generator (e.g. `(true, error("x"))`, independent of the
child), so the per-child call to `cond` was switched from
`eval_owned_multi` to `eval_owned_multi_keep_partial` too — otherwise an
already-truthy output within *that same call* was discarded along with
its own later error, before my sibling-level fix even got a chance to see
it. Confirmed this reproduces jq's real behavior byte-for-byte, including
the fact that jq's *reported error message differs* in this case (it
recurses fully into the approved child before ever asking `cond` for its
second output, so the error that actually surfaces there — e.g. "Cannot
iterate over number" — supersedes `cond`'s own deferred one via the
existing "later escape supersedes" rule).

Stayed within `resolve_recurse`'s and `builtin_recurse_cond`'s
`cond`-evaluation loops specifically, per #842's precedent — no change to
`eval_owned_multi`/`resolve_against_cow`'s shared contracts used by other
callers.

## Repro (verified against real jq 1.7.1)

```
$ echo '[1,2,3]' | jq -c 'recurse(if type=="array" then .[] else empty end; if . == 2 then error("boom") else true end)'
[1,2,3]
1
$ succinctly jq -c 'recurse(if type=="array" then .[] else empty end; if . == 2 then error("boom") else true end)' <<< '[1,2,3]'
[1,2,3]
1
```

(Before this fix, succinctly printed only `[1,2,3]`, dropping `1`.)

Fixes #854

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 2406 lib tests + all integration suites green; one CLI test flaked under local worktree contention, reproduced green in isolation and in a full solo `--test jq_cli_tests` run)
- [x] `cargo fmt --check`
- [x] Diffed output against real jq 1.7.1 directly for both repros in the issue, plus a deeper `cond`-own-fanout case and a subtree case, at both value and path position
- [x] Added 9 unit tests in `src/jq/eval.rs` and 3 CLI tests in `tests/jq_cli_tests.rs`; confirmed 100% line coverage on every new/changed branch via a scoped local `cargo llvm-cov --lib -- _854` run

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -9442,10 +9442,7 @@ fn eval_owned_multi<S: EvalSemantics>(
     input: &OwnedValue,
 ) -> Result<Vec<OwnedValue>, EvalEscape> {
     let (values, escape) = eval_owned_multi_keep_partial::<S>(expr, input);
-    match escape {
-        Some(e) => Err(e),
-        None => Ok(values),
-    }
+    escape.map_or(Ok(values), Err)
 }
 
 /// Like [`eval_owned_multi`], but keeps `collect_owned`'s old "just give me
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -36264,9 +36264,10 @@ mod tests {
         // comment above, "One divergence remains, deliberately") applying
         // identically whether or not `cond` is present: `is_null_current`
         // must terminate this in exactly the same 2 outputs the no-`cond`
-        // case (`path(recurse(.a?))`, already covered by other tests)
-        // produces, not silently regress to unbounded (or `MAX_ITEMS`-only)
-        // growth just because a `cond` was added.
+        // case (`path(recurse(.a?))`, asserted directly below as this
+        // test's own baseline -- no other test happens to pin this exact
+        // shape) produces, not silently regress to unbounded (or
+        // `MAX_ITEMS`-only) growth just because a `cond` was added.
         assert_eq!(
             outputs(br#"{"a":null}"#, r"path(recurse(.a?; true))"),
             vec!["[]", r#"["a"]"#]
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -10555,20 +10555,30 @@ fn resolve_catch<'a, S: EvalSemantics>(
 
 /// Shared "defer the escape, queue what's left" step for `recurse`'s three
 /// stack-driven implementations (`resolve_recurse`,
-/// `builtin_recurse_f`/`builtin_recurse_cond`) (#842).
+/// `builtin_recurse_f`/`builtin_recurse_cond`) (#842, extended to `cond`'s
+/// own fan-out by #854).
 ///
-/// `next` is this node's already-`cond`-gated (where applicable) children —
-/// from a fully-successful `f` call, or from whatever `f` itself produced
-/// before ending in an error/break/halt. Called every iteration regardless
-/// of which case it was: on success `deferred_error` is `None` and this is
-/// just the ordinary "push `next` reversed" step; on an escape, `stack` is
-/// cleared first (nothing already queued may run once an escape is pending
-/// — jq's error/break/halt aborts the *entire* remaining traversal, not
-/// just the current node) before `next` replaces it, and `pending_error` is
-/// set so the caller raises it once `next`'s own subtree(s) have fully
-/// drained — which may itself overwrite `pending_error` with a *later*
-/// call's own escape, correctly superseding this one (it happens first,
-/// chronologically, in jq's real generator order).
+/// `next` is this node's already-`cond`-approved children — from a
+/// fully-successful `f` call with every candidate child either passing or
+/// failing `cond` normally, or truncated partway through by an escape from
+/// either `f` itself or `cond` (whichever ends this node's fan-out first).
+/// Called every iteration regardless of which case it was: on success
+/// `deferred_error` is `None` and this is just the ordinary "push `next`
+/// reversed" step; on an escape, `stack` is cleared first (nothing already
+/// queued may run once an escape is pending — jq's error/break/halt aborts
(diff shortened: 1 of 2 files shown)


newhoggy:
Follow-up #896 filed for the 4-site `resolve_node` prefix-loss bug class noted in #854's status comment.

Also found and fixed a real regression in this PR's own `d723242f` commit during rebase: the `Some(cond)` arm's switch to `eval_owned_multi_keep_partial` had dropped the pre-existing `is_null_current` gate (#856) that bounds recursion past a null node — without it, a truthy `cond` on a null current's candidate children would reopen the quadratic-blowup risk that gate exists to prevent. Restored during conflict resolution, pinned with new tests.

Applied the rest of the review punch list: `eval_owned_multi_keep_partial` now returns `(values, escape)` directly instead of `Result<Vec<_>, (Vec<_>, _)>` (removes the duplicated Ok/Err unwrap at all 5 call sites, and the shadowing concern along with it since the match itself is gone), fixed its stale doc comment, added the missing path-position `cond`+`break` test. Rebased onto current `main`, full suite + both clippy invocations + fmt all clean.

newhoggy:
Filed follow-up #897 for the `cond`-gate-and-defer duplication between `resolve_recurse` and `builtin_recurse_cond` that several review passes flagged — worth doing, but scoped out of this PR to avoid further expanding an already-large diff on a file with heavy concurrent traffic right now.

Issue #693, opened 2026-08-09
jq: `optional`/`?` conflates a caught error with genuine `empty` at leaf evaluators (pre-existing, amplified by #534)


newhoggy:
**Severity: High** — silent wrong answers under ordinary `?`/`try` usage, not errors. Pre-existing, predates #534.

## Summary

`optional`/`?` is implemented in `src/jq/eval.rs` by threading an `optional: bool` parameter down through the entire evaluator. At leaf error-raising sites (`eval_error`, and presumably others with the same shape), `optional=true` converts a real error directly into `QueryResult::None` — indistinguishable from a genuinely empty stream. Every combinator that sequences sub-evaluations (`eval_comma`, and by extension `reduce`/`foreach`/`while`/`until`'s per-step evaluation via `eval_owned_expr_fork`) receives that same ambient `optional` flag and passes it straight through to its children, so a masked error looks exactly like ordinary `empty` to the combinator — which then wrongly *continues* instead of stopping the rest of the `?`/`try`-wrapped expression's evaluation, which is what real jq does.

## Repro (verified against pinned jq 1.7.1, and against `main` at `e5ccc83c`, i.e. before #534 merged)

```
$ jq -n '(1, error("x"), 2)?'
1
$ succinctly jq -n '(1, error("x"), 2)?'   # wrongly continues past the masked error
1
2
```

```
$ jq -n '(reduce (1) as $x (0; error("boom")))?'
                                              # (no output at all)
$ succinctly jq -n '(reduce (1) as $x (0; error("boom")))?'
null                                         # wrongly treats the masked error as if UPDATE produced empty
```

Both repros reproduce identically on `main` (commit `e5ccc83c`, before #534's parser/evaluator changes), confirming this predates #534 and is not something that PR introduced.

## Why this matters more after #534

#534 taught `reduce`/`foreach`/`while`/`until` to correctly fork over multi-output INIT/UPDATE/COND/EXTRACT. That fork logic amplifies this same pre-existing leaf-level bug into much more visible symptoms:

```
$ jq -n '[(foreach (1,2) as $x ((10,20); if .==10 then error("boom") else .+$x end))?]'
[]
$ succinctly jq -n '[(foreach (1,2) as $x ((10,20); if .==10 then error("boom") else .+$x end))?]'
[2,21,23]        # fork 1's masked error wrongly "recovers" to null, and fork 2 (which
                  # real jq's ? never attempts once fork 1 aborts) is wrongly tried too
```

[...]

PR #863, opened 2026-08-14
fix(jq): let strftime accept a raw Unix timestamp number

### LINKED ISSUE #759: jq: strftime rejects a raw timestamp number; real jq auto-converts via gmtime
## Summary

Real jq's `strftime(fmt)` accepts either a broken-down-time array (from
`gmtime`/`localtime`) **or** a raw Unix timestamp number, auto-converting the
number via `gmtime` internally. `succinctly jq`'s `builtin_strftime`
(`src/jq/eval.rs`) only accepts an array and raises a type error otherwise.

```
$ echo '1435677542.822351' | jq 'strftime("%A, %B %e, %Y")'
"Tuesday, June 30, 2015"

$ echo '1435677542.822351' | succinctly jq 'strftime("%A, %B %e, %Y")'
jq: error (at <stdin>:1): expected array, got strftime
```

## Impact

This isn't just a narrow gap — it breaks any pipeline that produces a
timestamp arithmetically and formats it directly, e.g.
`mktime + 86400 | strftime(...)`, without an explicit intervening `gmtime`.
That's exactly the shape of jq's own day-of-week/day-of-year round-trip
torture test (`tests/optional.test`), which is now pinned as
`date_roundtrip_leap_year_torture` in `tests/data/jq-golden/cases/` (see #716)
and fails for this reason — it's in
`tests/data/jq-golden-known-failures.txt` pending this fix, alongside
`strftime_raw_number_input`.

## Suggested fix

In `builtin_strftime`, when the input value is a number rather than an array,
convert it via the same logic `builtin_gmtime` uses before formatting,
mirroring how `builtin_todate` already accepts a raw number.

## Discovery context

Found while adding jq-oracle golden coverage for date/time builtins (#716).

newhoggy:
## Summary

- Real jq's `strftime(fmt)` accepts either a broken-down-time array (from `gmtime`/`localtime`) or a raw Unix timestamp number, auto-converting the number the same way `gmtime` does. `succinctly jq`'s `builtin_strftime` only accepted an array and raised a type error otherwise, breaking pipelines like `mktime + 86400 | strftime(...)` that produce a timestamp arithmetically without an explicit intervening `gmtime`.
- `builtin_strftime` now branches on whether the input is a `StandardJson::Number` (converts via the same Howard-Hinnant-algorithm broken-down-time computation `builtin_gmtime` uses) or an array (existing path, unchanged).
- Also fixed the type-error wording for a genuinely invalid input (string/null/etc): it was the generic `"expected array, got strftime"` (literally naming the builtin as the "got" type, a pre-existing bug); it's now jq's own `"strftime/1 requires parsed datetime inputs"`, matching real jq exactly. Fixed inline since it's the same match arm being restructured for this change.
- Dropped `strftime_raw_number_input` and `date_roundtrip_leap_year_torture` from `tests/data/jq-golden-known-failures.txt` — both now pass against the pinned jq oracle.

Fixes #759

## Test plan

- [x] Manually diffed output against the pinned real `jq` oracle for: raw float timestamp, raw integer timestamp, `mktime + 86400 | strftime(...)` roundtrip, existing array input (regression check), invalid-type error wording (string/null), negative (pre-1970) timestamp, and the `?` optional-suppression form — all byte-identical to real jq
- [x] `jq_golden_conformance` test passes with the two known-failure entries removed (437/440 → still passing after removing them from the manifest)

### DIFF
--- a/docs/plan/yq.md
+++ b/docs/plan/yq.md
@@ -333,11 +333,12 @@ now  # → 1705766400.123456 (seconds since Unix epoch)
 .timestamp = now
 ```
 
-**`strftime(fmt)`** - Format broken-down time array as string (jq-compatible) ✅
+**`strftime(fmt)`** - Format broken-down time array, or a raw Unix timestamp number, as string (jq-compatible) ✅
 ```bash
 # Uses POSIX strftime format specifiers (%Y, %m, %d, etc.)
 now | gmtime | strftime("%Y-%m-%d")  # → "2026-01-20"
 now | gmtime | strftime("%Y-%m-%dT%H:%M:%SZ")  # → "2026-01-20T15:04:05Z"
+now | strftime("%Y-%m-%d")  # → "2026-01-20" (raw number, auto-converted like gmtime)
 ```
 
 **`strptime(fmt)`** - Parse string to broken-down time array (jq-compatible) ✅
--- a/docs/reference/jq-language.md
+++ b/docs/reference/jq-language.md
@@ -226,7 +226,7 @@ The implementation covers ~95% of jq functionality and is production-ready.
 - [x] `gmtime` - Convert Unix timestamp to broken-down UTC time
 - [x] `localtime` - Convert Unix timestamp to broken-down local time
 - [x] `mktime` - Convert broken-down time to Unix timestamp
-- [x] `strftime(fmt)` - Format broken-down time as string
+- [x] `strftime(fmt)` - Format broken-down time (or a raw Unix timestamp number) as string
 - [x] `strptime(fmt)` - Parse string to broken-down time
 - [x] `todate` / `todateiso8601` - Convert Unix timestamp to ISO 8601 string
 - [x] `fromdate` / `fromdateiso8601` - Parse ISO 8601 string to Unix timestamp
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -730,6 +730,12 @@ impl EvalError {
         Self::new("mktime requires array inputs")
     }
 
+    /// `strftime/1 requires parsed datetime inputs` — `strftime` on an input
+    /// that's neither a broken-down-time array nor a raw number.
+    pub fn strftime_requires_parsed_datetime_inputs() -> Self {
+        Self::new("strftime/1 requires parsed datetime inputs")
+    }
+
     /// `date "<input>" does not match format "<fmt>"` — any `strptime`
     /// parse failure. jq reports this one wording regardless of which
     /// format specifier the input failed to satisfy.
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -16742,15 +16742,18 @@ fn builtin_strftime<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     // Value should be a broken-down time array, or a raw Unix timestamp
     // number (auto-converted the same way `gmtime` converts one).
     let (year, month, day, hour, minute, second, weekday, yearday) = match &value {
-        StandardJson::Number(_) => {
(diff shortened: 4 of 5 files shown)


newhoggy:
Filed #869 for the pre-existing overflow/trunc-vs-floor/NaN bugs in the epoch↔civil-time math that review surfaced (reproducible via `gmtime` alone on unmodified `main`, predating this PR) — out of scope for #759's narrow fix, tracked separately.

newhoggy:
Ran a multi-angle standard-effort `/code-review` pass on this PR. Findings and how they were handled:
- **Fixed** (new commit): extracted `broken_down_time_from_unix_secs` shared by `gmtime` and `strftime`'s new number branch (was inlining a 4th/5th copy of the epoch-conversion algorithm), added a unit test for the number-input path, fixed two stale doc references, updated the known-failures manifest's stale bug-count narration, and satisfied a `manual_rem_euclid` clippy hit surfaced by the extraction.
- **Filed as follow-up** (#869): three real, reproducible bugs (integer-underflow panic / silent wraparound on extreme timestamps, `.trunc()`-vs-floor mishandling negative sub-second timestamps, NaN silently mapping to epoch) — all pre-existing in `gmtime` before this PR, just newly reachable through `strftime`'s new path. Out of scope for #759's narrow fix.

If a deeper pass is wanted before merge, a human can trigger `/code-review ultra` — not run here since it needs explicit user confirmation and billing.

newhoggy:
## Coverage

Patch coverage: **91.3%** (84/92 new lines). The remaining 8 uncovered lines in `builtin_strftime`'s restructured match are genuinely unreachable via any currently-expressible jq syntax, not gaps I chose to skip:

- **`f64::NAN` / `n.as_f64()`-failure branches** (NaN-sentinel and malformed-number-literal paths): these require a `StandardJson::Number` cursor whose raw bytes are a NaN sentinel or a number token that fails `as_f64()`. No valid JSON literal encodes NaN (confirmed: `echo 'NaN' | succinctly jq '.'` errors as an invalid token, doesn't parse as a number), and the `nan` jq builtin's output never reaches this code path as a `StandardJson::Number` cursor.
- **Every `if optional { return QueryResult::None }` branch inside `builtin_strftime`** (both the number-branch and array-branch variants, plus the outer "neither array nor number" fallback): dead by design since #693 — plain `expr?` (`Expr::Optional`) evaluates its inner expression with the *ambient* `optional` value, not a forced `true`; only `.foo?`/`.[0]?` (`IndexExpr`/`SliceExpr`) force `optional=true` on themselves specifically. `strftime(...)?` is a plain `FuncCall`, so it's evaluated with `optional=false` and its `QueryResult::Error` is instead caught and converted to `None` one level up, in `eval_try` — correct end-to-end behavior (tested), but it never exercises these internal branches. Confirmed by testing `strftime(...)?` on both a too-short array and an invalid-type input — both correctly produce no output, but coverage still shows the internal `if optional` lines as unhit, which is consistent with this mechanism.
- **`get_int`'s bare `Some(OwnedValue::Int(n))`/`Some(OwnedValue::Float(f))` arms**: unreachable because every value that reaches this match — JSON array literals, arithmetic results (`map(.+0)`), and other builtins' array output (`gmtime`) — normalizes to `OwnedValue::NumberLiteral` before arriving here. Verified empirically with two different constructions (piping `gmtime`'s output directly, and forcing arithmetic via `map(.+0)`); neither reached the bare `Int`/`Float` arms.

All of this is pre-existing code shape (the `optional: bool` threading convention and `get_int`'s 4-arm match both predate #759 verbatim) — my diff just relocated it into a new match arm, which is why patch-coverage flags it as "new."

Issue #716, opened 2026-08-09
test(jq): date/time builtins have no golden/CLI oracle coverage against real jq


newhoggy:
## Summary

`strftime`/`strptime`/`mktime`/`gmtime`/`localtime`/`todate`/`fromdate` have
unit tests (`test_gmtime`, `test_mktime`, `test_strftime`, `test_strptime`,
`test_todate`, `test_fromdate`, `test_date_roundtrip` in `src/jq/eval.rs`),
but **zero entries in the 338-case golden corpus and zero entries in
`jq_cli_tests.rs`** reference any date/time filter. Every date/time
assertion in the suite is self-consistency-checked against hand-written
expectations — none of it is cross-checked against real jq-1.7.1's actual
output, even though `tests/data/jq-golden/` already has jq pinned as an
oracle and the CLI tests have `jq` on `PATH` (confirmed at
`jq-1.7.1-apple`, matching `tests/data/jq-golden/JQ_VERSION`).

## Why this matters

Date/time formatting is exactly the kind of area — format-string edge cases
(`%j`/`%U`/`%e`), leap-year handling, weekday calculation, timezone/locale
interaction — where a reimplementation is most likely to diverge from libc's
`strftime`/`strptime` in a subtle way. jq's own test suite treats this as
important enough to include a dedicated leap-year/weekday torture test
(`tests/optional.test:8-13`): round-tripping `strftime`→`strptime` across
every day from a fixed 1970 epoch for 67 years (365×67 days), specifically
to catch date-arithmetic bugs that a handful of hand-picked examples would
miss. succinctly has no equivalent.

## Suggested test cases

- Add golden cases under `tests/data/jq-golden/cases/` for `gmtime`,
  `mktime`, `strftime("%Y-%m-%dT%H:%M:%SZ")`, `strptime(...)`, `todate`,
  `fromdate` on a handful of representative epochs (including a leap-day
  boundary and a DST-transition-adjacent date, if timezone-sensitive paths
  are involved).
- Consider porting (or approximating) jq's day-of-week/day-of-year
  round-trip torture test as an integration test that runs both
  `succinctly jq` and real `jq` over the same date range and diffs the
  output, rather than hand-typing hundreds of expected values.

## Discovery context

Found via a coverage audit cross-referencing succinctly's jq test suite
against jq-1.7.1's own `tests/optional.test`/`tests/man.test`.

Issue #869, opened 2026-08-14
jq: gmtime/localtime/strftime panic or silently corrupt on extreme or sub-second-negative timestamps; epoch<->civil-time math still duplicated 4x


newhoggy:
**Severity: Medium**

## Summary

Multi-angle review of #759 (PR #863, which let `strftime` accept a raw Unix timestamp number) surfaced three correctness bugs in the shared epoch-to-broken-down-time conversion, all pre-existing in `builtin_gmtime` before #863 but newly reachable through a second path once `strftime` started accepting raw numbers too. #863 consolidated `gmtime` and `strftime`'s number path into one shared helper (`broken_down_time_from_unix_secs`, `src/jq/eval.rs:16431`), so these bugs now live in one place instead of two — but the bugs themselves were out of scope for that PR and are unfixed. Three more copies of the same duplicated math (`builtin_localtime`, `builtin_todate`, `format_datetime_with_offset`) still exist independently and are not covered by the consolidation.

## Repro (verified against `succinctly jq`, this repo's own binary — not a jq divergence, real jq is presumably also worth checking but wasn't compared here)

**1. Integer-underflow panic (debug) / silent wraparound (release) on extreme-magnitude timestamps:**
```console
$ echo null | succinctly jq '(-1e300) | strftime("%Y-%m-%d")'    # debug build
thread 'main' panicked at src/jq/eval.rs:16436:9: attempt to subtract with overflow

$ echo null | succinctly jq '(-1e300) | strftime("%Y-%m-%d")'    # release build
"292277026596-12-03"     # silently wrong, no error, exit 0
```
Root cause: `broken_down_time_from_unix_secs` (`src/jq/eval.rs:16431-16481`) does `let secs = timestamp.trunc() as i64;` then, for negative `secs`, `(secs - 86399) / 86400`. A sufficiently negative/positive float saturates the `as i64` cast to `i64::MIN`/`i64::MAX`, and `i64::MIN - 86399` underflows. Reachable via `gmtime(-1e300)` directly, and now also via `strftime(-1e300)` since #863.

**2. `.trunc()` instead of floor mishandles negative sub-second timestamps:**
```console
$ echo null | succinctly jq '(-0.5) | strftime("%Y-%m-%dT%H:%M:%SZ")'
"1970-01-01T00:00:00Z"   # should be "1969-12-31T23:59:59Z" (floor semantics)
```
[...]

PR #872, opened 2026-08-14
feat(jq): implement strftime's %s/%U/%V/%W/%G/%g specifiers

### LINKED ISSUE #760: jq: strftime doesn't implement %s/%U/%V/%W/%G/%g format specifiers
## Summary

`format_strftime` in `src/jq/eval.rs` falls back to echoing an unrecognized
`%<c>` specifier literally (`Some(other) => { result.push('%'); result.push(other); }`).
That's the right behavior for genuinely libc/locale-dependent specifiers
(`%z`, `%Z`, `%P`, `%x`, `%X`, `%c` vary by host TZ/libc even under real jq —
confirmed by direct comparison, which is why they're deliberately *not*
golden-tested), but six specifiers are pure, portable calendar arithmetic
that real jq computes correctly and succinctly does not implement at all:

| Specifier | Meaning                              | jq (epoch 1435677542, UTC) | succinctly |
|-----------|---------------------------------------|------------------------------|------------|
| `%s`      | seconds since epoch                   | `1435641542`                 | `%s`       |
| `%U`      | week number, Sunday-first             | `26`                         | `%U`       |
| `%V`      | ISO 8601 week number                  | `27`                         | `%V`       |
| `%W`      | week number, Monday-first             | `26`                         | `%W`       |
| `%G`      | ISO 8601 week-based year               | `2015`                       | `%G`       |
| `%g`      | ISO 8601 week-based year, 2-digit      | `15`                         | `%g`       |

Reproduced with real jq vs `succinctly jq` on `gmtime | strftime("%s %U %V %W %G %g")`.

## Impact

Pinned as `strftime_unsupported_specifiers` in
`tests/data/jq-golden/cases/` (see #716) and recorded in
`tests/data/jq-golden-known-failures.txt` pending this fix.

## Suggested fix

Add the six specifiers to `format_strftime`'s `match`:
- `%s`: the original Unix timestamp (needs threading through or recomputing
  from the broken-down fields via the same math `mktime` uses).
- `%U`/`%W`: week number from `yearday`/`weekday`, differing only in which
  day starts the week (Sunday vs Monday).
- `%V`/`%G`/`%g`: ISO 8601 week-based year/week, which requires the
  Thursday-of-this-week rule (a date belongs to the ISO year of the Thursday
  in its week) — more involved than `%U`/`%W`.

## Discovery context

Found while adding jq-oracle golden coverage for date/time builtins (#716).

newhoggy:
## Summary

- Real jq's `strftime` supports six portable calendar-arithmetic specifiers that `succinctly` echoed back literally instead of implementing: `%s` (seconds since epoch), `%U`/`%W` (Sunday-/Monday-first week number), and the ISO 8601 trio `%V`/`%G`/`%g` (week number, week-based year, 2-digit week-based year).
- `%s` is recomputed from the broken-down-time fields via the same inverse-Hinnant-algorithm math `mktime` uses — extracted into a shared `unix_secs_from_broken_down_time` helper (now used by both `mktime` and `strftime`) rather than duplicating it a third time in this file. This is the portable choice per the issue's own follow-up correction: real jq's `%s` actually round-trips through libc `mktime()`, which reinterprets the UTC-valued broken-down time as *local* time and shifts by the capturing host's TZ offset — porting that quirk would just reproduce the same host-dependence #716 already excluded `localtime` for, in a different builtin.
- `%U`/`%W` use the standard `(yearday - weekday + 7) / 7` week-number formula, offset for Sunday- vs Monday-first weeks.
- `%V`/`%G`/`%g` implement the ISO 8601 "Thursday rule" (a date belongs to the ISO year of the Thursday in its week) via a new `iso_week_date` helper, including the 52-vs-53-week-year edge case.

Fixes #760

## Test plan

- [x] Verified against the pinned jq oracle across 4037 individual days spanning 11 years chosen for known ISO-week edge cases (leap years, 53-week years like 2015/2020, both directions of the year-boundary week reassignment — e.g. a Monday in late December belonging to next year's ISO week 1, and a Friday Jan 1 belonging to the previous year's ISO week 53) — zero mismatches
- [x] `jq_golden_conformance` passes with `strftime_unsupported_specifiers` removed from the known-failures manifest (now empty of datetime entries)
- [x] Confirmed `%s` matches `mktime`'s output exactly, and `mktime` itself is unaffected by the shared-helper extraction

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -16810,13 +16810,19 @@ fn builtin_strftime<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         }
         _ => match to_owned(&value) {
             OwnedValue::Array(arr) => {
-                if arr.len() < 6 {
+                // Real jq requires the full 8-element array — weekday and
+                // yearday included — not just the 6 fields mktime needs.
+                // Confirmed empirically: `[2024,0,15,0,0,0] | strftime(...)`
+                // errors in real jq for every length 1-7, only succeeding at
+                // 8. Below 8, this codebase used to default weekday/yearday
+                // to 0 instead of erroring, which was harmless before this
+                // PR added specifiers that read those fields (#760) — now it
+                // would silently produce a wrong date instead of matching
+                // jq's error.
+                if arr.len() < 8 {
                     if optional {
                         return QueryResult::None;
                     }
-                    // Real jq reports the same generic wording for a
-                    // too-short array as for any other invalid input —
-                    // it has no distinct "at least 6 elements" message.
                     return QueryResult::Error(
                         EvalError::strftime_requires_parsed_datetime_inputs(),
                     );
@@ -16839,8 +16845,8 @@ fn builtin_strftime<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
                     get_int(3),
                     get_int(4),
                     get_int(5),
-                    if arr.len() > 6 { get_int(6) } else { 0 },
-                    if arr.len() > 7 { get_int(7) } else { 0 },
+                    get_int(6),
+                    get_int(7),
                 )
             }
             _ if optional => return QueryResult::None,
@@ -16960,7 +16966,7 @@ fn format_strftime(
                 }
                 Some('W') => {
                     // Monday-first week number.
-                    let weekday_mon = (weekday + 6) % 7; // Sunday=0..Sat=6 -> Monday=0..Sun=6
+                    let weekday_mon = (weekday + 6).rem_euclid(7); // Sunday=0..Sat=6 -> Monday=0..Sun=6
                     let week = (yearday - weekday_mon + 7).div_euclid(7);
                     result.push_str(&format!("{week:02}"));
                 }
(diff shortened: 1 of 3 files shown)


newhoggy:
Ran a multi-angle standard-effort `/code-review` pass on this PR. Findings and how they were handled:

**Fixed** (new commits):
- **Real jq actually requires the full 8-element array for `strftime`, not 6** (the threshold my earlier #759 fix got wrong, though it did get the error message right). Confirmed empirically against the pinned oracle for every array length 1-7. Before this fix, a 6-7 element array silently defaulted `weekday`/`yearday` to 0 — harmless before this PR (nothing consumed those fields), but this PR's new `%U`/`%V`/`%W`/`%G`/`%g` specifiers *do* consume them, so it would have silently produced a wrong date (including a wrong year) instead of matching jq's error. This was the most severe finding, independently reproduced by multiple review passes.
- `%W`'s weekday-to-Monday-first remap used plain `%` instead of `rem_euclid`, inconsistent with every other new arm in this same diff — fixed for consistency, plus a regression test.
- Added a doc comment to `weeks_in_iso_year`'s `p()` closure explaining the ISO week-date "long year" algorithm and its magic constants.
- Fixed two pre-existing tests (`jq_evaluator_parity_tests.rs`) that used a 6-element `strftime` array — their expectations never actually matched real jq (verified: real jq also errors on that exact input), they were asserting succinctly's own prior buggy leniency.

**Filed as follow-up** (updated #869, already open from #760's own prior review): several integer-overflow panics in both the reused `mktime`-derived math (`%s`, pre-existing, now reachable via a second path) and the brand-new `%U`/`%W`/ISO-week arithmetic — all require a hand-crafted array with deliberately extreme field values (not reachable via `gmtime`/`localtime`/`strptime` output), consistent with #869's existing scope. Also noted two more duplicate copies of the shared civil-calendar-day formula (`strptime`, `parse_iso8601`) and a closely-related discovery that real jq's `mktime` *also* requires 8 elements (out of scope for this PR, which only touches `strftime`).

**Not changed**: minor near-duplication between `%U`/`%W` arms, and `%V`/`%G`/`%g` each independently calling `iso_week_date` (up to 3x redundant O(1) computation per format string) — both flagged as low-severity/nice-to-have by every reviewing pass, skipped to keep the diff focused.

If a deeper pass is wanted before merge, a human can trigger `/code-review ultra` — not run here since it needs explicit user confirmation and billing.

Issue #852, opened 2026-08-13
yq: a bare top-level/navigated scalar document root doesn't drop its styling like real yq does


newhoggy:
**Severity: Low**

## Summary

Real `yq` drops *all* styling (quotes, block-scalar indicators) for a
scalar value when it is the entire top-level output — either because the
whole document is a bare scalar, or because a query navigated to and
returned a single scalar as its result. `succinctly yq` keeps the styling
in this case, for every scalar style (quoted, plain, and block).

## Repro

```
$ printf '"hello world"\n' | yq '.'
hello world
$ printf '"hello world"\n' | succinctly yq '.'
"hello world"

$ printf 'a: "hello world"\n' | yq '.a'
hello world
$ printf 'a: "hello world"\n' | succinctly yq '.a'
"hello world"

$ printf 'a: |\n  line1\n  line2\n' | yq '.a'
line1
line2
$ printf 'a: |\n  line1\n  line2\n' | succinctly yq '.a'
|
line1
line2
```

Note this is *not* about dropping the styling when the scalar is nested
inside a mapping/sequence — `a: "hello world"` correctly keeps its quotes
when output as part of the whole document (`yq '.'` on that input). It's
specifically when a scalar is the outermost/document-level result itself
(via `.` on a bare-scalar document, or via navigation like `.a`) that real
yq strips all styling.

## Root cause

Not investigated in depth. `YamlCursor::stream_yaml`/`stream_yaml_as_
document` (src/yaml/light.rs) already special-case a bare scalar/null/
alias document *root* for one thing — dropping its own trailing comment,
per the existing doc comment on `stream_yaml_as_document`:

> Scalar/null/alias document roots are excluded even here... Only
> Mapping/Sequence roots... keep it.

This suggests real yq has a broader "document-root scalar" special case
that this codebase has only partially implemented (comment-dropping, not
style-dropping). The style-dropping behavior likely needs a similar
top-level-only code path rather than a change to `stream_yaml_value`
itself (which is also used recursively for nested scalars, where styling
must NOT be dropped).

## Why this is separate from #836

#836 is specifically about block scalars (`|`/`>`) losing their style
entirely and coming back as a double-quoted string with `\n` escapes —
that bug is now fixed for the streaming/cursor path (PR #851). This issue
is different in two ways: (1) it affects *every* scalar style, not just
block scalars, and (2) it's about a bare top-level/navigated scalar
incorrectly *keeping* its styling, the opposite direction from #836's
[...]

Issue #712, opened 2026-08-09
yq: merge key (`<<`) is eagerly flattened on identity output


newhoggy:
**Severity: Medium** — silent structural rewrite of untouched input on a plain identity query.

## Summary

A YAML merge key (`<<: *anchor`) is expanded/flattened into its resolved form even on an unmodified identity query — the `<<:` syntax itself doesn't survive a round-trip. Real `yq` keeps the merge key as-is unless a query actually touches the affected mapping.

## Repro (verified against installed `yq` v4.53.3)

```
$ printf 'default: &d\n  a: 1\nitem:\n  <<: *d\n  b: 2\n' | yq '.'
default: &d
  a: 1
item:
  <<: *d
  b: 2
$ printf 'default: &d\n  a: 1\nitem:\n  <<: *d\n  b: 2\n' | succinctly yq '.'
default:
  a: 1
item:
  a: 1
  b: 2                     # `<<: *d` is gone, flattened into item on a no-op query
```

## Related, deeper design note already in the codebase

`src/yaml/light.rs:9581` (`test_merge_key_source_own_merge_key_not_recursively_expanded`) already documents that succinctly's merge-key model deliberately diverges from real `yq`'s traversal-order-dependent mutation for merge-of-a-merge chains — that's a separate, already-acknowledged tradeoff. This issue is about something upstream of that: even a *single-level*, non-chained merge key loses its own syntax on a plain identity query, which isn't about traversal order at all.

## Why this matters

This is the same failure mode as the alias-staleness issue (see #711) — likely a shared root cause where anchors/merge keys are resolved to a value once rather than kept as a lazy/live structure. Combined with the total absence of merge-flag support (`*+`/`*?`/`*n`/`*d`/`*c`, see #713), this suggests succinctly's merge-key handling is currently closer to "parse-time normalization" than "yq-compatible merge operator."

## Notes

- Found via a systematic gap-audit comparing succinctly's yq test coverage against mikefarah/yq's own test suite (2026-08-10).

Issue #710, opened 2026-08-09
yq: comments are silently dropped from all output


newhoggy:
**Severity: Medium-High** — silent data loss on every single output, with no builtin to even read a comment back, let alone preserve one.

## Summary

`succinctly yq` has no `comment`/`line_comment`/`head_comment`/`foot_comment` builtin at all — it's not merely broken, the function doesn't exist:

```
$ printf 'a: 1 # keep this\n' | succinctly yq '.a | comment'
Error: undefined function: comment
$ printf 'a: 1 # keep this\n' | succinctly yq 'line_comment'
Error: undefined function: line_comment
```

And comments are silently dropped from every output, including a no-op identity query:

```
$ printf 'a: 1 # keep this\nb: 2\n' | succinctly yq '.a = 5'
a: 5
b: 2                       # "# keep this" is gone, no warning
```

## What real yq does

Real `yq` (v4.53.3) treats comments as first-class node metadata via its comment operators (`line_comment`, `head_comment`, `foot_comment`, get and set forms), backed by `operator_comments_test.go` (~28 test scenarios covering attach/detach across map keys, arrays, `split_doc`, and the `...` deep-comment-strip form):

```
$ printf 'a: 1 # keep this\nb: 2\n' | yq '.a = 5'
a: 5 # keep this
b: 2
```

## Why this matters

This project's own docs treat silent data loss as a worse failure mode than an explicit error (`docs/compliance/yaml/limitations.md`, describing a since-fixed parser bug: *"this yielded `[null]` — the content was silently discarded, which is a worse failure mode than either erroring or keeping the text"*). Comment-stripping is the same failure mode, just not yet fixed, and it happens on every document that has comments, not an edge case.

## Scope note

This is a real feature gap, not a one-line fix — it likely needs comment text to be threaded through the semi-index (similar to how anchor/style metadata would need to be, see #709) and then through the write path (see #707). Filing this primarily to make the gap trackable; whether preserving comments is in scope for a semi-indexing/streaming architecture at all is a scope decision the maintainers should make explicitly rather than it staying an undocumented silent gap.

## Notes

- Found via a systematic gap-audit comparing succinctly's yq test coverage against mikefarah/yq's own test suite (2026-08-10).

newhoggy:
## Investigation + scoping (2026-08-10)

Traced how comments are currently handled end-to-end (parser → index → cursor → write path) to turn "this likely needs comment text threaded through the semi-index and write path" into a concrete plan.

[...]

PR #873, opened 2026-08-14
fix(yaml): drop a bare top-level/navigated scalar's styling on the M2 streaming path (#852)

newhoggy:
## Summary

- Real `yq` drops all styling (quotes, `|`/`>` block-scalar indicators)
  for a scalar value when it is the *entire* top-level output — either
  the whole document is a bare scalar, or a query navigated to and
  returned a single scalar. `succinctly yq` kept the styling in this
  case on the M2/cursor-streaming read path (issue #739, merged, fixed
  the separate DOM/write path only).
- `YamlCursor::stream_yaml_as_document` (`src/yaml/light.rs`) is the
  shared entry point for both the bare-document (`.`) and
  navigated-field (`.a`) cases (confirmed via #793a's
  `OneCursor`/`ManyCursor` → `stream_yaml_as_document` wiring in
  `eval_generic.rs`), so one fix covers both: intercept `String` roots
  there and write the decoded value directly, bypassing
  `stream_yaml_value`'s normal quote/block-scalar logic entirely — the
  same way the function already special-cases dropping a scalar root's
  own trailing comment (#710) and anchor (#712, via
  `write_leading_anchor`).
- Also fixes a stale doc comment claiming the bare `stream_yaml` (not
  `stream_yaml_as_document`) backs those `GenericResult` arms.
- This also resolves a previously-real data-loss bug (re-emitting block
  style at the root's zero indent produced invalid, unparseable output)
  by construction, since no block scalar is written at the root at all
  anymore — superseding the "fall back to quoting" workaround that used
  to guard it. Verified the new round-trip behavior matches real `yq`'s
  own (YAML plain-scalar line-folding, not a succinctly-specific loss).

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full workspace suite green, 532 yq CLI tests)
- [x] New tests for double/single-quoted and ambiguous-content root scalars, a nested-scalar-keeps-style contrast case, and the block-scalar round-trip — every assertion verified against the pinned real `yq` binary
- [x] Rewrote the one existing test that explicitly pinned the superseded "fall back to quoting" workaround
- [x] `omni-dev coverage diff` reviewed for patch coverage

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -1295,7 +1295,21 @@ fn output_value<W: Write>(
     // For YAML output format (default)
     if config.output_format == OutputFormat::Yaml {
         // For YAML, scalars are printed without quotes by default (like -r in yq)
-        let body = emit_yaml_value(value, comments, config, "", false);
+        // A bare top-level result drops all of its own styling (#852,
+        // mirroring `YamlCursor::stream_yaml_as_document`'s and
+        // `OwnedValue::stream_yaml`'s identical root-only special case for
+        // the cursor/streaming paths) - `output_value` is the actual top
+        // level for every result that reaches it (its own recursion never
+        // calls back into `output_value`, only `emit_yaml_value`), so this
+        // covers every caller uniformly: the main eval path, `-n`,
+        // `--raw-input`, `--split-exp`, ... Redundant with (but harmless
+        // alongside) `evaluate_yaml_cursor`'s equivalent root-scalar pass
+        // for the cursor-based DOM path specifically.
+        let body = if let OwnedValue::String(s) = value {
+            s.clone()
+        } else {
+            emit_yaml_value(value, comments, config, "", false)
+        };
         // Every non-root node's own trailing comment is appended by its
         // *parent* during `emit_yaml_value`'s recursion (see its Array/Object
         // arms), but the root has no parent call site to do that for it —
--- a/src/jq/stream.rs
+++ b/src/jq/stream.rs
@@ -820,6 +820,23 @@ mod tests {
         assert_eq!(buf, "null");
     }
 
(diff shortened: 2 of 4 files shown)


newhoggy:
Patch coverage: 80% (4/5 new lines). The one uncovered line (`light.rs:1326`) is the invalid-UTF-8 fallback (`Err(_) => out.write_str("\"\"")`) — it mirrors `stream_yaml_string_value`'s own identical, already-uncovered fallback for the same reason: the test harness's `run_yq_stdin` takes `&str` input, which can't represent invalid UTF-8 at the type level, so there's no way to reach this branch through the existing string-based test infrastructure without unrelated raw-bytes plumbing. Left uncovered as genuinely untestable through the established harness, consistent with the pre-existing sibling gap.

newhoggy:
Code review (4 independent agents) found a real gap: the fix only covered the M2/cursor path (`stream_yaml_as_document`), leaving two sibling top-level entry points still quoting ambiguous root strings — `OwnedValue::stream_yaml` (`src/jq/stream.rs`, the computed-value path: `-n` construction, arithmetic, etc.) and `output_value` (`src/bin/succinctly/yq_runner.rs`, the actual top-level entry for `-n`/`--raw-input`/`--split-exp`). Fixed both with the identical special case. This also surfaced 8 pre-existing tests that hardcoded the old (pre-#852) quoted-empty-string/quoted-ambiguous-string behavior — one of them (`test_line_comment_builtin_empty_when_absent_710`) had already documented this exact gap in its own doc comment as a known, unrelated discrepancy. All updated to the corrected, oracle-verified output. Several other review findings (control characters, YAML document markers `---`, values starting with `#`) turned out to be exact byte-for-byte matches with real `yq`'s own (surprising) root-scalar behavior, not regressions — verified against the pinned binary.

PR #886, opened 2026-08-14
fix(jq): path(paths(node_filter)) raises before a later break, not silence

### LINKED ISSUE #861: jq: path(paths(node_filter)) diverges from real jq's Invalid-path-expression error
**Severity: Low**

## Summary

`path(paths(node_filter))` diverges from real jq: wrapping the `paths(node_filter)` builtin's own output inside `path(...)`'s trackable-path machinery produces empty output in succinctly, while real jq raises `Invalid path expression with result ...`.

## Repro

```
$ echo '[1,"x",2]' | jq -c 'label $out | path(paths(if type=="string" then break $out else true end))'
jq: error (at <stdin>:0): Invalid path expression with result [0]
$ echo '[1,"x",2]' | succinctly jq -c 'label $out | path(paths(if type=="string" then break $out else true end))'
(no output, exit 0)
```

Real jq treats the plain array `[0]` (the value `paths(...)` yields, since it's a value-position evaluation there, not itself further path-tracked) as an invalid path expression and raises. succinctly instead silently produces no output at all.

## Root cause

Not diagnosed. Found while verifying #850's fix (`builtin_paths_filter`'s per-node loop no longer swallows an ordinary error/break, matching real jq for the *plain* value-position `paths(node_filter)` case — confirmed exhaustively, including `break` correctly reaching an outer `label`). That verification also tried the same repro wrapped in `path(...)`, which surfaced this separate, pre-existing divergence — `path(...)`'s own trackable-path resolution of an expression containing `paths(...)` is a different code path (`resolve_node`, not `builtin_paths_filter`'s `eval_owned_input`-based evaluator) and isn't touched by #850's fix.

## Why this is separate from #850

#850 is scoped to `builtin_paths_filter`'s own per-node control-signal handling in *value* position (plain `paths(node_filter)`, not wrapped in `path(...)`). That fix is confirmed complete and correct for every case tested, including nested matches and `break`-to-outer-label. This issue is about a different evaluator path (`path(...)` wrapping `paths(...)`) that #850 never touches.

## Suggested fix direction

Needs investigation first: trace how `path(EXPR)` handles `EXPR` when `EXPR` contains `paths(node_filter)` — whether `resolve_node` has a `Builtin::PathsFilter` arm at all, and if not, what value-position fallback it uses that produces empty output instead of raising real jq's "Invalid path expression" error.

newhoggy:
## Summary

`path(paths(node_filter))` diverges from real jq: wrapping `paths(node_filter)`'s own output inside `path(...)`'s trackable-path machinery produced empty output in succinctly, where real jq raises `Invalid path expression with result ...`.

Root cause: `resolve_node` has no `Builtin::PathsFilter` arm, so it falls through to `resolve_leaf`'s generic fallback, which evaluated the whole argument via `eval_owned_multi` — an all-or-nothing evaluator that discards any already-produced output the instant a *later* output errors or breaks. Real jq's `path(...)` checks each output as it streams: the first non-path-shaped value raises `Invalid path expression` immediately, before a later output — even one that would itself `break` to an outer label — is ever produced. `resolve_leaf` was losing the already-produced first value to the later break, ending up with zero outputs and no error at all.

Fix: switch to `eval_owned_multi_keep_partial`, the same "prefix survives a later escape" evaluator already used for the identical all-or-nothing pitfall in `recurse`'s own fan-out (#842). The already-produced first value now survives the later break and gets checked, matching jq's per-output semantics.

## Repro

```
$ echo '[1,"x",2]' | jq -c 'label $out | path(paths(if type=="string" then break $out else true end))'
jq: error (at <stdin>:0): Invalid path expression with result [0]
$ echo '[1,"x",2]' | succinctly jq -c 'label $out | path(paths(if type=="string" then break $out else true end))'
(no output, exit 0)      # before this fix
jq: error (at <stdin>:1): Invalid path expression with result [0]   # after this fix, matches jq
```

## Scope note

While investigating, I confirmed `path(range(3))` is a **separate, pre-existing** divergence (succinctly raises "Cannot use a computed index after a multi-output path component"; real jq raises "Invalid path expression with result 0" on the first output, same pattern as this issue but without an escape involved). Left untouched — it's a different code path (the existing `_ =>` 2+-output arm in `resolve_leaf`, not exercised by this issue's repro) and out of scope for #861's own repro. Noted inline in the diff; can file as a follow-up if wanted.

Fixes #861

## Test plan

- [x] Diffed output against pinned real `jq` (1.7.1) directly for the repro, `path(range(3))`, and the existing `path(1)`/`path(length)`/`path(keys)`/`path(.a + 1)` cases (all still match)
- [x] Added `test_path_paths_filter_raises_before_reaching_a_later_break_861` pinning the fix

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -10282,7 +10282,15 @@ fn resolve_leaf<'a, S: EvalSemantics>(
             ));
         }
     }
-    let mut values = eval_owned_multi::<S>(expr, value).map_err(|e| (Vec::new(), e))?;
+    // `eval_owned_multi_keep_partial`, not `eval_owned_multi`: a later output
+    // of `expr` erroring or breaking must not retroactively un-produce an
+    // earlier one it already yielded (#842's precedent for `recurse`'s own
+    // fan-out, applying equally here) — see the "streams as it goes" case
+    // below (#861).
+    let (mut values, trailing) = match eval_owned_multi_keep_partial::<S>(expr, value) {
+        Ok(values) => (values, None),
+        Err((prefix, escape)) => (prefix, Some(escape)),
+    };
     if trackable
         && matches!(
             expr,
@@ -10292,8 +10300,15 @@ fn resolve_leaf<'a, S: EvalSemantics>(
         let mut components = Vec::new();
         push_path_components(&mut components, expr);
         return match values.len() {
-            // No output prunes the branch.
-            0 => Ok(Vec::new()),
+            // No output prunes the branch — unless there never would have
+            // been one because evaluating this leaf itself escaped (these
+            // primitives don't fan out, so a non-empty prefix can't coexist
+            // with a trailing escape in practice, but propagate rather than
+            // silently prune if that ever changes).
+            0 => match trailing {
+                Some(escape) => Err((Vec::new(), escape)),
+                None => Ok(Vec::new()),
+            },
             1 => Ok(vec![(
                 components,
                 Cow::Owned(values.pop().expect("len checked")),
@@ -10311,7 +10326,21 @@ fn resolve_leaf<'a, S: EvalSemantics>(
         };
     }
     match values.len() {
-        0 => Ok(Vec::new()),
+        0 => match trailing {
+            Some(escape) => Err((Vec::new(), escape)),
+            None => Ok(Vec::new()),
+        },
+        // Real jq's `path(...)` checks each output as it streams: the first
+        // non-path-shaped value raises "Invalid path expression" immediately,
+        // before a later output of the same expression is ever produced —
+        // even one that would itself error or break (confirmed live:
+        // `path(range(3))` raises on `0` alone, never reaching `1`/`2`; #861's
+        // own repro, `path(paths(if type=="string" then break $out else true
+        // end))` on `[1,"x",2]`, raises on `[0]` alone, never reaching the
+        // second candidate where the `break` sits). `values[0]` is that first
+        // output regardless of whether `trailing` also carries an escape from
+        // evaluating further here — real jq never reaches whatever would have
+        // caused it, so the escape is discarded rather than propagated.
         1 => Err((
             Vec::new(),
             EvalError::invalid_path_expression(&values[0]).into(),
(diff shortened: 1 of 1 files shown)


newhoggy:
## Review update

Ran `/code-review` (standard effort) against this branch. Multiple independent review passes converged on one real finding, which I independently verified against real `jq 1.7.1` before acting:

**Confirmed and fixed**: switching `resolve_leaf` from `eval_owned_multi` (auto-propagate any escape via `?`) to `eval_owned_multi_keep_partial` (keep the prefix, thread the escape separately) meant a trailing `Halt` could get discarded and folded into an ordinary catchable "Invalid path expression" error once 1+ values had already been produced — verified live: `path(paths(if type=="string" then (.|halt_error) else true end))` on `[1,2,"trigger"]` no longer halted, and worse, the halt became catchable via `try`/`catch`. `halt`/`halt_error` must reach the CLI unconditionally per `EvalEscape::Halt`'s own doc comment, so this is fixed in a follow-up commit: Halt is now always propagated regardless of prefix length, hoisted ahead of both length-based matches.

**Investigated and NOT changed** (verified against real jq, the review's initial framing was based on comparing against this branch's own pre-diff behavior rather than real jq, which turned out to also diverge): the review also flagged `Break` and `Error` discarding as potential regressions. I reproduced the exact scenarios against real `jq 1.7.1` directly and confirmed jq's own `path(...)` never even reaches a later candidate once an earlier one fails its path-shape check — e.g. `path(paths(if type=="string" then error("msg") else true end))` on `[1,2,"trigger"]` raises `Invalid path expression with result [0]` in real jq too, never surfacing `"msg"`. So discarding Break/Error here is the objectively correct match for jq's real behavior, not a regression — this branch's pre-diff behavior (propagating them) was itself the divergence.

**Filed as a follow-up**: #891, covering the pre-existing (not introduced by this PR) `values.len() >= 2` message-shape divergence (`path(range(3))` already showed the wrong message on `main`) and the related eager-vs-lazy evaluation architecture gap that both issues share a root cause in. Out of scope for #861's own narrow repro.

Also fixed inline: a stale doc comment on `eval_owned_multi_keep_partial` that didn't list `resolve_leaf` as a caller, and de-duplicated a repeated match arm the review flagged.

All commits pushed. Full test suite (0 failures), both required clippy invocations, and manual diffs against real jq all still green after these changes.

PR #887, opened 2026-08-14
fix(jq): recurse(f; cond) keeps cond-approved siblings before a later cond error

### LINKED ISSUE #854: jq: recurse(f; cond) drops cond's already-approved siblings on a later cond error

newhoggy:
## Summary

- `recurse(f; cond)`'s `cond`-evaluation loop (both the value-position `builtin_recurse_cond` and the path-position `resolve_recurse`, in `src/jq/eval.rs`) dropped the already-`cond`-approved siblings of a node's children when `cond` itself errored on a *later* child in the same batch — real jq's generator semantics give each already-approved child its own full recursive descent before `cond` is ever asked about the next candidate child.
- Reuses #842's `queue_recurse_children`/`pending_error` deferred-error mechanism (already used for `f`'s own fan-out) for `cond`'s error too: on a `cond` error, the loop now stops considering further children but lets whatever already passed `cond` this batch (`next`) get its own full recursive descent before the error surfaces.
- `cond`'s error supersedes `f`'s own later deferred error (not the other way around) — verified against real jq: a `cond` error on an already-`f`-produced child always happens before `f` would ever be asked to produce a later one, in real generator order.

```
$ echo '[1,2,3]' | jq -c 'recurse(if type=="array" then .[] else empty end; if . == 2 then error("boom") else true end)'
[1,2,3]
1
```

Before this fix, `succinctly jq` printed only `[1,2,3]`, dropping the already-approved `1`. After the fix, output matches real jq exactly (value position and `path(...)` position both verified).

Fixes #854

## Test plan

- [x] Manually diffed output against real jq 1.7.1 across several scenarios: original repro, multiple approved siblings, error on the very first child, no-error sanity check, `break` through `cond`, and the `cond`-error-vs-`f`-own-later-error priority ordering — all match exactly
- [x] Added unit tests in `src/jq/eval.rs` (value + path position, plus the priority-ordering case) and CLI-level tests in `tests/jq_cli_tests.rs` (stdout/stderr/exit-code), all pinned against real jq 1.7.1

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -35663,6 +35663,65 @@ mod tests {
     }
 
     #[test]
+    fn test_recurse_cond_keeps_already_approved_siblings_before_cond_error_854() {
+        // #854's own bug shape, distinct from #842 above: here `f` never
+        // errors (it fans out cleanly over every array element); `cond`
+        // itself errors on a later child, after already approving an
+        // earlier one in the same batch. Confirmed against real jq 1.7.1:
+        // `echo '[1,2,3]' | jq -c 'recurse(if type=="array" then .[] else
+        // empty end; if . == 2 then error("boom") else true end)'` prints
+        // the root `[1,2,3]` *and* `1` (the first child, already
+        // `cond`-approved and fully recursed before `cond` errors on the
+        // second child) before erroring. Before this fix, succinctly
+        // printed only the root, dropping `1`.
+        assert_eq!(
+            outputs(
+                br"[1,2,3]",
+                r#"recurse(if type=="array" then .[] else empty end; if . == 2 then error("boom") else true end)"#
+            ),
+            vec!["[1,2,3]", "1"]
+        );
+    }
+
+    #[test]
+    fn test_resolve_recurse_cond_keeps_already_approved_siblings_before_cond_error_854() {
+        // Path-position counterpart of the previous test (`resolve_recurse`'s
+        // `Some(cond)` arm). Confirmed against real jq 1.7.1: `echo
+        // '[1,2,3]' | jq -c 'path(recurse(if type=="array" then .[] else
+        // empty end; if . == 2 then error("boom") else true end))'` prints
+        // `[]` *and* `[0]` before erroring.
+        assert_eq!(
+            outputs(
+                br"[1,2,3]",
+                r#"path(recurse(if type=="array" then .[] else empty end; if . == 2 then error("boom") else true end))"#
+            ),
+            vec!["[]", "[0]"]
+        );
+    }
+
+    #[test]
+    fn test_recurse_cond_error_supersedes_f_own_later_deferred_error_854() {
+        // When both `f` and `cond` would eventually error at the same node,
+        // `cond`'s error on an already-`f`-produced child must win over `f`'s
+        // own later deferred error (from #842) — real jq's generator
+        // semantics evaluate `select(cond)` on each child as `f` produces
+        // it, so a `cond` error on an earlier child always happens before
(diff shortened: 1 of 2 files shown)


newhoggy:
Superseded by #859, which was discovered mid-review to already fix this issue (a prior session's work, more complete — it also covers a deeper "cond's own multi-output fan-out" bug my review found missing here). Closing this one to avoid duplicating work; continuing on #859 instead. See the comment on #854 for the full trail.

Issue #878, opened 2026-08-14
yaml: trailing content after a flow collection's closing delimiter silently drops sibling fields instead of erroring


newhoggy:
## Summary

Real YAML (and real `yq`) treats content after a flow collection's closing `]`/`}` on the same line (other than whitespace/a comment) as a hard parse error. `succinctly yq` doesn't validate this - it silently continues, and depending on where the malformed value sits, either loses the collection's own trailing content, or drops every sibling field that follows it in the enclosing mapping, all with exit code 0 and no error.

This is a pre-existing gap (confirmed identical on `main` before #864/#875, both at plain top level and inside an ordinary block mapping - not something #864/#875 introduced), surfaced while verifying #875's review findings against a temporary comparison worktree.

## Repro

Top level:
```
$ printf -- 'a: [note] see below\nb: 2\nc: 3\nd: 4\n' | yq -o json '.'
Error: bad file '-': yaml: while parsing a block mapping ...: did not find expected key
$ printf -- 'a: [note] see below\nb: 2\nc: 3\nd: 4\n' | succinctly yq -o json '.'
{"a":["note"],"see below":"b","2":"c","3":"d"}
```

Ordinary block-mapping field, non-first:
```
$ printf -- '- x: 1\n  a: [note] see below\n  b: 2\n  c: 3\n  d: 4\n' | succinctly yq -o json '.'
[{"x":1,"a":["note"]}]
```
(`b`, `c`, `d` vanish entirely, exit 0.)

Same shape as a compact block-sequence item's *first* field (the #864/#875 code path) behaves identically - not worse, just now also present there since #875 correctly gave that path real flow-collection parsing instead of a scalar-fallback that happened to avoid triggering this specific gap by never really parsing the collection at all.

## Scope note

Per `CLAUDE.md`'s documented semi-indexing trade-off ("Semi-indexing performs minimal validation compared to full parsers"), *not* matching `yq`'s hard-error behavior in general may be an acceptable trade-off - but silently and inconsistently dropping sibling fields (rather than, say, consistently treating the rest of the line as part of the collection, or consistently erroring) seems like an accident of the scalar-scanning implementation rather than a deliberate choice, and is worth a real decision either way.

## Suggested fix direction

[...]

Issue #876, opened 2026-08-14
yaml: unify the 4 hand-duplicated flow/block-scalar inline-value dispatch arms in parser.rs


newhoggy:
## Summary

`src/yaml/parser.rs` has four separate functions that each hand-roll an identical 3-arm dispatch table for inline values:

```rust
Some(b'[') => { self.parse_flow_sequence()?; }
Some(b'{') => { self.parse_flow_mapping()?; }
Some(b'|' | b'>') => { self.parse_block_scalar(indent)?; }
```

The four sites (as of `main` after #875):

- `parse_value` (~line 2868) - the general-purpose value parser, already reused by `parse_sequence_item_inner`, `parse_explicit_key`, and others via `self.parse_value(indent)`.
- `parse_mapping_entry` (~line 2438)
- `parse_explicit_value` (~line 2751)
- `parse_compact_mapping_entry` (~line 2080, added by #864/#875)

(There are 7 further `Some(b'[') =>` sites in the file - `parse_explicit_key`, `parse_explicit_flow_key_node`, `parse_explicit_flow_mapping_entry`, `parse_implicit_flow_mapping_entry`, `parse_flow_sequence_inner`, `parse_flow_mapping_inner`, `parse_flow_key` - but those are flow-*context* dispatch with different semantics (already inside `{...}`/`[...]`), not copies of this specific block-context value dispatch, and are out of scope here.)

## Why this matters

This is the same bug class recurring, not a new one: #325, #372, #406, #224, #785, and most recently #864 were all "one of these hand-copied dispatch tables is missing an arm (or has a stale ordering bug) that a sibling table already has." #864 specifically was `parse_compact_mapping_entry` missing the `[`/`{`/`|`/`>` arms that `parse_mapping_entry` already had - two of these same four sites silently diverging from each other.

`parse_explicit_value` already demonstrates the fix pattern one arm above this exact dispatch: its `-` (block-sequence-indicator) arm used to be a third divergent copy of the same decision and was rewritten to call the shared `parse_sequence_item` helper specifically to close this class of bug (see the comment there citing #106: "duplicated predicates diverge silently - one definition, plus a test that the call sites agree"). The `[`/`{`/`|`/`>` arms right below it were left as hand-rolled duplicates when that fix landed.

Two independent review angles on #875 (`git diff` cleanup/reuse review and an altitude/conventions review) both surfaced this independently and pointed at the same root cause.

## Suggested fix direction

Extract a shared helper, e.g.:

```rust
[...]

Issue #864, opened 2026-08-14
yaml: compact block-sequence item's first field loses a flow-collection value entirely, corrupting sibling parsing


newhoggy:
**Severity: High**

## Summary

A block-sequence item using the compact `- key: value` form (dash and first
key share one line) silently corrupts parsing when that **first** field's
value is a flow collection (`{...}`/`[...]`). The flow value's own content
is lost, and for a flow *mapping* specifically, the parser's position gets
stranded mid-line, causing the *next* sibling field to be misread as
garbled structure instead of a normal key.

Found incidentally while verifying #739's oracle diffs — unrelated to that
issue's own scope (style/comment preservation), reproduces identically on
unpatched `main`.

## Repro

```
$ printf -- '- a: {x: 1}\n  b: 1\n' | yq -o json '.'
[{"a":{"x":1},"b":1}]
$ printf -- '- a: {x: 1}\n  b: 1\n' | succinctly yq -o json '.'
[{"a":{}}]
```

`b` disappears entirely, not just `x: 1`.

A flow **array** loses its own content too, but `b` survives (see root
cause below for why object vs array differ):

```
$ printf -- '- a: [1, 2]\n  b: 1\n' | succinctly yq -o json '.'
[{"a":[],"b":1}]
```

Nesting one level deeper doesn't just lose data locally — it corrupts the
*outer* mapping too:

```
$ printf 'items:\n  - a: {x: 1}\n    b: 1\n' | succinctly yq -o json '.'
{"items":[{"a":{}}],"1} b":1}
```

Control cases that work correctly (isolating the trigger to this exact
combination):
- Flow value as a *non-first* field of a compact item (`- a: 1\n  b: {x: 2}\n`) — correct.
- The same flow-first-field shape at the plain top level, not inside a block sequence (`a: {x: 1}\nb: 1\n`) — correct.
- A flow mapping as the *entire* sequence-item value (`- {x: 1}\n`, no compact key) — correct (different code path).

## Root cause

`src/yaml/parser.rs`'s `parse_compact_mapping_entry` (parser.rs:1953) is
the compact item's own first-key/value parser. Its inline-value dispatch
fallback arm (parser.rs:2095-2103) calls `parse_inline_value` directly with
no flow-collection awareness:

```rust
_ => {
    self.set_ib();
    self.write_bp_open();
    let end_pos = self.parse_inline_value(indent)?;   // no {/[ handling
    self.set_bp_text_end(end_pos);
    self.write_bp_close();
[...]

PR #898, opened 2026-08-14
fix(yaml): unify flow/block-scalar dispatch, reject trailing flow content

newhoggy:
## Summary

Bundled fix for #876 and #878 — #878's fix is a one-place addition once #876's refactor lands, so implementing them separately would mean either duplicating the trailing-content check across 4 call sites first (then deleting 3 of the copies) or leaving it duplicated. Same file, same functions, cited by both issues.

**#876**: `parse_compact_mapping_entry`, `parse_mapping_entry`, `parse_explicit_value`, and `parse_value` each hand-rolled an identical copy of the `[`/`{`/`|`/`>` inline-value dispatch table — the same bug class recurring six times now (#325, #372, #406, #224, #785, #864), where one copy silently diverges from a sibling's (most recently: #864 was `parse_compact_mapping_entry` missing arms `parse_mapping_entry` already had). Extracted a shared `try_dispatch_flow_or_block_value` helper, called from all four sites. Each site keeps its own remaining arms (`-`/quoted/alias/plain-scalar) — they differ enough in content and relative ordering between sites that folding them in too would trade one duplication bug class for a worse one, per #876's own scope note.

**#878**: real YAML treats content after a flow collection's closing `]`/`}` on the same line — other than whitespace or a comment — as a hard parse error. This loader silently continued instead, corrupting the rest of the document (dropping every sibling field after a malformed value) with no error and exit code 0. Added `reject_trailing_flow_content`, called from the new shared helper right after `parse_flow_sequence`/`parse_flow_mapping` return.

### The one wrinkle: implicit mapping keys

A naive "always check trailing content" version broke the YAML test suite: real YAML allows `[a, b]: value` / `{a: 1}: value` — a flow collection used as an **implicit mapping key**, where `:` legitimately follows the closing delimiter. `parse_value` is reached in that ambiguous position (a sequence item's value once `looks_like_mapping_entry` has already ruled out a *scalar*-keyed compact mapping, and document-root/deferred content starting with `[`/`{`) — a flow collection there might turn out to be a key, not a standalone value. The other three call sites are always reached *after* an unambiguous `key:` has already been parsed, where a following `:` cannot be anything but real garbage.

Fixed by threading a `check_trailing: bool` through the shared helper: `true` for the three unambiguous-value sites, `false` for `parse_value`. Caught this via the full `yaml_test_suite_conformance` run (3 previously-passing test-suite cases newly failing) before it ever reached review — see commit history on this branch.

Note: succinctly's own handling of `[a, b]: value` doesn't actually implement flow-collection-keyed mappings correctly (produces two separate top-level values instead of `{"": "value"}` like real yq) — confirmed this is a **pre-existing, unrelated** gap (identical output on `main` before this branch), not something this PR introduces or fixes. Out of scope here.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 0 failures, including `yaml_test_suite_conformance`'s known-failures manifest — no drift)
- [x] `cargo fmt --check`
- [x] Diffed output against pinned real `yq` for #878's repro and the implicit-mapping-key cases
- [x] Added `test_flow_or_block_value_dispatch_agrees_across_all_four_sites_876`, extending #864's two-site parity test to all four unified sites (explicit key/value, bare sequence item) against the ordinary-mapping-entry baseline
- [x] Added `test_trailing_content_after_flow_collection_errors_878` (document root + non-first block-mapping field repros, plus a trailing-comment-is-still-legal case) and `test_flow_collection_as_implicit_mapping_key_still_permitted_878`

### DIFF
--- a/src/yaml/parser.rs
+++ b/src/yaml/parser.rs
@@ -2074,6 +2074,23 @@ impl<'a, const HAS_CR: bool> Parser<'a, HAS_CR> {
             // `parse_sequence_item_inner` and `parse_explicit_value` where the
             // placeholder is conditional. `test_every_anchor_targets_an_open_bit`
             // is the whole-corpus guard on that.
+        } else if self.try_dispatch_flow_or_block_value(indent, true)? {
+            // Flow sequence/mapping value (`- a: [1, 2]` / `- a: {x: 1}`), or
+            // block scalar value (`- a: |`) — handled. Missing this used to
+            // leave every flow-value fallback through the scalar arm's
+            // `parse_inline_value`, which treats `{`/`[`/`,`/`}`/`]` as
+            // ordinary plain-scalar content instead of structure: a flow
+            // *array* value gets consumed whole as one bogus scalar token
+            // (its real content lost — reads back as `[]`), and a flow
+            // *mapping* value is worse, since the scalar scanner stops at its
+            // first inner `key:`+space and strands `self.pos` mid-line,
+            // corrupting everything parsed after it too (#864). A block
+            // scalar hit the same class of corruption for the same reason:
+            // the plain-scalar scanner doesn't know it's inside literal
+            // content, so a body line shaped like `key: value` or starting
+            // with `#` (legal in a block literal, meaningless as YAML there)
+            // stopped the scan early (confirmed via `git worktree` bisection
+            // against the pre-#864 fallback during review).
         } else {
             // Inline value (`parse_node_properties` ran above)
             match self.peek() {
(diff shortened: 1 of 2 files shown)


newhoggy:
## Review update

Ran `/code-review` (standard effort, multi-agent). Two real regressions found and fixed, both confirmed live against real `yq` before and after the fix:

1. **`reject_trailing_flow_content` wasn't tab-aware** (it reused `at_line_end()`, which only treats a plain space as skippable, not a tab). `a: [1, 2]\t# comment` — valid YAML real `yq` accepts — was being false-positive-rejected. Fixed by giving the check its own tab-aware scan instead of reusing the buggy helper.

2. **`parse_explicit_value`'s `check_trailing: true` was wrong.** Its value position (`? k\n: ...`) is exactly as ambiguous as `parse_value`'s — the node after `: ` can itself be a compact mapping keyed by a flow collection (`: [a, b]: value`), mirroring the scalar-keyed case its own `looks_like_mapping_entry` arm already handles a few lines below. Confirmed live: real `yq` parses `? k\n: [a, b]: value\n` as `{"k":{"":"value"}}`; the unconditional check rejected it outright. Now passes `check_trailing: false`, matching `parse_value`'s reasoning.

Also fixed a test anti-pattern the review flagged: `test_flow_collection_as_implicit_mapping_key_still_permitted_878` only asserted exit code 0, which couldn't catch a regression that kept exit 0 while corrupting output. Now pins actual output for all 4 reachable shapes (document root, bare sequence item, and both `parse_explicit_value` cases newly unblocked by fix #2).

**Filed as a follow-up** (#902), not fixed here — all confirmed pre-existing (identical on `main` before this PR), not regressions:
- `parse_value`'s `check_trailing: false` is a blanket disable rather than narrowed to the one ambiguous shape, so real garbage after a flow collection through that path still silently corrupts instead of erroring.
- `parse_explicit_key`'s own dispatch (deliberately excluded from #876's unification — it wraps the flow-collection calls in an extra BP-node pair the shared helper can't represent) never gained #878's validation either, reproducing the same corruption class.
- A separate, unrelated pre-existing bug: a tab after a flow collection now hits "tab character used for indentation" instead of my old wrong error (progress, but still not matching real yq, which accepts it) — narrow to the flow-collection case, confirmed a plain scalar with the same shape works fine.
- Some reuse/duplication observations (8x duplicated error-construction, `at_line_end()`'s tab-blindness affecting its 9 other callers too, a couple more hand-rolled copies of the `[`/`{`/`|`/`>` byte-set).
- The root feature gap: succinctly doesn't actually implement flow-collection-keyed implicit mappings (`[a, b]: value` doesn't fold into `{"": "value"}` like real `yq` — the new test documents this inline).

All commits pushed. Full test suite, both required clippy invocations, and `cargo fmt --check` all green after these changes.

PR #903, opened 2026-08-14
fix(jq): resolve explicit YAML tags in the generic evaluator

### LINKED ISSUE #747: yaml: generic jq evaluator (type/==/arithmetic/select) ignores explicit tags
## Summary

#224 made YAML tag resolution (`!!str`, `!!int`, `!!bool`, `!!float`, `!!null`) apply
consistently everywhere *except* one path: the generic jq evaluator's cursor-free
materializer, `eval_generic::to_owned`.

```console
$ echo 'a: !!str 1' | succinctly yq '.a | type'
number          # should be "string" — !!str forces string resolution
$ echo 'a: !!str 1' | succinctly yq '.'
{"a": "1"}      # JSON output gets it right
$ echo 'a: !!str 1' | succinctly yq '.a == "1"'
false           # should be true, for the same reason
```

## Root cause

`to_owned` (`src/jq/eval_generic.rs`) is shared between the JSON and YAML
`DocumentValue` implementations and only ever sees a bare `DocumentValue`, never a
cursor. `YamlValue`'s typed getters (`is_null`, `as_bool`, `as_i64`, `as_f64`,
`type_name`) — called by `to_owned`, and in turn by `select`, `==`, arithmetic, and
`type` on the cursor-evaluator path — resolve a plain scalar's shape from its text but
have no way to look up an explicit tag, because that lookup is keyed by `bp_pos`
(`YamlIndex::get_tag`), which only a `YamlCursor` carries. A bare `YamlValue` has
already lost it by the time `to_owned` sees it.

Every *other* path is already tag-aware because it never loses the cursor:
`YamlCursor::write_json_to`/`stream_json_value` (JSON output), the CLI's
`yaml_to_owned_value` (`src/bin/succinctly/yq_runner.rs`), and `load()`'s
`yaml_value_to_owned` (`src/jq/eval.rs`) all recurse via `uncons_cursor`/
`value_cursor` rather than a plain value, precisely so an explicit tag on a nested
node isn't lost. `to_owned` is the one materializer that still recurses on bare
values.

See the `KNOWN LIMITATION (#224)` comment directly above `impl DocumentValue for
YamlValue` in `src/yaml/light.rs` for the full analysis, written when this was
scoped out of #224.

## Fix sketch (per that comment, unverified)

Two options, neither attempted yet:

1. Thread a cursor through `to_owned`'s recursion. Plausible: `DocumentField::value_cursor`
   and `DocumentElements::uncons_cursor` already exist for exactly this. Requires an audit
   of every `to_owned` call site currently passing a bare value with no cursor available
   (`eval_index_expr`/`eval_slice_bound`'s `One`/`Many` arms reach `to_owned` from a
   cursor-less ambient `.`, per `eval_generic.rs`'s own `test_computed_index_key_bare_one_and_many_via_cursor_less_entry`).
2. Embed the tag directly in `YamlValue::String` instead. Rejected in #224 as fanning out
   to ~90 pattern-match sites across 6 files.

## Scope

Confined to the cursor-evaluator path for YAML tags. JSON is unaffected (JSON has no
tag syntax). `succinctly yq '.'`'s JSON/YAML output is unaffected (already correct).

Split out of #224.

newhoggy:
## Summary

- `to_owned`, the generic cursor-evaluator's materializer backing `type`, `==`, arithmetic, and `select`, only ever saw a bare `DocumentValue` — never a cursor — so it had no way to look up an explicit YAML tag (`!!str`, `!!int`, `!!bool`, `!!float`, `!!null`). That lookup is keyed by byte position, which only a `YamlCursor` carries. JSON/YAML output was already correct (`stream_json`/`stream_yaml` stream straight from a cursor), so `succinctly yq '.'` on `a: !!str 1` correctly said `{"a": "1"}` while `.a | type` still said `"number"`.
- Adds `DocumentCursor::explicit_tag` (default `None`; `YamlCursor` overrides it, delegating to the pre-existing inherent `YamlCursor::explicit_tag`), a new `to_owned_cursor` that mirrors `to_owned`'s container-first structure but recurses via cursors (`field.value_cursor`/`elems.uncons_cursor`, the same pattern `to_owned_with_comments` already established) and checks the tag on each scalar, and `tagged_type_name` for `type` specifically (it reads `DocumentValue::type_name()` directly and never goes through `to_owned` at all, so it needed its own fix).
- Every call site in `eval_generic.rs` that already holds a cursor now routes through `to_owned_cursor`/`to_owned_with_cursor` instead of the bare, tag-blind `to_owned`.

## Test plan

- [x] Verified against the pinned real `yq` oracle (v4.53.3) for identity/`==`/arithmetic output; `type`'s own output format is succinctly's own deliberate jq-style choice (already diverges from mikefarah `yq`'s `!!str`-style `type`, pre-existing and unrelated to this fix), verified instead against the issue's own stated expected values and internal consistency with untagged equivalents
- [x] New CLI tests in `tests/yq_cli_tests.rs` covering `type`/`==`/arithmetic/`select`/nested-tag resolution
- [x] New unit test in `eval_generic.rs` for the one reachable-but-obscure shape CLI queries can't easily hit directly (a bare `GenericResult::Many` fan-out from `select`'s cursor-less `pass_n` arm, whose per-item re-evaluation lands back on `OneCursor`/`ManyCursor`)
- [x] 100% patch coverage on all touched source lines (`cargo llvm-cov` + diff cross-reference)

Fixes #747

### DIFF
--- a/src/jq/document.rs
+++ b/src/jq/document.rs
@@ -111,6 +111,16 @@ pub trait DocumentCursor: Sized + Copy + Clone {
         None
     }
 
+    /// Get the explicit YAML tag at this node, if any (e.g. `"!!str"`).
+    ///
+    /// Returns `None` when the node has no explicit tag, and for formats
+    /// without a tag concept (JSON). Only YAML cursors override this —
+    /// tag lookup is keyed by byte position, which only a cursor carries;
+    /// a bare [`DocumentValue`] has already lost it (#747).
+    fn explicit_tag(&self) -> Option<&str> {
+        None
+    }
+
     /// Get the YAML style indicator for this node (e.g. `"flow"`, `"double"`).
     ///
     /// Returns `""` (no explicit style) by default; only YAML cursors
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -5220,6 +5220,40 @@ mod tests {
         );
     }
 
+    /// #903 review round: `Builtin::Shuffle`'s array branch now materializes
+    /// via `collect_cursors`/`to_owned_cursor` instead of
+    /// `collect_values`/`to_owned`, the same fix as `to_entries`/`reverse`/
+    /// `pivot`. An in-process unit test (rather than a CLI subprocess test,
+    /// like the sibling cases in `tests/yq_cli_tests.rs`) because
+    /// `cargo-llvm-cov`'s workspace report doesn't reliably attribute
+    /// coverage back through `Command::new(env!("CARGO_BIN_EXE_succinctly"))`
+    /// for this arm specifically, despite the CLI binary demonstrably taking
+    /// it when run directly. Order isn't checked (`shuffle` permutes), only
+    /// that every element still carries its own resolved type.
+    #[test]
+    fn test_yaml_shuffle_resolves_explicit_tag_903() {
+        use crate::yaml::YamlIndex;
+
+        let yaml = b"a:\n  - !!str 1\n  - !!str 2\n";
+        let index = YamlIndex::build(yaml).unwrap();
+        let cursor = index.root(yaml);
+        let mapping_cursor = cursor
+            .first_child()
+            .expect("YAML document should have content");
+        let value = mapping_cursor.value();
+
+        let result = eval(&crate::jq::parse(".a | shuffle").unwrap(), value);
+        match result.into_owned().unwrap() {
+            OwnedValue::Array(items) => {
+                assert_eq!(items.len(), 2);
+                for item in &items {
+                    assert_eq!(item.type_name(), "string", "{item:?}");
+                }
+            }
+            other => panic!("expected array, got {other:?}"),
+        }
+    }
+
     #[test]
     fn test_yaml_generic_field_access() {
         use crate::yaml::YamlIndex;
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -5120,6 +5120,55 @@ mod tests {
         }
     }
 
+    /// #747: mirrors `test_json_multi_stage_pipe_first_stage_bare_many_lazy_index_range_684`'s
+    /// shape — `select(true,true)` on a cursor-less top-level `eval()` call
+    /// produces a bare `GenericResult::Many` (`Builtin::Select`'s `pass_n`
+    /// closure, `(n, None)` arm), whose per-item re-evaluation in the pipe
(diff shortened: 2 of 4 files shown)


newhoggy:
Ran the mandatory multi-angle review before merge. It found 6 confirmed, real gaps beyond the initial fix — all independently reproduced against the built binary before and after fixing:

1. **`YamlCursor::explicit_tag()` didn't dereference through an alias** to the anchor definition's tag, unlike every other alias-transparent getter on `YamlValue`. A tagged value reached via `*anchor` silently kept its untagged type/comparison/arithmetic result.
2. **Several type-mismatch error messages still read the raw `value.type_name()`** (`Expr::Field`/`Expr::Index`, and `First`/`Last`/`Reverse`/`Pivot`/`Shuffle`'s non-array-input errors) even though `cursor` was already in scope — so the error text could contradict what `type` now correctly reports for the same node.
3. **`to_entries`, `reverse`, `pivot`, and `shuffle` materialized elements via bare `to_owned`/`collect_values`** instead of `to_owned_cursor`/`collect_cursors`, silently dropping tags on any element they touched.
4. **The `is*` builtin family** (`isnull`/`isboolean`/`isnumber`/`isstring`/`isarray`/`isobject`) called the untagged `DocumentValue` predicates directly — the same gap `type` had. Routing through `tagged_type_name` also incidentally fixed an unrelated pre-existing bug for these call sites: `is_number`/`is_string`'s default implementations could both answer `true` for the same untagged plain scalar.
5. **`to_owned_key_shape`** (computed index keys, slice bounds) took a bare value with no cursor — added a `to_owned_key_shape_cursor` sibling.
6. Plus a correctness-adjacent tightening: `tagged_type_name` was building and discarding a full `OwnedValue` (including a heap allocation) just to read its type name — now resolves straight to `ResolvedScalar::type_name()`.

All fixed as new commits, each independently verified with a live repro against `./target/debug/succinctly` before and after, plus new regression tests (CLI-level for everything reachable that way, one in-process unit test for `shuffle`'s array branch since `cargo-llvm-cov` wasn't attributing coverage back through the CLI subprocess for that specific arm). Full test suite, both clippy invocations, and 100% patch coverage (except one test-internal `panic!` assertion-failure arm, matching this file's existing test convention) all confirmed green after.

Filed #907 for the remaining lower-priority, unconfirmed-or-cosmetic findings from the same review round (an audit gap in the write-path comment-tree materializer, an audit gap in two `yq_runner.rs` call sites, and two pre-existing code-duplication concerns) rather than blocking merge on them.

Issue #730, opened 2026-08-09
Regex flags n/l are unimplemented, and unknown flag characters are silently accepted instead of erroring


newhoggy:
## Summary

`build_regex` in [`src/jq/eval.rs`](../blob/main/src/jq/eval.rs) parses the
flags string character-by-character:

```rust
for c in flags.chars() {
    match c {
        'i' => prefix.push('i'),
        'x' => prefix.push('x'),
        's' => prefix.push('s'),
        'm' => prefix.push('m'),
        'g' => {}                // global - handled at call site
        'p' => {}                // PCRE mode - not fully supported
        _ => {}
    }
}
```

Two gaps fall out of this:

1. **`n` (ignore/suppress empty matches) and `l` (POSIX leftmost-longest) are
   not implemented at all** — there is no special-case logic for either
   anywhere in the file. Both silently fall into the catch-all `_ => {}` and
   have zero effect.
2. **Any unrecognized flag character is silently accepted** rather than
   producing an error. Real jq validates the flags string and raises
   `<char> is not a valid modifier string` for anything it doesn't
   recognize.

## Repro (verified against pinned jq-1.7.1)

```console
$ echo '"abc"' | jq 'test("abc";"z")'
jq: error (at <stdin>:1): z is not a valid modifier string
$ echo '"abc"' | succinctly jq 'test("abc";"z")'
true    # should be a "not a valid modifier string" error
```

`n` and `l` don't currently have a clean, isolated repro: probing `n` via
`scan` (e.g. `scan(" *";"n")` vs `scan(" *")` on `"  "`) is confounded by a
*separate* bug where succinctly's global regex iteration (`find_iter`/
`captures_iter`) never enumerates a trailing zero-width match at all — with
or without `n` — so `n`'s "ignore empty matches" semantics can't be
distinguished from that iteration bug using this pattern. That trailing
zero-width-match gap looks related to (or possibly the same root cause as)
the zero-width case tracked in #729; worth checking there first, since
fixing it may make `n`/`l` easy to test cleanly.

## Suggested fix

- Add flags-string validation to `build_regex`: reject any character outside
  the recognized set (`g i x s m n l p` — see #727/#728/#729 for the
  semantics some of these still need fixing) with an error matching jq's
  wording.
- Implement `l` (leftmost-longest): Rust's `regex` crate doesn't support
  POSIX leftmost-longest semantics directly at the flag level; this may
  require `RegexBuilder::unicode`/a different match-selection strategy,
[...]

Issue #729, opened 2026-08-09
match(re;"g") and capture(re;"g") return a single value instead of streaming one output per match


newhoggy:
## Summary

In real jq, `match(re; "g")` is a generator: it emits one match object **per
match** as separate stream outputs (so `[match(re;"g")]` collects N objects,
and `match(re;"g") | .offset` runs once per match). Likewise, idiomatic jq
use of `capture` with `"g"`-like iteration (`foreach match(re;"g") as $m
...`, or `[capture(re;"g")]` in newer jq) is expected to yield one object per
match.

succinctly's `builtin_match` (`src/jq/eval.rs`) instead collects every match
into a single `OwnedValue::Array` and returns it as **one**
`QueryResult::Owned` value:

```rust
if global {
    let matches: Vec<OwnedValue> = re
        .find_iter(&input)
        .map(|m| build_match_object(&re, m.as_str(), m.start(), &input))
        .collect();
    QueryResult::Owned(OwnedValue::Array(matches))   // one value, not a stream
}
```

`builtin_capture_with_flags` has the same shape problem: it always returns
just the *first* match's object (`re.captures(&input)`), regardless of `g`.

## A second, related bug found in the same area

Even setting the streaming-vs-array shape aside, succinctly's global match
also fails to find a **trailing zero-width match** that oniguruma/jq's
global search does find. For a zero-width-capable pattern like `( )*`
against `"  "` (two spaces), jq's global search reports a full match
(`"  "`, offset 0) *and* an empty match at the end of the string (offset 2);
succinctly's `find_iter`-based loop only reports the first. This affects
`match` today and would affect any future streaming fix, so it's worth
diagnosing together with the shape fix rather than separately.

## Repro (verified against pinned jq-1.7.1)

```console
$ echo '"a1b22c333"' | jq -c 'match("[0-9]+";"g")'
{"offset":1,"length":1,"string":"1","captures":[]}
{"offset":3,"length":2,"string":"22","captures":[]}
{"offset":6,"length":3,"string":"333","captures":[]}
$ echo '"a1b22c333"' | succinctly jq -c 'match("[0-9]+";"g")'
[{"offset":1,...},{"offset":3,...},{"offset":6,...}]   # one array, not 3 lines
```

```console
$ echo '"  "' | jq -c '[match("( )*";"g")]'
[{"offset":0,"length":2,...},{"offset":2,"length":0,"string":"","captures":[...]}]
[...]

Issue #703, opened 2026-08-09
test(jq): regex flag combinations (g/i/x/s/m/n/l) have zero test coverage


newhoggy:
## Summary

`test`/`match`/`capture`/`scan`/`sub`/`gsub`/`split` all accept a regex-flags
argument (`Builtin::TestFlags`, `MatchFlags`, `CaptureFlags`, `SubFlags`,
`GsubFlags`, `ScanFlags`, `SplitRegex`, `SplitsFlags` in `src/jq/expr.rs`),
but **no test in the repo exercises any flag, alone or in combination.**

Confirmed by grep across every test file and `src/jq/eval.rs`'s
`#[cfg(feature = "regex")]` unit tests (`test_regex_test`, `test_regex_scan`,
`test_regex_splits`, `test_regex_sub`, `test_regex_gsub`, `test_regex_match`,
`test_regex_capture`) — all seven call the un-flagged 1-arg form only. The
190-probe `jq_error_message_tests.rs` corpus and all 338
`tests/data/jq-golden/cases` are also flag-free: zero filters anywhere match
`test(...; "..."`/`match(...;"...")`/etc.

## Why this matters

jq-1.7.1's own oniguruma test suite (`tests/onig.test`) treats flag handling
as one of its densest areas, and several of these are exactly the kind of
thing that breaks silently:

- `"g"` global matching against a zero-width-capable pattern (one match per
  position).
- `"gn"` (global + suppress-empty-matches) — must yield **zero** matches for
  a pattern that can only match empty, not the same result as bare `"g"`.
- `"gi"` case-insensitive matching against non-ASCII text (must NOT
  ASCII-fold accented characters).
- `"x"` (extended/free-spacing, ignores literal whitespace and `#` comments
  in the pattern).
- `"s"`/`"m"` (single-line / multi-line `.`/`^`/`$` semantics).
- `gsub`'s zero-width-match cluster: `gsub(""; "a"; "g")` on `""` vs on a
  1-char string, `gsub("^"; ...)`, `gsub("$"; ...)` — each has a distinct,
  easy-to-get-wrong expected insertion count.
- Lookahead assertions combined with `"g"`.

succinctly's flags are live-verified to work correctly today (spot-checked
`test("ABC"→"abc";"i")`, `match(...;"g")` multi-match, `test(...;"x")` with a
commented pattern) — but **nothing pins any of it**, so a refactor of the
flag-dispatch code (or an upgrade of the underlying regex crate) could break
[...]

PR #905, opened 2026-08-14
fix(jq): reject an unrecognized regex flag instead of ignoring it

newhoggy:
## Summary

`build_regex`'s flags loop silently accepted any character outside the recognized set (`g i x s m n l p`), falling into a no-op catch-all arm. Real jq validates the flags string and raises `<flags> is not a valid modifier string` for anything it doesn't recognize.

```
$ echo '"abc"' | jq 'test("abc";"z")'
jq: error (at <stdin>:1): z is not a valid modifier string
$ echo '"abc"' | succinctly jq 'test("abc";"z")'
true    # should be a "not a valid modifier string" error
```

Confirmed the error message names the **whole flags string**, not just the offending character, and is identical whether the bad character stands alone or sits next to a genuinely valid one:

```
$ echo '"abc"' | jq 'test("abc";"zq")'
jq: error (at <stdin>:1): zq is not a valid modifier string
$ echo '"abc"' | jq 'test("abc";"iz")'
jq: error (at <stdin>:1): iz is not a valid modifier string
```

## Scope note

This is a partial fix for #730, which bundles three sub-problems of very different complexity:

1. **Unknown-flag validation** — this PR. Clean, isolated, matches jq's exact wording.
2. **`l` (POSIX leftmost-longest)** — not implemented. Rust's `regex` crate doesn't support this directly; needs its own investigation into what's achievable.
3. **`n` (suppress empty matches)** — not implemented, and per #730's own text, depends on #729 (confirmed closed) landing cleanly-testable ground first.

`n`/`l`/`p` are still accepted as valid flag characters (matching real jq's grammar) — this PR only rejects characters real jq itself rejects, and doesn't attempt to implement `n`/`l`'s actual semantics. Left #730 open rather than closing it, since it isn't fully resolved; commented there with this scoping so the remaining `n`/`l` work is easy to pick up separately.

## Test plan

- [x] `cargo fmt --check`
- [x] Diffed output against pinned real `jq` for single/multiple/mixed invalid flags, and confirmed `n`/`l`/`p`/`i`/`g` still work as before
- [x] Un-pinned `regex_test_z_unknown_flag` from `tests/data/jq-golden-known-failures.txt` (filed alongside #728 from #703's regex-flag coverage sweep) — golden conformance now covers this case
- [x] Added `test_regex_unknown_flags_rejected_730` pinning the exact repro shapes plus `n`/`l`/`p` acceptance

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -7510,8 +7510,17 @@ fn builtin_scan_with_flags<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         _ => return QueryResult::Error(EvalError::cannot_be_matched(&to_owned(&value))),
     };
 
+    // `scan` always operates in jq's "global" mode (it finds every match, the
+    // way `gsub` always replaces every match) — jq's own bootstrap prepends
+    // `g` to the flags string before validation for exactly that reason.
+    // `g` is a no-op for `build_regex`'s own compiled-pattern behavior, so
+    // this only affects the wording of an invalid-flags error, matching
+    // jq's own message: confirmed live, `scan("a"; "z")` reports "gz is not
+    // a valid modifier string", not "z is not a valid modifier string".
+    let global_flags = format!("g{}", flags.unwrap_or(""));
+
     // Build regex
-    let re = match build_regex(&pattern, flags) {
+    let re = match build_regex(&pattern, Some(&global_flags)) {
         Ok(r) => r,
         Err(_e) if optional => return QueryResult::None,
         Err(e) => return e.into(),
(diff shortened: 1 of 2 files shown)


newhoggy:
## Review update

Ran \`/code-review\` (standard effort, multi-agent). One real bug found and fixed, confirmed live against real jq before and after:

**\`scan\`/\`split\`/\`splits\` reported the wrong error message for invalid flags.** These three builtins always operate in jq's "global" mode (finding/splitting on every match), and jq's own bootstrap prepends \`"g"\` to the flags string before validation for exactly that reason — \`gsub\` already did the equivalent (appending \`"g"\`), but my first pass left \`scan\`/\`split\`/\`splits\` passing the raw user flags straight to \`build_regex\`. Confirmed live: \`scan("a"; "z")\` should report \`"gz is not a valid modifier string"\`, not \`"z is not a valid modifier string"\`. This was previously silent (no error at all) before this PR's own fix, so there was no wrong text to notice before — now visible since these three builtins raise an error at all. Fixed by prepending \`"g"\` at all three call sites, matching \`gsub\`'s existing pattern (just prepend instead of append, matching jq's own observed ordering).

**Also found, filed separately**: bare \`splits(re)\` (no flags argument) returns a single array instead of streaming each substring the way \`splits(re; flags)\` and real jq both do — confirmed pre-existing on \`main\`, unrelated to this PR's own changes. Filed as **#906**.

Other review angles (efficiency, reuse, language-pitfalls, altitude, conventions) found only minor style observations — comment verbosity, a possible named-error-constructor opportunity matching an existing pre-existing gap in the same function, and a test/fixture-overlap note — none of which are bugs, so left as-is given this fix's narrow, already-well-scoped goal.

All commits pushed. Full test suite, both required clippy invocations, and \`cargo fmt --check\` all green after these changes.

Issue #880, opened 2026-08-14
jq: builtin_in swallows/misreports a Halt or Break from its xs argument


newhoggy:
## Summary

`builtin_in`'s (`in(xs)`, `src/jq/eval.rs` ~line 2992) match on `eval_single(obj_expr, ...)`'s result has no arm for `Halt`, `Break`, or their `Partial` variants — all fall into either `_ if optional => None` or the final `_ => Error("in() requires an object or array argument")`, misreporting a halt/break as a bogus type error (or silently swallowing it under `?`) instead of propagating it.

Found during #867's code review (which fixed the identical bug shape in `isvalid`/`isempty`) — same bug class, different builtin, not covered by that PR.

## Repro (verified against pinned jq-1.7.1)

```
$ echo null | jq -n -c '"a" | in(halt_error(3))'
a
$ echo null | succinctly jq -n -c '"a" | in(halt_error(3))'
jq: error (at <unknown>): in() requires an object or array argument
```
(real jq prints `xs`'s own single already-in-flight side effect - the raw `a` written by `halt_error(3)`'s message-printing - then exits 3; succinctly misreports a type error, exit 5)

```
$ echo null | jq -n -c 'label $out | ("a" | in(break $out)), "after"'
(no output, exit 0)
$ echo null | succinctly jq -n -c 'label $out | ("a" | in(break $out)), "after"'
jq: error (at <unknown>): in() requires an object or array argument
```

## Why this needs its own investigation, separate from the mechanical isvalid/isempty fix

Unlike `isvalid`/`isempty` (which evaluate their argument as a single logical value and just needed the missing match arms added), `in(xs)` genuinely **fans out** over multiple `xs` outputs at the real-jq level:

```
$ jq -n -c '"a" | [in({a:1}, {a:9})]'
[true,true]
```

[...]

PR #908, opened 2026-08-14
fix(jq): builtin_in fans out over every xs output and propagates Halt/Break

newhoggy:
## Summary

- `in(xs)` = `. as \$x | xs | has(\$x)` (confirmed against jq 1.7.1) — a plain pipe, so it forks once per `xs` output (`"a" | [in({a:1}, {a:9})]` is `[true,true]`), not just the first. `builtin_in`'s own `eval_single` call only ever consulted the first output (`os.into_iter().next()`) and had no match arm for `Halt`, `Break`, or their `Partial` variants — a halt/break from `xs` fell into either the `optional` guard or a bogus "in() requires an object or array argument" type error instead of propagating.
- Switches to `eval_owned_multi_keep_partial` and checks each candidate in order, matching real jq's generator truncation: a later candidate's escape (whether from `xs` itself or a per-candidate type mismatch) keeps whatever booleans were already produced by earlier candidates before surfacing — the same "keep partial, defer the escape" shape #842/#854 established for `recurse`.
- `?` suppresses only a genuine error, never a `Break` or `Halt`, matching `EvalEscape`'s own documented contract.

```
$ jq -n -c '"a" | in(halt_error(3))'
a
$ succinctly jq -n -c '"a" | in(halt_error(3))'
jq: error (at <unknown>): in() requires an object or array argument
```
(real jq writes `a`, halt_error's own message, to stderr, then exits 3; succinctly misreported a type error, exit 5)

```
$ jq -n -c '"a" | [in({a:1}, {a:9})]'
[true,true]
$ succinctly jq -n -c '"a" | [in({a:1}, {a:9})]'
[true]
```
(pre-fix: only the first `xs` output was ever consulted)

## Note on coverage

4 lines (the `Some(EvalEscape::Error(_)) if optional => ...` branch) are genuinely unreachable via any real jq/yq syntax and left uncovered. `?` on a whole expression desugars to `eval_try` (external Partial/Error catching), not to forcing this function's own `optional: bool` parameter true (confirmed empirically and via the `#693` doc comment on `Expr::Optional`'s dispatch, which deliberately stopped forcing `optional=true` down a subtree). The sibling `builtin_has` has the identical pre-existing, equally-uncovered pattern (`_ if optional => QueryResult::None`), confirming this isn't new to this PR — I verified via the raw lcov data rather than assuming.

## Test plan

- [x] `cargo fmt --check`
- [x] Manually diffed output against real jq 1.7.1 across ~12 scenarios: original repros, fan-out (basic/mixed), error mid-stream with/without `?`, error-first with `?`, empty `xs`, break/halt after partial success, array bounds — all match exactly (streams captured separately, not merged, since `halt_error` writes to stderr)
- [x] Added unit tests (`src/jq/eval.rs`) and CLI tests (`tests/jq_cli_tests.rs`, `tests/yq_cli_tests.rs` for the yq-only negative-index branch), all pinned against real jq/yq
- [x] Patch coverage: 96% (48/50), remaining 4 lines documented above as genuinely untestable

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -2969,10 +2969,13 @@ fn builtin_has<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
             let len = (*elements).count() as i64;
             let in_bounds = if S::NEGATIVE_INDEX_IN_HAS {
                 // yq behavior: negative indices are valid if abs(idx) <= len
+                // -- `unsigned_abs`, not `abs`, since `i64::MIN.abs()`
+                // overflows (panics in debug, silently wraps back to a
+                // still-negative i64::MIN in release, #908 review).
                 if *idx >= 0 {
                     *idx < len
                 } else {
-                    idx.abs() <= len
+                    idx.unsigned_abs() <= len as u64
                 }
             } else {
                 // jq behavior: only non-negative indices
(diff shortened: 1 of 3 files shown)


newhoggy:
Review found real bugs plus real duplication — all addressed:

**Fixed** (new commits):
- Null \`xs\` candidate errored instead of contributing \`false\` (missing arm, unlike \`builtin_has\`'s own null short-circuit) — worse, it also truncated a multi-candidate stream partway through.
- \`idx.abs()\` overflow on \`i64::MIN\` in yq mode's negative-index check — panics in debug, silently wraps to a wrong answer in release. Fixed in both \`builtin_in\` and the identical duplicated code in \`builtin_has\`.
- The trailing escape-combining match hand-rolled what the existing \`finish_fork\`/\`owned_vec_to_result\`/\`partial\` helpers already do (multiple reviewers independently found this, one verified via live substitution) — replaced with a single \`finish_fork\` call.

**Filed as follow-ups** (out of scope for this PR):
- #909 — \`has()\`/\`in()\` error on a float/NaN array key instead of truncating (real jq behavior); pre-existing in \`has()\`, not introduced here.
- #910 — \`IN(s)\`/\`IN(src;s)\` (the builtins immediately adjacent to \`in()\`) eagerly evaluate their whole generator instead of short-circuiting on first match like real jq's \`any\`; same bug family as #880 but a different builtin, needs its own investigation since \`any\`'s short-circuit semantics differ from \`in(xs)\`'s keep-every-output fan-out.
- Also noted but not filed separately: the deeper "eager evaluation of \`xs\` can leak a side effect real jq's lazy pipe would never reach" gap (e.g. \`in(5, halt_error(3))?\` leaks \`halt_error\`'s stderr message even though real jq never gets there) — this predates #908 (same eager \`eval_single\` in the old code) and would need real laziness work to fix properly, not a quick patch.

All fixes verified against real jq/yq 1.7.1, full test suite green, both clippy invocations clean, patch coverage 100% on the new commits.

PR #911, opened 2026-08-14
fix(jq): checked arithmetic for mktime/strftime/tz timestamp math

### LINKED ISSUE #893: jq: mktime/strftime("%s") panics on overflow in unix_secs_from_broken_down_time
**Severity: Medium**

## Summary

`unix_secs_from_broken_down_time` (`src/jq/eval.rs`), the inverse of
`broken_down_time_from_unix_secs`, used by `mktime` and by `strftime`'s `%s`
specifier (#760), does unchecked `era * 146097` arithmetic on a
user-controlled `year` and panics instead of erroring when that overflows.

This is the mirror-image bug of #869 (fixed in #889), on the opposite
conversion direction — #869 was scoped to the seconds-to-broken-down-time
direction (`gmtime`/`localtime`/`strftime`'s raw-number path/`todate`/`tz`);
this function was not one of the sites #869 named, and #889 did not touch it.

## Repro

```
$ echo null | succinctly jq '[9223372036854775807,0,1,0,0,0,0,0] | mktime'
thread 'main' panicked at src/jq/eval.rs:16837:16: attempt to multiply with overflow

$ echo null | succinctly jq '[9223372036854775807,0,1,0,0,0,0,0] | strftime("%s")'
thread 'main' panicked at src/jq/eval.rs:16837:16: attempt to multiply with overflow
```

No `overflow-checks` override exists in `Cargo.toml`, so a release build would
instead silently return a wrong integer rather than panicking or erroring —
the same silent-corruption failure mode #869 was filed to eliminate on the
other conversion direction.

## Root cause

`unix_secs_from_broken_down_time` (`src/jq/eval.rs`, near line 16796):

```rust
let era = if y >= 0 { y } else { y - 399 } / 400;
let yoe = (y - era * 400) as u32;
...
let days = era * 146097 + doe as i64 - 719468;
```

`era * 400` and `era * 146097` are unchecked and overflow for a sufficiently
large `year` array element.

## Why this is separate from #869 / #889

#869/#889 fixed the seconds→broken-down-time direction
(`broken_down_time_from_unix_secs`, shared by `gmtime`/`localtime`/`todate`/`tz`/
`strftime`'s raw-number input path) using `checked_sub` plus an `i32`
year-range check. `unix_secs_from_broken_down_time` is the *inverse*
direction, used by `mktime`/`strftime("%s")`, with its own independent
arithmetic — not fixed by that PR, and not named in #869's own "Duplication
context" list of sites sharing the vulnerable pattern.

## Suggested fix direction

Apply the same treatment: `checked_mul`/`checked_add` through the `era`
arithmetic, returning a clean error (matching real jq's own behavior/message
if it has one for this direction — verify against the pinned jq-1.7.1 oracle)
instead of panicking or wrapping.

### LINKED ISSUE #894: jq: localtime panics on an extreme TZ env var via parse_simple_tz_offset overflow
**Severity: Low**

## Summary

`parse_simple_tz_offset` (`src/jq/eval.rs`, ~line 16703), reached from
`builtin_localtime` via `estimate_local_offset`, does an unchecked
`hours * 3600` multiply on an integer parsed straight out of the `TZ`
environment variable, and panics if that overflows — reachable *before*
#889's new `checked_add` overflow guard in `builtin_localtime` ever runs, so
that guard gives a false sense of safety against this specific path.

## Repro

```
$ TZ="EST9999999999999999" succinctly jq -n 'localtime'
thread 'main' panicked at src/jq/eval.rs:16703: attempt to multiply with overflow
```

Note the input timestamp here is not extreme at all (`0`, via `-n`'s `null`
input) — the panic comes purely from the malformed `TZ`-derived offset, not
from the timestamp.

## Root cause

```rust
let offset_secs = (hours * 3600 + minutes * 60) * if negative { 1 } else { -1 };
```

`hours` is parsed directly from `TZ` via `.parse()` with no bound, and
`hours * 3600` overflows `i64` for a sufficiently long digit string in `TZ`.

## Why this is separate from #869 / #889

#869/#889 fixed the *timestamp* magnitude causing this class of panic in
`gmtime`/`localtime`/`strftime`/`todate`/`tz`. This is a distinct trigger
(malformed environment input, not a filter argument) in a different function
entirely, pre-existing on `main` before and after that fix.

## Suggested fix direction

Use `checked_mul`/`checked_add` in `parse_simple_tz_offset`, returning `None`
(falling back to the existing UTC/0 fallback path, consistent with how this
function already returns `None` for other unparseable `TZ` formats) instead
of panicking.

### LINKED ISSUE #895: jq: tz()'s DST heuristics use truncating division, giving wrong pre-1970 results
**Severity: Low**

## Summary

`is_dst_us_eastern`/`is_dst_europe`/`is_dst_australia` (`src/jq/eval.rs`,
~lines 17823/17849/17872), called from `builtin_tz`'s `get_timezone_offset`
for named abbreviation zones (`EST`/`CET`/etc.), each inline their own copy
of the civil-calendar day/month derivation using truncating `secs / 86400`
instead of the floor-corrected division (`checked_sub(86399) / 86400`) that
`broken_down_time_from_unix_secs` uses after #889.

This is the exact kind of hand-duplicated conversion math #869 was filed
about, in three more places not named in that issue's own "Duplication
context" list (which named `builtin_localtime`, `builtin_todate`, and
`format_datetime_with_offset` — all fixed in #889 — but not these three DST
helpers).

## Impact

For any negative (pre-1970) timestamp, `secs / 86400` truncates toward zero
instead of flooring (e.g. `secs = -1` gives `days = 0` instead of the correct
`-1`), feeding the wrong month/day into the DST match arms. Near a DST
transition boundary on a pre-1970 date (e.g. the 2nd Sunday of March or last
Sunday of October, in `America/New_York`'s or `CET`'s heuristic), this
silently selects the wrong standard/daylight offset with no error raised.

## Why this is separate from #869 / #889

#869/#889 fixed the primary Unix-timestamp-to-broken-down-time conversion
(`broken_down_time_from_unix_secs`) and migrated `localtime`/`todate`/`tz`'s
own formatting step onto it. These three DST-heuristic functions are a
*different* call path (feeding `get_timezone_offset`, upstream of
`format_datetime_with_offset`) with their own independent, still-unmigrated
copy of the day/month math — not touched by that PR.

## Suggested fix direction

These functions already note in their own comments that they're "a
simplified subset" of real IANA timezone rules — full correctness may not be
worth chasing here. At minimum, though, migrating their day/month derivation
onto the shared, floor-corrected `broken_down_time_from_unix_secs` (or
extracting just the day/month portion into something these three and it all
share) would fix the pre-1970 truncation bug without expanding scope beyond
what #869/#889 already centralized.

newhoggy:
## Summary

Three sibling gaps left by #869/#889's fix, all in the same `src/jq/eval.rs` timestamp/TZ subsystem — bundled together per the same shared root cause (unchecked/truncating arithmetic on user-controlled datetime components), each cross-referenced by its own issue against #869/#889 as the direct precedent:

- **#893** — `unix_secs_from_broken_down_time` (backing `mktime` and `strftime`'s `%s` specifier) did unchecked `era`/day/hour/minute/second arithmetic on a user-controlled broken-down-time array element and panicked on overflow — the mirror-image bug of #869 on the opposite (broken-down-time → seconds) conversion direction. Verified by hand that *every* field (year, month, day, hour, minute, second) independently reaches this panic, not just year as the issue's own repro showed. Now returns `Result<i64, EvalError>`, propagated through `builtin_mktime`, `builtin_strftime`'s array branch (whose own eager `month + 1` step had the identical bug, reachable independently of `%s`), and `format_strftime`'s `%s` arm.
- **#894** — `parse_simple_tz_offset` did an unchecked `hours * 3600`/`minutes * 60` multiply on a value parsed straight out of the `TZ` env var and panicked on a long enough digit string — reachable *before* #889's own overflow guard in `builtin_localtime` ever runs, since it's a different trigger (malformed environment input, not the timestamp argument). Now checked throughout, falling back to the function's existing `None` → UTC path.
- **#895** — `is_dst_us_eastern`/`is_dst_europe`/`is_dst_australia` each hand-duplicated the pre-#889 truncating `secs / 86400` instead of the shared, floor-corrected `broken_down_time_from_unix_secs`, giving the wrong civil day — and thus the wrong DST answer — for any pre-1970 timestamp not exactly on a day boundary. Now derive month/day from that shared function instead of re-deriving the civil date inline, fixing the truncation bug and the "duplicated 4x" code smell #869's own title flagged in one step.

## Test plan

- [x] `cargo test --verbose` (no `cli` feature — catches feature-gated test assumptions)
- [x] Verified every overflow panic reproduced by hand pre-fix (year/month/day/hour/minute/second for `mktime`/`strftime("%s")`; malformed `TZ` for `localtime`), then confirmed each now errors cleanly instead
- [x] Verified the #895 DST boundary fix against real `gmtime` output at exact day boundaries (both pre- and post-fix behavior traced by hand and confirmed against the pinned oracle's own civil-date math)
- [x] New regression tests mirroring #869's own precedent style (`query!` macro, same error-message constant) for all three fixes
- [x] 97.81% patch coverage (`omni-dev coverage diff`); the 3 remaining uncovered lines are a pre-existing `Err(_)/None if optional => QueryResult::None` idiom already present elsewhere in the same functions, empirically confirmed unreachable via `?` syntax (the outer `Expr::Optional` wrapper does the None-conversion instead, not this builtin's own `optional` parameter) — kept for consistency with the surrounding code rather than removed

Fixes #893, Fixes #894, Fixes #895

### DIFF
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -736,6 +736,24 @@ impl EvalError {
         Self::new("error converting number of seconds since epoch to datetime")
     }
 
+    /// `mktime/strftime: broken-down time value out of representable
+    /// range` — a `mktime`/`strftime` broken-down-time array element
+    /// (year, month, day, hour, minute, second, weekday, or yearday) too
+    /// extreme for this codebase's checked civil-date arithmetic to
+    /// convert without overflow (#893, #911). Deliberately a distinct
+    /// message from [`Self::datetime_out_of_range`] rather than reusing
+    /// it: that message is anchored to real jq's own text for the
+    /// *opposite* (seconds -> broken-down-time) direction, and real jq has
+    /// no equivalent error for *this* direction at all — confirmed
+    /// empirically it silently computes a wrapped/nonsensical result
+    /// instead (`[9223372036854775807,0,1,0,0,0,0,0] | mktime` succeeds in
+    /// real jq with a bogus timestamp) — so there is no oracle text to
+    /// match here, and reusing the other message would misdescribe which
+    /// conversion actually failed.
+    pub fn broken_down_time_out_of_range() -> Self {
+        Self::new("mktime/strftime: broken-down time value out of representable range")
+    }
+
     /// `mktime requires array inputs`.
     pub fn mktime_requires_array() -> Self {
         Self::new("mktime requires array inputs")
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -16977,7 +16977,7 @@ fn builtin_mktime<W: Clone + AsRef<[u64]>>(
     };
     let month = match get_int(1).and_then(|m| {
         m.checked_add(1)
-            .ok_or_else(EvalError::datetime_out_of_range)
+            .ok_or_else(EvalError::broken_down_time_out_of_range)
     }) {
         Ok(m) => m, // jq uses 0-indexed months
         Err(_) if optional => return QueryResult::None,
@@ -17037,7 +17037,7 @@ fn unix_secs_from_broken_down_time(
     minute: i64,
     second: i64,
 ) -> Result<i64, EvalError> {
-    let overflow = EvalError::datetime_out_of_range;
+    let overflow = EvalError::broken_down_time_out_of_range;
 
     let y = year
         .checked_sub(i64::from(month <= 2))
@@ -17165,7 +17165,7 @@ fn builtin_strftime<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
                 let month = match get_int(1).checked_add(1) {
                     Some(m) => m,
                     None if optional => return QueryResult::None,
-                    None => return QueryResult::Error(EvalError::datetime_out_of_range()),
+                    None => return QueryResult::Error(EvalError::broken_down_time_out_of_range()),
                 };
 
                 (
@@ -17207,6 +17207,7 @@ fn format_strftime(
     weekday: i64,
     yearday: i64,
 ) -> Result<String, EvalError> {
+    let overflow = EvalError::broken_down_time_out_of_range;
     let mut result = String::new();
     let mut chars = fmt.chars().peekable();
 
@@ -17234,7 +17235,10 @@ fn format_strftime(
                 Some('S') => result.push_str(&format!("{second:02}")),
                 Some('p') => result.push_str(if hour < 12 { "AM" } else { "PM" }),
                 Some('P') => result.push_str(if hour < 12 { "am" } else { "pm" }),
-                Some('j') => result.push_str(&format!("{:03}", yearday + 1)), // 1-indexed
+                Some('j') => result.push_str(&format!(
+                    "{:03}",
+                    yearday.checked_add(1).ok_or_else(overflow)? // 1-indexed
+                )),
                 Some('w') => result.push_str(&format!("{weekday}")),
                 Some('u') => {
                     result.push_str(&format!("{}", if weekday == 0 { 7 } else { weekday }));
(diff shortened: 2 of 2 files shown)


newhoggy:
Ran the mandatory multi-angle review before merge. It found — and every angle independently, empirically confirmed via live reproduction against the built binary — a real gap: the initial fix hardened \`unix_secs_from_broken_down_time\` and the two \`month + 1\` sites, but \`format_strftime\`'s \`%j\`/\`%U\`/\`%W\`/\`%V\`/\`%G\`/\`%g\` specifiers each do their own independent, unchecked arithmetic on the same user-controlled \`weekday\`/\`yearday\`/\`year\` array fields — panicking just the same, in the very function this PR already touches. \`iso_week_date\`/\`weeks_in_iso_year\` (backing \`%V\`/\`%G\`/\`%g\`) needed the identical checked-arithmetic treatment.

Fixed as a new commit, with all 8 previously-panicking repros re-verified as now erroring cleanly, and byte-for-byte output confirmed unchanged against the pinned real \`jq\` oracle for ordinary (non-adversarial) inputs. Also gave the mktime/strftime overflow direction its own error message (\`EvalError::broken_down_time_out_of_range\`) rather than reusing \`datetime_out_of_range\`, whose text is anchored to real jq's own wording for the *opposite* direction — flagged by several review angles as misleading, and confirmed real jq has no equivalent error for this direction at all (it silently computes a wrapped/nonsensical result instead), so there was no oracle text being matched by the reuse.

New regression tests cover all the newly-fixed panic sites plus a byte-for-byte oracle-matched normal-input case. Full test suite (including the no-\`cli\`-feature legs), both clippy invocations, and 98.56% patch coverage all confirmed green after.

Filed #912 for the remaining lower-priority findings from the same review round (civil-date-math duplicated in \`parse_strptime\`/\`parse_iso8601\` — not currently exploitable, digit-count-bounded — plus a few style/consistency nits) rather than blocking merge on them.

PR #913, opened 2026-08-14
fix(jq): bare splits(re) streams one output per substring

### LINKED ISSUE #906: jq: bare splits(re) returns one array instead of streaming, unlike splits(re; flags)
**Severity: Medium**

## Summary

`splits(re)` (bare, no flags argument) returns a single array value instead of streaming each substring as a separate output the way `splits(re; flags)` and real jq's `splits(re)` both do.

## Repro

```
$ echo '"a1b2"' | jq -c '[splits("[0-9]")]'
["a","b",""]
$ echo '"a1b2"' | succinctly jq -c '[splits("[0-9]")]'
[["a","b",""]]

$ echo '"a1b2"' | jq -c 'splits("[0-9]")'
"a"
"b"
""
$ echo '"a1b2"' | succinctly jq -c 'splits("[0-9]")'
["a","b",""]
```

`splits(re; flags)` (with an explicit, even empty, flags argument) is correct:

```
$ echo '"a1b2"' | succinctly jq -c '[splits("[0-9]"; "")]'
["a","b",""]
```

## Root cause

`builtin_splits` (`src/jq/eval.rs:7077`, dispatched from `Builtin::Splits(re)` for the bare `splits(re)` form) returns `QueryResult::Owned(OwnedValue::Array(parts))` — a single value. Its sibling `builtin_splits_with_flags` (dispatched for `splits(re; flags)`) correctly returns `QueryResult::ManyOwned(parts)` — one output per substring, matching real jq's `def splits($re): splits($re; null);` definition (which should route the bare form through the *same* streaming logic, just with `null` flags).

## Repro history

Found while fixing #730 (regex flag validation) — confirmed identical on `main` before that fix, not introduced or affected by it.

## Suggested fix direction

Either have `builtin_splits` delegate to `builtin_splits_with_flags` with `flags: None` (matching how `builtin_gsub_with_flags` and others delegate to their `_with_flags` sibling), or fix its own return to `QueryResult::ManyOwned(parts)` directly.

newhoggy:
## Summary

`builtin_splits` (the bare `splits(re)` form, no flags argument) had its own standalone implementation that collected every substring into a single `QueryResult::Owned(Array(...))` instead of streaming, unlike its `splits(re; flags)` sibling (`builtin_splits_with_flags`), which correctly returns `QueryResult::ManyOwned` — one output per substring, matching real jq's `def splits($re): splits($re; null);` definition.

```
$ echo '"a1b2"' | jq -c '[splits("[0-9]")]'
["a","b",""]
$ echo '"a1b2"' | succinctly jq -c '[splits("[0-9]")]'
[["a","b",""]]
```

Delegate to `builtin_splits_with_flags` with `flags: None` instead of duplicating the split logic — matching the same pattern `builtin_sub` already uses for its own bare/flagged pair (`sub(re;str)` delegates to `sub_with_flags(...,None,...)`).

Found while fixing #730 (regex flag validation) — confirmed identical on `main` before that fix, unrelated to it.

**Update:** a second round of review on this PR's own diff surfaced a matching bug in `builtin_scan` (bare `scan(re)`), which ignored capture groups entirely (`caps.get(0)` regardless of `capture_count`). Filed as #915 and fixed here the same way — delegate to `builtin_scan_with_flags(..., None, ...)`. #915 also notes `test(re)` has the same duplication risk (unconfirmed as a live bug) and stays open for that.

Fixes #906

## Test plan

- [x] `cargo fmt --check`
- [x] Diffed output against pinned real `jq` for `[splits(re)]`, bare streamed `splits(re)`, `splits(re; flags)` (unaffected), and `?`-optional behavior
- [x] Diffed output against pinned real `jq` for `[scan(re)]` with and without capture groups, confirming array-of-groups vs bare-string dispatch matches `scan(re; flags)`
- [x] Fixed the two existing unit tests (`test_regex_splits`, and added `test_regex_scan_bare_form_honors_capture_groups_915`) that were pinning the wrong behavior / missing coverage
- [x] Rebased onto latest `main` post-#911 merge; build/test/clippy re-verified clean after rebase

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -27392,6 +27392,50 @@ mod tests {
 
     #[cfg(feature = "regex")]
     #[test]
+    fn test_regex_scan_bare_form_honors_capture_groups_915() {
+        // #915: bare `scan(re)` used to duplicate `scan(re; flags)`'s logic
+        // instead of delegating, and the duplicate never checked
+        // `capture_count` — so a pattern with capture groups returned the
+        // whole match (group 0) instead of the array of captured groups.
+        // Verified live against jq 1.7.1: `scan("(a)(b)")` on `"ab"` is
+        // `["a","b"]`, matching what `scan(re; "")` already produced here.
+        query!(br#""ab""#, r#"scan("(a)(b)")"#,
+            QueryResult::ManyOwned(matches) => {
+                assert_eq!(matches.len(), 1);
+                assert_eq!(
+                    matches[0],
+                    OwnedValue::Array(vec![
+                        OwnedValue::String("a".to_string()),
+                        OwnedValue::String("b".to_string()),
+                    ])
+                );
+            }
+        );
+
+        // Two matches of a capture-grouped pattern.
+        query!(br#""a1b2""#, r#"scan("([a-z])([0-9])")"#,
+            QueryResult::ManyOwned(matches) => {
+                assert_eq!(matches.len(), 2);
+                assert_eq!(
+                    matches[0],
+                    OwnedValue::Array(vec![
+                        OwnedValue::String("a".to_string()),
+                        OwnedValue::String("1".to_string()),
+                    ])
+                );
+                assert_eq!(
+                    matches[1],
+                    OwnedValue::Array(vec![
+                        OwnedValue::String("b".to_string()),
+                        OwnedValue::String("2".to_string()),
+                    ])
+                );
+            }
+        );
+    }
+
+    #[cfg(feature = "regex")]
+    #[test]
     fn test_regex_splits() {
         // `splits(re)` streams one output per substring — like real jq, and
         // like `splits(re; flags)` already did before #906's fix made the
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -7123,46 +7123,22 @@ fn builtin_scan<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
 }
 
 /// Builtin: splits(re) - split by regex
+///
+/// jq defines this as `splits($re; null)` — delegate to the flags-aware
+/// implementation directly rather than duplicating its (streaming) split
+/// logic here. The former standalone body returned a single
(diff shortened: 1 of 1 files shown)


newhoggy:
## Review update: second round of `/code-review` found a real capture-group bug in `builtin_scan`

The first round of multi-agent review (10 angles) on the original splits-only diff came back clean on 6 angles; the **altitude** and **reuse/duplication** angles both independently flagged that `builtin_scan` (bare `scan(re)`, no flags) has the same standalone-implementation smell that `builtin_splits` just had — and it turned out to hide a real bug, not just duplication:

```
$ echo '"a1b2"' | jq -c '[scan("([a-z])([0-9])")]'
[["a","1"],["b","2"]]
$ echo '"a1b2"' | succinctly jq -c '[scan("([a-z])([0-9])")]'
["a1","b2"]
```

`builtin_scan`'s standalone implementation always took `caps.get(0)` (the whole match) regardless of `capture_count`, ignoring the capture groups entirely — while `builtin_scan_with_flags` (the `scan(re; flags)` sibling) already had the correct capture-group handling. Verified live against real jq that the correct behavior is capture groups → array of groups, no captures → bare string, matching `scan(re; flags)`'s existing logic exactly.

Fixed the same way as the `splits` fix in this PR: delegate `builtin_scan` to `builtin_scan_with_flags(re_expr, None, value, optional)` instead of duplicating (and diverging from) the logic. Filed as #915, which also notes `test(re)`'s bare/flagged pair carries the same duplication risk — not confirmed as a live bug there, so left as a documented open item rather than fixed speculatively here. This PR closes #915's `scan` half only; #915 stays open for the `test` follow-up.

Commits: `726fa6baa` (splits fix, unchanged), `d4d3c040f` + `056e8f1ae` (scan fix + test).

Rebased onto latest `main` (picked up #911's mktime/strftime overflow-panic fixes, unrelated area, clean rebase, no conflicts). Build/test/clippy (both invocations) all green post-rebase.

PR #914, opened 2026-08-14
fix(jq): has()/in() truncate a float array key instead of erroring

### LINKED ISSUE #909: jq: has()/in() error on a float or NaN array key instead of truncating (or false for NaN), like real jq
**Severity: Low**

## Summary

`builtin_has`'s and `builtin_in`'s array-index match arm only matches `OwnedValue::Int`/`NumberRepr::Int`, never `Float`/`NaN`. Real jq truncates a non-integer numeric key toward zero before bounds-checking, and treats NaN as simply out of range (`false`). Both builtins in succinctly instead raise a bogus "Cannot check whether array has a number key" error.

Found during #908's review (a rewrite of `builtin_in` for issue #880); confirmed the identical gap already exists in the untouched, pre-existing `builtin_has`.

```
$ jq -n -c '1.5 | in([1,2,3])'
true
$ succinctly jq -n -c '1.5 | in([1,2,3])'
jq: error (at <unknown>): Cannot check whether array has a number key

$ echo '[1,2,3]' | jq -c 'has(1.5)'
true
$ echo '[1,2,3]' | succinctly jq -c 'has(1.5)'
jq: error (at <stdin>:1): Cannot check whether array has a number key

$ jq -n -c 'nan | in([1,2,3])'
false
$ succinctly jq -n -c 'nan | in([1,2,3])'
jq: error (at <unknown>): Cannot check whether array has a number key
```

## Root cause

`src/jq/eval.rs`: both `builtin_has`'s array-index arm (~line 2966) and `builtin_in`'s (~line 3026) match only `OwnedValue::Int(idx) | OwnedValue::NumberLiteral(NumberRepr::Int(idx), _)` against `OwnedValue::Array(elements)`. A `Float`/`NumberLiteral(Float(_), _)` key falls through to the generic type-mismatch catch-all instead of being truncated toward zero (matching jq's actual C-level index-coercion behavior) or, for NaN specifically, short-circuited to `false` (jq's `#530`-adjacent NaN-as-no-element-address convention, already documented for a different function — see `key_to_path_component`'s own NaN handling doc comment in this file).

## Why this is separate from #880/#908

That issue/PR fixed `in(xs)`'s fan-out and Halt/Break propagation; it touched this exact match arm but this specific float/NaN gap already existed in `builtin_has` beforehand and wasn't introduced or widened by that fix — it's a pre-existing divergence this issue documents separately.

## Suggested fix direction

Add a shared numeric-key-to-index coercion (truncate `Float`/`NumberLiteral(Float(_), _)` toward zero via `as i64` after a NaN check, matching this file's existing `numeric_key_to_index`/`key_to_path_component` NaN-handling precedent) to both `builtin_has`'s and `builtin_in`'s array-index arms. Given both functions already duplicate their array-bounds logic verbatim (a separate finding from #908's review, not yet filed as its own issue), this may be worth doing as part of a shared `check_has`/`index_in_bounds`-style helper rather than patching each copy independently.

newhoggy:
## Summary

- Both `builtin_has`'s and `builtin_in`'s array-index arm matched only `OwnedValue::Int`/`NumberLiteral(Int)`, so a `Float` or `NaN` key fell through to the generic type-mismatch error instead of matching real jq's own index coercion: truncate a `Float` key toward zero, or — for NaN specifically — simply answer `false` (never in bounds).
- Widens both arms' key pattern to accept any numeric `OwnedValue` variant and routes the actual coercion through the existing `numeric_key_to_index` helper (already used by `index_one` for the equivalent `.[key]` case) — no new truncation/NaN logic needed, just reuse of an already-correct helper.

```
$ jq -n -c '1.5 | in([1,2,3])'
true
$ succinctly jq -n -c '1.5 | in([1,2,3])'
jq: error (at <unknown>): Cannot check whether array has a number key

$ echo '[1,2,3]' | jq -c 'has(1.5)'
true
$ echo '[1,2,3]' | succinctly jq -c 'has(1.5)'
jq: error (at <stdin>:1): Cannot check whether array has a number key

$ jq -n -c 'nan | in([1,2,3])'
false
$ succinctly jq -n -c 'nan | in([1,2,3])'
jq: error (at <unknown>): Cannot check whether array has a number key
```

Fixes #909

## Test plan

- [x] `cargo fmt --check`
- [x] Manually diffed output against real jq/yq 1.7.1: float truncation (in-bounds/out-of-bounds), NaN, negative float index in both jq and yq modes, negative float index bounds-checking (yq-only), plus a sanity check that plain Int keys are unaffected — all match exactly
- [x] Added unit tests (`src/jq/eval.rs`) and a yq CLI test (`tests/yq_cli_tests.rs`), all pinned against real jq/yq 1.7.1

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -2963,33 +2963,28 @@ fn builtin_has<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         }
         // Array has index - jq returns false for negative, yq returns true if
         // in range. Any numeric key type is accepted here, not just Int --
-        // `numeric_key_to_index` truncates a Float toward zero (matching
-        // jq's own `.[1.5]` coercion) and reports `None` for NaN, which is
-        // never in bounds (#909 review: this arm used to match only Int,
-        // erroring on a Float/NaN key instead of truncating/answering
-        // `false` like real jq's `has(1.5)`/`has(nan)` do).
+        // `numeric_key_to_array_index` truncates a Float toward zero in jq
+        // mode (matching jq's own `.[1.5]` coercion) and reports `None` for
+        // NaN, which is never in bounds. yq mode never accepts a Float key
+        // at all (that helper's own doc comment has the real-yq-verified
+        // rationale) -- #909 review: this arm used to match only Int,
+        // erroring on a Float/NaN key in *either* mode instead of matching
+        // each mode's own real behavior.
         (
             StandardJson::Array(elements),
             OwnedValue::Int(_) | OwnedValue::Float(_) | OwnedValue::NumberLiteral(..),
         ) => {
-            let len = (*elements).count() as i64;
-            let in_bounds = match numeric_key_to_index(&key_owned) {
+            let in_bounds = match numeric_key_to_array_index::<S>(&key_owned) {
+                // NaN: a number, so the container check above still
+                // applies, but no element -- never in bounds. `elements`
+                // is a lazy O(n) cursor walk (`StandardJson::Array`, #909
+                // review), so this short-circuits before paying for it,
+                // matching the `Some` arm's own `len` computation being
+                // needed at all.
                 None => false,
                 Some(idx) => {
-                    if S::NEGATIVE_INDEX_IN_HAS {
-                        // yq behavior: negative indices are valid if abs(idx) <= len
-                        // -- `unsigned_abs`, not `abs`, since `i64::MIN.abs()`
-                        // overflows (panics in debug, silently wraps back to a
-                        // still-negative i64::MIN in release, #908 review).
-                        if idx >= 0 {
-                            idx < len
-                        } else {
-                            idx.unsigned_abs() <= len as u64
-                        }
-                    } else {
-                        // jq behavior: only non-negative indices
-                        idx >= 0 && idx < len
-                    }
+                    let len = (*elements).count() as i64;
+                    index_in_array_bounds::<S>(idx, len)
                 }
             };
             QueryResult::Owned(OwnedValue::Bool(in_bounds))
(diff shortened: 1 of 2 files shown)


newhoggy:
Review found real, significant findings — all addressed:

**Fixed** (new commits):
- Duplication: the new `None`/`Some(idx)` escape-and-bounds block was hand-duplicated between `builtin_has`/`builtin_in` (on top of the `S::NEGATIVE_INDEX_IN_HAS` branch already duplicated since #908) — extracted into a shared `index_in_array_bounds` helper (multiple reviewers independently converged on this, one verified via live substitution).
- Efficiency: `builtin_has`'s array-index arm walked the whole (lazy, O(n)) array via `.count()` unconditionally, even for a NaN key whose answer (`false`) never needed it — reordered so the count only runs once a real index exists to check.
- **A real regression from my own fix**: applying jq's float-truncation uniformly to yq mode too was wrong. Independently verified live against the pinned yq v4.53.3: real yq's `has(2.0)` on `[1,2,3]` is `false`, not `true` — it never accepts a Float key for array indexing at all, regardless of value. Added `numeric_key_to_array_index`, which applies each semantics' own real policy (jq truncates; yq requires a genuine integer key) instead of forcing jq's contract onto both.
- Minor: hoisted `numeric_key_to_array_index` out of `builtin_in`'s per-candidate loop (the key is invariant across candidates).
- One of my own new tests asserted the *wrong* (pre-verification) yq behavior — fixed to match the real, verified answer.

**Filed as follow-ups** (out of scope for #909, all pre-existing, none introduced by this fix):
- #917 — yq mode's `has()`/`in()` diverge from real yq on three fronts, all independently verified live: (a) the negative-index bounds check (`abs(idx) <= len`) should just be unconditionally `true` for any negative integer; (b) `has()` never errors on a type-mismatched key in real yq, always answers `false`; (c) `in(xs)` as a jq-style function call doesn't even parse in real yq — every "confirmed against real yq" claim on succinctly's yq-mode `in()` tests (across #880/#908/#909) has no real oracle to check against, only the desugared `has($x)` form does.
- #918 — an integer-valued YAML float scalar (`2.0`) silently loses its float type during input parsing, becoming indistinguishable from the integer `2` (confirmed: `2.0 == 2` is `true` in succinctly yq mode, `false` in real yq). This is what caused my own float-key test to initially fail for the wrong reason — traced to this separate, upstream YAML-parsing issue, not a bug in this PR's fix.

All fixes verified against real jq/yq (pinned v4.53.3), full test suite green, both clippy invocations clean.

PR #916, opened 2026-08-14
fix(jq): keep already-produced generator prefix at 4 more path-position sites

### LINKED ISSUE #896: jq: select/if/computed-index/getpath drop an already-produced generator prefix on a later error, like #842/#854 did for recurse
**Severity: Low**

## Summary

The same "drops an already-produced generator prefix on a later error" bug class #842 and #854 fixed for `recurse`'s `f` and `cond` loops is independently live at 4 more call sites in `resolve_node`/`src/jq/eval.rs`, all using `eval_owned_multi` (all-or-nothing) where the caller's own argument can be a multi-output generator that produces a value before erroring on a later output.

Found and independently re-verified against real jq 1.7.1 while finishing #854's review punch list (originally surfaced by a prior session that ran out of time before filing it — see the "Status" comment on #854).

1. `select`'s own argument (`resolve_node`'s `Select` arm, `src/jq/eval.rs` ~9881-9900 area):
```
$ echo '1' | jq -c 'path(select((true, error("x"))))'
jq: error (at <stdin>:1): x
[]
$ echo '1' | succinctly jq -c 'path(select((true, error("x"))))'
jq: error (at <stdin>:1): x
```

2. `if`'s cond arm (`resolve_node`'s `If` arm):
```
$ echo '1' | jq -c 'path(if (true, error("x")) then . else empty end)'
jq: error (at <stdin>:1): x
[]
$ echo '1' | succinctly jq -c 'path(if (true, error("x")) then . else empty end)'
jq: error (at <stdin>:1): x
```

3. Computed-key index (`resolve_index_expr`'s computed-key eval):
```
$ echo '{"a":1,"b":2}' | jq -c 'path(.[("a", error("x"))])'
jq: error (at <stdin>:1): x
["a"]
$ echo '{"a":1,"b":2}' | succinctly jq -c 'path(.[("a", error("x"))])'
jq: error (at <stdin>:1): x
```

4. `getpath`'s argument (`GetPath`'s arg eval):
```
$ echo '{"a":1}' | jq -c 'path(getpath((["a"], error("x"))))'
jq: error (at <stdin>:1): x
["a"]
$ echo '{"a":1}' | succinctly jq -c 'path(getpath((["a"], error("x"))))'
jq: error (at <stdin>:1): x
```

In every case, real jq prints the value(s) already produced by the earlier output(s) of the multi-output generator, *then* raises the error; succinctly prints only the error, silently dropping the already-produced prefix.

## Root cause

Value-position code already has a general "keep every value, defer the escape" primitive (`push_owned_values`, backing `eval_binary_fanout` and — as of #854 — `eval_owned_multi_keep_partial`, built on `QueryResult::Partial`). Path position's `PathResolveResult` has no type-level equivalent: each call site has to hand-thread a prefix correctly through its own `Err`, and most of `resolve_node`'s call sites just use plain `eval_owned_multi` (all-or-nothing) instead, which is right for most callers (see that function's own doc comment) but wrong for these 4, whose argument the caller doesn't otherwise expect to be a generator that can partially succeed.

## Why this is separate from #842/#854

Those two issues (and their fixes) are scoped specifically to `recurse`'s own `f`/`cond` loops. This is the identical bug shape, independently occurring at 4 unrelated call sites in `resolve_node` that also happen to evaluate a caller-supplied sub-expression as a `Vec`-collecting multi-output generator.

## Suggested fix direction

Two options, worth deciding explicitly rather than defaulting to the first one found:

1. **Piecemeal**: fix each of the 4 sites individually, the same way #842/#854 did for `recurse` (switch to `eval_owned_multi_keep_partial` or an equivalent keep-partial call, thread the prefix through the site's own `Err` path).
2. **Architectural**: add a `PathResolveResult`-side `eval_fanout`-style primitive (mirroring `eval_owned_multi_keep_partial`'s value-position role) that every path-position multi-output-argument call site can use, closing this bug class categorically instead of one site at a time — matching how #627/#636/#842/#854 have each incrementally chased a variant of the same underlying gap.

Given `resolve_node` likely has other call sites not enumerated here that could share the same shape, option 2 may be worth the investment; left as a decision for whoever picks this up.

## Why this is Low severity

Requires a caller-supplied argument expression to itself be a multi-output generator whose later output errors after an earlier one already succeeded — narrow, and every confirmed repro needs `path(...)` to observe (the value-position equivalents already correctly propagate the error without needing to preserve a value prefix, since there's no separate "value" vs "path" split there for these specific arms).

newhoggy:
## Summary

The same "drops an already-produced generator prefix on a later error" bug class #842/#854 fixed for `recurse`'s own `f`/`cond` loops was independently live at 4 more call sites in `resolve_node`/`resolve_index_expr` (`src/jq/eval.rs`), all using the all-or-nothing `eval_owned_multi` where the caller-supplied argument can itself be a multi-output generator that produces a value before erroring on a later output:

- `select`'s own condition
- `if`'s `cond` arm
- a computed index key (`.[EXPR]`)
- `getpath`'s argument

Each now uses `eval_owned_multi_keep_partial` and threads the trailing escape through `PathResolveResult`'s own `Err` alongside whatever branches were already produced, instead of discarding them entirely — mirroring #842/#854's precedent and `resolve_leaf`'s (#861) existing use of the same primitive.

Went with the piecemeal fix direction (the issue's own "option 1") rather than the architectural `PathResolveResult`-side primitive (`option 2`) — lower risk and narrower blast radius given these 4 sites are structurally simple, and matches the precedent #842/#854 already established for the identical bug shape.

## Test plan

- [x] All 4 repros verified byte-for-byte against the pinned real `jq` (1.7.1) oracle, both stdout and exit code
- [x] Verified the bare-escape case (zero prior generator output) still propagates unchanged at every site — the partial-keeping fix must not regress this
- [x] Verified ordinary (non-erroring) multi-output cases at each site are unaffected, also checked against the oracle
- [x] New regression tests (`tests/jq_cli_tests.rs`) for all 4 sites plus the bare-escape guard
- [x] 100% patch coverage (`omni-dev coverage diff`)

Fixes #896

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -10519,19 +10519,32 @@ fn resolve_node<'a, S: EvalSemantics>(
         // unchanged or prune it, exactly like `Optional` above but driven by a
         // predicate instead of an error.
         //
-        // `cond` runs through `eval_owned_multi`, not `eval_owned_expr`, and
-        // republishes `value` once per truthy output rather than collapsing
-        // every output into one always-truthy array (#628, mirroring #627's
+        // `cond` runs through `eval_owned_multi_keep_partial`, not
+        // `eval_owned_multi`/`eval_owned_expr`, and republishes `value` once
+        // per truthy output rather than collapsing every output into one
+        // always-truthy array (#628, mirroring #627's
         // `builtin_recurse_cond`/`resolve_recurse` fix): confirmed against jq
         // 1.7.1, `path(select((false,false)))` is empty (not `[[]]`), and
         // `path(select((true,true)))` forks into two branches, `[[],[]]`.
+        // Keeping the partial prefix (rather than discarding it on a later
+        // error/break) matches jq's backtracking generator semantics: a
+        // later output of `cond` failing does not retroactively un-emit a
+        // branch an earlier truthy output already produced (#842/#854's
+        // precedent for `recurse`, independently live here too — #896).
+        // Confirmed against jq 1.7.1: `path(select((true, error("x"))))`
+        // prints `[]` (the branch for the already-produced `true`) before
+        // raising `x`.
         Expr::Builtin(Builtin::Select(cond)) => {
-            let cond_outputs = eval_owned_multi::<S>(cond, value).map_err(|e| (Vec::new(), e))?;
-            Ok(cond_outputs
+            let (cond_outputs, escape) = eval_owned_multi_keep_partial::<S>(cond, value);
+            let branches: Vec<PathBranch<'a>> = cond_outputs
                 .into_iter()
                 .filter(OwnedValue::is_truthy)
                 .map(|_| (Vec::new(), Cow::Borrowed(value)))
-                .collect())
+                .collect();
+            match escape {
+                Some(e) => Err((branches, e)),
+                None => Ok(branches),
+            }
         }
         Expr::Builtin(
             builtin @ (Builtin::Values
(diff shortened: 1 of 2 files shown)


newhoggy:
Ran the mandatory multi-angle review before merge. It found — and independently, empirically confirmed across multiple angles via live reproduction against the built binary and the pinned real \`jq\` (1.7.1) oracle — 2 real bugs beyond the initial fix:

1. **\`resolve_index_expr\`'s target-error handling discarded a non-empty prefix.** \`target_branches = resolve_node::<S>(target, value, trackable)?\` used the \`?\` operator to bail immediately on any \`target\` error — harmless before this PR, since none of \`Select\`/\`If\`/\`GetPath\`/computed-index could ever return a non-empty prefix on \`Err\`. Now that they can (this PR's own fix), that prefix needs indexing by \`keys\` like any other target branch instead of being silently returned unindexed. Confirmed: \`path(select((true, error("t")))[("x", error("k"))])\` on \`{"a":1,"x":9}\` printed \`[]\` instead of \`["x"]\`. jq's own \`K as $k | E | .[$k]\` compilation (cited in this function's own doc comment) exhausts \`E\`'s whole generator, escape included, before ever asking \`K\` for its next value — so \`target\`'s escape now correctly takes priority over \`key\`'s.

2. **\`GetPath\`'s multi-output fallback fabricated a wrong error message.** When \`arg\` produces 2+ outputs before erroring, the shape doesn't match the fast-path's exact single-array check, so it fell through to \`resolve_leaf\`, which re-evaluates \`arg\` from scratch via a *different*, single-shot code path that silently discards the escape it rediscovers. Confirmed: \`path(getpath((["a"],["b"],error("x"))))\` on \`{"a":1,"b":2}\` reported \`"Invalid path expression with result 1"\` instead of the real error \`x\` — strictly worse than the pre-#896 behavior (which at least surfaced \`x\` correctly, just without the fan-out). Fixed by propagating the real escape directly instead of falling through for this specific shape — not a further improvement for this shape's own fan-out (that needs \`resolve_leaf\`'s own multi-output handling fixed separately, tracked in the follow-up below), just not a regression from it either.

Also worth noting: every angle of the review that investigated a third candidate — a downstream error in a branch/index step outranking an *earlier-captured* \`halt_error\` at 3 of the sites — independently confirmed this is **not a bug**. I verified it directly: real jq itself exits the same way (ordinary error wins, not the halt) for all 3 repros, because jq's lazy generator model never demands the halt-producing output once an earlier stage already threw. The pre-#896 code's opposite behavior (halt always winning) was itself diverging from jq in this exact race; this PR's change happens to fix that divergence as a side effect, not introduce one.

Both fixes verified byte-for-byte against the real jq oracle, with new regression tests. Full test suite, both clippy invocations, and 100% patch coverage all confirmed green after.

Filed #962 for the remaining lower-priority findings (an eager-vs-lazy interaction between \`first()\`/\`limit()\` and every keep-partial resolver including \`recurse\`'s pre-existing #842/#854 fix — pre-existing, not newly introduced, but worth its own scoped fix; plus some escape-handling duplication and doc-comment staleness) rather than blocking merge on them.

Issue #920, opened 2026-08-14
jq: regex l flag (POSIX leftmost-longest) needs an architecture decision, not implementable with current regex engine


newhoggy:
**Severity: Low**

## Summary

The regex `l` flag (POSIX leftmost-longest matching) is one of two remaining gaps from #730 — the other, `n` (suppress empty matches), is fixed by the PR that files this issue. `l` is split out separately because it needs an architectural decision, not a mechanical fix.

## What `l` does (verified against pinned jq 1.7.1)

```
$ echo '"aaa"' | jq -c 'match("a|aa|aaa") | .string'
"a"
$ echo '"aaa"' | jq -c 'match("a|aa|aaa";"l") | .string'
"aaa"
```

Without `l`, jq's oniguruma backend uses Perl-style leftmost-first matching (first alternative that matches wins — `"a"`). With `l`, it switches to POSIX leftmost-longest (longest match at the leftmost starting position wins — `"aaa"`). This isn't limited to top-level alternation; POSIX semantics apply recursively to every subexpression (nested groups, quantifiers), which is what makes this hard to approximate.

## Why this isn't a quick fix

Checked the vendored `regex-automata` 0.4.16 source directly (`src/util/search.rs`, `MatchKind` enum). It has exactly two variants:

```rust
pub enum MatchKind {
    /// Report all possible matches.
    All,
    /// Report only the leftmost matches. When multiple leftmost matches exist,
    /// report the match corresponding to the part of the regex that appears
    /// first in the syntax.
    #[default]
    LeftmostFirst,
    // There is prior art in RE2 that shows that we should be able to add
    // LeftmostLongest too. ... it's not clear whether it's really worth
    // adding this. [comment continues, explaining why literal-sorting
    // tricks don't generalize past simple cases]
}
```

`LeftmostLongest` doesn't exist anywhere in the crate — it's a documented "we could add this, not clear it's worth it" comment from the `regex-automata` maintainers themselves, never implemented. This isn't a succinctly gap in using the crate; the semantics genuinely aren't available at any layer of the `regex`/`regex-automata` stack succinctly depends on.

## Options (needs a decision, not obviously one right answer)

1. **Swap regex engines** for something with native POSIX support (e.g. bindings to a POSIX-compliant engine). Large blast radius — `regex` is used throughout `src/jq/eval.rs`'s regex builtins; would need evaluating alternatives, licensing, `no_std`/portability fit, and whether it's worth carrying two regex dependencies just for one flag.
[...]

PR #921, opened 2026-08-14
fix(jq): regex n flag suppresses empty matches

### LINKED ISSUE #730: Regex flags n/l are unimplemented, and unknown flag characters are silently accepted instead of erroring

newhoggy:
## Summary

The regex `n` flag (suppress empty matches) was accepted as a valid flag character but had zero functional effect. Real jq filters zero-length matches out of the global-match iteration, and for the first-match-only builtins (`match`/`test`/`capture`/`sub` without `"g"`), searches forward past a leading empty match to find the first non-empty one instead of settling for it.

```
$ echo '"a  b   c"' | jq -c '[scan(" *")]'
["","  ","","   ","",""]
$ echo '"a  b   c"' | jq -c '[scan(" *";"n")]'
["  ","   "]

$ echo '"aaxaa"' | jq -c 'match("x*";"n") | .offset'
2
```

Tractable now that #729 unified every global-match consumer (`match(;g)`, `capture(;g)`, `scan`, `gsub`, `split`, `splits`) behind one `global_captures` helper — filtering zero-length matches there (guarded by a new `JqRegex::suppress_empty` flag, set from `n`) covers all of them in one place. Added a parallel `first_captures` helper for the non-global case, which stops at the first qualifying match instead of scanning the whole input the way `global_captures` does — `JqRegex::captures`/`is_match` now route through it, so `match`/`test`/`capture`/`sub` (without `"g"`) pick it up automatically.

Every case verified live against pinned jq 1.7.1, including the no-non-empty-match-anywhere case, where jq's generators produce **no output at all** (not `null`), `test` is `false`, and `sub` leaves its input unchanged.

`l` (POSIX leftmost-longest) remains unimplemented — split into #920, since implementing it needs an architecture decision (checked the vendored `regex-automata` 0.4.16 source directly: `MatchKind::LeftmostLongest` doesn't exist in the crate at all) rather than a mechanical fix. With this PR, #730 is otherwise fully addressed (unknown-flag validation via #905, `n` here).

**Update:** review found a real gap in the same family — the skip-forward search only finds a non-empty match reachable by moving to a *later* position, not one reachable only by backtracking to a different alternative at the *same* position (lazy quantifiers, alternations with an empty branch listed first). Same root cause as `l`'s gap (no `ONIG_OPTION_FIND_NOT_EMPTY` equivalent in Rust's non-backtracking `regex` crate). Filed as #922 and pinned with a test; doesn't block this PR since every documented/tested `n` use case (greedy quantifiers) is correctly handled. Also deduped `first_captures`/`global_captures`'s near-identical loop bodies into a shared `next_match_step` helper, and restored a fast path for the common non-`n` case that a first pass had accidentally routed through the more general (but slower) path.

Fixes #730

## Test plan

- [x] `cargo fmt --check`
- [x] Diffed output against pinned real `jq` for `n` on `scan`/`gsub`/`split`/`splits`/`match(;g)` (global) and `match`/`test`/`capture`/`sub` (non-global, including no-non-empty-match-anywhere and `n` composed with `i`)
- [x] Added `test_regex_n_flag_suppresses_empty_matches_730` covering all of the above

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -6576,21 +6576,27 @@ fn dv_captures_at<'h>(
 /// A compiled jq regex plus whatever extra state its builtins need to
 /// replicate jq semantics that Rust's `regex` crate doesn't natively
 /// support (currently: jq's Perl-style trailing-newline exception for `$`
-/// — see `dv_captures_at`/`dv_is_match`).
+/// — see `dv_captures_at`/`dv_is_match` — and the `n` flag's suppression of
+/// zero-length matches — see `first_captures`/`global_captures`).
 #[cfg(feature = "regex")]
 struct JqRegex {
     re: regex::Regex,
     needs_trailing_nl_check: bool,
+    suppress_empty: bool,
 }
 
 #[cfg(feature = "regex")]
 impl JqRegex {
     fn is_match(&self, input: &str) -> bool {
-        dv_is_match(&self.re, input, self.needs_trailing_nl_check)
+        if self.suppress_empty {
+            first_captures(self, input).is_some()
+        } else {
+            dv_is_match(&self.re, input, self.needs_trailing_nl_check)
+        }
     }
 
     fn captures<'h>(&self, input: &'h str) -> Option<regex::Captures<'h>> {
-        dv_captures_at(&self.re, input, 0, self.needs_trailing_nl_check)
+        first_captures(self, input)
     }
 
     fn capture_names(&self) -> regex::CaptureNames<'_> {
@@ -6608,6 +6614,7 @@ fn build_regex(pattern: &str, flags: Option<&str>) -> Result<JqRegex, EvalError>
     let mut pattern = pattern.to_string();
 
     // Apply flags
+    let mut suppress_empty = false;
     if let Some(flags) = flags {
         let mut case_insensitive = false;
         let mut extended = false;
@@ -6628,12 +6635,13 @@ fn build_regex(pattern: &str, flags: Option<&str>) -> Result<JqRegex, EvalError>
                 // `p` enables both s and m; s is a no-op here, so p reduces
                 // to the same effect as m.
                 'p' => dot_matches_newline = true,
-                'g' => {} // global - handled at call site
+                'g' => {}                     // global - handled at call site
+                'n' => suppress_empty = true, // suppress empty matches (#730)
                 // Recognized by real jq but not yet implemented here (#730):
-                // `n` (suppress empty matches) and `l` (POSIX leftmost-longest)
-                // are accepted as valid flag characters, silently without
-                // effect, rather than raising the error below.
-                'n' | 'l' => {}
+                // `l` (POSIX leftmost-longest) is accepted as a valid flag
+                // character, silently without effect, rather than raising
+                // the error below — see #920 for why this one stays open.
+                'l' => {}
                 _ => {
                     // Real jq's own wording, and its own choice: the message
                     // names the whole flags string, not just the offending
@@ -6676,6 +6684,7 @@ fn build_regex(pattern: &str, flags: Option<&str>) -> Result<JqRegex, EvalError>
     Ok(JqRegex {
         re,
         needs_trailing_nl_check,
+        suppress_empty,
     })
 }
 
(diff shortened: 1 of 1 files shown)


newhoggy:
## Review update: `/code-review` (10 angles) found a real correctness bug

**Angle A** (line-by-line diff scan) fuzz-tested the branch against pinned jq 1.7.1 across ~16,000 generated `(string, pattern, flags, builtin)` combinations and found a real, previously-undetected bug — my own hand-picked test cases only used greedy quantifiers, which all happen to work correctly.

**Root cause**: `first_captures`/`global_captures` treat "the leftmost-first match at this position is empty" as proof no non-empty match is achievable there, and unconditionally skip forward. That's true for greedy quantifiers (the common, documented `n` use case — filtering zero-width separator matches from patterns like ` *`, `[0-9]*`) but false for **lazy quantifiers** (`a*?`) and **alternations with an empty-matching branch listed first** (`(?:|a)`), where a non-empty match is achievable at the *same* position via backtracking to a different alternative — which the skip-forward search never attempts.

Verified live, e.g. on `"x\nabb"`: `test("a*?";"n")` → jq `true`, this branch (before the fix below) `false` — an actual regression from `main`'s pre-existing (accidental, since `n` was previously a total no-op) behavior for this specific input.

I independently reproduced every case Angle A reported before acting on it, then confirmed via the vendored `regex-automata`/`regex` source (grepped for `FIND_NOT_EMPTY`/`NotEmpty`) that this is the same class of gap as #920 (`l` flag): oniguruma's backtracking-based `ONIG_OPTION_FIND_NOT_EMPTY` has no equivalent in Rust's non-backtracking `regex` crate. Filed as **#922**, with the same "engine limitation, not a quick fix" framing as #920. Added `test_regex_n_flag_known_gap_lazy_quantifier_922` pinning the current known-divergent behavior (not jq's), so this stays visible rather than silently wrong.

**Angles Reuse, Simplification, and Altitude** independently converged on the same finding: `first_captures` and `global_captures` had near-duplicate loop bodies (match extraction, the acceptance test, the advance-past-empty-match arithmetic) — the exact "two copies of one predicate diverge silently" pattern this codebase's own notes call out for #106, and the same shape #906/#915 hit in unrelated builtins. Extracted `next_match_step` as the one place that logic lives now; both functions are just that step driven in a loop.

**Angle Efficiency** found `JqRegex::captures()` routed through `first_captures` unconditionally, even for the common non-`n` case where the loop always returns on its first iteration anyway (~5.8% measured overhead for no behavioral difference). Restored the direct fast path, mirroring `is_match`'s existing dispatch. Two more Efficiency findings (redundant `caps.get(0)` recomputation by callers, and using the allocating `captures_at` instead of `find_at` for rejected trial positions in the skip loop) are real but pre-existing-in-spirit / minor-constant-factor — confirmed the skip loop is O(n) overall (not O(n²), per empirical scaling tests from both Angle B and Angle Efficiency), so I've left those as noted-but-not-actioned rather than risk more churn on top of the correctness fix.

**Angles B, C, D, and Conventions** came back clean — no removed-behavior regressions, no cross-file bypass, no panics/overflow/UTF-8 boundary issues, no CLAUDE.md violations. Angle B specifically disproved a self-raised O(n²) hypothesis with empirical timing (linear scaling to 4M chars).

Commits: `2aa6270f4`/`8c1e3363c` (original fix+test), `bc46af39e` (dedup + fast path + gap documentation), `a9dcb369e` (gap-pinning test).

Issue #917, opened 2026-08-14
yq: has()/in() semantics diverge from real yq on negative-index bounds, type-mismatch errors, and in()'s own syntax


newhoggy:
**Severity: Medium**

## Summary

While fixing #909 (a narrower float/NaN key issue), review surfaced three separate, pre-existing divergences between succinctly's yq-mode `has()`/`in()` and real yq (mikefarah/yq), all independently verified live against the pinned oracle version (`v4.53.3`, matching `tests/data/yq-golden/YQ_VERSION`). None of these are introduced by #909's own diff — they predate it (from whenever `S::NEGATIVE_INDEX_IN_HAS`/`in()` support was first added, likely #880/#908 or earlier) — but #909's fix touches the exact same code and didn't have the scope to also correct them.

## Repro (verified against pinned yq v4.53.3)

**1. Negative-index bounds check is unconditionally `true`, not `abs(idx) <= len`:**
```
$ printf 'a: [1,2,3]\n' | yq '.a | has(-1)'
true
$ printf 'a: [1,2,3]\n' | yq '.a | has(-1000000)'
true
$ printf 'a: [1,2,3]\n' | yq '.a | has(-9223372036854775808)'
true
$ printf 'a: []\n' | yq '.a | has(-1)'
true
```
Every negative integer tested (from `-1` through `i64::MIN`, including on an *empty* array) returns `true` — the magnitude doesn't matter at all. succinctly's `index_in_array_bounds` (`src/jq/eval.rs`) implements `abs(idx) <= len` instead (documented in this repo's own CLAUDE.md as "yq behavior: negative indices are valid if abs(idx) <= len" — that documented rule itself appears to be wrong, not just the implementation).

**2. `has()` never errors on a type-mismatched key — always `false`, not `cannot_check_has`:**
```
$ printf 'a: [1,2,3]\n' | yq '.a | has("x")'
false
$ printf 'a: {b: 1}\n' | yq '.a | has(0)'
false
$ printf 'a: 5\n' | yq '.a | has(0)'
false
$ printf 'a: 5\n' | yq '.a | has("x")'
false
```
succinctly's `builtin_has` raises `EvalError::cannot_check_has` for exactly these shapes (a string key on an array, a numeric key on an object, `has()` on a scalar at all) — real yq is fully permissive and just answers `false`.

**3. `in(xs)` as a jq-style function-call filter does not exist in real yq's grammar at all:**
```
[...]

PR #924, opened 2026-08-15
fix(jq): isvalid evaluates its argument unforced, no longer masks a comma sibling's error

### LINKED ISSUE #881: jq: isvalid forces optional=true across its whole argument, masking a later comma-sibling's genuine error
## Summary

`builtin_isvalid` (`src/jq/eval.rs` ~line 13395) forces `optional=true`
for its *entire* argument subtree via a single `eval_single::<W,
S>(expr, value, true)` call. This is necessary to avoid a spurious
`Partial` result from a genuinely-erroring single-value argument (see
the existing test `test_isvalid_reaches_every_builtins_own_optional_guard`
and its comment), but it has an unintended, undocumented, and untested
side effect: when `expr` is a **multi-output comma expression** and a
*later* sibling errors, that error is silently swallowed - `isvalid`
answers `true` instead of `false` - because `optional=true` degrades
the later sibling's genuine error to `QueryResult::None` (e.g. via
`eval_error`/`builtin_explode`'s own `optional`-guard), and
`eval_comma` treats a `None` sibling as a pure no-op, leaving whatever
output the earlier sibling(s) already produced untouched.

Found and independently re-confirmed twice during #867's code review
(which fixed `isvalid`'s separate, unrelated `Break`-swallowing bug -
this is a different root cause, not fixed by that PR).

## Repro

```
$ echo null | succinctly jq -c 'isvalid((1|explode))'      # single value, correctly false
false
$ echo null | succinctly jq -c 'isvalid(1, (1|explode))'   # comma sibling, wrongly true
true
$ echo null | succinctly jq -c 'isvalid(1, error("boom"))' # wrongly true
true
$ echo null | succinctly jq -c 'isvalid(1, (1/0))'          # wrongly true
true
```

This contradicts `isvalid`'s own doc comment ("check if expr succeeds
without errors").

## Root cause

- `eval_error` and `builtin_explode` (both `src/jq/eval.rs`) degrade a
  genuine error to `QueryResult::None` when `optional` is true.
- `eval_comma` (`src/jq/eval.rs` ~line 786) treats a `None` sibling's
  result as `QueryResult::None => {}` - a no-op that neither
  contributes to nor fails the overall comma evaluation.
- `builtin_isvalid` forces `optional=true` unconditionally for the
  whole argument, so *every* nested optional-aware error site within
  it (not just the top-level expression) gets this treatment - not
  just the single-value case the existing test/comment covers.

## Why this is separate from #867

#867 (PR #879) fixed `isvalid`/`isempty` swallowing an unresolved
`break $label` from their argument - a *different* mechanism (missing
match arms for `QueryResult::Break`/`Partial(_, Control::Break(_))`),
unrelated to how `optional=true` propagates through comma-sibling
error handling. Neither PR's fix touches this bug.

## Suggested fix direction

Needs investigation: `optional=true` can't simply be dropped (that's
what the existing `test_isvalid_reaches_every_builtins_own_optional_guard`
test guards against - it would turn a normal `Partial` result from a
plain erroring single-value argument into an outright process-level
error escape instead of `isvalid` answering `false`). The real fix is
likely narrower: `isvalid` needs to distinguish "the expression
overall degraded to a `Partial`/`None` because a top-level error was
suppressed by the `optional` guard" (should still answer `false`, the
correctly-working case today) from "a comma sibling's error was
suppressed by an `optional` guard applied too broadly, several levels
down" (should also answer `false`, but currently doesn't because the
suppression already erased the evidence by the time `isvalid`'s own
match runs). This may need a way to evaluate the argument that
distinguishes "produced values, no errors" from "produced values, but
also had an error the optional-forcing swallowed" - which the current
`QueryResult`/`optional`-guard machinery doesn't expose. No existing
test in the codebase covers `isvalid(<success>, <error>)`.

Found during #867's code review (PR #879).

newhoggy:
## Summary

`isvalid(EXPR)` used to force `optional=true` down the whole subtree, which made every nested error-suppression site (not just the top-level one) self-suppress to `None` before `eval_comma`'s own match ever saw an `Error` — so a later comma sibling's genuine error was silently swallowed whenever an earlier sibling had already produced real output: `isvalid(1, error("x"))` wrongly answered `true`.

- Fixed by evaluating with the argument's own ambient (unforced) `optional`, matching every other builtin's post-#693 model, and catching the `Partial(_, Control::Error(_))` shape a fork construct (comma/reduce/foreach) naturally produces when an unforced error interrupts it after some output.
- Real jq has no `isvalid` builtin to diff against directly, but its own "check if expr succeeds without errors" doc comment matches the standard community desugaring `def isvalid(f): try (f|true) catch false;` — checked live against jq 1.8.2 (`/opt/homebrew/Cellar/jq/1.8.2`), which also confirms two existing tests had pinned succinctly-only divergences the old forcing accidentally introduced (`@format`'s own optional-dependent empty-string success, and `paths_filter`'s node-error swallowing): both are corrected too, not just the comma-sibling bug this issue was filed for.
- All other existing `isvalid` CLI/unit tests (Halt/Break propagation, `sub` replacement type errors, yq `has()`/`in()` type-mismatch-never-errors from #917) verified unaffected.

Fixes #881

## Test plan
- [x] `cargo fmt --check`
- [x] Manually re-verified the fix and all corrected/unaffected cases against real jq 1.8.2 and 1.7.1

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -13720,14 +13720,44 @@ fn walk_impl<S: EvalSemantics>(
 }
 
 /// Builtin: isvalid(expr) - check if expr succeeds without errors.
+/// #881 fix: `expr` is evaluated with its own *ambient* `optional` (always
+/// `false` here, matching #693's "ambient, not broadcast" model every other
+/// builtin follows) rather than the old forced `optional=true` broadcast
+/// down the whole subtree. The old forcing existed to make a genuinely
+/// error-producing leaf convert into a `None`/`Owned` shape isvalid's match
+/// could recognize, but it had the unintended side effect of making
+/// `eval_error` (and every other builtin's own `_ if optional => ...` arm)
+/// self-suppress *every* nested error the same way -- including a comma
+/// sibling's error several levels down, which `eval_comma` then silently
+/// no-op'd, discarding the evidence that anything went wrong while an
+/// earlier sibling's real output survived (`isvalid(1, error("x"))` wrongly
+/// answered `true`). Evaluating unforced instead lets a genuine error
+/// surface as `eval_comma`/`eval_reduce`/`eval_foreach`'s own natural
+/// `Partial(_, Control::Error(_))` shape (built by `partial(...)`, same as
+/// every other "keep the prefix, thread the escape" site in this file),
+/// which the new arm below catches directly -- no per-builtin
+/// optional-forcing needed. See
+/// `test_isvalid_comma_sibling_error_not_masked_881` and
+/// `test_isvalid_reaches_every_builtins_own_optional_guard` for the full
+/// before/after case-by-case verification, including two cases
+/// (`@format`'s own `_ if optional => Ok(String::new())` arm, and
+/// `paths_filter`'s node-error swallowing) where the old forcing had
+/// produced a succinctly-only divergence from real jq's own reference
+/// `isvalid` desugaring that this fix incidentally corrects too.
 fn builtin_isvalid<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     expr: &Expr,
     value: StandardJson<'a, W>,
     _optional: bool,
 ) -> QueryResult<'a, W> {
-    match eval_single::<W, S>(expr, value, true) {
+    match eval_single::<W, S>(expr, value, false) {
         QueryResult::Error(_) => QueryResult::Owned(OwnedValue::Bool(false)),
         QueryResult::None => QueryResult::Owned(OwnedValue::Bool(false)),
+        // A fork construct (comma/reduce/foreach) that produced some output
+        // before a later, unforced error interrupted it -- the shape
+        // `eval_comma` et al. build via `partial(...)`, now reachable here
+        // since evaluation is no longer forced-optional (see this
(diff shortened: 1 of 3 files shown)


newhoggy:
Coverage note: the bot's \"indirect coverage changes\" section flags ~12 pre-existing lines (scattered across unrelated builtins, e.g. lines 554, 2943, 7151, 8911, 12480/12484, 12775/12797, 20322) that lost coverage as a side effect of this fix. All are `_ if optional => QueryResult::None`-shaped guards in builtins this PR doesn't otherwise touch — their only prior coverage came from being incidentally reached through `isvalid`'s old, over-broad forced `optional=true` broadcast (the exact mechanism this PR removes), not from any dedicated test of their own `?`-postfix behavior. Closing all of them would mean auditing each builtin's own direct `?` test coverage individually — a much larger, separate scope than this issue. I added direct tests for the specific cases traceable to this diff (the six `@format` builtins, `paths_filter`'s node-error swallowing, and `isvalid`'s own now-relocated `None` arm), all called out in the commits/tests themselves; the rest is pre-existing latent coverage debt, not a regression introduced by this change.

PR #925, opened 2026-08-15
fix(jq): bare test(re) delegates to test(re; flags) instead of duplicating

### LINKED ISSUE #923: jq: builtin_test_regex duplicates builtin_test_flags instead of delegating (split from #915)
**Severity: Low**

## Summary

Split from #915, which was mostly about `builtin_scan` returning wrong output for capture groups (already fixed in #913). The other half of that issue — `builtin_test_regex` (bare `test(re)`, no flags argument) independently reimplementing pattern/input fetch and regex build instead of delegating to `builtin_test_flags` — is still unfixed.

Unlike `scan`, this hasn't drifted into a wrong-output bug yet: `test()` only returns a boolean `is_match`, so there's no capture-group-shaped output to diverge on. It's a maintenance-drift risk, not a live bug — any future change to the flags-aware matching logic only lands in one of the two copies.

## Root cause

`builtin_test_regex` (`src/jq/eval.rs`, ~line 6684) has its own standalone body for pattern/input fetch and `build_regex` + match, rather than delegating to `builtin_test_flags` (~line 7197) with `flags: None` — the same duplication-instead-of-delegation anti-pattern #906 fixed in `builtin_splits` and #913 (via #915) fixed in `builtin_scan`.

## Suggested fix

Delegate: `builtin_test_regex` → `builtin_test_flags::<W, S>(re_expr, None, value, optional)`, matching the established local convention `builtin_sub`/`builtin_gsub`/`builtin_splits`/`builtin_scan` already use post-#906/#913.

newhoggy:
## Summary

`builtin_test_regex` (bare `test(re)`, no flags argument) had its own standalone body for pattern/input fetch and regex build, rather than delegating to `builtin_test_flags` with `flags: None` — the same duplication-instead-of-delegation anti-pattern #906/#913 fixed in `builtin_splits`/`builtin_scan`.

Unlike `scan`, this hadn't drifted into a wrong-output bug: `test()` only returns a boolean `is_match`, so there's no capture-group-shaped output to diverge on. Still a maintenance-drift risk worth closing off, since any future change to the flags-aware matching logic previously only landed in one of the two copies.

Extracted `builtin_test_with_flags` (taking `flags: Option<&str>`, matching `build_regex`'s own signature) as the shared implementation; both `builtin_test_regex` and `builtin_test_flags` are now thin wrappers around it, following the established local convention.

Fixes #923

## Test plan

- [x] `cargo fmt --check`
- [x] Diffed output against pinned real `jq` for bare `test()` (match/non-match, case-sensitivity, non-string pattern/input errors, `?`-optional on a bad pattern) and confirmed the flagged form is unaffected
- [x] Added `test_regex_bare_test_delegates_to_flagged_923` covering all of the above

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -27320,6 +27320,47 @@ mod tests {
 
     #[cfg(feature = "regex")]
     #[test]
+    fn test_regex_bare_test_delegates_to_flagged_923() {
+        // #923: bare test(re) now delegates to builtin_test_with_flags
+        // instead of duplicating pattern/input fetch and regex build.
+        // Every case verified live against jq 1.7.1.
+
+        // Ordinary match/non-match, unaffected by the delegation.
+        query!(br#""abc""#, r#"test("abc")"#,
+            QueryResult::Owned(OwnedValue::Bool(b)) => {
+                assert!(b);
+            }
+        );
+        query!(br#""ABC""#, r#"test("abc")"#,
+            QueryResult::Owned(OwnedValue::Bool(b)) => {
+                assert!(!b, "bare test() must stay case-sensitive (no flags)");
+            }
+        );
+
+        // Non-string pattern and non-string input error the same way as
+        // the flagged form (both route through the same shared checks now).
+        query!(br#""abc""#, r#"try test(5) catch "ERR""#,
+            QueryResult::Owned(OwnedValue::String(s)) => {
+                assert_eq!(s, "ERR");
+            }
+        );
+        query!(b"5", r#"try test("abc") catch "ERR""#,
+            QueryResult::Owned(OwnedValue::String(s)) => {
+                assert_eq!(s, "ERR");
+            }
+        );
+
+        // `?` suppresses a bad pattern with no output, matching
+        // test(re;flags)'s existing optional behavior.
+        query!(br#""abc""#, r#"[test("(")?]"#,
+            QueryResult::Owned(OwnedValue::Array(vs)) => {
+                assert!(vs.is_empty());
+            }
+        );
+    }
+
+    #[cfg(feature = "regex")]
+    #[test]
     fn test_regex_flags_s_m_p() {
         // The #727 repro: jq's `s` (single-line mode) leaves dot NOT
         // matching newline.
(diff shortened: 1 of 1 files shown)


newhoggy:
## Review: `/code-review` (10 angles) — clean, one real finding split out

All 10 angles confirmed this is a faithful, behavior-preserving delegation: argument order, `None`/`Some("")` equivalence in `build_regex`, evaluation order, `optional`-suppression at every checkpoint, clone counts, and generic bounds all verified unchanged. Angles A and D independently traced every deleted line back to its verbatim reappearance in `builtin_test_with_flags`. Angle Conventions found 0 CLAUDE.md/testing-skill violations. Full test suite, clippy (both invocations), and fmt all clean.

**Angles Reuse, Altitude, and Simplification** independently converged on a pre-existing (not introduced here) duplication: the pattern-eval/input-fetch/flags-eval blocks this PR's new `builtin_test_with_flags` shares its shape with are hand-duplicated across 7-9 sibling `_flags`/`_with_flags` builtins, with no generic helper.

Digging into *why* that matters (Angle Altitude's lead) surfaced a real, live bug rather than just a style nit — verified against pinned jq 1.7.1:

```
$ echo '"x"' | jq -c 'try test(1) catch .'
"number not a string or array"
$ echo '"x"' | succinctly jq -c 'try test(1) catch .'
"number not a string or array"          # matches — test/match/capture use EvalError::not_string_or_array

$ echo '"x"' | jq -c 'try sub(1; "y") catch .'
"number (1) is not a string"
$ echo '"x"' | succinctly jq -c 'try sub(1; "y") catch .'
"expected string, got pattern"          # doesn't match jq at all — scan/gsub/sub/splits use a generic type_error() fallback
```

jq itself uses two different wordings for the same conceptual error depending on the builtin family; succinctly's `test`/`match`/`capture` happen to match one of them, but `scan`/`gsub`/`sub`/`splits` match neither. Filed as **#926**, out of scope for this PR since it spans four builtins this diff doesn't touch — this PR only delegates `test`'s implementation, it doesn't change what error `test` raises (still correctly `not_string_or_array`, confirmed live above).

PR #927, opened 2026-08-15
fix(jq): scan/gsub/sub/splits report jq's real non-string-pattern wording

### LINKED ISSUE #926: jq: scan/gsub/sub/splits report the wrong wording for a non-string pattern argument
**Severity: Low**

## Summary

`/code-review` on PR #925 (multiple independent angles: Reuse, Altitude, Simplification) flagged that `test`/`match`/`capture`'s non-string-pattern error uses a different wording than `scan`/`gsub`/`sub`/`splits`'s. Verifying live against pinned jq 1.7.1 shows this is worse than internal inconsistency — real jq itself uses two different wordings depending on the builtin family, and succinctly's `scan`/`gsub`/`sub`/`splits` match **neither**:

```
$ echo '"x"' | jq -c 'try test(1) catch .'
"number not a string or array"
$ echo '"x"' | succinctly jq -c 'try test(1) catch .'
"number not a string or array"          # matches jq ✓

$ echo '"x"' | jq -c 'try sub(1; "y") catch .'
"number (1) is not a string"
$ echo '"x"' | succinctly jq -c 'try sub(1; "y") catch .'
"expected string, got pattern"          # does NOT match jq ✗
```

Confirmed the same divergence for `scan(1)`, `gsub(1; "y")`, and `splits(1)` — all report jq's `"<type> (<value>) is not a string"` for the real oracle, but succinctly's generic `EvalError::type_error("string", "pattern")` (`src/jq/eval.rs`) produces `"expected string, got pattern"` instead: wrong wording, and the literal `"pattern"` isn't even the actual runtime type/value.

`match`/`capture`/`test` are unaffected — they already use `EvalError::not_string_or_array`, which happens to match jq's wording for *that* family.

## Root cause

`docs/compliance/jq/limitations.md`'s "message vocabulary" table documents `not_string_or_array` as the constructor for jq's "not a string or array" family, but there's no equivalent named constructor for jq's other family — `"<type> (<value>) is not a string"`, used specifically for the pattern argument to `scan`/`gsub`/`sub`/`splits`. Those four builtins fall back to the generic `EvalError::type_error("string", "pattern")`, which `error.rs`'s own doc comment says is meant only for "error sites that have no jq counterpart" — but this one does have a jq counterpart, just not the same wording `not_string_or_array` produces.

## Suggested fix

Add a named constructor (e.g. `EvalError::not_a_string(type_name, rendered_value)`) producing jq's exact `"<type> (<value>) is not a string"` wording, and use it at the pattern-eval site in `builtin_sub_with_flags`, `builtin_gsub_with_flags`, `builtin_scan_with_flags`, and `builtin_splits_with_flags` (`src/jq/eval.rs`) instead of `type_error("string", "pattern")`. Add golden/probe coverage (`tests/data/jq-error-probes.tsv` or a direct unit test) pinning the corrected wording against the pinned jq oracle.

Split out from PR #925's review rather than fixed there, since #925 only touches `test`'s delegation and this spans four unrelated builtins.

newhoggy:
## Summary

`scan`/`gsub`/`sub`/`splits` used the generic `EvalError::type_error("string", "pattern")` for a non-string pattern argument, producing `"expected string, got pattern"` — a message that doesn't even substitute the actual runtime type, and doesn't match real jq's wording at all.

```
$ echo '"x"' | jq -c 'try sub(1; "y") catch .'
"number (1) is not a string"
$ echo '"x"' | succinctly jq -c 'try sub(1; "y") catch .'
"expected string, got pattern"
```

Verified live against jq-1.7.1: jq uses **two different sentences** for a non-string pattern depending on the builtin family. `test`/`match`/`capture` report `"<type> not a string or array"` (already correct here, via `not_string_or_array`). `scan`/`gsub`/`sub`/`splits` instead report `"<v> is not a string"` (with a value preview) — added `EvalError::is_not_a_string` for this second family and switched all four call sites to it. `split(re)` has its own distinct jq wording (`"split input and separator must be strings"`) and is untouched.

Extended `tests/data/jq-error-probes.tsv` with the four missing probes — `scan`/`gsub`/`sub`/`splits`'s non-string-pattern case existed for `test`/`match`/`capture` already, but not for this family, which is the exact gap that let this wording drift go undetected — and regenerated `jq-error-messages.tsv` from the pinned jq oracle. Updated `docs/compliance/jq/limitations.md`'s asserted probe count (192 → 196), which the test suite itself asserts against.

Found by `/code-review` on #925 — three independent angles (Reuse, Altitude, Simplification) converged on the underlying duplication between `test`'s implementation and its siblings; digging into *why* it mattered (rather than just deduping) surfaced this live wording bug.

Fixes #926

## Test plan

- [x] `cargo test --features cli,regex --test jq_error_message_tests` (196/196 probes match jq, including the 4 new ones)
- [x] `cargo fmt --check`
- [x] Diffed output against pinned real `jq` for `scan`/`gsub`/`sub`/`splits` (bare and flagged) with a number/object/array pattern, and confirmed `test`/`match`/`capture`/`split` are unaffected
- [x] Added `test_regex_non_string_pattern_wording_matches_jq_per_family_926` covering all of the above

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -24,14 +24,14 @@ For jq *feature* coverage rather than error wording, see
 
 ## Summary
 
-Measured against jq-1.7.1 over the 196 probes in
+Measured against jq-1.7.1 over the 197 probes in
 [`tests/data/jq-error-probes.tsv`](../../../tests/data/jq-error-probes.tsv), through
 **both** evaluators — the full one (`src/jq/eval.rs`) and the generic one
 (`src/jq/eval_generic.rs`, which the CLI uses):
 
 | Dimension                                    | Result               | Meaning                                            |
 |----------------------------------------------|----------------------|----------------------------------------------------|
-| **Message text** (both evaluators, verbatim) | **196/196 = 100.0%** | Byte-identical to jq                               |
+| **Message text** (both evaluators, verbatim) | **197/197 = 100.0%** | Byte-identical to jq                               |
 | **Wording divergences**                      | **0**                | Every probe that errors in both errors identically |
 | **Behaviour / parser gaps**                  | **0**                | succinctly does not raise the error at all         |
 
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -24,14 +24,14 @@ For jq *feature* coverage rather than error wording, see
 
 ## Summary
 
-Measured against jq-1.7.1 over the 192 probes in
+Measured against jq-1.7.1 over the 196 probes in
 [`tests/data/jq-error-probes.tsv`](../../../tests/data/jq-error-probes.tsv), through
 **both** evaluators — the full one (`src/jq/eval.rs`) and the generic one
 (`src/jq/eval_generic.rs`, which the CLI uses):
 
 | Dimension                                    | Result               | Meaning                                            |
 |----------------------------------------------|----------------------|----------------------------------------------------|
-| **Message text** (both evaluators, verbatim) | **192/192 = 100.0%** | Byte-identical to jq                               |
+| **Message text** (both evaluators, verbatim) | **196/196 = 100.0%** | Byte-identical to jq                               |
 | **Wording divergences**                      | **0**                | Every probe that errors in both errors identically |
 | **Behaviour / parser gaps**                  | **0**                | succinctly does not raise the error at all         |
 
@@ -78,6 +78,7 @@ iterate over number` for the same condition, and `cannot parse 'a' as number` ag
 | `<v> cannot be sorted, as it is not an array`            | `cannot_be_sorted`                            |
 | `<v> cannot be matched, as it is not a string`           | `cannot_be_matched`                           |
 | `<t> not a string or array`                              | `not_string_or_array`                         |
+| `<v> is not a string`                                    | `is_not_a_string`                             |
 | `<v> cannot be parsed as a number`                       | `cannot_parse_as_number`                      |
 | `<v> only strings can be parsed`                         | `only_strings_can_be_parsed`                  |
 | `<v> only strings have UTF-8 byte length`                | `no_utf8_byte_length`                         |
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -690,6 +690,21 @@ impl EvalError {
         Self::new(format!("{type_name} not a string or array"))
     }
 
+    /// `<v> is not a string`.
+    ///
+    /// jq's wording when scan/gsub/sub/splits's *pattern* argument isn't a
+    /// string — a genuinely different sentence from `not_string_or_array`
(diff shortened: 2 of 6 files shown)


newhoggy:
## Review update: `/code-review` (10 angles) — one coverage gap fixed, four findings split out

**Angle C** (cross-file tracer) found a real coverage gap: `builtin_split_regex` (the `split(re;flags)` two-arg regex form) was one of the 4 sites this PR fixed, but only `scan`/`gsub`/`sub`/`splits` got a golden probe. Fixed: added `split_regex_arg_non_string` to `tests/data/jq-error-probes.tsv`, regenerated the message table (197 probes now), and verified `split(1;"x")` matches jq's `"number (1) is not a string"` — distinct from `split(1)`'s own wording, which stays untouched.

**Angle Simplification** found the new unit test duplicated existing TSV-driven coverage for its `test`/`match`/`capture` loop and trailing `split(1)` check — `jq_error_message_parity` already asserts those pairs through both evaluators, stronger than a single-evaluator inline copy. Trimmed the test down to what the TSV corpus can't cover on its own: the flagged sibling sharing its bare form's wording, and the object/array value-preview rendering.

**Angle Altitude** dug into whether this PR's fix was scoped at the right depth and found it was — but also surfaced that the *flags* argument to the same 8 builtins has an analogous wording gap, and it's more subtle than the pattern-argument fix: `test`/`match`/`capture`/`sub`'s flags argument needs the same `is_not_a_string` wording, but `gsub`/`scan`/`split`/`splits`'s needs a **third, different** wording (an arithmetic `"cannot be added"` error, since jq implements the flags check by string-concatenating `"g"` onto it before ever checking the type — operand order flips per builtin, matching the g-prepend-vs-append split #730 already found for valid flags). Verified every case live. Also surfaced that `EvalError::type_error` misuse recurs well beyond the regex family (`min_by`/`max_by`/`join`/`flatten`/`limit`, several with a matching unused constructor already sitting in `error.rs`). Filed as **#928** (flags-argument wording, directly adjacent scope) and **#929** (broader systematic audit, open-ended).

**Angle D** (language-pitfall) and **Angle Efficiency** independently found two more real gaps in the shared `describe()`/`dump_truncated` value-preview machinery — pre-existing, but newly *reachable* because this PR is the first to route the regex pattern argument through it at all: an overflowed/infinite float renders as `null` instead of jq's literal text (**#930**), and the `Object` preview arm does an unbounded `.collect()` over the whole object before the 14-byte truncation budget is ever applied (**#931**). Both verified live / confirmed by reading `stream.rs`; both belong in the shared rendering code, not this PR.

**Angles Reuse, Conventions, B (removed-behavior), E (wrapper/proxy), and A (line-by-line)** came back clean — no correctness regressions, no CLAUDE.md violations, no dead code, no wiring bugs, both evaluators confirmed to share the fixed code path.

Commits: `88a1090f5`/`db447d614` (original fix+test), `9ea211d75` (split_regex coverage + test trim, this update).

Issue #932, opened 2026-08-15
jq: any(gen;cond)/all(gen;cond)/IN/IN(src;s) leak halt_error/stderr's print side effect from a discarded generator tail


newhoggy:
**Severity: Low**

## Summary

`any(gen; cond)`/`all(gen; cond)` (and, as of #910's fix, `IN(s)`/`IN(src; s)`, which now shares the same implementation shape) evaluate `gen` eagerly to completion via `eval_owned_expr_fork`, then check `cond` against the already-produced outputs, discarding any trailing error/break if a match is already found among them. This is a deliberate, documented trade-off (see `any_all_gen_cond`'s own doc comment) that's *usually* externally indistinguishable from true step-by-step generator laziness -- **except** for builtins whose Rust implementation prints as an immediate side effect of being evaluated, rather than deferring the print until their `QueryResult` is actually consumed by the top-level runner. `halt_error(n)` and `stderr` are both implemented this way; `error(...)` and `debug` are not (their "output" is just data carried in the `QueryResult`, with no direct I/O at evaluation time).

Concretely: even though the *logical* result of `any`/`all`/`IN` correctly matches real jq (a match found before the discarded tail short-circuits, and the process does *not* actually halt), `halt_error`/`stderr`'s side-printed bytes have already leaked to stdout/stderr by the time that discard decision is made -- something a real, lazily-pulling generator would never have printed at all, since it would never have asked `gen` for that output in the first place.

## Repro (verified against pinned jq 1.7.1)

```
$ echo 2 | jq -c 'any(2, halt_error(3); . == 2)'
true
$ echo 2 | succinctly jq -c 'any(2, halt_error(3); . == 2)'
2
true
```
(the `2` is `halt_error(3)`'s own stderr print of `.` -- confirmed via `2>&1` merging -- leaking even though the overall answer, `true`, is correct and the process does not actually halt)

```
$ echo 2 | jq -c 'any(2, (5|stderr); . == 2)'
true
$ echo 2 | succinctly jq -c 'any(2, (5|stderr); . == 2)'
5true
```
(same leak for `stderr`)

```
$ echo 2 | jq -c 'any(2, (5|debug); . == 2)'
true
$ echo 2 | succinctly jq -c 'any(2, (5|debug); . == 2)'
true
```
(`debug` does *not* leak -- confirms the issue is specific to builtins with an immediate I/O side effect at evaluation time, not a general property of the discard mechanism)

[...]

PR #933, opened 2026-08-15
fix(jq): IN(s)/IN(src;s) short-circuit on an earlier match instead of evaluating eagerly

### LINKED ISSUE #910: jq: IN(s)/IN(src;s) eagerly evaluate their whole generator instead of short-circuiting on first match, unlike real jq
**Severity: Low**

## Summary

`IN(s)` (`def IN(s): any(s == .; .);`) and `IN(src; s)` should short-circuit as soon as a match is found, exactly like `any/2`'s real generator semantics — once `s == .` is truthy for one output of `s`, real jq's `any` never asks `s` for its next output. `builtin_upper_in`/`builtin_upper_in_src` (`src/jq/eval.rs`, immediately following `builtin_in`) instead evaluate `s`'s entire generator up front before checking any candidate, so a side-effecting or erroring later candidate still fires even when an earlier one already produced the match.

Found during #908's review (which fixed the sibling `in(xs)`'s own eager-evaluation/escape-swallowing bug, #880) while cross-file-tracing `builtin_in`'s neighbors for the same bug family; confirmed independently against real jq.

```
$ echo 2 | jq -c 'IN(2, error("boom"))'
true
$ echo 2 | succinctly jq -c 'IN(2, error("boom"))'
```

Real jq's `any(s == .; .)` finds the match on `s`'s first output (`2 == 2`) and never asks `s` for its second output (`error("boom")`) at all — `any` short-circuits on the first truthy condition, per its own generator semantics. succinctly instead evaluates `s`'s whole comma expression eagerly before ever checking a candidate, so the later `error("boom")` fires regardless of whether an earlier candidate already matched.

```
$ echo 2 | jq -c 'IN(3, error("boom"))'
```
(Confirms this isn't just "IN never errors" — when no earlier candidate matches, real jq does reach and raise the later error; the divergence is specifically about *skipping unreached candidates once a match already fired*.)

## Root cause

Not yet traced to a specific line/mechanism — likely the same "materialize the whole generator via `eval_owned_multi`/`eval_owned_multi_keep_partial` before consulting any candidate" shape #880 diagnosed for `builtin_in`, applied to `builtin_upper_in`/`builtin_upper_in_src`'s own generator evaluation instead. Needs its own investigation before scoping a fix, since `any`'s short-circuit-on-first-truthy semantics are a different shape than `in(xs)`'s "keep every candidate, fork per output" semantics #880 implemented — porting that fix directly isn't necessarily correct here.

## Why this is separate from #880/#908

Different builtins (`IN`/`IN(src;s)`, not `in`), immediately adjacent in the file and sharing the general "eager generator evaluation loses real jq's laziness" bug class, but not touched by that PR's diff.

## Suggested fix direction

Investigate whether this codebase already has a short-circuiting multi-output evaluator suitable for `any`-style "stop on first truthy" semantics (as opposed to `eval_owned_multi_keep_partial`'s "keep every output" semantics), or whether one needs to be added. `any`/`all`'s own existing builtin implementations (if they don't share this eager-evaluation problem) may already have the right shape to model this on.

newhoggy:
## Summary

`IN(s)`/`IN(src; s)` used to evaluate their generator argument(s) fully via `eval_owned_multi` before ever checking for a match, so a later output's error fired even when an earlier output already satisfied the condition — unlike real jq's `any`-based definitions, which stop asking for more outputs once a match is found.

```
$ echo 2 | jq -c 'IN(2, error("boom"))'
true
$ echo 2 | succinctly jq -c 'IN(2, error("boom"))'
```

- Fixed by reusing `eval_owned_expr_fork`'s "keep the produced prefix, thread the trailing escape separately" split — the exact same shape `any(gen;cond)`/`all(gen;cond)` already use in this file (`any_all_gen_cond`). A match already present in the prefix short-circuits before the trailing escape is ever consulted.
- For `IN(src;s)`, when neither side matches and both carry a trailing escape, `src`'s wins — confirmed against real jq: `IN(3, error("src-boom"); 4, error("s-boom"))` reports `"src-boom"`.
- All existing error-propagation tests for both builtins (no-match case) still pass unchanged.

Also files and documents a known, narrow, **pre-existing** gap this fix inherits from `any`/`all`'s own established shape (not introduced by this PR, just discovered while verifying it): `halt_error`/`stderr` print as an immediate side effect of evaluation rather than of their result being consumed, so their bytes can still leak from a discarded generator tail even though the overall answer and exit behavior stay correct. Filed as #932 (needs true lazy generator evaluation to fully close — out of scope here).

Fixes #910

## Test plan
- [x] `cargo fmt --check`
- [x] Manually re-verified every case (short-circuit, no-match propagation, tie-break, Halt/Break propagation) against real jq 1.7.1

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -3144,17 +3144,43 @@ fn builtin_in<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
 
 /// Builtin: `IN(s)` - true if any output of `s` (run against the current
 /// value) equals the current value. Matches jq's `any(s == .; .)`.
+///
+/// #910 fix: `s` used to be evaluated eagerly to completion via
+/// `eval_owned_multi` before any equality check ran, so a later output's
+/// error fired even when an earlier output already matched -- unlike real
+/// jq's `any`-based definition, which stops asking `s` for outputs the
+/// moment a match is found. Reuses `eval_owned_expr_fork` (the same "keep
+/// the already-produced prefix, thread the trailing escape separately"
+/// split `any_all_gen_cond`/`eval_reduce`/`eval_foreach` already use
+/// throughout this file) so a match already present in the prefix
+/// short-circuits before the trailing escape is ever consulted. This
+/// isn't true step-by-step generator laziness -- `s` is still evaluated to
+/// completion internally -- but it's externally indistinguishable from it
+/// for `error(...)`/ordinary values, the same trade-off `any_all_gen_cond`'s
+/// own doc comment documents. Confirmed against jq 1.7.1: `IN(2,
+/// error("boom"))` on `2` is `true` (the error is never reached), `IN(3,
+/// error("boom"))` on `2` still raises it (no earlier candidate matched, so
+/// the trailing escape is the only verdict left). Known gap (#932, shared
+/// with `any`/`all`, pre-existing there): `halt_error`/`stderr` print as an
+/// immediate side effect of being *evaluated*, not of their result being
+/// consumed, so a match found before reaching one in `s` still leaks its
+/// printed bytes even though the overall answer and process-exit behavior
+/// stay correct -- `debug`/`error` don't have this problem, since neither
+/// performs I/O at evaluation time.
 fn builtin_upper_in<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     s: &Expr,
     value: StandardJson<'a, W>,
     _optional: bool,
 ) -> QueryResult<'a, W> {
     let current = to_owned(&value);
-    let candidates = match eval_owned_multi::<S>(s, &current) {
-        Ok(c) => c,
-        Err(e) => return e.into(),
-    };
-    QueryResult::Owned(OwnedValue::Bool(candidates.contains(&current)))
+    let (candidates, trailing) = eval_owned_expr_fork::<S>(s, &current, false);
+    if candidates.contains(&current) {
+        return QueryResult::Owned(OwnedValue::Bool(true));
+    }
+    match trailing {
+        Some(control) => control_to_result(control),
+        None => QueryResult::Owned(OwnedValue::Bool(false)),
+    }
 }
 
 /// Builtin: `IN(src; s)` - true if any output of `src` equals any output of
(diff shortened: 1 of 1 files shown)


newhoggy:
Update: review (10 parallel finder angles) found a real, confirmed correctness regression in the original commit's \`builtin_upper_in_src\` fix — the hand-rolled "fork both operands, then check the cross-product" approach doesn't respect real jq's actual nested-loop evaluation order for \`src == s\` (\`s\` is jq's outer generator loop, \`src\` is the inner loop). Concretely:

- \`IN(4, error("boom"); 9, 4)\` wrongly answered \`true\` (should raise \`"boom"\` — \`s\`'s first output \`9\` drives a full inner pass over \`src\` that hits the error before \`s\`'s second output \`4\`, where the match lives, is ever reached)
- \`IN(9, error("boom"); empty)\` wrongly raised the error (should be \`false\` — \`s\` produces zero outputs, so \`src\` is never touched)
- \`IN(error("src-boom"); error("s-boom"))\` wrongly reported \`"src-boom"\` (should be \`"s-boom"\` — same reasoning)
- \`IN(error("src-boom"); halt_error(7))\` wrongly masked the halt with an ordinary error (should actually halt, exit 7)

Fixed in a follow-up commit (\`8ee4a5828\`) by deleting the hand-rolled cross-product entirely and instead constructing the real \`src == s\` comparison expression, routed through \`any_all_gen_cond\` (the same shared engine \`any(gen;cond)\` already uses) — reusing \`eval_compare\`/\`binary_fanout_core\`'s own already-correct generator order instead of re-deriving it by hand. All four cases above now match real jq exactly (independently re-verified against the pinned jq 1.7.1 oracle). \`builtin_upper_in\` (the single-argument form) was not affected — it has no operand-ordering ambiguity and was independently confirmed correct by multiple reviewers.

Also filed additional findings from this review round onto the already-open #932 (self-recursive/infinite generators crash with a native stack overflow in \`any\`/\`all\`/\`IN\`/\`first\`/\`limit\` even in the best case for short-circuiting; \`first\`/\`limit\` share the same \`halt_error\`/\`stderr\` side-effect-leak shape) since they're the same pre-existing architectural gap (no true lazy generator evaluation), not something in scope for this PR.

PR #936, opened 2026-08-15
fix(jq): report jq's real wording for a non-string regex flags argument

### LINKED ISSUE #928: jq: test/match/capture/sub/gsub/scan/split/splits report the wrong wording for a non-string flags argument
**Severity: Low**

## Summary

Split from #926/#927's review: the *flags* argument to `test`/`match`/`capture`/`sub`/`gsub`/`scan`/`split`/`splits` has the same `EvalError::type_error("string", "flags")` problem #926 fixed for the *pattern* argument — but the correct fix is more subtle than just swapping in `is_not_a_string`, because jq implements the flags-argument type check differently depending on the builtin family.

**`test`/`match`/`capture`/`sub`'s flags argument** — same wording as #926's `is_not_a_string`:

```
$ echo '"a"' | jq -c 'try test("a"; 1) catch .'
"number (1) is not a string"
$ echo '"a"' | succinctly jq -c 'try test("a"; 1) catch .'
"expected string, got flags"
```

(Same divergence for `match("a"; 1)`, `capture("(?<a>a)"; 1)`, `sub("a";"b";1)`.)

**`gsub`/`scan`/`split`/`splits`'s flags argument** — a *third*, distinct wording: an arithmetic error, because jq implements these by string-concatenating the flags with `"g"` (`flags + "g"` or `"g" + flags`) before ever checking the type, so a non-string flags value hits jq's `+` operator's own type check instead:

```
$ echo '"a"' | jq -c 'try gsub("a";"b";1) catch .'
"number (1) and string (\"g\") cannot be added"
$ echo '"a"' | succinctly jq -c 'try gsub("a";"b";1) catch .'
"expected string, got flags"

$ echo '"a"' | jq -c 'try scan("a";1) catch .'
"string (\"g\") and number (1) cannot be added"
$ echo '"a"' | succinctly jq -c 'try scan("a";1) catch .'
"expected string, got flags"
```

Note the operand order flips between `gsub` (`flags + "g"`) and `scan`/`split`/`splits` (`"g" + flags"`) — matching the g-prepend-vs-append distinction #730 already found and fixed for *valid* flags strings, now showing up again on the type-error path.

## Root cause

Same as #926: `EvalError::type_error("string", "flags")` is used at 8 call sites in `src/jq/eval.rs` (the flags-eval block shared across `test_flags`/`match_flags`/`capture_flags`/`sub_flags`/`gsub_flags`/`scan_flags`/`split_regex`/`splits_flags`) instead of jq's real wording. Unlike the pattern-argument fix, this one needs **two** different corrected constructors depending on family, not one:
- `test`/`match`/`capture`/`sub`: `EvalError::is_not_a_string` (the constructor #926/#927 already added)
- `gsub`/`scan`/`split`/`splits`: `EvalError::binary_op` (jq's arithmetic-error constructor, already exists and is already used elsewhere in this file) — with the right operand order per builtin

## Suggested fix

At each of the 8 sites, replace `EvalError::type_error("string", "flags")` with the correct constructor for that builtin's family, verified against the live jq wording/operand-order shown above. Add matching probes to `tests/data/jq-error-probes.tsv` (mirroring #926's approach) so the pinned-oracle corpus catches any future regression here the way it now does for the pattern argument.

Found by `/code-review`'s Altitude angle on #927 while judging whether #926/#927's fix was scoped at the right depth — deliberately not fixed there since it needs a different mechanism (two constructors, not one) and touches 8 more call sites.

newhoggy:
## Summary

Split from #926: the flags argument to `test`/`match`/`capture`/`sub`/`gsub`/`scan`/`split`/`splits` used the same placeholder `EvalError::type_error("string", "flags")` the pattern argument did, producing `"expected string, got flags"` instead of jq's real wording.

Unlike the pattern argument, this splits into **two** families rather than one:

- `test`/`match`/`capture`/`sub`: `EvalError::is_not_a_string` (the constructor #926 added) — same wording as the pattern argument.
- `gsub`/`scan`/`split`/`splits`: a **third** wording entirely. jq builds these builtins' internal global-flag string via concatenation (`flags + "g"` for `gsub`, `"g" + flags` for `scan`/`split`/`splits` — the same append-vs-prepend split #730 found for *valid* flags strings) **before** ever checking the flags argument's type, so a non-string value hits jq's own `+` operator type check instead: `"<a> and <b> cannot be added"`, via the already-existing `EvalError::binary_op`. Operand order flips per builtin, matching which side of the concatenation the flags value sits on.

```
$ echo '"a"' | jq -c 'try test("a"; 1) catch .'
"number (1) is not a string"
$ echo '"a"' | jq -c 'try gsub("a";"b";1) catch .'
"number (1) and string (\"g\") cannot be added"
$ echo '"a"' | jq -c 'try scan("a";1) catch .'
"string (\"g\") and number (1) cannot be added"
```

Every wording and operand order verified live against jq 1.7.1 for all 8 call sites, including object/array flags values, `?`-suppression, and `null`-flags (unaffected, already worked).

**Update:** review found a real, higher-severity bug in the same functions: `test`/`match`/`capture`/`sub` evaluated the flags argument *before* the pattern argument, but jq validates pattern first. Once the flags error stopped being an obviously-bogus placeholder, this silently blamed the wrong operand when both arguments were invalid, and skipped a side-effecting pattern expression's `error()` entirely. Fixed by evaluating pattern first, flags second, in all four (`gsub`/`scan`/`split`/`splits` deliberately left unchanged — their own flags-before-pattern order already matches jq's for type mismatches). Filed **#937** (pre-existing, unrelated wording gap the reorder made visible: `test`/`match`/`capture`'s pattern error doesn't switch wording based on flags-argument presence) and **#938** (gsub/scan/split/splits still lose a pattern-argument `error()` to a flags-argument one — a narrower, differently-shaped bug needing a different fix) as follow-ups.

Extended `tests/data/jq-error-probes.tsv` with 8 new probes (one per builtin), regenerated the message table, added all 8 to the `FEATURE_GATED` allowlist (learned from #927: without `regex`, every one of these fails `"regex feature not enabled"` before reaching the type check — including `test`'s flags form specifically, since its non-regex substring-match fallback has no flags support at all, unlike its pattern-argument case, which stays un-gated). Bumped `docs/compliance/jq/limitations.md`'s asserted probe count (197 → 205). Also fixed one pre-existing test's weak assertion (`e.to_string().contains("flags")`) that happened to still pass with the wrong placeholder message — now asserts the exact corrected wording.

Fixes #928

## Test plan

- [x] `cargo test --features cli,regex --test jq_error_message_tests` and `cargo test --test jq_error_message_tests` (205/205 probes match jq, both with and without `regex`)
- [x] `cargo fmt --check`
- [x] Diffed output against pinned real `jq` for all 8 builtins (bare and object/array flags values), `?`-suppression, and `null`-flags
- [x] Added `test_regex_non_string_flags_wording_matches_jq_per_family_928` covering all of the above

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -24,14 +24,14 @@ For jq *feature* coverage rather than error wording, see
 
 ## Summary
 
-Measured against jq-1.7.1 over the 197 probes in
+Measured against jq-1.7.1 over the 205 probes in
 [`tests/data/jq-error-probes.tsv`](../../../tests/data/jq-error-probes.tsv), through
 **both** evaluators — the full one (`src/jq/eval.rs`) and the generic one
 (`src/jq/eval_generic.rs`, which the CLI uses):
 
 | Dimension                                    | Result               | Meaning                                            |
 |----------------------------------------------|----------------------|----------------------------------------------------|
-| **Message text** (both evaluators, verbatim) | **197/197 = 100.0%** | Byte-identical to jq                               |
+| **Message text** (both evaluators, verbatim) | **205/205 = 100.0%** | Byte-identical to jq                               |
 | **Wording divergences**                      | **0**                | Every probe that errors in both errors identically |
 | **Behaviour / parser gaps**                  | **0**                | succinctly does not raise the error at all         |
 
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -6779,21 +6779,26 @@ fn builtin_match<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     value: StandardJson<'a, W>,
     optional: bool,
 ) -> QueryResult<'a, W> {
-    // Get the pattern
+    // Get the pattern. No `Ok(_) if optional` guard: same reasoning as the
+    // flags-eval block below — `optional` is never forced `true` reaching a
+    // nested `Call` like this one (see `Expr::Optional`'s dispatch, #693).
     let pattern = match result_to_owned(eval_single::<W, S>(re_expr, value.clone(), optional)) {
         Ok(OwnedValue::String(s)) => s,
-        Ok(_) if optional => return QueryResult::None,
         Ok(v) => return QueryResult::Error(EvalError::not_string_or_array(v.type_name())),
         Err(e) => return e.into(),
     };
 
-    // Get the flags, now that the pattern is known valid
+    // Get the flags, now that the pattern is known valid. No `Ok(_) if
+    // optional` guard here: `optional` is only ever forced `true` for the
+    // *final* step of `.[EXPR]?`/`.[S:E]?` (see `Expr::Optional`'s dispatch
+    // in `eval_single`, #693) — a nested `Call` like this one never sees it,
+    // so a bare `?` suppresses this error via the ancestor `eval_try`'s
+    // catch, not locally. Confirmed dead: 0 coverage hits pre- and post-#928.
     let flags = match flags_expr {
         None => None,
         Some(fe) => match result_to_owned(eval_single::<W, S>(fe, value.clone(), optional)) {
             Ok(OwnedValue::String(s)) => Some(s),
             Ok(OwnedValue::Null) => Some(String::new()),
-            Ok(_) if optional => return QueryResult::None,
             Ok(v) => return QueryResult::Error(EvalError::is_not_a_string(&v)),
             Err(e) => return e.into(),
         },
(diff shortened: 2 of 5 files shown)


newhoggy:
## Review update: `/code-review` (11 angles) found a real, higher-severity bug — fixed

**Two independent angles** (Wrapper/proxy correctness, and the orchestrator's own direct verification) converged on the same finding: `test`/`match`/`capture`/`sub` evaluated and type-checked the **flags** argument before the **pattern** argument, but real jq validates the pattern first. While the flags-argument error was still the old placeholder (`"expected string, got flags"`), this went unnoticed — nobody could mistake a bogus message for correct attribution. But now that PR #936 gives the flags error real jq wording, a call with *both* arguments invalid started silently blaming the wrong operand, and worse — a side-effecting pattern expression's `error()` call was never evaluated at all:

```
$ echo '"abc"' | jq 'test(1;2)'
number (1) is not a string        # blames the pattern, correctly
$ echo '"abc"' | succinctly jq 'test(1;2)'   # before this fix
number (2) is not a string        # wrongly blames the flags

$ echo '"abc"' | jq 'sub(error("P");"b";2)'
P                                  # pattern's error() wins
$ echo '"abc"' | succinctly jq 'sub(error("P");"b";2)'  # before this fix
number (2) is not a string        # pattern's error() never even ran
```

Fixed by changing `builtin_test_with_flags`/`builtin_match`/`builtin_capture_with_flags`/`builtin_sub_with_flags` to take `flags_expr: Option<&Expr>` instead of a pre-evaluated `flags: Option<&str>`, evaluating it *after* the pattern check. `builtin_sub_with_flags`'s post-resolution logic (input fetch, regex build, replace) is split into a new `sub_with_resolved_pattern` so `builtin_gsub_with_flags` can keep reusing it directly with its own synthesized `flags + "g"` string — verified live that gsub's own flags-before-pattern order already matches jq's, so gsub is deliberately unchanged.

While investigating, I also found (independently, going one step further than the review) that `gsub`/`scan`/`split`/`splits` have a **narrower, differently-shaped** version of the same class of bug: their type-mismatch blame order is already correct, but a genuine `error()` in the pattern argument still loses to one in the flags argument (`gsub(error("P");"b";error("F"))` → jq: `"P"`, succinctly: `"F"`). This needs a different fix (jq's real model appears to bind-and-propagate the pattern expression's own errors *before* the flags-concat type check, but defer the pattern's *type* check to after) — filed separately as **#938**, out of scope for this PR.

Also filed **#937** for a related, pre-existing (not introduced by this PR) gap the reordering fix made directly visible: `test`/`match`/`capture`'s pattern-check always uses `not_string_or_array` regardless of whether a flags argument is present, but jq switches to `is_not_a_string` wording once the 2-arg form is used (even with valid flags). `sub` doesn't have this gap — its pattern-check already used the right wording via #926.

**All other angles** (line-by-line ×2, cross-file ×2, removed-behavior, language-pitfall, altitude/conventions, reuse/simplification/efficiency) came back clean or with only minor polish notes (the four near-identical `binary_op` arms for gsub/scan/split/splits, judged not worth extracting given the per-site "verified live" documentation value each carries).

Commits: `f2c10a315`/`8a3538b2f` (original flags-wording fix+test), `47ac706ea`/`bc953035a` (evaluation-order fix+test, this update).

Issue #939, opened 2026-08-15
jq: keys/.[] on a document-sourced overflow literal previews the internal reindex sentinel, not the source text


newhoggy:
**Severity: Low**

## Summary

Found while implementing #930 (which fixes `describe()`/`dump_truncated()`'s error-message value previews to stop rendering a non-finite float as `null`). For an overflowed numeric literal read from a JSON *document* (as opposed to one written directly in a jq filter expression — see #930's own scope note on that distinction), an operation that reaches `describe()` via `OwnedValue::to_json_for_reindex`'s round-trip (`keys`, `.[]`, and other cursor-materializing operations) previews the internal `1e999`/`9e999e999` round-trip sentinel instead of the document's actual literal text.

## Repro

```
$ echo '123e400' | jq -c 'try keys catch .'
"number (1.23E+402) has no keys"
$ echo '123e400' | succinctly jq -c 'try keys catch .'   # after #930's fix
"number (1E+999) has no keys"
```

Every overflow literal shows the *same* `1E+999`/`-1E+999` text here, regardless of its own mantissa digits — a strong signal the original text ("123e400") never reaches `describe()`, having already been replaced upstream.

## Root cause

`OwnedValue::to_json_for_reindex` (`src/jq/value.rs`) deliberately substitutes a self-overflowing sentinel literal (`"1e999"`/`"-1e999"` via `overflow_literal`, `"9e999e999"` for NaN via `NAN_SENTINEL`) when serializing a non-finite value for a JSON-text round-trip — this is intentional, existing infrastructure (issue #561) whose whole point is that the sentinel re-parses back to the same non-finite value, keeping further cursor operations correct. It was never designed to preserve *display* fidelity, only round-trip *value* fidelity.

`keys`/`.[]`-style operations on a document-sourced value go through exactly this round-trip (`eval.rs`/`eval_generic.rs`, multiple call sites — see `to_json_for_reindex`'s callers) before any error is raised, so by the time `describe()` sees the resulting `NumberLiteral`, its `literal` field already holds the sentinel text, not the document's own.

## Why this is separate from #930

[...]

Issue #387, opened 2026-07-26
jq: tostring/tojson lose a number's source literal (1e100 becomes a 101-digit expansion)


newhoggy:
**Severity: Low** — reproduced against the built CLI (debug, `e94771cd`), oracle `jq-1.7.1-apple`.

`OwnedValue` stores a number as `Int(i64)` or `Float(f64)` and does not carry the literal the document was written with. Anything that materialises therefore re-renders the number from the parsed value, losing jq's formatting:

```console
$ echo 1e100 | jq  -r 'tostring'    # 1E+100
$ echo 1e100 | sjq -r 'tostring'    # 1000000000000...0000  (101 digits)

$ echo 1.0   | jq  -r 'tostring'    # 1.0
$ echo 1.0   | sjq -r 'tostring'    # 1

$ echo 1e100 | jq  -r 'tojson'      # 1E+100
$ echo 1e100 | sjq -r 'tojson'      # 1000000000000...0000
```

The streaming identity path is correct, because it copies source text rather than materialising:

```console
$ echo 1e100 | jq  -c .    # 1E+100
$ echo 1e100 | sjq -c .    # 1E+100   (agrees)
$ echo -0.0  | sjq -c .    # -0.0     (agrees)
```

So the divergence is exactly "did this value get materialised", which is not something a user can predict from the filter they wrote. The same document prints two different ways depending on the query.

Other values in the class: `-0.0` materialises to `0`, `1e-7` to `0.0000001`.

## Fix

Teach `OwnedValue` to carry the source literal for numbers that came from a document — e.g. an optional `literal: Option<Box<str>>` alongside the parsed value, preferred by the serialiser when present and dropped once arithmetic produces a new number (which is what jq does: computed numbers get canonical formatting).

That is a change to a core public type and touches every construction site, so it is well beyond the scope of the issue that surfaced it.

## Related

Found while reviewing #358. The containment error preview renders its operand through `OwnedValue`, so jq's `number (1E+100) and string ("a") ...` comes out as `number (10000000000...) and string ("a") ...`. That case is pinned rather than fixed, by `number_previews_are_canonicalised` in `tests/jq_containment_tests.rs`, with a note to delete the test if this issue is ever fixed.

Not the same as #169, which is the YAML streaming path collapsing integer-valued floats during *parsing*; this one is the jq materialising path in a document whose text was correct.

PR #940, opened 2026-08-15
fix(jq): error-message value previews render non-finite floats like jq (#930)

### LINKED ISSUE #930: jq: error-message value previews render an overflowed/infinite float as null instead of jq's literal text
**Severity: Low**

## Summary

`describe()`/`dump_truncated()` (`src/jq/error.rs`, `src/jq/stream.rs`, `src/jq/value.rs`) — the shared machinery every `subject`-based `EvalError` constructor uses to render a value preview like `number (1)` — renders an overflowed or truly-infinite float as JSON `null` instead of jq's actual text. Pre-existing, but newly exposed on 4 more call sites by #926/#927 (which is the first PR to route `scan`/`gsub`/`sub`/`splits`'s pattern argument through `describe()` at all — previously that argument's error path used a static string regardless of value).

```
$ echo '"x"' | jq -c 'try scan(1e400) catch .'
"number (1E+400) is not a string"
$ echo '"x"' | succinctly jq -c 'try scan(1e400) catch .'
"number (null) is not a string"

$ echo '"x"' | jq -c 'try scan(infinite) catch .'
"number (1.797693134...) is not a string"
$ echo '"x"' | succinctly jq -c 'try scan(infinite) catch .'
"number (null) is not a string"
```

jq's `jv` keeps the literal source text for an overflowed float literal (`1E+400`), and represents `infinite` as `DBL_MAX` (`1.797693134...`, truncated the same way other long previews are) rather than IEEE `inf`. `nan` renders correctly on both sides — only the infinity/overflow path diverges.

## Root cause

`src/jq/value.rs`'s float-to-JSON rendering (used by `stream_owned_value_json_with` / `dump_truncated`, called from `describe()`) maps a non-finite `f64` to JSON `null`, which is valid JSON output behavior in general but wrong specifically for jq's own error-message value previews, where jq preserves the literal/DBL_MAX text instead.

## Scope note

This isn't specific to #926/#927's fix — `describe()` is shared by ~20 other `EvalError` constructors (`cannot_be_matched`, `cannot_iterate`, `binary_op`, etc.), so any of them previewing an overflowed/infinite float value would show the same `(null)` instead of jq's text. #926/#927 just happens to be the PR that first exercises this on the regex-pattern-argument path; the fix belongs in the shared rendering code, not in that PR.

Found by `/code-review`'s language-pitfall angle on #927.

newhoggy:
## Summary

- `describe()`/`dump_truncated()` (`src/jq/error.rs`), the shared machinery every `subject`-based `EvalError` constructor uses to render a value preview like `number (1)`, mapped **any** non-finite float to JSON `null` in error-message previews — but jq's own previews aren't JSON and aren't constrained by RFC 8259, so it shows real text instead:
  - `try scan(infinite) catch .` → jq: `number (1.797693134...) is not a string`; succinctly (before): `number (null) is not a string`
  - `try scan(-infinite) catch .` → jq: `number (-1.79769313...) is not a string`; succinctly (before): `number (null) is not a string`
  - `nan` stays `null` on both sides (unchanged) — it has no literal spelling to fall back to either.
- Root cause was `src/jq/stream.rs`'s shared `stream_owned_value_json_with`, used by **both** the error-preview convention (`stream_owned_value_json_jq`) and real `-o json`/`yq` output (`stream_owned_value_json`) — the fix threads two new per-convention function pointers through it (`infinite_float`/`infinite_literal`) so the **real-output convention is completely unchanged** (still always `null`, matching RFC 8259, pinned by the pre-existing `test_stream_json_number_literal_nan_and_infinite`), while the preview convention now renders real text:
  - A computed infinite `OwnedValue::Float` (the `infinite`/`-infinite` builtins, or any arithmetic overflow — no source literal of its own) renders jq's `DBL_MAX` text (hardcoded, since it's a fixed constant).
  - An infinite `OwnedValue::NumberLiteral` (an overflowed number read from a JSON *document*, which carries its own source text) reformats that text into jq's normalized scientific notation via a new `format_overflow_literal_mantissa` helper in `src/jq/value.rs` — pure string-based mantissa/exponent shifting, **not** a round-trip through `f64`, since `format_number_jq_compat`'s existing `log10`/`pow` renormalization produces garbage (`"NaNE+2147483647"`) on a non-finite input (it's built to renormalize a *finite* value's exponent, and the value here is definitionally out of `f64` range).

## Known, documented, out-of-scope gap

A numeric literal written **directly in a jq filter expression** (e.g. `scan(1e400)`, as opposed to one read from a JSON *document*) still shows `DBL_MAX` text rather than jq's literal-preserving `1E+400`, because `Expr::Literal(Literal::Float(f64))` never carried source text in the first place — only document-sourced values do, via `OwnedValue::NumberLiteral` (issue #387's fix, scoped specifically to document literals). Closing that gap would mean threading literal text through the parser/expr/eval layers for filter-embedded numbers — a materially bigger, separate change. `-1e400` (negated) is unaffected by this gap and matches jq exactly, since jq **itself** shows `DBL_MAX` text for a negated literal (verified live: `-` forces materialization in real jq too, same as any other arithmetic).

Also filed **#939** (separate, pre-existing bug found while implementing this): `keys`/`.[]`-style operations on a document-sourced overflow literal round-trip the value through `OwnedValue::to_json_for_reindex` before any error is raised, which substitutes an internal `1e999` sentinel for the value's real text — so those specific operations' previews show the sentinel rather than the literal, even after this fix. The `NumberLiteral`-arm fix here is still correct and tested directly (unit tests bypass the eval pipeline, constructing the value directly) for when the original literal text *does* reach `describe()` intact.

## Test plan

- [x] New oracle probes (`infinite_preview`, `neg_infinite_preview`, `nan_preview`, `arithmetic_overflow_preview`) added to `tests/data/jq-error-probes.tsv`, captured against pinned jq-1.7.1 via `sync-jq-error-messages.sh` — `jq_error_message_parity` passes 201/201
- [x] Direct unit tests in `src/jq/stream.rs` (`stream_owned_value_json_jq` on constructed `Float`/`NumberLiteral` values) and `src/jq/value.rs` (`format_number_jq_compat` on 8 oracle-verified overflow shapes: `1e400`, `-1e400`, `123e400`, `12.34e400`, `0.5e400`, `100e400`, `9.999e400`, `1E400`, plus a no-exponent 400-digit literal)
- [x] `docs/compliance/jq/limitations.md`'s probe-count summary updated (197 → 201)

Closes #930

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -24,14 +24,14 @@ For jq *feature* coverage rather than error wording, see
 
 ## Summary
 
-Measured against jq-1.7.1 over the 208 probes in
+Measured against jq-1.7.1 over the 212 probes in
 [`tests/data/jq-error-probes.tsv`](../../../tests/data/jq-error-probes.tsv), through
 **both** evaluators — the full one (`src/jq/eval.rs`) and the generic one
 (`src/jq/eval_generic.rs`, which the CLI uses):
 
 | Dimension                                    | Result               | Meaning                                            |
 |----------------------------------------------|----------------------|----------------------------------------------------|
-| **Message text** (both evaluators, verbatim) | **208/208 = 100.0%** | Byte-identical to jq                               |
+| **Message text** (both evaluators, verbatim) | **212/212 = 100.0%** | Byte-identical to jq                               |
 | **Wording divergences**                      | **0**                | Every probe that errors in both errors identically |
 | **Behaviour / parser gaps**                  | **0**                | succinctly does not raise the error at all         |
 
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -1952,10 +1952,16 @@ struct JqCompatFormatter;
 impl LiteralFormatter for JqCompatFormatter {
     fn format_raw_number<'a>(&self, raw: &'a [u8]) -> Cow<'a, str> {
         // A source literal that overflows to ±Infinity/NaN (e.g. `1e400`)
-        // must not be fed to `format_number_jq_compat`, which assumes a
-        // finite value and produces garbage like "NaNE+2147483647" for one
-        // that isn't (#561). Match `format_float`'s guard below: JSON output
-        // substitutes "null" for values RFC 8259 can't represent.
+        // must not be fed to `format_number_jq_compat` here: this is real
+        // JSON output (`--jq-compat`), which RFC 8259 forbids an Infinity/NaN
+        // literal from, so it must substitute "null" regardless of what text
+        // the function would produce - match `format_float`'s guard below.
+        // (`format_number_jq_compat` itself now handles a non-finite input
+        // correctly rather than the "NaNE+2147483647"-style garbage #561
+        // fixed for the plain-overflow case - #930 closed the same gap for
+        // the *literal*-with-exponent case - but that's jq's real-text
+        // convention for error-message previews, the wrong one for this
+        // RFC-8259-constrained call site.)
         let overflows = core::str::from_utf8(raw)
             .ok()
             .and_then(|s| s.parse::<f64>().ok())
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -296,7 +296,7 @@ impl EvalError {
     /// The message renders `value` the way jq reports it: a string payload is
     /// used as-is (`error("boom")` reports `boom`, not `"boom"`), anything else
     /// is serialized via the jq-error-message convention (`stream_owned_value_json_jq`,
-    /// the same one [`describe`]/[`dump_truncated`] use — not [`OwnedValue::to_json`],
+    /// the same one `describe`/`dump_truncated` use — not [`OwnedValue::to_json`],
     /// whose RFC-8259-safe `null` for a non-finite float is wrong here: `error(infinite)`
     /// reports jq's real `DBL_MAX` text, not `null` (#930)). Unlike `dump_truncated`,
     /// this has no length budget — `error(v)`'s message is the whole value, verbatim,
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -295,11 +295,20 @@ impl EvalError {
     ///
     /// The message renders `value` the way jq reports it: a string payload is
     /// used as-is (`error("boom")` reports `boom`, not `"boom"`), anything else
-    /// is serialized as JSON.
+    /// is serialized via the jq-error-message convention (`stream_owned_value_json_jq`,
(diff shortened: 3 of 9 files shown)


newhoggy:
Code review (10-agle multi-pass) found two genuine defects and one missed sibling path in the initial commit — all fixed as a follow-up commit, verified against the pinned jq oracle:

1. **Panic**: `format_overflow_literal_mantissa` panicked on a leading-dot overflow literal (`.5e400`, no digit before the decimal point) — `int_part` was empty rather than `"0"`, falling into the wrong branch and slicing an empty string.
2. **Silent wrong output**: the exponent digit string can itself overflow `i32`, which the original fix's `unwrap_or(0)` silently zeroed, producing `"1E+0"` for an astronomically large value. Also had no ceiling matching jq's own literal-preservation limit (~1,000,000,000, decNumber's limit) — beyond it, jq falls back to `DBL_MAX` text, which this now does too. Both oracle-verified.
3. **Missed sibling path**: `EvalError::from_value` (backing `error(v)`'s uncaught diagnostic) still rendered its non-string payload via `OwnedValue::to_json()` — the real-output convention, not the jq-error-preview convention `describe()`/`dump_truncated()` were fixed to use. `error(infinite)` uncaught showed `null` where real jq shows `DBL_MAX` text. Now routes through `stream_owned_value_json_jq`.

Also bounded `format_overflow_literal_mantissa`'s own cost (review flagged it could do unbounded work on a document-controlled mantissa before `dump_truncated`'s ~14-byte truncation ever kicks in) — capped to 32 rendered digits, with a test pinning both the cap and that the exponent math stays correct regardless of mantissa length.

Two stale comments (claiming `format_number_jq_compat` "produces garbage" on non-finite input, no longer true) were also corrected.

Lower-priority style findings (bundling the 4 new fn-pointer params into a struct, avoiding a few small allocations on the non-finite path, unifying two structurally-similar branches) were left as-is — non-blocking, and the fn-pointer/allocation ones are already partially addressed as a side effect of the correctness rewrite (shared `assemble_scientific` helper, sign passed from the caller instead of re-derived from text).

All local checks green (full test suite, both clippy invocations, oracle parity 209/209) after these fixes and a rebase onto latest `main`.

newhoggy:
CI caught one more issue: the 4 new oracle probes used `scan(...)` as their vehicle to reach a non-finite value's preview, but CI also runs the default-features suite (no `regex`), where `scan()` itself fails first. Feature-gated them via the existing `FEATURE_GATED` mechanism (same as 21 other regex-dependent probes) rather than switching vehicles — tried `keys`/`.[]` first, but those route the value through `OwnedValue::to_json_for_reindex` before `describe()` ever sees it, hitting #939's sentinel-substitution bug and defeating the whole point of the probe. Verified locally both with and without `regex`.

Issue #945, opened 2026-08-15
jq: 'as $x' variable binding drops a document literal's source-text fidelity


newhoggy:
**Severity: Low**

## Summary

Found while implementing #939 (which fixes `keys`/`.[]`-style operations losing a document-sourced overflow literal's real text via `OwnedValue::to_json_for_reindex`'s round-trip sentinel). `as \$x` variable binding has a *broader*, separate version of the same symptom: it drops a `NumberLiteral`'s source-text fidelity even for perfectly ordinary **finite** values, not just overflow ones.

## Repro

```
$ echo '1.50' | succinctly jq -c 'try (.[]) catch .'
"Cannot iterate over number (1.50)"

$ echo '1.50' | succinctly jq -c 'try (. as $x | $x[]) catch .'
"Cannot iterate over number (1.5)"
```

Direct `.` access preserves the document's own trailing zero (matching real jq); routing the same value through `. as \$x | \$x` first loses it. The same divergence reproduces for #939's overflow case too — `. as \$x | \$x | keys` still shows the stale `1E+999` sentinel text even after #939's own fix, while direct `keys` (no `as \$x`) now shows the correct reformatted literal.

## Root cause

Not fully diagnosed. `eval_owned_input`'s own reindex step already correctly calls `to_json_for_reindex` (not `to_json`), so by the time *that* function's own reindex runs, the literal text is already gone — meaning whatever evaluates the `as \$x` binding itself must be collapsing the `NumberLiteral` into a plain `Int`/`Float` (dropping the source text, the same operation `OwnedValue::into_plain_number()`/similar "computed value" paths already do deliberately elsewhere) before the bound variable is ever stored, rather than at the reindex step this issue's investigation was focused on.

## Why this is separate from #939

#939 was specifically about `to_json_for_reindex`'s own sentinel substitution discarding a `NumberLiteral`'s text before `keys`/`.[]`/`reduce`/`with_entries` reach `describe()` - fixed there by reusing the literal's own text when it's JSON-number-syntax-safe. This issue is upstream of that: `as \$x` apparently loses the `NumberLiteral` entirely (not just its text during a reindex round-trip) at the point the variable is bound, so #939's own fix has nothing to work with once a value has passed through `as \$x` - the gap reproduces for ordinary finite values too, which #939 never touched at all.

## Suggested fix direction

[...]

Issue #561, opened 2026-08-01
jq: overflowed NumberLiteral renders as garbage in @uri/@html/@sh/tostring instead of inf/-inf/NaN


newhoggy:
## Summary

A `NumberLiteral` whose parsed value overflows to `inf`/`-inf`/`NaN` renders as garbled text in the non-JSON text formats, instead of `"inf"`/`"-inf"`/`"NaN"` the way a plain `Float` already does.

```console
$ echo '1e400' | sjq '@uri'
"NaNE%2B2147483647"
$ echo '1e400' | sjq '@html'
"NaNE+2147483647"
$ echo '1e400' | sjq 'tostring'
"NaNE+2147483647"
```

Expected (matching how `OwnedValue::Float(f64::INFINITY)` already renders in these same code paths):

```console
$ echo '1e400' | sjq '@uri'
"inf"
```

## Root Cause

`OwnedValue::number_str()` (`src/jq/value.rs`) documents that it *deliberately* does not special-case NaN/Infinity:

> Deliberately does not special-case NaN/Infinity -- callers that need JSON's "null" substitution (`to_json`, the error-preview streamer) check that themselves before falling back to this, e.g. via `self.as_f64().is_some_and(f64::is_nan)`.

`to_json()` follows that contract -- it checks `f.is_nan() || f.is_infinite()` on the `NumberLiteral(NumberRepr::Float(f), _)` arm before falling back to `number_str()`/`format_number_jq_compat()`. But the non-JSON text-format call sites never added the same check, so an overflowed literal falls straight into `format_number_jq_compat()` re-parsing the raw source text (e.g. `"1e400"`), which was never designed for that input and produces garbage.

Affected call sites (all in `src/jq/eval.rs` / `src/jq/eval_generic.rs`):
- `owned_to_string` (string interpolation)
- `format_uri` (`@uri`)
- `format_html` (`@html`)
- `shell_quote_value` / `format_sh` (`@sh`)
- `builtin_tostring` (`tostring`)
- the CSV/TSV/DSV cell formatter's fallback to `owned_to_string`
- `eval_generic`'s mirror of the `tostring`-style number branch

## Notes

- Discovered incidentally while working on #124 (SIMD escape scanning for jq format functions) -- unrelated to that issue's actual scope, filing separately per the existing pattern for incidentally-discovered `NumberLiteral` bugs (see #472, #488).
[...]

Issue #472, opened 2026-07-29
jq: nan does not survive as a number (round-trips to null via JSON)


newhoggy:
## Summary

While confirming the fix in #421, `nan` turns out not to stay a NaN *number*
in either evaluator:

```console
$ echo 'null' | jq  -c 'nan | type'                    # "number"
$ echo 'null' | sjq -c 'nan | type'                    # "null"

$ echo 'null' | jq  -c 'nan | isnan'                   # true
$ echo 'null' | sjq -c 'nan | isnan'                   # false

$ echo 'null' | jq  -c '[nan] | sort | .[0] | isnan'   # true
$ echo 'null' | sjq -c '[nan] | sort | .[0] | isnan'   # false
```

The `nan` builtin does produce `OwnedValue::Float(NaN)` -- `eval.rs`'s own
`test_infinite_nan` asserts it -- but a value that round-trips through JSON
comes back as an actual `null`, since that is how jq renders NaN and
succinctly re-reads its own rendering. jq prints NaN as `null` while keeping
it a number internally.

This also explains why constructing a fresh array containing 2+ NaNs and
then materializing it through JSON text collapses them: JSON has no NaN
literal, so each NaN becomes a genuine, mutually-`Equal` `Null` *before*
`compare_values` ever runs.

```console
$ echo 'null' | jq  -c '[nan,nan] | unique'     # [null,null]
$ echo 'null' | sjq -c '[nan,nan] | unique'     # [null]

$ echo 'null' | jq  -c '[nan,1,nan] | unique'   # [null,null,1]
$ echo 'null' | sjq -c '[nan,1,nan] | unique'   # [null,1]
```

Two more `bsearch` rows are explained by the same mechanism:
`[nan,1] | bsearch(nan)` gives `1` here against jq's `-2`, and
`[nan] | bsearch(nan)` gives `-2` against jq's `-1` -- the constructed array
element became `null` while the search argument stayed NaN, so the two
compare across type ranks (`null` < `number`) rather than as numbers.

## Relationship to #421

Different mechanism from the comparator bug #421 fixed (that was
`compare_values` folding NaN comparisons incorrectly; this is NaN losing its
number-ness on the way through a JSON round trip). #421's fix pins the
current, still-wrong answer to the `unique`/`bsearch` cases above as a
documented known-divergence test
(`test_nan_container_ordering_known_divergence_421` in
`tests/jq_evaluator_parity_tests.rs`) rather than leaving them unpinned, so
[...]

PR #946, opened 2026-08-15
fix(jq): reuse a JSON-safe overflow literal's own text through reindex (#939)

### LINKED ISSUE #939: jq: keys/.[] on a document-sourced overflow literal previews the internal reindex sentinel, not the source text

newhoggy:
## Summary

- `OwnedValue::to_json_for_reindex` (`src/jq/value.rs`) substituted a generic `"1e999"`/`"-1e999"` sentinel for **every** infinite `NumberLiteral`, unconditionally discarding the literal's own source text — so once `keys`/`.[]`/`reduce`/`foreach`/`with_entries` routed a document-sourced overflow literal (e.g. `123e400`) through this round-trip bridge before an error could be raised, #930's `describe()` fix had nothing left to reformat: every overflow literal previewed identically (`1E+999`) regardless of its actual digits.
  ```
  $ echo '123e400' | jq -c 'try keys catch .'
  "number (1.23E+402) has no keys"
  $ echo '123e400' | succinctly jq -c 'try keys catch .'   # before this fix
  "number (1E+999) has no keys"
  ```
- A document-sourced literal's raw text is, by construction, already valid JSON number syntax that reparses to the exact same value — so `to_json_for_reindex` now just reuses it directly (via a new `is_json_number_literal` check) instead of substituting the disconnected sentinel, letting `describe()` show the real text and matching real jq exactly.
- YAML's own special `.inf`/`-.inf` tokens are **not** JSON-number-syntax-shaped (writing `.inf` back out as JSON text to reparse would itself be invalid) — `is_json_number_literal` keeps those on the existing sentinel path, completely unchanged, since there's no jq equivalent to match against for that shape anyway.
- NaN is unaffected either way: `NAN_SENTINEL` is deliberately unparseable as `f64` by design (#472), so there was never any literal text to reuse for it regardless of origin.

## Follow-up filed

Found and filed **#945** while verifying this more broadly: `as $x` variable binding drops `NumberLiteral` fidelity even for ordinary **finite** values (`1.50` → `1.5`), which also defeats this fix specifically when a value passes through a binding first (`. as $x | $x | keys` still shows the stale sentinel). That's a separate, broader, upstream gap — not expanding this fix's scope to chase it.

## Test plan

- [x] Diffed output directly against pinned real `jq`/`yq` for both the JSON-overflow case (now matches exactly) and the YAML `.inf`/`-.inf` case (confirmed unchanged, still the sentinel — no jq equivalent to match)
- [x] Unit tests for `is_json_number_literal` and `to_json_for_reindex`'s new/unchanged branches (`src/jq/value.rs`)
- [x] `query!`-based tests for the `reduce`/`foreach`/`with_entries` reindex bridges (`src/jq/eval.rs`), mirroring #561's existing bridge test
- [x] CLI-level tests for both the JSON overflow case (`tests/jq_cli_tests.rs`) and the YAML special-float case staying on the sentinel (`tests/yq_cli_tests.rs`)

Closes #939

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -27230,6 +27230,45 @@ mod tests {
         );
     }
 
+    /// #939: `to_json_for_reindex` (the same bridge #561's test above
+    /// exercises) substituted a generic `1e999`/`-1e999` sentinel for *any*
+    /// infinite `NumberLiteral`, including a document-sourced overflow
+    /// literal that already had real source text of its own - so once
+    /// `keys`/`.[]`-style operations routed the value through this bridge,
+    /// #930's fix (reformatting a `NumberLiteral`'s own text) had nothing
+    /// left to work with, and every overflow literal previewed as the same
+    /// sentinel-derived text regardless of its actual digits. Oracle-
+    /// verified against real jq 1.7.1: a document literal's own text is
+    /// reused directly (matching jq exactly) whenever it's already valid
+    /// JSON number syntax, since it's guaranteed to reparse to this same
+    /// value either way.
+    #[test]
+    fn test_document_overflow_literal_preserves_own_text_through_reindex_bridges_939() {
+        query!(br"123e400", "reduce (1) as $x (.; .) | keys",
+            QueryResult::Error(e) => {
+                assert_eq!(e.message, "number (1.23E+402) has no keys");
+            }
+        );
+
+        query!(br"-1e400", "reduce (1) as $x (.; .) | .[]",
+            QueryResult::Error(e) => {
+                assert_eq!(e.message, "Cannot iterate over number (-1E+400)");
+            }
+        );
+
+        query!(br#"{"a":12.34e400}"#, "with_entries(.value |= (. | keys))",
+            QueryResult::Error(e) => {
+                assert_eq!(e.message, "number (1.234E+401) has no keys");
+            }
+        );
+
+        query!(br"0.5e400", "foreach (1) as $x (.; .) | keys",
+            QueryResult::Error(e) => {
+                assert_eq!(e.message, "number (5E+399) has no keys");
+            }
+        );
+    }
+
     #[test]
     fn test_builtin_tonumber() {
         query!(br#""42""#, "tonumber",
--- a/src/jq/stream.rs
+++ b/src/jq/stream.rs
@@ -1035,13 +1035,13 @@ mod tests {
     /// #930: a `NumberLiteral` carries its document source text, so an
     /// overflowed one gets jq's literal-preserving text instead of `DBL_MAX`
     /// text - matching how real jq distinguishes an echoed literal from a
-    /// computed value. `format_number_jq_compat` is exercised directly
-    /// (rather than via a CLI-level `keys`/`.[]` probe) because those
-    /// operations round-trip an input document's value through
-    /// `OwnedValue::to_json_for_reindex` before an error can reach this
-    /// code, which substitutes its own `1e999` round-trip sentinel for the
-    /// original literal text (pre-existing, unrelated to this fix - see the
-    /// `#930` follow-up filed for it).
+    /// computed value. `format_number_jq_compat` is exercised directly here
+    /// rather than via a CLI-level `keys`/`.[]` probe purely to keep this a
+    /// focused unit test of the formatting logic on its own - `keys`/`.[]`
+    /// do reach this same text today (`OwnedValue::to_json_for_reindex`
+    /// reuses a document-sourced overflow literal's own text rather than
+    /// substituting its `1e999` round-trip sentinel, fixed by #939; see
+    /// `tests/jq_cli_tests.rs`'s CLI-level coverage of that bridge).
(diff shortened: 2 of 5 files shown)


newhoggy:
Code review found two real issues in the initial commit, both fixed as a follow-up commit:

1. **DoS-shaped performance regression**: reusing the document literal's own text unconditionally turned an O(1) sentinel substitution into O(literal length), repeated every time `reduce`/`foreach`'s non-fast-pathed update step re-runs its reindex bridge over an unchanged accumulator. Reviewer measured ~700x slowdown on a 300KB literal in a 20K-iteration loop against unmodified `main`. Fixed by capping literal reuse to 256 bytes (comfortably above any realistic overflow literal — reaching `f64::MAX` needs on the order of ~300 exponent digits at most) and falling back to the O(1) sentinel beyond that. Independently reproduced the reviewer's exact scenario against the fix — now completes in under a second regardless of iteration count.

2. **Unreachable shape-check**: my original `is_json_number_literal` guard (distinguishing JSON-syntax-safe overflow literals from YAML's `.inf`/`-.inf`) turned out to defend a case that can't currently occur — `OwnedValue::NumberLiteral` only ever comes from JSON parsing (`document.rs`'s `number_literal()` defaults to `None`; only `json/light.rs` overrides it, so YAML's special float tokens always become a plain `Float`, never reaching the branch I added). Removed the now-unnecessary check (and the duplicated `looks_like_number`-style predicate it introduced) in favor of just the length bound, and retitled the YAML CLI test to describe what it actually verifies (that path is unaffected by this fix) rather than implying it exercised the removed branch.

Also fixed a stale test comment (in `stream.rs`) that described the reindex sentinel substitution as still-unresolved — this PR is that resolution.

All local checks green (full test suite, both clippy invocations) after these fixes and a rebase onto latest `main`.

PR #951, opened 2026-08-15
fix(yaml): preserve integer-valued float scalars through YAML parsing

### LINKED ISSUE #918: yq: an integer-valued float scalar (2.0) loses its float type during YAML input parsing, unlike real yq
**Severity: Medium**

## Summary

An integer-valued YAML float scalar (`2.0`, `-5.0`, etc.) parsed as *input* silently loses its float-ness somewhere in the YAML-to-`OwnedValue` conversion path in `succinctly yq` mode — internally it becomes indistinguishable from the plain integer `2`. Real yq (mikefarah/yq, pinned `v4.53.3`) keeps the two genuinely distinct types, unlike jq (where `2.0 == 2` is `true` — jq has no strict int/float type distinction at all).

Found while fixing #909 (`has()`/`in()`'s float-key handling): a test using `2.0` as YAML input to exercise the array-key-Float-rejection path unexpectedly answered `true` instead of `false`, tracing back to this normalization rather than a bug in #909's own fix.

## Repro (verified against pinned yq v4.53.3)

```
$ echo '2.0' | yq '. == 2'
false
$ echo '2.0' | succinctly yq '. == 2'
true

$ echo '2.0' | jq '. == 2'
true
```

Real yq keeps `2.0` and `2` as distinct, non-equal types; succinctly's yq mode collapses them to equal, matching jq's own (looser) numeric equality instead of yq's (stricter) one.

A non-integer-valued float doesn't exhibit this — it's preserved correctly:
```
$ echo '2.5' | succinctly yq '. == 2.5'
true
```
(Direct inspection during #909's investigation confirmed `2.5` parses to `OwnedValue::NumberLiteral(Float(2.5), "2.5")` as expected, while `2.0` parses to a bare `OwnedValue::Int(2)` — the float marker is lost specifically when the value has no fractional part.)

Downstream effect also confirmed via #909's own investigation: `.a | has($x)` with `$x` bound from a YAML-input `2.0` treats the key as the integer `2` (matching an in-bounds array index) rather than correctly failing to match any array index the way a genuine float key does in yq mode.

## Why this is separate from #909

#909 fixed `has()`/`in()`'s own array-key-arm logic (adding Float/NaN support, matching real jq, and — after review — correctly *excluding* Float from yq-mode array indexing to match real yq). This issue is one level upstream: the YAML scalar `2.0` shouldn't have been converted into `OwnedValue::Int(2)` in the first place, regardless of what any particular builtin then does with it. #909's own fix works around this correctly for `has()`/`in()`'s array-key path specifically (by not truncating Float in yq mode at all), but the root normalization bug still affects every *other* operation on such a value (equality, arithmetic, output formatting round-trip, etc.).

## Suggested fix direction

Needs investigation into where in the YAML parsing/cursor pipeline (`src/yaml/`) or the `to_owned`/`OwnedValue`-construction path (`src/jq/eval.rs` or `src/jq/value.rs`) an integer-valued float scalar's `NumberRepr::Float` marker gets dropped in favor of `NumberRepr::Int`/`OwnedValue::Int`. Given this codebase's own documented care around type preservation (CLAUDE.md's P10 "Type Preservation" optimization note: `"1.0"` stays a string, not a number — a similar-shaped bug already fixed once for quoted strings), this may be the same class of gap resurfacing for *unquoted* float-looking-like-int scalars specifically.

newhoggy:
## Summary

- YAML's `DocumentValue` impl had no `number_literal()` override (unlike JSON's), so any YAML float scalar always fell through to `as_f64`, producing a bare `OwnedValue::Float` whose `Display`-based formatting drops the decimal point on a whole number (`2.0` becomes indistinguishable from a genuine `2` once materialized, e.g. via `has()`/`in()`).
- The new override only fires for a **finite** float whose literal text is independently confirmed to be valid JSON number syntax (`is_json_number_syntax`, added in `src/yaml/scalar.rs`): YAML's plain-scalar grammar accepts spellings JSON's doesn't (leading-dot `.5`, hex/octal `0x2A`/`0o52`), and echoing those into the internal JSON-reindex bridge used by builtins like `tag` (which has no native cursor-aware implementation) silently misclassified them as parse errors instead of numbers — discovered live via the `scalar_tag_resolution` golden fixture regressing.
- `Int` is excluded entirely: `resolve_plain`'s own doc comment already warns its numeric variants' source text "cannot be re-parsed... emitters must use the carried value" (hex/octal), and plain decimal ints never lose information through the existing `as_i64` path, so there's no bug there to fix.
- All finite-float behavior verified live against the pinned `yq` oracle (v4.53.3), including confirming real yq *also* normalizes a leading-dot float (`.5` → `0.5`) once forced through materialization — validating that excluding it from literal-preservation isn't just a defensive JSON-safety measure but the actually-correct output.

Fixes #918

## Test plan

- [x] New regression tests: integer-valued float never matches via `in()`/`has()` (#918), materialized float literal fidelity through `[.]` (`2.0`→`[2.0]`, `.5`→`[0.5]`), `tag` on a leading-dot float (`!!float`, not `!!null`)
- [x] Updated stale #909 test comment now that #918 is fixed
- [x] `yq_golden_conformance` (`scalar_tag_resolution` fixture) passes

### DIFF
--- a/docs/compliance/yaml/1.2.md
+++ b/docs/compliance/yaml/1.2.md
@@ -333,7 +333,6 @@ yq v4.53.3 with `-o=json`):
 | `-0x2A` (signed hex)   | `"-0x2A"`      | Error          | 1.2 core forbids signs on based ints       |
 | `0xFFFFFFFFFFFFFFFF`   | string         | Error          | Overflows `i64`; yq errors on JSON output  |
 | `-0`                   | `0`            | `-0.0`         | go-yaml quirk (signed-zero float)          |
-| `18.0` (compact `-I=0`)| `18`           | `18.0`         | Float formatting, tracked in #169          |
 
 The `Error` rows are yq resolving a scalar as `!!int`/`!!float` and then failing to
 serialize it (`strconv` rejects the source text) — byte-parity there is impossible, so
--- a/src/jq/document.rs
+++ b/src/jq/document.rs
@@ -283,10 +283,14 @@ pub trait DocumentValue: Sized + Clone {
     /// (`1e100`, `1.0`, `-0.0`) instead of re-rendering the parsed value —
     /// see issue #387.
     ///
-    /// Defaults to `None`: only JSON overrides this today. YAML numbers are
-    /// resolved plain scalars (hex/octal/underscore forms `yq` may or may not
-    /// echo back verbatim), which is a separate judgment call this default
-    /// deliberately leaves alone.
+    /// Defaults to `None`. JSON overrides this unconditionally (every
+    /// `Number` token is JSON-legal number syntax by construction). YAML
+    /// overrides it too (#918), but only for a finite float whose source
+    /// text is independently confirmed safe and worthwhile to echo — see
+    /// `YamlValue::number_literal`'s doc comment (`src/yaml/light.rs`);
+    /// YAML's own plain-scalar grammar accepts spellings (hex/octal,
+    /// leading-dot) that aren't JSON-legal, and ints never lose information
+    /// through the bare-`Int` path, so both stay on this `None` default.
     fn number_literal(&self) -> Option<Cow<'_, str>> {
         None
     }
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -131,7 +131,21 @@ fn tagged_scalar_to_owned<V: DocumentValue>(tag: &str, value: &V) -> Option<Owne
         crate::yaml::ResolvedScalar::Null => OwnedValue::Null,
         crate::yaml::ResolvedScalar::Bool(b) => OwnedValue::Bool(b),
         crate::yaml::ResolvedScalar::Int(n) => OwnedValue::Int(n),
-        crate::yaml::ResolvedScalar::Float(f) => OwnedValue::Float(f),
+        // Mirrors YamlValue::number_literal (src/yaml/light.rs) for the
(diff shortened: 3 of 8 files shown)


newhoggy:
Code review surfaced two additional gaps beyond the original PR body's scope, filed as follow-ups (both fixed/addressed in this PR where in-scope, both fully investigated):

- #953 — an i64-overflow decimal integer scalar (classified `Float` only because it exceeds `i64::MAX`) shows different rounded/unrounded output across query paths. **Confirmed pre-existing** via a pristine pre-#918 build — not a regression from this PR, and this PR's `is_preservable_float_literal` gate correctly declines to touch it (no literal `.` in the source text).
- #954 — a few YAML-legal-but-JSON-unsafe float spellings (`+2.0`, `1.`, `007e2`, and `!!float 5`'s int-widened case) still show #918's original symptom, since this PR's fix is deliberately scoped to the JSON-safe subset (echoing JSON-unsafe text into the internal reindex bridge was shown to break `tag` during review — see the `is_json_number_syntax`/`is_preservable_float_literal` doc comments).

Also fixed during review: the initial commit only patched the plain-scalar path (`YamlValue::number_literal`) — an explicit `!!float 2.0` tag converges on a separate materialization path (`tagged_scalar_to_owned`) that bypassed the fix entirely. Now covered, with a regression test (`test_materialized_yaml_explicit_float_tag_literal_fidelity_918`).

PR #964, opened 2026-08-15
fix(jq): hoist the duplicated ResolvedScalar->OwnedValue conversion

### LINKED ISSUE #907: jq: remaining tag-blind paths and duplicated ResolvedScalar->OwnedValue logic after #747/#903
## Summary

Follow-up from the #903 review (fix for #747). The review round fixed several confirmed tag-resolution bugs in the generic evaluator (alias dereferencing, `to_entries`/`reverse`/`pivot`/`shuffle` materialization, the `is*` builtin family, `to_owned_key_shape` for computed index keys/slice bounds, and several type-mismatch error messages). A few lower-priority items surfaced during that review that weren't fixed, either because they're unconfirmed (no live-failing repro found) or because they're pure code-quality concerns rather than correctness bugs:

## 1. `to_owned_with_comments`'s scalar leaf isn't tag-aware (unconfirmed)

`src/jq/eval_generic.rs`, `to_owned_with_comments`'s leaf case still calls plain `to_owned(value)` rather than a cursor-aware equivalent, even though `cursor: Option<&V::Cursor>` is in scope. This function backs the YAML *write* path (`yq_runner.rs`'s `evaluate_yaml_cursor`, used by `=`/`|=`/`del()` etc. for comment/style-preserving reconciliation).

Direct manual testing (`.a |= .`, `.b = .a`, `del(.b)` combined with an untouched tagged sibling) did **not** reproduce a wrong value — every case correctly materialized the tag-resolved value. It's possible some other path already resolves the tag before reaching this function, or the bug requires a more specific shape. Worth a closer audit rather than a blind fix.

## 2. `yq_runner.rs:988`/`:1022` still call tag-blind `to_owned` (unconfirmed)

`src/bin/succinctly/yq_runner.rs` has two call sites — `alias_sync_ctx`'s pristine-document snapshot (issue #711) and the `owned_with_comments` closure's `need_comments == false` branch — that still call `to_owned(&cursor.value())`/`to_owned(&c.value())` directly on a `YamlCursor` rather than the new `crate::jq::eval_generic::to_owned_cursor`. No concrete failing CLI repro was found in the time available — every non-M2-streamable expression shape tried already routed through `eval_generic.rs`'s own tag-aware materialization before reaching these lines — but the pattern contradicts the module comment's claim that "every reachable caller now checks the tag first." Worth an explicit audit.

## 3. Duplicated `ResolvedScalar -> OwnedValue` mapping (code quality, not a bug)

The same 5-arm match (`ResolvedScalar::{Null,Bool,Int,Float,Str} -> OwnedValue`) now exists in at least three places:
- `src/jq/eval_generic.rs`'s `tagged_scalar_to_owned` (added by #903)
- `src/bin/succinctly/yq_runner.rs:170-178`'s `resolved_scalar_to_owned`
- `src/jq/eval.rs:18131-18138`, inside `yaml_value_to_owned`'s local `resolved_scalar_to_owned` helper — in the *same lib crate* as the new copy, so the "can't share across the bin/lib boundary" rationale that applies to `yq_runner.rs`'s copy doesn't apply here

Consider hoisting a shared `ResolvedScalar::to_owned_value(text: &str) -> OwnedValue` (or similar) next to the existing `ResolvedScalar::type_name()`/`tag()` methods in `src/yaml/scalar.rs`, and having all three call sites use it. A future `ResolvedScalar` variant would otherwise need three synchronized edits.

## 4. `to_owned`/`to_owned_cursor` container-recursion duplication (code quality, not a bug)

`to_owned`'s and `to_owned_cursor`'s `as_object`/`as_array` recursion branches (`src/jq/eval_generic.rs`) are structurally identical except for `field.value` vs `field.value_cursor` and `elems.uncons()` vs `elems.uncons_cursor()`. A future `DocumentValue`/`DocumentField`/`DocumentElements` change applied to one loop and missed in the other would silently reintroduce a #747-style bug in whichever function is missed. No proposed fix — the two functions have genuinely different signatures (`&V` vs `&C`), so de-duplicating isn't a trivial extraction; flagging for awareness.

## Scope note

None of these four items are confirmed, live, user-visible correctness bugs — #1 and #2 are audit gaps with no reproduced failure, and #3/#4 are maintainability concerns. Lower priority than #747/#903's confirmed fixes, tracked here so they aren't lost.

newhoggy:
## Summary

#907's item 3 catalogued three independent copies of the same `ResolvedScalar -> OwnedValue` match arm:
- `src/jq/eval_generic.rs`'s `tagged_scalar_to_owned`
- `src/bin/succinctly/yq_runner.rs`'s `resolved_scalar_to_owned`
- `src/jq/eval.rs`'s local helper inside `yaml_value_to_owned`

Collapsed into one `ResolvedScalar::to_owned_value` (`src/yaml/scalar.rs`), used by all three call sites.

**This wasn't just a duplication cleanup.** #918's literal-preservation fix for integer-valued floats (`2.0` losing its decimal point) only landed in `eval_generic.rs`'s copy — the other two silently missed it:
- `yq_runner.rs`'s copy backs `--slurp`/`--eval-all`'s DOM materialization path
- `eval.rs`'s copy backs the `load()` builtin

Both still showed the original #918 symptom (`2.0` → `2`) until this fix.

**A fourth, previously uncatalogued duplicate surfaced during investigation**: `yq_runner.rs`'s `standard_json_to_owned` hand-rolled its own JSON `Number -> OwnedValue` conversion via `as_i64()`/`as_f64()`, completely bypassing `DocumentValue::number_literal()` — unlike the crate's own `to_owned()`, which `StandardJson` already implements `DocumentValue` for. This is what actually caused `--slurp`/`--eval-all`'s JSON-round-trip evaluation bridge (`evaluate_input`) to still show the bug even after the `ResolvedScalar` fix alone. Replaced with a direct call to the existing `to_owned()`.

#907's items 1 and 2 (unconfirmed "tag-blind path" audit gaps in `to_owned_with_comments` and two other `yq_runner.rs` call sites) were investigated again with different targeted test cases (an explicit tag that actually changes resolution, e.g. `!!str 5`, combined with an untouched sibling through `.b |= . + 1` / `del()` under `-o json`) and remained unreproducible — matching the original author's own finding. Left as-is rather than fixed speculatively, per this project's convention against unverified defensive fixes.

Item 4 (`to_owned`/`to_owned_cursor` container-recursion duplication) has no proposed fix per the issue itself; not addressed here either.

Fixes #907

## Test plan

- [x] `cargo doc --no-deps --all-features` (RUSTDOCFLAGS=-D warnings)
- [x] New regression test `test_integer_valued_float_survives_slurp_eval_all_load_907` covering `--slurp`, `--eval-all`, and `load()`
- [x] Live-verified fix via direct CLI testing before/after (confirmed `--slurp '.'` went from `[2]` → `[2.0]`, `--eval-all '.'` from `2` → `[2.0]`, `load()` from `2` → `2.0`)

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -11,7 +11,8 @@ use std::path::Path;
 
 use succinctly::jq::document::{DocumentCursor, DocumentFields, IndentSpec};
 use succinctly::jq::eval_generic::{
-    eval_with_cursor_using, to_owned, to_owned_with_comments, CommentTree, GenericResult,
+    eval_with_cursor_using, to_owned as generic_to_owned, to_owned_with_comments, CommentTree,
+    GenericResult,
 };
 use succinctly::jq::{
     self, sync_aliased_paths, Builtin, EvalError, Expr, OwnedValue, QueryResult, YqSemantics,
@@ -406,7 +407,7 @@ fn parse_input(bytes: &[u8], format: InputFormat) -> Result<Vec<OwnedValue>> {
             // Parse as JSON
             let index = JsonIndex::build(bytes);
             let cursor = index.root(bytes);
-            Ok(vec![to_owned(&cursor.value())])
+            Ok(vec![generic_to_owned(&cursor.value())])
         }
         InputFormat::Yaml | InputFormat::Auto => {
             // Parse as YAML (Auto defaults to YAML when no extension hint)
@@ -553,9 +554,9 @@ fn query_result_to_owned_values(
     sink: &mut ErrorSink,
 ) -> Vec<OwnedValue> {
     match result {
-        QueryResult::One(v) => vec![to_owned(&v)],
-        QueryResult::OneCursor(c) => vec![to_owned(&c.value())],
-        QueryResult::Many(vs) => vs.iter().map(to_owned).collect(),
+        QueryResult::One(v) => vec![generic_to_owned(&v)],
+        QueryResult::OneCursor(c) => vec![generic_to_owned(&c.value())],
+        QueryResult::Many(vs) => vs.iter().map(generic_to_owned).collect(),
         QueryResult::None => vec![],
         QueryResult::Error(e) => {
             sink.report(DiagStyle::Yq, &e, &no_location());
@@ -967,8 +968,13 @@ fn evaluate_yaml_cursor<W: AsRef<[u64]> + Clone>(
     // assignment-family expression against a document that actually has
     // aliases. Everything else (JSON, plain reads, alias-free YAML) pays
     // nothing beyond this one bool check.
-    let alias_sync_ctx = (is_alias_sensitive_assign(expr) && cursor.index().has_aliases())
-        .then(|| (to_owned(&cursor.value()), collect_alias_groups(cursor)));
+    let alias_sync_ctx =
+        (is_alias_sensitive_assign(expr) && cursor.index().has_aliases()).then(|| {
+            (
+                generic_to_owned(&cursor.value()),
+                collect_alias_groups(cursor),
+            )
+        });
 
     // Snapshot the pristine presentation tree *before* evaluation too
     // (#739, ADR-0017): same shape-preserving-write gate as `alias_sync_ctx`
@@ -1002,7 +1008,7 @@ fn evaluate_yaml_cursor<W: AsRef<[u64]> + Clone>(
         if need_comments {
             to_owned_with_comments(&c.value(), Some(c))
         } else {
-            no_comments(to_owned(&c.value()))
+            no_comments(generic_to_owned(&c.value()))
         }
     };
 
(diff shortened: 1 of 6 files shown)


newhoggy:
Multi-round `/code-review` (10 finder angles, several passes) found and this PR now fixes:

- **A real no_std break**: unconditional `use std::borrow::Cow;` in `scalar.rs` — fixed by matching `light.rs`'s existing `#[cfg(not(test))] alloc::borrow::Cow` / `#[cfg(test)] std::borrow::Cow` pattern. Confirmed via `cargo check --no-default-features`.
- **A `ToOwned` naming footgun**: `yq_runner.rs`'s bare `to_owned` import (now aliased `to_owned as generic_to_owned`, matching the sibling `jq_runner.rs`'s existing convention) — `StandardJson`/`YamlValue` derive `Clone`, so an unaliased import risks a future `.to_owned()` method-call rewrite silently invoking `Clone`'s blanket `ToOwned` instead.
- **Two doc/test-comment overclaims**: `to_owned_value`'s doc comment claimed to be "the single source of truth" for every `ResolvedScalar` match in the crate — `light.rs`'s `write_resolved_scalar_as_json`/`stream_resolved_scalar_as_json` (JSON-text output, not `OwnedValue`) are two more, unfixed here. A test comment conflated the unrelated `standard_json_to_owned`/`StandardJson` duplicate with the `ResolvedScalar` mapping it happened to share a bug class with. Both corrected; the remaining duplicates filed as #965.
- **A genuine correctness improvement, verified against the oracle**: replacing `standard_json_to_owned` also changes `--slurp`/`--eval-all --input-format json`'s exponent-notation number formatting (`1e2` now preserves as `1E+2` instead of normalizing to `100.0`) — confirmed live that real `jq` also renders `1e2` as `1E+2`, so this brings that path into line with jq's own literal-preservation behavior rather than introducing a divergence. Pinned with a new regression test.

Also investigated and **confirmed pre-existing, not caused by this PR**: review flagged that leniently-parsed-but-RFC-8259-invalid JSON numbers (`007`, `1.2.3`) now echo verbatim / fall back to `0` instead of `null` through the new `--slurp`/`--eval-all` path. Reproduced the identical behavior directly against a pristine `main` build via plain `succinctly jq '[.a]'` (no `yq`, no `--slurp`, nothing this PR touches) — this PR made `yq`'s materialization *consistent* with the rest of the codebase's already-established (if imperfect) behavior, rather than introducing it. Filed as #966 for proper investigation/fix.

PR #969, opened 2026-08-15
fix(jq): validate strptime's month/day before array-indexing them

### LINKED ISSUE #968: jq: strptime panics on out-of-range month (00 or 13+) instead of erroring
**Severity: Medium**

## Summary

`parse_strptime`'s yearday computation (`src/jq/eval.rs`, the `month_days[(month - 1) as usize]` indexing) has no bounds check on the parsed `month` field before indexing a fixed 12-element array. An out-of-range month from `%m`/`%D`/`%F` (`0`, or `13`+) reaches this line unchecked and panics the process, rather than reporting a parse error the way real jq does.

Found during #912's review (a pure-refactor PR that extracted this same code into a shared `yearday_for_civil` helper) — confirmed both that the panic reproduces identically on unmodified `main` (pre-#912, so not a regression from that PR) and that it's untouched by #912's own diff (only the *weekday* computation two lines above was refactored there).

## Repro

```
$ echo '"2024-00-15"' | succinctly jq 'strptime("%Y-%m-%d")'
thread 'main' panicked at src/jq/eval.rs:...: index out of bounds: the len is 12 but the index is 18446744073709551615
(month=0: `(0 - 1) as usize` wraps to usize::MAX)

$ echo '"2024-13-15"' | succinctly jq 'strptime("%Y-%m-%d")'
thread 'main' panicked at src/jq/eval.rs:...: index out of bounds: the len is 12 but the index is 12
```

Real jq 1.7.1:
```
$ echo '"2024-00-15"' | jq 'strptime("%Y-%m-%d")'
jq: error (at <stdin>:1): date "2024-00-15" does not match format "%Y-%m-%d"
```
(exit 5, a normal catchable error — no crash, no adversarial `i64::MAX` needed, just an ordinary malformed date string)

## Root cause

`parse_strptime` parses `month`/`day` via `parse_digits` with no range validation (a 2-digit `%m` accepts `00`-`99` verbatim), then both the weekday computation (via `checked_days_from_civil`, which does *arithmetic* bounds-checking but never validates that `month` is actually in `1..=12`) and the yearday computation (`month_days[(month - 1) as usize]`) proceed unchecked. The array index is where it actually panics, but the missing validation is upstream of both.

## Why this is separate from #912

#912 was explicitly scoped to pure deduplication/consistency cleanup with no behavior change (confirmed via byte-identical output testing against real jq and the pre-refactor build across every code path it touched). This is a genuine, pre-existing correctness bug — real jq rejects this input as a normal parse error, succinctly crashes the process — not a duplication or style concern, and fixing it would be a behavior change #912 deliberately avoided making.

## Suggested fix direction

Validate `month` (and likely `day`, similarly unchecked against the calendar) is in range before either the weekday or yearday computation in `parse_strptime`, returning `Err("date \"...\" does not match format \"...\"")` (or similar, matching real jq's wording family) instead of proceeding — consistent with how `parse_strptime`'s other malformed-input paths already report a parse error rather than panicking.

newhoggy:
## Summary

`parse_strptime` parsed `%m`/`%d`/`%e`/`%D`/`%F` with no range check (`parse_digits` just accepts `00`-`99`), so an out-of-range month reaching `month_days[(month - 1) as usize]` panicked the process:
- month=0: `(0 - 1) as usize` wraps to `usize::MAX`
- month=13+: a plain out-of-bounds index

Real jq rejects both as an ordinary catchable parse error instead of crashing.

## Fix

Validates `month` is `1..=12` and `day` is `0..=31` before the weekday/yearday computation, matching real jq's own boundary — confirmed live against jq 1.7.1, including the surprising but genuine acceptance of day `00` (`2024-01-00` succeeds; `2024-01-32` errors).

The caller (`builtin_strptime`) already discards `parse_strptime`'s specific error text in favor of `EvalError::strptime_no_match`'s jq-matching wording (`date "..." does not match format "..."`), so any `Err(..)` returned here produces the correct final message — no new error-formatting logic needed.

A second `month_days[(month - 1) as usize]` site exists (`broken_down_time_from_unix_secs`), but its `month` is derived algorithmically from a mathematically-bounded civil-date computation (Howard Hinnant's algorithm), never from unchecked user input — confirmed it can't produce an out-of-range value, so it needs no fix.

Fixes #968

## Test plan

- [x] New regression test `test_strptime_out_of_range_month_or_day_errors_instead_of_panicking_968` covering month 0/13+, day 32+, day 00 (accepted), a valid date, and `%D`'s separate parsing arm
- [x] Live-verified every case against the pinned jq oracle (exit codes and output match exactly)

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -18555,8 +18555,8 @@ fn parse_strptime(input: &str, fmt: &str) -> Result<BrokenDownTime, String> {
     // instead of a hand-duplicated, unchecked copy of the same formula.
     // Its `i64`, checked arithmetic throughout is also what fixes #968's
     // second panic: `day == 0` needs this formula's `doy` term to go
-    // momentarily negative for March specifically, which a `u32` (the
-    // pre-#912 hand-duplicated copy's type) couldn't represent without its
+    // momentarily negative for March specifically, which the pre-#912
+    // hand-duplicated copy's `u32` type couldn't represent without its
     // subtraction underflowing.
     let days = checked_days_from_civil(year, month, day).map_err(|e| e.to_string())?;
     weekday = (days % 7 + 4 + 7) % 7;
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -18532,9 +18532,32 @@ fn parse_strptime(input: &str, fmt: &str) -> Result<BrokenDownTime, String> {
         }
     }
 
+    // `%m`/`%d`/`%e`/`%D`/`%F` parse up to 2 digits with no range check of
+    // their own (`parse_digits` just accepts `00`-`99`), so an out-of-range
+    // month reaching `yearday_for_civil`'s `month_days[(month - 1) as
+    // usize]` below panics (month=0 wraps `(0 - 1) as usize` to
+    // `usize::MAX`; month=13+ is a plain out-of-bounds index) - matching
+    // real jq's own rejection of month 0 and 13+, and day > 31 (confirmed
+    // live against jq 1.7.1: `2024-01-32` errors). `checked_days_from_civil`
+    // (#912) doesn't validate this itself - it only guards against actual
+    // `i64` overflow, which an out-of-semantic-range month/day doesn't
+    // trigger - so this check is still required even after that refactor.
+    // `day`'s lower bound is never checked: it's only ever assigned from
+    // `parse_digits`, which builds its output from `is_ascii_digit()`
+    // characters alone, so it can never be negative (`day == 0` is real,
+    // jq-valid input - `2024-01-00` succeeds) (#968).
+    if !(1..=12).contains(&month) || day > 31 {
+        return Err(format!("month {month} or day {day} out of range"));
+    }
+
     // Calculate weekday from days-since-epoch, via the same checked
     // days-from-civil math `mktime`/`strftime`'s %s specifier use (#912)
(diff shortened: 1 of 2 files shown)


newhoggy:
Multi-round `/code-review` (four independent finder agents) found a second, distinct panic the first commit's month/day range check didn't close: the *weekday* computation (a separate Howard Hinnant `days_from_civil` formula, `u32`-typed) assumes `day >= 1`, but `day == 0` is real, jq-valid input the range check deliberately keeps accepting. For March specifically (the one month where the formula's leading term truncates to exactly `0`), `day == 0` made `0u32 - 1` underflow — confirmed live, `"2024-03-00"` still panicked even after the array-index fix alone. Fixed by widening that computation's intermediate types from `u32` to `i64`; verified by sweeping all 12 months' day-0 case (and a broader valid-date sweep) against the pinned jq oracle.

Also dropped the day range check's vacuous `0..=` lower bound (day can never be negative — it's only ever assigned from `parse_digits`, which can't produce one).

While sweeping a broader date range for regressions, found and filed two out-of-scope follow-ups:
- #970 — a pre-existing, unrelated weekday-computation bug for century-non-leap years (1900, 2100, 2200, 2300, ...), reproduced identically on a pristine pre-#968 `main` build.
- #971 — three lower-priority strptime gaps (unvalidated `%H`/`%M`, duplicated `month_days` indexing between `parse_strptime` and `broken_down_time_from_unix_secs`, silent `%b`/`%B`/`%h` no-match).

Issue #972, opened 2026-08-15
jq: first()/limit() don't truncate the keep-partial path resolvers, causing a spurious trailing error after correct partial output


newhoggy:
> ⚠️ **Attempted and reverted (PR #985, not merged) — see the closing comment on that PR and the comment below for the full writeup.** The suggested fix direction below ('shared primitive, few-line fix') is unsafe: it relies on an invariant (`PathResolveResult`'s `Err` prefix is always exactly what jq's generator would have emitted before the escape) that `resolve_index_expr`/`resolve_slice_expr`/multi-stage `resolve_seq` violate, causing silent data destruction on `del`/assignment and silently-wrong computed paths. Fixing this properly needs those producers audited/fixed first (or a structurally different mechanism), not a drop-in helper. Severity below should be read as 'the observable symptom is low' — the *fix* is materially harder than originally scoped.

**Severity: Low**

## Summary

Split out from #962 item 1 (found reviewing #896/PR #916). `resolve_node`'s `FirstExpr`/`Builtin::FirstStream` and `resolve_limit` call the inner generator via `resolve_node::<S>(inner, ...)?` and then `.take(n)` on the resulting `Vec` — but the "keep-partial" resolvers (`recurse`'s `f`/`cond` loops from #842/#854, and `select`/`if`/`getpath`/computed-index from #896) eagerly evaluate the *entire* underlying generator, including any error *after* the nth output, before `first`/`limit` ever gets a chance to truncate.

Real jq's lazy generator model stops pulling from the inner generator once `n` outputs have been taken, so a later error (after the nth output) is never reached at all.

## Repro (verified against pinned jq 1.7.1, both pre- and post-#896 succinctly builds)

```
$ echo '1' | jq -c 'path(first(select((true, error("x")))))'
[]
$ echo '1' | succinctly jq -c 'path(first(select((true, error("x")))))'
# pre-#896: no output, exit 5 (error silently swallows the correct [] output)
# post-#896: prints [] (looks right!) then still raises "x", exit 5 (correct output, but a
#            trailing error jq never raises — arguably a more confusing symptom, since the
#            correct partial output is now visible right before the spurious error)
```

Same shape confirmed for `first(if ... then ... else ... end)`, `first(.[(...)])`, `first(getpath(...))`, and `limit(1; select(...))`.

## Root cause

[...]

PR #973, opened 2026-08-15
refactor(jq): collapse path-resolver escape-handling into one path_result helper

### LINKED ISSUE #962: jq: path-resolver escape-handling duplication, and first()/limit() interaction with keep-partial resolvers
## Summary

Follow-up from the #916 review round (fix for #896). The review found and this session fixed two real, live-reproduced bugs (a wrong-error-masking regression in \`GetPath\`'s multi-output fallback, and \`resolve_index_expr\` discarding a non-empty target prefix on error). A few lower-priority, non-regression findings from the same review round weren't acted on:

## 1. \`first(f)\`/\`limit(n; f)\` interaction with the keep-partial resolvers (#842/#854/#896)

\`resolve_node\`'s \`FirstExpr\`/\`Builtin::FirstStream\` and \`resolve_limit\` call \`resolve_node::<S>(inner, ...)?\` then \`.take(n)\` — but the keep-partial resolvers (\`recurse\`'s \`f\`/\`cond\` from #842/#854, and this issue's own \`select\`/\`if\`/\`getpath\`/computed-index sites) now eagerly evaluate their *entire* generator (including any trailing error) before \`first\`/\`limit\` ever gets a chance to truncate. Real jq's lazy generator model stops pulling from \`f\` once \`n\` outputs are taken, so a later error in \`f\` (after the nth output) is never reached.

Confirmed via direct reproduction (jq 1.7.1 vs both pre- and post-#896 builds):

\`\`\`
echo '1' | jq -c 'path(first(select((true, error("x")))))'
# real jq: prints [], exit 0
# pre-#896 succinctly: no output, exit 5 (already wrong, silently)
# post-#896 succinctly: prints [] (looks right!) then still raises "x", exit 5
\`\`\`

Same class confirmed for \`first(if ... then ... else ... end)\`, \`first(.[(...)])\`, \`first(getpath(...))\`, and \`limit(1; select(...))\`.

**Root cause pre-dates #896** (eager whole-generator evaluation ignoring \`take()\`-style truncation already caused a wrong exit code before it, for \`recurse\`) — #896 doesn't introduce the false error, but it does introduce a more confusing symptom (correct-looking partial stdout immediately followed by an error jq never raises) for the 4 sites it touches, on top of the pre-existing \`recurse\` case. Not covered by any existing test.

**Scope note**: fixing this needs \`first\`/\`limit\` to actually truncate the underlying generator lazily, or at least stop threading a trailing escape through once enough outputs have been taken — likely a shared primitive (\`resolve_node\`'s own analogue to \`eval_binary_fanout\`'s early-exit), not a per-site patch. Worth scoping as its own issue given the blast radius (affects `recurse` too, not just #896's 4 sites).

## 2. Escape-handling idiom duplicated across 6+ sites

The \`match escape { Some(e) => Err((prefix, e)), None => Ok(prefix) }\` shape (or its close variants) now appears at: \`eval_owned_multi\`'s own body, \`resolve_leaf\`'s \`empty_case\` closure, \`resolve_recurse\`'s \`pending_error\` match, and all 4+ of #896's new sites (\`Select\`, \`If\`, \`GetPath\`, \`resolve_index_expr\`'s two return points). A small \`fn path_result<'a>(branches: Vec<PathBranch<'a>>, escape: Option<EvalEscape>) -> PathResolveResult<'a>\` helper would collapse every site to one call, and would have structurally prevented the two review-round bugs above (a stray unmatched arm, an escape variable going unread on a fallthrough path) rather than relying on manual care at each site.

## 3. Doc-comment staleness

\`eval_owned_multi\`'s and \`eval_owned_multi_keep_partial\`'s own doc comments still describe their "direct callers" as only the \`recurse\` family plus \`resolve_leaf\`'s fallback — neither list was updated to include #896's 4 new direct callers (\`Select\`'s cond, \`If\`'s cond, \`GetPath\`'s arg, \`resolve_index_expr\`'s key). Minor, but exactly the kind of staleness this codebase's own CLAUDE.md flags ("duplicated predicates diverge silently").

## Scope note

None of these are live, reproducible bugs on their own (item 1 is a real but pre-existing-and-widened UX confusion, not a new false-positive; items 2-3 are pure maintainability). Lower priority than #896's confirmed fixes (shipped in #916). Tracked here so they aren't lost.

newhoggy:
## Summary

- #962 item 2: the `Some(e) => Err((prefix, e)), None => Ok(prefix)` idiom was hand-rolled independently at 7 `PathResolveResult`-returning call sites in `src/jq/eval.rs` (`resolve_node`'s `Select`/`If`/`GetPath` arms, `resolve_index_expr`'s two return points, `resolve_leaf`'s `empty_case` closure, `resolve_recurse`'s `pending_error` match) — 4 of them added by #896. Collapsed into one `path_result(branches, escape) -> PathResolveResult` helper, placed right next to the `PathResolveResult` type alias it wraps.
- #962 item 3: `eval_owned_multi_keep_partial`'s doc comment listed its "direct callers today" but hadn't been updated for #896's 4 new callers (`Select`/`If`'s `cond`, `GetPath`'s `arg`, `resolve_index_expr`'s `key`). Updated.
- #962 item 1 (a real but wider-blast-radius bug: `first`/`limit` don't truncate the keep-partial resolvers lazily, so a trailing error after the nth output still surfaces) is **out of scope here** — filed separately as #972, since it also affects pre-existing `recurse` behavior, not just #896's 4 sites, and needs its own lazy-truncation design rather than a per-site patch.

This is a pure refactor — every touched arm's control flow is unchanged. Verified by spot-checking each of the 6 previously-`#896`/`#854`/`#861`-covered repros against both this branch and real jq 1.7.1 (identical output/exit codes before and after), and confirmed via `omni-dev coverage diff` that all 13 changed lines are already covered by the existing test suite (no new test needed, since no new behavior was introduced).

## Test plan

- [x] Spot-checked every touched arm's output against real jq 1.7.1 (select/if/computed-index/getpath partial-fanout cases, recurse cond partial, resolve_leaf non-primitive fallback, resolve_index_expr target-prefix case) — all identical before/after
- [x] `omni-dev coverage diff`: 100% patch coverage (13/13 new lines), consistent with this being a pure refactor of already-tested code

Fixes #962

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -10128,7 +10128,11 @@ fn eval_owned_multi_first<S: EvalSemantics>(
 /// same child), and `resolve_leaf`'s non-primitive fallback, also path
 /// position (#861 — the same all-or-nothing pitfall reaching
 /// `path(...)`-wrapped expressions whose argument isn't one of the
-/// specially-tracked navigation primitives). `resolve_recurse`'s own
+/// specially-tracked navigation primitives). Also `resolve_node`'s own
+/// `Select`/`If` arms (their `cond`), its `GetPath` arm (its `arg`), and
+/// `resolve_index_expr` (its `key`) — the same argument-can-itself-be-a-
+/// generator shape #842/#854 fixed for `recurse`, independently live at
+/// these 4 unrelated call sites too (#896). `resolve_recurse`'s own
 /// `f`-evaluation step is the one exception: it resolves `f` through
 /// `resolve_node`/`resolve_against_cow` instead — a structurally different
 /// call whose `PathResolveResult` already threads a prefix through its own
@@ -10240,6 +10244,24 @@ type PathBranch<'a> = (Vec<Expr>, Cow<'a, OwnedValue>);
 /// treats this exactly like a plain `Err`, matching jq's atomic write-side
 /// semantics (confirmed live: `(.a, 1) = 5` produces no output at all).
 type PathResolveResult<'a> = Result<Vec<PathBranch<'a>>, (Vec<PathBranch<'a>>, EvalEscape)>;
+
+/// Combine an already-resolved branch prefix with an optional trailing
+/// escape into this resolver's `PathResolveResult` shape.
+///
+/// Every "keep-partial" call site below ends by matching a
+/// `(prefix, Option<EvalEscape>)` pair the exact same way — `Some(e) =>
+/// Err((prefix, e)), None => Ok(prefix)` — hand-written independently at
+/// each one. Collapsing that into a single call removes the chance of a
+/// stray unmatched arm or an escape variable going unread at a *new* call
+/// site by omission, the way two of #896's own review-round bugs did
+/// (#962).
+#[inline]
+fn path_result(branches: Vec<PathBranch<'_>>, escape: Option<EvalEscape>) -> PathResolveResult<'_> {
+    match escape {
+        Some(e) => Err((branches, e)),
+        None => Ok(branches),
+    }
+}
 
 /// Resolve one path node against one value, yielding a branch per output.
 ///
@@ -10541,10 +10563,7 @@ fn resolve_node<'a, S: EvalSemantics>(
                 .filter(OwnedValue::is_truthy)
                 .map(|_| (Vec::new(), Cow::Borrowed(value)))
                 .collect();
-            match escape {
-                Some(e) => Err((branches, e)),
-                None => Ok(branches),
-            }
+            path_result(branches, escape)
         }
         Expr::Builtin(
             builtin @ (Builtin::Values
@@ -10615,10 +10634,7 @@ fn resolve_node<'a, S: EvalSemantics>(
                     }
                 }
             }
-            match escape {
-                Some(e) => Err((out, e)),
-                None => Ok(out),
-            }
+            path_result(out, escape)
         }
 
         // `a // b`: resolve `a`, keep only its truthy branches — jq's `//`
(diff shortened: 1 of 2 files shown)


newhoggy:
Mandatory review completed. Six findings, all independently verified before acting:

1. **`eval_owned_multi_keep_partial`'s caller list missing `builtin_in`** — real gap, fixed.
2. **`path_result`'s doc comment overclaimed what it prevents**, citing #896's own review-round bugs as something this helper would have caught. Checked the actual fix commit (84551d7f9): neither bug was a hand-rolled `Some`/`None` match failure (one was a bare `?` discarding a prefix, the other a fallthrough that skipped checking an escape at all) — `path_result` doesn't address either's root cause. Corrected the comment to be accurate about what it does and doesn't guard against.
3. **A live divergence the review found via a `resolve_slice_expr` repro** — real and reproducible, but **misattributed**: instrumented the function directly and confirmed it is never reached by the review's own repro (`(select(...))[0:1]`, literal bounds). The parser desugars a *literal*-bounded `TARGET[N:M]` into a static `Pipe([TARGET, Slice{N,M}])` at parse time, which routes through `resolve_seq`'s fan-out loop instead — a code path #896 never touched, and a genuinely new, more general bug (drops a static tail — literal `.foo`/`[N]`/`[N:M]` — whenever an earlier pipe element's keep-partial escape fires). Filed as #977 rather than fixed here, since it needs its own design (the fix has to thread a second fallible step, not just reorder an existing one).
4. `resolve_slice_expr` **does** have the same bare-`?` bug #896 fixed in `resolve_index_expr`, just only reachable with a genuinely *dynamic* bound (`[(0+1):2]`). Fixed for real this time, verified against jq 1.7.1 with a repro that actually exercises the function (confirmed via the same instrumentation), and covered by a new regression test.
5. Orphaned comment on `resolve_leaf`'s second `empty_case`-inlining call site — fixed.
6. `#[inline]` on `path_result` inconsistent with every neighboring path-resolution helper (none of them use it) — dropped.

Second commit pushed with all of the above.

Issue #980, opened 2026-08-15
jq: path()'s Alternative arm doesn't apply // 's truthy filter per-output when a later Comma sibling is the one that's falsy/non-path-shaped


newhoggy:
**Severity: Low**

## Summary

Split out from #845, whose fix covers the narrow (and most common) shape: `left` resolving as a *whole* to a single non-path-shaped value, with nothing already resolved ahead of it. This issue is the harder, more general shape #845 explicitly didn't attempt: a `Comma`-fanned `left` where an *earlier* sibling already resolved successfully as a path, and a *later* sibling is the one that's falsy/non-path-shaped — real jq still silently drops that later falsy sibling (via `//`'s own per-output truthy filter, applied before any path-shape check), but this resolver's `Comma` arm has already committed to raising on the first sibling that fails, never getting the chance to apply `//`'s filter to it.

## Repro (verified against pinned jq 1.7.1)

```
$ echo '{"a":10}' | jq -c 'path((.a, false) // .b)'
["a"]
$ echo '{"a":10}' | succinctly jq -c 'path((.a, false) // .b)'
jq: error (at <stdin>:1): Invalid path expression with result false
```

```
$ echo '{"a":10}' | jq -c 'path((false, .a) // .b)'
["a"]
$ echo '{"a":10}' | succinctly jq -c 'path((false, .a) // .b)'
jq: error (at <stdin>:1): Invalid path expression with result false
```

Both raise a spurious error where real jq exits cleanly with `["a"]` — real jq never even considers `false` an error, since `//`'s truthy filter discards it before path-shape is ever checked, and the sibling ordering doesn't matter (`.a` first or `false` first both still succeed).

Contrast with the case #845's fix *does* handle correctly (a later truthy-but-non-path-shaped sibling, which legitimately still raises, prefix and all):

```
$ echo '{"a":10}' | jq -c 'path((.a, 1) // .b)'
["a"]
jq: error (at <stdin>:1): Invalid path expression with result 1
$ echo '{"a":10}' | succinctly jq -c 'path((.a, 1) // .b)'
["a"]
jq: error (at <stdin>:1): Invalid path expression with result 1
```

## Root cause

[...]

Issue #820, opened 2026-08-11
jq: comma-expression evaluation isn't lazy, leaking side effects a short-circuiting consumer never needed


newhoggy:
## Summary

Comma-expression argument evaluation isn't fully lazy: a side-effecting builtin (`stderr`, `halt_error`, `debug`) sitting in a later branch of a comma expression can fire even when the consumer only needed an earlier branch's value and, in real jq's generator semantics, would never reach the later one at all.

Found during #791's review (PR #816) while verifying `isempty`'s halt-propagation fix; confirmed live against real jq 1.7.1.

## Repro

```
$ jq -n 'isempty(1, stderr)'
false
$ succinctly jq -n 'isempty(1, stderr)'
nullfalse
```

Real jq prints nothing to stderr — `isempty(g)` desugars to `label $out | (g | false, break $out), true`, so as soon as `g` produces its first value (`1`), the `break $out` unwinds and `g`'s continuation (the `stderr` branch) is never evaluated at all.

succinctly's `isempty` calls `eval_single`/`eval_comma` on the whole `g` expression before ever inspecting how many outputs were needed, so `eval_comma` runs to its own natural stopping point (successfully finishing the whole comma list, since nothing here raises/halts/breaks) — evaluating `stderr` as a side effect along the way, which writes `null` (the current value, compact JSON, no trailing newline) to stderr before `isempty` ever gets to decide it only needed the first item.

The same eagerness reaches `halt_error(n)`'s own exit-code argument if `n` is itself a comma expression:

```
$ jq -n '"outer" | halt_error(1, ("inner"|halt_error(2)))'
outer          # only the outer's own message; exits 1
$ succinctly jq -n '"outer" | halt_error(1, ("inner"|halt_error(2)))'
innerouter     # inner's message leaks too; exits 1
```

Real jq's `halt_error(ec)` also only takes the first value of `ec` — confirmed by the matching exit code (1, from the first comma branch) — but never evaluates the second branch at all, so `"inner"|halt_error(2)`'s own stderr write never happens.

## Scope

Not systemic: `first(f)`/`limit(n; f)` already short-circuit correctly —

```
$ succinctly jq -cn 'first(1, debug)'
1
```

[...]

PR #982, opened 2026-08-15
fix(jq): path()'s Alternative arm no longer requires a falsy left side to be path-shaped

### LINKED ISSUE #845: jq: path()'s Alternative (//) arm requires the left operand to be path-shaped even when it's simply falsy and filtered out

newhoggy:
## Summary

`resolve_node`'s `Expr::Alternative` arm (`src/jq/eval.rs`) resolved `left` as a full path expression via `?` *before* checking whether any of its outputs were truthy. Real jq only requires `//`'s left side to be path-shaped for outputs that actually survive the truthy filter — a left side whose only output is falsy never needs to have been path-trackable at all, since it's filtered out before ever being part of the result.

```
$ echo '{"a":10}' | jq -c 'path(false // .b)'
["b"]
$ echo '{"a":10}' | succinctly jq -c 'path(false // .b)'   # before this fix
jq: error (at <stdin>:1): Invalid path expression with result false
```

Fixed for the shape this resolver can distinguish without a second, structurally different evaluation pass: `left` resolving as a *whole* to a single non-path-shaped value, with nothing already resolved ahead of it (`prefix` empty on the `#530` error). In that case, `left`'s plain *value* is cheap and side-effect-safe to re-check — there's no earlier branch whose own evaluation this could double up on. A truthy-but-non-path-shaped left (`1 // .b`) still raises, matching jq exactly.

A harder, related shape — a `Comma`-fanned `left` where an *earlier* sibling already resolved successfully as a path and a *later* sibling is the falsy/non-path one (`(.a, false) // .b`, real jq: `["a"]`, no error) — is **not** fixed here. This resolver's `Comma` arm commits to the first sibling's failure eagerly, before `Alternative`'s own truthy filter ever gets a chance to see it; fixing that needs the resolver's path resolution to be genuinely lazy per output, the same limitation #820 already tracks. Filed as #980.

## Test plan

- [x] 8 new regression tests covering: falsy-left fallthrough (`false`, `null`), truthy-non-path-left still raises, genuine `error(...)` still propagates, truthy-path-left unaffected, bare `path(false)` (no `//`) still raises, the already-correct prefix-carrying case (`(.a, 1) // .b`), and `try`/`catch` interaction — all verified directly against real jq 1.7.1
- [x] `omni-dev coverage diff`: 100% patch coverage (16/16 new lines)

Fixes #845

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -10274,6 +10274,16 @@ fn path_result(branches: Vec<PathBranch<'_>>, escape: Option<EvalEscape>) -> Pat
     }
 }
 
+/// Peel any wrapping `Expr::Paren` down to the inner expression — `(false)`
+/// and `false` are the same node for a caller that only cares about literal
+/// shape, not surface syntax (`Expr::Alternative`'s `#845` fix, below).
+fn unwrap_paren(expr: &Expr) -> &Expr {
+    match expr {
+        Expr::Paren(inner) => unwrap_paren(inner),
+        other => other,
+    }
+}
+
 /// Resolve one path node against one value, yielding a branch per output.
 ///
 /// `trackable` is `false` for exactly one caller (`resolve_catch`, for the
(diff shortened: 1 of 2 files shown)


newhoggy:
Mandatory review completed. The reviewer found real, serious problems with the first commit's approach — all independently verified before acting:

1. **Double evaluation of `left`** — the first fix re-checked truthiness by calling `eval_owned_multi::<S>(left, value)` a second time once `resolve_node`'s own path-mode resolution already failed with `#530`. Confirmed live: `path(stderr // .b)` wrote its input to stderr **twice**; real jq writes it once.
2. **Halt swallowed** — the catch-all on the second evaluation's outcome converted a `halt_error(3)` into an ordinary catchable `#530` error. Confirmed live: `path((1|halt_error(3)) // .b)` printed `1` then raised instead of halting with exit 3.
3. **Break swallowed** — same catch-all, same mechanism, for `break $label`.
4. **Downstream error masked** — a genuine `error(...)` past the point the first evaluation stopped got hidden behind the stale `#530` message from the *first* run.
5. **Wrong value named for a Comma's eagerly-aborted first sibling** (`(false, 5) // .b`) — real, but traced to the exact same pre-existing "Comma doesn't apply `//`'s truthy filter lazily" gap #980 already tracks (this resolver's `Comma` arm commits to the first sibling's own failure before `Alternative` ever sees it), not something the first fix's `prefix.is_empty()` guard could safely resolve without also risking bug #1 above.

**Fix**: replaced the whole "catch the error, re-evaluate to check truthiness" approach with a static, zero-evaluation check — only a bare `Expr::Literal` left side (`false`, `null`, ...) gets the new leniency, since a literal's value is known without running anything at all. Every other shape (including the Comma case) falls through unchanged to the original `?`-based propagation, so none of findings 1–4 have any surface area left, and finding 5 remains exactly as before this PR (still tracked by #980, not newly introduced).

While verifying finding 2/3/4 against the oracle, discovered they're **not actually caused by this PR at all** — `path(1|halt_error(3))`, `path(1|break $out)`, and `path(1|error("boom"))` all diverge from jq identically on a completely bare `path(...)` call with no `//` involved, reproducing on unmodified `main`. `resolve_seq` treats a non-terminal literal/value-producing pipe element as its own path leaf, unrelated to `Alternative`. Filed as #986.

Second commit pushed with the rewritten fix, 3 new regression tests (no-double-eval, `del`/`=`/`|=` blast-radius coverage, `catch`+untracked-payload interaction), and updated doc comments. All local verification green (build, full suite, both clippy passes, 100% patch coverage).

PR #985, opened 2026-08-15
fix(jq): path(first(...))/path(limit(...)) discard a trailing error jq never reaches

### LINKED ISSUE #972: jq: first()/limit() don't truncate the keep-partial path resolvers, causing a spurious trailing error after correct partial output

newhoggy:
## Summary

`resolve_node`'s `first(...)`/`limit(...)` handling used a bare `?` on the inner generator's `PathResolveResult`, propagating any error unconditionally — even when the "keep-partial" resolvers (`select`/`if`/`getpath`/computed-index from #896, `recurse` from #842/#854) had already produced enough output for `first`/`limit` to be fully satisfied. jq's own lazy generator model never asks a source for a value past what it needs, so it never even reaches an error that comes *after* that point.

```
$ echo '1' | jq -c 'path(first(select((true, error("x")))))'
[]
$ echo '1' | succinctly jq -c 'path(first(select((true, error("x")))))'   # before
jq: error (at <stdin>:1): x
[]
```

`PathResolveResult`'s `Err` side already carries the partial prefix collected before the error — every keep-partial resolver already threads it there for exactly this reason (that's the whole point of `PathResolveResult<'a> = Result<Vec<PathBranch<'a>>, (Vec<PathBranch<'a>>, EvalEscape)>`). The bug was purely that `first`/`limit` never looked at it.

Added `take_path_branches`, a shared helper (per the issue's own suggested direction — "likely wants a shared primitive... since the same truncation gap applies uniformly") that checks whether the already-collected partial output is enough before discarding the trailing error, used by both `first` (n=1) and `limit` (n=arbitrary) so the two can't drift on the same comparison.

Fixes #972

## Verified against real jq 1.7.1

All 5 shapes named in the issue, plus `recurse` (the pre-#896 case):

```
$ echo '1' | succinctly jq -c 'path(first(select((true, error("x")))))'
[]
$ echo '1' | succinctly jq -c 'path(limit(1; select((true, error("x")))))'
[]
$ echo '1' | succinctly jq -c 'path(first(if (true, error("x")) then . else empty end))'
[]
$ echo '[1,2,3]' | succinctly jq -c 'path(first(.[(0,error("x"))]))'
[0]
$ echo '{"a":1}' | succinctly jq -c 'path(first(getpath((["a"],error("x")))))'
["a"]
$ echo '{"a":{"b":1}}' | succinctly jq -c 'path(first(recurse(if type=="object" then .[] else error("x") end)))'
[]
```

All exit 0, byte-identical to real jq, no unwanted stderr.

Plus two sanity checks confirming the fix doesn't overreach — an error still surfaces normally when the source doesn't produce *enough* output first:

```
$ echo '1' | succinctly jq -c 'path(first(select((error("x")))))'
jq: error (at <stdin>:1): x     # exit 5, matches real jq

$ echo '1' | succinctly jq -c 'path(limit(3; select((true, true, error("x")))))'
[]
[]
jq: error (at <stdin>:1): x     # only 2 of the needed 3 outputs, error correctly surfaces
```

## Test plan
- [x] `cargo build --features cli,regex`
- [x] `cargo test --features cli,simd,regex,serde --test jq_cli_tests -- 972` — 8 tests covering all 5 keep-partial resolver shapes plus 2 sanity checks
- [x] `cargo test --features cli,simd,regex,serde` (full suite) — all green across all 54 test binaries, 0 failed
- [x] Manually verified every repro case above against real jq 1.7.1


newhoggy:
Closing without merging. Code review found the fix is unsafe as designed, and I've reverted the branch back to `main` (revert commit `5c82a1c42`) rather than continuing to patch around it.

## What went wrong

`take_path_branches`'s core assumption — that `PathResolveResult`'s `Err((partial, escape))` always carries *exactly* the branches jq's own generator would have already emitted, in order, strictly before `escape` — does not hold across every producer in the `resolve_node` family. Two independent review angles found live, severe violations:

**1. Silent data destruction on the write side.** `resolve_index_expr`/`resolve_slice_expr` loop `for key in keys { for target in target_branches }` and defer their own escape to the end, so `partial` can contain branches for keys jq's own lazy generator never reaches:

```
$ echo '{"x":9,"y":8}' | jq -c 'del(limit(2; select((true,error("t")))[("x","y")]))'
jq: error (at <stdin>:1): t          # exit 5, document untouched
$ echo '{"x":9,"y":8}' | succinctly jq -c 'del(...)'   # with this PR's fix
{}                                    # exit 0 — BOTH keys deleted
```

**2. Silently wrong computed paths on multi-stage pipes.** `resolve_seq`'s own doc comment already discloses it rebuilds branches stage-by-stage rather than depth-first per branch, "not a byte-for-byte reproduction of jq's stream order" for pipes with more than one dynamic element. `take_path_branches` promotes that known-incomplete intermediate prefix into a false "final success":

```
$ echo '{"a":{"p":{"q":1}},"b":3}' | jq -c 'path(first(.[]|.[]|.[]))'
["a","p","q"]
$ echo '...' | succinctly jq -c '...'   # with this PR's fix
["a","p"]        # exit 0 — wrong, truncated path, silently
```

Both are confirmed live against real jq and against this branch's own build. Given the original issue was explicitly Low severity ("not a silent-data-loss bug... an extra, unwanted, jq-incompatible trailing error"), shipping a fix that introduces genuine silent data loss/corruption on the write path is a clear net regression, not a net improvement.

## Why this needs more than a per-callsite patch

Properly fixing this requires either:
- Auditing and fixing every producer that constructs a `PathResolveResult::Err` (`resolve_index_expr`, `resolve_slice_expr`, `resolve_seq`'s multi-stage case, and any others not yet found) to genuinely uphold "partial is exactly what jq would have emitted before this escape, in order" — which for `resolve_seq` specifically means addressing its documented depth-first-vs-stage-by-stage simplification, a materially larger piece of work than this issue's own scope; or
- A structurally different mechanism that doesn't rely on this invariant at all (e.g. a real lazy/coroutine-style path generator, which the codebase doesn't currently have for path resolution).

Neither is a "shared primitive, few-line fix" the way the original issue described — the true scope only became visible once code review specifically probed multi-key index/slice fan-outs and multi-stage pipes, which the fix's own test suite (all single-stage, single-key cases, matching the issue's own repro list) didn't happen to exercise.

## Also found along the way (not blocking, filed for the record)

- `nth(n; g)` has no path-position arm in `resolve_node` at all (`path(nth(0; .a))` errors instead of resolving) — a separate, pre-existing gap from #972, and note for whoever eventually adds it: jq's `nth` has the identical lazy-truncation shape, so it would need the same care this issue's fix needed, not a bare `?`.
- `Expr::While` also has no path-position arm — unrelated, different shape, no truncation/laziness component.
- The doc comments I wrote overclaimed "never reached" laziness — succinctly's resolver is eager, so a discarded escape's *side effects* (e.g. `stderr`) still execute even though the escape itself is dropped; only real jq is genuinely lazy there.

Re-opening this as future work with the above context, rather than leaving it looking done. Thank you to the review — this is exactly the kind of regression the gate exists to catch before merge.

Issue #987, opened 2026-08-15
jq: path()'s non-primitive resolver evaluates its whole argument eagerly, so a discarded later output's side effects and error message still leak through


newhoggy:
**Severity: Low**

## Summary

Split out from #891, which fixed the wrong error message `resolve_leaf`'s non-primitive fallback (`src/jq/eval.rs`) raised for a multi-output `path(...)` argument (`path(range(3))`), bringing it in line with real jq's per-output-checked wording. That fix only addressed the *message*; the deeper issue #891 also described — that the underlying evaluation is eager, not lazy — remains.

`resolve_leaf`'s fallback calls `eval_owned_multi_keep_partial`, which internally runs `expr` to full completion (via `eval_owned_input`, materializing every output into a `QueryResult::Many`/`ManyOwned`/`Partial` *before* `resolve_leaf` ever gets to inspect `values[0]`'s path-shape). Real jq's `path(...)` tracking instead checks path-shape *as each value streams*, and never even asks the underlying generator for a value past the first non-path-shaped one — so any side effect or error embedded in a *later*, never-actually-needed output simply never happens in real jq, but does happen in succinctly (its side effect already fired, and/or its error message is discarded in favor of the first output's).

## Repro (verified against pinned jq 1.7.1)

```
$ echo '[1,2,"trigger"]' | jq -c 'path(paths(if type=="string" then error("my custom message") else true end))'
jq: error (at <stdin>:1): Invalid path expression with result [0]
$ echo '[1,2,"trigger"]' | succinctly jq -c 'path(paths(if type=="string" then error("my custom message") else true end))'
jq: error (at <stdin>:1): Invalid path expression with result [0]
```
(Both now show the *same message* after #891 — the first output's own shape is what both actually report — so this specific repro no longer demonstrates the gap on its own. A repro that actually observes eager evaluation needs a *side effect*, not just a differently-worded error, on a later candidate:)

```
$ echo '[1,2,3]' | jq -c 'path(paths(if . == 3 then (stderr, true) else true end))'
jq: error (at <stdin>:1): Invalid path expression with result [0]
(no stderr output -- jq never reaches the `stderr` candidate, since [0] already fails first)
$ echo '[1,2,3]' | succinctly jq -c 'path(paths(if . == 3 then (stderr, true) else true end))'
3
jq: error (at <stdin>:1): Invalid path expression with result [0]
(stderr write for the `3` candidate happens anyway, even though the reported error is for `[0]`, not it)
```

[...]

Issue #986, opened 2026-08-15
jq: resolve_seq treats a non-terminal literal/value-producing pipe element as its own path leaf, mishandling Halt/Break/downstream errors and misreporting navigation failures


newhoggy:
**Severity: Medium**

## Summary

Found while investigating a #845/PR #982 review finding that initially looked like a bug in the `Expr::Alternative` fix itself, but reproduces identically on a completely bare `path(...)` call with no `//` involved at all — confirming it's pre-existing on `main`, unrelated to #845.

`resolve_seq` (`src/jq/eval.rs`) flattens a `Pipe` into components via `push_path_components`, then runs every element that `needs_fanout_pass` (anything that isn't a bare `Identity`/`Field`/`Index`/`Slice`/`Iterate`) through a fan-out loop via `resolve_against_cow`/`resolve_node`. This treats **every** such element as something that must itself resolve to a valid path branch — including an element that is only providing a *value* for the *next* pipe stage to consume, not a navigational step of its own.

For `1 | halt_error(3)` (`Expr::Pipe([Literal(1), <halt_error call>])`), both `Literal(1)` and the `halt_error(...)` call need a fan-out pass (neither is Field/Index/Slice/Identity), so the loop resolves `Literal(1)` **first, on its own**, via `resolve_node`/`resolve_leaf` — which immediately raises `#530`'s "Invalid path expression with result 1" for the bare literal, since a `Literal` never has any path-shape at all. Real jq never treats `1` as needing to be a path leaf here: it is simply the value `halt_error(3)` (or whatever the next stage is) operates on.

## Repro (verified against pinned jq 1.7.1, no `//` anywhere)

```
$ echo '{"a":10}' | jq -c 'path(1|halt_error(3))'
1
(exit 3 — halts)
$ echo '{"a":10}' | succinctly jq -c 'path(1|halt_error(3))'
jq: error (at <stdin>:1): Invalid path expression with result 1
(exit 5 — wrong: raises an ordinary, catchable error instead of halting)
```

```
$ echo '{"a":10}' | jq -c 'label $out | path(1|break $out)'
(no output, exit 0)
$ echo '{"a":10}' | succinctly jq -c 'label $out | path(1|break $out)'
jq: error (at <stdin>:1): Invalid path expression with result 1
(exit 5 — wrong: raises instead of unwinding to the label)
```

```
$ echo '{"a":10}' | jq -c 'path(1|error("boom"))'
jq: error (at <stdin>:1): boom
[...]

Issue #977, opened 2026-08-15
jq: resolve_seq's fan-out loop drops a static tail (literal .foo/[N]/[N:M]) when an earlier pipe element's keep-partial escape fires


newhoggy:
**Severity: Low-Medium**

## Summary

Found while investigating a #973 review finding that misattributed this bug to `resolve_slice_expr` (that attribution was checked and is wrong — see below — but the underlying divergence it pointed at is real and reproducible, just via a different mechanism).

`resolve_seq` (`src/jq/eval.rs`) flattens a pipe into static components (`push_path_components`) and, when one or more of those components needs a dynamic "fan-out" pass (`needs_fanout_pass` — e.g. `select`, `if`, a computed index/slice, `..`/`recurse`, ...), runs only the elements up to and including the *last* such dynamic element through a fan-out loop, then applies whatever purely-static components remain after it (`tail`) to each surviving branch.

The bug: if that fan-out loop's element (not necessarily the pipe's last element overall) itself escapes with a non-empty partial prefix — one of #842/#854/#896's "keep-partial" sites (`select`, `if`, `getpath`, a computed index) — the loop returns immediately with that raw prefix, **never applying `tail` to it**. Any purely-static suffix after the last dynamic element (a literal `.foo`, `[N]`, or `[N:M]` — anything the parser can constant-fold into `Expr::Field`/`Expr::Index`/`Expr::Slice` at parse time) is silently dropped from the streamed partial output.

This is architecturally distinct from #896's own 4 sites and from #973's `resolve_slice_expr` fix: those all cover a *dynamic* (non-constant-foldable) `IndexExpr`/`SliceExpr`/`GetPath` argument, which routes through `resolve_index_expr`/`resolve_slice_expr`/the `GetPath` arm directly. A **literal** subscript after a dynamic target, by contrast, gets desugared by the parser into a flat `Pipe` of static components at parse time (confirmed via `push_path_components`/`needs_path_prepass`'s doc comments) and goes through `resolve_seq`'s own fan-out loop instead — a code path #896/#973 never touched.

## Repro (verified against pinned jq 1.7.1)

```
$ echo '[10,20,30]' | jq -c 'path((select(true, error("t")))[0])'
jq: error (at <stdin>:1): t
[0]
$ echo '[10,20,30]' | succinctly jq -c 'path((select(true, error("t")))[0])'
jq: error (at <stdin>:1): t
(no path printed)
```

Same shape for a literal slice:

```
[...]

PR #988, opened 2026-08-15
fix(jq): path()'s non-primitive fallback names the first value on a multi-output result

### LINKED ISSUE #891: jq: path(...)'s non-primitive fallback is eager, not lazy — wrong message and discarded escapes at 2+ outputs
**Severity: Low**

## Summary

`resolve_leaf`'s fallback for `path(...)`-wrapping a non-path-trackable expression (e.g. `range(3)`, `paths(node_filter)`) evaluates the whole argument *eagerly* before checking path-shape on any output. Real jq's `path(...)` checks path-shape *per output as it streams*, raising `Invalid path expression` on the first non-path-shaped value and never even evaluating a later output — including one that would itself `error`/`break`/`halt`.

This produces two observable divergences from real jq whenever `resolve_leaf`'s non-primitive fallback (the `match values.len() { 0 | 1 | _ => ... }` block, `src/jq/eval.rs`) sees **2 or more** produced values:

1. **Wrong error message.** `path(range(3))` raises `"Cannot use a computed index after a multi-output path component"` in succinctly, but real jq raises `"Invalid path expression with result 0"` — same message shape #530/#861 already use for the single-output case, just never applied to the `values.len() >= 2` arm. This divergence pre-dates #861 (confirmed: `main`'s pre-#861 binary shows the same wrong message for `path(range(3))`) and was left untouched by #861's own fix, which is scoped to the `values.len() == 1` case.

2. **Discarded escape content/observable side effects**, newly surfaced by #861's PR review (rust-works/succinctly#886). Because `resolve_leaf` now evaluates via `eval_owned_multi_keep_partial` (keeps the prefix instead of discarding everything via `eval_owned_multi`'s `?`), a `Break`/`Error` trailing 2+ already-produced values is deliberately discarded (matches real jq's own laziness, confirmed live — see PR #886's discussion) — but:
   - A real `error("msg")` on the third-or-later candidate has its actual message replaced by the generic multi-output message, rather than surfacing `"msg"` (pre-#861 behavior, via the old `?`-based propagation, at least showed the real message, even though it also didn't match jq's own "never reached" semantics).
   - Any side-effecting builtin evaluated as part of producing that later, ultimately-discarded output (e.g. `halt_error`'s stderr write) still executes, since evaluation is eager — a visible side effect real jq's laziness would never produce at all.
   - `Halt` specifically is now defensively *always* propagated regardless of prefix length (#886's own fix), which is the one case handled correctly today — intentionally diverging from jq's true "never reached, so never halts" semantics in favor of never letting `halt_error` become silently catchable.

## Repro

```
$ echo null | jq -c 'path(range(3))'
jq: error (at <stdin>:1): Invalid path expression with result 0
$ echo null | succinctly jq -c 'path(range(3))'
jq: error (at <stdin>:1): Cannot use a computed index after a multi-output path component

$ echo '[1,2,"trigger"]' | jq -c 'path(paths(if type=="string" then error("my custom message") else true end))'
jq: error (at <stdin>:1): Invalid path expression with result [0]
$ echo '[1,2,"trigger"]' | succinctly jq -c 'path(paths(if type=="string" then error("my custom message") else true end))'
jq: error (at <stdin>:1): Cannot use a computed index after a multi-output path component
```

## Root cause

`resolve_leaf` (`src/jq/eval.rs`) materializes `expr`'s entire output list via `eval_owned_multi_keep_partial` before ever checking whether any individual output is path-shaped. Real jq's `path(...)` bytecode-level tracking checks incrementally and never resumes the generator once one output fails the check. A genuinely correct fix would make `resolve_leaf` check path-shape *as each value is produced* (e.g. hooking into `push_owned_values`'s per-value callback, short-circuiting on the first non-path-shaped value) instead of gating behavior on a post-hoc `values.len()` bucket.

## Why this is separate from #861

#861 (PR #886) fixed the `values.len() == 1` case specifically: an already-produced single value no longer gets erased by a later escape. That fix is confirmed complete and correct for every case tested, including the Halt-safety follow-up found during its own review. This issue is about the adjacent, structurally different `values.len() == 0` (already correct) and `values.len() >= 2` cases, which #861 explicitly declined to touch (see its PR description's "Scope note" and inline code comments).

## Suggested fix direction

Restructure `resolve_leaf`'s fallback to check path-shape per value as `expr` is evaluated (a streaming/short-circuiting check), rather than materializing the full output list first. This would fix both the pre-existing message-shape divergence and the discarded-escape-content/side-effect issue with one mechanism, and would let `resolve_leaf` finally match real jq's actual per-output-checked laziness instead of approximating it with length-based special cases.

newhoggy:
## Summary

`resolve_leaf`'s non-primitive fallback (`src/jq/eval.rs`) raised a bespoke "Cannot use a computed index after a multi-output path component" message whenever a non-path-shaped `path(...)` argument produced 2+ outputs (e.g. `path(range(3))`), instead of the same `#530` "Invalid path expression with result \<v\>" wording its single-output sibling already uses. Real jq checks path-shape per output as it streams and raises on the first failure alone, so its message always names that first value regardless of how many outputs there would have been — now matched exactly:

```
$ echo null | jq -c 'path(range(3))'
jq: error (at <stdin>:1): Invalid path expression with result 0
$ echo null | succinctly jq -c 'path(range(3))'   # before this fix
jq: error (at <stdin>:1): Cannot use a computed index after a multi-output path component
```

Updates the one existing test that pinned the old message (`test_unsupported_path_prefixes_report_rather_than_misfire`, an assignment-target repro where real jq actually uses a third, still-unmatched "near attempt to access element ..." wording — a pre-existing, unrelated divergence that already affects the single-output case identically, not introduced or fixed here).

Doesn't touch the underlying eager evaluation itself — a later output's side effect can still fire even though its own error is never the one reported (`path(paths(if . == 3 then (stderr, true) else true end))` on `[1,2,3]` still writes `3` to stderr on this branch, where real jq never reaches that candidate at all). That needs genuine per-output laziness, the same architectural gap #820 tracks and two other instances found this session (#977, #986) also trace back to. Filed as #987.

## Test plan

- [x] 2 new regression tests (`path(range(3))`, `path(paths(...))` with a container-valued first output) verified directly against jq 1.7.1
- [x] `omni-dev coverage diff`: 100% patch coverage (1/1 new line)

Fixes #891


newhoggy:
Mandatory review completed — 10 finder angles, several independently converging on the same real findings:

1. **The two match arms became byte-identical** (3 angles: simplification, reuse, altitude-adjacent) — merged into one, dropping the now-redundant `1 =>` guard.
2. **Undocumented, untested `?`/`try`/`catch`/`del(...?)` fix** (3 independent angles: cross-file tracer, removed-behavior auditor, line-by-line scan, all verified live against pre-PR and jq 1.7.1) — `EvalError::is_invalid_path_expression()` is a string-prefix check, and the multi-output case now correctly matches it, so `?` no longer silently swallows the error and `catch` no longer wrongly runs its handler on the caught-then-still-invalid value. Added 2 regression tests pinning this (`?` and `try`/`catch` for `path()`, plus `del(...?)`).
3. **Two stale doc pages** (`docs/reference/jq-language.md`, `docs/compliance/jq/limitations.md`) still described the old bespoke message as current behavior — updated both.
4. **Missing tracking issue** (altitude angle) — the PR's own comment acknowledged a remaining "near attempt to access element..." wording gap for the assignment-target case, but per this file's own convention (every other deferred divergence nearby has a filed issue — #845→#980, #861→#843, this PR's own #987), that gap had none. Filed as #989.

One suggestion (reuse `recurse_untracked_error` for the merged arm) was investigated and not applied — that helper's name and doc comment scope it specifically to #843's recurse-family untracked-value case (its only call site), and reusing it here for an unrelated purpose would misuse a semantically-scoped helper rather than genuinely deduplicate.

Second commit pushed with the arm merge, 2 new regression tests, and both doc updates. All local verification green (build, full suite, both clippy passes, 100% patch coverage).

Issue #943, opened 2026-08-15
jq: bare test/match/capture with a 2-element array pattern unpacks it as [regex, flags]


newhoggy:
**Severity: Low**

## Summary

Real jq's bare (1-arg) `test`/`match`/`capture` accepts a 2-element array as the pattern argument and unpacks it as `[regex, flags]` — i.e. `test(["a","i"])` behaves exactly like `test("a"; "i")`. This is undocumented in jq's manual as far as I can find, but is real, observable behavior. succinctly doesn't implement this at all: any array pattern (valid `[regex, flags]` shape or not) falls into the plain non-string-pattern error path.

## Repro (verified live against jq 1.7.1)

```
$ echo '"abc"' | jq -c 'test(["a","i"])'
true
$ echo '"abc"' | succinctly jq -c 'try test(["a","i"]) catch .'
"array not a string or array"           # wrong — should behave like test("a";"i")

$ echo '"abc"' | jq -c 'try test([1,2]) catch .'
"number (1) is not a string"            # unpacks: index 0 (1) as pattern, index 1 (2) present as flags
                                         # → hits the *flagged*-form wording (#937), not the bare one
$ echo '"abc"' | succinctly jq -c 'try test([1,2]) catch .'
"array not a string or array"           # wrong on two counts: doesn't unpack, and uses the wrong
                                         # (bare-form) wording even by its own non-unpacking logic
```

Same for `match`/`capture`.

## Scope note

This is a **pre-existing** gap — confirmed present identically on `main` before #937/#941 (those PRs only changed which wording a *non-array* bad pattern gets; they didn't touch array-pattern handling at all). Found incidentally while reviewing #941's array/object pattern-preview sweep. Not a regression, and #941 doesn't need to fix it — but it means the jq-error-probes corpus's "matches jq" claim doesn't actually cover array-shaped patterns for these three builtins, since no probe exercises this path.

## Suggested fix

[...]

PR #990, opened 2026-08-15
fix(jq): non-regex test() fallback unpacks a bare array pattern

### LINKED ISSUE #956: jq: non-regex builtin_test fallback doesn't unpack bare array patterns (#943 follow-up)
## Background

#943 (fixed in #952) taught `test`/`match`/`capture`'s *bare* form (no explicit flags argument) to unpack a non-empty array pattern as `[pattern, flags, ...ignored]`, matching real jq 1.7.1's undocumented behavior. That fix lives in `eval_regex_pattern_and_flags` in `src/jq/eval.rs`, which is entirely gated behind `#[cfg(feature = "regex")]`.

## The gap

The non-regex substring-match fallback for bare `test(re)` — `builtin_test` in `src/jq/eval.rs` (used only when the `regex` feature is compiled out) — was not updated and still rejects a bare array pattern outright.

Confirmed live: building with `--no-default-features --features std` (no `regex`) and evaluating `"abc" | test(["a","i"])` returns `Error("array not a string or array")`, whereas the same query with `regex` enabled returns `true`.

`match`/`capture` don't have this asymmetry — they don't exist at all without `regex` (their non-regex fallback is a blanket `"regex feature not enabled"` error regardless of pattern shape), so this is specific to bare `test`.

## Severity / scope

Low severity and narrow: unreachable via the shipped CLI binary, since `Cargo.toml`'s `cli` feature unconditionally pulls in `regex` (Cargo features are additive-only). Only reachable by a downstream crate depending on `succinctly` as a library with default features (no `cli`, no explicit `regex`).

Filed as a follow-up rather than folded into #952/#943, since it's a pre-existing architecture gap (the non-regex fallback is already a documented approximation of full regex semantics), not a regression introduced by that fix.

Found during `/code-review` on PR #952.

newhoggy:
## Summary

`builtin_test` — the non-regex fallback for bare `test(re)`, only reachable when the `regex` feature is compiled out (the shipped CLI always enables it via `cli`'s feature dependencies, so this only affects a downstream crate depending on `succinctly` as a library with default features) — rejected a bare array pattern outright:

```
$ echo '"abc"' | succinctly-no-regex jq -c 'test(["a","i"])'   # before this fix
Error("array not a string or array")
```

The regex-enabled path (`eval_regex_pattern_and_flags`) already unpacks a non-empty array pattern as `[pattern, flags, ...ignored]` for the bare form (#943, fixed in #952) — confirmed real, if undocumented, jq 1.7.1 behavior. This fallback wasn't updated at the time, since it's a separate function gated on the opposite `cfg`.

Mirrors #943's exact rule (only the bare form, only non-empty arrays, extra elements past index 1 ignored, a 1-element array supplies no flags), but the unpacked flags element is then discarded regardless of value — this fallback does plain substring matching with no flag support at all, an existing, documented approximation of full regex semantics.

## Test plan

- [x] `cargo test --no-default-features --features std` (full suite under the feature combination that actually exercises this fix, 0 failures)
- [x] New `#[cfg(not(feature = "regex"))]`-gated unit test (`test_builtin_test_unpacks_bare_array_pattern_956`) covering: 2-element unpack, extra-elements-ignored, 1-element (no flags), non-matching pattern, non-string unpacked element's error wording, empty-array's error wording, and `optional` (`?`) suppression — verified directly via `cargo test --no-default-features --features std --lib -- test_builtin_test_unpacks_bare_array_pattern_956`
- [x] `omni-dev coverage diff`: reports zero new executable lines, expected — the report is generated with `regex` enabled (matching CI's own convention), so this entirely `cfg(not(feature = "regex"))`-gated diff is compiled out of that build and can't be instrumented by it; the new code path is verified by the no-regex test run above instead

Fixes #956

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -6044,6 +6044,61 @@ fn builtin_implode<W: Clone + AsRef<[u64]>>(
     }
 }
 
+/// Unpack a `test`/`match`/`capture` bare-form (no explicit flags argument)
+/// pattern if it's a non-empty array, as `[pattern, flags, ...ignored]` —
+/// real, if undocumented, jq 1.7.1 behavior (#943): `test(["a","i"])`
+/// behaves like `test("a";"i")`, extra elements past index 1 are silently
+/// ignored, and a 1-element array supplies `None` for flags. Only a
+/// genuinely *empty* array (no element 0 to unpack) is left alone.
+///
+/// Shared by the regex-enabled `eval_regex_pattern_and_flags` and the
+/// non-regex `builtin_test` fallback (#956) — the two are mutually
+/// exclusive, `cfg`-gated on opposite sides of the `regex` feature and can
+/// never both compile into the same binary, but this helper isn't gated on
+/// either, so both reach the one definition instead of two independently
+/// hand-written copies of the same boundary drifting apart (CLAUDE.md:
+/// "duplicated predicates diverge silently").
+///
+/// Returns `(pattern, flags, was_unpacked)` — `was_unpacked` tells the
+/// caller to use the array-unpacked error wording
+/// (`EvalError::is_not_a_string`) rather than the plain
+/// `not_string_or_array` one if `pattern` still isn't a string after this.
+fn unpack_bare_array_pattern(pattern: OwnedValue) -> (OwnedValue, Option<OwnedValue>, bool) {
+    match pattern {
+        OwnedValue::Array(items) if !items.is_empty() => {
+            let mut items = items.into_iter();
+            let pattern = items.next().expect("checked non-empty above");
+            let flags = items.next();
+            (pattern, flags, true)
+        }
+        other => (other, None, false),
+    }
+}
+
+/// Validate a `test`/`match`/`capture` flags value once it's been evaluated
+/// (or unpacked from an array pattern, `unpack_bare_array_pattern`):
+/// missing is `None`, `null` is treated as "no flags" (an empty string), a
+/// string is used as-is, and anything else is the same `is_not_a_string`
+/// error every pattern-position type mismatch in this family already uses.
+///
+/// Shared by `eval_regex_pattern_and_flags` and the non-regex `builtin_test`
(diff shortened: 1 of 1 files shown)


newhoggy:
Mandatory review completed — 8 finder angles, 5 of them independently converging on the same core finding.

**The consensus finding**: my non-regex `builtin_test` fallback hand-rolled its own copy of #943's array-unpacking rule instead of sharing it with `eval_regex_pattern_and_flags` — and the copy itself was needlessly convoluted (a boolean flag double-checking the same `Array` match twice, plus an `.expect()` whose soundness depended on a fact computed 10 lines earlier). Extracted two small, ungated helpers (`unpack_bare_array_pattern`, `validate_regex_flags`) that both `cfg`-exclusive functions now call, so the #943/#956 boundary has one definition instead of two hand-maintained copies.

**A real correctness bug found while extracting** (cross-file-tracer angle, independently corroborated by direct oracle testing): the fallback unpacked the array's flags slot but never *validated* it — `test(["a", 5])` (a non-string, non-null flags element) silently succeeded instead of raising `"number (5) is not a string"` like both real jq and the regex-enabled path do. Confirmed live against jq 1.7.1 before fixing. This wasn't in scope of the review's own stated ask (duplication) but is exactly the kind of thing duplicated-and-never-cross-checked code hides — the regex-enabled sibling validates this correctly, and my copy quietly didn't.

**Also fixed**: a `clippy::needless_raw_string_hashes` in the new test, invisible to every CI clippy job (they all enable `regex`, compiling this `cfg(not(feature = "regex"))` test out) but real under a plain local `cargo clippy --features std,simd,serde`.

Two new regression tests added for the flags-validation fix (malformed flags still errors; `null` flags is valid and still a no-op). Second commit pushed. All local verification green: both feature configurations' full test suites, both required clippy passes, and — since the shared helpers are ungated and now reachable through the regex-enabled path too — 100% patch coverage under the standard CI coverage build (previously 0 lines were visible to it at all, since the whole diff was `cfg`-gated off).

Issue #991, opened 2026-08-15
jq: remaining EvalError::type_error wording gaps found during #929's audit (join, sub replacement, limit, combinations, @csv nested-array validation)


newhoggy:
**Severity: Low**

## Summary

Split out from #929's systematic sweep of `EvalError::type_error` call sites. #929 fixed the clear, contained cases (`map_values`/`min_by`/`max_by`/`unique_by` → `cannot_iterate`, `nth`'s two sites → `cannot_index`, `@csv`/`@tsv`/`@sh` → new dedicated wording, `strptime`/`fromdate` → a shared new wording) and documented the "jq accepts, succinctly refuses" cases (`@uri`/`@base64`/`@base64d` on non-string) in `docs/compliance/jq/limitations.md`. This issue tracks what's left, each investigated but deferred for a specific reason.

## 1. `join`/`sub`'s replacement function — real jq wording exists but comes from internal accumulator state, not the argument directly

```
$ echo '[1,2]' | jq -c 'try join(1) catch .'
"string (\"1\") and number (1) cannot be added"
$ echo '[1,2]' | succinctly jq -c 'try join(1) catch .'
"expected string, got non-string"
```

```
$ echo '"a"' | jq -c 'try sub("a"; 5) catch .'
"string (\"\") and number (5) cannot be added"
$ echo '"a"' | succinctly jq -c 'try sub("a"; 5) catch .'
"expected number, got non-number"  # approximate -- actual succinctly wording not re-checked against this exact repro
```

Both are `EvalError::binary_op`-shaped messages (`<left> and <right> cannot be added`), but the operands jq names are **not** simply "the array/string being joined and the bad argument" — they're whatever the accumulator holds at the point jq's own `reduce`-based definition (`join`) or replacement-function pipeline (`sub`) actually fails, which depends on internal implementation state (e.g. `join`'s first accumulator value is `""`, converted from the first element, then `+`'d against the separator or next element). Reproducing this exactly likely needs restructuring these builtins' error paths to fail at the same *point* jq's own definition does, not just swapping the message constructor — a real behavioral change, not a wording swap. Needs its own investigation into exactly where succinctly's `join`/`sub`-replacement implementations diverge structurally from jq's own `reduce`/pipeline-based definitions.

## 2. `limit`/`combinations`/`combinations(n)` — real jq wording is an accidental artifact of a jq-language macro definition

```
$ echo '[1,2,3]' | jq -c 'try limit("x"; .[]) catch .'
[...]

PR #992, opened 2026-08-15
fix(jq): correct EvalError::type_error wording at 10 call sites

### LINKED ISSUE #929: jq: audit remaining EvalError::type_error call sites against real jq wording
**Severity: Low**

## Summary

#926/#927 fixed one instance (the regex pattern argument) of a broader pattern: `EvalError::type_error(expected, got)` — documented in its own doc comment as reserved for "error sites that have no jq counterpart" — is used at sites that actually *do* have a real jq wording, producing a message that doesn't match jq at all. `/code-review`'s Altitude angle on #927 found this recurs well beyond the regex family, several times with jq's real wording being a *different existing constructor* that's already unused-but-available.

```
$ echo '5' | jq -c 'try min_by(.) catch .'
"Cannot iterate over number (5)"
$ echo '5' | succinctly jq -c 'try min_by(.) catch .'
"expected array, got number"

$ echo '[1,2]' | jq -c 'try join(1) catch .'
"array ([1,2]) and number (1) cannot be added"
$ echo '[1,2]' | succinctly jq -c 'try join(1) catch .'
"expected string, got non-string"

$ echo '[1,2]' | jq -c 'try flatten("x") catch .'
[1,2]
$ echo '[1,2]' | succinctly jq -c 'try flatten("x") catch .'
"expected number, got non-number"
```

(`flatten("x")` is a different kind of gap — real jq doesn't even error here, it silently no-ops; worth confirming during the audit rather than assuming a wording fix is all that's needed.)

The reviewer identified `min_by`/`max_by`/`unique_by` (should likely use the already-existing `EvalError::cannot_iterate`), `join`/`limit` (should likely use the already-existing `EvalError::binary_op`), and `flatten` as concrete additional sites, but explicitly did not treat that as an exhaustive list — this issue is to do the systematic sweep, not to enumerate every site by hand ahead of time.

## Suggested approach

Grep `src/jq/eval.rs` for every `EvalError::type_error(` call site, and for each one:
1. Reproduce the same filter/input shape against pinned jq to see if jq actually errors, and with what wording.
2. If jq has real wording, either reuse an existing named constructor (checking the vocabulary table in `docs/compliance/jq/limitations.md` first) or add a new one following the `#926` pattern.
3. If jq doesn't error at all (like `flatten("x")` above), that's a different class of gap — succinctly refuses something jq accepts — and belongs in `docs/compliance/jq/limitations.md`'s "Where succinctly errors and jq does not" table instead of a wording fix.
4. Add oracle-verified probes to `tests/data/jq-error-probes.tsv` for every corrected wording, so the corpus keeps this from drifting back.

This is intentionally an investigation/audit issue rather than a bounded fix — the actual scope (how many sites, how many need new constructors vs. reusing existing ones vs. a "jq doesn't error" note) isn't known until the sweep is done.

Split from #927's review, and adjacent to #928 (the same `type_error` misuse pattern, but for the regex builtins' flags argument specifically — narrower and already scoped there).

newhoggy:
## Summary

`EvalError::type_error` is documented as reserved for error sites with no jq counterpart, but several call sites in `src/jq/eval.rs` had a real jq wording they didn't match (found in #926/#927's review, this issue is the systematic sweep). Every fix below was verified live against pinned jq 1.7.1.

**Wording fixes** (reusing existing constructors or adding narrow new ones):
- `map_values`/`min_by`/`max_by`/`unique_by` on a non-object/array input → `EvalError::cannot_iterate` ("Cannot iterate over number (5)"), matching how real jq's own `.[]`-based definitions fail.
- `nth(n)`'s two type checks (non-numeric `n`, non-array input) → `EvalError::cannot_index`, matching jq's own `.[n]` definition exactly (a non-numeric `n` names itself as the bad key; a non-array input names the successfully-parsed `n` as the key).
- `@csv`/`@tsv` on a non-array, and `@sh` on an object → two new dedicated constructors (`cannot_be_dsv_formatted`, `cannot_be_shell_escaped`) matching jq's exact wording.
- `strptime`'s two type checks and `fromdate`'s share a new `strptime_requires_string` constructor — real jq validates both the input and format argument with **one combined check**, raising the identical message regardless of which side is wrong (confirmed live for all three combinations, except the one where jq itself crashes with a C assertion failure — not reproduced, since a clean error beats a crash).

**Documented, not code-fixed** (`docs/compliance/jq/limitations.md`'s existing "where succinctly errors and jq does not" table): `@uri`/`@base64` auto-`tostring` a non-string argument in real jq rather than refusing it; `flatten` silently ignores a non-numeric depth rather than erroring.

**Deferred, filed as #991**: `join`/`sub`'s replacement function (real jq's wording depends on internal accumulator state from a `reduce`-based definition, not a direct argument — needs restructuring, not a wording swap), `limit`/`combinations`/`combinations(n)` (jq's wording is an accidental artifact of a jq-language macro definition, not a deliberately authored message — worth a second opinion on whether it's even worth chasing), and a **missing** (not just wrongly-worded) `@csv` validation gap found along the way (`[[1,2]] | @csv` silently succeeds in succinctly, where jq rejects a nested-array row element).

## Test plan

- [x] New regression test (`test_type_error_sites_match_real_jq_wording_929`) covering all 10 fixed call sites, each assertion verified directly against jq 1.7.1
- [x] `omni-dev coverage diff`: 100% patch coverage (66/66 new lines)

Fixes #929

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -299,13 +299,13 @@ A probe is only admitted to the corpus if jq errors on it, so the corpus is blin
 opposite divergence: a filter that jq answers with a value and succinctly refuses. Those
 have to be recorded here.
 
-| Filter            | Input         | jq                          | succinctly                            |
-|-------------------|---------------|-----------------------------|---------------------------------------|
-| `.[1:2] = ["x"]`  | `null`        | `["x"]`                     | `Cannot index null with object`       |
-| `.[1:2] \|= ["x"]`| `null`        | `["x"]`                     | `Cannot index null with object`       |
-| `@uri`            | `[1,2]`       | `"%5B1%2C2%5D"`             | `expected string, got array`          |
-| `@base64`         | `5`           | `"NQ=="`                    | `expected string, got number`         |
-| `flatten("x")`    | `[1,2]`       | `[1,2]` (ignores non-integer depth) | `expected number, got non-number` |
+| Filter             | Input   | jq                                  | succinctly                        |
+|--------------------|---------|-------------------------------------|-----------------------------------|
+| `.[1:2] = ["x"]`   | `null`  | `["x"]`                             | `Cannot index null with object`   |
+| `.[1:2] \|= ["x"]` | `null`  | `["x"]`                             | `Cannot index null with object`   |
+| `@uri`             | `[1,2]` | `"%5B1%2C2%5D"`                     | `expected string, got array`      |
+| `@base64`          | `5`     | `"NQ=="`                            | `expected string, got number`     |
+| `flatten("x")`     | `[1,2]` | `[1,2]` (ignores non-integer depth) | `expected number, got non-number` |
 
 [#929](https://github.com/rust-works/succinctly/issues/929) found these while auditing
 `EvalError::type_error` wording: real jq's `@uri`/`@base64` (and every other format string
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -303,6 +303,19 @@ have to be recorded here.
 |-------------------|---------------|-----------------------------|---------------------------------------|
 | `.[1:2] = ["x"]`  | `null`        | `["x"]`                     | `Cannot index null with object`       |
 | `.[1:2] \|= ["x"]`| `null`        | `["x"]`                     | `Cannot index null with object`       |
+| `@uri`            | `[1,2]`       | `"%5B1%2C2%5D"`             | `expected string, got array`          |
+| `@base64`         | `5`           | `"NQ=="`                    | `expected string, got number`         |
+| `flatten("x")`    | `[1,2]`       | `[1,2]` (ignores non-integer depth) | `expected number, got non-number` |
+
+[#929](https://github.com/rust-works/succinctly/issues/929) found these while auditing
+`EvalError::type_error` wording: real jq's `@uri`/`@base64` (and every other format string
+except `@csv`/`@tsv`/`@sh`) auto-`tostring`s a non-string argument before formatting rather
+than refusing it outright, and `flatten`'s depth argument is silently ignored if it isn't a
+number rather than validated. `@base64d` is a related but distinct case — jq *does* still
+error on `5 | @base64d` (`string ("5") trailing base64 byte found`), just from attempting to
+base64-decode the auto-stringified `"5"` rather than refusing the number up front, so it
+isn't purely a "jq doesn't error" gap; matching it needs the same underlying
+auto-`tostring`-first change `@uri`/`@base64` do, not just a different error message.
 
 Both rows are a slice write walking a path *in place*, and the gap is the same
 auto-vivification jq performs everywhere else it writes through a path: `null` grows into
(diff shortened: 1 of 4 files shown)


newhoggy:
Review response pushed in c162aad92.

Verified all 7 findings against the pinned jq 1.7.1 oracle; 6 were real and are now fixed, 1 (the `@csv`/`@tsv` nested-array-element gap) was already tracked as #991 item 3, so no new action there.

**Fixed:**
1. **`nth(n)` semantics** — rewrote `builtin_nth` to delegate to `index_one`, the same generic indexing operator `.[$k]` itself uses (jq really does define `nth(n)` as `n as $n | .[$n]`). This one change fixes three bugs at once:
   - `nth("x")` on an object with a missing key now returns `null` instead of erroring (`{"a":1} | nth("x")` → `null`, matching jq).
   - `nth(1.5)` now truncates toward zero and indexes normally instead of being rejected (`[1,2,3] | nth(1.5)` → `2`).
   - `n` is now evaluated before the null short-circuit, so `null | nth(error("boom"))` raises `"boom"` again instead of silently returning `null`.
2. **`min_by`/`max_by`/`unique_by` on an object input** — these now raise jq's real two-operand wording, pairing the input with the `map([f])` array jq itself computes on it (e.g. `{"a":3,"b":1} | min_by(.)` → `object ({"a":3,"b":1}) and array ([[3],[1]]) cannot be iterated over`), instead of the generic single-value "Cannot iterate over" a bare scalar gets. Added a shared `map_bracketed_over_object_fields` helper (used by all three) and `EvalError::pair_cannot_be_sorted` (unique_by's distinct suffix; min_by/max_by reuse the existing `pair_cannot_be_iterated`).
3. **`@sh`'s `shell_quote_value`** — now returns `Result` and validates nested array/object elements the same way the top-level value is validated (`[{"a":1}] | @sh` now errors instead of silently emitting `""`).
4. **Markdown table width** — fixed the `flatten("x")` row's column-width violation in `docs/compliance/jq/limitations.md`.

**Deferred (not fixed, with reason):**
- The pre-existing `nth`-skips-evaluating-`n`-on-null-input bug is fixed as a side effect of item 1 above (reordering `n`'s evaluation ahead of the null check was needed for the float/object fixes anyway), so nothing was deliberately left out here.
- Two internal `if optional { QueryResult::None }` branches in the new object-input arms (min_by/max_by/unique_by) are structurally unreachable via real jq syntax — confirmed by checking full-suite (non-patch) coverage: the *pre-existing* scalar-input `_ if optional => QueryResult::None` arm these mirror is *also* uncovered on `main` today. Since #693, `E?` evaluates its inner expression with the *ambient* `optional` rather than forcing `true`, so a bare `min_by(.)?` is caught externally by `eval_try`, never by this function's own `optional` parameter. Kept for signature consistency with every other builtin in this file, not for reachability.

Full local verification: build, full test suite (`cargo test --features cli,simd,regex,serde`, 2474+ tests), non-regex path (`cargo test --no-default-features --features std --lib`), yq golden conformance, both required clippy configs, and `omni-dev coverage diff` (96.39% patch coverage on this commit — the 6 uncovered lines are the structurally-dead branches noted above). All new cases re-verified live against jq 1.7.1.

PR #1000, opened 2026-08-15
fix(jq): group_by/sort_by object-input wording matches jq's pairing error

### LINKED ISSUE #995: jq: group_by/sort_by on an object input use the wrong two-value error wording
## Summary

Found while auditing `min_by`/`max_by`/`unique_by`'s object-input error wording for #929's
code review. `group_by`/`sort_by` have the identical bug but are **not** touched by #929's
diff (they already used `EvalError::cannot_iterate`/`cannot_be_sorted` before that PR, not
`type_error`, so they were out of #929's original grep-based scope).

Real jq computes `min_by`/`max_by`/`unique_by`/`sort_by`/`group_by` all the same way: they
unconditionally build `map([f])` first (which succeeds on an object input too, since bare
`.[]` iterates an object's values), then pair the *original input* with that *computed
array* in a two-operand type error if the input itself isn't an array. succinctly's
`group_by`/`sort_by` instead fall through to the generic single-value `cannot_iterate`
wording a bare scalar gets, which is wrong for an object input specifically.

## Repro (jq 1.7.1 vs `succinctly jq`)

```
$ echo '{"a":3,"b":1}' | jq 'group_by(.)'
jq: error (at <stdin>:1): object ({"a":3,"b":1}) and array ([[3],[1]]) cannot be sorted, as they are not both arrays

$ echo '{"a":3,"b":1}' | jq 'sort_by(.)'
jq: error (at <stdin>:1): object ({"a":3,"b":1}) and array ([[3],[1]]) cannot be sorted, as they are not both arrays

$ echo '{"a":3,"b":1}' | succinctly jq 'group_by(.)'
jq: error (at <stdin>:1): Cannot iterate over object ({"a":3,"b":1})

$ echo '{"a":3,"b":1}' | succinctly jq 'sort_by(.)'
jq: error (at <stdin>:1): Cannot iterate over object ({"a":3,"b":1})
```

The computed array in jq's message tracks the actual filter, not a fixed placeholder:

```
$ echo '{"a":3,"b":1}' | jq 'group_by(. + 100)'
jq: error (at <stdin>:1): object ({"a":3,"b":1}) and array ([[103],[101]]) cannot be sorted, as they are not both arrays
```

## Fix

`group_by`/`sort_by` (`src/jq/eval.rs`) should share the `map_bracketed_over_object_fields`
helper and `EvalError::pair_cannot_be_sorted` constructor #929's PR (#992) just added for
`unique_by`'s identical case — both already exist and match this wording exactly.

## Scalar-input case is already correct

Only the *object*-input case is wrong. A non-object, non-array scalar input already gets
the right single-value wording (`5 | group_by(.)` → `Cannot iterate over number (5)`,
matching jq), so that arm needs no change.

newhoggy:
## Summary

- `group_by`/`sort_by` on an object input raised the generic single-value "Cannot
  iterate over" wording, instead of jq's real two-operand wording that pairs the
  input with the `map([f])` array jq itself computes on it (`.[]` iterates an
  object's values just as readily as an array's, so that computation succeeds even
  though the input itself isn't an array).
- This is the exact same bug #929/#992 already fixed for `min_by`/`max_by`/
  `unique_by`, which share this identical computation — `group_by`/`sort_by` were
  simply outside that issue's scope (they never called `EvalError::type_error`,
  which is what #929 grepped for).
- Reuses the existing `map_bracketed_over_object_fields` helper and
  `EvalError::pair_cannot_be_sorted` constructor #992 added, unchanged.

## Repro (jq 1.7.1 vs succinctly, before this fix)

```
$ echo '{"a":3,"b":1}' | jq 'group_by(.)'
jq: error (at <stdin>:1): object ({"a":3,"b":1}) and array ([[3],[1]]) cannot be sorted, as they are not both arrays

$ echo '{"a":3,"b":1}' | succinctly jq 'group_by(.)'
jq: error (at <stdin>:1): Cannot iterate over object ({"a":3,"b":1})
```

Fixes #995.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 2475+ tests)
- [x] `cargo test --no-default-features --features std --lib` (non-regex path)
- [x] `cargo test --features cli,simd,regex,serde --test yq_golden_tests`
- [x] Every wording change verified live against the pinned jq 1.7.1 oracle
- [x] New regression test (`test_group_by_sort_by_object_wording_995`) covers object input, filter-tracking in the computed array, scalar input (unchanged wording), `?`-suppression, and error propagation through that suppression

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -4349,7 +4349,31 @@ fn builtin_group_by<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
 
             QueryResult::Owned(OwnedValue::Array(groups))
         }
+        // #995: group_by shares min_by/max_by/unique_by's `map([f])`-pairing
+        // computation (see `map_bracketed_over_object_fields`) and
+        // unique_by's own suffix -- confirmed live: `{"a":3,"b":1} |
+        // group_by(.)` raises `object ({"a":3,"b":1}) and array
+        // ([[3],[1]]) cannot be sorted, as they are not both arrays` in jq
+        // 1.7.1.
+        StandardJson::Object(fields) => {
+            let original = to_owned(&StandardJson::Object(fields));
+            let computed = match map_bracketed_over_object_fields::<W, S>(f, fields, optional) {
+                Ok(v) => v,
+                Err(result) => return result,
+            };
+            if optional {
+                QueryResult::None
+            } else {
+                QueryResult::Error(EvalError::pair_cannot_be_sorted(
+                    &original,
+                    &OwnedValue::Array(computed),
+                ))
+            }
+        }
         _ if optional => QueryResult::None,
+        // #995: same "Cannot iterate over" wording min_by/max_by/unique_by's
+        // own scalar arm uses -- confirmed live: `5 | group_by(.)` raises
+        // "Cannot iterate over number (5)" in jq 1.7.1.
         _ => QueryResult::Error(EvalError::cannot_iterate(&to_owned(&value))),
     }
 }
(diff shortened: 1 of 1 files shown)


newhoggy:
Review response pushed in 6845947bc.

Verified the review's finding against the pinned jq 1.7.1 oracle: bare `unique` (real jq defines it as `unique_by(.)`) had the identical object-input pairing bug, confirmed live — `{"a":3,"b":1} | unique` raised `Cannot iterate over object (...)` instead of jq's real `object (...) and array ([[3],[1]]) cannot be sorted, as they are not both arrays`. It was missed by both #929's grep-based audit (which only found `EvalError::type_error` call sites) and this issue's own initial scope, since `unique` takes no `f` argument at all and so didn't fit the `min_by`/`max_by`/`unique_by`/`group_by`/`sort_by` pattern being ported.

**Fixed:**
1. **`unique`'s missing fix** — added an `S: EvalSemantics` bound to `builtin_unique` (it didn't need one before) and a `StandardJson::Object` arm that reuses the same machinery via `f = &Expr::Identity` (matching real jq's `unique_by(.)` definition).
2. **Duplication across 5 (now 6) call sites** — extracted `object_pair_type_error` (parameterized by which `EvalError` constructor to use: `pair_cannot_be_iterated` for `min_by`/`max_by`, `pair_cannot_be_sorted` for `unique_by`/`group_by`/`sort_by`/`unique`), replacing all 6 near-identical ~15-line blocks with one line each.
3. **Deferred the `original` clone** — moved inside the helper's `!optional` branch, so the `?`-suppressed and inner-`f`-error paths no longer pay for a clone they discard.

**Not changed:**
- No table-driven/oracle-enumeration test was added — the review's suggestion is reasonable for the long term, but building one is a larger, separate effort than this issue's scope; the existing hand-written regression tests (now covering all 6 call sites) are sufficient for this fix.
- `sort_by`'s test previously lacked the "computed array tracks the actual filter" assertion `group_by`'s had — added the missing `sort_by(. + 100)` case to close that asymmetry.

Full local verification re-run after the changes: build, full test suite (2476 tests), non-regex path, yq golden conformance, both required clippy configs — all clean. Every wording change (including `unique`'s) re-verified live against jq 1.7.1.

PR #1002, opened 2026-08-15
fix(jq): add recursion-depth guard against adversarial JSON nesting

### LINKED ISSUE #998: yq: JSON-input DOM conversion has no recursion-depth guard (stack-overflow on adversarial deep nesting)
## Summary

Neither the JSON→\`OwnedValue\` materialization path (\`generic_to_owned\`/\`to_owned\`, shared
by \`succinctly jq\` and \`succinctly yq\`) nor the new \`canonicalize_json_numbers\` pass added
by #978/PR #994 has a recursion-depth guard. Both walk the parsed tree recursively with no
depth limit, so extremely deeply-nested adversarial JSON input (thousands of nested arrays
or objects) can overflow the stack.

## How this surfaced

Found during code review of PR #994 (issue #978). \`generic_to_owned\`'s lack of a depth
guard is pre-existing and out of that PR's scope, but #978 adds a *second* unguarded
recursive pass (\`canonicalize_json_numbers\`) over the same tree for \`--input-format json\`
(or auto-detected \`.json\` input), measurably shrinking the crash threshold: a reviewing
agent measured roughly 10000–10500 levels of nesting before stack overflow with both passes
active, versus roughly 15000–16000 with only the original pass — a ~30-35% reduction in
adversarial-input crash resistance, specific to JSON input on \`succinctly yq\`.

\`src/json/validate.rs\` already has a precedent for this: \`MAX_NESTING_DEPTH = 128\`, but
it's only enforced by the separate \`--validate\` strict-RFC-8259 path, not by ordinary
parsing/evaluation.

## Suggested fix

Add a shared depth-limit check to the recursive walk(s) — ideally by giving
\`generic_to_owned\`/\`to_owned\` (src/jq/eval_generic.rs) a depth parameter/guard that both
\`succinctly jq\` and \`succinctly yq\` inherit, rather than only patching the new
\`canonicalize_json_numbers\` pass in isolation (which would still leave the older, deeper
pass unguarded and only partially close the gap). Needs a decision on the right limit and
the right failure mode (jq-style error vs. hard process abort) — worth scoping as its own
small design discussion rather than folding into an unrelated bug-fix PR.

## Scope note

Not blocking #978/PR #994 — that PR's canonicalization logic is correct; this is a
pre-existing robustness gap (this PR only narrows the existing margin, it doesn't
introduce the vulnerability class) best fixed as a standalone hardening change covering
both jq and yq.

newhoggy:
## Summary

- Neither `eval_generic::to_owned`/`to_owned_cursor`/`to_owned_with_comments` (the DOM-materialization path shared by `jq` and `yq`) nor `jq_runner`'s `print_json` (the lazy identity-query streaming path, which `to_owned`'s own guard can't reach) had a recursion-depth guard, so deeply nested adversarial JSON input could overflow the stack.
- Confirmed live before this fix: `succinctly jq '.'` and `succinctly yq --input-format json '.'` both aborted with a raw SIGABRT stack overflow on a 200,000-level-deep document.
- `to_owned`/`to_owned_cursor`/`to_owned_with_comments` are called pervasively throughout the core evaluator's hot path (truthiness checks, streaming, comparisons — I initially tried threading a `Result` through their signatures for a clean, catchable error matching real jq's own parse-time depth limit, but that rippled into 55+ call sites deep inside the evaluator's control flow and was too invasive/risky to land safely). A controlled panic closes the actual safety concern instead — a raw stack overflow is undefined behavior, a panic is not — without that blast radius; #998's own text names a hard process abort as an acceptable outcome alongside a clean error.
- `print_json` already returns a `Result` and already threads a `level` parameter (previously used only for indentation), so its own guard is a clean, catchable `anyhow` error with **no signature change at all**.

## Why 256, not 128

The existing YAML parser/JSON validator use `MAX_NESTING_DEPTH = 128` for their own, unrelated guards. Reusing that value broke a pre-existing regression test (`test_walk_deep_nesting_does_not_overflow_the_stack`) that pins `walk(.)` working at depth 200 — that query's output still passes through this same materialization step.

Chosen empirically instead — measured directly on a debug build (larger per-frame stack cost than release): the real crash boundary sits between depth 2000 (safe) and 3000 (overflow) for `to_owned`, and between 600 (safe) and 700 (overflow) for `print_json`. 256 clears the 200 floor with margin and sits comfortably under both boundaries, including headroom for a CI runner with a smaller default stack than the dev machine this was measured on.

## Repro (before this fix)

```
$ succinctly jq '.' deeply-nested-200000-levels.json
thread 'main' panicked... fatal runtime error: stack overflow, aborting  (SIGABRT, exit 134)

$ succinctly yq --input-format json '.' deeply-nested-200000-levels.json
thread 'main' panicked... fatal runtime error: stack overflow, aborting  (SIGABRT, exit 134)
```

After this fix, both reject the adversarial input cleanly (`print_json`: exit 1, `anyhow` error; `to_owned`: exit 101, controlled panic — not a raw stack overflow) while legitimately-nested input under the limit is unaffected.

Fixes #998

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 2477+ tests, including the pre-existing `walk(.)` depth-200 regression test)
- [x] `cargo test --no-default-features --features std --lib` (non-regex path)
- [x] `cargo test --features cli,simd,regex,serde --test yq_golden_tests`
- [x] New library test (`to_owned_panics_past_nesting_depth_limit_998`) pins the exact 255/256 boundary
- [x] New CLI tests (`test_identity_query_rejects_adversarial_nesting_998`/`test_identity_query_accepts_nesting_under_limit_998` in `jq_cli_tests.rs`; `test_json_input_rejects_adversarial_nesting_998`/`test_json_input_accepts_nesting_under_limit_998` in `yq_cli_tests.rs`) exercise both commands end-to-end via the real compiled binary
- [x] Manually verified against a 200,000-level-deep adversarial document and against depth 126/127/128/129 to confirm the exact boundary

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -11,7 +11,9 @@ use std::path::{Path, PathBuf};
 
 use succinctly::dsv::{build_index as build_dsv_index, DsvConfig, DsvRows};
 use succinctly::jq::document::DocumentFields;
-use succinctly::jq::eval_generic::{eval_with_cursor, to_owned as generic_to_owned, GenericResult};
+use succinctly::jq::eval_generic::{
+    eval_with_cursor, to_owned as generic_to_owned, GenericResult, MAX_NESTING_DEPTH,
+};
 use succinctly::jq::{self, format_number_jq_compat, Expr, JqValue, OwnedValue, Program};
 use succinctly::json::light::{JsonCursor, StandardJson};
 use succinctly::json::validate::{self, ValidationError};
@@ -2031,18 +2033,23 @@ impl LiteralFormatter for PreserveFormatter {
 ///
 /// This is the unified printer that handles JSON structure (arrays, objects,
 /// indentation) while delegating literal formatting to the formatter.
-/// Recursion-depth ceiling for [`print_json`] (#998) -- adversarially deep
-/// JSON input (thousands of nested arrays/objects) recurses this writer once
-/// per nesting level; unlike `eval_generic::to_owned`'s own guard (which
-/// this mirrors, same limit -- see that constant's doc comment for how 256
-/// was chosen, including this function's own measured debug-build crash
-/// boundary of depth 600-700), a query like the bare identity `.` never
-/// materializes an `OwnedValue` tree at all (it stays lazy, streaming
-/// straight from the `JsonCursor`), so `to_owned`'s guard never gets a
-/// chance to fire for it -- this is the one recursive step that shape
-/// always goes through, so it needs its own guard.
-const MAX_NESTING_DEPTH: usize = 256;
-
+///
+/// Guarded against adversarially deep JSON input (thousands of nested
+/// arrays/objects, which would otherwise recurse this writer once per
+/// nesting level and overflow the stack) by the shared
+/// [`succinctly::jq::eval_generic::MAX_NESTING_DEPTH`] ceiling -- see that
+/// constant's own doc comment for how 256 was chosen, including this
+/// function's own measured debug-build crash boundary of depth 600-700.
+/// Unlike `eval_generic::to_owned`'s own guard (a panic, since that
+/// function sits too deep in the evaluator's hot path for a `Result`-based
+/// fix), this one is a clean, catchable `anyhow` error: a query like the
+/// bare identity `.` never materializes an `OwnedValue` tree at all (it
+/// stays lazy, streaming straight from the `JsonCursor`), so `to_owned`'s
+/// own guard never gets a chance to fire for it -- this is the one
+/// recursive step that shape always goes through, and it already returns
+/// `Result` and already threads a `level` parameter (previously used only
+/// for indentation, reused here as the depth counter), so there's no reason
+/// to give up the clean failure mode the way `to_owned` had to.
 fn print_json<F, Out, Wrd>(
     out: &mut Out,
     value: &JqValue<'_, Wrd>,
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -2031,6 +2031,18 @@ impl LiteralFormatter for PreserveFormatter {
 ///
 /// This is the unified printer that handles JSON structure (arrays, objects,
 /// indentation) while delegating literal formatting to the formatter.
+/// Recursion-depth ceiling for [`print_json`] (#998) -- adversarially deep
+/// JSON input (thousands of nested arrays/objects) recurses this writer once
+/// per nesting level; unlike `eval_generic::to_owned`'s own guard (which
+/// this mirrors, same limit -- see that constant's doc comment for how 256
+/// was chosen, including this function's own measured debug-build crash
+/// boundary of depth 600-700), a query like the bare identity `.` never
+/// materializes an `OwnedValue` tree at all (it stays lazy, streaming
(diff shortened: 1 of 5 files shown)


newhoggy:
Review response pushed in a536c9d18.

Verified the review's headline finding against the pinned live binary: `--exit-status`/`-e` forces `JqValue::materialize()` on every result before `print_json`'s own guarded output path ever runs, reaching `lazy.rs`'s independent `cursor_to_owned` materializer directly — a second, unguarded recursive tree-walker with the identical shape `eval_generic::to_owned_cursor` had. Confirmed live: `succinctly jq -e -c '.[0]'` on a 200,000-level-deep document raw-stack-overflowed (SIGABRT, exit 134) even with the original PR's guards already in place.

**Fixed:**
1. **`cursor_to_owned` bypass** — added the identical depth-tracking wrapper the other three functions already had.
2. **Duplicated `MAX_NESTING_DEPTH` constant** — `eval_generic::MAX_NESTING_DEPTH` is now `pub` and shared; `jq_runner.rs` imports it instead of carrying an independent copy that could silently drift out of sync.
3. **Triplicated `assert!` block** — extracted a shared `assert_nesting_depth` helper, used by all four now-guarded functions (`to_owned`, `to_owned_cursor`, `to_owned_with_comments`, `cursor_to_owned`).
4. New regression test (`test_exit_status_query_rejects_adversarial_nesting_998`) pins the `-e` bypass specifically.

**Investigated and confirmed real, but filed separately as #1005 rather than folded into this PR:**

The review also surfaced a structurally distinct, much larger vulnerability class: a value **constructed at query-evaluation time** via `reduce`/`foreach`/`while`/`until`/`repeat` (growing one level per loop iteration) can exceed the guard depth with **no adversarial document at all** — plain `null` input is enough — and crashes through several entirely different, still-unguarded recursive functions this PR doesn't touch: `value.rs`'s `to_json`/`to_json_for_reindex`, `compare_values`/`PartialEq`, `eval.rs`'s own separate `to_owned`, `yq_runner.rs`'s `reconcile_presentation`, `output.rs`'s `format_json_impl`, YAML's own `stream_json_value` family, and even `OwnedValue`'s compiler-generated recursive `Drop`. Confirmed live (both independently by me and corroborated by two review angles): `echo null | succinctly yq 'reduce range(50000) as $i (null; [.]) | empty'` raw-stack-overflows with zero adversarial input.

This is out of #998's own stated scope (adversarial *document* nesting, which this PR and its predecessor now fully close — verified across `jq`, `jq -e`, `jq -S`, `jq -C`, `yq --input-format json` in every combination I could construct) and touches at least 6 more call sites across 4 files plus a structural `Drop` question that needs its own design pass, not a rushed addition here. Full detail and repro cases are in #1005.

**Not changed:** the exit-code asymmetry between `to_owned`'s panic (exit 101) and `print_json`'s clean `anyhow` error (exit 1) for what's conceptually the same failure — this is a known, deliberate tradeoff (documented in both functions' doc comments and reflected in the tests, which assert the different codes explicitly) rather than an oversight: unifying it would mean either giving `print_json` a panic too (losing its clean-error UX) or giving `to_owned` a `Result` (the exact 55+-call-site ripple through the evaluator's hot path this PR's own design explicitly avoided).

Full local verification re-run after these changes: build, full test suite (2477+ tests), non-regex path, yq golden conformance, both required clippy configs, and the new + existing `-e`/`-S`/`-C`/adversarial-nesting CLI tests — all clean.

PR #1007, opened 2026-08-15
fix(cli): use scientific notation for large/small computed yq floats

### LINKED ISSUE #997: yq: bare-float formatter never uses scientific notation for very large/small magnitudes
## Summary

\`succinctly yq\`'s formatter for computed (non-literal-preserved) float values never
switches to scientific notation, unlike real \`yq\` (and unlike \`format_number_jq_compat\`,
which succinctly already has and uses for JSON-literal-preserving output in \`succinctly jq\`).

This is **pre-existing and unrelated to JSON input** — it reproduces on a plain YAML-input
computed value, with no JSON literal anywhere in the pipeline:

\`\`\`bash
$ echo 'a: 1' | succinctly yq '.a * 1e100' -o json
10000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000.0

$ echo 'a: 1' | yq '.a * 1e100' -o json     # real yq (mikefarah/yq v4.53.3)
1e+100
\`\`\`

Real \`yq\` switches to scientific notation for large-magnitude floats; succinctly's bare
\`OwnedValue::Float\` formatting path (\`format!("{f}")\` for JSON via \`to_json()\`,
\`format_float_with_fraction\` for YAML output) never does, producing a 100+ digit decimal
expansion instead.

## How this surfaced

Found during code review of #978 (yq JSON-literal-spelling leak fix, PR #994).  That fix
adds \`canonicalize_json_numbers\`, which collapses JSON-sourced \`NumberLiteral\` values
(which *do* use \`format_number_jq_compat\`'s scientific-notation-aware formatting) down to
bare \`Int\`/\`Float\` — matching real yq's own float64-canonicalization behavior (confirmed:
real yq also rounds \`99999999999999999999999999\` to \`1e+26\`, discarding the exact digits).
So #978's fix is correct to canonicalize, but the canonicalized value then inherits this
pre-existing formatter gap for large magnitudes. Before #978, JSON numbers took the
literal-preserving path and never hit this formatter, which is why it went unnoticed for
JSON input specifically — but it was always reachable via ordinary yq arithmetic on YAML
input, as the repro above shows.

## Suggested fix

Give \`OwnedValue::Float\`'s yq-mode formatting the same jq-canonical scientific-notation
threshold/algorithm that \`format_number_jq_compat\` already implements (or route through it
after synthesizing an exponential-form literal string), rather than maintaining two float
formatters that disagree on when to switch to scientific notation. Needs oracle
verification against real \`yq\` for the exact magnitude threshold (jq/yq's threshold for
switching may not be identical to Rust's or JS's default).

## Scope note

Not blocking #978/PR #994 — that PR's canonicalization direction (round to float64,
discard exact literal digits) is independently confirmed correct against the real yq
oracle; only the *display formatting* of the resulting large-magnitude float is wrong, and
was already wrong before #978 touched this code.

newhoggy:
## Summary

- `OwnedValue::Float`'s yq-mode formatting never switched to scientific
  notation, unlike real `yq` (v4.53.3) — a computed magnitude like
  `.a * 1e100` produced a 100+ digit decimal expansion instead of `1e+100`.
- Added `format_float_yq` in `output.rs`, gated on `ControlEscape::Yq` so
  it only affects yq mode — `succinctly jq`'s own separate, differently-
  thresholded formatter gap (out of scope here) is untouched.
- Threshold (`exponent >= 6` or `<= -5`) and spelling (`e+NN`/`e-NN`,
  lowercase, signed, exponent padded to at least 2 digits) are both
  oracle-verified against real `yq` v4.53.3, on both JSON (`format_json_impl`)
  and YAML (`emit_yaml_value`) output, compact and pretty alike.
- Deliberately does **not** touch `OwnedValue::NumberLiteral`'s formatting
  (YAML-sourced identity/navigation output) or the M2 streaming fast path —
  both correctly preserve the original source's decimal-vs-scientific style
  regardless of magnitude (confirmed against real yq: a huge *decimal*
  literal stays fully expanded on identity, only a *computed* result gets
  the magnitude-based switch), and per-existing tests already cover that.

## Test plan

- [x] New CLI tests in `tests/yq_cli_tests.rs` covering the issue's own
      repro, both threshold boundaries (positive and negative exponent),
      YAML output, and a jq-mode non-regression check
- [x] New unit tests in `output.rs` for `format_float_yq` directly and for
      the `format_json`/`ControlEscape::Yq` integration
- [x] All new/changed behavior spot-checked live against real `yq` v4.53.3

Fixes #997

### DIFF
--- a/src/bin/succinctly/output.rs
+++ b/src/bin/succinctly/output.rs
@@ -318,6 +318,45 @@ pub fn format_json(value: &OwnedValue, opts: &JsonFormatOpts) -> String {
     format_json_impl(value, opts, 0)
 }
 
+/// Render a computed (non-literal-preserved) `f64` the way real yq does:
+/// decimal for everyday magnitudes, scientific notation once the value's
+/// decimal exponent is `>= 6` or `<= -5`.
+///
+/// Only for [`OwnedValue::Float`] -- a value with no source literal left to
+/// preserve, either because it was actually computed (arithmetic) or because
+/// it came from JSON input, which real yq always re-serializes through
+/// float64 rather than preserving spelling. [`OwnedValue::NumberLiteral`]
+/// (YAML-sourced identity/navigation output) keeps its own source spelling
+/// regardless of magnitude and must never route through this function --
+/// confirmed against real yq v4.53.3: `12345678901234567890123` (a decimal
+/// literal) stays fully expanded on identity, while the equivalent
+/// *computed* magnitude switches to scientific notation. See issue #997.
+///
+/// The threshold and the `e+NN`/`e-NN` (lowercase, signed, exponent padded
+/// to at least 2 digits) spelling are both oracle-verified against real yq;
+/// this is yq's own threshold, distinct from `format_number_jq_compat`'s
+/// jq-mode one (which real jq only reformats when the source literal itself
+/// already used exponent notation).
+#[must_use]
+pub fn format_float_yq(f: f64) -> String {
+    if f == 0.0 {
+        return format_float_with_fraction(f);
+    }
+    let sci = format!("{f:e}");
+    let (mantissa, exp_str) = sci
+        .split_once('e')
+        .expect("Rust's exponential formatter always includes a lowercase 'e'");
+    let exp: i32 = exp_str
+        .parse()
+        .expect("exponent from Rust's exponential formatter is always a valid i32");
+    if (-4..6).contains(&exp) {
+        format_float_with_fraction(f)
+    } else {
+        let sign = if exp < 0 { '-' } else { '+' };
+        format!("{mantissa}e{sign}{:02}", exp.abs())
+    }
+}
+
 /// Recursive JSON formatter behind [`format_json`].
 fn format_json_impl(value: &OwnedValue, opts: &JsonFormatOpts, level: usize) -> String {
     let indent = opts.indent;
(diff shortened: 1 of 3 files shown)


newhoggy:
Mandatory review complete (8-agent multi-angle review, standard effort). Findings addressed in 4c8071cc9:

- **Confirmed real bug, fixed**: \`first(f)\`/\`last(f)\` were unconditionally M2-streamable regardless of \`f\`, so \`first(.a * 1e100)\` still hit \`src/jq/stream.rs\`'s hardcoded \`format_float_with_fraction\`, bypassing this PR's fix entirely. \`can_use_m2_streaming\` now recurses into the inner expression for both AST spellings, matching how \`Pipe\`/\`Optional\`/\`Paren\` already do — closing the gap without touching the M2 fast path's writer itself (pure-navigation \`first(.[])\` etc. still take the fast path, dup-key preservation intact).
- **Latent NaN/Infinity panic** (found independently by 2 agents): \`format_float_yq\` is unreachable with a non-finite value today (both call sites guard first), but nothing enforced that. Added a \`debug_assert!\` and documented the precondition explicitly, matching its sibling \`format_float_with_fraction\`.
- Dropped a dead \`f == 0.0\` special case (verified: the general path already produces the same output for both signs of zero) and fixed a STYLE-0006 doc-comment mood violation.
- One finding was investigated and found to be **not a regression**: compact yq JSON dropping \`.0\` on a computed whole float pre-PR was itself wrong — verified against real yq v4.53.3 (\`.a + 0.0\` on \`a: 1\` → \`1.0\` even with \`-I=0\`), so this PR's uniform (compact-and-pretty-alike) handling is the fix, not a side effect.
- **Found and filed as a separate, pre-existing issue**: #1008. Two review agents independently found that YAML scientific-notation literals (\`1e2\`, \`1e100\`, ...) get corrupted into decimal expansions by ordinary \`.field\` navigation (both the M2 fast path *and* the DOM path) — reproduced identically on unmodified \`main\` @ 54d1a23f7, so it predates and is unrelated to this PR (which only touches genuinely *computed*, non-literal-preserved float formatting).

Issue #1008, opened 2026-08-15
yq: scientific-notation YAML literals corrupted by any non-identity navigation


newhoggy:
## Summary

A YAML scalar written in scientific notation (`1e2`, `1.5e100`, etc.) prints
correctly on bare identity (`.`), but is silently expanded into a huge
decimal string by *any* other navigation shape — `.field`, `.field[]`,
`first(.field)`, and even the DOM path forced via `--arg`. This is
pre-existing on `main` (confirmed at `54d1a23f7`, unrelated to #997/PR #1007
— that PR only touches genuinely *computed* (arithmetic) float formatting
and does not go anywhere near this).

```bash
$ printf 'a: 1e2\n' | succinctly yq '.'
a: 1e2
$ printf 'a: 1e2\n' | succinctly yq '.a'
100.0
$ printf 'a: 1e2\n' | yq '.a'      # real yq v4.53.3
1e2
```

More severe with a larger exponent — the value is not just re-typed, it's
fully expanded to a 100+ digit decimal:

```bash
$ printf 'a: 1e100\n' | succinctly yq '.a' -o json -I=0
10000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
$ printf 'a: 1e100\n' | yq '.a' -o json -I=0     # real yq
1e+100
```

## Reproductions (all against `main` @ 54d1a23f7)

| Filter                  | Input          | succinctly gives         | real yq gives |
|--------------------------|----------------|---------------------------|----------------|
| `.`                       | `a: 1e2`       | `a: 1e2` (correct)         | `a: 1e2`       |
| `.a`                      | `a: 1e2`       | `100.0`                    | `1e2`          |
| `.a[]`                    | `a: [1e2, 2e3]`| `1e2` / `2e3` (correct, oddly) | `1e2` / `2e3` |
| `first(.a)`               | `a: 1e2`       | `1e2` (correct, oddly)     | `1e2`          |
| `.a` with `--arg x y` (forces DOM) | `a: 1e100` | 105-digit decimal | `1e+100` |
| `.a` (M2 fast path, no `--arg`) | `a: 1e100` | 105-digit decimal, no `.0` | `1e+100` |

Note the `.a[]`/`first(.a)` rows currently print correctly — the bug isn't a
clean "any navigation" rule, it's inconsistent across shapes, which suggests
several different code paths resolve a navigated-to scalar differently (some
preserve `NumberLiteral`'s source text correctly, some don't).

[...]

PR #1012, opened 2026-08-15
fix(yaml): preserve a YAML float literal's trailing zero on -o json

### LINKED ISSUE #993: yq: -o json output loses a YAML-sourced float's trailing zero, unlike default YAML output
**Severity: Low**

## Summary

Found while verifying #978 (yq's JSON-input literal-spelling fix). Unrelated to that issue: a genuinely YAML-sourced float scalar with a trailing zero (`1.50`) correctly preserves its spelling through default (YAML) output, but loses the trailing zero specifically when output format is `-o json`.

## Repro

```
$ echo 'a: 1.50' | succinctly yq '.a'
1.50
$ echo 'a: 1.50' | yq '.a'
1.50

$ echo 'a: 1.50' | succinctly yq -o json -I0 '.a'
1.5     # wrong
$ echo 'a: 1.50' | yq -o json -I0 '.a'
1.50    # correct
```

Also reproduces for the whole-document case (`.` instead of `.a`) and isn't specific to field access.

## Root cause (not fully investigated)

Confirmed this is on the M2 (`can_json_fast_path`) streaming fast path — `src/yaml/light.rs`'s `stream_resolved_scalar_as_json`/`format_float_with_fraction` most likely doesn't fully replicate #918's literal-text-echo behavior for the JSON-output case, only for YAML output. Not confirmed whether the DOM path (non-M2-streamable filters) has the same gap.

## Why this is separate from #978

#978 is about `--input-format json` (JSON-*sourced* values) incorrectly over-preserving spelling, fixed by disabling the M2 fast path for that case and canonicalizing the DOM path. This issue is the opposite: a genuinely YAML-sourced value under-preserving spelling, specific to `-o json` output. Confirmed unrelated and unaffected by #978's fix — reproduces identically before and after that change, since `#978` only touches the `--input-format json` gate, and this repro uses no `--input-format` flag at all (default `Auto`, genuine YAML source).

## Suggested fix direction

Compare `stream_resolved_scalar_as_json`'s float-formatting path against the YAML-output equivalent (`src/yaml/light.rs`) and against #918's `is_preservable_float_literal`/literal-echo mechanism to find where the JSON-output path diverges.

newhoggy:
## Summary

- `stream_resolved_scalar_as_json` — the M2 fast path's JSON scalar writer
  (used by ordinary `.field`/`.`/navigation queries when `-o json` is
  requested) — always reconstructed a float's text from its parsed `f64` via
  `format_float_with_fraction`, silently dropping any extra trailing zero a
  source literal carried (`1.50` → `1.5`), even though default (YAML) output
  preserved the same literal correctly through a different writer.
- Fixed by echoing the source text when `is_preservable_float_literal`
  confirms it's safe — the exact predicate the DOM path's existing
  `number_literal()` mechanism already uses for the same decision (#918),
  so this doesn't introduce a new preservation policy, just applies the
  existing one to a writer that had been missing it.
- Deliberately does not touch the cases `is_preservable_float_literal`
  excludes on purpose (exponent notation — that's #1008's separate scope;
  >17 significant digits) — those keep falling back to
  `format_float_with_fraction` exactly as before.

## Test plan

- [x] New CLI tests in `tests/yq_cli_tests.rs`: the issue's own repro
      (`.a` and whole-doc `.`), a non-regression check that excluded
      spellings (exponent, >17 digits) keep their existing fallback
      behavior, and a check that default YAML output (already correct)
      didn't regress
- [x] Spot-checked live against real `yq` v4.53.3

Fixes #993

### DIFF
--- a/src/yaml/light.rs
+++ b/src/yaml/light.rs
@@ -3682,8 +3682,9 @@ fn stream_yaml_scalar_as_json<Out: core::fmt::Write>(
 }
 
 /// Stream an already-resolved scalar as JSON. `str_val` is the original
-/// source text, used only for the `Str` case (and only if `resolved` didn't
-/// come from resolving it, e.g. tag-forced `!!str` on non-string content).
+/// source text, used for the `Str` case (and only if `resolved` didn't come
+/// from resolving it, e.g. tag-forced `!!str` on non-string content), and
+/// for a preservable `Float` literal (#993).
 fn stream_resolved_scalar_as_json<Out: core::fmt::Write>(
     out: &mut Out,
     resolved: ResolvedScalar,
@@ -3694,7 +3695,16 @@ fn stream_resolved_scalar_as_json<Out: core::fmt::Write>(
         ResolvedScalar::Bool(true) => out.write_str("true"),
         ResolvedScalar::Bool(false) => out.write_str("false"),
         ResolvedScalar::Int(n) => write!(out, "{n}"),
-        // Not `write!(out, "{f}")`: that drops the `.0` from a whole float.
+        // Echo the source text when it's already safe, valid JSON number
+        // syntax (`number_literal()` below uses the same predicate for the
+        // DOM path) -- this is what keeps a trailing zero (`1.50`) intact;
+        // `format_float_with_fraction` only reconstructs the value's
+        // shortest round-trip spelling, which silently drops it (#993).
+        // Not `write!(out, "{f}")` either way: that drops the `.0` from a
+        // whole float.
+        ResolvedScalar::Float(f) if f.is_finite() && is_preservable_float_literal(str_val) => {
+            out.write_str(str_val)
+        }
         ResolvedScalar::Float(f) if f.is_finite() => out.write_str(&format_float_with_fraction(f)),
         // JSON cannot represent the `.inf`/`.nan` family.
         ResolvedScalar::Float(_) => out.write_str("null"),
--- a/src/yaml/light.rs
+++ b/src/yaml/light.rs
@@ -3237,8 +3237,10 @@ fn write_yaml_scalar_as_json(output: &mut String, str_val: &str) {
 }
 
 /// Write an already-resolved scalar as JSON. `str_val` is the original
-/// source text, used only for the `Str` case (and only if `resolved` didn't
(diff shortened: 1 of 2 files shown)


newhoggy:
Mandatory review complete (standard effort). Multiple independent finder agents converged on the same real bug, addressed in 7f445db1b:

- **Confirmed real bug, fixed**: this PR's first commit only patched `stream_resolved_scalar_as_json` (the M2/streaming JSON emitter). Its documented sibling `write_resolved_scalar_as_json` — the buffered/DOM emitter backing the public `YamlCursor::to_json()`/`to_json_document()` API and the `yaml_test_suite.rs` conformance harness (not reachable from the shipped `succinctly yq` CLI, which only uses the streaming path) — still had the original unconditional `format_float_with_fraction` call, silently reproducing #993 for any direct library caller. Applied the identical `is_preservable_float_literal` echo, and added a lockstep test (`test_float_trailing_zero_lockstep_between_emitters_993`) pinning that both emitters agree, matching this file's own established convention for buffered/streaming pairs.
- **Simplification**: dropped a redundant `f.is_finite()` conjunct from both echo arms — `is_preservable_float_literal` already bounds the digit count (≤17) well below where a finite `f64` could overflow, matching `number_literal()`'s identical (already `is_finite()`-free) guard at the same call site's precedent.
- **Test correctness**: one of my own regression tests used `007e2` as an "exponent notation is excluded" case, but that literal has no `.` at all and is rejected by `is_preservable_float_literal`'s *first* condition, never reaching the exponent check its comment claimed to exercise. Replaced with `1.5e2`, which actually has both a dot and an exponent.
- A few findings were investigated and found to be pre-existing, out-of-scope, or non-issues: `is_preservable_float_literal`'s digit budget counting all digit characters (not just significant ones) is a pre-existing, already-tested, documented "deliberately conservative" limitation, not something this PR introduces or narrows. A performance-efficiency finding (new validation calls on the P9/M2 streaming hot path) was noted; given the issue's own "Low" severity and the tiny (typically <15 byte) strings involved, a full CLAUDE.md-compliant cross-architecture A/B benchmark wasn't run, but local full-suite tests plus targeted checks (`cargo build --release --features bench-runner` builds clean) showed no obvious red flags.

Issue #626, opened 2026-08-04
perf(jq): eval_index_expr/eval_slice_bound eagerly materialize a computed key/bound's full subtree before checking whether it's usable


newhoggy:
**Severity: Medium** — follow-up to #491, which named two separate contributors to the `..`/`recurse` + computed-key quadratic blowup. #491's fix addressed the second (the per-node `to_json_for_reindex` + `JsonIndex::build` round trip in `eval_owned_expr`/`eval_owned_input`, now bypassed for bare `.`/`.foo`/`.[n]` via a fast path). This issue covers the residual quadratic cost that fast path left behind.

### Correction (2026-08-08)

This issue originally named `push_recursive_branches`/`resolve_recurse` (the `path(..)`/`path(recurse(...))` resolver, both in `src/jq/eval.rs`) as the cause, based on their `value.clone()`/`current.clone()` calls being visibly O(subtree). That diagnosis was wrong for the benchmark below: those two functions are only reachable through `resolve_dynamic_indexes`, whose only 4 callers are `eval_assign` (`=`), `eval_update` (`|=`), `builtin_path` (`path()`), `builtin_del` (`del()`) — none of which a bare `.. | .[.k]?` filter (no `path()`, no assignment) ever reaches. That query instead goes through `eval_recursive_descent`/`collect_recursive`, which clones a `StandardJson`/`JsonCursor` — confirmed `Copy`, O(1) per node, not the source of the quadratic cost.

`push_recursive_branches`/`resolve_recurse` do have a real, independently-verified O(subtree)-per-node clone bug — it's just a different bug from the one measured here, reachable only via `path()`/write-context. Split out to #668 to avoid conflating the two.

[...]

PR #1015, opened 2026-08-16
fix(jq): guard reduce/foreach-constructed values against stack overflow

### LINKED ISSUE #1005: jq/yq: reduce/foreach-constructed deep values crash via unguarded recursion (value.rs, compare_values, eval.rs's own to_owned, reconcile_presentation, format_json_impl) — distinct from #998
## Summary

Found during #998/PR #1002's mandatory review. #998 fixed the stack-overflow-on-adversarially-deep-*input* class (`eval_generic::to_owned`/`to_owned_cursor`/`to_owned_with_comments`, `jq_runner::print_json`, `lazy::cursor_to_owned` — all guarded at depth 256 now). This issue is a **structurally different** vulnerability: a value **constructed at query-evaluation time** (via `reduce`/`foreach`/`while`/`until`/`repeat`, growing one level per loop iteration) can exceed the same depth with **no adversarial document at all** — plain `null` input is enough — and crashes through several entirely different, still-unguarded recursive functions that #998's fix doesn't touch.

## Repro (confirmed live, `null` input, no file needed)

```
$ echo null | succinctly yq 'reduce range(50000) as $i (null; [.]) | empty'
thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting   (SIGABRT, exit 134)
```

Binary search: `range(5000)` succeeds, `range(8000)` already takes >10s (consistent with an O(n²) re-serialize-every-iteration cost) en route to eventually overflowing. The crash happens **during evaluation**, before any output/comparison/materialization step — `| empty` discards the result entirely and it still crashes.

## Root cause: at least 4 more unguarded recursive tree-walkers, plus a structural `Drop` issue

1. **`OwnedValue::to_json_for_reindex`** (`src/jq/value.rs:692`) — recurses over `Array`/`Object` with no depth guard. Called on **every loop iteration** by `eval_owned_input`/`eval_owned_expr_ctrl` (`src/jq/eval.rs:13502`/`:13405`), which back `reduce`/`foreach`/`while`/`until`/`repeat`/`walk`'s accumulator-update step. This is the confirmed proximate cause of the repro above.
2. **`compare_values`** (`src/jq/eval.rs:2087`) and `OwnedValue`'s derived `PartialEq` (`src/jq/value.rs:910`) — recurse over `Array`/`Object` pairs with no depth guard, reachable from `==`/`!=`/`<`/`<=`/`>`/`>=`. Not hit by the specific repro above (shadowed by finding 1 firing first during construction), but independently unguarded — a tree built past depth 256 by any other route and then compared would hit this.
3. **`OwnedValue::to_json`** (`src/jq/value.rs:643`) — recurses with no depth guard. Used by `tostring`, `@json`, `debug`/`stderr`, and roughly a dozen other formatting call sites in `eval.rs` (lines 4861, 4913, 5364, 5499, 5700, 19552, 20961, 21019, 23763, etc.).
4. **`eval.rs`'s own private `to_owned<W>`** (`src/jq/eval.rs:328`) — a *second*, entirely separate copy of the cursor-to-`OwnedValue` conversion from the one #998 patched in `eval_generic.rs`, with no depth guard. Feeds `collect_tostream_events`/`builtin_tostream` (line 16108) and `flatten_owned` (line 4268). Document-sourced *input* is protected transitively (the top-level value already passed through `eval_generic`'s now-guarded conversion before reaching `eval.rs`), but a value `eval.rs` constructs internally afterward (e.g. a `reduce`/`foreach` accumulator) bypasses that upfront guard and hits this unguarded copy directly.
5. **`reconcile_presentation`** (`src/bin/succinctly/yq_runner.rs:811`) — recurses on `(Object, Object)`/`(Array, Array)` pairs with no depth guard; used by `yq -i`'s comment-preserving merge. The pristine side is bounded by #998's now-guarded `to_owned_with_comments`, but the filter's *evaluated output* can still be constructed past the limit via `reduce`/`foreach` (same mechanism as finding 1) and reaches this unguarded.
6. **`format_json_impl`** (`src/bin/succinctly/output.rs:322`) — a third, independent recursive JSON formatter (distinct from `print_json`, guarded by #998, and `stream_json`/`stream_yaml`) with no depth guard. Reached via `write_output`/`format_json` for the "eager" output path (`--sort-keys`, `--slurp`, color/ASCII output, DSV streaming rows). Input depth *is* bounded on this path (`serde_json::from_str`/`Deserializer` enforce their own ~128-deep limit and return a clean `Err`), but that only bounds *input*, not the filter's *evaluated output* — a `reduce`/`foreach`-constructed value bypasses `serde_json`'s guard entirely before reaching this formatter.
7. **Structural**: `OwnedValue`'s enum (`Array(Vec<OwnedValue>)`, `Object(IndexMap<String, OwnedValue>)`) has no custom `Drop` impl, so the compiler-generated recursive drop glue walks the whole tree with one stack frame per nesting level whenever a sufficiently deep value goes out of scope — independent of any comparison, printing, or materialization ever happening. Even after every guard above lands, a value that *does* reach depth 256 legitimately (e.g. via `walk` operating within the limit) still needs an iterative `Drop` if it can ever be dropped while deeper than a stack-safe bound; likely moot in practice once findings 1-6 cap construction at 256, but worth naming as a reason a "check depth at every recursion site" strategy can't be the *whole* answer — a value already at the ceiling still has one more unguarded recursive walk ahead of it the instant it's dropped.

## Why this is a separate issue, not a #998 scope-creep fix

#998's own title and repro are specifically about adversarial **document input** nesting; the guards #1002 added are exactly the ones on that path (confirmed via extensive live testing across `jq -e`, `jq -C`, `jq -S`, `yq --input-format json` in various combinations — all now correctly guarded or already safe via `serde_json`'s own limit on the color/sort-keys path). This issue's mechanism is entirely different (query-time construction via unbounded loop iteration, no adversarial document needed) and touches a much larger set of call sites (at least 6 more functions across 4 files, plus a structural `Drop` concern) — properly scoping and fixing it needs its own PR with its own dedicated verification pass, not a rushed addition to #998's already-large diff.

## Suggested fix direction

Given the sheer number of independent "walk an `OwnedValue` tree" call sites found here (and the pre-existing one #998 already fixed), a single shared, depth-checked traversal primitive (or a depth field carried in `OwnedValue` itself, cheap to maintain incrementally at construction time) that all of `to_json`/`to_json_for_reindex`/`compare_values`/`PartialEq`/`reconcile_presentation`/`format_json_impl`/`eval.rs`'s own `to_owned` route through would close this systemically, rather than the "hand-add a depth parameter to each function as it's individually discovered" pattern #998 already had to use twice (once for the original 3 functions, once more for `cursor_to_owned` after review). Worth scoping as a design discussion given the number of call sites and the `Drop` question above, rather than assuming a mechanical per-function patch is sufficient.

newhoggy:
## Summary

`reduce`/`foreach`/`while`/`until`/`repeat` can build a value one level deeper per loop iteration with no adversarial *document* involved — plain `null` input is enough:

```
$ echo null | succinctly yq 'reduce range(50000) as $i (null; [.]) | empty'
thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting   (SIGABRT, exit 134)
```

#998 guarded the document-input conversion path (`eval_generic::to_owned`/`to_owned_cursor`/`to_owned_with_comments`, `print_json`, `lazy::cursor_to_owned`) against adversarially deep *input*. This issue is a structurally different vulnerability: a value constructed at query-evaluation time reaches six entirely different, still-unguarded recursive tree-walkers that #998's fix never touched:

1. `OwnedValue::to_json` (`src/jq/value.rs`) — the confirmed proximate cause; called on every `reduce`/`foreach` accumulator-update iteration via `to_json_for_reindex`.
2. `OwnedValue::to_json_for_reindex` (`src/jq/value.rs`)
3. `OwnedValue`'s `PartialEq`/`==` (`src/jq/value.rs`)
4. `compare_values` (`src/jq/eval.rs`) — reachable from `sort`/`<`/`>`/etc.
5. `eval.rs`'s own private `to_owned<W>` (`src/jq/eval.rs`) — a second, independent copy of the cursor-to-`OwnedValue` conversion, distinct from the one #998 patched in `eval_generic.rs`.
6. `reconcile_presentation` (`src/bin/succinctly/yq_runner.rs`) — `yq -i`'s comment-preserving merge.
7. `format_json_impl` (`src/bin/succinctly/output.rs`) — the eager JSON output formatter.

## Fix

Each function now threads an explicit `depth` parameter through a private `_at_depth` helper (mirroring #998's own pattern) and panics past a depth ceiling rather than recursing unbounded — a controlled panic (catchable, clean exit code) instead of a raw stack overflow (SIGABRT/undefined behavior).

**New, separately-tuned constant** (`value::MAX_VALUE_TREE_DEPTH = 384`), not a reuse of `eval_generic::MAX_NESTING_DEPTH` (256): reusing 256 broke `tests/jq_recurse_depth_tests.rs`'s pre-existing, deliberately-pinned depth-300 document-navigation capability (`..`/`recurse` correctness at depth 300, a #626 de-risk step) — that test routes through eval.rs's own `to_owned` during ordinary traversal, not just accumulator construction, so it needs a higher ceiling. I empirically measured each guarded function's real stack-overflow boundary (debug build, default 2MiB test-thread stack — the more fragile combination, matching how #998's own 256 was measured):

| Function | Measured crash boundary |
|---|---|
| `reconcile_presentation` | 580 (safe) / 600 (crash) — tightest |
| `format_json_impl` | 600 (safe) / 650 (crash) |
| `eval.rs`'s `to_owned` | 1800 (safe) / 2000 (crash) |

384 clears the 300-deep pinned test with the same proportional margin #998's own 256 used to clear its 200-deep `walk` floor (`256 = 200 * 1.28`; `384 = 300 * 1.28`), while sitting comfortably under the tightest measured boundary (580) with margin to spare for a CI runner with a smaller default stack.

## Testing

- New unit test per guarded function (6 total, one per file/function) confirming success just under `MAX_VALUE_TREE_DEPTH` and a panic exactly at it.
- Full `cargo test --features cli,simd,regex,serde` suite green, including `tests/jq_recurse_depth_tests.rs` (previously would have regressed with a naive reuse of the 256 constant — caught locally before this PR, not left in the diff).
- `cargo clippy --all-targets --features cli,simd,regex,serde -- -D warnings` and the `scalar-yaml`-inclusive second CI invocation both clean.
- Manually confirmed the original repro now exits 101 with a clean panic message instead of SIGABRT/134.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, including new #1005 tests and `jq_recurse_depth_tests.rs`)
- [x] `cargo clippy --all-targets --features cli,simd,regex,serde -- -D warnings`
- [x] `cargo fmt --check`
- [x] Manual repro verification (both the original crash and the depth-300 regression test)

Fixes #1005

### DIFF
--- a/src/bin/succinctly/output.rs
+++ b/src/bin/succinctly/output.rs
@@ -9,7 +9,7 @@ use succinctly::jq::escape::{
     escape_json_body as run_escaper, write_json_body_jq, write_json_body_jq_ascii,
     write_json_body_yq, write_json_body_yq_ascii,
 };
-use succinctly::jq::{EvalError, OwnedValue, StreamError};
+use succinctly::jq::{assert_value_tree_depth, EvalError, OwnedValue, StreamError};
 use succinctly::yaml::format_float_with_fraction;
 
 /// Exit codes matching jq behavior
@@ -365,7 +365,14 @@ pub fn format_float_yq(f: f64) -> String {
 }
 
 /// Recursive JSON formatter behind [`format_json`].
+///
+/// Panics past `succinctly::jq::MAX_VALUE_TREE_DEPTH` levels of nesting
+/// (#1005) — see that constant's own doc comment for why. `value` is the
+/// filter's evaluated output, constructible via `reduce`/`foreach`/etc.
+/// with no adversarial document involved; `serde_json`'s own input-depth
+/// limit (used elsewhere on this eager-output path) never sees it.
 fn format_json_impl(value: &OwnedValue, opts: &JsonFormatOpts, level: usize) -> String {
+    assert_value_tree_depth(level);
     let indent = opts.indent;
     let compact = indent.is_empty();
     let current_indent = if compact {
@@ -1210,5 +1217,43 @@ mod tests {
         // Diagnostic output; assert it runs without panicking for both tools.
         print_build_configuration("jq");
         print_build_configuration("yq");
+    }
+
+    /// `depth` levels of single-element array nesting: `[[[...[Null]...]]]`.
+    fn linear_array_nest(depth: usize) -> OwnedValue {
+        let mut v = OwnedValue::Null;
+        for _ in 0..depth {
+            v = OwnedValue::Array(vec![v]);
+        }
+        v
+    }
+
+    /// #1005: `value` is a filter's evaluated output (e.g. a `reduce`
+    /// accumulator growing one level per iteration), which has no
+    /// adversarial *document* behind it — `format_json_impl` must
+    /// independently refuse to recurse past the same limit rather than
+    /// overflow the stack.
+    #[test]
+    fn format_json_panics_past_nesting_depth_limit_1005() {
+        use succinctly::jq::MAX_VALUE_TREE_DEPTH;
+
+        let opts = JsonFormatOpts {
+            indent: "",
+            sort_keys: false,
+            ascii: false,
+            float_style: FloatStyle::Shortest,
+            control_escape: ControlEscape::Jq,
+        };
+
+        let under = linear_array_nest(MAX_VALUE_TREE_DEPTH - 1);
+        let _ = format_json(&under, &opts);
+
+        let over = linear_array_nest(MAX_VALUE_TREE_DEPTH);
+        let result =
+            std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| format_json(&over, &opts)));
+        assert!(
+            result.is_err(),
+            "format_json should panic at MAX_VALUE_TREE_DEPTH"
+        );
     }
 }
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -15,7 +15,8 @@ use succinctly::jq::eval_generic::{
     GenericResult,
 };
 use succinctly::jq::{
-    self, sync_aliased_paths, Builtin, EvalError, Expr, OwnedValue, QueryResult, YqSemantics,
+    self, assert_value_tree_depth, sync_aliased_paths, Builtin, EvalError, Expr, OwnedValue,
+    QueryResult, YqSemantics,
 };
 use succinctly::json::JsonIndex;
 use succinctly::yaml::{
@@ -812,6 +813,20 @@ fn reconcile_presentation(
     pristine_tree: &CommentTree,
     result_value: &OwnedValue,
 ) -> CommentTree {
+    reconcile_presentation_at_depth(pristine_value, pristine_tree, result_value, 0)
+}
+
+/// Panics past `succinctly::jq::MAX_VALUE_TREE_DEPTH` levels of nesting
+/// (#1005) — see that constant's own doc comment for why. `result_value` is
+/// the evaluated filter output, which can be constructed via `reduce`/
+/// `foreach`/etc. with no adversarial document involved on either side.
+fn reconcile_presentation_at_depth(
+    pristine_value: &OwnedValue,
+    pristine_tree: &CommentTree,
+    result_value: &OwnedValue,
+    depth: usize,
+) -> CommentTree {
+    assert_value_tree_depth(depth);
     match (pristine_value, result_value) {
         (OwnedValue::Object(p_fields), OwnedValue::Object(r_fields)) => {
             let own_comment = pristine_tree.own().map(str::to_string);
@@ -820,7 +835,9 @@ fn reconcile_presentation(
             let mut key_comments = IndexMap::new();
             for (k, r_v) in r_fields {
                 let child = match p_fields.get(k) {
-                    Some(p_v) => reconcile_presentation(p_v, pristine_tree.field(k), r_v),
+                    Some(p_v) => {
+                        reconcile_presentation_at_depth(p_v, pristine_tree.field(k), r_v, depth + 1)
+                    }
                     None => CommentTree::empty(),
                 };
                 fields.insert(k.clone(), child);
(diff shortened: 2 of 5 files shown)


newhoggy:
## Code review summary

Ran a multi-angle review (8 finder agents: diff scan, removed-behavior audit, cross-file trace, language-pitfall check, duplicate-implementation check, reuse/existing-helper check, simplification/efficiency, CLAUDE.md conventions).

**No blocking issues** — all 6 named guards in the PR body are correct, verified via 60+ tool calls across the review including live rebuild-and-repro testing on both this branch and `main`. One inaccuracy in my own doc comment was found and fixed (commit `fb4c5a17e`): the claim that `eval.rs`'s `to_owned` needs `MAX_VALUE_TREE_DEPTH=384` for "CLI-reachable" document navigation overstated the case — the CLI binary actually hits a different, earlier `eval_generic::MAX_NESTING_DEPTH` guard (256) first via a separate code path. The underlying need for 384 (clearing `tests/jq_recurse_depth_tests.rs`'s pinned depth-300 capability via the library's own `eval()` API) still holds; only the CLI-reachability framing was wrong.

Three legitimate but out-of-scope findings were filed as follow-ups rather than scope-creeping this PR:
- **#1016 (High)**: user-defined recursive jq functions (`def deep(n): ...`) overflow the stack via a completely different mechanism (the evaluator's own call-stack recursion, not an `OwnedValue` tree-walk) — confirmed live at just depth 500. Needs its own investigation given it sits on `eval()`'s hot path.
- **#1017 (Medium)**: several more `OwnedValue`/`CommentTree` walkers (`standard_json_to_owned`, `emit_yaml_value`, `format_props_recursive`/`owned_to_yaml`, `merge_values` family, `flatten_owned`, a few lower-risk ones) are still unguarded, currently safe only via an implicit "always fed through a guarded reindex bridge first" invariant that isn't tested or documented.
- **#1018 (Low)**: `assert_value_tree_depth` duplicates `assert_nesting_depth`'s exact shape instead of a shared parameterized helper — a maintainability cleanup, no functional risk.

One review claim (that `emit_yaml_value` independently stack-overflows via `syq -n 'reduce range(1000) as $i (null; [.])'`) was empirically refuted during synthesis — that query already panics cleanly via this PR's new guard before ever reaching `emit_yaml_value`.

Issue #978, opened 2026-08-15
yq: JSON-sourced literal spelling leaks decimal-point preservation, unlike real yq


newhoggy:
**Severity: Low**

## Summary

Discovered while investigating #949 (computed whole-number float decimal-point handling). Real yq v4.53.3 never preserves a JSON-sourced literal's exact source spelling — `1.0` from JSON input always renders as `1`, in every output format, treating JSON-sourced values the same whether they're literal or computed. succinctly's `NumberLiteral` machinery doesn't distinguish input format and currently preserves `1.0` verbatim for JSON-sourced input too, in every output format.

## Repro (verified against pinned yq v4.53.3)

```
$ echo '1.0' | yq --input-format json -o json '.'
1
$ echo '1.0' | succinctly yq --input-format json -o json '.'
1.0     # wrong, should be 1

$ echo '1.0' | yq --input-format json -o json -I0 '.'
1
$ echo '1.0' | succinctly yq --input-format json -o json -I0 '.'
1.0     # wrong

$ echo '1.0' | yq --input-format json '.'   # YAML output
1
$ echo '1.0' | succinctly yq --input-format json '.'
1.0     # wrong
```

For comparison, YAML-sourced literals correctly preserve their spelling in every output format (unaffected, not a regression):

```
$ echo 'a: 1.0' | succinctly yq -o json -I0 '.a'
1.0     # correct, matches real yq
```

## Hypothesis

Real yq's `--input-format json` path apparently never carries literal source text through at all — it's most plausibly parsed via a generic value representation (Go's `encoding/json` → `float64`) rather than yq's native YAML-node-based parser that preserves scalar style/text. This would explain why JSON-sourced values behave identically whether touched or untouched: there's no "untouched literal" concept for them at all in real yq, only computed float64 values.

succinctly's JSON input path (`parse_input` in `src/bin/succinctly/yq_runner.rs`, `InputFormat::Json` arm) builds a `JsonIndex` and converts via `generic_to_owned`, which (like YAML's `yaml_to_owned_value`) constructs `OwnedValue::NumberLiteral` to preserve source spelling — this mirrors `succinctly jq`'s own (correct, for jq) convention of preserving JSON literal text, but yq's JSON-input path needs to NOT do this, unlike jq's.

## Suggested fix direction

[...]

Issue #442, opened 2026-07-28
yq/jq: pretty-printed JSON output silently drops duplicate mapping keys (compact output is correct)


newhoggy:
**Severity: Medium (silent data loss)** — reproduced.

Pretty-printed JSON output silently drops duplicate mapping keys (last one wins); compact output (`-I=0`) is correct and matches `yq`. Real `yq` v4.53.3 preserves both keys in **both** modes, so this is a regression relative to the reference, not an acceptable divergence.

```console
$ printf 'a: 1\na: 2\n' | succinctly yq -o json '.'
{
  "a": 2
}
$ printf 'a: 1\na: 2\n' | succinctly yq -o json -I0 '.'
{"a":1,"a":2}
$ printf 'a: 1\na: 2\n' | yq -o json '.'
{
  "a": 1,
  "a": 2
}
```

Not specific to YAML's explicit-key syntax — plain `a: 1\na: 2` triggers it, and it's how the `explicit_key_non_scalar_pretty` golden fixture (added while scoping #172) tripped over it: two explicit non-scalar keys both render to `""`, and the second silently overwrote the first in pretty mode only.

## Cause

`-I`/`--indent` (`src/bin/succinctly/main.rs:1021`) sets `output_config.compact = (indent == 0)`, which gates a fast/slow fork in `src/bin/succinctly/yq_runner.rs:1229-1247` (`can_yaml_fast_path` requires `output_config.compact`).

- **Compact path** (`-I0`): `root.stream_json_document(...)` → `stream_yaml_value_as_json`/`write_yaml_value_as_json` (`src/yaml/light.rs:1122`, `2435`, `1845-1877`) iterate the mapping's `fields` and write `"key":value` straight to the buffer — no map involved, so duplicate keys pass through untouched.
- **Pretty path** (default indent): falls through to `evaluate_yaml_direct_filtered` → `evaluate_yaml_cursor` (`yq_runner.rs:271-386`), which for identity calls `to_owned()` (`src/jq/eval_generic.rs:34-45`):
  ```rust
  pub fn to_owned<V: DocumentValue>(value: &V) -> OwnedValue {
      if let Some(fields) = value.as_object() {
          let mut map = IndexMap::new();
          let mut f = fields;
          while let Some((field, rest)) = f.uncons() {
              if let Some(key) = field.key_str() {
                  map.insert(key.into_owned(), to_owned(&field.value)); // last-write-wins
              }
              f = rest;
          }
[...]

PR #1019, opened 2026-08-16
fix(cli): restore M2 streaming for JSON input, fixing duplicate-key collapse

### LINKED ISSUE #996: yq: --input-format json now collapses duplicate object keys for M2-eligible filters (regression from #978)
**Severity: Medium**

## Summary

#978 (PR #994) fixed `yq --input-format json` incorrectly preserving a JSON-sourced number's literal spelling, by disabling the M2 cursor-streaming fast path for JSON input and falling back to the DOM path (`parse_input`'s `InputFormat::Json` arm). This fix is correct for its stated scope, but M2 streaming was *also* incidentally providing a second, unrelated invariant: duplicate JSON object keys survive unchanged, since M2 never materializes a map at all. The DOM path's `OwnedValue::Object` is `IndexMap`-backed and can only keep the last value for a repeated key — the same class of bug #442 already fixed for YAML input (~15 dedicated `test_duplicate_mapping_key_survives_*` regression tests exist for YAML; none exist for JSON).

## Repro (confirmed against pinned yq v4.53.3)

```
$ echo '{"a":1,"a":2}' | succinctly yq --input-format json -o json -I0 '.'
{"a":2}                                          # wrong — collapsed
$ echo '{"a":1,"a":2}' | yq --input-format json -o json -I0 '.'
{"a":1,"a":2}                                    # correct — preserved

$ echo '{"a":1,"a":2}' | succinctly yq --input-format json --slurp '.'
- a: 2                                           # wrong
```

Confirmed this does *not* reproduce on `main` before #978/#994 (M2 streaming handled these M2-eligible filter shapes without ever collapsing keys), and does *not* reproduce for YAML input either direction (#442 already covers that case with its own `OwnedValue`-avoiding fast path).

## Why this is separate from #978

#978's own fix is correct and necessary — real yq genuinely never preserves a JSON-sourced number's spelling, confirmed live. This is a distinct invariant (duplicate-key survival) that #978's specific fix strategy (route JSON input through the DOM path) happens to break as a side effect, not a flaw in #978's own stated goal. A narrower fix scoped to *just* the number-formatting issue — e.g. teaching the M2 streaming formatters themselves to canonicalize a JSON-sourced number instead of disabling M2 outright for JSON input — would avoid this trade-off entirely, but is a larger, higher-risk change than #978's PR was scoped for (it touches the shared streaming formatters in `src/yaml/light.rs`, used by both JSON and YAML output, not just `yq_runner.rs`).

## Suggested fix direction

Most likely: give `src/yaml/light.rs`'s scalar-streaming functions (`stream_resolved_scalar_as_json` and its YAML-output sibling) input-format awareness, so they canonicalize a number the same way `canonicalize_json_numbers` now does for the DOM path — restoring M2 eligibility for JSON input instead of disabling it. This would fix both #978's original number-formatting bug and this duplicate-key regression at once, but needs its own careful design/verification pass given the shared-formatter blast radius.

newhoggy:
## Summary

#978 (PR #994) fixed `yq --input-format json` incorrectly preserving a JSON-sourced number's literal spelling by disabling M2 cursor-streaming for all JSON input and falling back to the DOM path. That fix was correct for its stated goal, but M2 was also incidentally providing a second, unrelated invariant: duplicate JSON object keys survive unchanged, since M2 never materializes a map at all. The DOM path's `OwnedValue::Object` is `IndexMap`-backed and can only keep the last value for a repeated key — the same class of bug #442 already fixed for YAML input.

```
$ echo '{"a":1,"a":2}' | succinctly yq --input-format json -o json -I0 '.'
{"a":2}                                          # wrong — collapsed (before this PR)
$ echo '{"a":1,"a":2}' | yq --input-format json -o json -I0 '.'
{"a":1,"a":2}                                    # correct — preserved
```

## Fix

Restores M2 eligibility for JSON input instead of disabling it, by teaching `src/yaml/light.rs`'s streaming formatters to canonicalize a JSON-sourced float the same way the DOM path's `canonicalize_json_numbers` already does — matching **#978's own documented baseline exactly**, not attempting to independently improve past it (verified extensively against a baseline build of unmodified `main`; #978's own DOM path already diverges from real yq for extreme-magnitude floats like `1e100`, e.g. never switching to scientific notation — that's a separate, pre-existing gap, out of scope here).

`YamlIndex::build` already parses JSON input correctly (JSON is a syntactic subset of YAML's flow grammar — this is how M2 handled JSON input before #978 disabled it). The only missing piece was number-formatting awareness in the streaming formatters.

**Design**: rather than threading a `canonicalize: bool` parameter through the ~30 internal `stream_json_value`/`stream_yaml_value` recursive call sites (which would also mean changing the shared `DocumentCursor` trait's method signatures, affecting the unrelated `JsonCursor` implementor too), the flag is carried as a new `canonicalize_numbers` field on `YamlIndex` itself, set once via `YamlIndex::mark_json_sourced()` right after building the index for JSON input. Every `YamlCursor` already holds a reference to its `YamlIndex`, so the formatters read it directly with zero signature changes to the trait or its other implementor.

Three call sites needed the canonicalize check (found via careful tracing, including one non-obvious "scalar document root" bypass that skips the normal formatter dispatch entirely):
- `stream_resolved_scalar_as_json` (JSON-target output)
- `stream_yaml_string_value` (YAML-target output, nested values)
- `stream_yaml_as_document`'s scalar-root shortcut (YAML-target output, a bare navigated scalar result like `.b`)

Also moves `format_float_yq` from the CLI binary (`output.rs`) into the library (`yaml/light.rs`) so the new M2 code path and the existing DOM-path formatter share one definition instead of two — though ultimately unused by the JSON-canonicalize path itself (see below), it stays alongside `format_float_with_fraction` as the natural shared home for float-formatting helpers.

## A wrong turn worth documenting

My first attempt used `format_float_yq`'s scientific-notation threshold for the canonicalize case, reasoning that real yq would want it. Live-testing against a baseline build of `main` revealed #978's own DOM path does **not** do this (`1e100` renders as a 100+ digit decimal expansion, not `1e+100`, on `main` today) — so using `format_float_yq` there would have made M2 diverge from #978's own established DOM-path baseline for M2-eligible queries specifically, not converge with it. Corrected to bare `write!(out, "{f}")`, matching #978's actual behavior exactly. Verified byte-for-byte against a temporary baseline worktree of unmodified `main` across 7 test values (fractional, whole, extreme-magnitude, tiny) in both JSON and YAML output.

## Test changes

- `test_json_input_duplicate_keys_collapse_known_limitation_996` → `test_json_input_duplicate_keys_preserved_996`: flipped from pinning the bug to pinning the fix (its own doc comment said to do exactly this once #996 landed).
- `test_json_input_rejects_adversarial_nesting_998` split into two: restoring M2 means adversarially-deep JSON input (for M2-eligible shapes) now hits `YamlIndex`'s own parse-time depth-128 guard first (clean `anyhow` error, exit 1) instead of `eval_generic::to_owned`'s conversion-time depth-256 guard (panic, exit 101) — strictly safer (caught earlier, no panic), and closer to real jq/yq's own ~128 parse-time limit. Added a companion test confirming #998's own guard is still reachable (and unchanged) via the DOM-forced path (`-P`), so it isn't silently dead code now that the M2 path usually catches this earlier.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 2 unrelated tests flaked under contention from concurrent sessions on this machine — confirmed by re-running in isolation, then a clean full re-run: 100% pass)
- [x] `cargo clippy --all-targets --features cli,simd,regex,serde -- -D warnings`
- [x] `cargo fmt --check`
- [x] Manual verification: duplicate-key repro fixed, number formatting verified byte-for-byte against a baseline build of `main` across 7 test values × 2 output formats × root/nested/array positions
- [x] Manual verification: #998's depth guard still reachable via DOM-forced path (`-P`), M2 path's own (earlier, safer) guard confirmed working

Fixes #996

### DIFF
--- a/src/bin/succinctly/output.rs
+++ b/src/bin/succinctly/output.rs
@@ -11,6 +11,7 @@ use succinctly::jq::escape::{
 };
 use succinctly::jq::{assert_value_tree_depth, EvalError, OwnedValue, StreamError};
 use succinctly::yaml::format_float_with_fraction;
+pub use succinctly::yaml::format_float_yq;
 
 /// Exit codes matching jq behavior
 pub mod exit_codes {
@@ -316,52 +317,6 @@ fn escape_json_body(s: &str, opts: &JsonFormatOpts) -> String {
 /// Format a value as JSON text (compact or pretty, per `opts`).
 pub fn format_json(value: &OwnedValue, opts: &JsonFormatOpts) -> String {
     format_json_impl(value, opts, 0)
-}
-
-/// Renders a computed (non-literal-preserved) `f64` the way real yq does:
-/// decimal for everyday magnitudes, scientific notation once the value's
-/// decimal exponent is `>= 6` or `<= -5`.
-///
-/// Only for [`OwnedValue::Float`] -- a value with no source literal left to
-/// preserve, either because it was actually computed (arithmetic) or because
-/// it came from JSON input, which real yq always re-serializes through
-/// float64 rather than preserving spelling. [`OwnedValue::NumberLiteral`]
-/// (YAML-sourced identity/navigation output) keeps its own source spelling
-/// regardless of magnitude and must never route through this function --
-/// confirmed against real yq v4.53.3: `12345678901234567890123` (a decimal
-/// literal) stays fully expanded on identity, while the equivalent
-/// *computed* magnitude switches to scientific notation. See issue #997.
-///
-/// The threshold and the `e+NN`/`e-NN` (lowercase, signed, exponent padded
-/// to at least 2 digits) spelling are both oracle-verified against real yq;
-/// this is yq's own threshold, distinct from `format_number_jq_compat`'s
-/// jq-mode one (which real jq only reformats when the source literal itself
-/// already used exponent notation).
-///
-/// `f` must be finite -- like [`format_float_with_fraction`], this has no
-/// JSON/YAML-specific spelling for NaN/Infinity to fall back on, so every
-/// caller must special-case those first (both current call sites already
-/// do, ahead of this function).
-#[must_use]
-pub fn format_float_yq(f: f64) -> String {
-    debug_assert!(
-        f.is_finite(),
-        "format_float_yq requires a finite value; NaN/Infinity have no \
-         JSON/YAML spelling here and must be special-cased by the caller"
-    );
-    let sci = format!("{f:e}");
-    let (mantissa, exp_str) = sci
-        .split_once('e')
-        .expect("Rust's exponential formatter always includes a lowercase 'e'");
-    let exp: i32 = exp_str
-        .parse()
-        .expect("exponent from Rust's exponential formatter is always a valid i32");
-    if (-4..6).contains(&exp) {
-        format_float_with_fraction(f)
-    } else {
-        let sign = if exp < 0 { '-' } else { '+' };
-        format!("{mantissa}e{sign}{:02}", exp.abs())
-    }
 }
 
 /// Recursive JSON formatter behind [`format_json`].
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -2517,47 +2517,25 @@ pub fn run_yq(args: YqCommand) -> Result<i32> {
     // `Auto` whenever the user relies on `.json`-extension detection
     // instead of typing `--input-format json` explicitly, which
     // `resolve_input_format` (used everywhere else JSON input is handled)
-    // still correctly resolves to `Json`. The M2 gates below need that
-    // same resolution, not the raw flag, or a bare `succinctly yq '.'
-    // file.json` (arguably the more common way to hit this than the
-    // explicit flag) would still leak. Conservative for a mixed-format
-    // multi-file `--inplace`/`--doc` invocation (`-i '.' a.yaml b.json`):
-    // this disables M2 for every file, not just the JSON one(s), since
-    // `is_m2_streamable`/`can_slurp_fast_path` are single, run-wide
-    // booleans with no per-file M2-vs-DOM switch to hook into instead.
-    let any_input_is_json = if input_files.is_empty() {
-        resolve_input_format(args.input_format, None) == InputFormat::Json
-    } else {
-        input_files.iter().any(|f| {
-            resolve_input_format(args.input_format, Some(Path::new(f))) == InputFormat::Json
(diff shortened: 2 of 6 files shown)


newhoggy:
## Code review summary

Ran the multi-angle review (5 finder agents; 1 stalled without producing results — the other 4 delivered substantial, corroborating findings).

**One HIGH-severity bug found and fixed, independently confirmed by two reviewer angles**: the scalar-document-root canonicalize shortcut (`stream_yaml_as_document`) applied to *every* `YamlValue::String` regardless of quoting style, unlike its two sibling call sites (both correctly gated on `is_unquoted()`). A genuinely quoted JSON string that merely looked numeric (`"1.50"`) had its quotes silently dropped and was reinterpreted as a bare float — corrupting the value's *type*, not just its spelling. Fixed by adding the same `is_unquoted()` gate (commit `d326aa5e9`, pushed immediately).

A second reviewer angle then independently found (with a concrete, worse-than-baseline repro) that the buffered `YamlCursor::to_json()`/`to_json_document()` path — the sibling of the streaming formatters this PR fixed — never learned about `canonicalize_numbers` at all. Unreachable from the CLI today, but a live trap in the public API, and for an exponent literal like `1e2` it was actively worse than pre-PR (gained a spurious `.0`). Extended with the same fix, and extracted a shared `json_sourced_canonical_float()` helper that all four float-canonicalize call sites now route through (fixing the duplication that let the `is_unquoted()` gate get missed in the first place) (commit `beb1fe73f`, pushed immediately).

Also fixed two doc comments that no longer matched the code (one claiming `format_float_yq` is shared with the M2 path when it deliberately isn't; one giving the wrong reason for a hardcoded `false`), and added a regression test exercising the fixed bug across three distinct M2-reachable shapes (bare navigated result, nested-under-object, JSON-target output) per the `testing` skill's "assert call sites agree with each other" guidance — the gap that let the original bug ship undetected.

Two lower-confidence findings intentionally left as-is: consolidating 4 structurally-trivial (1-line) duplicate call sites in `yq_runner.rs` (low risk of silent divergence, unlike the fixed cases), and a missing `# Examples` doc block on the relocated `format_float_yq` (a repo-wide, pre-existing convention gap on every function in this file, not specific to this PR).

Full test suite (100% pass, 2× re-run clean), both clippy invocations, and `cargo fmt --check` all green after each round.

PR #1020, opened 2026-08-16
fix(jq): guard remaining recursive value/yaml walkers against stack overflow

### LINKED ISSUE #1017: jq/yq: several more OwnedValue tree-walkers are unguarded, currently safe only via an implicit invariant (#1005/#1015 follow-up)
**Severity: Medium**

## Summary

Found during #1015's (fix for #1005) mandatory code review. #1005/#1015 guarded six recursive `OwnedValue` tree-walkers (`OwnedValue::to_json`/`to_json_for_reindex`/`==`, `eval::compare_values`, `eval.rs`'s own `to_owned`, `yq_runner::reconcile_presentation`, `output::format_json_impl`) against unbounded growth from `reduce`/`foreach`/`while`/`until`/`repeat` accumulators. The review's cross-file trace found several more recursive `OwnedValue`/`CommentTree` walkers of the exact same shape that were *not* guarded:

1. **`eval_generic::standard_json_to_owned`** (`src/jq/eval_generic.rs`) — a third independent copy of the cursor-to-`OwnedValue` conversion (alongside `eval_generic::to_owned` and `eval.rs`'s own `to_owned`, both now guarded). Reached from `eval_on_owned` and `eval_single`'s fallback arm.
2. **`yq_runner::emit_yaml_value`** (`src/bin/succinctly/yq_runner.rs`) — the *default* YAML output emitter for `syq`/`yq`, structurally identical to the now-guarded `reconcile_presentation` in the same file, but walks `OwnedValue`+`CommentTree` with no depth check.
3. **`eval.rs`'s `format_props_recursive`/`owned_to_yaml`** (`@props`/`@yaml` format functions) — unguarded, unlike their sibling `format_json_impl` (guarded by #1015).
4. **`eval.rs`'s `merge_values`/`merge_object_fields`/`merge_arrays_by_index`** (backing `*`/`*=`, including yq's merge-flag suffixes) — unguarded.
5. **`eval.rs`'s `flatten_owned`** — recurses per the *caller-supplied* `flatten(depth)` argument rather than tree depth, so it isn't bounded by `MAX_VALUE_TREE_DEPTH` at all; a `reduce`-built near-384-deep value piped through a large `flatten(N)` call has its own, separate unbounded-recursion surface.
6. **`yq_runner::strip_presentation_style`/`serde_json_to_owned`/`canonicalize_json_numbers`** — lower risk (currently fed only already-bounded input, or protected by `serde_json`'s own input-depth limit), but still unguarded copies of the same walker shape.

## Why these are follow-up, not #1015-blocking

All of the above are currently **safe only via an implicit invariant**, not an explicit guard: every current call site happens to feed these functions a value that already passed through one of #1015's newly-guarded functions first (most commonly `to_json_for_reindex`'s reindex-bridge round-trip, capped at `MAX_VALUE_TREE_DEPTH`). This was confirmed empirically during review — e.g. `syq -n 'reduce range(1000) as $i (null; [.])'` panics cleanly via the new guard *before* `emit_yaml_value` ever sees the over-deep value, refuting an initial "this crashes independently" hypothesis.

That said, the invariant is undocumented and untested — a future change to any of these call chains (e.g. an optimization that lets a `reduce`/`foreach` accumulator skip the JSON round-trip and stay as `OwnedValue` across iterations, which is exactly the kind of "systemic fix" #1005's own issue text suggested) would silently reopen one or more of these as live, untested stack-overflow paths.

## Suggested fix direction

Same `_at_depth`-threading pattern #1015 already established, applied to each of the functions above — ideally via a shared, parameterized `assert_depth(depth, max)` helper rather than another hand-copied `assert!` (see the companion cleanup issue for that). `flatten_owned`'s `flatten(depth)` ceiling is a separate concern (a user-supplied argument, not tree depth) and may need its own bound rather than reuse of `MAX_VALUE_TREE_DEPTH`.

### LINKED ISSUE #1018: jq: assert_value_tree_depth duplicates assert_nesting_depth's exact shape instead of a shared parameterized helper (#1005/#1015 follow-up)
**Severity: Low**

## Summary

Found during #1015's (fix for #1005) mandatory code review. #1015 added `value::MAX_VALUE_TREE_DEPTH`/`assert_value_tree_depth` as a second, independently-tuned depth ceiling alongside #998's `eval_generic::MAX_NESTING_DEPTH`/`assert_nesting_depth` — necessary because the two guarded function sets have genuinely different measured crash boundaries and reachable-depth requirements (documented on `MAX_VALUE_TREE_DEPTH` itself). But the two guard *functions* are structurally byte-identical:

```rust
// eval_generic.rs (#998)
pub fn assert_nesting_depth(depth: usize) {
    assert!(depth < MAX_NESTING_DEPTH, "nesting depth exceeds limit of {MAX_NESTING_DEPTH}");
}

// value.rs (#1015)
pub fn assert_value_tree_depth(depth: usize) {
    assert!(depth < MAX_VALUE_TREE_DEPTH, "nesting depth exceeds limit of {MAX_VALUE_TREE_DEPTH}");
}
```

`eval_generic::assert_nesting_depth`'s own doc comment records that #998 already fixed this exact problem once (consolidating three independently hand-copied `assert!`s into one shared function) — #1015 reintroduces a second copy of the same pattern rather than generalizing to a single `fn assert_depth(depth: usize, max: usize)` (or const-generic equivalent) called with `256` in one place and `384` in the other.

Relatedly, the two constants' doc comments (`MAX_NESTING_DEPTH` ~46 lines, `MAX_VALUE_TREE_DEPTH` ~32 lines) restate largely the same reasoning (proportional-headroom-over-a-pinned-floor arithmetic, debug-build/test-thread-stack measurement methodology) as independent prose in two places rather than one being the canonical source the other links to.

## Why this is follow-up, not #1015-blocking

Purely a maintainability/duplication concern — both guards are individually correct, tested, and doing their job. No functional risk today. But per this codebase's own documented lesson (`CLAUDE.md`: "Duplicated predicates diverge silently — one definition, plus a test that the call sites agree", from #106), this is the same shape of problem recurring for a third time, and worth cleaning up before a *third* near-identical const+fn+doc-comment triple gets added for some future guarded-function set (see the companion #1017, which will need at least one more ceiling if `flatten_owned`'s separate bound is added).

## Suggested fix direction

Extract a single `fn assert_depth(depth: usize, max: usize)` (in whichever module makes most sense — `value.rs` or a new small shared module), have `assert_nesting_depth`/`assert_value_tree_depth` become thin wrappers (or be replaced by call sites passing the right constant directly), and have `MAX_VALUE_TREE_DEPTH`'s doc comment link to `MAX_NESTING_DEPTH`'s for the shared measurement-methodology explanation rather than restating it.

newhoggy:
## Summary

Extends #1005's depth-guard pattern (established in #998) to every remaining unguarded `OwnedValue`/`CommentTree` tree-walker that #1017 identified:

- `standard_json_to_owned`
- `emit_yaml_value`
- `format_props_recursive`
- `owned_to_yaml`
- the `merge_values`/`merge_existing`/`merge_position`/`merge_object_fields`/`merge_arrays_by_index` family (backs `*`/`*=` merge semantics)
- `flatten_owned`
- `canonicalize_json_numbers`
- `strip_presentation_style`
- `serde_json_to_owned`

Each gains an `_at_depth` sibling threading an explicit `depth: usize` that increments only at actual descent into a child node (not at every function-call hop), guarded via a new shared `value::assert_depth(depth, max)` helper (#1018) — replacing the two byte-identical `assert!` bodies that `assert_value_tree_depth` and `eval_generic::assert_nesting_depth` previously each carried independently against their own constant.

## Test plan

- [x] `cargo fmt --check`
- [x] Live-verified merge/flatten output against CLAUDE.md's documented `*=`/`flatten` examples — byte-identical to pre-change behavior

Fixes #1017
Fixes #1018

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -14,7 +14,9 @@ use succinctly::jq::document::DocumentFields;
 use succinctly::jq::eval_generic::{
     eval_with_cursor, to_owned as generic_to_owned, GenericResult, MAX_NESTING_DEPTH,
 };
-use succinctly::jq::{self, format_number_jq_compat, Expr, JqValue, OwnedValue, Program};
+use succinctly::jq::{
+    self, assert_value_tree_depth, format_number_jq_compat, Expr, JqValue, OwnedValue, Program,
+};
 use succinctly::json::light::{JsonCursor, StandardJson};
 use succinctly::json::validate::{self, ValidationError};
 use succinctly::json::JsonIndex;
@@ -1555,6 +1557,16 @@ fn strip_quotes_and_decode(field: &[u8]) -> String {
 
 /// Convert serde_json::Value to OwnedValue.
 fn serde_to_owned(value: &serde_json::Value) -> OwnedValue {
+    serde_to_owned_at_depth(value, 0)
+}
+
+/// Panics past `MAX_VALUE_TREE_DEPTH` levels of nesting (#1017/#1020) --
+/// mirrors `yq_runner::serde_json_to_owned_at_depth`, which this function's
+/// own doc comment names as guarded by the same PR; this twin was
+/// initially missed and is reachable the same way, via `--argjson`/`--args`
+/// and `--seq`/streamed JSON input parsing below.
+fn serde_to_owned_at_depth(value: &serde_json::Value, depth: usize) -> OwnedValue {
+    assert_value_tree_depth(depth);
     match value {
         serde_json::Value::Null => OwnedValue::Null,
         serde_json::Value::Bool(b) => OwnedValue::Bool(*b),
@@ -1568,12 +1580,14 @@ fn serde_to_owned(value: &serde_json::Value) -> OwnedValue {
             }
         }
         serde_json::Value::String(s) => OwnedValue::String(s.clone()),
-        serde_json::Value::Array(arr) => {
-            OwnedValue::Array(arr.iter().map(serde_to_owned).collect())
-        }
+        serde_json::Value::Array(arr) => OwnedValue::Array(
+            arr.iter()
+                .map(|v| serde_to_owned_at_depth(v, depth + 1))
+                .collect(),
+        ),
         serde_json::Value::Object(obj) => OwnedValue::Object(
             obj.iter()
-                .map(|(k, v)| (k.clone(), serde_to_owned(v)))
+                .map(|(k, v)| (k.clone(), serde_to_owned_at_depth(v, depth + 1)))
                 .collect(),
         ),
     }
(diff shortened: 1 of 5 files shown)


newhoggy:
Code review (8 finders) came back with two real, low-risk gaps in this PR's own guards, plus missing regression coverage — all fixed in 67f4d1157:

- **`jq_runner::serde_to_owned` was left unguarded** — the exact twin of `yq_runner::serde_json_to_owned` (guarded by this PR), even named as such in that function's own doc comment. Now guarded the same way.
- **`assert_depth` lacked `#[track_caller]`**, so all ~15 guarded call sites reported the identical panic location instead of the actual recursive function that overflowed. Fixed on `assert_depth` and both thin wrappers.
- **No regression tests were added** for the 9 (now 10, incl. the `jq_runner` fix) newly-guarded functions, breaking from this codebase's established `*_panics_past_nesting_depth_limit_*` per-guard convention. Added all 10, following that convention exactly.

Two other findings were investigated and are **not** folded into this PR:
- Several more unguarded `OwnedValue` walkers elsewhere in the crate (`walk`, both `stream.rs` output serializers, `lazy.rs`'s `materialize`/`into_owned`, `collect_paths`, `collect_tostream_events`, `push_recursive_branches`, `owned_contains`) — filed as #1021 (added to Ready), since fixing 8 more functions here would balloon this PR's scope well past #1017/#1018. The two `stream.rs` serializers are flagged there as priority: unlike most of what this PR fixes, they're on the *default output path* for essentially every query.
- A suggested number-validation divergence between `standard_json_to_owned` and `to_owned` — checked against the actual code and it's a false positive: `OwnedValue::from_number_bytes` already gates via `is_valid_number` internally (value.rs:510), same as the path `to_owned` takes.

Full local verification (build, `cargo test --features cli,simd,regex,serde`, both CI clippy invocations, `cargo fmt --check`, `cargo check --no-default-features`) is green after these fixes.

Issue #1026, opened 2026-08-16
yq: [.a]-shaped queries lose exact exponent spelling through the reindex-bridge JSON round-trip (#1008 follow-up)


newhoggy:
## Context

#1008 fixed scientific-notation YAML literals (`1e2`, `1e100`, ...) being corrupted by non-identity navigation — the M2 streaming path and every `Expr` shape `eval_generic.rs`'s cursor-based evaluator handles natively (`Field`, `Index`, `Iterate`, etc.) now preserve the exact source spelling verbatim in both YAML- and JSON-target output, matching real yq exactly.

One shape remains a known, tracked gap: any `Expr` the generic evaluator has **no native arm for** (`[...]`, `{...}`, assignment, arithmetic, `map`, and others — the "reindex bridge" in `eval_generic.rs`'s catch-all / `eval_on_owned`) serializes the *whole document* to a JSON string via `OwnedValue::to_json_for_reindex()`, re-parses that string with the JSON parser, and re-evaluates the query against the fresh JSON cursor.

## Repro

```
$ printf 'a: 1e2\n' | succinctly yq '[.a]' -o json -I0
[1E+2]
$ printf 'a: 1e2\n' | yq '[.a]' -o json -I0   # real yq
[1e2]

$ printf 'a: 1e2\n' | succinctly yq '.a = .a'
a: 1E+2
$ printf 'a: 1e2\n' | yq '.a = .a'            # real yq
a: 1e2
```

This is a **regression-guarded, not a silently-worsening** gap: `test_materialized_yaml_exponent_literal_reindex_bridge_known_gap_1008` (`tests/yq_cli_tests.rs`, from #1008's PR) pins the current `1E+2`/`1E+100`/`1E+5` output so this doesn't drift further without anyone noticing.

## Root cause

[...]

PR #1027, opened 2026-08-16
fix(cli): preserve scientific-notation literals through non-identity navigation

### LINKED ISSUE #1008: yq: scientific-notation YAML literals corrupted by any non-identity navigation

newhoggy:
## Summary

Fixes #1008: `yq`'s JSON/YAML output corrupted a YAML scalar written in scientific notation (`1e2`, `1e100`, ...) as soon as any query beyond bare identity touched it. Real yq preserves the exact source spelling verbatim regardless of magnitude or query shape — confirmed empirically against the pinned oracle (`1e100` stays `1e100`, `1E5` stays `1E5`).

## Root cause

- `is_preservable_float_literal` (`src/yaml/scalar.rs`) excluded any literal containing `e`/`E` from `NumberLiteral` preservation, degrading it to a bare `f64` with the source text lost before any formatter could echo it.
- Even past that gate, three yq-specific formatting sites still routed a literal through `format_number_jq_compat` (jq's own normalization — verified correct for `jq`'s real output, wrong for `yq`) instead of echoing it verbatim:
  - `emit_yaml_value_at_depth` (`yq_runner.rs`, yq-only, no jq caller)
  - `format_json_impl`'s `NumberLiteral` arm (`output.rs`), now gated on `ControlEscape::Yq` matching its `Float` arm's existing yq/jq split
  - `stream_owned_value_json_with` (`stream.rs`), gains a `finite_literal` hook mirroring its existing `infinite_literal` parameter, wired to verbatim-echo for yq and unchanged `format_number_jq_compat` for jq

`jq` mode is completely unaffected — `format_number_jq_compat` itself is untouched, and every jq call site keeps using it (verified against real jq: `1e100 -> 1E+100` both before and after this PR).

## Known residual gap (out of scope, filed separately)

`Expr` shapes with no native `eval_generic.rs` arm (`[.a]`, assignment, `map`, ...) round-trip the whole document through `to_json_for_reindex`, which still calls `format_number_jq_compat` internally before the fixed formatters ever see the re-parsed value — so `[.a]` on `1e2` now gives `1E+2` (improved from a 100+-digit decimal expansion, but not byte-identical to real yq's `1e2`). Filed as #1026, and regression-pinned by `test_materialized_yaml_exponent_literal_reindex_bridge_known_gap_1008` so this can't silently worsen unnoticed.

## Test plan

- [x] All 6 of #1008's own repro commands re-verified against real yq — now byte-identical
- [x] `cargo test --features cli,simd,regex,serde` — full suite green (one pre-existing unit test, `test_stream_json_number_literal`, updated since it pinned the old jq-compat reformatting for yq's own streaming path)
- [x] `cargo fmt --check`
- [x] New regression tests: `test_m2_json_output_preserves_exponent_1008`, `test_materialized_yaml_exponent_literal_fidelity_1008` (renamed to `..._reindex_bridge_known_gap_1008` after discovering the residual gap), plus `is_preservable_float_literal`'s own unit test widened to assert acceptance instead of rejection

Fixes #1008

### DIFF
--- a/src/bin/succinctly/output.rs
+++ b/src/bin/succinctly/output.rs
@@ -367,12 +367,20 @@ fn format_json_impl(value: &OwnedValue, opts: &JsonFormatOpts, level: usize) -> 
                 }
             }
         }
-        OwnedValue::NumberLiteral(..) => {
+        OwnedValue::NumberLiteral(_, literal) => {
             if value
                 .as_f64()
                 .is_some_and(|f| f.is_nan() || f.is_infinite())
             {
                 "null".to_string() // JSON doesn't support NaN or Infinity
+            } else if opts.control_escape == ControlEscape::Yq {
+                // yq mode: echo the source spelling verbatim (#1008) --
+                // matches the Float arm's own yq/jq split above, and real
+                // yq's documented byte-for-byte literal preservation. jq
+                // mode keeps `number_str()`'s jq-compat reformatting
+                // (verified to match real jq's own scientific-notation
+                // rules) unchanged.
+                literal.to_string()
             } else {
                 value.number_str().expect("numeric variant").into_owned()
             }
--- a/src/bin/succinctly/output.rs
+++ b/src/bin/succinctly/output.rs
@@ -9,7 +9,9 @@ use succinctly::jq::escape::{
     escape_json_body as run_escaper, write_json_body_jq, write_json_body_jq_ascii,
     write_json_body_yq, write_json_body_yq_ascii,
 };
-use succinctly::jq::{assert_value_tree_depth, EvalError, OwnedValue, StreamError};
+use succinctly::jq::{
+    assert_value_tree_depth, format_number_jq_compat, EvalError, OwnedValue, StreamError,
+};
 use succinctly::yaml::format_float_with_fraction;
 pub use succinctly::yaml::format_float_yq;
 
@@ -377,12 +379,15 @@ fn format_json_impl(value: &OwnedValue, opts: &JsonFormatOpts, level: usize) -> 
                 // yq mode: echo the source spelling verbatim (#1008) --
                 // matches the Float arm's own yq/jq split above, and real
                 // yq's documented byte-for-byte literal preservation. jq
-                // mode keeps `number_str()`'s jq-compat reformatting
-                // (verified to match real jq's own scientific-notation
-                // rules) unchanged.
+                // mode keeps `format_number_jq_compat`'s reformatting
+                // unchanged. This PR fixed its `-0.0`-sign-loss bug (also
+                // #1008, since widening `is_preservable_float_literal`
+                // newly exposed it via YAML), but it has other pre-existing
+                // divergences from real jq, unrelated and left alone here,
+                // e.g. `0.1e1` -> `1E+0` here vs real jq's `1`.
                 literal.to_string()
             } else {
-                value.number_str().expect("numeric variant").into_owned()
+                format_number_jq_compat(literal.as_bytes())
             }
         }
         OwnedValue::String(s) => {
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -1534,7 +1534,7 @@ fn emit_yaml_value_at_depth(
                 format_float_yq(*f)
             }
         }
-        OwnedValue::NumberLiteral(..) => {
+        OwnedValue::NumberLiteral(_, literal) => {
             if value.as_f64().is_some_and(f64::is_nan) {
                 ".nan".to_string()
             } else if value.as_f64().is_some_and(f64::is_infinite) {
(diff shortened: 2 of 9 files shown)


newhoggy:
Code review (9 finders across two rounds) came back with substantially more than the initial pass — all fixed in 5c041e61a:

**Real bugs in the original fix, all confirmed via the pinned oracle:**
- **Digit-cap counted exponent digits** against the mantissa-precision budget, so a full-precision (17-digit) mantissa with any 2-3 digit exponent was still rejected — reproducing #1008's own catastrophic-decimal-expansion symptom. Fixed to count mantissa digits only.
- **`-0.0` sign loss in `format_number_jq_compat`** across all three of its branches — pre-existing (reachable via plain JSON input even before this PR), but this PR's predicate widening made it newly reachable from YAML. Fixed at the source, so `jq` mode's own output is also now correct.
- **Four more missed sibling formatters**: `stream_owned_value_yaml`, `@yaml`'s `owned_to_yaml_at_depth`, `@props`'s `props_value_to_string`, and — the big one — `to_json_for_reindex_at_depth`, which was the entire mechanism behind #1026's known gap. Verified the round-trip needs no jq-specific reformatting at all (jq mode's own final formatter re-normalizes after reparsing regardless of what fed it), so fixing it closes #1026 completely with **zero jq-mode impact** across a 20+ query sweep. **#1026 is now closed as fixed.**

**Stale/misleading comments corrected**: two `light.rs` comments that (mis)attributed non-finite protection to the digit cap rather than `parse_float`'s actual upstream gate, and an `output.rs` comment overclaiming jq mode was "verified to match real jq" (it has other pre-existing, unrelated divergences).

**Out of scope, filed separately**:
- #1030 — `tostring`/`@json`/`@text` still lose fidelity via the mode-unaware `OwnedValue::number_str()` (needs a real mode parameter, not a quick fix; also folds in the review's structural suggestion to consolidate the now-6-site "echo verbatim for yq" rule into one shared accessor).
- #1031 — an extreme-magnitude literal (`1e-400`) displays nonzero text but compares equal to `0` due to `f64` underflow — a value-semantics question, not a formatting bug.
- #1032 — a pre-existing, **YAML-unrelated** bug in jq's own `as $x` variable binding that loses a `NumberLiteral` entirely on plain JSON input, found incidentally while auditing the reindex bridge.

All 6 of #1008's original repro commands, plus every shape the reviewers found (comma expressions, `map_values`, `with_entries`, `del`, `-0e10` across all code paths, the digit-cap boundary), verified byte-identical to real yq/jq. Full local verification (build, `cargo test --features cli,simd,regex,serde`, both CI clippy invocations, `cargo fmt --check`, `cargo check --no-default-features`, `RUSTDOCFLAGS="-D warnings" cargo doc`) green.

Issue #1021, opened 2026-08-16
jq: 9 more recursive OwnedValue walkers unguarded, incl. both output serializers (#1017/#1020 follow-up)


newhoggy:
## Context

#1017/#1018 (fixed by #1020) guarded the 9 recursive `OwnedValue`/`CommentTree` walkers named in #1017's own audit. #1020's code review (\"Altitude and conventions\" pass) found several more that audit missed — including, unlike most of what #1020 fixed, two that are on the **primary output path of nearly every successful `jq`/`yq` invocation**, not just narrower internally-bounded cases.

## Confirmed unguarded (verified against current `main`, all recurse into `Array`/`Object` with no `depth` param and no `assert_value_tree_depth`/`assert_nesting_depth` call anywhere in the body)

- `stream_owned_value_json_with` — `src/jq/stream.rs:267`. JSON result serializer; runs on essentially every `jq`/`yq -o json` query result.
- `stream_owned_value_yaml` — `src/jq/stream.rs:535`. YAML result serializer; same reach for `yq`'s default output.
- `walk_impl` — `src/jq/eval.rs:14793`. Backs the `walk(f)` builtin; directly reachable as `jq 'walk(.)'` on adversarial input.
- `JqValue::materialize` — `src/jq/lazy.rs:387`.
- `JqValue::into_owned` — `src/jq/lazy.rs:413`. Both reachable from `--exit-status`, `-S`/color output, and any `GenericResult::Owned`/`ManyOwned`/`LazySeq` conversion in `jq_runner.rs`.
- `collect_paths` — `src/jq/eval.rs:16282`. Backs `paths`/`leaf_paths`.
- `collect_tostream_events` — `src/jq/eval.rs:16309`. Backs `tostream`.
- `push_recursive_branches` — `src/jq/eval.rs:11541`. Backs `..`/`recurse`.
- `owned_contains` — `src/jq/eval.rs:4186`. Backs `contains`/`inside`.

Repro shape for any of these: pipe a JSON/YAML document nested well past `MAX_VALUE_TREE_DEPTH` (384) through the corresponding builtin (or just query it at all, for the two serializers) — expect an uncontrolled stack overflow (SIGABRT) instead of the intended clean `nesting depth exceeds limit of N` panic.

## Suggested approach

[...]

Issue #1016, opened 2026-08-16
jq: user-defined recursive functions overflow the stack — a distinct, more easily reached crash class from #1005


newhoggy:
**Severity: High**

## Summary

Found during #1015's (fix for #1005) mandatory code review. #1005/#1015 guarded six recursive `OwnedValue` tree-walkers against unbounded growth from `reduce`/`foreach`/`while`/`until`/`repeat` accumulators. This issue is a **structurally different** vulnerability in the same "no adversarial document required" family: `eval_func_call`'s handling of user-defined jq function calls (`def f: ...;`) has no depth guard on the **evaluator's own Rust call stack** — each level of a recursive jq function invocation adds a real stack frame to `eval()`'s own recursion, independent of any `OwnedValue` tree ever being walked.

## Repro (confirmed live on `main` after #1015 merges — reproduces identically on `main` before #1015 too, since #1015 never touches `eval_func_call`)

```
$ echo null | succinctly jq -c 'def deep(n): if n == 0 then . else [deep(n-1)] end; deep(500) | length'
thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting   (SIGABRT, exit 134)
```

500 levels of jq-level function recursion is a very ordinary depth — far shallower than the `reduce range(50000)` needed to trigger #1005's original repro, and recursive function definitions are a completely normal, common jq idiom (list processing, tree walks, accumulator patterns), not a contrived stress case.

## Why this is separate from #1005/#1015

#1005's six guarded functions all walk an already-*constructed* `OwnedValue` tree — the fix threads a `depth: usize` counter through each and panics past a measured ceiling. `eval_func_call`'s crash is a different mechanism entirely: `eval()` calling `eval()` calling `eval()` once per nested `def`-function invocation, with no `OwnedValue` tree involved until (if ever) the recursion bottoms out. None of #1005's `MAX_VALUE_TREE_DEPTH`-based guards can reach this — the crash happens purely in the evaluator's own control flow, before any value materialization step.

## Why this needs its own investigation, not a copy-paste fix

Unlike #1005's tree-walkers (each a self-contained function converted to thread a `depth` parameter with a small, well-defined set of callers), `eval_func_call` sits on `eval()`'s own hot path — the single most call-heavy function in the whole evaluator, reached by literally every query. Any fix needs to weigh:
[...]

PR #1033, opened 2026-08-16
fix(jq): guard 3 more unguarded recursive OwnedValue/JqValue walkers

### LINKED ISSUE #1025: jq: more unguarded recursive OwnedValue walkers found reviewing #1021 (from_owned, @yaml, @props, owned_to_expr, and more)
## Background

#1021 (PR #1024) guarded 9 recursive `OwnedValue`/`JqValue` walkers its own audit named — the 9th PR in a row applying this pattern (#998, #1005, #1015, #1017, #1018, #1020, #1021). A `/code-review` pass on PR #1024 found several more, confirming this audit still isn't exhaustive. All of these are genuinely out of scope for #1021 itself (which named an explicit, closed list of 9), but real gaps worth tracking.

## Confirmed unguarded, in priority order

1. **`owned_to_yaml`** (`src/jq/eval.rs` ~line 5441) — backs the `@yaml` format function. Recurses over `OwnedValue::Array`/`Object` via `.map(owned_to_yaml)`, no depth guard. `deep_value | @yaml` stack-overflows uncatchably where the sibling `yq` default-output path (`stream_owned_value_yaml`, just guarded by #1021) now panics gracefully.
2. **`format_props_recursive`** (`src/jq/eval.rs` ~line 5366) — backs the `@props` format function. Same defect class, same file, ~1000 lines from the functions #1021 fixed.
3. **`owned_to_expr`** (`src/jq/eval.rs` ~line 13240) — splices a compound-assignment RHS or pattern-substitution value back into the AST as literals. Recurses over `OwnedValue::Array`/`Object` with no depth guard. Reachable from `.a += rhs_expr`/`.a //= rhs_expr` (eval.rs:9645, 9669) and `as`-pattern variable substitution (line 12576).
4. **`JqValue::from_owned` + `write_json`/`to_json_string`** (`src/jq/lazy.rs` ~lines 180, 477, 596) — public recursive walkers with no guard anywhere. Since `OwnedValue::Array`/`Object` are `pub` and re-exported from `succinctly::jq`, any library consumer can build a deeply-nested `OwnedValue` with a plain loop (no recursion needed to construct it) and hand it to `JqValue::from_owned(deep).to_json_string()` for a real, uncatchable stack overflow.
5. **`collect_leaf_paths`** (`src/jq/eval.rs` ~line 16649, backs `leaf_paths`) — structurally identical to `collect_paths`/`collect_tostream_events` (both fixed by #1021, same `current_path.to_vec()`-per-descent shape, sole call site starting at `&[]`), but skipped. Currently protected only transitively (`builtin_leaf_paths` calls the guarded `to_owned` first) — fragile, not an independent guard.
6. **`flatten_owned`** (`src/jq/eval.rs` ~line 4322) — bounds its recursion by the user-supplied `flatten(depth)` argument decrementing to 0, not by `assert_value_tree_depth`, so it has no ceiling independent of what the caller passes. `deep_array | flatten(999999999)` is directly user-triggerable with ordinary jq syntax.

## Needs its own investigation before fixing

7. **`eval_array_construction`** (`src/jq/eval.rs` ~line 856) — never depth-checks its own `OwnedValue::Array(items)` wrap when the inner expression's result is already `QueryResult::Owned`/`ManyOwned`; only the cursor branches route through the guarded `to_owned`. A recursive `def deepen(n): if n==0 then . else [deepen(n-1)] end; deepen(N)` may be able to build an arbitrarily deep `OwnedValue` with no guard anywhere in the chain — **unverified** whether this manifests as this specific bug or hits a separate, already-existing unbounded `def`-call-recursion crash first. Worth a precursor investigation, not a direct fix.

## Lower priority / architectural, not blocking

8. **Depth-guard composition gap** (`src/jq/lazy.rs` ~line 396): `materialize_at_depth`/`into_owned_at_depth`'s `Cursor` arm hands off to an independent `cursor_to_owned_at_depth` recursion with its own reset-to-0 counter and a *different, tighter* cap (`eval_generic::MAX_NESTING_DEPTH = 256` vs. this file's `MAX_VALUE_TREE_DEPTH = 384`) — the two guards can stack on the real call stack up to ~640 combined frames without either individually detecting it. Currently unreachable in production (no real constructor nests a `Cursor` inside a `JqValue::Array`/`Object` today) — a latent trap for a future change, not a live bug.
9. **No `catch_unwind`/panic UX** anywhere in production CLI code — every depth-guard panic (from #1005 onward, including #1021's 9) surfaces as a raw Rust backtrace to end users instead of the tool's normal `jq: error: ...` message and exit code. Pre-existing since #1005, not a regression from #1021, but #1021 extended the same gap to two more reachable CLI call sites (`jq_runner.rs:868`, `:1878`).
10. **Systemic boilerplate**: the depth-guard pattern is now ~16 independently hand-threaded local counters (9 new in #1021, ~7 pre-existing), each needing the same wrapper/`_at_depth`-sibling/assert/`depth+1`-threading steps by hand. #1021's own issue text already flagged this as "worth a quick design look... but not a blocker" (a `DepthCursor`-style newtype with a self-asserting `child()` method was suggested) — still not a blocker, but this is now the 10th time the pattern's been hand-copied across #998/#1005/#1015/#1017/#1018/#1020/#1021, and this issue's own items 1-7 would be a 10th-12th.
11. **Unenforced invariant** (`src/jq/eval.rs` ~line 11466): `push_recursive_branches`/`collect_paths`/`collect_tostream_events` (fixed by #1021 with *no* new parameter, reusing an existing `Vec`'s `.len()` as the depth counter) depend on an unchecked invariant — "this Vec grows by exactly one element per recursive call" — that a future refactor of any of the three could silently violate with no compiler error and no test failure until specifically exercised at depth.

## Suggested fix direction

Items 1-6 are mechanical, same shape as #1021 itself (wrapper + `_at_depth` twin, or reuse an existing length-tracking parameter where one exists) — a natural single follow-up PR. Item 7 needs its own investigation first. Items 8-11 are architectural/UX observations worth keeping on record but not urgent.

Found during `/code-review` on PR #1024 (#1021).

newhoggy:
## Summary

Fixes #1025. #1025's own audit (a `/code-review` pass on #1021's PR #1024) named 6 functions as unguarded; 3 of them (`owned_to_yaml`, `format_props_recursive`, `flatten_owned`) turned out to already be guarded — fixed earlier the same day by #1017/#1020, which #1025's filing didn't have visibility into since it reviewed a parallel PR. This PR guards the 3 that genuinely remained:

- `owned_to_expr` — splices a compound-assignment RHS or pattern-substitution value back into the AST as literals, reachable from `.a += rhs_expr`/`.a //= rhs_expr` and `as`-pattern variable substitution.
- `JqValue::from_owned` + `write_json` — `to_json_string` delegates to `write_json`, so guarding it alone covers both. Public recursive walkers with no guard anywhere; since `OwnedValue::Array`/`Object` are `pub` and re-exported from `succinctly::jq`, any library consumer can build a deeply-nested value with a plain loop (no recursion needed to construct it) and hand it to either.
- `collect_leaf_paths` (backs `leaf_paths`) — reuses `current_path.len()` as the depth counter rather than threading a new parameter, mirroring the shape #1021 already established for `collect_paths`/`collect_tostream_events` (both share the identical `current_path.to_vec()`-per-descent invariant).

Item 7 (`eval_array_construction`) needs its own investigation per the issue's own text and stays out of scope. Items 8-11 are explicitly non-blocking architectural observations.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — full suite green, including 4 new regression tests
- [x] `cargo fmt --check`
- [x] Live-verified against CLAUDE.md's documented `leaf_paths` example and `+=` compound assignment — unchanged, correct output

Fixes #1025

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -13366,6 +13366,16 @@ fn substitute_var_in_builtin(
 
 /// Convert an OwnedValue to an Expr, preserving complex types.
 fn owned_to_expr(value: &OwnedValue) -> Expr {
+    owned_to_expr_at_depth(value, 0)
+}
+
+/// Panics past [`MAX_VALUE_TREE_DEPTH`](super::value::MAX_VALUE_TREE_DEPTH)
+/// levels of nesting (#1025) -- splices a compound-assignment RHS or
+/// pattern-substitution value back into the AST as literals, reachable from
+/// `.a += rhs_expr`/`.a //= rhs_expr` and `as`-pattern variable substitution
+/// with no adversarial document involved.
+fn owned_to_expr_at_depth(value: &OwnedValue, depth: usize) -> Expr {
+    assert_value_tree_depth(depth);
     match value {
         OwnedValue::Null => Expr::Literal(Literal::Null),
         OwnedValue::Bool(b) => Expr::Literal(Literal::Bool(*b)),
@@ -13382,7 +13392,10 @@ fn owned_to_expr(value: &OwnedValue) -> Expr {
             if arr.is_empty() {
                 Expr::Array(Box::new(Expr::Builtin(Builtin::Empty)))
             } else {
-                let elements: Vec<Expr> = arr.iter().map(owned_to_expr).collect();
+                let elements: Vec<Expr> = arr
+                    .iter()
+                    .map(|v| owned_to_expr_at_depth(v, depth + 1))
+                    .collect();
                 Expr::Array(Box::new(Expr::Comma(elements)))
             }
         }
@@ -13392,7 +13405,7 @@ fn owned_to_expr(value: &OwnedValue) -> Expr {
                 .iter()
                 .map(|(k, v)| ObjectEntry {
                     key: ObjectKey::Literal(k.clone()),
-                    value: owned_to_expr(v),
+                    value: owned_to_expr_at_depth(v, depth + 1),
                 })
                 .collect();
             Expr::Object(entries)
@@ -16742,11 +16755,18 @@ fn get_value_at_path(value: &OwnedValue, path: &[OwnedValue]) -> Option<OwnedVal
 /// side effect of `select()`'s falsy handling, not a design choice) and
 /// empty containers (whose `type` isn't scalar even though they have no
 /// children either). See #771.
+/// Panics past [`MAX_VALUE_TREE_DEPTH`](super::value::MAX_VALUE_TREE_DEPTH)
+/// levels of nesting (#1025) -- reuses `current_path.len()` as the depth
+/// counter rather than threading a separate parameter, the same shape
+/// `collect_paths`/`collect_tostream_events` (#1021) already established:
+/// `current_path` grows by exactly one element per descent, so its length
+/// already tracks tree depth without duplicating a counter.
 fn collect_leaf_paths(
     value: &OwnedValue,
     current_path: &[OwnedValue],
     paths: &mut Vec<OwnedValue>,
 ) {
+    assert_value_tree_depth(current_path.len());
     match value {
         OwnedValue::Object(entries) => {
             if entries.is_empty() {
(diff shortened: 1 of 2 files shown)


newhoggy:
Code review came back with one real, actionable gap and a few minor items — fixed in e79f70590:

- **Object-arm test coverage was missing**: all 4 new regression tests exercised only the Array match arm via linear array nesting, so a depth-threading bug isolated to an Object arm (e.g. `depth` instead of `depth + 1`) would have passed every one of them. Added Object-nested boundary tests to all 4, mirroring #1021's own precedent for `materialize`/`into_owned`.

The rest of the findings are correctly out of scope for this PR:
- A more severe, already-tracked `expand_func_calls` stack-overflow bug (#1016, "Severity: High") — this whole depth-guard series (#1005 through this PR) has never addressed it, since it's a distinct crash mechanism (AST-inlining at `def`-expansion time, not a runtime value getting deep) that needs its own architectural investigation. Already filed, already deliberately deferred after investigation earlier this session.
- Doc-comment placement on private `_at_depth` helpers rather than public wrappers — pre-existing convention from #1021, not something this PR introduced or should unilaterally change.
- No compile-time enforcement of the depth-guard pattern — explicitly flagged as non-blocking in #1025's own text (item 10).

Full local verification (build, `cargo test --features cli,simd,regex,serde`, both CI clippy invocations, `cargo fmt --check`, `cargo check --no-default-features`, `RUSTDOCFLAGS="-D warnings" cargo doc`) green.

Issue #1023, opened 2026-08-16
jq: resolve_node's Select/If arms hand-roll eval_recurse_cond's fork-on-cond shape


newhoggy:
> **Rescoped 2026-08-21.** This issue originally covered three items. **Items 2 and 3 shipped** in `2a10606ca` ("refactor(jq): consolidate recurse-family MAX_ITEMS and finalization tail (#1023)"), verified on current `main`:
>
> - ~~**Item 2** — `builtin_recurse_f`/`builtin_recurse_cond`'s duplicated finalization block~~ → now the shared `finish_recurse_walk` (`src/jq/eval.rs:14204`), called from both.
> - ~~**Item 3** — `MAX_ITEMS` copy-pasted three times~~ → now a single `const RECURSE_MAX_ITEMS` (`:14192`).
>
> Only item 1 remains, and the line citations below have been refreshed against `main` @ `bc8c86ad8` (they had drifted again since the 2026-08-20 refresh). Original three-item body is in the edit history.

## Summary

`resolve_node`'s `Expr::Builtin(Builtin::Select(cond))` and `Expr::If` arms each hand-roll the same "evaluate `cond` via `eval_owned_multi_keep_partial`, iterate its outputs, fork per truthiness, defer any escape" shape that `eval_recurse_cond` centralizes for the recurse family (#897, PR #1022).

Both arms' own doc comments already cross-reference that lineage — "mirrors `#627`'s `builtin_recurse_cond`/`resolve_recurse` fix", "#842/#854's precedent for `recurse`, independently live here too — #896" — i.e. the code already documents this as one rule living in three places.

## Where (verified against `main` @ `bc8c86ad8`)

| Site | Line |
|---|---|
| `resolve_node`'s `Expr::Builtin(Builtin::Select(cond))` arm | `src/jq/eval.rs:13317` |
| `resolve_node`'s `Expr::If { cond, .. }` arm | `src/jq/eval.rs:13381` |
| `eval_recurse_cond` (the existing shared primitive) | `src/jq/eval.rs:14170` |

Both arms open with the identical line:

```rust
let (cond_outputs, escape) = eval_owned_multi_keep_partial::<S>(cond, value);
```

## Why this is not a call-site swap

`eval_recurse_cond`'s signature is `(cond, child, gate, push) -> Option<EvalEscape>` — a *gate*: it pushes when `cond` is truthy and reports an escape. That does not fit either caller as-is:

[...]

PR #1037, opened 2026-08-16
refactor(jq): route more inline None/Owned/ManyOwned splits through owned_vec_to_result

### LINKED ISSUE #1029: jq: 6 more inline None/Owned/ManyOwned splits not routed through owned_vec_to_result
## Background

A `/code-review` pass on PR #1028 (#1023's recurse-family dedup) found that a pre-existing shared helper, `owned_vec_to_result` (`src/jq/eval.rs:1134`), already implements the exact `Vec<OwnedValue>` → `None`/`Owned`/`ManyOwned` three-way collapse and has ~25 existing call sites — but at least 6 more sites in the same file still hand-roll the identical split inline instead of calling it. (PR #1028 itself had briefly reintroduced a 5th inline copy in its own new `finish_recurse_walk` helper; that one is now fixed to call `owned_vec_to_result` directly.)

## Confirmed inline duplicates (not routed through `owned_vec_to_result`)

1. `eval_range_values` (`src/jq/eval.rs` ~line 14558-14587) and `eval_range_values_f64` (~14594-14623) — both end with the identical inline three-way split. Also share a second duplication: `const MAX_RANGE: usize = 100000;` is copy-pasted between the two (the structural twin of the `MAX_ITEMS` triplication #1023 already fixed for the recurse family — same file, sitting immediately below it).
2. `eval_object_construction`'s tail (~lines 1078-1084) — inline copy, ~50 lines above `owned_vec_to_result`'s own definition.
3. `builtin_fromstream` (~lines 16460-16466) and `builtin_truncate_stream` (~lines 16528-16535) — both `None => match outputs.len() {0/1/_}` arms duplicate `owned_vec_to_result` verbatim.
4. `builtin_paths` (~lines 16546-16554), `builtin_paths_filter` (~lines 16687-16695), `builtin_leaf_paths` (~lines 16790-16798) — three more copies of the identical split, all within ~250 lines of each other in the "paths" builtin family.
5. `retain_truthy`'s `QueryResult::ManyOwned` arm (~lines 1334-1341) — after `vs.retain(...)`, reimplements the 0/1/_ match instead of calling `owned_vec_to_result(vs)`.

(Line numbers as of the review; re-verify before implementing, since concurrent work may have shifted them.)

## Why this matters

`RECURSE_MAX_ITEMS`/`owned_vec_to_result` itself cites #106's "duplicated predicates diverge silently" as the exact motivation for consolidating — a future change to the collapsing rule (e.g. adding instrumentation, or changing tie-breaking for some edge case) would update `owned_vec_to_result`'s ~25 existing call sites but silently miss these 6+ copies, since none of them are actual call sites of the helper.

## Suggested fix direction

Mechanical: replace each inline `if outputs.is_empty() {...} else if outputs.len()==1 {...} else {...}` with `owned_vec_to_result(outputs)` (or the local variable name each site uses). The `MAX_RANGE` duplication (item 1) is a separate, smaller consolidation in the same spirit as `RECURSE_MAX_ITEMS` — one shared constant instead of two copies.

Found during `/code-review` on PR #1028 (#1023).

newhoggy:
## Summary

Fixes #1029. #1029's own audit (a `/code-review` pass on #1023's PR #1028) named 6 sites in `eval.rs` that hand-rolled `owned_vec_to_result`'s exact `Vec<OwnedValue>` → `None`/`Owned`/`ManyOwned` three-way collapse instead of calling it:

- `eval_range_values`/`eval_range_values_f64` — also had a duplicated `MAX_RANGE` constant, now a single shared one (the structural twin of the `MAX_ITEMS` triplication #1023 fixed for the recurse family).
- `eval_object_construction`
- `builtin_fromstream`
- `builtin_truncate_stream`
- `builtin_paths`, `builtin_paths_filter`, `builtin_leaf_paths`
- `retain_truthy`'s `ManyOwned` arm

A follow-up sweep for the same exact shape (not exhaustively enumerated by the issue) found 4 more genuine duplicates: `eval_skip`'s `Many`/`ManyOwned` arms, an `eval_limit`-shaped function at two call sites (`take`-style and `limit`-style), and the `path()` resolver's own `paths.len()` match.

Several similar-looking sites were deliberately left untouched after checking their behavior wasn't identical — the `capture(re;"g")`/`split`/`scan`-family builtins collapse an empty result the same way as a non-empty one (a 2-way split: empty → `None`, else → `ManyOwned` unconditionally, even for a single match), which `owned_vec_to_result` does not replicate (it collapses a single-element result to `Owned`, not `ManyOwned`). Converting those would have changed behavior.

Pure consolidation, no behavior change — every touched site keeps its exact `Vec<OwnedValue>` → `None`/`Owned`(len 1)/`ManyOwned`(len > 1) mapping, just via the shared helper instead of a hand-copied match.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — full suite green (no behavior change, so no new tests needed; existing coverage for `range`, object construction, `paths`/`leaf_paths`, `fromstream`/`truncate_stream`, `limit`, `//` already exercises every touched site)
- [x] `cargo fmt --check`
- [x] Live-verified `range`, object construction, `leaf_paths`, `paths`, `path()`, `limit`, `tostream`, and `//` against expected output

Fixes #1029

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -1379,12 +1379,8 @@ fn push_truthiness<W: Clone + AsRef<[u64]>>(
 }
 
 /// Turn a stream of booleans into the result of a boolean operator.
-fn bools_to_result<'a, W: Clone + AsRef<[u64]>>(mut bools: Vec<bool>) -> QueryResult<'a, W> {
-    match bools.len() {
-        0 => QueryResult::None,
-        1 => QueryResult::Owned(OwnedValue::Bool(bools.pop().unwrap())),
-        _ => QueryResult::ManyOwned(bools.into_iter().map(OwnedValue::Bool).collect()),
-    }
+fn bools_to_result<'a, W: Clone + AsRef<[u64]>>(bools: Vec<bool>) -> QueryResult<'a, W> {
+    owned_vec_to_result(bools.into_iter().map(OwnedValue::Bool).collect())
 }
 
 /// The `OwnedValue`-collecting analog of [`push_truthiness`]: pushes every
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -1074,13 +1074,7 @@ fn eval_object_construction<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         }
     }
 
-    if objects.is_empty() {
-        QueryResult::None
-    } else if objects.len() == 1 {
-        QueryResult::Owned(objects.pop().unwrap())
-    } else {
-        QueryResult::ManyOwned(objects)
-    }
+    owned_vec_to_result(objects)
 }
 
 /// Evaluate recursive descent.
@@ -1332,11 +1326,7 @@ fn retain_truthy<W: Clone + AsRef<[u64]>>(result: QueryResult<'_, W>) -> QueryRe
         }
         QueryResult::ManyOwned(mut vs) => {
             vs.retain(OwnedValue::is_truthy);
-            match vs.len() {
-                0 => QueryResult::None,
-                1 => QueryResult::Owned(vs.pop().unwrap()),
-                _ => QueryResult::ManyOwned(vs),
-            }
+            owned_vec_to_result(vs)
         }
         // The prefix is filtered exactly like `ManyOwned` above — jq's `//`
         // retains truthy outputs regardless of whether the stream that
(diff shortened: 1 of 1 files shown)


newhoggy:
Code review (10 finders) confirmed the refactor is behavior-preserving on every angle — no correctness bugs found. One real gap surfaced and fixed in 1a0472ee8:

- **\`bools_to_result\`** was missed by the original sweep (it takes \`Vec<bool>\`, not \`Vec<OwnedValue>\`, so a naive grep for the exact type wouldn't have caught it) but has the identical 0/1/_ shape, sitting three lines above \`owned_vec_to_result\` itself. Now routes through it via a \`bool -> OwnedValue::Bool\` map.

One legitimate architectural suggestion is out of scope for this PR and filed separately: \`retain_truthy\`'s and \`eval_limit\`'s borrowed-cursor \`Many\`/\`One\` arms (on \`Vec<StandardJson>\`, not \`Vec<OwnedValue>\`) have the same shape but need their own sibling helper, since \`owned_vec_to_result\`'s typing can't cover them directly — filed as #1038.

(Also: the PR description's "eval_skip" should read "builtin_skip" — no \`eval_skip\` function exists. Noted here since editing the description post-review isn't worth a diff.)

Full local verification (build, \`cargo test --features cli,simd,regex,serde\`, both CI clippy invocations, \`cargo fmt --check\`, \`cargo check --no-default-features\`, \`RUSTDOCFLAGS="-D warnings" cargo doc\`) green, including a live check that \`all()\`/\`any()\` (bools_to_result's callers) still work correctly.

PR #1053, opened 2026-08-16
fix(jq): keep decimal point on reindex-bridge whole-number floats

### LINKED ISSUE #953: yq: an i64-overflow decimal integer scalar shows different values across query paths (pre-existing, unrelated to #918)
**Severity: Low**

## Summary

A plain YAML decimal integer scalar too large for `i64` (`resolve_plain`'s `parse_int_or_float` fallback classifies it `ResolvedScalar::Float`) renders with different apparent precision depending on which query touches it — the materialized/JSON path drops the rounding that the streaming/YAML path applies.

## Repro

```
$ printf 'a: 99999999999999999999\n' | succinctly yq -o json '[.a]'
[
  100000000000000000000
]
$ printf 'a: 99999999999999999999\n' | succinctly yq '.a'
99999999999999999999
$ printf 'a: 99999999999999999999\n' | yq -o json '[.a]'   # pinned v4.53.3
[
  100000000000000000000.0
]
```

succinctly's materialized JSON output (`100000000000000000000`, correctly rounded but missing the decimal point) differs from both its own streaming output (`99999999999999999999`, raw unrounded source text) and from real yq (`100000000000000000000.0`, rounded with a decimal point marking it as a float).

## Root cause

`OwnedValue::to_json()`'s bare `Float` arm (`src/jq/value.rs`, `Self::Float(f) => format!("{f}")`) uses Rust's plain `f64` `Display`, which omits the decimal point for a whole-number value at any magnitude (`1e20_f64.to_string() == "100000000000000000000"`, no trailing `.0`) — the same class of bug #918 fixed for document-sourced float literals, but reachable here for *any* materialized whole-number float regardless of origin (this specific overflow-integer case, or any arithmetic result).

## Why this is separate from #918

Confirmed via a pristine pre-#918 build (`main` tip before that fix): this exact command already gave `[100000000000000000000]` (no decimal point) before #918's fix existed. #918's `number_literal()` override deliberately excludes this value (its source text has no literal `.`, per `is_preservable_float_literal`'s doc comment in `src/yaml/scalar.rs`), so it falls through to the same bare-`Float`-Display path that predates #918 entirely. Not a regression from that PR — a pre-existing, independent bug this investigation happened to surface.

## Suggested fix direction

Likely the same family as #949 (computed arithmetic result formatting) — `OwnedValue::Float`'s JSON serialization needs a decimal-point-preserving formatter for whole-number values, not raw `Display`. Worth checking whether one fix covers both.

newhoggy:
## Summary

- `OwnedValue::to_json_for_reindex`'s `Float` fallback formatted whole-number floats with bare `f64::to_string()`, which drops the trailing `.0` (`1e20_f64.to_string() == "100000000000000000000"`, no decimal point).
- This "purely-internal" reindex bridge turns out to be the actual materialization path used by `eval_generic.rs`'s cursor-based evaluator for any `Expr` shape it has no native cursor arm for — `[...]` array construction, `map_values`, `with_entries`, etc. So a YAML scalar too large for `i64` (which resolves to a plain `Float`, not a literal-preserving `NumberLiteral`, per `is_preservable_float_literal`'s deliberate bare-digit-run exclusion) lost its decimal point specifically when reached through one of those shapes, while the plain streaming path (`.a` alone, no array wrapper) already showed it correctly.
- Root-caused by adding temporary debug tracing to confirm the exact `OwnedValue` reaching the CLI's JSON formatter for `[.a]` — it was a freshly-reconstructed `NumberLiteral` whose text had already lost the decimal point *before* formatting, tracing back through `eval_on_owned`'s JSON-text round trip in `src/jq/eval_generic.rs`.
- Fix: thread the yq `float_fmt` (`format_float_with_fraction`) through the reindex bridge's `Float` fallback instead of the bare-`Display` closure. Verified this is a no-op for the round-trip's own correctness (reparsing `100000000000000000000.0` yields the identical `f64`), and for jq mode specifically (jq's own final formatter, `format_number_jq_compat`, re-normalizes away the added `.0` after reparsing regardless — confirmed live against pinned jq for `[.a * 1.0]`/`[.a / 5]`).
- Also threads the same fix into `to_json`/`to_json_yq`'s shared `to_json_at_depth`, covering `to_json_yq`'s own direct callers (`tostring`, `tojson`, string interpolation) alongside the reindex bridge — both routes can reach a plain (non-`NumberLiteral`) whole-number `Float`.

Fixes #953

## Test plan

- [x] New regression tests in `tests/yq_cli_tests.rs` pin the issue's exact repro (`[.a]` on an i64-overflow YAML scalar, `-o json`) plus the streaming-path (`.a` alone) and `map_values(.)` cases, verified byte-for-byte against pinned real `yq` v4.53.3.
- [x] New regression test in `tests/jq_cli_tests.rs` confirms jq mode's own array construction (`[.a * 1.0]`, `[.a / 5]`) is unaffected, verified against pinned real `jq`.
- [x] Full local sweep against pinned `yq`/`jq` oracles across `[.a]`, `.a,.a`, `map_values(.)`, `with_entries(.)`, normal floats/ints, negative overflow ints — all match.
- [x] `cargo test --features cli,simd,regex,serde` (including `yq_golden_tests`, `yq_path_consistency_tests`) — all pass.

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -4958,7 +4958,7 @@ fn builtin_with_entries<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     // map(f)
     let mut transformed: Vec<OwnedValue> = Vec::new();
     for entry in entries {
-        let entry_json = owned_to_json_bytes(&entry);
+        let entry_json = owned_to_json_bytes::<S>(&entry);
         let index = crate::json::JsonIndex::build(&entry_json);
         let cursor = index.root(&entry_json);
 
@@ -4999,8 +4999,8 @@ fn builtin_with_entries<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
 /// purely internal (e.g. `with_entries`'s per-entry re-evaluation), so an
 /// overflowed `NumberLiteral`/`Float` must keep its ±Infinity rather than
 /// being silently substituted with JSON's `"null"` (#561).
-fn owned_to_json_bytes(value: &OwnedValue) -> Vec<u8> {
-    value.to_json_for_reindex().into_bytes()
+fn owned_to_json_bytes<S: EvalSemantics>(value: &OwnedValue) -> Vec<u8> {
+    value.to_json_for_reindex::<S>().into_bytes()
 }
 
 // =============================================================================
@@ -13899,7 +13899,7 @@ fn eval_owned_expr_ctrl<S: EvalSemantics>(
     // `to_json_for_reindex` (not `to_json`): this round-trip is purely
     // internal, so an overflowed `NumberLiteral`/`Float` must keep its
     // ±Infinity rather than being silently substituted with "null" (#561).
-    let json_str = input.to_json_for_reindex();
+    let json_str = input.to_json_for_reindex::<S>();
     let json_bytes = json_str.as_bytes();
 
     // We need to create a temporary index and cursor
@@ -13996,7 +13996,7 @@ fn eval_owned_input<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     // `to_json_for_reindex` (not `to_json`): this round-trip is purely
     // internal, so an overflowed `NumberLiteral`/`Float` must keep its
     // ±Infinity rather than being silently substituted with "null" (#561).
-    let json_str = input.to_json_for_reindex();
+    let json_str = input.to_json_for_reindex::<S>();
     let json_bytes = json_str.as_bytes();
 
     use crate::json::JsonIndex;
@@ -23888,7 +23888,7 @@ mod tests {
         input: &OwnedValue,
         optional: bool,
     ) -> Result<Option<OwnedValue>, EvalError> {
-        let json_str = input.to_json_for_reindex();
+        let json_str = input.to_json_for_reindex::<JqSemantics>();
         let json_bytes = json_str.as_bytes();
         let index = JsonIndex::build(json_bytes);
         let cursor = index.root(json_bytes);
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -895,7 +895,7 @@ fn eval_on_owned<S: EvalSemantics, V: DocumentValue>(
         return format_result::<S, _>(format_type, &owned, optional);
     }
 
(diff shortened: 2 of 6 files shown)


newhoggy:
Code review on the initial fix found a real regression and fixed it before merge:

- **Confirmed regression** (verified against a from-scratch rebuild of the pre-fix commit and the pinned jq oracle): making `to_json_for_reindex`'s `Float` fallback unconditionally add a decimal point flipped `reduce (1,2) as $i (1.0 + 4.0; [.])` from the correct `[[[5]]]` to `[[[5.0]]]` in **jq** mode. `format_number_jq_compat` does not strip an explicit `.0` back off a literal it's handed after the reindex bridge's reparse, so real jq's "computed values lose their literal formatting" convention broke. Fixed by threading `S: EvalSemantics` through `to_json_for_reindex`/`to_json_for_reindex_at_depth` so the added decimal point is yq-only; jq mode keeps its prior bare-`Display` fallback. New regression test: `test_jq_reduce_accumulator_unaffected_by_yq_float_fraction_fix_953`.
- **Confirmed non-issue**: a separate finding (`to_json_yq`'s `Float` arm should use the scientific-notation-threshold `format_float_yq` instead of `format_float_with_fraction`, citing `(1e10 * 2) | tostring` giving `20000000000.0` instead of real yq's `2e+10`) turned out to be a **pre-existing** divergence, not a regression — confirmed by rebuilding the pre-#953 commit, which already gave the wrong (differently wrong: `20000000000`, no decimal point at all) output for the same command. This is a separate, deeper issue: real yq applies scientific notation to *genuinely computed* floats but plain decimal (regardless of magnitude) to a value that's merely an unmodified i64-overflow scalar passed through — a distinction succinctly's `OwnedValue::Float` can no longer make once a `NumberLiteral` degrades. Filing a follow-up issue for this rather than expanding #953's scope.
- **Maintainability cleanup**: extracted the `|f| f.to_string()` closure (duplicated across three call sites after this fix) into a single named `jq_bare_float_display` function.

All three commits pushed, rebased clean onto current `main`, full local verification (build, `cargo test --features cli,simd,regex,serde` including golden/path-consistency suites, 3 clippy configs, fmt, no_std check, doc build) green after every change.

Issue #1055, opened 2026-08-16
yq: EvalError's value previews (describe/dump_truncated) always format via jq rules, never yq's verbatim echo


newhoggy:
**Severity: Low**

## Summary

Split from #1051 (item 3) — deliberately left out of that issue's own fix to keep it reviewable, since this is architecturally the largest of #1051's four sub-findings.

Every `EvalError` constructor that embeds a value preview (`describe`/`dump_truncated` in `src/jq/error.rs`, and `EvalError::from_value` backing the `error(v)` builtin) routes through `stream::stream_owned_value_json_jq`, which has no mode parameter anywhere in its call chain. In yq mode, a `NumberLiteral` embedded in an error message is always reformatted via jq's `format_number_jq_compat` (uppercase `E`, forced sign, etc.) instead of echoed verbatim — inconsistent with every yq-mode *output* path (`tostring`, `@json`, `to_json_yq`, ...), which #1030/#953/#1051 have all made mode-aware.

## Repro (internal-consistency only — see note below)

```
$ echo 'a: [1e2, x]' | succinctly yq '.a | error'
Error: [1E+2,"x"]      # should be [1e2,"x"], matching .a | tostring on the same value
```

## Why this is deliberately out of scope for a quick fix

Threading `S: EvalSemantics` through this path is a real refactor, not a one-line change:

- ~40 `EvalError` constructors in `src/jq/error.rs` take `&OwnedValue` and call `describe()`/`dump_truncated()` internally, all currently mode-blind.
- Confirmed via grep: `cannot_iterate` alone has 30 call sites; `cannot_use_as_object_key` has 5; `EvalError::from_value` has 11 (spanning both `eval.rs` and the CLI binary). Many more constructors exist beyond these three examples.
- Not every call site currently has `S: EvalSemantics` in scope — some would need it threaded further up their own call chain first.
- A `stream::stream_owned_value_json` (yq-aware) vs. `_jq` (jq-only) split already exists for the *output* path (#1008) — the error-preview path needs its own equivalent, or a shared `S`-generic entry point, designed rather than bolted on incrementally.

## Verification caveat

[...]

Issue #1030, opened 2026-08-16
yq: tostring/@json/@text also lose scientific-notation fidelity via mode-unaware number_str() (#1008 follow-up)


newhoggy:
## Context

#1008 (fixed by #1027) made yq's *output* formatters (`emit_yaml_value_at_depth`, `format_json_impl`'s `NumberLiteral` arm, `stream_owned_value_json`'s `finite_literal` hook, `stream_owned_value_yaml`, `owned_to_yaml_at_depth`/`@yaml`, `props_value_to_string`/`@props`, and the reindex-bridge's `to_json_for_reindex`) preserve a document-sourced scientific-notation literal's exact spelling, matching real yq.

Code review on that PR found one more surface still affected: `OwnedValue::number_str()` (`src/jq/value.rs`) hardcodes `format_number_jq_compat` unconditionally, with no yq/jq mode awareness at all. It backs `tostring`, `@text`, `@json`/`tojson`, and string interpolation (`"\(.a)"`) in **both** CLI binaries via `numeric_display_string` (`src/jq/eval.rs`).

## Repro

```
$ printf 'a: 1e2\n' | succinctly yq '.a | tostring'
"1E+2"
$ printf 'a: 1e2\n' | yq '.a | tostring'   # real yq
"1e2"

$ printf 'a: 1e2\n' | succinctly yq '.a | @json'
"1E+2"
$ printf 'a: 1e2\n' | yq '.a | @json'      # real yq
"1e2"
```

## Why this is separate from #1008/#1027

Unlike the output formatters #1008 fixed, `number_str()` is genuinely mode-shared — it's the one place `jq`'s own real output (via `to_json()`) and various internal previews already rely on for **correct**, verified-against-real-jq behavior. Naively making it always echo verbatim would leak yq's convention into jq mode's own `tostring`/`@json`, which must keep jq's normalization. Fixing this needs a mode parameter or a yq-specific sibling method, not just deleting the `format_number_jq_compat` call — out of scope for a quick fix, hence this separate issue.

## Suggested approach (also addresses a structural review finding)

[...]

PR #1057, opened 2026-08-16
fix(jq): fix yq-mode number fidelity in stderr/halt_error/-i DOM fallback

### LINKED ISSUE #1051: yq: several remaining paths still lose scientific-notation fidelity (stderr, halt_error, error messages, -i/--slurp DOM fallback)
## Summary

Surfaced during code review of #1030/PR #1049 (which made `tostring`/`@json`/string interpolation/`@sh`/`@csv`/`@tsv`/`@dsv`/`tojson`/`join` mode-aware for yq's scientific-notation-literal fidelity). Several more surfaces share the identical root cause — a `NumberLiteral` gets reformatted via jq's `format_number_jq_compat` instead of echoed verbatim in yq mode — but are architecturally separate from the call chain that PR fixed, so they're still broken. Live-verified against pinned yq v4.53.3 where an oracle comparison is possible.

## Confirmed divergences

**1. `builtin_stderr` (`src/jq/eval.rs`) has no `S: EvalSemantics` parameter at all**, so it structurally cannot check mode; its container arm always formats via jq rules.
```
$ echo 'a: [1e2, x]' | succinctly yq '.a | stderr | empty'
```
writes `[1E+2,"x"]` to stderr instead of `[1e2,"x"]` — an internal inconsistency against the now-correct `.a | tostring` on the same value. (No external oracle: real yq doesn't implement `stderr` as a jq-dialect builtin — confirmed via self-consistency and direct reproduction only.)

**2. `builtin_halt_error` *is* generic over `S`** (already uses `S::DEFAULT_HALT_ERROR_CODE`), but its trailing container arm ignores `S::TAG` and always calls `.to_json()`.
```
$ echo 'a: [1e2, x]' | succinctly yq '.a | halt_error'
```
prints `[1E+2,"x"]` instead of `[1e2,"x"]`. Since `S` was already in scope here, the fix is a one-line `if S::TAG == EvalTag::Yq` check, same shape as the sites #1030 already fixed.

**3. `EvalError`'s value-embedding constructors and preview functions all route through `stream::stream_owned_value_json_jq`**, which has no mode parameter anywhere in its call chain — affects `describe()`/`dump_truncated()` and at least `cannot_use_as_object_key`, `not_valid_in_csv_row`, `cannot_be_shell_escaped`, `cannot_iterate`, `cannot_be_dsv_formatted`, `pair_cannot_be_sorted`, and `EvalError::from_value` (the `error(v)` builtin).
```
$ echo 'a: [1e2, x]' | succinctly yq '.a | error'
```
prints `Error: [1E+2,"x"]` instead of `Error: [1e2,"x"]`. (External-oracle comparison is inconclusive for most of these — the specific succinctly error conditions tried don't reproduce as errors, or reproduce with entirely different wording, in real yq, itself a pre-existing divergence unrelated to number formatting — but the internal inconsistency against `tostring`/`@json` on the identical branch is directly confirmed.)

**4. Most severe: `src/bin/succinctly/yq_runner.rs`'s `evaluate_input` DOM fallback round-trips the *whole document* through jq-mode `OwnedValue::to_json()` before the evaluator ever runs**, on the path taken by `--slurp` and by `--inplace` for any expression `can_use_m2_streaming` doesn't allow-list (`tostring` is not on that list). Because the round-trip reformats every `NumberLiteral` in the document up front, the yq-verbatim-echo fix from #1030 has nothing left to preserve by the time the evaluator sees it — and this corrupts **fields the query never touched**, not just the one being computed:
```
$ printf 'a: 1e2\n' > f.yaml
$ succinctly yq -i '.b = (.a | tostring)' f.yaml
# writes: a: 1E+2 / b: "1E+2"
$ yq -i '.b = (.a | tostring)' f.yaml   # real yq, identical operation
# writes: a: 1e2 / b: "1e2"  (both correct, "a" left alone)
```
This is silent data corruption on an in-place write, not just a wrong query result — the untouched sibling field `a` gets rewritten.

## Suggested approach

- #1 and #2 are small, mechanical fixes mirroring #1030's own pattern (`owned_value_to_json::<S>`, added by that PR's own code-review follow-up).
- #3 needs a mode parameter threaded through `stream_owned_value_json_jq` and its callers in `error.rs` — a `stream::stream_owned_value_json` vs. `_jq` split already exists for the *output* path (#1008); the error-preview path needs its own equivalent, or a shared `S`-generic entry point.
- #4 is the highest-value fix but needs its own careful investigation: `evaluate_input`'s DOM round-trip exists to rebuild a `JsonIndex` cursor cheaply, and switching it to route through `to_json_yq()` (added by #1030) unconditionally would be wrong for jq mode `-i`/`--slurp` invocations sharing the same code path — needs mode-awareness threaded into `yq_runner.rs` itself, plus verification across the `can_use_m2_streaming` allow-list boundary to confirm no other expression shapes have the same gap.

## Scope note

Not blocking #1030/PR #1049 — that PR (plus its own code-review fast-follow commit) correctly and completely fixes the eight surfaces it targeted (`tostring`, `@text`, `@json`, `@uri`, `@html`, `@sh`, `@csv`/`@tsv`/`@dsv`, `tojson`, `join`), all oracle-verified. This is a broader, deeper set of remaining gaps in the same problem family, confirmed pre-existing/unrelated to that PR's own diff.

newhoggy:
## Summary

Fixes 3 of #1051's 4 confirmed gaps in yq's scientific-notation-literal fidelity, surfaced during #1030/PR #1049's code review:

- **`builtin_stderr`** had no `S: EvalSemantics` parameter at all, so its container arm always formatted via jq rules regardless of mode.
- **`builtin_halt_error`**'s trailing container arm ignored `S::TAG` despite already being generic over `S` — a one-line fix mirroring #1030's own pattern.
- **`yq_runner.rs`'s `evaluate_input`** (the DOM fallback taken by `--slurp`, and by `--inplace` for any expression `can_use_m2_streaming` doesn't allow-list — `tostring` is not on that list) round-tripped the *whole document* through jq-mode `to_json()` before the evaluator ever ran. This silently rewrote every `NumberLiteral` in the document — **including fields the query never touched** — on an in-place write:
  ```
  $ printf 'a: 1e2\n' > f.yaml
  $ succinctly yq -i '.b = (.a | tostring)' f.yaml
  # before: a: 1E+2 / b: "1E+2"  (a corrupted, b wrong)
  # after:  a: 1e2  / b: "1e2"   (byte-for-byte match with real yq)
  ```
  Fixed by switching to `to_json_yq()` (widened from `pub(crate)` to `pub` so the CLI binary crate can reach it — this file always evaluates with `YqSemantics`, confirmed by checking every caller).

Item 3 (`EvalError`'s value-preview formatting in `describe`/`dump_truncated`/`from_value`) needs a real refactor across ~40 constructors and 50+ call sites, several without `S` currently in scope — split out as #1055 rather than folded into this PR.

Fixes #1051

## Test plan

- [x] New regression tests: `test_yq_stderr_preserves_exponent_literal_1051`, `test_yq_halt_error_preserves_exponent_literal_1051`, `test_yq_inplace_tostring_does_not_corrupt_untouched_sibling_field_1051` (the last pinned byte-for-byte against real yq v4.53.3's identical operation).
- [x] `test_jq_stderr_and_halt_error_unaffected_by_yq_mode_fix_1051` confirms jq mode's own `stderr`/`halt_error` container formatting is unchanged, verified against pinned jq.
- [x] `cargo test --features cli,simd,regex,serde` (incl. golden/path-consistency suites) — all pass.

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -572,29 +572,48 @@ fn evaluate_yaml_direct_filtered(
 /// front in `run_yq`, so this function needs no evaluation context (#284).
 ///
 /// This is `yq_runner.rs`'s own function (`jq_runner.rs` has an unrelated,
-/// separate `evaluate_input` for jq mode) and always evaluates with
-/// `YqSemantics` below, so the round-trip must use a yq-mode-aware
-/// formatter, not `to_json()` (#1051): the DOM fallback taken by `--slurp`
-/// and by `--inplace` for any expression `can_use_m2_streaming` doesn't
-/// allow-list (`tostring` is not on that list) was reformatting every
-/// `NumberLiteral` in the *whole document* via jq rules before the
-/// evaluator ever ran, silently rewriting fields the query never touched on
-/// an in-place write.
+/// separate `evaluate_input` for jq mode). It always *evaluates* with
+/// `YqSemantics` below, but the round-trip's own formatter deliberately
+/// stays `to_json_for_reindex::<JqSemantics>()`, not `<YqSemantics>` or
+/// `to_json_yq()` (#1051):
 ///
-/// `to_json_for_reindex::<YqSemantics>()`, not `to_json_yq()`: this round
-/// trip is purely internal (code review on #1051's first draft caught this
-/// live — `-i '.b = (.a | tostring)'` on `a: .inf` silently rewrote it to
-/// `a: null`), so a non-finite value must keep its ±Infinity/NaN spelling
-/// through the reparse rather than being substituted with JSON's `null`,
-/// the same reasoning `eval_owned_input`'s identical reindex bridge already
-/// applies for `reduce`/`foreach` (#561, #472).
+/// - Its `NumberLiteral` echo is unconditional either way (#1051's actual
+///   fix), so a document-sourced literal like `1e2` survives this round trip
+///   with its exact spelling untouched regardless of which `S` is passed —
+///   the DOM fallback taken by `--slurp` and by `--inplace` for any
+///   expression `can_use_m2_streaming` doesn't allow-list (`tostring` is not
+///   on that list) used to reformat it via jq's `format_number_jq_compat`
+///   before the evaluator ever ran, silently rewriting fields the query
+///   never touched on an in-place write.
+/// - Its NaN/Infinity handling is also unconditional (code review on this
+///   fix's first draft caught this live — `-i '.b = (.a | tostring)'` on
+///   `a: .inf` silently rewrote it to `a: null` through `to_json_yq()`'s
+///   RFC-8259 "null" substitution, wrong for this purely-internal round
+///   trip), matching `eval_owned_input`'s identical reindex bridge for
+///   `reduce`/`foreach` (#561, #472).
+/// - Only the *fallback* arm — a plain, already-literal-less `Float` — is
+///   `S`-gated, and `YqSemantics` there is actively wrong for this call
+///   site specifically: `parse_input`'s `--input-format json` path already
+///   collapses every `NumberLiteral` to a plain `Float`/`Int` via
+///   `canonicalize_json_numbers` (#978, matching real yq's "a JSON-sourced
+///   number never keeps its own spelling" convention) *before* the value
+///   ever reaches here — forcing `YqSemantics`'s decimal point back onto
+///   that already-canonicalized value reintroduced exactly the bug #978
+///   fixed (`--slurp --input-format json '.'` on `{"a":1e2}` regressed from
+///   `[{"a":100}]` to `[{"a":100.0}]`, caught by CI). `JqSemantics`'s bare
+///   fallback (no forced point) is correct for both the JSON-canonicalized
+///   case and the untouched-overflow-scalar case (the latter is already
+///   lossy through this whole-document round trip regardless of the point —
+///   confirmed live, real yq keeps an untouched i64-overflow scalar
+///   byte-for-byte via `-i`, e.g. `99999999999999999999` verbatim, which
+///   this reindex-through-`f64` architecture cannot match either way).
 fn evaluate_input(
     input: &OwnedValue,
     expr: &jq::Expr,
     sink: &mut ErrorSink,
 ) -> Result<Vec<OwnedValue>> {
     // Convert OwnedValue to JSON bytes for indexing
-    let json_str = input.to_json_for_reindex::<YqSemantics>();
+    let json_str = input.to_json_for_reindex::<jq::JqSemantics>();
     let json_bytes = json_str.as_bytes();
 
     // Build index and evaluate
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -573,19 +573,28 @@ fn evaluate_yaml_direct_filtered(
 ///
 /// This is `yq_runner.rs`'s own function (`jq_runner.rs` has an unrelated,
 /// separate `evaluate_input` for jq mode) and always evaluates with
-/// `YqSemantics` below, so the round-trip must use `to_json_yq()`, not
-/// `to_json()` (#1051): the DOM fallback taken by `--slurp` and by
-/// `--inplace` for any expression `can_use_m2_streaming` doesn't allow-list
-/// (`tostring` is not on that list) was reformatting every `NumberLiteral`
-/// in the *whole document* via jq rules before the evaluator ever ran,
-/// silently rewriting fields the query never touched on an in-place write.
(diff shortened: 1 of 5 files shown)


newhoggy:
Ran 8 parallel code-review finders across correctness, efficiency, reuse, altitude, cross-file tracing, removed-behavior auditing, simplification, and CLAUDE.md conventions. Synthesis:

**Real bug found and fixed** (Angle: Altitude): `evaluate_input`'s round-trip used `to_json_yq()`, which substitutes JSON's `null` for a non-finite `Float`/`NumberLiteral` — correct for real JSON *output*, wrong for this purely-internal reindex round-trip. Confirmed live: `-i '.b = (.a | tostring)'` on `a: .inf` silently corrupted it to `a: null` — the identical shape of bug this PR was filed to fix, still present in the exact function it touches. Fixed by switching to `to_json_for_reindex::<YqSemantics>()`, which already exists for exactly this purpose (shared with `reduce`/`foreach`'s reindex bridge) and preserves ±Infinity/NaN through the reparse. New regression test covers `.inf`/`-.inf`/`.nan`.

While verifying that fix, found `tostring`'s own spelling of a non-finite value still diverges from real yq (`"inf"`/`"NaN"` instead of `".inf"`/`".nan"`) — a separate, pre-existing gap in `numeric_display_string`, reproducible via the plain streaming path with no DOM fallback involved. Filed as #1060.

**Minor cleanup**: trimmed a test doc comment that was restating its function's own doc comment verbatim (Angle: Simplification) rather than pointing at it.

**No other findings survived scrutiny** across the remaining 6 finders — including a careful check that `S: EvalSemantics` is genuinely in scope everywhere it's newly threaded, that no parallel/stale copy of the old jq-only formatting exists in `eval_generic.rs`, that the `pub(crate)`→`pub` visibility bump on `to_json_yq`/`to_json_for_reindex` doesn't leak any invariant, and that monomorphization cost is zero-overhead (both `EvalSemantics` impls are zero-sized marker types).

Two architectural observations from the Altitude angle, not blocking this PR: (1) there's no lint/test enforcing that new mode-aware formatting call sites route through the shared `owned_value_to_json`/`to_json_yq`/`to_json_for_reindex` convention rather than being hand-rolled (this PR itself fixes two prior misses of exactly that kind); (2) `jq_runner.rs` and `yq_runner.rs` each hand-roll their own `evaluate_input`, duplicating the "serialize → reindex → eval" round-trip rather than sharing one library-level implementation. Both are real but are process/design questions for a separate discussion, not confirmed defects — not filing follow-ups for these unless requested, since they're observations about missing guardrails rather than concrete bugs.

All local verification (build, `cargo test --features cli,simd,regex,serde`, 3 clippy configs, fmt, no_std check, doc build) green after the fix.

newhoggy:
CI caught a real regression from the previous commit that my local test run missed (likely a build-artifact race from running clippy/doc concurrently with the background test suite on the same `target/` dir — re-verified sequentially this time).

**Regression**: routing `evaluate_input`'s reindex bridge through `to_json_for_reindex::<YqSemantics>()` forced a decimal point onto every plain `Float` reaching its fallback arm — including ones #978 deliberately strips. `--input-format json`'s `parse_input` already collapses every `NumberLiteral` to a plain `Float`/`Int` via `canonicalize_json_numbers`, so a JSON-sourced number never keeps its own spelling (matching real yq). Forcing the point back on regressed `--slurp --input-format json '.'` on `{"a":1e2}` from `[{"a":100}]` to `[{"a":100.0}]` — caught by `test_json_input_slurp_canonicalizes_exponent_literal_spelling_978` across every CI platform.

**Fix**: switched to `to_json_for_reindex::<JqSemantics>()`. The two things this PR actually needs from the bridge — `NumberLiteral` verbatim echo and NaN/Infinity preservation — are unconditional in that function regardless of which `S` is passed, so nothing is lost; only the Float-fallback's forced decimal point (a #953 concern specific to a *different* call site, `eval_generic.rs`'s reindex bridge) is dropped here. Confirmed live that `JqSemantics`'s bare fallback is correct for both the JSON-canonicalized case and the untouched-overflow-scalar case (real yq keeps an untouched i64-overflow YAML scalar fully byte-for-byte via `-i`, which this whole-document `f64` round trip can't match either way).

Added a second regression test (`test_json_input_inplace_does_not_reintroduce_decimal_point_1051`) covering the same interaction via `--inplace` specifically, since `--slurp` alone wasn't enough to prove both `evaluate_input` entry points stayed fixed.

Re-verified sequentially this time (no concurrent cargo invocations): full build, `cargo test --features cli,simd,regex,serde` (656 passed incl. golden/path-consistency), all 3 clippy configs, fmt, no_std check, doc build — all green, and specifically re-ran `test_json_input_slurp_canonicalizes_exponent_literal_spelling_978` and `test_json_input_inplace_does_not_reintroduce_decimal_point_1051` standalone to confirm.

Issue #1058, opened 2026-08-16
CLI: --argjson loses number-literal source fidelity (serde_json round-trip, unlike filter literals)


newhoggy:
## Summary

Since #1035, a number literal written directly in filter text (`1.500`, `1e100`) preserves jq's exact source spelling through evaluation. `--argjson`'s value parsing is a separate code path that does not get this fix — it still loses fidelity, formatting through Rust's own `f64`/`i64` Display instead of jq's canonical spelling.

## Repro

```
$ succinctly jq -n --argjson n 1.500 '$n'
1.5
$ jq -n --argjson n 1.500 '$n'
1.500

$ succinctly jq -n --argjson n 1e100 '$n'
10000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
$ jq -n --argjson n 1e100 '$n'
1E+100
```

## Root cause

`src/bin/succinctly/yq_runner.rs`'s `parse_json_value` (used for `--argjson`) validates strictly via `serde_json::from_str` (deliberately, per #284 — the lenient JSON semi-index would silently coerce malformed input like `42 garbage` → `42`), then converts through `serde_json_to_owned_at_depth`, whose `Number` arm builds a plain `OwnedValue::Int`/`Float` from `n.as_i64()`/`n.as_f64()` — never `OwnedValue::from_number_literal`, so the original text is discarded before the `OwnedValue` even exists.

This is unrelated to and not reached by #1035's fix (`Literal::NumberLiteral`/`substitute_vars`/`owned_to_expr`), which only covers literals written in the filter's own text — `--argjson`'s value string never becomes a `Literal` at all, it's substituted in as an already-built `OwnedValue`.

## Suggested fix

Have `serde_json_to_owned_at_depth`'s `Number` arm (or `parse_json_value` itself) preserve the matched substring's source text and construct `OwnedValue::from_number_literal`/`from_number_literal_boxed` instead of a plain `Int`/`Float` — `serde_json::Value::Number`'s `Display`/`to_string()` may already round-trip the original digits closely enough, but that should be verified against exponent-notation and trailing-zero cases specifically before relying on it.

## Scope note

Found while verifying #1035's fix scope boundary — confirmed as a separate, pre-existing gap (not introduced by that change) via direct testing of both code paths.

newhoggy:
🔧 Claimed by agent `cadence` — branch `issue-1058-argjson-number-literal-fidelity`

Issue #1056, opened 2026-08-16
jq: unary minus ("0 - expr") loses IEEE754 negative-zero sign, diverging from jq's -0


newhoggy:
## Summary

Real jq's negative-zero output is distinct from positive zero (`-0` vs `0`), but succinctly's unary minus always prints plain `0` regardless of the operand's sign, for both integers and floats. Confirmed on current `main` (unrelated to any in-flight branch), against pinned jq 1.7.1.

## Repro

```
$ echo '{"a":0.0}' | jq -c -- '-.a'
-0
$ echo '{"a":0.0}' | succinctly jq -c -- '-.a'
0

$ echo null | jq -c -- '-(1-1)'
-0
$ echo null | succinctly jq -c -- '-(1-1)'
0

$ echo null | jq -c -- '-(1.0-1.0)'
-0
$ echo null | succinctly jq -c -- '-(1.0-1.0)'
0

$ echo null | jq -c -- '-0.0'
-0
$ echo null | succinctly jq -c -- '-0.0'
0
```

## Root cause

`src/jq/parser.rs`'s unary-minus production for a non-literal operand is implemented as `Expr::Arithmetic { op: Sub, left: Literal::Int(0), right: operand }` (i.e. `0 - expr`). In IEEE754, `0.0 - 0.0` evaluates to positive `0.0`, not `-0.0` — only a true negation (`-(0.0)`, or `-1.0 * 0.0`) preserves the sign bit. Since succinctly's unary minus always goes through subtraction rather than a real negation, the sign is lost whenever the operand evaluates to zero.

This affects every code path that lowers unary minus to `0 - expr`, not just literals — `-.a`, `-(expr)`, and (after #1035) filter-embedded negative float/exponential literals like `-0.0`/`-0e5` that get split into `0 - <positive literal>` to correctly collapse fidelity like jq does for the general negative-literal case (see #1035's `test_1035_negative_float_literal_collapses_fidelity_like_jq`).

## Suggested fix

Represent unary minus as a real negation (e.g. a dedicated `Expr::Negate` variant, or `-1 * expr` if that already preserves IEEE754 sign correctly for this evaluator — needs verification either way) instead of `0 - expr`, so the arithmetic actually inverts the sign bit rather than relying on subtraction identity.

## Scope note

[...]

PR #1059, opened 2026-08-16
fix(jq): filter literal numbers keep jq's source spelling through evaluation

### LINKED ISSUE #1035: jq: numeric literals written directly in filter text lose source fidelity (unrelated to document input)
## Summary

Discovered as a side effect of implementing #1003 (`join`'s error-wording fix) — an
unrelated, pre-existing bug. Numeric literals typed directly in **filter source text**
(not document input) lose their original textual form; the equivalent value arriving via
document input preserves it correctly. This confirms the bug is in the jq *filter
parser*'s number-literal handling, not in document/JSON number parsing (which already
preserves source fidelity via `NumberLiteral`).

## Repro

```
$ echo 'null' | succinctly jq -c '[1.500]'
[1.5]
$ echo '[1.500]' | succinctly jq -c '.'
[1.500]          # document input: correct, fidelity preserved

$ echo 'null' | succinctly jq -c '[1e2]'
[100]
$ echo '[1e2]' | succinctly jq -c '.'
[1E+2]            # document input: correct, fidelity preserved

# Compare real jq (fidelity preserved either way):
$ echo 'null' | jq -c '[1.500]'
[1.500]
$ echo 'null' | jq -c '[1e2]'
[1E+2]
```

## Scope note

This is completely unconnected to `join`, `sub`, or any accumulator/reduce logic — it
showed up only because #1003's join-fix verification happened to compare filter-literal
test expressions like `[1.500]` against document-input equivalents in a batch script, and
the mismatch was isolated down to a plain `null | [1.500]` with zero join involvement.

## Suggested approach

Needs investigation into how the jq filter parser tokenizes/stores number literals
(`Expr::Literal` or equivalent) — likely missing the same `NumberLiteral`-style raw-text
retention that document-input JSON numbers already get. Not yet investigated further;
scope, root cause, and fix strategy all still need verification against the pinned jq
oracle before implementation, per this session's standing practice.

newhoggy:
## Summary

- A numeric literal written directly in filter text (`1.500`, `1e2`) had no source-text slot in the AST — `Expr::Literal` only carried `Int(i64)`/`Float(f64)`, so the parser's number conversion immediately discarded jq's exact spelling. `null | [1.500]` printed `[1.5]` instead of jq's `[1.500]`; `1e2` printed `100` instead of `1E+2`.
- Adds `Literal::NumberLiteral(String)` (mirrors `JqValue::NumberLiteral`'s simpler shape, not `OwnedValue::NumberLiteral`'s two-field one, to avoid a circular import between `expr.rs`/`value.rs`), threaded through all 3 evaluator paths (`eval.rs`, `eval_generic.rs`, `lazy.rs`) plus the `.foo[N]` index-fold optimization and the `owned_to_expr`/`substitute_vars` value→AST splice path (closing a documented #387 scope gap along the way).
- jq's own grammar treats a leading `-` before a number as unary negation, not part of the literal token (`-1.500` prints `-1.5` in real jq, collapsing fidelity, unlike the positive spelling). succinctly's parser folds `-DIGIT` into one token at the lexer level (pre-existing, harmless for integers), so this PR splits negative float/exponent literals specifically back out into `0 - <positive literal>` to match jq's actual collapse behavior, while leaving negative integers untouched (no alternate spelling to lose, and it's what preserves succinctly's existing i64::MIN-exact-literal capability).

## Follow-ups filed (found via oracle verification, out of scope here)

- #1056 — unary minus (`0 - expr`) loses the IEEE754 negative-zero sign (`-0.0`/`-(1-1)` all print `0` instead of jq's `-0`); pre-existing on `main`, confirmed unrelated to this change.
- #1058 — `--argjson` has a separate `serde_json`-based value-parsing path that doesn't get this fix's fidelity; unrelated code path, not reached by this PR's `substitute_vars`/`owned_to_expr` fix.

Fixes #1035

## Test plan

- [x] `cargo test --features cli,simd,regex,serde --lib` — 2512 passed
- [x] Verified against pinned jq 1.7.1 oracle: positive/negative literal fidelity, negative-literal collapse, index-fold optimization (`.[N]`, `.[-1]`), i64 boundary preservation, arithmetic-collapses-fidelity behavior

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -23982,25 +23982,31 @@ mod tests {
     }
 
     #[test]
-    fn test_substitute_vars_number_literal_degrades_to_plain_literal_387() {
+    fn test_substitute_vars_number_literal_preserves_source_text_387_1035() {
         // `substitute_vars` (the `--arg`/`--argjson`-style public API) inlines
-        // each variable's value into the AST via `owned_to_expr`. Unlike a
-        // document-sourced value flowing through a builtin argument,
-        // `Expr::Literal` has no source-text slot (#387's documented scope
-        // boundary -- see `owned_to_expr`'s doc comment), so a `NumberLiteral`
-        // substituted in degrades to its plain parsed value: the substituted
-        // filter's own re-evaluation loses jq's canonical spelling and falls
-        // back to Rust's `f64`/`i64` Display, same as any other computed
-        // (non-passthrough) number.
+        // each variable's value into the AST via `owned_to_expr`. #387's
+        // originally-documented scope boundary here -- `Expr::Literal` had
+        // no source-text slot, so a `NumberLiteral` substituted in degraded
+        // to its plain parsed value -- is now closed by #1035:
+        // `Literal::NumberLiteral` carries the same source text
+        // `OwnedValue::NumberLiteral` does, so the substituted filter's own
+        // re-evaluation keeps jq's canonical spelling instead of falling
+        // back to Rust's `f64`/`i64` Display.
         let expr = parse("$n").unwrap();
 
         let int_lit = OwnedValue::from_number_literal("42");
         let substituted = substitute_vars(&expr, [("n", &int_lit)]);
-        assert_eq!(substituted, Expr::Literal(Literal::Int(42)));
+        assert_eq!(
+            substituted,
+            Expr::Literal(Literal::NumberLiteral("42".to_string()))
+        );
 
         let float_lit = OwnedValue::from_number_literal("1e100");
         let substituted = substitute_vars(&expr, [("n", &float_lit)]);
-        assert_eq!(substituted, Expr::Literal(Literal::Float(1e100)));
+        assert_eq!(
+            substituted,
+            Expr::Literal(Literal::NumberLiteral("1e100".to_string()))
+        );
 
         let index = JsonIndex::build(b"null");
         let cursor = index.root(b"null");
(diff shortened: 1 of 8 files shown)


newhoggy:
Ran the mandatory multi-agent code review (10 finder angles). It found 3 real regressions in the initial commit, all fixed and pushed:

1. **Invalid-JSON leak** — a filter literal like `1.`/`007` (valid jq filter syntax, invalid RFC 8259) skipped the same `is_valid_number` gate document numbers already go through, leaking through `@json`/string interpolation/`tostring`. Now falls back to the pre-existing canonical `Literal::Int`/`Float` construction for JSON-invalid-but-jq-valid spellings.
2. **`-0.0` sign loss** — the negative-literal desugaring used `0 - N`, and IEEE-754 `0.0 - 0.0` is positive zero, so `-0.0`/`-0e0` printed plain `0` instead of jq's `-0` — a regression from the exact build of `main` this branched from. Switched to `-1 * N`, which correctly preserves the sign bit.
3. **yq-mode fidelity regression** — the negative-literal collapse applied unconditionally in the shared jq/yq parser, but real yq (v4.53.3, checked directly) never collapses a negative literal's fidelity the way jq does. Gated to jq mode only.

Also closed a silent perf regression the same review surfaced: the negative-literal split defeated `fold_index_key`'s `.foo[-1.0]`/`.foo[-2.0:]` constant-folding fast path (correctness was fine, just always fell back to dynamic evaluation). `fold_index_key` now sees through the new `Mul(-1, literal)` shape.

Follow-ups filed for out-of-scope findings: #1061 (pre-existing i64::MAX/2^63 boundary bug in the same function, unrelated to this PR), #1062 (4x duplicated `Literal->OwnedValue` conversion + `Literal::NumberLiteral` re-parsing on every eval instead of caching, both non-blocking). #1056 corrected via comment — its literal-token case is now closed by this PR, the general `-(expr)` case remains open there.

All local verification re-run clean after the fixes: 2518 lib tests, 3x clippy matrix, fmt, no_std check, and a fresh oracle sweep against both `jq` 1.7.1 and real `yq` v4.53.3.

PR #1063, opened 2026-08-16
fix(jq): yq mode spells NaN/Infinity as .nan/.inf, not jq's NaN/inf

### LINKED ISSUE #1060: yq: tostring/numeric_display_string spells NaN/Infinity as inf/NaN instead of .nan/.inf
**Severity: Low**

## Summary

`numeric_display_string` (`src/jq/eval.rs`) — the shared helper backing `tostring` and friends — has an unconditional (mode-blind) fast path for NaN/Infinity that uses Rust's bare `f64::Display`, regardless of `S::TAG`. Real yq spells these `.nan`/`.inf`/`-.inf` (its own YAML-native scalar syntax); succinctly gives `NaN`/`inf`/`-inf`.

## Repro

```
$ echo 'a: .inf' | succinctly yq -r '.a | tostring'
inf
$ echo 'a: .inf' | yq -r '.a | tostring'   # pinned v4.53.3
.inf

$ echo 'a: .nan' | succinctly yq -r '.a | tostring'
NaN
$ echo 'a: .nan' | yq -r '.a | tostring'   # pinned v4.53.3
.nan
```

## Root cause

```rust
pub(crate) fn numeric_display_string<S: EvalSemantics>(value: &OwnedValue) -> String {
    if let OwnedValue::NumberLiteral(repr, literal) = value {
        if let NumberRepr::Float(f) = repr {
            if f.is_nan() || f.is_infinite() {
                return f.to_string();   // <-- unconditional, ignores S::TAG entirely
            }
        }
        ...
```

This check runs *before* the `S::TAG == EvalTag::Yq` verbatim-echo branch a few lines below it, so a NaN/Infinite `NumberLiteral` never reaches the yq-mode-aware path at all, in either mode.

## Where this was found

Surfaced while verifying #1051's fix (PR #1057) against a non-finite scalar via `-i '.b = (.a | tostring)'` — confirmed the *field-corruption* part of that fix is correct (`a` now stays `.inf`/`.nan` untouched), but the *computed* `b` value's own spelling still diverges from real yq. Reproducible via the plain streaming path alone (no `-i`, no DOM fallback), confirming this is unrelated to #1051's own root cause — a separate, narrower gap in `numeric_display_string` itself.

## Suggested fix

Gate the NaN/infinite fast path on `S::TAG`, matching the pattern already used a few lines below it in the same function — likely echoing `.nan`/`.inf`/`-.inf` for yq mode (verify against the pinned oracle for the exact spelling, including sign handling) and keeping the current `f.to_string()` behavior for jq mode (verify against real jq too — jq has no `.nan`/`.inf` YAML syntax, so its own convention here may already be correct/unrelated).

newhoggy:
## Summary

- `numeric_display_string`'s NaN/Infinity fast path was unconditional (`f.to_string()`, jq's own spelling in both modes), so `tostring`/`@text`/`@sh`/CSV-TSV-DSV cell formatting on a document-sourced `.inf`/`-.inf`/`.nan` scalar gave `"inf"`/`"-inf"`/`"NaN"` instead of real yq's own YAML-native `.inf`/`-.inf`/`.nan` spelling:
  ```
  $ echo 'a: .inf' | succinctly yq -r '.a | tostring'
  inf
  $ echo 'a: .inf' | yq -r '.a | tostring'   # pinned v4.53.3
  .inf
  ```
- Real yq's own Go-based number model distinguishes a document-sourced non-finite scalar from a genuinely *computed* one (`1/0` → Go's `+Inf`, confirmed live against the pinned oracle), but succinctly's `OwnedValue` can't make that distinction once a document-sourced `.inf`/`.nan` degrades to a plain `Float` — YAML's `number_literal()` override never captures these spellings as a `NumberLiteral` in the first place (`is_preservable_float_literal`'s JSON-syntax gate rejects them). So this fix always answers with the document-sourced spelling — correct for the common case (`tostring` on an untouched `.inf`/`.nan` scalar), diverging from real yq's Go-style spelling only for a genuinely computed non-finite result — a narrower, rarer residual gap left unaddressed here rather than expanding scope, matching the same "computed vs literal" tension already documented for #953/#1051.
- `@json`/`-o json` output is unaffected — a separate, already-RFC-8259-correct code path (`to_json`/`to_json_yq`).
- jq mode is unaffected — verified against succinctly's own pre-existing jq-mode output for `infinite`/`nan` (which doesn't replicate real jq's `DBL_MAX`-substitution quirk either, an unrelated, pre-existing divergence not touched by this fix).

Fixes #1060

## Test plan

- [x] New regression tests: `test_yq_tostring_special_floats_use_yaml_spelling_1060`, `test_yq_at_csv_special_floats_use_yaml_spelling_1060` (verified byte-for-byte against pinned real yq v4.53.3), `test_yq_json_output_special_floats_still_null_after_1060` (confirms the separate JSON-output path is untouched).
- [x] `test_jq_tostring_special_floats_unaffected_by_yq_mode_fix_1060` confirms jq mode's own output is unchanged.
- [x] `cargo test --features cli,simd,regex,serde` (incl. golden/path-consistency suites) — 659 passed, 0 failed.

All verification steps run sequentially (not concurrently) to avoid a build-artifact race that caused a false-green local result on a previous PR this session (#1057).

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -5857,8 +5857,7 @@ fn builtin_tostring<W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         OwnedValue::Bool(true) => "true".to_string(),
         OwnedValue::Bool(false) => "false".to_string(),
         OwnedValue::Int(n) => format!("{n}"),
-        OwnedValue::Float(f) => format!("{f}"),
-        OwnedValue::NumberLiteral(..) => numeric_display_string::<S>(&owned),
+        OwnedValue::Float(_) | OwnedValue::NumberLiteral(..) => numeric_display_string::<S>(&owned),
         OwnedValue::Array(_) | OwnedValue::Object(_) => owned_value_to_json::<S>(&owned),
     };
     QueryResult::Owned(OwnedValue::String(s))
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -5068,17 +5068,34 @@ fn eval_string_interpolation<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
 /// comment) because `to_json` needs "null" instead. Text formats (`@text`,
 /// `@uri`, `@html`, `@sh`, and the CSV/TSV/DSV cell formatter that falls back
 /// to `owned_to_string`) want the same `"inf"`/`"-inf"`/`"NaN"` rendering a
-/// plain `Int`/`Float` already gets from `f64::to_string()` -- a
+/// plain `Int`/`Float` already gets from `f64::to_string()` in jq mode -- a
 /// `NumberLiteral` just needs that check made explicit: `format_number_jq_compat`
 /// now formats an overflowed literal's source text correctly (#930), but its
 /// jq-error-message-preview text (e.g. `1E+400`) still isn't the Rust-style
 /// `"inf"`/`"-inf"` these non-JSON text formats want, so the explicit check
 /// stays regardless.
+///
+/// yq mode instead wants YAML's own `.nan`/`.inf`/`-.inf` spelling (#1060) --
+/// this was previously unconditional (`f.to_string()`, jq's own spelling in
+/// both modes), the same class of gap #1030 fixed for finite literals.
+/// Real yq's own Go-based number model actually distinguishes a
+/// document-sourced non-finite scalar (`.inf`) from a genuinely *computed*
+/// one (`1/0` -> Go's own `+Inf`, confirmed live against the pinned oracle)
+/// -- but succinctly's `OwnedValue` can't make that distinction once a
+/// document-sourced `.inf`/`.nan` degrades to a plain `Float`: YAML's
+/// `number_literal()` override never captures these spellings as a
+/// `NumberLiteral` in the first place (`is_preservable_float_literal`'s
+/// JSON-syntax gate rejects them, since neither starts with a digit or
+/// `-digit`). So this always answers with the document-sourced spelling --
+/// correct for the common case (`.a | tostring` on an untouched `.inf`/
+/// `.nan` scalar), diverging from real yq's Go-style spelling only for a
+/// genuinely computed non-finite result, a narrower, rarer case left
+/// unaddressed here.
 pub(crate) fn numeric_display_string<S: EvalSemantics>(value: &OwnedValue) -> String {
     if let OwnedValue::NumberLiteral(repr, literal) = value {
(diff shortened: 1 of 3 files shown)


newhoggy:
Ran 8 parallel code-review finders. Found and fixed one real bug before merge; the rest were duplication/style observations, filed as a follow-up.

**Confirmed bug, fixed**: `builtin_tostring` (`src/jq/eval.rs`) had its own separate, hardcoded `OwnedValue::Float(f) => format!("{f}")` arm that the original commit never touched — so the `.nan`/`.inf` fix only applied to document-sourced scalars reaching `eval_generic.rs`'s own (already-correct) `Builtin::ToString` cursor arm, not to a value routed through the reindex bridge into `eval.rs`'s `builtin_tostring` (`nan`/`infinite`'s own jq-builtin evaluation, `--slurp`, `-i` with a non-allowlisted expression, `reduce`/`foreach`, ...). Confirmed live: `nan | tostring` gave `"NaN"` while `nan | @text` (documented in CLAUDE.md as identical to `tostring`) already gave `.nan` — a live self-inconsistency. Fixed by merging the `Float` arm into the existing `NumberLiteral` arm's call to `numeric_display_string::<S>`, which already handles both variants identically for the (unaffected) finite case. New regression test (`test_yq_tostring_computed_nan_uses_yaml_spelling_1060`) specifically targets the reindex-bridge path the original test missed.

**Filed as follow-up** (#1064, not blocking this PR): the yq-mode NaN/Infinity-spelling decision is now independently reimplemented in several places (`nonfinite_display_string` here, plus similar logic in `props_value_to_string`/`owned_to_yaml_at_depth` in the same file, and more in `stream.rs`/`yq_runner.rs`/`json/light.rs`) — the "duplicated predicates diverge silently" pattern CLAUDE.md's own #106 lesson names. Consolidating existing, already-correct call sites is a pure refactor with no behavior change, so it's scoped as its own follow-up rather than expanded into this fix.

Two lower-severity style findings (internal duplication of the `is_nan()||is_infinite()` check within `numeric_display_string` itself, and a small heap allocation for compile-time-constant yq spelling strings) were judged not worth blocking on given #1060's own "narrower, rarer case" scope.

Re-verified everything sequentially after both the fix and a rebase onto main (which advanced 7 commits, #1035's filter-literal fidelity work, unrelated area, clean rebase): full build, `cargo test --features cli,simd,regex,serde` (660 passed incl. golden/path-consistency), all 3 clippy configs, fmt, no_std check, doc build — all green.

Issue #1041, opened 2026-08-16
yq: join(s) doesn't match real yq's semantics (elements, separator, objects)


newhoggy:
## Summary

Split out of #1003/PR #1036's code review. `builtin_join` is shared between `jq` and `yq`
mode with no `S`-gated branch. #1003's fix made `join`'s jq-mode error wording exactly
match real jq — but real yq's `join` (v4.53.3) diverges from jq's semantics in multiple
confirmed ways, none of which succinctly currently implements correctly for `yq` mode.
PR #1036 deliberately kept the **pre-#1003** `join_parts`-based behavior for `YqSemantics`
(so #1003's fix doesn't newly regress yq — see `builtin_join`'s doc comment), but that
pre-existing behavior was never yq-accurate either. This issue is to actually implement
real yq-compatible `join` semantics.

## Confirmed divergences (live-verified against pinned yq v4.53.3)

1. **Non-scalar elements**: real yq silently treats an array/object element as an
   empty-string part instead of erroring.
   ```
   $ yq '[[1,2], "a"] | join(",")'
   ,a
   $ succinctly yq '[[1,2], "a"] | join(",")'
   [1,2],a
   ```
2. **Non-string separator**: real yq implicitly stringifies the separator instead of
   erroring.
   ```
   $ yq '[1, 2] | join(1)'
   112
   $ succinctly yq '[1, 2] | join(1)'
   Error: expected string, got non-string
   ```
3. **`null` separator**: real yq renders it as the literal text `"null"`, inserted as a
   real separator (not a no-op).
   ```
   $ yq '["a","b"] | join(null)'
   anullb
   $ succinctly yq '["a","b"] | join(null)'
   Error: expected string, got non-string
   ```
4. **Object input**: real yq rejects it outright.
   ```
   $ yq '{"a":"x","b":"y"} | join(",")'
   Error: cannot join with !!map, can only join arrays of scalars
   $ succinctly yq '{"a":"x","b":"y"} | join(",")'
   x,y
   ```

## Suggested approach

Add a `join`-specific `S`-gated branch (matching this codebase's existing pattern of
narrow named `EvalSemantics` consts like `ARRAY_MULTIPLY_MERGES`/`NULL_MERGES_AS_EMPTY`)
implementing yq's actual algorithm: reject objects with yq's exact error wording,
implicitly stringify a non-string separator instead of erroring, render `null` as literal
text, and treat non-scalar elements as empty-string parts. Needs its own live-verification
[...]

Issue #1064, opened 2026-08-16
jq: NaN/Infinity-to-YAML-spelling logic duplicated across eval.rs/stream.rs/yq_runner.rs/json/light.rs


newhoggy:
**Severity: Low**

## Summary

Code review on #1060 (PR #1063) found that the yq-mode `.nan`/`.inf`/`-.inf` spelling decision (`is_nan()`/`is_infinite()`/sign check → the three-way YAML-native string) is now independently reimplemented in multiple places rather than sharing one definition:

- `nonfinite_display_string` (new, `src/jq/eval.rs`, added by #1060)
- `props_value_to_string` (`src/jq/eval.rs:~5665`)
- `owned_to_yaml_at_depth` (`src/jq/eval.rs:~5717`)
- Similar copies noted in `src/jq/stream.rs`, `src/bin/succinctly/yq_runner.rs`, `src/json/light.rs`

This is the exact "duplicated predicates diverge silently" pattern CLAUDE.md's own optimization log names for issue #106 (three copies of one predicate, one of them quadratic) — here, a future spelling change (case, sign handling, an edge case fix) applied to only one copy would leave the others silently stale.

## Suggested approach

Consolidate the NaN/Infinity → yq-spelling decision into one shared function (mirroring how `format_number_jq_compat`/`format_float_with_fraction`/`real_output_finite_literal` are each a single definition reused across call sites), and update each of the sites above to call it instead of reimplementing the `is_nan()`/`sign()`/`.nan`/`-.inf`/`.inf` logic locally.

## Where this was found

Surfaced during code review of #1060/PR #1063 (which added `nonfinite_display_string` for `numeric_display_string`'s own NaN/Infinity gap) — deliberately not expanded into that fix's own scope, since consolidating existing, already-correct call sites is a refactor with no behavior change, not a bug fix, and touches several files beyond that PR's diff.

newhoggy:
🔧 Claimed by agent `cadence` — branch `issue-1064-nan-infinity-yq-spelling-dup`

PR #1068, opened 2026-08-16
fix(jq): yq join() renders empty-object as {} and keeps NaN/Inf spelling

### LINKED ISSUE #1047: yq: join(s) still diverges on an empty-object separator and NaN/Infinity elements
## Summary

Surfaced during code review of #1041/PR #1044 (which fixed yq's `join` to match real yq's four main confirmed divergences from jq's algorithm). Two narrower edge cases still diverge, found after that fix landed — deliberately left out of that PR's scope rather than expanding it further.

## Confirmed divergences (live-verified against pinned yq v4.53.3)

**1. An empty-object (`{}`) separator renders as literal text, unlike a non-empty object or any array (which all become `""`):**
```
$ yq '[1,2] | join({})'
1{}2
$ succinctly yq '[1,2] | join({})'
12
```
Compare: `join({"a":1})`, `join([1,2])`, `join([])` all correctly give `"12"` in both implementations — the bug is specific to an *empty* map.

`yq_join_separator` (`src/jq/eval.rs`) currently treats `OwnedValue::Object(_)` uniformly (always `""`, regardless of whether it's empty) — needs to special-case an empty object as its own case, distinct from array and non-empty-object handling.

**2. A NaN/Infinity element or separator loses its YAML literal spelling, rendering as JSON's `"null"` instead:**
```
$ printf -- '- .nan\n- 2\n' | yq 'join(",")'
.nan,2
$ printf -- '- .nan\n- 2\n' | succinctly yq 'join(",")'
null,2
```
Same divergence for `.inf`/`-.inf`, and for a NaN/Infinity *separator* (not just element).

Root cause: `yq_join_element_part`/`yq_join_separator`'s shared `other => other.to_json()` fallback arm inherits `to_json()`'s existing NaN→`"null"` rendering (this part is *not* new — it's the same behavior `join_step`'s jq-mode path and pre-#1041's `join_parts` both already had, and is *correct* for jq mode, where `nan | tostring` really does give `"null"`). yq's own scalar spelling for these values (`.nan`/`.inf`/`-.inf`) is already implemented elsewhere in this file (`src/jq/eval.rs` ~5596-5660, inline in at least two places) but not reused here.

## Suggested fix

- For the empty-object case: add a dedicated `OwnedValue::Object(m) if m.is_empty() => "{}".to_string()` (or similar) arm to `yq_join_separator`, ahead of the general `Object(_) => String::new()` arm. Needs verification of whether the same distinction applies to an empty *array* separator (`join([])`) — not yet checked.
- For NaN/Infinity: extract the existing inline `.nan`/`.inf`/`-.inf` spelling logic (`src/jq/eval.rs` ~5596-5660) into a shared helper, and route `yq_join_element_part`/`yq_join_separator`'s `other` arm through it instead of `to_json()` for `Float`/`NumberLiteral(Float)` specifically. Given the same inline logic already appears at least twice, extracting it would also address a pre-existing minor duplication independent of `join`.

## Scope note

Not blocking #1041/PR #1044 — that PR's own four named divergences (array-only input, non-scalar-element-to-empty-string, non-string-separator-stringification, null-separator-as-literal-text) are all correctly fixed and oracle-verified; these are two additional, narrower edge cases found during that PR's code review, out of its original scope.

newhoggy:
## Summary

Two narrower divergences from real yq's `join()` found after #1041 (PR #1044), tracked in #1047:

1. **Empty-object special case**: `[1,2] | join({})` gives `"1{}2"` in real yq — an *empty* object element or separator renders as literal `{}` text, unlike a non-empty object or any array (empty or not), which all become the empty string.

   The tricky part: `yq_join_element_part`'s element goes through `to_owned_key_shape` first (a deliberate optimization from #1044's own review, avoiding a deep-copy of a container whose contents the algorithm doesn't need) — and that helper *always* produces an empty `IndexMap` for any object, regardless of its real field count. So an `.is_empty()` check on the already-collapsed value can't distinguish "genuinely empty" from "non-empty, contents unread" — confirmed live: my first attempt at this fix wrongly rendered `{"a":1}` as `{}` too. Fixed by checking emptiness on the live `StandardJson::Object` cursor (`JsonFields::is_empty()`, O(1)) *before* the collapse, at the call site in `builtin_join`. `yq_join_separator`'s own `sep` value goes through `result_to_owned` (the full materializer) instead, so it needed no such cursor-level check — a plain `.is_empty()` guard works there.

2. **NaN/Infinity spelling**: a document-sourced `.nan`/`.inf`/`-.inf` element or separator loses its YAML spelling, rendering as `to_json_yq()`'s RFC-8259 `"null"` substitution instead. Fixed by routing through `nonfinite_display_string` — the same helper #1060 (merged just before this PR) added for the identical gap in `tostring`/`@text`/etc.

Fixes #1047

## Test plan

- [x] `test_yq_join_empty_object_renders_as_literal_text_1047` — covers empty-object element, non-empty-object element (the specific case my first attempt got wrong), and empty-object separator.
- [x] `test_yq_join_special_float_element_and_separator_use_yaml_spelling_1047` — covers `.nan`/`.inf` element and `.nan` separator.
- [x] Full oracle sweep against pinned real yq v4.53.3 across empty/non-empty object and array, both as element and separator — all match.
- [x] jq mode confirmed unaffected (both succinctly and real jq error identically on `join` with an object input, since all changes are inside the `S::TAG == EvalTag::Yq` branch).
- [x] `cargo test --features cli,simd,regex,serde` (incl. golden/path-consistency suites) — 662 passed, 0 failed.

All verification run sequentially (no concurrent cargo invocations against the same worktree).

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -4102,7 +4102,28 @@ fn builtin_split<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
 /// site, not [`to_owned`]) so a container element's contents are never
 /// materialized at all -- they're discarded unread here regardless -- and a
 /// string element is moved rather than cloned a second time.
+/// yq's own `.nan`/`.inf`/`-.inf` spelling for a non-finite element or
+/// separator (#1047, matching #1060's identical fix for `tostring`/`@text`/
+/// etc.), shared by [`yq_join_element_part`] and [`yq_join_separator`]
+/// rather than hand-copied at each -- the #106 "duplicated predicates
+/// diverge silently" lesson in `CLAUDE.md`, caught in this PR's own code
+/// review. `None` for anything else, so the caller's own catch-all
+/// (`other.to_json_yq()`) handles every other variant unchanged.
+fn yq_join_nonfinite_part(value: &OwnedValue) -> Option<String> {
+    match value {
+        OwnedValue::Float(f) | OwnedValue::NumberLiteral(NumberRepr::Float(f), _)
+            if f.is_nan() || f.is_infinite() =>
+        {
+            Some(nonfinite_display_string::<YqSemantics>(*f))
+        }
+        _ => None,
+    }
+}
+
 fn yq_join_element_part(elem: OwnedValue) -> String {
+    if let Some(s) = yq_join_nonfinite_part(&elem) {
+        return s;
+    }
     match elem {
         OwnedValue::String(s) => s,
         // NOT a `OwnedValue::Object(m) if m.is_empty()` guard here (#1047
@@ -4114,17 +4135,6 @@ fn yq_join_element_part(elem: OwnedValue) -> String {
         // special case is instead handled at the call site, on the live
         // cursor, before that collapse ever runs.
         OwnedValue::Null | OwnedValue::Array(_) | OwnedValue::Object(_) => String::new(),
-        // YAML's own `.nan`/`.inf`/`-.inf` spelling (#1047, matching #1060's
-        // identical fix for `tostring`/`@text`/etc.), not `to_json_yq()`'s
-        // RFC-8259 `"null"` substitution -- confirmed live,
-        // `[.nan, 2] | join(",")` on a document-sourced `.nan` element is
-        // `".nan,2"` in real yq, not `"null,2"`.
-        OwnedValue::Float(f) if f.is_nan() || f.is_infinite() => {
-            nonfinite_display_string::<YqSemantics>(f)
-        }
-        OwnedValue::NumberLiteral(NumberRepr::Float(f), _) if f.is_nan() || f.is_infinite() => {
-            nonfinite_display_string::<YqSemantics>(f)
-        }
         // `to_json_yq()`, not `to_json()` (#1030 code review): this function
         // is only ever reached from `builtin_join`'s `S::TAG == EvalTag::Yq`
         // branch, so a scientific-notation literal here must echo verbatim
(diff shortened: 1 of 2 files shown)


newhoggy:
Ran 8 parallel code-review finders. No correctness bugs found — every finder independently traced the empty-object/NaN-Infinity handling through `to_owned_key_shape`'s collapse behavior, `result_to_owned`'s full materialization, and every call site, and confirmed the fix is sound (including edge cases beyond the new tests: merge keys, aliases, nested empty objects, computed non-finite values via the owned-pipe reindex path).

**One real finding, fixed**: three independent finders converged on the same duplication — the NaN/Infinity match-arm pair (`Float`/`NumberLiteral(Float)` with an `is_nan()||is_infinite()` guard) was hand-copied identically into both `yq_join_element_part` and `yq_join_separator`, with one finder explicitly citing CLAUDE.md's own `#106` "duplicated predicates diverge silently" lesson by name (the same predicate also exists in `numeric_display_string`). Extracted `yq_join_nonfinite_part`, called once at the top of each function. No behavior change — re-verified against the full oracle sweep and the existing test suite.

Two finders also noted the pre-existing `"{}".to_string()` literal now appears twice (the separator's match arm and `builtin_join`'s inline element check) — judged not worth a shared constant given they're structurally different code paths (one operates on `OwnedValue`, the other on a live `StandardJson` cursor before collapse), and the broader "consolidate all yq NaN/Infinity spelling call sites" work is already tracked in #1064.

Local verification hit a known false-positive this session: the first full-suite run under heavy concurrent load (many other `grinding-backlog` sessions active) showed 17 spurious `jq_cli_tests` failures, all in tests completely unrelated to this diff (`test_sort_keys`, `test_slurpfile`, etc.). Re-running `jq_cli_tests` alone: 553/553 passed — a known CPU-contention flakiness pattern, not a real regression (confirmed and documented in project memory).

Re-verified everything after the fix, each test binary run in isolation to avoid the contention issue above: lib unit tests (2519 passed), `jq_cli_tests` (553 passed), `yq_cli_tests` (662 passed), `yq_golden_tests`/`yq_path_consistency_tests` (6 passed), all 3 clippy configs, fmt, no_std check, doc build — all green.

Issue #1069, opened 2026-08-16
yq: plain sub(re; s) replaces every match, not just the first (real yq diverges from jq's sub/gsub distinction)


newhoggy:
## Summary

Found while verifying #1052's replacement-value coercion fix. Real yq's own `sub`/`gsub` builtins appear to diverge from jq's match-count semantics in a way not previously documented: real yq's plain 2-arg `sub(re; s)` replaces **every** match in the input, not just the first the way jq's `sub` does (jq's `sub` = first match only, `gsub` = all matches — yq's `sub` seems to behave like jq's `gsub` unconditionally).

## Repro

```
$ jq -n '"aaa" | sub("a";"X")'
"Xaa"                              # jq: sub replaces only the first match

$ echo '"aaa"' | yq 'sub("a";"X")'
XXX                                # real yq: sub replaces every match

$ echo '"aaa"' | succinctly yq 'sub("a";"X")'
Xaa                                # succinctly currently matches jq's semantics here, not yq's
```

## Also observed, not yet understood

`sub(re; s; flags)` (3-arg form) with an explicit `"g"` flags string gave empty output in real yq during probing, which doesn't fit any hypothesis tried so far — worth re-verifying carefully (quoting/escaping issues in the probe are also possible) before drawing conclusions about the 3-arg form specifically.

## Scope note

Discovered during #1052 (which was scoped narrowly to replacement-*value* type coercion, not match-count semantics) — deliberately not investigated further or fixed there. This is a distinct, and probably more impactful, divergence: it affects every existing `sub(...)` call under yq mode with a pattern that can match more than once in the input, not just calls with a non-string replacement.

## Suggested approach

Needs careful live-probing against real yq (v4.53.3, the pinned golden version) across:
- Multiple non-overlapping matches (confirmed above)
- The 3-arg flags form, especially with `"g"`/without it, to understand what (if anything) yq's flags argument actually controls if plain `sub` already replaces everything
- Whether `gsub` even exists as valid yq filter syntax at all (a preliminary probe hit a lexer error on bare `gsub(...)`, worth double-checking with correct quoting before concluding yq's `gsub` name is unsupported)
- Capture-group interaction, since the replacement is a filter over `.` bound to captures per real jq's `sub` definition — confirm yq's captures-per-match semantics still hold when every match is processed, not just the first

[...]

Issue #1065, opened 2026-08-16
yq: slicing a non-array/string scalar with ? gives no output, but real yq gives []


newhoggy:
## Summary

Found while verifying #1048's fix. Real yq has its own special slicing semantics for a scalar target that succinctly doesn't replicate: `<scalar> | .[S:E]?` produces a literal empty array `[]` in real yq, but produces no output at all in succinctly (matching real *jq*'s behavior instead, which also gives no output for the equivalent).

## Repro

```
$ echo '5' | yq -o=json '.[0:1]?'
[]
$ echo '5' | succinctly yq -o json '.[0:1]?'
(no output)

$ echo null | jq -c '5 | .[0:1]?'
(no output)   # jq agrees with succinctly here — this is a yq-only divergence
```

## Scope note

This is unrelated to #1048 (which was about `eval_generic.rs` internally collapsing a
zero-result computed index/slice to `Many`/`ManyOwned(vec![])` instead of `None` — a
library-API-only bug with no CLI-observable difference, since both variants print "no
output" at the top level). This is a different, CLI-observable divergence: real yq
appears to treat a non-array/string target as an empty-array-like container for slicing
purposes (matching the null-as-empty-container pattern already documented elsewhere in
this project's yq merge-flag semantics), rather than suppressing the slice to zero
results via `?`.

## Suggested approach

Needs live-probing against real yq across a few target shapes (number, string, bool,
null, object) and slice-bound combinations before implementing, to confirm the exact
rule (is it *any* non-sliceable target, or something narrower?) — not yet investigated
further.

newhoggy:
🔧 Claimed by agent `cadence` — branch `issue-1065-yq-slice-scalar-empty-array`

Issue #1048, opened 2026-08-16
jq: eval_generic.rs's index/slice-expr collapses have the same missing 0 => None bug as #1043


newhoggy:
## Summary

Found during PR #1046's code review. `src/jq/eval_generic.rs` — the generic-document
(yq) counterpart to `src/jq/eval.rs`'s `eval_index_expr`/`eval_slice_expr` — has the
identical missing-`0 => None`-arm bug #1043 just fixed in `eval.rs`, in three places, all
untouched by that PR.

## Confirmed sites (current line numbers as of PR #1046's branch)

1. **`eval_index_expr`'s "keys empty, target has zero results after halt" collapse**
   (~`eval_generic.rs:2925`):
   ```rust
   None => match out.len() {
       1 => GenericResult::Owned(out.pop().expect("len checked")),
       _ => GenericResult::ManyOwned(out),
   },
   ```
   Same shape as `eval.rs`'s now-fixed `eval_index_expr` Owned arm. Trivial fix: swap to
   the existing `owned_vec_to_generic_result(out)` helper (`eval_generic.rs:950`, already
   documented as mirroring `eval::owned_vec_to_result`).

2. **`eval_index_expr`'s main "key outer, target inner" final collapse**
   (~`eval_generic.rs:2992-3002`):
   ```rust
   if any_owned {
       match owned.len() {
           1 => GenericResult::Owned(owned.pop().expect("len checked")),
           _ => GenericResult::ManyOwned(owned),
       }
   } else {
       match cursors.len() {
           1 => GenericResult::OneCursor(cursors.pop().expect("len checked")),
           _ => GenericResult::ManyCursor(cursors),
       }
   }
   ```
   The `owned` arm is a trivial swap to `owned_vec_to_generic_result`. The `cursors` arm
   has **no existing helper** — `owned_vec_to_generic_result` only covers `Vec<OwnedValue>`,
   not `Vec<V::Cursor>`. Needs a new `cursor_vec_to_generic_result`-style helper (or an
   inline `0 => None` arm) before this site can be fixed.

3. **`eval_slice_expr`'s final collapse** (~`eval_generic.rs:3107-3110`):
   ```rust
   match out.len() {
       1 => GenericResult::Owned(out.pop().expect("len checked")),
       _ => GenericResult::ManyOwned(out),
   }
   ```
   Mirrors `eval.rs`'s now-fixed `eval_slice_expr` exactly. Trivial fix: swap to
[...]

Issue #1034, opened 2026-08-16
jq: sub/gsub replacement-function error wording doesn't match jq's accumulator-state-derived messages


newhoggy:
## Summary

Split out of #1003 (which is now scoped to `join` only — see the PR fixing #1003). Real
jq's error wording for a type mismatch inside `sub`'s replacement-function evaluation
comes from *internal accumulator/pipeline state* at the point jq's own replacement-
pipeline definition fails, not a bespoke "expected X, got Y" message.

## Repro (verified against pinned jq 1.7.1 oracle, and against current `main`)

```
$ echo '"a"' | jq -c 'try sub("a"; 5) catch .'
"string (\"\") and number (5) cannot be added"
$ echo '"a"' | succinctly jq -c 'try sub("a"; 5) catch .'
"expected string, got replacement"
```

## Why this is a separate issue from #1003's `join` fix

`join`'s fix (this issue's sibling) worked because jq's own `join` definition is a simple
`reduce`-based fold — reimplementing it directly as a `reduce`-shaped Rust loop over
`arith_add` naturally reproduced jq's exact wording as a side effect.

`sub`/`gsub` are structurally different: they're regex-based, evaluate a replacement
*filter expression* per match (with capture groups bound as `.` context), and build the
result string by concatenating literal spans with per-match replacement output — not a
single accumulator folded via `+`. See `builtin_sub`/`builtin_sub_with_flags` /
`builtin_gsub_with_flags` in `src/jq/eval.rs` (~line 7903 onward). Reproducing jq's exact
wording here likely needs tracing jq's own C implementation or a from-scratch
reverse-engineering pass (same empirical-probing approach used for `join` in #1003) to
find the real definition/accumulator shape `sub` uses internally — it is not a direct port
of `join`'s fix.

## Suggested approach

Needs its own investigation into where succinctly's `sub`/`gsub` replacement-evaluation
diverges structurally from jq's own implementation before a fix strategy can be chosen.
Per this session's own experience on #1003/#989/#1030/#1010: don't trust an issue's
suggested fix direction (including this one) without live-verifying it against the pinned
jq oracle first.

newhoggy:
🔧 Claimed by agent `cadence` — branch `issue-1034-sub-gsub-accumulator-wording`

PR #1070, opened 2026-08-16
fix(jq): yq's sub/gsub coerces a non-string replacement instead of erroring

### LINKED ISSUE #1052: yq: sub/gsub replacement coercion doesn't match real yq's semantics
## Summary

Split out of #1034/PR #1050's code review. `combine_sub_gap` (the shared helper
`stitch_replacements_evaluated`/`sub_with_resolved_pattern` use to combine a match's
gap text with its replacement) is jq-specific for non-string replacements — it deliberately
keeps yq mode on its pre-#1034 "replacement must already be a string" check rather than
adopting jq's `+`-based accumulator rules, since real yq's own `sub`/`gsub` coercion
diverges from jq's in a different way that #1034 didn't attempt to implement.

## Confirmed divergence (live-verified against pinned yq v4.53.3)

```
$ yq 'sub("a"; 5)' <<< '"abc"'
5bc
$ succinctly yq 'sub("a"; 5)' <<< '"abc"'
Error: expected string, got replacement

$ yq 'sub("a"; null)' <<< '"abc"'
nullbc
$ succinctly yq 'sub("a"; null)' <<< '"abc"'
Error: expected string, got replacement
```

Real yq stringifies a non-string replacement and concatenates it (matching yq's general
looser-coercion philosophy documented elsewhere in this codebase, e.g.
`NULL_MERGES_AS_EMPTY`/`ARRAY_MULTIPLY_MERGES`), rather than erroring. succinctly's yq
mode currently errors either way (both pre- and post-#1034 — this is a pre-existing gap,
not a regression #1034 introduced; #1034 was careful to keep yq's behavior unchanged
rather than accidentally moving it further from real yq under jq's `+`-based rules).

## Suggested approach

Add yq-specific coercion inside `combine_sub_gap` (gated the same way `builtin_join`
gates its own jq-vs-yq algorithm split, `S::TAG == EvalTag::Yq` — see #1041/PR #1044 for
the precedent) that stringifies a non-string replacement (matching yq's own `tostring`-ish
coercion) instead of requiring it to already be a `String`. Needs its own live-verification
pass across the full case matrix (numbers, booleans, arrays, objects, `null`) against the
pinned yq oracle, plus dedicated `yq_query!`-based test coverage — none of which currently
exists for `sub`/`gsub` under yq mode.

newhoggy:
## Summary

Split from #1034/#1050's review: `combine_sub_gap` deliberately kept yq mode on its pre-#1034 "replacement must already be a string" check, since real yq's own coercion rules diverge from jq's `+`-based accumulator and #1034 was scoped to jq-mode wording only.

Live-verified against yq v4.53.3:
- A scalar replacement (number, float, bool, null) stringifies and concatenates instead of erroring — `sub("a";5)` on `"abc"` is `"5bc"`.
- An array/object replacement is the one surprising exception: it contributes an **empty** string, not a JSON representation the way `tostring()` gives for the same value (`sub("a";[1,2])` is `"bc"`, not `"[1, 2]bc"`) — presumably an artifact of yq's underlying go-yaml node model, where only scalar nodes carry a raw text `Value` this call site reads directly.

## Follow-ups filed

Two significant, unrelated divergences surfaced while probing the oracle for this fix:
- #1065 — slicing a scalar with `?` gives `[]` in real yq but no output in succinctly (found during #1048, not this PR).
- #1069 — real yq's plain 2-arg `sub(re; s)` appears to replace **every** match, not just the first the way jq's `sub` does. This is a bigger, unrelated divergence in match-count semantics (this PR only touches replacement-*value* coercion) — deliberately not investigated or fixed here.

Fixes #1052

## Test plan

- [x] `cargo test --features cli,simd,regex,serde --no-fail-fast` — all green (full integration suite, not just `--lib`, per the #1035 lesson)
- [x] Verified against real yq v4.53.3 for the full replacement-type matrix (number/float/null/bool/array/object) on both `sub` and `gsub`

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -30157,6 +30157,19 @@ mod tests {
                 assert_eq!(s, "bc");
             }
         );
+        yq_query!(br#""abc""#, r#"sub("a"; [])"#,
+            QueryResult::Owned(OwnedValue::String(s)) => {
+                assert_eq!(s, "bc");
+            }
+        );
+        // An *empty* object is the one exception to the container-is-empty
+        // rule: it stringifies to the literal text "{}", not "" the way an
+        // empty array or a non-empty object both do (confirmed live).
+        yq_query!(br#""abc""#, r#"sub("a"; {})"#,
+            QueryResult::Owned(OwnedValue::String(s)) => {
+                assert_eq!(s, "{}bc");
+            }
+        );
         // Never errors any more, so isvalid reports true (was false
         // pre-#1052).
         yq_query!(br#""abc""#, r#"isvalid(sub("a"; 5))"#,
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -8053,14 +8053,17 @@ fn stitch_replacements_evaluated<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
 /// unreachable-arm coupling a fully generic `+` call would add to this
 /// per-match hot path (#1034 review). Anything else falls back to
 /// jq-mode's real `+`-based accumulator (`arith_add`, `null` included via
-/// its own identity rule) or yq-mode's pre-#1034 "must already be a
-/// string" check -- **not** real yq's own `sub`/`gsub` coercion (confirmed
-/// live against yq v4.53.3: a non-string replacement stringifies and
-/// concatenates instead of erroring, e.g. `sub("a";5)` on `"abc"` is
-/// `"5bc"`); #1034 only fixed jq-mode's wording, so yq keeps its prior
-/// (still not yq-accurate, but not newly regressed) behavior here rather
-/// than silently changing under jq's `+`-based rules -- real yq-compatible
-/// `sub`/`gsub` coercion is its own follow-up (#1052).
+/// its own identity rule) or, in yq mode, real yq's own looser coercion
+/// (#1052): a scalar replacement stringifies and concatenates instead of
+/// erroring (`sub("a";5)` on `"abc"` is `"5bc"`, confirmed live against
+/// yq v4.53.3 for numbers/floats/bools/`null`), while an array/object
+/// replacement contributes an *empty* string rather than a JSON
+/// representation (`sub("a";[1,2])` on `"abc"` is `"bc"`, not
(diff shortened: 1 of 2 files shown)


newhoggy:
Code review (10 finder angles) found one real oracle-verified bug in the initial commit, plus several quality findings — all addressed, pushed:

**Fixed:**
- An *empty* object replacement (`sub("a"; {})`) is the one exception to the "container stringifies to empty" rule — real yq renders it as literal `"{}"` text, not `""` the way an empty array or non-empty object both do. Confirmed and fixed independently before implementing (mirrors this session's discipline of live-verifying before trusting a guess).
- Deduplicated `combine_sub_gap`'s two near-identical concat-building blocks into one, and made the yq-mode match exhaustive over `OwnedValue`'s variants instead of a wildcard.
- Stale doc-comment sentence ("both branches") left over from before the fix removed one of the two error paths.
- Added a CLI-level integration test — every existing coercion test called `eval()` directly, which can't observe the actual exit-code flip (non-zero → zero) a script piping the real binary would see.

**Filed as follow-ups** (real, but out of scope for this narrow fix):
- #1071 — a *computed* (not document-sourced) NaN/Infinity replacement uses the wrong spelling (`.inf` instead of real yq's `+Inf`). Confirmed this also affects `join` (#1047, merged during this PR's review) via the same shared helper — not unique to sub/gsub.
- #1072 — the "container stringifies to empty" quirk is now independently implemented in 3 places (join's 2 helpers + this fix's new one), with real disagreements between them on `null` handling and numeric formatting that need resolving before a mechanical merge.
- #1073 — `@sh` on an array doesn't error under yq mode the way real yq does (found while checking whether other builtins share this coercion rule — they don't, several error instead).
- #1069 — a much bigger, unrelated finding from the original oracle probing: real yq's plain `sub(re;s)` appears to replace *every* match, not just the first the way jq's `sub` does.

Rebased once mid-review onto #1047 (unrelated, but touches adjacent yq-number-formatting code) — clean, zero conflicts. All local verification re-run clean after the fixes: full `--no-fail-fast` test suite, 3x clippy matrix, fmt, no_std check, and a fresh oracle sweep against real yq v4.53.3 including the empty-object/empty-array edge cases.

Issue #954, opened 2026-08-15
yq: JSON-unsafe-but-YAML-legal float spellings (+2.0, 1., 007e2) still lose their decimal point (#918's residual scope)


newhoggy:
**Severity: Low**

## Summary

#918 fixed the "integer-valued float scalar loses its decimal point" bug for the common case (plain decimal spelling like `2.0`, and the explicitly-tagged `!!float 2.0`), by preserving the scalar's literal source text — but only when that text is also valid JSON number syntax. A few YAML 1.2 core-schema-legal float spellings aren't valid JSON syntax and so are deliberately excluded from that fix, leaving #918's original symptom open for them.

## Repro

```
$ printf 'a: +2.0\n' | succinctly yq '[.a]' -o json -I0
[2]
$ printf 'a: +2.0\n' | yq '[.a]' -o json -I0   # pinned v4.53.3
[2.0]

$ printf 'a: 1.\n' | succinctly yq '[.a]' -o json -I0
[1]
$ printf 'a: 1.\n' | yq '[.a]' -o json -I0
[1.0]

$ printf 'a: 007e2\n' | succinctly yq '[.a]' -o json -I0
[700]
$ printf 'a: 007e2\n' | yq '[.a]' -o json -I0
[700.0]
```

Also affects an explicit `!!float` tag on int-shaped text (e.g. `!!float 5` → `[5]`, real yq → `[5.0]`) — there the source text has no literal `.` at all to preserve, since `resolve_tagged` widens the int under the tag rather than the text itself being float-shaped.

## Root cause

`is_preservable_float_literal` (`src/yaml/scalar.rs`) requires the literal to already be valid JSON number syntax — no leading `+`, no leading-zero-then-more-digits, and (per #918's own review) at least one digit after a trailing `.`. `+2.0`, `007e2`, and `1.` all fail one of those, so `number_literal()` declines and they fall through to the pre-existing `as_f64()` → bare `OwnedValue::Float` → `Display` path, which drops the decimal point on a whole number exactly as #918 originally described.

## Why this is separate from #918

#918's fix is deliberately scoped to the JSON-safe subset: echoing JSON-unsafe text into the internal JSON-reindex bridge used by builtins like `tag` was demonstrated (during #918's own review) to silently misclassify values as parse errors rather than numbers, so widening the accepted grammar isn't a safe drop-in extension of that same mechanism — it needs its own design (e.g. normalizing the text to JSON-safe form before storing it, rather than echoing it verbatim).

[...]

PR #1074, opened 2026-08-16
refactor(jq): share yq's NaN/Infinity YAML spelling across 5 sites

### LINKED ISSUE #1064: jq: NaN/Infinity-to-YAML-spelling logic duplicated across eval.rs/stream.rs/yq_runner.rs/json/light.rs

newhoggy:
## Summary

The "is_nan → `.nan`, is_infinite && positive → `.inf`, else → `-.inf`" yq YAML-native spelling decision was independently reimplemented in 5 places beyond its canonical definition (`nonfinite_display_string`, added by #1060): `eval.rs`'s `props_value_to_string` (`@props`) and `owned_to_yaml_at_depth` (`@yaml`), `stream.rs`'s `stream_owned_value_yaml_at_depth` (the default `yq` YAML streamer), `json/light.rs`'s `stream_json_as_yaml` (the M2.5 JSON-input YAML-output fast path), and `yq_runner.rs`'s `emit_yaml_value_at_depth` (the CLI's own non-streaming YAML emitter, a separate crate) — CLAUDE.md's #106 "duplicated predicates diverge silently" lesson, previously applied to `join`'s own two internal copies of this same decision (#1047).

All 5 sites now route through `nonfinite_display_string::<YqSemantics>`, which required making it `pub` (not just `pub(crate)`): `yq_runner.rs` is a separate binary crate depending on this one as an external dependency, so `pub(crate)` alone can't reach it. Re-exported through `jq/mod.rs`'s existing `pub use eval::{...}` list, the same mechanism already used for `eval`, `substitute_vars`, and friends.

**Pure refactor, no behavior change.** Verified two ways:
- The full existing test suite (golden fixtures included) passes unchanged.
- Each site's prior inline logic used a stylistically different but behaviorally identical sign test (`f > 0.0` vs `f.is_sign_negative()`, or an `Option<f64>` comparison) — none of which can actually diverge, since none of these sites is ever reached with a signed-zero float (all gated on `is_infinite()` first, and infinity is never `±0.0`).

## Test coverage added

Several sites had thin or missing direct coverage before this PR:
- `@props`/`@yaml` had no NaN/Infinity tests at all.
- `emit_yaml_value`'s existing test only covered the `NumberLiteral` arm, with the plain `Float` arm's parity merely *claimed* in a comment, never exercised.
- `json/light.rs`'s `stream_json_as_yaml` (the M2.5 fast path) has no CLI-observable trigger to verify against directly — covered at the unit level via an exponent overflowing to `f64::INFINITY` (JSON has no direct Infinity/NaN literal).

Fixes #1064

## Test plan

- [x] `cargo test --features cli,simd,regex,serde --no-fail-fast` — all green (full integration suite)
- [x] Live-verified `@props`, `@yaml`, and plain `yq` YAML output against document-sourced `.nan`/`.inf`/`-.inf` input

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -3905,7 +3905,6 @@ pub fn run_yq(args: YqCommand) -> Result<i32> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use succinctly::jq::NumberRepr;
 
     /// `Builtin::FirstStream`/`LastStream` (the "Phase 13" builtin-table
     /// spelling of `first(expr)`/`last(expr)`, distinct from the
@@ -4004,6 +4003,39 @@ mod tests {
             OwnedValue::NumberLiteral(NumberRepr::Float(f64::NEG_INFINITY), "-1e400".into());
         assert_eq!(
             emit_yaml_value(&neg_inf, &CommentTree::empty(), &config, "", false),
+            "-.inf"
+        );
+
+        // #1064: the plain `Float` arm (this test's own doc comment above
+        // claims parity with it, but never actually exercised it).
+        assert_eq!(
+            emit_yaml_value(
+                &OwnedValue::Float(f64::NAN),
+                &CommentTree::empty(),
+                &config,
+                "",
+                false
+            ),
+            ".nan"
+        );
+        assert_eq!(
+            emit_yaml_value(
+                &OwnedValue::Float(f64::INFINITY),
+                &CommentTree::empty(),
+                &config,
+                "",
+                false
+            ),
+            ".inf"
+        );
+        assert_eq!(
+            emit_yaml_value(
+                &OwnedValue::Float(f64::NEG_INFINITY),
+                &CommentTree::empty(),
+                &config,
+                "",
+                false
+            ),
             "-.inf"
         );
     }
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -15,8 +15,8 @@ use succinctly::jq::eval_generic::{
     GenericResult,
 };
 use succinctly::jq::{
-    self, assert_value_tree_depth, sync_aliased_paths, Builtin, EvalError, Expr, OwnedValue,
-    QueryResult, YqSemantics,
+    self, assert_value_tree_depth, nonfinite_display_string, sync_aliased_paths, Builtin,
+    EvalError, Expr, NumberRepr, OwnedValue, QueryResult, YqSemantics,
 };
 use succinctly::json::JsonIndex;
 use succinctly::yaml::{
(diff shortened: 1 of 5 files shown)


newhoggy:
Code review (10 finder angles) confirmed the consolidation is behaviorally correct (zero divergence across all 5 sites) and found no correctness bugs — but surfaced several real findings, all addressed or filed:

**Fixed:**
- `nonfinite_display_string` now returns `&'static str` instead of `String` (independently confirmed by 2 angles: every branch is one of a fixed 6-string set, since Rust's own `f64::Display` always normalizes any NaN/Infinity to exactly `"NaN"`/`"inf"`/`"-inf"`) — eliminates a heap allocation at the two streaming call sites that used to write a `&'static str` literal directly.
- Added the `# Examples` doc block `docs/STYLE_GUIDE.md` requires for a newly-`pub` function (this file's other `pub fn`s already carry one — a real, quotable gap).
- Cleaned up `json/light.rs`'s fully-qualified `crate::jq::`-prefixed call into a proper import.
- Tightened my own new `stream_json_as_yaml` test's doc comment to state plainly it's pinning internal-consistency only, not oracle-verified (real yq hard-errors on this exact input rather than producing comparable output) — a fair concern from the language-pitfall angle about a future reader assuming oracle parity.

**Filed as follow-ups** (both need careful live-oracle scoping before touching, not same-PR drive-bys):
- #1075 — `nonfinite_display_string`'s jq-mode branch uses Rust's raw `f64::Display` ("NaN"/"inf") instead of real jq's own convention (`null` for NaN, `DBL_MAX`'s decimal expansion for infinity) — confirmed live, pre-existing (this PR only changed the function's visibility/canonicalness, not this branch's logic).
- #1076 — the adjacent finite-float arms in these same functions don't reuse `format_float_yq`/`format_float_with_fraction` the way other YAML emitters do. **Important:** I live-verified the review's own suggested "obvious" fix (route through those helpers) would likely be wrong — real yq's own top-level output and `@yaml` output *agree with each other* for a computed whole float (`4.0/2` → `2`, no `.0`, in both), while succinctly's currently *disagree* (`2.0` top-level vs `2` via `@yaml`) — so the "buggy" `@yaml` behavior is incidentally correct for that case, and blindly adding `.0` would regress it. Documented this contradiction explicitly in the issue so whoever picks it up doesn't fall into the same trap; it connects to the already-open #954/#1054 "computed vs document-sourced" formatting gap.

All local verification re-run clean after the fixes: full `--no-fail-fast` test suite (40 tests incl. the new doctest), 3x clippy matrix, fmt, no_std check.

newhoggy:
Final commit adds doc comments explaining why `eval.rs:5769`/`5818` (the plain finite-`Float` arms in `props_value_to_string`/`owned_to_yaml_at_depth`) stay uncovered: every real filter reaches these functions via `eval_format`'s `to_owned(&value)`, which always reconstructs a `NumberLiteral` for finite RFC-8259 number text — a bare `Float` only ever arrives as NaN/Infinity, already caught by the preceding guarded arm. Confirmed via `cargo llvm-cov --html` plus direct entry-point probes (not assumed). The arm can't be deleted (Rust requires match exhaustiveness over `Float`), so it's documented rather than chased with an artificial test. Patch coverage holds at 98.15% (106/108) with this as the only, now-explained, gap. All CI green — merging.

Issue #1080, opened 2026-08-16
yq: chained deferred anchors (`a: &x # c1` immediately followed by `b: &y # c2`) render the second anchor on its own line


newhoggy:
**Severity: Low**

## Summary

Found while implementing #784, and left out of that PR's scope — an even
narrower, compound edge case than #1077/#1078/#1079, likely entangled with
#1077's null-anchor-rendering bug but not fully explained by it.

Two mapping keys in a row, each with its own anchor whose value is deferred
to a following line and each with its own trailing comment
(`a: &x # c1` immediately followed by `b: &y # c2`, where `a`'s deferred
value turns out null and floats `c1` to `b`'s line, while `b`'s own
deferred value is a nested mapping):

```
$ printf 'a: &x # c1\nb: &y # c2\n  c: 1\n' | yq '.'
a: &x
b: # c1
&y
  c: 1 # c2
$ printf 'a: &x # c1\nb: &y # c2\n  c: 1\n' | succinctly yq '.'
a: &x ""
b: # c1 &y
  c: 1 # c2
```

Note real yq's own output is already unusual here — `&y` renders on its
*own*, un-indented standalone line, separate from both `b:`'s line and
`c: 1`'s line. succinctly's output instead puts `&y` at the end of `b:`'s
line (same ordering bug shape as #1078), and (per #1077) shows `&x ""`
instead of bare `&x`.

## Scope note

This is a compound of #1077 (null-anchor rendering) and something like
#1078's anchor/comment ordering issue, but real yq's own rendering (`&y` on
its own bare line) suggests its internal comment/anchor model may be more
structurally different from succinctly's `line_comments: BTreeMap<bp_pos,
(u32,u32)>` than a straightforward fix to either of those two issues would
account for. Recommend investigating *after* #1077 and #1078 land, in case
fixing those simplifies or resolves this case as a side effect — don't
implement in isolation.

newhoggy:
## Implementation plan (2026-08-21) — re-measured; #1077 and #1078 have both landed

The issue says "recommend investigating *after* #1077 and #1078 land, in
case fixing those simplifies or resolves this case as a side effect". Both
are now **CLOSED**, and the picture has changed materially — the issue's
recorded succinctly output is stale.

### Current state (measured on `main`, 2026-08-21)

```
$ printf 'a: &x # c1\nb: &y # c2\n  c: 1\n' | yq '.'
a: &x
b: # c1
&y
  c: 1 # c2

[...]

Issue #1079, opened 2026-08-16
yq: a sequence item's own trailing comment (no anchor) is dropped when its value is deferred


newhoggy:
**Severity: Low**

## Summary

Found while implementing #784 (deliberately out of that issue's scope — no
anchor is involved here at all, so it isn't #784's bug pattern, but a
sibling gap in the same neighborhood: #765 fixed this exact shape for
*mapping keys*, but never for *sequence items*).

A bare sequence item (`- `) with a trailing comment and no scalar on its own
line, whose value is deferred to a following line, drops the comment
entirely — worse, when that deferred value turns out null, succinctly's
output diverges from real yq in an additional way (null renders as the
literal string `null` instead of an empty scalar, *and* the comment
relocates to a bare, un-indented standalone line in real yq's own output,
which succinctly doesn't attempt to replicate).

```
$ printf -- '- # comment\n  b: 1\n' | yq '.'
- # comment
  b: 1
$ printf -- '- # comment\n  b: 1\n' | succinctly yq '.'
- b: 1
```

```
$ printf -- '- # comment\n- 2\n' | yq '.'
-
# comment
- 2
$ printf -- '- # comment\n- 2\n' | succinctly yq '.'
- null
- 2
```

## Root cause (partial)

`parse_sequence_item_inner`'s `if self.at_line_end() { ... }` branch (`src/yaml/parser.rs`)
has no `maybe_capture_line_comment`/`defer_line_comment` call at all,
regardless of whether an anchor prefixed the item (`had_property`) — #784's
fix added a `defer_line_comment()` call there, but gated it on
`had_property`, deliberately leaving the no-anchor case (`had_property ==
false`) untouched, since real yq's own null-value-plus-comment rendering for
this shape (comment relocates to its own standalone line, not attached to
any single node's trailing position) doesn't fit the "attach to a specific
bp_pos" model #784's fix uses — it would need its own investigation into
whether that's representable at all in this codebase's current
`line_comments: BTreeMap<bp_pos, (u32,u32)>` model (a comment on its own
line, not trailing any particular node, may need a different data
structure entirely — a `head_comment`/foot-comment concept, which per
existing coverage-gap notes elsewhere in this repo isn't implemented at
all).

## Scope note

Two entangled sub-problems here (comment capture at all for this shape, and
null-value rendering matching real yq's standalone-comment-line format) —
recommend scoping/investigating before implementing, not a quick fix.

newhoggy:
[...]

Issue #1077, opened 2026-08-16
yq: an anchor's own null value renders as `&anc ""` instead of bare `&anc`


newhoggy:
**Severity: Low**

## Summary

Found while implementing #784. An anchor whose value is truly null (deferred
to a following line that never materializes any content — EOF right after,
or a sibling at the same/lower indent immediately follows) renders as
`&anc ""` (anchor plus an explicit empty-string marker) instead of real
yq's bare `&anc`.

```
$ printf 'a: &anc\n' | yq '.'
a: &anc
$ printf 'a: &anc\n' | succinctly yq '.'
a: &anc ""
```

This is **unrelated to comments** — reproduces identically with no comment
anywhere in the input, and reproduces on `main` prior to #784 (confirmed via
`git stash` before implementing that issue's fix). #784 only made this more
visible, since its own tests for the "anchor's deferred value floats a
comment to the next sibling" shapes had to pin the current (buggy) `""`
output rather than real yq's bare `&anc`:

```
$ printf 'a: &anc # comment\nb: 2\n' | yq '.'
a: &anc
b: 2 # comment
$ printf 'a: &anc # comment\nb: 2\n' | succinctly yq '.'
a: &anc ""
b: 2 # comment
```

## Scope note

The underlying `OwnedValue`/JSON representation is correct (`{"a": null}` —
confirmed via `succinctly yq -o json`), so this is purely a YAML-emission
formatting bug for a null anchor target, not a parsing/data bug. Likely
somewhere in the YAML emitter's anchor-rendering logic (`owned_to_yaml_at_depth`
or the streaming equivalent) treating a null anchored value differently from
a null anchored value's key-line rendering. Whoever picks this up should
verify against the pinned real `yq` binary before assuming the fix, per this
repo's usual discipline — narrow, single-emitter-path shape, filing to make
it trackable.

newhoggy:
🔧 Claimed by agent `bruce` — branch `issue-1077-yq-null-anchor-empty-string`

PR #1081, opened 2026-08-16
fix(yaml): attach a comment trailing a deferred anchor to its value, not its key

### LINKED ISSUE #784: yq: comment after an anchor is dropped when the anchored value is deferred to the next line
## Summary

Follow-up from #765 (itself a follow-up from #710). When a node property (`&anchor`) prefixes a mapping value that is then deferred to the next line, a `#` comment trailing that same line is silently dropped from identity/output — no warning, no data loss for the anchor itself, just the comment.

```
$ printf 'a: &anc # comment\n  b: 1\n' | succinctly yq '.'
a: &anc
  b: 1
```

Real `yq` (v4.53.3) keeps it, but attaches it to the anchored value's own first field rather than the key line:

```
$ printf 'a: &anc # comment\n  b: 1\n' | yq '.'
a: &anc
  b: 1 # comment
```

Same shape with a deferred sequence value:

```
$ printf 'a: &anc # comment\n  - 1\n  - 2\n' | yq '.'
a: &anc
  - 1 # comment
  - 2
$ printf 'a: &anc # comment\n  - 1\n  - 2\n' | succinctly yq '.'
a: &anc
  - 1
  - 2
```

## Why this is distinct from #765

#765's fix (`Parser::parse_mapping_entry` in `src/yaml/parser.rs`, around line 2201-2206) only calls `maybe_capture_line_comment` in the branch where nothing but the comment/newline follows `key:` directly. When an anchor prefixes the value, parsing instead goes through the separate branch a bit further down in the same function (`src/yaml/parser.rs`, around line 2333-2392 — `parse_node_properties()` followed by its own `if self.at_line_end()` check), which calls `self.skip_to_eol()` with no `maybe_capture_line_comment` call beforehand — so the comment is scanned past and lost before any node is open to own it.

Real `yq`'s behavior (attaching to the *value's first child*, not the key) also means this almost certainly isn't a simple copy of #765's key-scoped `CommentTree`/`key_comment` machinery — it looks like real `yq` treats it as an ordinary head-of-block comment on whatever follows, not a property of the key or the anchor. Whoever picks this up should verify against the pinned real `yq` binary before assuming the attachment point.

## Scope note

Narrow, low-traffic shape (anchor + trailing comment + deferred value) — filing to make the gap trackable, per the same pattern #710 → #752/#765 already used. Not blocking either.

### LINKED ISSUE #1078: yq: compact-mapping key's anchor+trailing-comment garbles output order (`key: # comment &anc`)
**Severity: Low**

## Summary

Found while implementing #784 (deliberately left out of that PR's scope —
this is a materially different bug, not just a missing capture site).

A compact-mapping entry (`- key: value`, i.e. a mapping entry directly on a
sequence item's own dash line) whose value is `&anchor # comment` with the
value deferred to a following line garbles the output: the comment and the
anchor swap positions and end up on the *same* physical line, instead of the
anchor staying on the key's line and the comment moving to the deferred
value's own line (the behavior #784 established for plain block-mapping
keys and bare sequence items).

```
$ printf -- '- key: &anc # comment\n    b: 1\n' | yq '.'
- key: &anc
    b: 1 # comment
$ printf -- '- key: &anc # comment\n    b: 1\n' | succinctly yq '.'
- key: # comment &anc
    b: 1
```

Note the anchor `&anc` renders *after* the comment on the key's own line —
not merely mis-attached, but visibly out of order. The underlying data is
unaffected (`b: 1` still nests correctly under `key`, confirmed via `-o
json`); this is purely an emission-ordering bug.

## Root cause (partial)

`parse_compact_mapping_entry` (`src/yaml/parser.rs`) has its own,
independent copy of #765's "capture a trailing comment on the key's own
line before it's deferred" logic — unlike `parse_mapping_entry`, it doesn't
distinguish "comment follows an anchor that itself needs deferring" from
"comment follows the key with no anchor at all": it captures *any* trailing
comment unconditionally onto the key's own bp before even checking whether
an anchor consumed part of the line. #784's fix pattern (defer the comment
until the value's own resolution is known, per `defer_line_comment`/
`take_pending_head_comment` in the same file) would need to be threaded
through this function's own anchor-handling branch too — but the *ordering*
symptom (anchor after comment on the same rendered line) suggests there may
be a second, independent bug in how the YAML emitter orders an anchor
decoration relative to a same-node trailing comment, worth understanding
before assuming the parser-side fix alone resolves it.

## Scope note

Narrow, low-traffic shape (anchor + comment + deferred value, specifically
in the compact-mapping-inside-sequence-item form) — filing to make it
trackable, not blocking.

newhoggy:
Fixes #784
Fixes #1078

## Summary

`&anchor # comment` whose value is deferred to a following line (`a: &anc # comment\n  b: 1`) silently dropped the comment. Real yq attaches it to the deferred value's own first line instead (`b: 1 # comment`), or, if that value turns out null, floats it further to the next sibling entirely — never to the anchor's own key line.

Added a two-phase deferred-comment mechanism in `src/yaml/parser.rs`:
- `defer_line_comment` stashes the comment's byte range in a new `pending_head_comment` field when the owning node doesn't exist yet (the anchor's own line). Never overwrites an already-pending comment (a nested anchor deferring a second comment before the first is claimed can't destroy it).
- `take_pending_head_comment` claims it at whichever bp_pos the renderer already reads a trailing comment from for that shape of node — a mapping key (`parse_mapping_entry`, `parse_compact_mapping_entry`), a plain-scalar value (`parse_block_node`'s two scalar arms — the single most common resolution of a deferred anchor), or a plain-scalar sequence item's own scalar bp (`parse_sequence_item_inner`, guarded against misrouting into a flow collection's last inner element).
- `drop_stale_pending_head_comment` bounds a comment to one line's grace period, so it can never leak onto an unrelated node further down the document.
- Every consumption site runs *after* the node's own ordinary same-line-comment capture, as a fallback — never before, which a first-pass code review caught as a real regression (an eager, key-open-time claim silently destroyed #765/#785's own already-working comment capture whenever it collided with an unrelated floated one).

Extracted `scan_trailing_comment` as the one scan both `maybe_capture_line_comment` and `defer_line_comment` now share.

31 tests in `tests/yq_cli_tests.rs` (11 from the initial pass + 10 added after code review), every shape verified byte-for-byte against the pinned real `yq` binary, matching this repo's existing #765/#710 discipline. Three of the added tests specifically guard the collision/misrouting bugs review found against reintroduction; one collision outcome (own comment survives, floated one drops) was confirmed byte-identical to unmodified `main`'s pre-existing behavior via `git stash` diff, not just asserted.

## Code review findings, all fixed
- **Regression**: eager key-open consumption destroyed a node's own genuine comment when it collided with a floated one — fixed by reordering every site to consume only as a fallback.
- **Missing coverage**: the single most common resolution of a deferred anchor (a plain/quoted scalar folded onto the next line, no container) was completely unhandled — added to `parse_block_node`'s two scalar arms.
- **#1078** (compact-mapping key's own anchor+comment garbled token order) — `parse_compact_mapping_entry` previously always used #765-style local capture regardless of whether an anchor caused the deferral; now splits on `had_property` like its siblings.
- **Missing coverage**: `parse_explicit_value` (the `: v` half of `? k`/`: v`) dropped an anchor's comment unconditionally — added the same defer-on-`had_property` branch its siblings have.
- **Structural corruption**: a floated comment landing on a flow-collection sequence-item value misrouted into the collection's last inner element instead of being safely dropped — guarded against.
- **Simplification**: extracted the duplicated comment-scan logic into one shared helper.

## Follow-ups filed (found while implementing, deliberately out of scope here)
- #1077 — an anchor's own null value renders as `&anc ""` instead of bare `&anc`, pre-existing and unrelated to comments (reproduces on `main` with no comment present at all).
- #1079 — a *bare* sequence item's own comment (no anchor) is dropped when its value is deferred — #765's fix never covered sequence items, only mapping keys.
- #1080 — chained double-anchor deferral renders unusually even in real yq; likely simplifies once #1077 lands, don't implement in isolation.
- #1085 — a floated comment and a target's own genuine comment can't both survive (single comment slot per bp_pos) — needs a data model change (multiple comments per node, or a head-comment concept) out of scope here.

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` (full suite, 684 passed)
- [x] `cargo fmt --check`
- [x] Every new test's expected output verified against the pinned real `yq` v4.53.3 binary
- [x] 10-angle code review, all findings fixed or explicitly deferred to filed follow-ups

### DIFF
--- a/src/yaml/parser.rs
+++ b/src/yaml/parser.rs
@@ -183,6 +183,18 @@ struct Parser<'a, const HAS_CR: bool> {
     /// See [`SemiIndex::line_comments`].
     line_comments: BTreeMap<usize, (u32, u32)>,
 
+    /// A comment trailing a `&anchor`/`!tag` whose value is deferred to a
+    /// later line, not yet attached to any node (#784).
+    ///
+    /// Real yq attaches such a comment to whatever node the deferred value
+    /// resolves to — its first child if one follows, or (if the value turns
+    /// out null) the next sibling entirely — never to the anchor's own key
+    /// line. That target doesn't exist yet at the point the comment is
+    /// scanned, so [`Self::defer_line_comment`] stashes the byte range here
+    /// and [`Self::take_pending_head_comment`] claims it once the next
+    /// primary node (a mapping key or sequence item) actually opens.
+    pending_head_comment: Option<(u32, u32)>,
+
     // Document tracking
     /// Whether we're currently inside a document
     in_document: bool,
@@ -276,6 +288,7 @@ impl<'a, const HAS_CR: bool> Parser<'a, HAS_CR> {
             aliases: BTreeMap::new(),
             tags: BTreeMap::new(),
             line_comments: BTreeMap::new(),
+            pending_head_comment: None,
             in_document: false,
             document_start_bp_pos: 0,
             pending_explicit_key: None,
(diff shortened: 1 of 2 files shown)


newhoggy:
Code review (10 finder angles, deep multi-round investigation) found the initial pass had a real regression plus several missing shapes. All fixed:

**Regression (confirmed via `git stash` diff against unmodified `main`, not just inspection):** eager "claim the pending comment at key-open" consumption ran *before* #765/#785's own established same-line-comment capture for the same bp_pos — when both existed, the eager claim won the `or_insert` race and silently destroyed the node's own genuine, already-working comment. Fixed by reordering every consumption site to run only as a fallback, after the real capture has already had its chance.

**Missing coverage, now added:**
- `parse_block_node`'s two scalar-writing arms — the single most common shape a deferred anchor resolves to (a plain/quoted scalar folded onto the next line, no container) had zero handling.
- `parse_compact_mapping_entry`'s own anchor+comment (closes #1078 — previously garbled token order on the key's own line instead of deferring).
- `parse_explicit_value` (`: v` half of `? k`/`: v`) — dropped the comment unconditionally before.

**Structural-corruption fix:** a floated comment landing on a flow-collection sequence-item value used to misroute into the collection's last inner element (`- [1, 2 # comment\n  ]`), corrupting the emitted structure. Now guarded — drops cleanly instead, matching real yq's own behavior for that shape.

**Simplification:** extracted the duplicated whitespace-skip/comment-scan loop (`defer_line_comment` vs `maybe_capture_line_comment`) into one shared `scan_trailing_comment`.

10 new tests pin all of the above, 3 specifically as regression guards. One remaining architectural gap — a floated comment and a target's own genuine comment can't both survive (single comment slot per bp_pos; real yq shows both) — is filed as #1085 rather than attempted here, since it needs a data-model change genuinely out of scope for this fix.

All local verification re-run clean: full `--no-fail-fast` test suite (684 tests), 3x clippy matrix, fmt, no_std check.

Issue #1083, opened 2026-08-16
jq: overflowed NumberLiteral (e.g. 1e400) substitutes DBL_MAX text instead of real jq's literal-preserving reformat


newhoggy:
**Severity: Low**

## Summary

Found while implementing #1075's fix. Real jq's decNumber-backed literal
preservation reformats a *positive* overflowed `NumberLiteral`'s own source
text (`1e400 | tostring` -> `"1E+400"`) rather than substituting `DBL_MAX`
text, up to its own re-normalization ceiling (`format_number_jq_compat`
already implements this exactly, ceiling included -- #930). succinctly's
`numeric_display_string` (`src/jq/eval.rs`) instead unconditionally routes
*every* jq-mode infinite `NumberLiteral` through `nonfinite_display_string`'s
`DBL_MAX` substitution, regardless of whether it's a genuine, small-magnitude
document literal or a computed value that only looks like one.

## Repro (verified against pinned jq 1.7.1)

```
$ echo null | jq -r '1e400 | tostring'
1E+400
$ echo null | succinctly jq -r '1e400 | tostring'
1.7976931348623157e+308
```

Real jq's own asymmetry: a *negative* overflowed literal (`-1e400`) does NOT
keep its literal spelling -- it also substitutes `DBL_MAX` text, which
succinctly already matches correctly:

```
$ echo null | jq -r '-1e400 | tostring'
-1.7976931348623157e+308
$ echo null | succinctly jq -r '-1e400 | tostring'
-1.7976931348623157e+308
```

This also affects `@uri`/`@html`/`@sh`/`@csv`/`@tsv`/string interpolation,
and every reindex-bridge path (`reduce`/`foreach`/`as $x`/multi-stage pipes),
which is exactly why the fix is not a one-line change -- see "Root cause"
below.

## Root cause

`numeric_display_string`'s `NumberLiteral` branch (`src/jq/eval.rs`)
intercepts *any* infinite `NumberLiteral` in jq mode and routes it through
`nonfinite_display_string`'s `DBL_MAX` substitution, rather than falling
through to `number_str()` -> `format_number_jq_compat`, which already
reformats an overflowed literal's own text correctly (oracle-verified up to
its own ~999999999-exponent ceiling, beyond which it *also* falls back to
`DBL_MAX` text -- see `format_number_jq_compat`'s own tests in
`src/jq/value.rs`).

Removing that interception for jq mode (the naive fix) is not safe, though:
[...]

PR #1084, opened 2026-08-16
fix(jq): correct nonfinite_display_string's jq-mode NaN/Infinity spelling

### LINKED ISSUE #1075: jq: nonfinite_display_string's jq-mode branch diverges from real jq's own NaN/Infinity spelling
## Summary

Found during #1064's code review (live-verified independently). `nonfinite_display_string`'s jq-mode branch spells a non-finite float via Rust's own `f64::Display` ("NaN"/"inf"/"-inf"), but real jq has its own, different convention for the same case.

## Repro

```
$ echo null | jq -r 'nan|tostring'
null
$ echo null | jq -r 'infinite|tostring'
1.7976931348623157e+308

$ echo null | succinctly jq -r 'nan|tostring'
NaN
$ echo null | succinctly jq -r 'infinite|tostring'
inf
```

Also reproducible via `[nan]|@sh`, `nan|@uri`, `nan|@html`, string interpolation (`"\(nan)"`), and any other caller of `numeric_display_string`'s jq-mode path.

## Root cause

`nonfinite_display_string` (`src/jq/eval.rs`) is the single shared "how does a non-finite float spell itself" decision, added by #1060 specifically for yq-mode's `.nan`/`.inf`/`-.inf` convention. Its jq-mode branch (the `S::TAG != EvalTag::Yq` arm) was left as a straight `f64::Display` pass-through, which doesn't match real jq's own behavior: jq spells NaN as JSON `null` (its RFC-8259 substitution convention, same as document-sourced NaN) and infinity as `DBL_MAX`'s literal decimal expansion (`1.7976931348623157e+308` for `+infinity`, presumably `-1.7976931348623157e+308` for `-infinity` — not yet verified).

This is **pre-existing** — #1064 only changed this function's visibility (`pub`) and made it the single canonical definition reused by 5 more call sites; it did not touch this branch's logic. The codebase has already hit and fixed this same bug class once, locally, in `join_step` (`src/jq/eval.rs`, referencing #1036: routing a NaN/Infinity element through `to_json()` instead of `owned_to_string`'s raw `Display`) — but `builtin_tostring` and every other `numeric_display_string`/`nonfinite_display_string` caller in jq mode still hits the unfixed path.

## Why this matters more after #1064

#1064 made `nonfinite_display_string` the crate's single canonical, `pub`, cross-crate-reachable definition for this decision — exactly the place a fix like this should live once, rather than needing to be re-applied at each caller. Filing this now while the context is fresh, rather than letting the bug's blast radius grow further as more call sites adopt the shared function.

## Suggested approach

Fix the `S::TAG != EvalTag::Yq` branch to match real jq's actual convention (`null` for NaN, `DBL_MAX`'s literal text for ±infinity) instead of `f64::Display`. Needs live-verification of the exact expected string for `-infinity` and confirmation this doesn't regress the existing #1036 `join` fix (which already produces correct output via a different mechanism — `to_json()` — so should be checked for consistency with whatever this fix does).

newhoggy:
## Summary

- `nonfinite_display_string`'s jq-mode branch spelled a non-finite float via Rust's own `f64::Display` (`"NaN"`/`"inf"`/`"-inf"`) instead of real jq's own JSON-substitution convention (`null` for NaN, `f64::MAX`'s literal decimal text for ±infinity) — confirmed live against jq 1.7.1: `nan | tostring` -> `"null"`, `infinite | tostring` -> `"1.7976931348623157e+308"`.
- Affected `tostring`, `@uri`, `@html`, `@sh`, `@csv`/`@tsv`/`@dsv`, and string interpolation in jq mode (yq mode's `.nan`/`.inf`/`-.inf` spelling is unaffected).
- Reuses the existing `infinite_float_preview_text` helper (already used by `format_number_jq_compat`'s own overflow ceiling and error-preview text) instead of duplicating the `DBL_MAX` literal — one definition, not two.

A related, deeper gap was found and deliberately **not** fixed here: a jq-mode `NumberLiteral` overflowing to infinity (e.g. `1e400`) still takes this same substitution rather than real jq's own literal-preserving reformat (`1e400 | tostring` -> `"1E+400"` in real jq). An earlier draft of this fix removed that interception, but it turned out unsafe: the `eval_owned_expr`/`eval_generic` reindex bridge (`reduce`/`foreach`/`as $x`/multi-stage pipes — reached by almost any non-trivial expression, not just explicit loops) round-trips a *computed* infinity through its own `1e999`/`-1e999` smuggling literal, which — unlike its `NAN_SENTINEL` sibling — is not guaranteed-unparseable as a genuine document literal. Letting jq mode fall through to `format_number_jq_compat` reformatted that internal sentinel as `"1E+999"` instead of substituting `DBL_MAX` text, a real regression caught during review. Filed as #1083 for whoever picks it up next; it needs the reindex bridge's own sentinel made collision-safe first (mirroring `NAN_SENTINEL`'s design).

Closes #1075.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full workspace suite, worktree-isolated build) — 0 failures
- [x] `cargo clippy --all-targets --features std,simd,serde,cli,regex,bench-runner,large-tests,mmap-tests -- -D warnings` — clean (catches YAML SIMD backends `--all-features` compiles out)
- [x] `cargo check --no-default-features` — `no_std` compiles clean
- [x] Diffed output against pinned real `jq` 1.7.1 directly for every affected format (`tostring`, `@uri`, `@html`, `@sh`, `@csv`, string interpolation, plus the `reduce`/`foreach`/`as $x`/`with_entries` reindex-bridge paths)
- [x] Updated every existing test that encoded the old (incorrect) `"inf"`/`"-inf"`/`"NaN"` expectations across `src/jq/eval.rs`, `src/jq/eval_generic.rs`, `tests/jq_cli_tests.rs`, and `tests/jq_evaluator_parity_tests.rs`
- [x] Added a new regression test (`test_builtin_tostring_computed_nonfinite_matches_jq_1075`) directly covering the issue's own repro

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -5156,7 +5156,11 @@ fn eval_string_interpolation<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
 /// Render a numeric value the way the non-JSON text formats want it.
 ///
 /// `number_str()` deliberately skips NaN/Infinity handling (see its doc
-/// comment) because `to_json` needs "null" instead. Text formats (`@text`,
+/// comment) because `to_json` needs its own substitution instead (currently
+/// "null" for both NaN and Infinity in jq mode -- correct for NaN, but not
+/// for a *computed* Infinity, which real jq instead gives `DBL_MAX` text the
+/// same as this function now does; that's issue #1087, separate from and
+/// out of scope for this function's own fix). Text formats (`@text`,
 /// `@uri`, `@html`, `@sh`, and the CSV/TSV/DSV cell formatter that falls back
 /// to `owned_to_string`) instead want jq's own JSON-substitution spelling for
 /// a non-finite `Float`/`NumberLiteral` (#1075: `null` for NaN, `DBL_MAX`'s
@@ -5239,13 +5243,10 @@ pub(crate) fn numeric_display_string<S: EvalSemantics>(value: &OwnedValue) -> St
 /// `infinite | tostring` -> `"1.7976931348623157e+308"`, `-infinite |
 /// tostring` -> `"-1.7976931348623157e+308"`).
 ///
-/// See `numeric_display_string`'s own doc comment for the
-/// "document-sourced vs computed" caveat this doesn't attempt to resolve
-/// (not a doc link: that function is `pub(crate)`, unresolvable from this
-/// one's now-`pub` docs) -- this includes jq mode's own version of that gap
-/// (a jq-mode `NumberLiteral` overflowing to infinity, e.g. `1e400`, is
-/// still routed through here rather than real jq's own literal-preserving
-/// reformat; see that function's doc comment for why, and issue #1083).
+/// See `numeric_display_string`'s own doc comment (`src/jq/eval.rs`, not
+/// doc-linkable from here since it's `pub(crate)`) for the
+/// "document-sourced vs computed" caveat this doesn't attempt to resolve,
+/// and issue #1083 for the gap that leaves open.
 ///
 /// `pub`, not `pub(crate)`: `src/bin/succinctly/yq_runner.rs` is a separate
 /// binary crate depending on this one as an external dependency, and is one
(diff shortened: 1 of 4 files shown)


newhoggy:
## Code review summary (6 finder agents, high effort)

**Fixed:**
- Stale doc comment in `eval_generic.rs` still describing pre-#1075 jq-mode spelling as current.
- Doc-comment duplication in `nonfinite_display_string` (3 independent reviewers flagged this) — trimmed to a bare cross-reference.
- **Real factual error**, caught by one reviewer's own oracle re-verification: my claim that a negative overflowed literal (`-1e400 | tostring`) "genuinely matches real jq's `DBL_MAX` substitution" was wrong — I'd tested that claim with the literal embedded in *filter* text (`(-1e400)|tostring`), which hits jq's unary-negation grammar (#1035), not the *input-document* shape every actual test here uses (`query!(br"-1e400", ...)`). Re-verified: real jq preserves **both** signs' literal spelling identically for the input-document shape, so there's no real sign asymmetry for what this PR's tests exercise. Corrected all affected comments/tests and posted a correction on #1083.

**Filed as follow-ups** (confirmed real, but out of scope for this PR):
- #1083 — jq-mode overflowed `NumberLiteral` doesn't get real jq's literal-preserving reformat (needs the reindex bridge's `1e999` sentinel made collision-safe first, mirroring `NAN_SENTINEL`'s design).
- #1087 — `OwnedValue::to_json`/`to_json_at_depth` nulls a *computed* Infinity, not just NaN (confirmed via `[infinite]`, `{"a":infinite}`, `tojson`, `join` all diverging from real jq; yq's `-o json` behavior also worth a look, since real yq errors instead of nulling).

No correctness bugs in the diff's actual runtime behavior were found — all findings were comment-accuracy/duplication issues, now fixed in a follow-up commit. Full test suite (2524 lib + 553 CLI + parity/golden suites) green, clippy clean (both feature-set invocations), `no_std` check clean.

PR #1092, opened 2026-08-17
fix(jq): unary minus preserves IEEE754 negative-zero sign

### LINKED ISSUE #1056: jq: unary minus ("0 - expr") loses IEEE754 negative-zero sign, diverging from jq's -0

### LINKED ISSUE #1091: jq: unary minus on an all-integer zero result still loses IEEE754 sign, and error wording still says 'cannot be subtracted' (#1056 follow-up)
**Severity: Low**

## Summary

Follow-up to #1056, whose fix (`(-1) * expr` instead of `(0 - expr)` for
unary minus on a non-literal operand) covers the common case — negating a
document field or computed expression that evaluates to a *float* zero
(`-.a` on `0.0`, `-(1.0-1.0)`) — but leaves two narrower gaps open that the
same rewrite can't reach:

1. **An all-integer computation that reduces to exactly zero still loses
   the sign**: `-(1-1)` gives `0`, not real jq's `-0`.
2. **Unary minus on a non-numeric operand still reports the wrong error**:
   `-"abc"` gives `number (-1) and string ("abc") cannot be multiplied`
   (previously `number (0) and string ("abc") cannot be subtracted`, before
   #1056's rewrite) instead of real jq's own dedicated `string ("abc")
   cannot be negated`.

```
$ echo null | jq -c -- '-(1-1)'
-0
$ echo null | succinctly jq -c -- '-(1-1)'
0

$ echo null | jq -c -- '-"abc"'
jq: error (at <stdin>:1): string ("abc") cannot be negated
$ echo null | succinctly jq -c -- '-"abc"'
jq: error (at <stdin>:1): number (-1) and string ("abc") cannot be multiplied
```

## Root cause

Both gaps trace back to the same thing: #1056 fixed unary minus by lowering
it to *binary* arithmetic (`(-1) * expr`), which is only an approximation of
real negation.

- Real jq has no separate integer type — `1 - 1` is already the IEEE-754
  double `0.0` internally, so negating it naturally yields `-0.0`. succinctly
  deliberately keeps a distinct `OwnedValue::Int` (for exact i64 literal
  preservation), and `Int * Int` correctly stays `Int` for genuine
  multiplication — `i64` has no negative-zero representation at all, so
  `(-1) * Int(0)` can only ever produce `Int(0)`. Special-casing "`-1 *
  0`" inside `arith_mul` to promote to `Float(-0.0)` would be *incorrect*:
  it would also fire for a literal `(-1) * (2-2)` a user wrote directly, not
  just the unary-minus rewrite, silently changing genuine multiplication's
  result type.
- The error wording is generic multiplication's own message
  (`arith_mul`'s type-mismatch path), not a message aware that this
  particular multiplication is standing in for negation.

## Suggested fix direction

Both need a representation that's unambiguously "this is a negation," not
an arithmetic rewrite that merely produces the same *value* in the common
case — e.g. a dedicated unary AST variant (an `Expr::Negate(Box<Expr>)`, or
a unary `ArithOp::Negate` reusing the existing `Expr::Arithmetic` shape with
`right` unused) evaluated by its own function: `Int(n) => Int(-n)` normally,
but `Int(0) => Float(-0.0)` specifically (mirroring real jq's uniform-double
model just for this one case), `Float(f) => Float(-f)` (plain IEEE-754
negation, no special-casing needed), and a dedicated `"{type} ({value})
cannot be negated"` error for everything else.

This needs threading through both evaluators (`src/jq/eval.rs`,
`src/jq/eval_generic.rs`, which must stay in parity per
`tests/jq_evaluator_parity_tests.rs`) and auditing any path-tracking code
that pattern-matches on `Expr::Arithmetic` in case unary minus is ever
reachable as (or adjacent to) a path expression — not investigated here,
scoped out of #1056 to keep that fix small and safe.

## Why this is separate from #1056

#1056's own fix is complete, verified, and safe for its stated scope (the
common float-zero case) — pinned by a regression test
(`test_unary_minus_integer_zero_still_loses_sign_1056_followup` in
`src/jq/eval.rs`) that explicitly documents this exact remaining gap rather
than silently leaving it uncovered. This issue tracks the deeper,
architecturally bigger fix (a real unary-negation representation) that
closing this gap and the error-wording divergence together would need.

newhoggy:
## Summary

- Unary minus on a non-literal operand (`-.a`, `-(expr)`) was lowered to `(0 - expr)`, whose IEEE-754 `0.0 - 0.0` evaluates to positive zero, silently dropping the sign of a zero-valued operand — real jq prints `-0` for these (`-.a` on `0.0`, `-(1.0-1.0)`, `-(0.0)`, `-(1-1)`), but succinctly printed plain `0`.
- **First draft rewrote it to `(-1 * expr)`** (matching #1035's precedent for the filter-literal case) — **code review caught a critical regression**: `arith_mul` has its own string-repetition (`Int * String`) and null-passthrough semantics that have nothing to do with negation, so `-"abc"`/`-null` silently returned `null` (exit 0) instead of erroring like real jq.
- **Final fix**: a dedicated `ArithOp::Negate`/`arith_negate`, reusing `Expr::Arithmetic`'s binary shape only for its existing fan-out machinery (a dummy `left` operand). `Int` negates via `checked_neg`/`wrapping_neg` (matching `arith_sub`'s existing `i64::MIN` overflow handling exactly); `Float` negates via Rust's own unary `-`, correctly preserving zero's sign. Non-numeric operands now error with real jq's own `"cannot be negated"` wording.
- Since `arith_negate` has exactly one caller (unlike `arith_mul`), it can also safely promote an all-integer computation that reduces to exactly zero (`-(1-1)`) to `Float(-0.0)` — real jq has no separate integer type, so this is already the double `0.0` there. This closes the fix's own remaining gap in full.

Closes #1056. Closes #1091 (opened mid-review for the two gaps the `(-1 * expr)` draft couldn't reach; both are fixed by the final `arith_negate` approach, so closed as resolved rather than left open).

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full workspace suite, twice — once per draft) — 0 failures
- [x] `cargo check --no-default-features` — `no_std` compiles clean
- [x] Diffed output against pinned real `jq` 1.7.1 for every original repro, the regression the review caught (`-"abc"`, `-null`, `-true`, `-[1,2]`, `-{}`), and sanity checks (ordinary negation, double negation, generator fan-out `-(1,2,3)`, `i64::MIN` overflow)
- [x] `test_unary_minus_zero_preserves_sign_1056` (all 4 original repro cases + sanity), `test_unary_minus_non_numeric_errors_1056` (the regression the review caught, previously zero coverage)

### DIFF
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -717,6 +717,14 @@ impl EvalError {
         Self::subject(value, "has no keys")
     }
 
+    /// `<v> cannot be negated` — unary minus (`-expr`) on a non-numeric
+    /// operand (#1056). Matches real jq's own dedicated wording, confirmed
+    /// live against jq 1.7.1 (`-"abc"` -> `string ("abc") cannot be
+    /// negated`).
+    pub fn cannot_be_negated(value: &OwnedValue) -> Self {
+        Self::subject(value, "cannot be negated")
+    }
+
     /// `<v> has no length`.
     pub fn has_no_length(value: &OwnedValue) -> Self {
         Self::subject(value, "has no length")
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -1718,19 +1718,31 @@ fn arith_sub<S: EvalSemantics>(
 /// sign bit for every value including `0.0`. `Int(n)` negates via
 /// `checked_neg`/`wrapping_neg` (jq/yq's usual overflow conventions,
 /// matching `arith_sub`'s `0 - n` handling exactly: only `i64::MIN` can
-/// overflow, since `-i64::MIN` doesn't fit in `i64`) -- *except* for `n ==
-/// 0`, which promotes to `Float(-0.0)` instead of staying `Int(0)`: unlike
-/// `arith_mul`, this function has exactly one caller (unary minus itself),
-/// so special-casing zero here can't accidentally reinterpret some
-/// unrelated expression the way special-casing `-1 * 0` inside `arith_mul`
-/// would have (see this function's own history above) -- real jq has no
-/// separate integer type, so an all-integer computation that reduces to
-/// exactly zero (`-(1-1)`) is already the double `0.0` there, and negating
-/// it naturally yields `-0.0`; this promotion is the one place `Int`
-/// deliberately gives up exact-integer fidelity to match that.
+/// overflow, since `-i64::MIN` doesn't fit in `i64`) -- *except* for jq mode
+/// specifically with `n == 0`, which promotes to `Float(-0.0)` instead of
+/// staying `Int(0)`: unlike `arith_mul`, this function has exactly one
+/// caller (unary minus itself), so special-casing zero here can't
+/// accidentally reinterpret some unrelated expression the way
+/// special-casing `-1 * 0` inside `arith_mul` would have (see this
+/// function's own history above) -- real jq has no separate integer type,
+/// so an all-integer computation that reduces to exactly zero (`-(1-1)`) is
+/// already the double `0.0` there, and negating it naturally yields `-0.0`;
+/// this promotion is the one place `Int` deliberately gives up
+/// exact-integer fidelity to match that.
+///
+/// **jq-mode only**: real yq has no unary-minus operator at all (confirmed
+/// live against yq v4.53.3: `-.a` is a hard parse error, `'-' expects 2
+/// args but there is 1`) -- succinctly's own support for it in yq mode is
+/// a pre-existing, unrelated extension (#1056 doesn't touch whether it
+/// exists), so there's no oracle to say a zero-valued yq integer field
+/// negating to a float (`-.replicas` on `0` becoming `-0.0` instead of
+/// staying `0`) is the *intended* behavior rather than a surprising type
+/// change to a config value -- caught by code review before reaching
+/// `main`. Leaving yq's `Int(0)` un-promoted keeps this fix scoped to
+/// jq-mode's own verified, oracle-backed convention.
 fn arith_negate<S: EvalSemantics>(operand: OwnedValue) -> Result<OwnedValue, EvalError> {
     match operand.into_plain_number() {
-        OwnedValue::Int(0) => Ok(OwnedValue::Float(-0.0)),
+        OwnedValue::Int(0) if S::TAG != EvalTag::Yq => Ok(OwnedValue::Float(-0.0)),
         OwnedValue::Int(n) => Ok(OwnedValue::Int(if S::OVERFLOW_WRAPS {
             n.wrapping_neg()
         } else {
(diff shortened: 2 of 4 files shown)


newhoggy:
## Code review summary (2 rounds, 5 finder agents total, high effort)

**Round 1** caught a critical regression: the initial \`(-1 * expr)\` rewrite silently returned \`null\` for \`-"abc"\`/\`-null\` instead of erroring (\`arith_mul\`'s string-repetition and null-passthrough semantics have nothing to do with negation). Replaced with a dedicated \`ArithOp::Negate\`/\`arith_negate\`, which also let the fix fully close the integer-zero gap (\`-(1-1)\` -> \`-0\`) that #1091 was tracking separately — #1091 closed as resolved.

**Round 2** (re-reviewed after that rewrite) found:
- **Real, fixed**: the \`Int(0) -> Float(-0.0)\` promotion applied unconditionally, silently changing a yq integer field's type when negated (no real-yq oracle exists for this at all — yq has no unary-minus operator). Gated to jq mode only, with new yq-mode test coverage.
- **Filed as #1100** (efficiency/simplification, not correctness): reusing \`Expr::Arithmetic\`'s binary shape with a dummy \`left\` operand re-evaluates that dummy once per output of a generator operand (e.g. \`-(1,2,3)\`) instead of once overall. A dedicated \`Expr::Negate(Box<Expr>)\` — matching the ~20+ existing single-child \`Expr\` variants — would avoid this, but is a larger, separately-reviewable change.

Full test suite (2532 lib + all integration suites) green, clippy clean (both feature-set invocations), \`no_std\` clean, output diffed against pinned real \`jq\` 1.7.1 and \`yq\` v4.53.3 throughout.

Issue #1094, opened 2026-08-17
jq: --argjson rejects leading-zero numbers (007, 00) that real jq accepts leniently


newhoggy:
**Severity: Low**

## Summary

Found incidentally while implementing #1058 (unrelated to that issue's own
fidelity scope — this is a validation-strictness divergence, not a
fidelity one).

Real jq's \`--argjson\` accepts JSON numbers with leading zeros (\`007\`,
\`007.5\`, \`00\`) — not RFC 8259-valid JSON, but jq's own number parser is
more lenient than strict JSON here. succinctly's \`--argjson\` rejects them
outright, via \`serde_json::from_str\`'s strict RFC 8259 validation:

```
$ jq -n --argjson n '007' '$n'
7
$ succinctly jq -n --argjson n '007' '$n'
Error: Invalid JSON for --argjson n

Caused by:
    0: Invalid JSON: 007
    1: invalid number at line 1 column 2
```

Same for \`007.5\` → \`7.5\` and \`00\` → \`0\` under real jq; both error under
succinctly.

## Root cause

\`src/bin/succinctly/jq_runner.rs\`'s \`parse_json_value\` (fixed for fidelity
in #1058) still gates on \`serde_json::from_str\`'s strict RFC 8259
validation first, before ever reaching this crate's own, more permissive
JSON semi-indexer. That strict gate is deliberate and load-bearing for
*trailing garbage* rejection (\`42 garbage\`, #284) — but it also rejects
leading-zero numbers that real jq's own parser tolerates, which #284's own
scope never considered.

## Scope note

Fixing this without reintroducing #284's regression needs a parser that
rejects trailing garbage *but* tolerates a leading-zero number — neither
\`serde_json\`'s strict mode nor this crate's own lenient
\`JsonIndex\`-based semi-index (accepts trailing garbage) does both today.
Low-traffic edge case (leading-zero \`--argjson\` values aren't common) —
filing to make the gap trackable, not blocking.

newhoggy:
🔧 Agent `bruce`: claimed, starting work.

Issue #1093, opened 2026-08-17
jq: --slurpfile/--seq lose number-literal source fidelity, same gap as --argjson (#1058)


newhoggy:
**Severity: Low**

## Summary

Follow-up from #1058, which fixed \`--argjson\`/\`--jsonargs\`'s number-literal
fidelity loss but deliberately left \`--slurpfile\` and \`--seq\` (RFC 7464 JSON
text sequences) untouched — they have the identical gap:

```
$ echo '1.500' > /tmp/n.json
$ succinctly jq -n --slurpfile n /tmp/n.json '$n'
[
  1.5
]
$ jq -n --slurpfile n /tmp/n.json '$n'
[
  1.500
]

$ printf '\x1e1.500\n' | succinctly jq --seq '.'
1.5
$ printf '\x1e1.500\n' | jq --seq '.'
1.500
```

## Root cause

\`src/bin/succinctly/jq_runner.rs\`'s \`parse_json_stream\` (backing
\`--slurpfile\`) and \`parse_json_seq\` (backing \`--seq\`) both convert through
\`serde_to_owned\`/\`serde_to_owned_at_depth\`, same as \`parse_json_value\` did
before #1058 — building a plain \`OwnedValue::Int\`/\`Float\` from
\`serde_json::Value::Number\` instead of preserving the source spelling.

## Why this needs a different fix than #1058

#1058's fix works because \`parse_json_value\` handles exactly *one* JSON
value per call — it validates the whole string via \`serde_json::from_str\`
(for strict rejection of trailing garbage, #284), then reparses the *same*
string through this crate's own fidelity-preserving semi-indexer
(\`JsonIndex::build\`).

\`parse_json_stream\`/\`parse_json_seq\` instead handle a single string
containing **multiple concatenated JSON values** with no separator
(\`parse_json_stream\`) or RS-separated (\`parse_json_seq\`) — \`JsonIndex\` is
built for exactly one JSON document per call, not a stream. Fixing this
needs either: (a) tracking each value's own byte span within the source
string (\`serde_json::StreamDeserializer\` exposes \`.byte_offset()\` after each
successful \`.next()\`, which could delimit each value's slice for a
per-value \`JsonIndex::build\` call), or (b) some other value-boundary
mechanism — needs scoping before implementing, not a copy-paste of #1058's
fix.

## Scope note

Low-traffic flags (\`--slurpfile\`/\`--seq\` are less commonly used than
[...]

Issue #284, opened 2026-07-22
CLI: yq --arg / --argjson variables are always "undefined" (context is parsed then discarded)


newhoggy:
Discovered while closing coverage gaps on #278 (the #271 fix).

`succinctly yq` accepts `--arg NAME VALUE` and `--argjson NAME VALUE` (defined in `main.rs`, documented under yq → Variables in `docs/guides/cli.md`), but the variables are never available to the filter — every reference errors as undefined:

```
$ echo 'a: 1' | syq --arg g hello '$g'
yq: error: undefined variable: $g

$ echo 'name: x' | syq --argjson n 42 '.n = $n' -o=json -I=0
yq: error: undefined variable: $n
```

The sibling `jq` command implements the same flags correctly:

```
$ echo '{}' | sjq --arg g hello '.g = $g' -c
{"g":"hello"}
$ echo '{}' | sjq --argjson n 42 '.n = $n' -c
{"n":42}
```

Since yq's `--arg`/`--argjson` are jq-inherited extensions (mikefarah yq has no equivalent), jq's behavior is the reference — as with #271.

## Root cause

`run_yq` builds the variable context but never applies it. `parse_variables()` is called at `yq_runner.rs:1087` and threaded as `&context` into `evaluate_input` at every call site (1309, 1335, 1345, 1392, 1444, 1513) — but `evaluate_input` **ignores it**:

```rust
// src/bin/succinctly/yq_runner.rs:279
fn evaluate_input(
    input: &OwnedValue,
    expr: &jq::Expr,
    _context: &EvalContext,   // <-- underscore: never read
) -> Result<Vec<OwnedValue>> {
    ...
    let result = jq::eval::<Vec<u64>, YqSemantics>(expr, cursor);  // :292 — no variable substitution
```

The direct-cursor path is the same: `evaluate_yaml_cursor` (`:316`) and the M2 streaming macros (`:1160`, `:1187`) call `eval_with_cursor(&program.expr, cursor)` with no variables either.

`jq` works because `jq_runner` **substitutes named variables into the expression AST before evaluating** (`jq_runner.rs:668-673`):

```rust
let mut all_vars: Vec<(&str, &OwnedValue)> =
    context.named.iter().map(|(k, v)| (k.as_str(), v)).collect();
all_vars.push(("ARGS", &args_value));
let expr = jq::substitute_vars(&expr, all_vars);
```

[...]

PR #1095, opened 2026-08-17
fix(jq): --argjson/--jsonargs preserve number-literal source fidelity

### LINKED ISSUE #1058: CLI: --argjson loses number-literal source fidelity (serde_json round-trip, unlike filter literals)

newhoggy:
Fixes #1058

## Summary

Since #1035, a filter-embedded number literal (`1.500`, `1e100`) preserves jq's exact source spelling. `--argjson`'s value parsing was a separate code path that didn't — `parse_json_value` built its `OwnedValue` via `serde_json::Value`'s own `Number` arm, which round-trips only through Rust's own `f64`/`i64` `Display`, discarding the original digit spelling before an `OwnedValue` even exists.

```
$ succinctly jq -n --argjson n 1.500 '$n'
1.5          # now: 1.500
$ succinctly jq -n --argjson n 1e100 '$n'
1000...000   # now: 1E+100 (matches real jq)
```

Fix: `parse_json_value` still validates strictly via `serde_json::from_str` first (unchanged — same error messages, same rejection of trailing garbage per #284), then discards that `Value` and reparses the same, now-known-valid text through this crate's own fidelity-preserving JSON semi-indexer (`JsonIndex::build` + `to_owned`) — the same path `evaluate_input` already uses for the program's primary input. Fixes both `--argjson` and `--jsonargs`, which share `parse_json_value`.

## Deliberately out of scope
- **`--slurpfile`/`--seq`** (`parse_json_stream`/`parse_json_seq`) have the identical gap but need per-value byte-span tracking within a concatenated JSON stream to fix the same way (`JsonIndex` handles one document per call, not a stream) — filed as #1093.
- **`succinctly yq`'s own `--argjson`** (a succinctly-only extension — real mikefarah/yq has no `--argjson` flag at all, confirmed via `yq --help`) is deliberately left unfixed. yq's main `--input-format json` path already discards JSON-sourced number fidelity on purpose (#978, `canonicalize_json_numbers`), and `--argjson`'s current output already matches that established convention — extending fidelity there would make `--argjson` *inconsistent* with yq's own JSON-input behavior, not fix a divergence from a real oracle, since none exists for this flag. This is a premise correction to the issue's own "Root cause" section, which pointed at `yq_runner.rs` without checking whether real yq even has this flag.
- **Leading-zero numbers** (`--argjson n '007'`) are rejected by succinctly (strict RFC 8259 via `serde_json`) but accepted leniently by real jq — a separate, pre-existing validation-strictness gap found incidentally while testing, filed as #1094.

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` (full suite, all green)
- [x] `cargo fmt --check`
- [x] Every new/updated test's expected output verified against the pinned real `jq` binary; `yq`'s unchanged output also verified

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -1453,10 +1453,20 @@ fn find_number_end(bytes: &[u8], pos: usize) -> Option<usize> {
 /// Validates strictly via `serde_json::from_str` first -- not for its
 /// resulting `Value` (discarded), but for its error message and its
 /// rejection of trailing garbage (`42 garbage`) that `JsonIndex`'s own
-/// semi-indexing wouldn't catch on its own -- then reparses the same,
-/// now-known-valid text through this crate's own fidelity-preserving JSON
-/// semi-indexer (the same one the primary input path already uses, see
-/// `evaluate_input` above).
+/// semi-indexing wouldn't catch on its own. Deliberately parses to
+/// `serde_json::Value` here rather than the cheaper `serde::de::IgnoredAny`
+/// (which drives the identical grammar/trailing-content check without
+/// allocating a parse tree this function then discards): `IgnoredAny`
+/// doesn't range-check numbers at all, so a magnitude-overflowing literal
+/// (`1e400`) that `Value` correctly rejects ("number out of range") would
+/// instead silently reach `json_bytes_to_owned_value` and materialize as
+/// `null` -- trading a clear error for silent data loss, a worse outcome
+/// than the allocation this would save (#1095 review).
+///
+/// Then reparses the same, now-known-valid text through this crate's own
+/// fidelity-preserving JSON semi-indexer (the same one the primary input
+/// path already uses, see `evaluate_input` above) via the shared
+/// `json_bytes_to_owned_value`.
 fn parse_json_value(s: &str) -> Result<OwnedValue> {
     let s = s.trim();
     if s.is_empty() {
@@ -1465,13 +1475,17 @@ fn parse_json_value(s: &str) -> Result<OwnedValue> {
 
     serde_json::from_str::<serde_json::Value>(s).with_context(|| format!("Invalid JSON: {s}"))?;
 
-    let bytes = s.as_bytes();
-    let index = JsonIndex::build(bytes);
-    let cursor = index.root(bytes);
-    Ok(generic_to_owned(&cursor.value()))
+    Ok(crate::output::json_bytes_to_owned_value(s.as_bytes()))
 }
 
-/// Parse a JSON stream (multiple JSON values) from a string.
+/// Parse a JSON stream (multiple JSON values) from a string (`--slurpfile`).
+///
+/// Still loses number-literal source fidelity the way `parse_json_value`
+/// did before #1058 -- `find_json_values` (below) already splits a
+/// multi-value buffer into byte spans for the main lazy-input path and
+/// `serde_json::Deserializer::byte_offset()` could delimit each value here
+/// the same way, but wiring that up is deliberately deferred to #1093 to
+/// keep #1058/#1095 scoped to `--argjson`/`--jsonargs`.
 fn parse_json_stream(s: &str) -> Result<Vec<OwnedValue>> {
     let s = s.trim();
     if s.is_empty() {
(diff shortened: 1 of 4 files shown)


newhoggy:
Code review (10 finder angles) found several real gaps and one attempted improvement that turned out to be a real regression. All resolved:

**Fixed:**
- Extracted `json_bytes_to_owned_value` (`output.rs`) — the shared "JSON bytes → fidelity-preserving `OwnedValue`" sequence that was independently hand-duplicated in `jq_runner.rs`'s `parse_json_value` and `yq_runner.rs`'s pre-existing `parse_input`. Both now share it.
- Tightened `yq_runner.rs`'s `parse_json_value` doc comment — its "matching jq's `--argjson`" claim was true for validation *strategy* but read as a broader behavioral claim once the two functions diverge on fidelity (one review angle read it exactly that way). Now explicit that the divergence is deliberate, matching yq's own established #978 convention (real mikefarah/yq has no `--argjson` at all, so there's no oracle to match fidelity against).
- Corrected #1093's own scope via an issue comment: its "needs a new value-boundary mechanism" framing was wrong — `find_json_values`/`StreamDeserializer::byte_offset` already exist in this file. Added source pointers at `parse_json_stream`/`parse_json_seq` so the deferral is visible from the code itself, not just a PR description.

**Tried and reverted (real regression caught before merge, not after):** review suggested swapping the validation-only `serde_json::Value` parse for the cheaper `serde::de::IgnoredAny` (avoids building a parse tree only to discard it). Implemented it, then live-tested — `IgnoredAny` doesn't range-check numbers, so `1e400` (which `Value` correctly rejects as "number out of range") would instead pass validation and silently materialize as `null` through the fidelity-preserving path. A wasted allocation is a better trade than a clear error turning into silent data loss, so reverted and documented in place.

**Filed as follow-ups** (both confirmed pre-existing via the unmodified `main` binary — unrelated to and unaffected by this fix, not something to block on):
- #1098 — `generic_to_owned`'s decode-failure-to-null fallback can theoretically swallow an error silently (no live repro found; every case tried is already caught by an earlier, stricter layer).
- #1099 — extreme-magnitude float literals mis-render in the shared number formatter: `1e-400` → `0E-400` (wrong mantissa, not just underflowed value) and high-precision mantissas lose digits beyond ~15-17 significant figures.

All local verification re-run clean after the fixes: full `--no-fail-fast` test suite, 3x clippy matrix, fmt, no_std check. One clean rebase (main advanced with unrelated #1089, `jq/eval.rs` fold_index_key work).

Issue #1043, opened 2026-08-16
jq: several empty-collapse sites lack a 0 => None arm, wrongly erroring instead of matching real jq's silent-empty behavior


newhoggy:
## Summary

Surfaced during code review of #1038/PR #1042 (which added `borrowed_vec_to_result`, the `Vec<StandardJson>` sibling of `owned_vec_to_result` — both centralize a `0 => None, 1 => One/Owned, _ => Many/ManyOwned` collapse). The review's reuse/altitude passes, auditing whether that PR's stated sweep ("three superficially similar sites are structurally different, not duplicates") was accurate, found that those three excluded sites — plus a fourth, byte-identical twin — share the *same underlying defect* the new helpers exist to fix: each is missing the `0 => None` arm entirely, so a genuinely empty result collapses to `Many(vec![])`/`ManyOwned(vec![])` (or, in one case, a constructed empty array) instead of `None`.

Downstream, a scalar-context consumer (`result_to_owned`) apparently treats a non-`None` empty collection as an error ("empty result"/"no value") rather than silently propagating no output — real jq's actual behavior for all of these.

## Confirmed repros (verified against pinned jq 1.7.1; all exit 0 with no output in real jq)

**`eval_index_expr`'s `Targets::Borrowed`/`Owned` collapses (`src/jq/eval.rs` ~9110-9138):**
```
$ echo 5 | succinctly jq 'contains(.[("a","b")]?)'
jq: error (at <stdin>:1): empty result
(exit 5 — wrong)
$ echo 5 | jq 'contains(.[("a","b")]?)'
(no output, exit 0)
```

**`eval_slice_expr`'s final collapse (`src/jq/eval.rs` ~9252, `match out.len() { 1 => Owned, _ => ManyOwned(out) }`):**
```
$ echo null | succinctly jq 'contains(({"a":1})[0:(1,2)]?)'
jq: error (at <stdin>:1): empty result
(exit 5 — wrong)
$ echo null | jq 'contains(({"a":1})[0:(1,2)]?)'
(no output, exit 0)
```
(`[({"a":1})[0:(1,2)]?]` shows `[]` in both implementations, confirming the underlying stream is genuinely empty in both — only the scalar-context consumer mishandles it.)

**`eval_comma`'s tail match (`src/jq/eval.rs` ~850) and its byte-identical twin in `eval_fanout` (~1504):**
```
$ echo 1 | succinctly jq 'contains((empty,empty))'
[...]

Issue #1038, opened 2026-08-16
jq: retain_truthy/eval_limit's borrowed-StandardJson Many/One arms need their own owned_vec_to_result sibling


newhoggy:
## Context

#1029 (PR #1037) consolidated ~15 hand-rolled `Vec<OwnedValue>` → `None`/`Owned`/`ManyOwned` collapses into calls to the existing `owned_vec_to_result` helper. Code review on that PR found the exact same shape recurring for a *different* type pair that `owned_vec_to_result` can't handle:

- `retain_truthy`'s `QueryResult::Many` arm (`src/jq/eval.rs:1318-1324`) hand-rolls `match vs.len() { 0 => None, 1 => One(...), _ => Many(vs) }` on `Vec<StandardJson>` — sitting directly beside the `ManyOwned` arm (lines 1326-1329) that #1029 *did* convert.
- `eval_limit`'s `QueryResult::Many` arm (`src/jq/eval.rs:~14015-14023`) has the identical shape, also sitting beside its already-converted `ManyOwned` sibling.

Both are `0 => None, 1 => One(v), _ => Many(vs)` on `Vec<StandardJson<'_, W>>` — the borrowed-cursor counterpart to what `owned_vec_to_result` already does for `Vec<OwnedValue>`.

## Suggested fix

Add a sibling helper (or genericize `owned_vec_to_result` over both, if a clean shared signature exists — `StandardJson`'s extra lifetime/generic parameters may make a single generic function awkward, worth checking before committing to that shape):

```rust
fn borrowed_vec_to_result<'a, W>(mut vs: Vec<StandardJson<'a, W>>) -> QueryResult<'a, W> {
    match vs.len() {
        0 => QueryResult::None,
        1 => QueryResult::One(vs.pop().unwrap()),
        _ => QueryResult::Many(vs),
    }
}
```

Then route `retain_truthy`'s and `eval_limit`'s `Many` arms (and any other exact-shape sites found during implementation — a broader sweep, mirroring #1029's own methodology, is worth doing rather than just fixing the two named here) through it.

## Scope note

Deliberately not folded into #1029/PR #1037 itself, since introducing a new helper for a different type family is a larger design step than routing existing call sites through an already-existing one — the review that found this explicitly flagged it as "the real fix belongs at a `Vec<T>`-generic (or dual-helper) level," which needs its own consideration rather than a same-PR bolt-on.

newhoggy:
[...]

PR #1110, opened 2026-08-17
refactor(jq): share the Vec-to-{None,One,Many} collapse across 4 sites

### LINKED ISSUE #1067: jq: 4 near-identical Vec-to-{None,One,Many} collapse helpers, same bug class recurred 3x (#1038/#1043/#1048)
## Summary

Found during #1048's code review (two independent finder angles, Reuse and Altitude). The "collapse a `Vec<T>` accumulator to `{0 => None, 1 => One(t), _ => Many(vec)}`" pattern now has 4 independently-maintained copies:

- `eval.rs`'s `owned_vec_to_result` (~line 1126)
- `eval.rs`'s `borrowed_vec_to_result` (~line 1138)
- `eval_generic.rs`'s `owned_vec_to_generic_result` (line 950)
- `eval_generic.rs`'s `cursor_vec_to_generic_result` (line 962, added by #1048)

All four implement the identical 3-arm length match, differing only in which enum variant (`QueryResult`/`GenericResult`) and which element type (`OwnedValue`/borrowed value/`V::Cursor`) gets plugged in.

## Why this matters

This exact bug class — a hand-rolled `match v.len() { 1 => One, _ => Many(v) }` missing the `0 => None` arm, silently collapsing an optional (`?`)-suppressed zero-result computed index/slice to an empty `Many`/`ManyOwned`/`ManyCursor` instead of `None` — has now been independently rediscovered and fixed **three times** in this codebase: originally in #1038, again for 5 more sites in #1043, and again for `eval_generic.rs`'s 3 sites in #1048. Each occurrence needed its own from-scratch investigation and live-oracle verification. A future call site anywhere in `eval.rs`/`eval_generic.rs` that hand-rolls this match instead of calling one of the four existing helpers will silently reintroduce the identical bug, with nothing but code review to catch it (as it has, three times).

## Suggested approach

Not yet investigated in detail, but a few directions worth considering:
- A shared trait (e.g. `impl<T> Collapsible for Vec<T>` with associated `None`/`One`/`Many` constructors supplied per call site) that all 4 helpers implement in terms of, so the *algorithm* has one definition even though the 4 concrete signatures stay distinct.
- A macro that generates each helper's body from a 3-line invocation, keeping the duplication mechanical/obviously-in-sync rather than eyeballed.
- At minimum, a shared doc-comment/test convention across all 4 (a parameterized "does this collapse correctly at 0/1/2+" test run against each) so a change to one's semantics can't drift from the others unnoticed.

`QueryResult`/`GenericResult` are different enums with different variant sets (`GenericResult` additionally has `OneCursor`/`ManyCursor` for the borrowed-cursor case `QueryResult` doesn't need), so full unification may not be achievable without either a shared marker trait or accepting some remaining duplication — needs scoping before implementation.

newhoggy:
## Summary

`owned_vec_to_result`/`borrowed_vec_to_result` (`src/jq/eval.rs`) and `owned_vec_to_generic_result`/`cursor_vec_to_generic_result` (`src/jq/eval_generic.rs`) each hand-rolled the identical 3-arm length match (`0 => None, 1 => the single element, _ => the whole Vec`), differing only in which enum variant (`QueryResult`/`GenericResult`) and element type (`OwnedValue`/`StandardJson`/`V::Cursor`) got plugged in.

This exact bug class — a hand-rolled match missing the `0 => None` arm, silently collapsing an optional (`?`)-suppressed zero-result computed index/slice into an empty `Many`/`ManyOwned`/`ManyCursor` instead of `None` — was independently rediscovered and fixed **three times** in this codebase (#1038, #1043, #1048), each needing its own from-scratch investigation and oracle verification.

Extracts `collapse_vec<T, R>`, generic over both the element type and the result enum via three constructor closures (`none`/`one`/`many`). Each of the 4 existing helpers becomes a thin wrapper plugging in its own constructors — every existing call site and function signature stays unchanged, this is a pure internal refactor with no behavior change (the actual collapse algorithm is identical to what each helper already did).

`QueryResult`/`GenericResult` are different enums with different variant sets (`GenericResult` additionally has `OneCursor`/`ManyCursor`), so full unification into one type wasn't attempted — closures over the constructors sidestep that without needing a shared trait or a macro.

Fixes #1067

## Test plan

- [x] `test_collapse_vec_1067` — new unit test pinning `collapse_vec`'s own 0/1/2+ shape directly (empty → `none`, one element → `one` receiving the element itself not a 1-vec, 2+ → `many` receiving the untouched `Vec`), independent of any particular caller's enum. A future regression here now surfaces in one place instead of silently affecting all 4 call sites.
- [x] Full lib unit tests (2532 passed, up 1 from the new test).
- [x] `jq_cli_tests` (555), `yq_cli_tests` (702), `yq_golden_tests`/`yq_path_consistency_tests` (6), plus every other jq-specific integration suite (`jq_golden_tests`, `jq_evaluator_parity_tests`, `jq_numeric_equality_tests`, `jq_computed_key_tests`, `jq_containment_tests`, `jq_error_message_tests`, `jq_recurse_depth_tests`, `jq_tests` — 247 more) — all pass, each run in isolation.

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -16961,12 +16961,16 @@ fn builtin_tostream<W: Clone + AsRef<[u64]>>(
     let mut events = Vec::new();
     collect_tostream_events(&owned, &[], &mut events);
     // Always non-empty: every value (including an empty container or a
-    // top-level scalar) produces at least one leaf event.
-    if events.len() == 1 {
-        QueryResult::Owned(events.pop().unwrap())
-    } else {
-        QueryResult::ManyOwned(events)
-    }
+    // top-level scalar) produces at least one leaf event -- routed through
+    // `collapse_vec` anyway (#1067) so a future change to
+    // `collect_tostream_events` that ever violates that invariant degrades
+    // to `QueryResult::None` instead of silently returning `ManyOwned(vec![])`.
+    collapse_vec(
+        events,
+        || QueryResult::None,
+        QueryResult::Owned,
+        QueryResult::ManyOwned,
+    )
 }
 
 /// Builtin: fromstream(f) - reconstruct values from a stream of tostream-style events
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -1151,27 +1151,55 @@ fn merge_owned<W: Clone + AsRef<[u64]>>(
     merged
 }
 
+/// Collapse a `Vec<T>` accumulator into the smallest shape representing it:
+/// empty calls `none`, exactly one calls `one` with that value, more than one
+/// calls `many` with the whole vec. The one definition every
+/// `Vec<T>`-to-`{None,One(T)/Owned(T),Many(Vec<T>)}` collapse in this module
+/// (and [`eval_generic`](super::eval_generic)) routes through, instead of
+/// hand-rolling the same 3-arm length match at each call site -- that
+/// pattern independently drifted the same way three times before
+/// consolidation (#1038, #1043, #1048): a hand-rolled match missing the
+/// `0 => None` arm silently collapses an optional (`?`)-suppressed
+/// zero-result computed index/slice into an empty `Many`/`ManyOwned`/
+/// `ManyCursor` instead of `None`. A shared definition can't drift per call
+/// site; a fresh call site gets the right shape for free instead of risking
+/// a fourth rediscovery of the same bug.
+pub(crate) fn collapse_vec<T, R>(
+    mut items: Vec<T>,
+    none: impl FnOnce() -> R,
+    one: impl FnOnce(T) -> R,
+    many: impl FnOnce(Vec<T>) -> R,
+) -> R {
+    match items.len() {
+        0 => none(),
+        1 => one(items.pop().expect("len() == 1 guarantees a first element")),
+        _ => many(items),
+    }
+}
+
 /// Normalize a `Vec<OwnedValue>` accumulator into the smallest `QueryResult`
 /// shape that represents it: empty stays `None`, one value collapses to
 /// `Owned`, more than one stays `ManyOwned`.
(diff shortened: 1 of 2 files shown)


newhoggy:
Code review found one real gap and fixed it before merge: the PR's own \`collapse_vec\` doc comment claimed "every \`Vec<T>\`-to-\`{None,One,Many}\` collapse in this module (and eval_generic) routes through" it, but two more hand-rolled instances existed — \`builtin_tostream\` (\`eval.rs\`) and \`Expr::Optional\`'s \`Partial\`-recovery arm (\`eval_generic.rs\`). Both are currently guarded by a documented "never empty" invariant, so neither was a live bug, but both are the identical drift-prone pattern this issue exists to eliminate — a future change violating either invariant would now degrade to \`None\` instead of silently returning an empty \`Many\`/\`ManyOwned\`. Routed both through \`collapse_vec\` too, making the doc comment's claim accurate (confirmed via an exhaustive re-sweep for any remaining \`.len()\`-adjacent-to-\`None\` pattern — none found).

Also corrected a small inaccuracy in my own fix: my first pass claimed the new \`eval_generic.rs\` arm "mirrors \`eval::eval_try\`'s own use of the shared collapse," but \`eval_try\` doesn't call \`collapse_vec\`/\`owned_vec_to_result\` directly — it delegates to \`eval::prepend\`, which does. Fixed the comment to name the actual mechanism.

Re-verified everything after both the fix and a 9-commit rebase onto main (unrelated area — unary minus / slice-scalar rules — one real conflict in the shared import list, resolved by keeping both an unrelated rename from main and this PR's own \`collapse_vec\` addition): full lib unit tests (2537 passed, incl. all 7 \`tostream\`-specific tests), \`jq_cli_tests\` (556), \`yq_cli_tests\` (708), every other jq-specific integration suite, all 3 clippy configs, fmt, no_std check, doc build — all green.

PR #1111, opened 2026-08-17
fix(yaml): render a deferred-absent value bare, not as an empty-string marker

### LINKED ISSUE #1077: yq: an anchor's own null value renders as `&anc ""` instead of bare `&anc`

newhoggy:
## Summary
- A mapping field or sequence item deferred to a following line that never materializes any content (EOF right after, or a sibling immediately follows at the same/lower indent) streamed as an explicit `""` marker instead of real yq's bare rendering.
- Example: `a: &anc ""` where real `yq` writes `a: &anc`; `- &anc ""` where real `yq` writes `- &anc`.
- The gap predates #784 and isn't limited to anchored values — it's a general deferred-absent-value formatting bug in `stream_yaml_value`. #784's own tests had to pin the buggy `""` output for this shape since it hadn't been fixed yet; this PR updates those 5 tests to the correct expected output.

## Fix
Both the mapping-field and sequence-item streaming branches in `src/yaml/light.rs` now use the existing `is_deferred_value_absent` helper to skip the value token (and its leading space) entirely when there's nothing to write, while still emitting an anchor if one is present.

Fixes #1077

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` (full suite green; one unrelated flaky test confirmed passing in isolation)
- [x] Manually verified against the pinned real `yq` binary, byte-exact
- [x] Patch coverage checked via `omni-dev coverage diff`

### DIFF
--- a/src/yaml/light.rs
+++ b/src/yaml/light.rs
@@ -1728,35 +1728,18 @@ impl<'a, W: AsRef<[u64]>> YamlCursor<'a, W> {
                             // nothing at all writes no value token here
                             // (matching the sibling branch above, #765) --
                             // but unlike that branch, this one can still
-                            // have an anchor to write, so it can't just
-                            // skip straight to the comment. A trailing
-                            // space is only owed to something that actually
-                            // follows it: the colon alone (`a:`, no anchor,
-                            // absent value) takes none at all, matching
-                            // real yq's own bare rendering byte-for-byte.
-                            let absent = is_deferred_value_absent(&value);
-                            let anchor = value.anchor();
-                            if anchor.is_some() || !absent {
-                                out.write_char(' ')?;
-                            }
-                            if let Some(anchor) = anchor {
-                                out.write_char('&')?;
-                                out.write_str(anchor)?;
-                                if !absent {
-                                    out.write_char(' ')?;
-                                }
-                            }
-                            if !absent {
-                                let child_indent = deeper_yaml_indent(indent, indent_spaces, unit);
-                                value.stream_yaml_value(
-                                    out,
-                                    &child_indent,
-                                    indent_spaces,
-                                    unit,
-                                    sort_keys,
-                                    false,
-                                )?;
-                            }
+                            // have an anchor/tag to write, so it can't just
+                            // skip straight to the comment. See
+                            // `write_deferred_value`'s own doc comment for
+                            // the byte-for-byte spacing rule.
+                            write_deferred_value(
+                                out,
+                                &value,
+                                indent,
+                                indent_spaces,
+                                unit,
+                                sort_keys,
+                            )?;
                             // The value's own comment takes priority; fall
                             // back to the key's own comment when the value
                             // has none - covers an explicit key's trailing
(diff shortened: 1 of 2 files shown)


newhoggy:
Ran the mandatory multi-agent review before merge. It surfaced a real, serious regression in the first draft: an explicit tag on a deferred-absent value (`a: !!str` with nothing after it) was silently dropped instead of preserved — worse than the bug #1077 itself reports, since the pre-fix code at least kept the tag alongside the buggy `""` marker.

Fixed in the latest commit: extracted the near-duplicated mapping-field/sequence-item logic into one `write_deferred_value` helper that accounts for anchor, tag, and absence together (review also flagged the duplication itself as a maintenance risk — a future fix landing in only one of the two call sites). Re-verified all anchor/tag/absent combinations byte-for-byte against the pinned real `yq` binary.

Review also surfaced three out-of-scope findings, filed as follow-ups rather than expanding this PR's scope:
- #1113 — pre-existing bug (not introduced here): an explicit key's own comment can drop an anchor on a deferred-absent value
- #1114 — possible double-resolve efficiency concern in the new absence-check path; flagged as needing real profiling before deciding whether it's worth fixing, not just doc-comment pattern-matching
- #1115 — #1077's fix doesn't extend to flow-style or `--slurp` rendering paths (neither has a live, exploitable regression; one has a doc comment that now overclaims parity)

Patch coverage: 91.67% (22/24 new lines); the 2 uncovered lines are an infallible-`Write` error-propagation branch and a diff-attribution artifact on an unrelated pre-existing match arm — not real gaps.

PR #1118, opened 2026-08-17
chore(build): untrack orphaned 1.yml fixture

### LINKED ISSUE #1086: test suite: full cargo test run corrupted repo-root fixture 1.yml
**Severity: Low**

## Summary

While running the full local test suite (`cargo test --features cli,simd,regex,serde`) in a feature-branch worktree, the tracked repo-root fixture `1.yml` was silently overwritten from its committed content to `a: 1.0` (no trailing newline). Caught by `git status`/`git diff` before committing, and reverted — this was not something my own changes wrote (my branch's diff never touches this file, and none of my own test commands targeted a relative `1.yml` path).

## Repro

Not yet reliably reproduced in isolation — observed once, during a full-suite run in `$HOME/wrk/work-trees/succinctly/issue-949-yq-computed-whole-float-decimal-point`. `git diff 1.yml` at the time showed:

```diff
--- a/1.yml
+++ b/1.yml
@@ -1,2 +1 @@
----
-name: Charlie
+a: 1.0
```

## Investigation so far

- `1.yml` was added in `f16c1262e` ("feat(cli): add --slurp option to yq command") as a fixture for multi-document YAML slurp testing.
- No test in `tests/yq_cli_tests.rs` currently references `1.yml` by that literal name (searched exhaustively) — only `f1.yml` inside a `TempDir` (e.g. `dir.path().join("f1.yml")`), suggesting whatever test originally read `1.yml` was refactored to use a temp directory at some point, leaving the fixture file orphaned but still tracked.
- Every `--split-exp` test that constructs an `$index`-based output filename (e.g. `"f" + ($index|tostring) + ".yml"`) correctly prefixes with a `TempDir`'s path (`dir.path().display()`) — none appear to write to a bare relative path.
- No literal string `"1.yml"` appears anywhere else in the repo (searched `*.rs`, `*.sh`, `*.md`).
- The corrupted content (`a: 1.0`, no trailing newline) doesn't obviously match any single test's fixture I could find, though it's a common enough value that this may be coincidental.
- Not yet checked: benchmark/data-generation code paths (`succinctly json generate`/`dsv generate`/bench runners) that might write output files with generic numeric names into the current working directory when run without an explicit `-o`, if any test or doctest invokes one of those without redirecting output.

## Why this matters

A tracked fixture silently changing content on a full test run is a correctness hazard for the test suite itself (a future test relying on `1.yml`'s original content would fail non-deterministically depending on run order/what ran before it), and risks an accidental commit of corrupted repo state if `git status` isn't checked before committing after a full suite run.

## Suggested next steps

- `git bisect`/binary-search which specific test (or test binary) writes to a relative `1.yml`/similar path, possibly by running subsets of the test suite and checking `git status` after each.
- Once found, fix it to use a `TempDir`-scoped path like its neighbors.
- Consider whether `1.yml` is still needed at all now that no test appears to read it directly, or whether it's safe to remove if genuinely orphaned.

newhoggy:
## Summary

`1.yml` was added in `f16c1262e` for `--slurp` multi-document testing, but the current tests embed that same content (Alice/Bob/Charlie) as inline byte-string literals instead of reading the file — nothing in the test suite references it.

As a tracked root fixture it's exposed to the same failure mode `0da3b8997` already fixed for the sibling `.yml` sample one commit before this issue was filed: an ad hoc `succinctly yq -i` command run against it during manual/interactive testing silently corrupts the tracked file.

Confirmed via an exhaustive automated sweep (doctests, CLI generate-command defaults, `--split-exp`, `-i`/`--inplace`, benches, examples, every raw `fs::write`/`File::create` in `src/`) that **no `cargo test` invocation writes to it** — every test that constructs a numbered/indexed filename prefixes it with a `TempDir` path, every CLI generate command requires an explicit `-o`, and grepping the entire tree for the literal string `1.yml` turns up nothing outside unrelated `f1.yml`-inside-TempDir fixtures. The corruption observed in the issue is most plausibly a manual command interleaved with a test run in the same terminal session, not the test suite itself.

## Fix

Removes the file and adds `/1.yml` to `.gitignore` alongside the existing `/.yml` entry, closing the same hole for this second fixture. `CLAUDE.md` already directs manual jq/yq CLI testing to `.ai/scratch/` (added by `0da3b8997`), so no further doc change is needed.

Fixes #1086

## Test plan

- [x] Confirmed nothing in `src/`, `tests/`, `benches/`, `examples/` references `1.yml` (grep across all four).
- [x] `cargo build --features cli` — clean.
- [x] Full lib unit tests (2537 passed), `jq_cli_tests` (556), `yq_cli_tests` (708), `corpus_stats_cli_tests`/`yq_golden_tests`/`yq_path_consistency_tests` — all pass unaffected, as expected for a fixture nothing reads.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,10 +25,15 @@ target
 # AI scratch directory (temporary files for AI assistants)
 .ai/
 
-# Manual yq/jq CLI testing fixture — untracked so ad hoc testing (e.g. `-i`
+# Manual yq/jq CLI testing fixtures — untracked so ad hoc testing (e.g. `-i`
 # inplace edits) never dirties the tree; use .ai/scratch/ for this instead.
-/.yml
-/1.yml
+# Root-anchored only (doesn't touch tests/data/, .github/workflows/, etc.).
+# Started as two specific filenames (/.yml, /1.yml) after each was hit by
+# this exact failure mode in turn (0da3b8997, #1086) -- generalized after
+# the second incident, since `git ls-files` confirms zero tracked root-level
+# .yml/.yaml files exist, so this costs nothing today.
+/*.yml
+/*.yaml
 
 # Generated benchmark data
 data/bench/generated/
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ target
 # Manual yq/jq CLI testing fixture — untracked so ad hoc testing (e.g. `-i`
 # inplace edits) never dirties the tree; use .ai/scratch/ for this instead.
 /.yml
+/1.yml
 
 # Generated benchmark data
 data/bench/generated/
--- a/1.yml
+++ b/1.yml
@@ -1,2 +1,0 @@
----
-name: Charlie


newhoggy:
Code review found one reasonable improvement, applied before merge: blocklisting specific filenames (\`/.yml\`, \`/1.yml\`) after each was independently hit by the identical failure mode is whack-a-mole — this is the second incident in three days (\`0da3b8997\`, this issue). \`git ls-files\` confirms zero tracked root-level \`.yml\`/\`.yaml\` files exist today, so generalized to a root-anchored \`/*.yml\` + \`/*.yaml\` pattern instead, which costs nothing today and prevents a third recurrence — verified it doesn't touch \`tests/data/**/*.yaml\`, \`.github/workflows/*.yml\`, or \`.omni-dev/*.yaml\` (all in subdirectories, unaffected by a root-only glob).

Rebased cleanly onto main (2 unrelated commits, jq-mode infinity sentinel work). Re-verified: full lib unit tests (2539 passed), \`jq_cli_tests\` (557), \`yq_cli_tests\` (708), clippy, fmt — all green.

Issue #1120, opened 2026-08-17
@base64d silently truncates non-multiple-of-4-length input instead of erroring


newhoggy:
## Summary

`format_base64d` (`src/jq/eval.rs`) silently drops a trailing base64 chunk shorter than 4 characters instead of erroring, producing a truncated-but-successful result. Real jq and real yq (v4.53.3, live-verified) both error on this:

```
$ echo '"false"' | jq -c '@base64d'
jq: error (at <stdin>:1): "false" is not valid base64 data
```

```
$ printf 'a: 1.5\n' | yq -o=json '.a | @base64d'
Error: illegal base64 data at input byte 1
```

succinctly instead breaks out of its chunking loop as soon as `chunk.len() < 4`, discarding the trailing partial chunk and returning whatever was decoded so far:

```
$ echo '"false"' | succinctly jq -c '@base64d'
"}�l"
```

## Context

Discovered while implementing #1109 (`format_urid`/`format_base64d` yq-mode scalar stringification). #1109's fix stringifies scalars like `1.5` → `"1.5"` (length 3) before decoding, which now also reaches this pre-existing leniency bug — but the bug itself is unrelated to #1109 and was already reachable via any plain string input of non-multiple-of-4 length (`"false"` is 5 chars), on both `jq` and `yq` modes, before #1109's change. Confirmed live against the pinned `main` binary predating #1109.

## Acceptance criteria

- `format_base64d` errors (matching jq's `"<input>" is not valid base64 data` / yq's `illegal base64 data at input byte N` wording, or a close succinctly-appropriate equivalent) when the input length (after whitespace stripping) is not a multiple of 4, instead of silently truncating.
- Existing valid-base64 decode tests continue to pass.

newhoggy:
Claimed by agent `cadence`. Working on branch `issue-1120-base64d-truncation`.

Issue #1072, opened 2026-08-16
jq: yq's "container stringifies to empty" quirk now duplicated 3x (join's 2 helpers + sub/gsub's new one)


newhoggy:
## Summary

Found during #1052's code review. Real yq's own go-yaml-node-model quirk — "only scalar nodes carry a raw text `Value`, so a container (array/non-empty-object) contributes an empty string when coerced to text" — is now independently reimplemented 3 times:

- `yq_join_element_part` (`src/jq/eval.rs`, ~line 4105, from #1041)
- `yq_join_separator` (`src/jq/eval.rs`, ~line 4120, from #1041)
- `combine_sub_gap`'s new yq branch (`src/jq/eval.rs`, from #1052)

## Why this matters

A future fix to the underlying model (e.g. #1041's own doc comment already flags real yq's go-yaml node model as "presumably" the cause, not confirmed) has to be found and applied in 3 places, with no compiler signal if one is missed. Worse: the 3 existing implementations **already disagree with each other** on `Null` handling — `yq_join_element_part` maps `Null` to an empty string, `yq_join_separator` leaves `Null` falling through to `to_json_yq()` (the literal text `"null"`), and `combine_sub_gap` (per #1052, live-verified) also spells `Null` as `"null"` text, matching `yq_join_separator`'s choice, not `yq_join_element_part`'s.

There's also a pre-existing, unrelated divergence in *numeric* formatting between the two paths: `owned_to_string`'s `Float` arm (used by `combine_sub_gap`) goes through `numeric_display_string` → `number_str()` (drops a trailing `.0`), while `yq_join_element_part`/`yq_join_separator`'s `to_json_yq()` path goes through `format_float_with_fraction` (keeps `.0`, per #953) — so the two existing "sources of truth" for yq stringification already disagree with each other on a plain computed float, independent of this issue's container-quirk question.

## Suggested approach

Not yet scoped. A straight merge into one function isn't safe as-is (see the `Null` disagreement above) — needs a `null_becomes_empty: bool`-style parameter or two thin wrappers sharing one core, plus a decision about which of the two numeric-formatting behaviors (`number_str()` vs `format_float_with_fraction`) is actually correct for each call site before consolidating. Needs its own live-verification pass across all 3 call sites (join separator, join element, sub/gsub replacement) against real yq before implementing, not a mechanical dedup.

## Scope note

[...]

PR #1121, opened 2026-08-17
fix(jq): yq's @base64d/@urid stringify a scalar instead of erroring

### LINKED ISSUE #1109: yq: @base64d/@urid should silently return "" for a container, not error
## Summary

`format_base64d`/`format_urid` (`src/jq/eval.rs`) aren't generic over `EvalSemantics` at all, so they behave identically in `jq`/`yq` mode: both error on an array/object input. Real yq v4.53.3 diverges here — it silently returns an empty string for a container instead of erroring:

```
$ printf 'a: [1,2]\n' | yq -o=json '.a | @base64d'
""
$ printf 'a: {b: 1}\n' | yq -o=json '.a | @base64d'
""
$ printf 'a: [1,2]\n' | yq -o=json '.a | @urid'
""
```

```
$ printf 'a: [1,2]\n' | succinctly yq -o json '.a | @base64d'
Error: expected string, got array
```

## Context

This is the decode-direction counterpart of #1096 (fixed in PR #1107), which made `@uri`/`@base64`/`@html` JSON-stringify a container in jq mode instead of erroring. #1096's own scope note flagged the decode direction (`@base64d`/`@urid`) as "not yet checked" — this issue is that check, filed as a separate follow-up since it needs different work: making `format_base64d`/`format_urid` generic over `S: EvalSemantics` (they currently aren't parameterized at all) so yq mode can special-case a container to `""` while jq mode keeps its existing (and still correct) error-on-container behavior.

Note this is scoped to **containers only** — real yq's `@base64d`/`@urid` still genuinely error on a string that isn't valid encoded data, same as jq; the divergence is specifically "wrong input *type*" (array/object), not "malformed input of the right type" (a bad string).

## Acceptance criteria

- `succinctly yq '.a | @base64d'` on an array/object `.a` returns `""`, matching real yq v4.53.3, without touching jq mode's existing container-rejection behavior for `@base64d`.
- Same for `@urid`.
- jq mode unaffected: `format_base64d`/`format_urid` continue to error on a container in jq mode (no real-jq oracle claims otherwise — jq has no `@urid` at all, and `@base64d` on a container still errors in real jq, just via a different code path: stringify-then-fail-to-decode).

newhoggy:
## Summary

- `format_urid`/`format_base64d` (`src/jq/eval.rs`) are now generic over `S: EvalSemantics`. In yq mode, a non-string scalar (`Int`/`Float`/`NumberLiteral`/`Bool`/`Null`) is stringified before decoding, and a container (array/object) stringifies to `""` — matching real yq v4.53.3 (live-verified) and the same "container stringifies to empty" convention already established for `@sh`/`@uri`/`@html` in #1072. jq mode is unaffected: it keeps erroring on every non-string type, unchanged.
- `format_base64d` now uses lossy UTF-8 conversion on the decoded bytes instead of erroring on invalid UTF-8 — confirmed live that both real jq 1.7.1 and real yq v4.53.3 never error here, they substitute the replacement character.

## Out of scope (filed as follow-up)

While verifying this fix I confirmed `format_base64d` also silently truncates a trailing base64 chunk shorter than 4 characters instead of erroring (e.g. `"false" | @base64d` gives garbage instead of an error), which both real jq and real yq do error on. This is a pre-existing bug, reachable via plain string input on both jq and yq mode before this PR — filed separately as #1120.

## Test plan

- [x] New regression tests in `tests/yq_cli_tests.rs`: scalar stringification for `@urid`/`@base64d`, container → `""`, jq mode still rejects non-strings, lossy-UTF8 decode.
- [x] `cargo test --features cli,simd,regex,serde` (full suite, 714 passed)
- [x] Live-verified against real yq v4.53.3 and real jq 1.7.1

Fixes #1109

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -5453,10 +5453,10 @@ pub(crate) fn format_owned<S: EvalSemantics>(
         FormatType::Tsv => format_tsv::<S>(owned, optional),
         FormatType::Dsv(delimiter) => format_dsv::<S>(owned, delimiter, optional),
         FormatType::Base64 => format_base64::<S>(owned, optional),
-        FormatType::Base64d => format_base64d(owned, optional),
+        FormatType::Base64d => format_base64d::<S>(owned, optional),
         FormatType::Html => format_html::<S>(owned, optional),
         FormatType::Sh => format_sh::<S>(owned, optional),
-        FormatType::Urid => format_urid(owned, optional),
+        FormatType::Urid => format_urid::<S>(owned, optional),
         FormatType::Yaml => format_yaml(owned),
         FormatType::Props => format_props(owned),
     }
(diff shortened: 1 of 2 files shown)


newhoggy:
Retriggering CI checks — the push at 9219f20c never fired a pull_request event.

PR #1127, opened 2026-08-17
fix(yaml): apply yq's scientific-notation threshold to a computed float's tostring/interpolation spelling

### LINKED ISSUE #1054: yq: tostring/@json on a computed whole-number float skips the scientific-notation threshold
**Severity: Low**

## Summary

Real yq applies its scientific-notation magnitude threshold (roughly: exponent outside `-4..6`) when formatting a *genuinely computed* float, but succinctly's `tostring`/`@json`/string-interpolation paths currently give a plain decimal-with-`.0` spelling instead, regardless of magnitude.

## Repro

```
$ printf 'a: 1\n' | succinctly yq -r '(1e10 * 2) | tostring'
20000000000.0
$ printf 'a: 1\n' | yq -r '(1e10 * 2) | tostring'   # pinned v4.53.3
2e+10
```

Confirmed **pre-existing**, not a regression: rebuilding the commit immediately before #953's fix landed already gives a different-but-still-wrong output for the same command (`20000000000`, no decimal point at all — the plain #953 symptom). #953's fix changed *which* wrong output this case produces, but did not introduce the underlying gap.

## Root cause (partial)

`format_float_yq` (`src/yaml/light.rs`) already implements the correct threshold logic (scientific past magnitude, `format_float_with_fraction` otherwise) and is already used correctly by the CLI's own `-o json` formatter (`src/bin/succinctly/output.rs`'s `format_json_impl`). But the *value-level* formatting paths — `OwnedValue::to_json_yq()`'s `Float` arm, and the `eval_generic.rs` reindex bridge's yq-mode fallback (`src/jq/value.rs`, both now use `format_float_with_fraction`) — do not apply this threshold, so a computed float reaching either path always gets a plain decimal spelling.

## Why this isn't a one-line swap

The reindex bridge specifically **needs** `format_float_with_fraction` (not `format_float_yq`) for #953's own repro: a plain YAML integer scalar too large for `i64` (`a: 99999999999999999999`) resolves to a bare `Float`, and real yq's own output for that value, reached via e.g. `[.a]`, is plain decimal *regardless of magnitude* (`100000000000000000000.0`, not `1e+20`) — confirmed live against the pinned oracle. Real yq is applying its scientific-notation threshold to *actually-computed* floats but not to a value that's merely an unmodified overflow-int scalar passed through unchanged, even though both end up as the same `OwnedValue::Float` internal representation with no way left to distinguish "computed" from "structurally repackaged, unmodified" by the time either code path sees it.

Fixing this properly likely needs either:
- Preserving an "origin" tag through more of the pipeline (e.g. extending `NumberLiteral` or adding a marker distinguishing "materialized without computation" from "arithmetic result"), or
- A narrower fix scoped to exactly the `tostring`/`@json`/string-interpolation call sites (which never touch the reindex bridge for a top-level scalar `tostring`), leaving the reindex bridge's fallback as-is.

## Where this was found

Surfaced during code review on #953's fix (PR #1053) as a live-reproduced divergence from real yq; deliberately not expanded into that issue's scope since it's a separate, pre-existing gap with a different (and more involved) root cause.

newhoggy:
## Summary
- `EXPR | tostring` and string interpolation on a genuinely computed float (e.g. `(1e10 * 2)`) always spelled it out as a plain decimal, never applying real yq's scientific-notation magnitude threshold (`2e+10`, not `20000000000.0`).
- Two distinct gaps, fixed separately:
  - `numeric_display_string` (`src/jq/eval.rs`) had no yq-mode branch for a genuinely computed `Float` at all. Fixes string interpolation on its own, since the embedded computation never leaves `OwnedValue::Float` before reaching this function.
  - `eval_on_owned` (`src/jq/eval_generic.rs`) — the reindex-bridge bypass used when a computed value needs a further pipe stage — already special-cased `Expr::Format` to skip the bridge, but not `Expr::Builtin(ToString)`. So `EXPR | tostring` still serialized through the bridge's decimal-only float formatter first, reparsed as a document-sourced-*looking* number, and echoed that baked text verbatim per #1008's literal-preservation rule. Fixed by giving `tostring` the same bypass `@text`/`@json` already have.

## Deliberately out of scope
- `@json`/`to_json_yq()`'s own bare-Float formatter is left unconverted: it's also the *only* formatter a document-sourced i64-overflow scalar reaches via plain `.field | @json` (no computation at all) once it degrades to a bare `Float`, and `OwnedValue::Float` carries no tag distinguishing that case from a genuinely computed one. Confirmed live: changing it regresses `.a | @json` on an overflow scalar (#953's own scope). Needs an origin-tracking mechanism this fix doesn't attempt — filed as a follow-up.
- `tostring` on a *container* holding computed floats surfaced a further, unrelated divergence (real yq's container-`tostring` doesn't even produce a JSON string in yq mode) — also filed as a follow-up, out of this issue's stated scope.

Fixes #1054

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` (full suite green; two unrelated flaky tests confirmed passing in isolation)
- [x] Verified byte-for-byte against the pinned real `yq` binary for the primary repro, threshold boundary, jq-mode non-regression, and #953's overflow-scalar cases
- [x] Patch coverage: 100% (7/7 new lines)

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -5333,6 +5333,18 @@ pub(crate) fn numeric_display_string<S: EvalSemantics>(value: &OwnedValue) -> St
         if f.is_nan() || f.is_infinite() {
             return nonfinite_display_string::<S>(*f).to_string();
         }
+        // A genuinely computed float (never captured as a `NumberLiteral`
+        // above -- e.g. `(1e10 * 2)`) applies real yq's scientific-notation
+        // magnitude threshold when stringified, unlike `-o json`/structured
+        // YAML output's `to_json_yq`/reindex-bridge formatting, which must
+        // keep a decimal spelling at any magnitude instead (#953) -- these
+        // are different formatting rules for different call sites, not a
+        // contradiction: confirmed live, `(1e10*2) | tostring` gives
+        // `"2e+10"` while `[.a]`'s own array-wrapped output keeps
+        // `100000000000000000000.0` for an i64-overflow scalar (#1054).
+        if S::TAG == EvalTag::Yq {
+            return crate::yaml::format_float_yq(*f);
+        }
     }
     value.number_str().expect("numeric variant").into_owned()
 }
@@ -5415,7 +5427,7 @@ pub(crate) fn owned_value_to_json<S: EvalSemantics>(value: &OwnedValue) -> Strin
 }
 
 /// Convert an owned value to a string representation (for interpolation).
-fn owned_to_string<S: EvalSemantics>(value: &OwnedValue) -> String {
+pub(crate) fn owned_to_string<S: EvalSemantics>(value: &OwnedValue) -> String {
     match value {
         OwnedValue::Null => "null".to_string(),
         OwnedValue::Bool(true) => "true".to_string(),
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -5342,8 +5342,16 @@ pub(crate) fn numeric_display_string<S: EvalSemantics>(value: &OwnedValue) -> St
         // contradiction: confirmed live, `(1e10*2) | tostring` gives
         // `"2e+10"` while `[.a]`'s own array-wrapped output keeps
         // `100000000000000000000.0` for an i64-overflow scalar (#1054).
+        //
+        // `format_float_yq_yaml`, not `format_float_yq`: a stringified
+        // value is never re-parsed back into the document, so there's no
+        // type-ambiguity-on-round-trip concern to preserve a forced `.0`
+        // for -- `format_float_yq`'s ordinary-magnitude formatter
+        // (`format_float_with_fraction`) is for exactly that concern, and
+        // wrongly forces `.0` here: confirmed live, `(50.0*2) | tostring`
+        // is `"100"` in real yq, not `"100.0"` (caught by review).
         if S::TAG == EvalTag::Yq {
-            return crate::yaml::format_float_yq(*f);
+            return crate::yaml::format_float_yq_yaml(*f);
         }
     }
     value.number_str().expect("numeric variant").into_owned()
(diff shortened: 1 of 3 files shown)


newhoggy:
Ran the mandatory multi-agent review before merge — it found several real issues, all fixed in the latest commit:

1. **Genuine bug**: the yq-mode branch used `format_float_yq`, whose ordinary-magnitude formatter forces a trailing `.0` (correct for `-o json`/structured output, #953, but wrong for `tostring`, which never re-parses its output — no round-trip type-ambiguity to preserve). Real yq gives `"100"` for `(50.0*2) | tostring`, not `"100.0"`. Fixed by switching to `format_float_yq_yaml`, whose sibling formatter drops the point unconditionally, matching real yq exactly. My own regression test didn't catch this because its inputs stayed `OwnedValue::Int` and never reached the `Float` branch at all.

2. **Real coverage gap**: the fix only patched `eval_generic.rs`'s `eval_on_owned`, used by normal document-input mode. `-n`/`--input-format json` route through `eval.rs`'s own separate `eval_owned_fast_path` instead, which never received the same bypass — confirmed live, `succinctly yq -n -r '(1e10*2)|tostring'` still gave the pre-fix decimal spelling. Added the same bypass there.

3. **Misleading comments**: my own comments claimed `owned_to_string` was "the exact same function" two pre-existing cursor-based `tostring` handlers already called — false in both cases (independent hand-rolled copies). Deduplicated both to call `owned_to_string` directly instead of just softening the wording, per review's "reuse" angle.

**Deliberately left unfixed, documented in code comments and filed as #1134**: the bypass only matches `tostring` as the *immediately next* pipe stage. Any intervening stage (even a no-op `.`), `(tostring)`, `tostring?`, or `map(...|tostring)` all still reopen the original bug — three independent review passes converged on this same finding. A real fix needs either threading reindex-provenance through every intermediate call, or the same origin-tracking mechanism #1128 (the `@json` sibling gap) already needs — both substantially larger than this PR's scope.

Multiple reviewers also confirmed no other correctness regressions via differential testing against a fresh `origin/main` baseline binary (NaN/Infinity, duplicate keys, i64 boundaries, string escaping, depth guards, `path()`/`optional` interaction).

Patch coverage: 100% (4/4 new lines).

PR #1136, opened 2026-08-17
fix(jq): @urid corrupts multi-byte UTF-8; @base64d over-strips whitespace

### LINKED ISSUE #1123: @urid byte-as-char decode corrupts multi-byte UTF-8; @base64d strips non-ASCII Unicode whitespace
## Summary

Two related, pre-existing decode-loop correctness bugs found while implementing/reviewing #1109 (both loops are unchanged by that PR, just newly reachable through its scalar-stringification path in addition to their prior plain-string reachability):

**1. `format_urid`'s percent-decode loop corrupts any non-ASCII input**, even with zero `%` escapes present:

```
$ echo '"café"' | succinctly jq '@urid'
"cafÃ©"
```

Root cause: the loop pushes each raw byte individually via `bytes[i] as char` (`src/jq/eval.rs`, `format_urid`'s decode loop), which mis-encodes any byte ≥ 0x80 as its own Latin-1 codepoint instead of preserving the original UTF-8 multi-byte sequence. This corrupts both literal pass-through bytes and genuinely percent-decoded bytes (`"caf%C3%A9" | @urid` should be `"café"`, not `"cafÃ©"`).

**2. `format_base64d`'s whitespace-stripping uses `char::is_whitespace()`**, which strips non-ASCII Unicode whitespace (U+00A0, U+2028, U+3000, etc.) that real yq treats as invalid input:

```
# a base64 payload with an injected U+00A0 silently "succeeds" instead of erroring
```

Real yq v4.53.3 errors (`illegal base64 data at input byte N`) on such input; succinctly silently strips it and decodes successfully.

## Acceptance criteria

- `format_urid` decodes into a byte buffer and converts once via `String::from_utf8_lossy` (or equivalent), so non-ASCII pass-through and percent-decoded bytes round-trip correctly.
- `format_base64d`'s whitespace handling only strips ASCII whitespace (or matches real yq's exact accepted character set), rejecting non-ASCII whitespace as invalid.
- Add regression tests for both, including non-ASCII pass-through for `@urid`.

newhoggy:
## Summary

Two decode-loop correctness bugs in `src/jq/eval.rs`, found during code review of #1109/#1121, filed as #1123.

**1. `format_urid` corrupts any non-ASCII byte**, whether percent-decoded or simply passed through:

```
$ echo '"café"' | succinctly jq '@urid'
"cafÃ©"          # should be "café" -- no % escapes even present
$ echo '"caf%C3%A9"' | succinctly jq '@urid'
"cafÃ©"          # should be "café"
```

Root cause: the decode loop pushed each output byte via `bytes[i] as char`, which mis-encodes any byte >= 0x80 as its own standalone Latin-1 codepoint instead of a UTF-8 sequence member. Fixed by accumulating into a `Vec<u8>` and converting once via `String::from_utf8_lossy` at the end, so multi-byte sequences (pass-through or reassembled from `%XX` escapes) round-trip intact.

**2. `format_base64d` silently discards non-ASCII Unicode whitespace** (U+00A0, U+2028, U+3000, ...) that real yq treats as invalid input:

```
$ printf '"aGVs bG8="' | succinctly jq '@base64d'
"hello"          # real yq errors: "illegal base64 data at input byte N"
```

Root cause: whitespace stripping used `char::is_whitespace()` (all Unicode whitespace) instead of ASCII-only. Fixed by switching to `char::is_ascii_whitespace()` — ordinary ASCII whitespace (space, tab, newline) is still stripped as before; non-ASCII whitespace now reaches the decoder as an invalid character and correctly errors.

Both functions are shared by jq and yq mode (`S: EvalSemantics`-generic), so both fixes apply identically to `succinctly jq` and `succinctly yq`.

## Test plan

- [x] New regression tests: `@urid` non-ASCII pass-through, percent-decode reassembly, and a mixed ASCII-escape/multi-byte-passthrough case; `@base64d` non-ASCII whitespace now errors, ASCII whitespace (space + literal newline) still strips correctly (regression guard).
- [x] Existing `#1109` scalar-stringification tests unaffected (11/11 pass in the targeted run).
- [x] Full `cargo test --features cli,simd,regex,serde` green.
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean.
- [x] `cargo fmt --check` clean.
- [x] `cargo check --no-default-features` (no_std) clean.

Fixes #1123

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -5629,8 +5629,16 @@ fn format_urid<S: EvalSemantics>(value: &OwnedValue, optional: bool) -> Result<S
         _ => return Err(EvalError::type_error("string", value.type_name())),
     };
 
-    let mut result = String::new();
     let bytes = s.as_bytes();
+    // Byte buffer, not `String` -- a decoded/pass-through byte is a UTF-8
+    // *sequence* member, not its own standalone codepoint; pushing each
+    // byte individually via `as char` (the pre-#1123 approach) mis-encoded
+    // any byte >= 0x80 as its own Latin-1 codepoint, corrupting every
+    // multi-byte character it touched (`"café" | @urid` produced
+    // `"cafÃ©"` even with zero `%` escapes present). Converting once at
+    // the end lets multi-byte sequences -- whether copied through
+    // verbatim or reassembled from `%XX` escapes -- round-trip intact.
+    let mut result = Vec::with_capacity(bytes.len());
     let mut i = 0;
     while i < bytes.len() {
         if bytes[i] == b'%' && i + 2 < bytes.len() {
@@ -5640,16 +5648,16 @@ fn format_urid<S: EvalSemantics>(value: &OwnedValue, optional: bool) -> Result<S
                 (bytes[i + 2] as char).to_digit(16),
             ) {
                 let decoded = (h1 * 16 + h2) as u8;
-                result.push(decoded as char);
+                result.push(decoded);
                 i += 3;
                 continue;
             }
         }
-        // Not a valid percent-encoded sequence, just copy the character
-        result.push(bytes[i] as char);
+        // Not a valid percent-encoded sequence, just copy the byte
+        result.push(bytes[i]);
         i += 1;
     }
-    Ok(result)
+    Ok(String::from_utf8_lossy(&result).into_owned())
 }
 
 /// Quote a CSV/DSV string field: wrap in `"..."`, doubling inner `"`.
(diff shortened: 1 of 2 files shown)


newhoggy:
Closing as a duplicate — #1137 addresses the same issue (#1123) and merged just now (both PRs were created within ~2.5 minutes of each other, a genuine race rather than one copying the other). #1137's version went through a full code-review pass and several rebases against a rapidly-advancing `main`, including reconciling with #1120's independent rewrite of the base64 decode algorithm along the way.

No offense intended to this PR's own approach — sorry for the wasted effort on a genuine timing collision. Thank you for the work either way.

Issue #1139, opened 2026-08-17
jq: object destructuring pattern shorthand {$a} (implicit key from var name) is not parsed


newhoggy:
**Severity: Low**

Found while implementing #720 (`?//` alternative patterns) — pre-existing, not introduced or worsened by that fix.

## Repro

Real jq accepts the object-destructuring-pattern shorthand `{$a}` as sugar for `{a: $a}` (the key is inferred from the variable name):

```
$ echo '{"a":1}' | jq '. as {$a} | $a'
1
```

succinctly rejects it outright, in both jq and yq mode, confirmed on `main`'s own existing binary (predates #720 entirely):

```
$ echo '{"a":1}' | succinctly jq '. as {$a} | $a'
jq: compile error: parse error at position 6: expected identifier, found '$'
Error: compile error
```

## Root cause (partial)

`parse_pattern`'s Object-pattern arm (`src/jq/parser.rs`, ~line 1617) always expects a `key: pattern` pair — it never checks for a bare `$var` entry as shorthand for `key = var, pattern = Pattern::Var(var)`.

## Scope note

Narrow, well-specified parser fix — `PatternEntry` already has the right shape (`key: String, pattern: Pattern`) to represent the desugared form directly, so this is likely a small addition to the object-pattern parsing loop's key-parsing step (recognize `$ident` as a shorthand key+pattern pair, distinct from the existing `ident:`/`"string":` explicit-key cases), not an architecture question. Not yet implemented — filing to make the gap trackable.

newhoggy:
🔧 Claimed by agent `aiden` — branch `issue-1139-object-pattern-var-shorthand`

PR #1141, opened 2026-08-17
feat(jq): implement ?// alternative destructuring patterns

### LINKED ISSUE #720: jq: ?// alternative destructuring patterns are not implemented
## Summary

jq's `?//` alternative-destructuring-pattern operator is not implemented.
Confirmed at three levels:

1. `Pattern` (`src/jq/expr.rs`, ~line 415-437) has exactly three variants —
   `Var(String)`, `Object(Vec<PatternEntry>)`, `Array(Vec<Self>)` — no
   `Alt`/`Or` variant to hold a list of alternatives.
2. `AsPattern` (~line 270) carries a single `pattern: Pattern` field, not a
   list.
3. Live-verified:
   ```
   $ echo '[1]' | succinctly jq '. as [$a] ?// {$a} | $a'
   compile error: parse error at position 10: expected '|', found '?'
   ```

Zero grep hits for `?//` anywhere in `src/` or `tests/` — no lexer/parser
token support exists for it at all. This is a complete feature absence, not
a testing gap.

## Why this matters

`?//` lets a filter try multiple destructuring shapes against the same
input, falling through to the next alternative on either a pattern-match
failure *or* an error raised in the body while bound to an earlier
alternative. It's real jq syntax (since jq 1.6) used for shape-tolerant
input handling. jq's own test suite treats it as its single densest tested
language feature — an exhaustive matrix over every 2-of-3 ordering of
object/array/plain-variable patterns (`jq.test:786-892`), plus subtler
cases: alternatives binding different variable sets (unbound vars from a
non-matching branch become `null`), and fallthrough triggered by an error
raised inside the matched branch's *body*, not just a pattern mismatch
(`man.test:842-844`).

A filter using `?//` currently fails loudly with a parse error rather than
silently misbehaving, so this is lower risk than a silent-wrong-answer bug —
but it blocks running a meaningful class of real-world jq scripts unmodified.

## Discovery context

Found via a coverage audit cross-referencing succinctly's jq test suite
against jq-1.7.1's own `tests/jq.test`.

newhoggy:
## Summary
- Implements jq's `?//` operator (since jq 1.6): `. as [$a] ?// {a: $a} | ...` lets a filter try multiple destructuring shapes against the same input, falling through to the next alternative on either a pattern-match failure or a body-evaluation error.
- `Expr::AsPattern`'s `pattern: Pattern` field becomes `patterns: Vec<Pattern>`. The pre-existing bare-`$var` `Expr::As` shape is untouched — the parser still produces it whenever no `?//` follows a plain variable pattern, so this is purely additive for the common non-`?//` case.
- Every AST-rewrite pass touching `AsPattern` (`substitute_var`, `expand_func_calls`, `substitute_func_param`, `jq_runner`'s namespaced-call rewrite) updated for the field rename; "shadowed" now means "any alternative binds this name."
- Gated to jq mode only: real yq's own parser rejects `?//` outright, confirmed live against yq v4.53.3 — succinctly's shared parser keeps erroring on it in yq mode too, rather than silently accepting broader syntax than the oracle.

## Semantics (all live-verified against jq 1.7.1)
- First alternative whose pattern matches *and* whose body doesn't error wins.
- A pattern-match failure or a body `error(...)`/type error falls through to the next alternative (retried from scratch — the whole body re-runs under the next alternative's bindings, not resumed).
- `break`, `halt`, and empty output do **not** trigger fallthrough — only a genuine error does.
- A variable bound only by a non-matching alternative resolves to `null` in the body, not an "undefined variable" error.
- The bind expression can be a generator; each output is independently destructured and fanned out.
- A body error's own partial output (if the body is itself a generator) is not discarded on fallthrough — each alternative actually tried contributes whatever it produced before failing.

## Deliberately out of scope (filed as follow-ups)
- #1139 — the `{$a}` object-pattern shorthand (implicit key from var name) isn't parsed at all; pre-existing, confirmed on `main`'s own binary, unrelated to `?//`.

Fixes #720

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` (full suite green; two unrelated flaky tests confirmed passing in isolation)
- [x] Verified against the pinned real `jq` binary across the full semantic matrix; verified yq mode correctly rejects `?//`
- [x] Patch coverage: 100% (18/18 new lines)

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -347,11 +347,11 @@ fn rewrite_namespaced_calls(expr: Expr) -> Expr {
         },
         Expr::AsPattern {
             expr,
-            pattern,
+            patterns,
             body,
         } => Expr::AsPattern {
             expr: Box::new(rewrite_namespaced_calls(*expr)),
-            pattern,
+            patterns,
             body: Box::new(rewrite_namespaced_calls(*body)),
         },
         Expr::StringInterpolation(parts) => {
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -23081,11 +23081,16 @@ fn eval_as_pattern<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     // still resolves to `null`, not "undefined variable" (confirmed live
     // against jq 1.7.1: `. as [$a] ?// {$b} | [$a,$b]` on `[1]` gives
     // `[1,null]`). A no-op allocation when `patterns` has one element (the
-    // common, non-`?//` case).
+    // common, non-`?//` case). Deduped (found by review): two alternatives
+    // commonly share a name (`. as [$a] ?// {a: $a}`), and each duplicate
+    // would otherwise cost `try_pattern_alternatives` a wasted
+    // `substitute_var` body clone-and-walk per matched alternative.
     let mut all_var_names: Vec<String> = Vec::new();
     for pattern in patterns {
         collect_pattern_var_names(pattern, &mut all_var_names);
     }
+    all_var_names.sort_unstable();
+    all_var_names.dedup();
 
     let mut all_results: Vec<OwnedValue> = Vec::new();
 
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -738,9 +738,9 @@ fn eval_single<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         // Phase 9: Variables & Definitions
         Expr::AsPattern {
             expr,
-            pattern,
+            patterns,
             body,
-        } => eval_as_pattern::<W, S>(expr, pattern, body, value, optional),
+        } => eval_as_pattern::<W, S>(expr, patterns, body, value, optional),
         Expr::FuncDef {
             name,
             params,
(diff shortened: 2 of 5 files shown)


newhoggy:
Ran the mandatory multi-agent review before merge — several independent passes (reuse/simplification/efficiency, cross-file rename tracer, removed-behavior audit, altitude/conventions) found **no correctness regressions**, which is a first for this session's issues. Fixed the one concrete improvement found:

- **Deduplicated `all_var_names`**: two `?//` alternatives commonly share a variable name (`. as [$a] ?// {a: $a}`), and each duplicate cost `try_pattern_alternatives` a wasted `substitute_var` body clone-and-walk per matched alternative. Fixed with a sort+dedup after collection.

Also found and fixed my own patch-coverage gap: several new tests used `run_jq_stdin`, which shells out via `cargo run` and builds a second, uninstrumented binary invisible to `cargo-llvm-cov` — a known trap already documented in this repo's own history, which I initially missed. Switched them to `run_jq_full` (spawns the pre-built, correctly-instrumented binary via `CARGO_BIN_EXE_succinctly`), and added 2 more targeted tests for shapes the existing ones didn't reach (a body referencing the original document via `QueryResult::One`, and one that also fans out via `QueryResult::Many`).

Remaining 4 uncovered lines are all either genuinely unreachable by design (the standard `OneCursor(_) => unreachable!()` pattern used throughout this file after `.materialize_cursor()`, and my own trailing `unreachable!()` after the always-returns loop) or diff-attribution false positives (two `.as_ref()` lines in the unrelated, unchanged `Range` arm that got swept into the "new lines" count by proximity to my edit).

Patch coverage: 96.95% (127/131 new lines).

Issue #1150, opened 2026-08-17
jq: --argjson can't accept a negative-number value at all (CLI arg parser rejects it, even with --)


newhoggy:
**Severity: Low**

Found while implementing #1094 (`--argjson` leading-zero tolerance) — while verifying a negative leading-zero case (`-007`), discovered the underlying problem is broader: **any** negative number as an `--argjson` value fails at the CLI argument-parsing layer, before `parse_json_value` (the JSON-content validator #1094 touches) ever runs.

## Repro

```
$ jq -n --argjson n '-7' '$n'                       # real jq: works fine
-7

$ succinctly jq -n --argjson n '-7' '$n'
error: unexpected argument '-7' found

  tip: to pass '-7' as a value, use '-- -7'

Usage: succinctly jq [OPTIONS] [FILTER] [FILES]...
```

Following clap's own suggested fix doesn't work either:

```
$ succinctly jq -n --argjson n -- '-7' '$n'
error: 2 values required for '--argjson <NAME> <VALUE>' but 1 was provided
```

## Root cause (partial)

Almost certainly a `clap` argument-definition issue: `--argjson` takes two values (`NAME VALUE`), and clap's negative-number detection (which normally lets a lone leading `-` argument through when no other flag could match it) doesn't extend correctly to the second of a two-value option. Not yet traced to the exact `clap::Arg` definition in `src/bin/succinctly/`.

## Scope note

Affects `--argjson` (and likely `--slurpfile`/`--rawfile`/similar two-value or value-taking options with the same clap shape) whenever the value legitimately starts with `-` — not just negative numbers with a leading zero. Filing to make the gap trackable; not investigated further since it's a CLI-parsing issue, not a JSON-validation one, and out of #1094's own scope.

newhoggy:
🔧 Claimed by agent `denna` — branch `issue-1150-argjson-negative-number`

Issue #1149, opened 2026-08-17
jq: a leading-zero number combined with scientific notation loses its exponent spelling (007e5 -> 700000, not 7E+5)


newhoggy:
**Severity: Low**

Found while implementing #1094 (`--argjson` leading-zero tolerance). Pre-existing, not introduced by that fix — reproduces identically via the plain document-input path, unrelated to `--argjson`.

## Repro

```
$ echo '007e5' | yq '.'         # via succinctly's own pinned jq oracle
7E+5
$ echo '007e5' | succinctly jq '.'
700000
```

A leading-zero number *without* scientific notation correctly preserves its spelling (minus the leading zero) already:

```
$ echo '007.500' | jq '.'
7.500
$ echo '007.500' | succinctly jq '.'   # correct, matches
7.500
```

So the gap is specific to the combination of a leading zero *and* an exponent (`e`/`E`) — `007e5`, `01.5e2`, etc. all lose their scientific-notation spelling and get expanded to a plain decimal instead, even though the underlying numeric *value* is correct either way (`700000 == 7E+5`).

## Root cause (partial)

Likely `is_preservable_float_literal` (referenced from `src/bin/succinctly/output.rs`, `src/jq/eval.rs`, `src/jq/value.rs`) rejecting a leading-zero-prefixed literal from the `NumberLiteral`-preservation fast path entirely (since leading zeros aren't valid RFC 8259 JSON syntax), falling back to full numeric parsing — which loses the exponent notation on re-display, unlike the non-exponent leading-zero case, which apparently takes some other, spelling-preserving path. Not yet traced further.

## Scope note

`--argjson`'s own `parse_json_value` (#1094) intentionally does *not* attempt to fix this — it materializes leading-zero `--argjson` values via this crate's own existing JSON semi-indexer unchanged, so `--argjson n '007e5' '$n'` inherits this exact gap (giving `700000`, not real jq's `7E+5`) rather than erroring outright as it did before #1094. That's still a net improvement (matches real jq's *acceptance* of the input), just not its exact spelling for this narrower sub-case.

newhoggy:
Broader than the title suggests — the same root cause also loses a *trailing* zero's spelling, not just the exponent-notation case:

\`\`\`
\$ echo '007.500' | yq '.'
7.500
\$ echo '007.500' | succinctly jq '.'
7.5
\`\`\`

[...]

PR #1152, opened 2026-08-17
fix(jq): --argjson tolerates a leading-zero number, matching real jq

### LINKED ISSUE #1094: jq: --argjson rejects leading-zero numbers (007, 00) that real jq accepts leniently

newhoggy:
## Summary
- Real jq's own number parser accepts leading-zero numbers (`007`, `00`, `007.5`) — not RFC 8259-valid JSON, but more lenient than strict JSON here. succinctly's `--argjson` rejected them outright via `serde_json::from_str`'s strict validation, which exists to reject trailing garbage (`42 garbage`, #284) and range-check magnitude overflow — neither of which this fix touches.
- Fixed by normalizing away a leading zero from every number token (skipping string contents/keys) only on the failure path, then retrying the same strict `serde_json` validation against the normalized text — so trailing-garbage rejection and the common (no-leading-zero) case's validation cost are both unaffected. The crate's own JSON semi-indexer already tolerates a leading zero on its own (confirmed live), so materialization still runs on the original text unchanged.

## Deliberately out of scope (filed as follow-ups)
- #1149 — a leading-zero literal loses full spelling fidelity on redisplay (`007.500` → `7.5`, `007e5` → `700000`, not real jq's `7.500`/`7E+5`). Reproduces identically via plain document input, unrelated to `--argjson` — a pre-existing gap in this crate's `NumberLiteral`-preservation fast path.
- #1150 — `--argjson` can't accept a negative-number value at all due to a CLI arg-parsing issue (affects every negative number, not just leading-zero ones).

Fixes #1094

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` (full suite green)
- [x] Verified against the pinned real `jq` binary across leading-zero variants, trailing-garbage combinations, nesting, string/key false-positives, and malformed input
- [x] Patch coverage: 100% (71/71 new lines)

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -1536,6 +1536,22 @@ fn normalize_leading_zero_numbers(s: &str) -> String {
             while i < bytes.len() && bytes[i].is_ascii_digit() {
                 i += 1;
             }
+            if int_start == i {
+                // A bare `-` with no digit following it at all (`-`,
+                // `[-,1]`, `{"a":-}`) isn't a number token in the first
+                // place -- only reachable when `b == '-'`, since a digit
+                // `b` always advances the scan above by at least one.
+                // Leaving it untouched (not fabricating a `0` digit) is
+                // required, not just tidier: filling one in turns a
+                // genuinely malformed token into a syntactically valid
+                // one (`-0`), which would make the retried `serde_json`
+                // validation below wrongly pass and this genuinely
+                // invalid input silently reach `json_bytes_to_owned_value`
+                // as `null` instead of erroring (found by review before
+                // merge).
+                out.push(b'-');
+                continue;
+            }
             let stripped = s[int_start..i].trim_start_matches('0');
             out.extend_from_slice(&bytes[start..int_start]);
             out.extend_from_slice(if stripped.is_empty() {
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -1474,9 +1474,106 @@ fn parse_json_value(s: &str) -> Result<OwnedValue> {
         return Ok(OwnedValue::Null);
     }
 
-    serde_json::from_str::<serde_json::Value>(s).with_context(|| format!("Invalid JSON: {s}"))?;
+    if let Err(e) = serde_json::from_str::<serde_json::Value>(s) {
+        // Real jq's own number parser tolerates a leading zero (`007`,
+        // `00`, `007.5`) that strict JSON doesn't (#1094) -- retry once
+        // with every number token's leading zero stripped (string
+        // contents/keys left untouched) before surfacing the original
+        // error. Only paid on this (rare) failure path, so the common
+        // case's validation cost is unaffected. `json_bytes_to_owned_value`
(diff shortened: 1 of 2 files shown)


newhoggy:
Ran the mandatory multi-agent review before merge — three independent passes found the same **critical bug**, now fixed:

**A bare `-` with no digit following it (`-`, `[-,1]`, `{"a":-}`) got silently fabricated into the valid number `-0`.** The digit-collection loop consumed zero characters for a lone hyphen, and the "all-zero run" fallback (meant for `00`-style input) inserted a `0` where none existed — turning a genuinely invalid token into a syntactically valid one. That made the retried strict validation wrongly pass, silently letting malformed input through as `null` instead of erroring — exactly the invariant `serde_json`'s strict gate exists to protect. Fixed by leaving a digit-less `-` untouched instead. All three failure scenarios the reviewers found now correctly error, matching real jq.

Also confirmed: this repo's own established "cargo run tests are invisible to coverage" trap (already documented in project memory) — my first test draft used a `cargo run`-based helper before I caught and fixed it.

Review also found real, legitimate duplication (not blocking): `normalize_leading_zero_numbers` reimplements string/escape scanning and JSON-number-grammar walking that already exist elsewhere in this same file (`find_string_end`, `find_number_end`) and in `src/json/validate.rs`'s `is_valid_number` — filed as #1154, since untangling it safely needs its own careful pass, not a rushed change under this PR's own review pressure.

Patch coverage: 100% (67/67 new lines).

Issue #1045, opened 2026-08-16
jq: generator-argument builtins (contains, ltrimstr, ...) don't fan out and error on a zero-output argument instead of propagating empty


newhoggy:
## Summary

Split out of #1043's own review — confirmed while verifying #1043's fix (routing 5
collapse sites through `owned_vec_to_result`/`borrowed_vec_to_result`) against its own
repro commands, all of which happened to be wrapped in `contains(...)`.

`result_to_owned` (`src/jq/eval.rs`) is the shared "collapse a generator argument's
`QueryResult` down to one scalar value" helper, used by ~37 call sites across the
codebase (`contains`, `ltrimstr`/`rtrimstr`, `split`, `join`'s separator, `strftime`'s
format, regex pattern/flags arguments, and many more). Its `QueryResult::None` arm
unconditionally errors:

```rust
QueryResult::None => Err(EvalError::new("no value").into()),
```

Real jq's actual semantics for a builtin taking a generator argument (`f(x)` desugars
roughly to `x as $b | ...`) is: if `x` produces zero outputs, the `as` binding runs its
body zero times, so the **whole call produces zero outputs** — not an error.

## Confirmed repros (verified against pinned jq 1.7.1)

```
$ jq -n '"abc" | ltrimstr((empty,empty))'
(exits 0, no output)
$ succinctly jq -n '"abc" | ltrimstr((empty,empty))'
jq: error (at <unknown>): no value
(exit 5)

$ echo 1 | jq 'contains(empty)'
(exits 0, no output)
$ echo 1 | succinctly jq 'contains(empty)'
jq: error (at <unknown>): no value
(exit 5)
```

This is a **pre-existing** divergence, not introduced by #1043 — two tests already in
`tests/jq_cli_tests.rs` (`test_result_to_owned_many_empty_arm_via_ltrimstr_argument` and
its `ManyOwned` sibling, added in #1038/PR containing `680a79095`) already documented and
pinned it as "a pre-existing, unrelated divergence from real jq worth flagging," predating
#1043. #1043's own fix changes *which* `result_to_owned` arm these particular repros hit
(from the `Many`/`ManyOwned`-empty arm to the `None` arm — different error message, "no
value" instead of "empty result") but doesn't resolve the underlying divergence, since
`result_to_owned`'s `None` arm errors unconditionally regardless of which upstream shape
produced the `None`.

## Also confirmed: `contains` doesn't fan out over a multi-output argument at all

[...]

Issue #867, opened 2026-08-14
jq: isvalid(...) swallows an unresolved break $label as true instead of letting it unwind


newhoggy:
**Severity: Low**

## Summary

`builtin_isvalid`'s match on its argument's evaluation result (`src/jq/eval.rs`, ~line 13363-13379) explicitly handles `Error`, `None`, `Halt`, and `Partial(_, Control::Halt(_))`, but has no arm for a bare `QueryResult::Break(_)` or `Partial(_, Control::Break(_))` — both fall into the `_ => QueryResult::Owned(OwnedValue::Bool(true))` catch-all, silently treating an unresolved `break $label` as "the argument is valid" instead of letting the break keep unwinding toward its enclosing `label`.

`isvalid` is a succinctly-only extension (not a real jq builtin — `jq: error: isvalid/1 is not defined` in real jq 1.7.1), so there's no oracle to diff against; this is about succinctly's own internal contract, which explicitly guards the analogous case for `halt` ("must exit the process, not report true/false and let evaluation continue") but not for `break`.

## Repro

```
$ echo '{"a":1}' | succinctly jq -c 'label $out | isvalid(paths(if type=="number" then break $out else true end))'
true
```

Expected: the break should keep unwinding toward `$out` (producing no output, exit 0), the same way it does today for `isvalid(break $out)` used directly — instead it's swallowed and reported as `true`.

## Root cause

`src/jq/eval.rs`, `builtin_isvalid`, the match on `eval_single(...)`'s result: only `Error`, `None`, `Halt`, and `Partial(_, Control::Halt(_))` have explicit arms; everything else (including `Break`, `Partial(_, Control::Break(_))`) falls into `_ => Bool(true)`.

## Why this is separate from #850

#850 fixed `builtin_paths_filter`'s per-node and root-level control-signal handling so that a `Break` from `paths(node_filter)` now correctly propagates out as a real `QueryResult::Break`/`Partial(_, Control::Break(_))` instead of always being swallowed internally. That's a *prerequisite* for this bug to be reachable through `paths(...)` at all — before #850, `builtin_paths_filter` could never produce a bare `Break` for `isvalid` to see. This issue is about `isvalid`'s own incomplete match, a pre-existing gap (also reachable via plain `isvalid(break $out)` even without `paths` involved) that #850 doesn't touch and isn't caused by.

## Suggested fix direction

[...]

Issue #850, opened 2026-08-13
jq: paths(node_filter) swallows an ordinary error/break per-node instead of aborting the whole call, diverging from real jq


newhoggy:
**Severity: Low**

## Summary

`builtin_paths_filter` (`src/jq/eval.rs`, the loop over `all_paths`) evaluates
`node_filter` once per node visited by the traversal. When that evaluation
ends in an ordinary `error(...)` or an unresolvable `break $label`, the
current policy is to silently swallow the escape and skip only that one
node, continuing on to evaluate `node_filter` against every remaining node.

Real jq's own `paths(node_filter)` is `path(recurse|select(node_filter))`.
An uncaught `error`/`break` inside `select(node_filter)` aborts the *entire*
`path(...)` generator, not just the current node — so real jq's `paths(...)`
call raises (or unwinds) and produces no further paths after the first
node whose filter escapes.

## Repro

```
$ jq -c '[1,"x",2] | paths(if type=="string" then error("bad") else true end)'
jq: error (at <stdin>:1): bad
[0]
$ succinctly jq -c 'paths(if type=="string" then error("bad") else true end)' <<< '[1,"x",2]'
[0]
[2]
```

Real jq streams `[0]` then raises, stopping there (never reaching index 2).
succinctly streams `[0]` **and** `[2]`, silently treating the string
element as if its filter had just been falsy instead of erroring.

## Root cause

`src/jq/eval.rs`, `builtin_paths_filter`'s per-node loop: an
`Error`/`Break` escape from the per-node filter evaluation is caught and
discarded (`// Error/Break: the prefix above is already credited; swallow
the escape itself and move to the next node.`), rather than aborting the
whole function the way real jq's generator-based `select` does.

## Why this is separate from #773

#773 (and its review-driven follow-up, merged via PR #831) fixed two
different bugs in the same function:
1. A multi-output `node_filter` collapsing into one always-truthy value
   before the truthiness check ran (#773's original report).
2. A node's own truthy output(s) produced *before* an escape in that same
   node's fan-out being incorrectly dropped along with the escape (found
   in #773's PR review).

Both of those are about **what gets credited for the current node**. This
issue is about a materially different question — **what happens to every
node *after* the one that errored** — which was already the pre-existing,
deliberately-unchanged behavior both fixes explicitly left alone (see the
`// Error/Break: ... swallow ... move to the next node` comment and the
pre-existing `test_paths_filter_still_swallows_ordinary_per_path_error`
[...]

Issue #824, opened 2026-08-12
jq: `break` inside a `path(...)` call can't reach a `label` outside it


newhoggy:
## Summary

A `break $label` raised inside a `path(...)` call's argument can't reach a `label $label` sitting *outside* that `path(...)` call: instead of unwinding cleanly to the label like real jq, it gets converted into an ordinary "break $label not in label" error, printed to stderr, and the process exits non-zero.

Found while adding coverage for #791/PR #816's `Expr::As` path-context fix; confirmed live against real jq 1.7.1. Not specific to `as`-bindings or to #791 — reproduces with a bare comma expression too, no `halt`/`halt_error` involved at all.

## Repro

```
$ echo '{"a":1}' | jq -c 'label $out | path((.a, break $out))'
["a"]
$ echo '{"a":1}' | succinctly jq -c 'label $out | path((.a, break $out))'
["a"]
jq: error (at <stdin>:1): break $out not in label
```
Real jq: exit 0, nothing on stderr. succinctly: exit 5, `["a"]` on stdout (correctly kept — the comma's first branch resolved before the break) but also the bogus error on stderr.

Same mismatch through an `as`-binding's bind source:

```
$ echo '{"a":1}' | jq -c 'label $out | path((1, break $out) as $x | .a)'
["a"]
$ echo '{"a":1}' | succinctly jq -c 'label $out | path((1, break $out) as $x | .a)'
["a"]
jq: error (at <stdin>:1): break $out not in label
```

Real jq: the `break $out` unwinds to the enclosing `label $out`, `path(...)` having already produced `["a"]` (from the `$x=1` binding), and the whole pipeline exits 0. succinctly instead treats the break as unmatched no matter where the matching `label` actually is.

## Root cause

`resolve_node`'s `PathResolveResult` (`Result<Vec<PathBranch>, (Vec<PathBranch>, EvalEscape)>`) has no way to carry a `Control::Break` — `EvalEscape` only has `Error`/`Halt` variants. Every path-context site that receives a `Control::Break` while resolving a sub-expression (`eval_owned_multi`'s pre-existing `QueryResult::Break` arm, and `Expr::As`'s new `Some(Control::Break(label))` trailing-control arm added by #791) has no choice but to convert it into an ordinary `EvalError::new("break $label not in label")` — even when a `label` with that exact name is sitting right outside the `path(...)` call in value context, fully able to catch it.

[...]

PR #1160, opened 2026-08-17
fix(jq): break can now reach an outer label through builtin-argument eval

### LINKED ISSUE #833: jq: break can't reach an outer label through most builtin-argument evaluation (result_to_owned/eval_owned_expr)
**Severity: Medium**

## Summary

A `break \$label` raised while evaluating a builtin's *argument* expression can't reach a `label \$label` sitting outside the builtin call, for the large majority of builtins whose argument goes through `result_to_owned` or `eval_owned_expr` — the same "no way to carry `Control::Break`" gap #824 fixed specifically for `path(...)`, but far broader in scope here: these two helpers back the argument evaluation of dozens of builtins (regex functions, `ltrimstr`/`rtrimstr`, `splits`, `strftime`, `parent(n)`, and more).

## Repro

```
$ echo '"abc"' | jq -c 'label $out | test("a"; (break $out))'
(no output, exit 0)
$ echo '"abc"' | succinctly jq -c 'label $out | test("a"; (break $out))'
jq: error (at <stdin>:1): break $out not in label
```

Real jq: the break unwinds cleanly to `$out`; the whole pipeline produces no output and exits 0. succinctly: the break is folded into a bogus "not in label" error at the point `test`'s second argument is evaluated, regardless of the label sitting right outside.

`ltrimstr` reproduces the same shape and is already pinned as a known, documented gap:

```
$ jq -n 'label $out | ("abc" | ltrimstr(break $out))'
(no output, exit 0)
```
— see `test_result_to_owned_none_error_break_arms_via_ltrimstr_argument` in `tests/jq_cli_tests.rs`, which already documents this exact divergence for `result_to_owned`.

`parent(n)`'s `n` argument reproduces it too, via `eval_owned_expr` instead:
```
$ echo '{"a":{"b":1}}' | succinctly jq -c '.a.b | parent(break $out)'
jq: error (at <stdin>:1): break $out not in label
```
— pinned (as expected/current behavior) by `test_parentn_n_expr_break_reported_as_break_not_in_label`.

## Root cause

- `result_to_owned` (`src/jq/eval.rs`) has a `QueryResult::Break(label) => Err(EvalError::new(format!("break ${label} not in label")).into())` arm — used by dozens of call sites across the evaluator for single-value argument evaluation (regex pattern/flags, `ltrimstr`/`rtrimstr` prefix, `splits` separator, `strftime` format string, and many more — see the call-site list in `src/jq/eval.rs`).
- `eval_owned_expr` similarly has `Control::Break(label) => EvalError::new(format!("break ${label} not in label")).into()` in its `Control`-to-`EvalEscape` conversion, used by e.g. `parent(n)`'s `ParentN` argument, `resolve_limit`'s `n` bound, and others.

Both are `EvalEscape`-returning wrappers around lower-level helpers (`eval_owned_expr_ctrl`, `eval_single`) that *do* carry `Control::Break` losslessly — the information exists right up until these two functions collapse it.

## Why this is separate from #824

#824 fixed the identical root cause (`EvalEscape` having no `Break` variant) but deliberately scoped to `resolve_node`'s own domain (`path()`, and the `=`/`|=`/`del()` computed-key machinery that shares it via `eval_owned_multi`/`Expr::As`) — see that PR's discussion of why `result_to_owned`/`eval_owned_expr` were left untouched. Those two helpers are used far more broadly (by evaluator internals having nothing to do with path context), so propagating `Break` through them safely needs its own audit of every call site — the same kind of "moderate-to-large blast radius" review #824's own issue text originally worried applied to `resolve_node` (and turned out not to, once actually investigated).

## Suggested fix direction

This is naturally a two-step precursor-and-slices shape, not a single PR:

1. **Precursor spike**: enumerate every `result_to_owned`/`eval_owned_expr` call site (there are dozens) and classify each as "safe to propagate `Break`" vs. "needs its own care" (e.g. a site already inside a loop construct that has its own break-catching, like #575's `while`/`until`/`repeat`/`reduce`/`foreach` fix, which already uses the `Control`-preserving `eval_owned_expr_ctrl` twin instead of `eval_owned_expr` for exactly this reason).
2. Either flip `result_to_owned`/`eval_owned_expr` to propagate a real `Break` (via `EvalEscape::Break`, added by #824) directly — auditing that no caller relies on the old collapse-to-error behavior — or introduce `_ctrl`-suffixed twins at the specific call sites that need it, mirroring the `eval_owned_expr`/`eval_owned_expr_ctrl` split #575 already established.

Given the number of call sites, this is likely Tier 2-or-3 sized work on its own, not a quick follow-up.

newhoggy:
## Summary

Fixes #833. `result_to_owned` and `eval_owned_expr` — the two shared helpers backing single-value argument evaluation for dozens of builtins (`ltrimstr`/`rtrimstr`, `has`, `contains`, `inside`, `test`, `getpath`, regex pattern/flags, `strftime`, `parent(n)`, and many more) — collapsed an unmatched `break $label` raised while evaluating a builtin's argument into a synthetic `"break $label not in label"` error, instead of propagating it so it could reach a `label $label` sitting outside the builtin call. Real jq unwinds cleanly; succinctly misreported it as a catchable error.

```bash
$ echo '"abc"' | jq -c 'label $out | test("a"; (break $out))'
(no output, exit 0)
$ echo '"abc"' | succinctly jq -c 'label $out | test("a"; (break $out))'   # before this PR
jq: error (at <stdin>:1): break $out not in label
```

## Fix

Both functions' `Break` arm now produces `Err(EvalEscape::Break(label))` — the variant #824 added specifically for this purpose — instead of collapsing it. The pre-existing `From<EvalEscape> for QueryResult`/`Control` conversions already thread `Break` through correctly, so every "safe" call site needed zero changes of its own.

## Precursor audit

Per the issue's own suggested methodology, I enumerated and classified all 41 call sites of the two functions before making the change (36 for `result_to_owned`, 5 for `eval_owned_expr`) — 40 were already safe (plain `?` or `Err(e) => return e.into()` forwarding), and one, `builtin_envvar`, had a dedicated `Halt` arm (from #791) but no `Break` arm, so it would have fallen into the generic `optional`/type-error fallback arms instead. Fixed alongside the main change. (`builtin_envvar` has no parser construction site anywhere in the codebase today — `env.VAR`/`$ENV.VAR` resolve to plain field access instead — so this is a latent gap, not a live user-facing bug, but worth closing for consistency.)

`builtin_paths_filter`, which the issue's own comment thread flagged as needing a follow-up decision once this landed, turned out to already be independently fixed (uses `eval_owned_expr_ctrl` directly, confirmed live against jq 1.7.1) — no change needed there.

## Testing

- New/updated in `tests/jq_cli_tests.rs`: `test_break_via_ltrimstr_argument_reaches_outer_label_833`, `test_break_via_parentn_argument_reaches_outer_label_833`, `test_isvalid_propagates_break_through_path_context_routing_833`, `test_isempty_propagates_break_through_path_context_routing_833` (the latter two cover a break reaching `isvalid`/`isempty` through `eval_pipe_with_path_context_internal`'s fallback → `eval_owned_expr` routing, a genuinely different call path from direct evaluation, called out explicitly in the issue's second comment).
- Renamed/re-split two tests that previously pinned the pre-fix buggy behavior as expected (`test_result_to_owned_none_error_break_arms_via_ltrimstr_argument` → `..._none_error_arms...` (its `None`-arm coverage is untouched, a separate tracked gap, #1045) + the new break test; `test_parentn_n_expr_break_reported_as_break_not_in_label` → `..._genuinely_uncaught_break_still_errors` (unchanged assertions — no enclosing label, so still correctly errors) + a new test with an actual enclosing label).
- New in `src/jq/eval.rs`: `builtin_envvar_propagates_break_from_variable_name_expression_833`.
- All pre-existing break/halt/error propagation tests (`#791`, `#824`, `#850`, `#867`, `#879`, path-context fanout) pass unmodified.
- Full local pipeline green: `cargo build --features cli`, full `cargo test --features cli,simd,regex,serde`, both required clippy invocations, `cargo check --no-default-features` (no_std), `cargo doc --all-features` with `-D warnings`.

## Out of scope

The issue's own `None`-arm divergence in `result_to_owned` (jq's `empty` argument yields zero output; succinctly errors) is a separate, already-tracked gap (#1045), left untouched here.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, clean)

### DIFF
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -81,19 +81,23 @@ pub enum Control {
 /// a synthetic `EvalError::new("break $label not in label")` — correct only
 /// when nothing outside actually catches it, and wrong (loud, wrongly-worded,
 /// wrongly-exit-coded) whenever a matching `label` sits further out, e.g.
-/// across a `path(...)` call boundary. Not every caller of this type has been
-/// audited to propagate `Break` correctly yet — `result_to_owned` and
-/// `eval_owned_expr` still collapse it into that same synthetic error by
-/// design, matching this type's `eval_owned_expr_ctrl`/`eval_owned_expr`
-/// split (#575's precedent: a `_ctrl`-suffixed twin that preserves
-/// [`Control`] losslessly exists at the few call sites that have been made to
-/// need it). Only `eval_owned_multi`/`eval_owned_multi_first` and
-/// `resolve_node`'s own arms (in `eval.rs`) have been updated to propagate a
-/// real `Break` so far, which is what `path()`, and the computed-key path
-/// this type also drives for `=`/`|=`/`del()`, needed. Auditing every
-/// `result_to_owned`/`eval_owned_expr` call site for the same fix is tracked
-/// separately (#833) — it is used far more broadly than path-context
-/// resolution, so it needs its own review rather than riding along here.
+/// across a `path(...)` call boundary. `eval_owned_multi`/
+/// `eval_owned_multi_first` and `resolve_node`'s own arms (in `eval.rs`)
+/// were the first to propagate a real `Break`, for `path()` and the
+/// computed-key path this type also drives for `=`/`|=`/`del()`. `#833`
+/// closed the far broader remaining gap: `result_to_owned` and
+/// `eval_owned_expr` (used by dozens of builtins' argument evaluation, not
+/// just path context) now propagate a *bare* unmatched `Break` too, matching
+/// this type's `eval_owned_expr_ctrl`/`eval_owned_expr` split (#575's
+/// precedent: a `_ctrl`-suffixed twin that preserves [`Control`] losslessly
+/// exists at the few call sites that need it). One shape is still open,
+/// though: when the argument generator produces one or more values *before*
+/// breaking/erroring (`QueryResult::Partial`), `result_to_owned` and
+/// `eval_owned_expr_ctrl` both still silently take the first value and drop
+/// the trailing escape — tracked as #1164, since fixing it means every
+/// caller becoming `Partial`-aware itself, not something these two
+/// functions can solve alone; see their own `Partial(vs, _control)` arms in
+/// `eval.rs` for the full rationale.
 ///
 /// Consumers should write `Err(EvalEscape::Error(e))` for the catchable case
 /// and let everything else flow through the `From` conversions into
@@ -105,10 +109,11 @@ pub enum EvalEscape {
     /// A genuine evaluation error — catchable by `try`/`catch`, suppressible
     /// by `?`.
     Error(EvalError),
-    /// `break $label`, still looking for a `label $label` to catch it (#824).
-    /// Every consumer that has not been specifically updated to propagate
-    /// this should keep converting it to a synthetic "break $label not in
-    /// label" `Error` — see this type's own doc comment.
+    /// `break $label`, still looking for a `label $label` to catch it (#824,
+    /// #833). A consumer that has not been specifically updated to
+    /// propagate this converts it to a synthetic "break $label not in
+    /// label" `Error` instead — see this type's own doc comment for which
+    /// consumers still do that.
     Break(String),
     /// `halt`/`halt_error(n)`: exit the whole process with this code. Must
     /// reach the CLI unconditionally; never catchable, never suppressible.
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -1362,15 +1362,28 @@ fn result_to_owned<W: Clone + AsRef<[u64]>>(
         // an argument stream that produced values and *then* halted must
         // halt, not quietly compute with the first value (#791).
         QueryResult::Partial(_, Control::Halt(code)) => Err(EvalEscape::Halt(code)),
+        // A trailing `Break`/`Error` here is a separate, harder gap #833
+        // does NOT fix: real jq would still abort after this point (e.g.
+        // `ltrimstr(("a", break $out))` prints one result then unwinds to
+        // `$out`, never running whatever follows), but this function's
+        // whole contract is collapsing a generator argument to one scalar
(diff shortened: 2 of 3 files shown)


newhoggy:
Code review (10-way fan-out) found a real gap the original fix's own comments overclaimed: **"propagates losslessly" only held for the bare, value-less `break`/`error` case.** When the argument generator produces a value *before* breaking/erroring (`QueryResult::Partial`), `result_to_owned` and `eval_owned_expr_ctrl` still silently drop the trailing escape and use the first value — live-verified against jq 1.7.1 for both `break` and `error`. Filed #1164 to track this properly: fixing it means making every one of `result_to_owned`'s ~40 callers `Partial`-aware (able to emit their own transformed value *and* a trailing escape), which is a materially larger, separately-reviewable change — not something foldable into this PR without ballooning its scope. Corrected the "losslessly" claims and added honest disclosure comments on both `Partial` arms instead, cross-referencing #1164.

Also fixed within this PR (same mechanical shape as the original fix, found auditing beyond `result_to_owned`'s own call sites):
- `range_arg` (backing `range()`'s bounds) — had no `Break` arm, misreported as "Range bounds must be numeric".
- `get_number_from_result`/`NumberError` (backing `pow`/`atan2`'s arguments, all 4 argument slots) — same missing-`Break`-variant shape.

Plus: simplified `builtin_envvar`'s two single-purpose Halt/Break arms into one OR-pattern (matching this file's own established precedent), and corrected three doc comments (`EvalEscape`'s canonical contract in `error.rs`, `eval_pipe_with_path_context_internal`'s `Expr::Break` arm, `builtin_paths_filter`'s rationale) that described the pre-fix collapse-to-error behavior as still current.

5 new tests added for the sibling-helper fixes. Full local pipeline re-verified clean after rebasing onto latest `main` (full test suite, both clippy invocations, no_std, docs).

Issue #1168, opened 2026-08-17
jq: Expr::Array/Comma have no cursor-native handling, so anything wrapped in [...] or joined with , loses duplicate YAML keys even after a builtin's own cursor-native fix (#443, #868)


newhoggy:
**Severity: Medium**

## Summary

Found while implementing #868's fix (PR for plain \`paths\`). \`src/jq/eval_generic.rs\`'s \`eval_single\` (the cursor-native fast-path evaluator shared by both \`jq_runner.rs\` and \`yq_runner.rs\`) has no native arm for \`Expr::Array\` or \`Expr::Comma\` — both fall into the function's final wildcard:

\`\`\`rust
// Fall back to the full evaluator for complex expressions
_ => {
    let owned = to_owned_with_cursor(&value, cursor);
    let json_str = owned.to_json_for_reindex::<S>();
    // ...reparse and hand the WHOLE expression to full_eval (eval.rs)...
}
\`\`\`

This materializes the **entire input document** into an \`OwnedValue\` (whose \`Object\` variant is backed by an \`IndexMap\`, collapsing duplicate YAML mapping keys to last-value-wins) *before* evaluating any part of the wrapped expression — including a builtin that has its own cursor-native, duplicate-key-preserving fix. The builtin's fix never runs; the whole subtree is re-evaluated against the already-collapsed document via \`eval.rs\`'s \`OwnedValue\`-based full evaluator instead.

## Repro (confirmed live)

\`to_entries\` already has a cursor-native fix for exactly this class of bug (#443) — and it works correctly on its own:

\`\`\`
$ printf 'a: 1\na: 2\nb: 3\n' | succinctly yq -o json 'to_entries'
[{"key":"a","value":1},{"key":"a","value":2},{"key":"b","value":3}]
\`\`\`

But wrap it in \`[...]\` (an extremely common, arguably the *idiomatic* way to consume a streaming builtin) or join two calls with \`,\`, and the fix silently stops applying:

\`\`\`
$ printf 'a: 1\na: 2\nb: 3\n' | succinctly yq -o json '[to_entries]'
[[{"key":"a","value":2},{"key":"b","value":3}]]
$ printf 'a: 1\na: 2\nb: 3\n' | succinctly yq -o json 'to_entries, to_entries'
{"key":"a","value":2} {"key":"b","value":3} {"key":"a","value":2} {"key":"b","value":3}
```

[...]

Issue #773, opened 2026-08-10
jq: paths(node_filter) collapses multi-output node_filter via eval_owned_expr, breaks truthiness check


newhoggy:
## Summary

`builtin_paths_filter` (`src/jq/eval.rs` ~line 12255) evaluates `node_filter`
via `eval_owned_expr` (~line 12274), which collapses a multi-output result
into a single value if `len()==1`, else wraps ALL outputs in one
`OwnedValue::Array`. A non-empty array is always truthy in jq (only `null`
and `false` are falsy), so if `node_filter` produces 2+ outputs,
`result.is_truthy()` sees an array, not the individual output values, and
treats it as truthy regardless of what those outputs actually are —
silently keeping every path when real jq's `path(recurse|select(node_filter))`
would keep none.

## Repro (confirmed against real jq and against this branch's release build)

```
$ echo '{"a":1,"b":{"c":2}}' | jq -c '[paths(false,false)]'
[]

$ echo '{"a":1,"b":{"c":2}}' | ./target/release/succinctly jq -c '[paths(false,false)]'
[["a"],["b"],["b","c"]]
```

Real jq's outer `select` checks each of `node_filter`'s outputs
independently and keeps `.` once per truthy output — here both outputs are
`false`, so nothing is kept. succinctly's collapse turns the two falsy
outputs into the truthy array `[false,false]` before the truthiness check,
so every path is incorrectly kept.

## Related pattern

This is the same "collapse via `eval_owned_expr` then check `.is_truthy()`"
shape already tracked and fixed at several other call sites: #570 (the
general collapsing behavior), #627 (`recurse(f; cond)`'s `cond`), #628
(`resolve_node`'s `Select`/`If` arms). `builtin_paths_filter` wasn't in that
audit's original list and appears to be a remaining un-migrated site — the
project has a cardinality-preserving sibling (`eval_owned_expr_fork`,
`src/jq/eval.rs` ~line 1021) already built for exactly this purpose by
#570.

Also note: even switching to a per-output truthiness check likely needs to
decide the fan-out question from #534 for a guard/filter position — does a
multi-output `node_filter` keep the path once (any truthy output) or
duplicate the path once per truthy output (jq's literal `select`
semantics, since `select(f)` re-emits `.` for each truthy output of `f`)?
Needs confirming against real jq before fixing; the repro above only proves
the current all-falsy case is wrong, not what the multi-truthy case should
do.

## Discovery context

[...]

Issue #443, opened 2026-07-28
YAML: to_entries and JSON/-o=json output collapse duplicate mapping keys to one entry


newhoggy:
**Severity: Medium (deviation)** — reproduced against real `yq` (mikefarah v4.53.3).

Follow-up to #174, which fixed `.key` field access (`YamlFields::find`) to be last-wins for duplicate YAML mapping keys. Two other query paths still silently *collapse* duplicates to a single entry instead of preserving both, diverging from `yq`'s raw pass-through behavior.

Repro:

```
$ printf 'a: 1\na: 2\n' > /tmp/dup.yaml

$ succinctly yq 'to_entries' /tmp/dup.yaml
- key: a
  value: 2                  # only 1 entry

$ yq 'to_entries' /tmp/dup.yaml
- key: a
  value: 1
- key: a
  value: 2                  # yq: 2 entries, unmerged

$ succinctly yq -o=json '.' /tmp/dup.yaml
{"a": 2}                    # merged to one key

$ yq -o=json '.' /tmp/dup.yaml
{"a": 1, "a": 2}             # yq: raw pass-through, both keys kept (technically invalid JSON, but that's yq's actual output)
```

For contrast, `.[]` and `keys` on the same input already show both entries unmerged and are correct:

```
$ succinctly yq '.[]' /tmp/dup.yaml
1
2

$ succinctly yq 'keys' /tmp/dup.yaml
- a
- a
```

**Root cause:** `to_owned()` in `src/jq/eval_generic.rs:34-45` converts a `DocumentValue` mapping into an owned `IndexMap` via `map.insert(key, value)` per field. `IndexMap::insert` on a repeated key overwrites the prior entry, so any builtin that materializes an owned value from a YAML mapping (`to_entries`, JSON output conversion, etc.) silently drops all but the last occurrence of a duplicate key — even though the raw field iterator it reads from (`DocumentFields::uncons`) does emit every field, duplicates included.

**Fix:** either (a) have `to_owned` build `OwnedValue::Object` from an ordered `Vec<(String, OwnedValue)>` instead of an `IndexMap` so duplicate keys survive, or (b) explicitly document that any owned-value materialization is last-wins-per-key by design and adjust `to_entries`/JSON output accordingly to match `yq`. Whichever is chosen, `to_entries` and `-o=json` output should agree with `.[]`/`keys` about whether duplicates are preserved.

Confirmed present on `main` (pre-existing, not introduced by #174's fix).

---
[...]

PR #1169, opened 2026-08-17
fix(jq): paths preserves duplicate YAML mapping keys

### LINKED ISSUE #868: yq: paths/paths(node_filter) silently drop duplicate mapping keys that identity output preserves
**Severity: Low**

## Summary

`builtin_paths_filter`/`builtin_paths`'s `to_owned(&value)` call (`src/jq/eval.rs`) collapses duplicate YAML mapping keys (last-value-wins) before `collect_paths`/`get_value_at_path` ever run, so `paths`/`paths(node_filter)` in yq mode silently ignore all but the last of a duplicate key — even though plain identity output preserves every duplicate.

## Repro

```
$ printf 'a: 1\na: 2\nb: 3\n' | succinctly yq -o json '.'
{
  "a": 1,
  "a": 2,
  "b": 3
}
$ printf 'a: 1\na: 2\nb: 3\n' | succinctly yq -o json '[paths]'
[["a"],["b"]]
```

Only one `"a"` path is reported, and `paths(node_filter)`/`path(...)`-based navigation into the duplicated key can only ever reach whichever copy `to_owned` happened to keep.

## Root cause

Not fully diagnosed — same shape as #631 (closed): a `to_owned()` conversion earlier in the pipeline collapses `OwnedValue::Object`'s duplicate keys via last-value-wins before path-producing logic runs. #631 was scoped specifically to `evaluate_yaml_cursor`/`first(.[])`/`last(.[])`/computed indexing; this is the same root-cause class surfacing at a different call site (`builtin_paths_filter`'s own `to_owned(&value)`), not covered by that fix.

## Why this is separate

Found while reviewing #850 (which hardens `builtin_paths_filter`'s control-signal handling, unrelated to this). Pre-existing, not introduced by that change — `to_owned` here is untouched code.

## Suggested fix direction

Needs investigation: whether `collect_paths`/`get_value_at_path` can walk the pre-`to_owned` representation directly (preserving duplicate keys the way identity output does), or whether `OwnedValue::Object` itself would need to stop deduplicating. Likely worth scoping alongside any other `to_owned()`-based builtins found to have the same issue, rather than fixing this one call site in isolation.

newhoggy:
## Summary

Addresses #868 (does not close it — see "Scope" below). `builtin_paths_filter`'s `to_owned(&value)` call collapsed duplicate YAML mapping keys into `OwnedValue::Object`'s `IndexMap` (last-value-wins) before `collect_paths` ever ran, so `paths` on YAML input with a repeated key reported only one path even though identity output preserves every occurrence:

```bash
$ printf 'a: 1\na: 2\nb: 3\n' | succinctly yq -o json '.'
{"a": 1, "a": 2, "b": 3}
$ printf 'a: 1\na: 2\nb: 3\n' | succinctly yq -o json 'paths'   # before this PR
["a"]
["b"]
```

## Fix

Adds a native `Builtin::Paths` arm to `eval_generic.rs`'s cursor-native evaluator, mirroring `Builtin::ToEntries`'s existing fix for the identical bug shape (#443): a new `collect_paths_cursor` walks `as_object()`/`as_array()` directly via `DocumentFields`/`DocumentElements::uncons()` (a cons-list, or for YAML's merge-resolved mappings an `Rc`-shared entry list — never a shared map) instead of requiring a `to_owned()`'d `OwnedValue` first, so no key ever passes through a structure that could collapse a duplicate. Depth-guarded via `assert_nesting_depth`, matching this module's own convention for recursive walkers.

## Scope — why this doesn't close #868

Investigating the fix surfaced a broader, separate gap: `Expr::Array` (and `Comma`) have no cursor-native handling at all in `eval_generic.rs::eval_single` — both fall to a wildcard that materializes the **entire input document** into an `OwnedValue` and hands the whole expression to `eval.rs`'s full evaluator, before the inner expression (including this fix's own `Builtin::Paths` arm) ever runs. That's exactly the shape of #868's own issue-text repro:

```bash
$ printf 'a: 1\na: 2\nb: 3\n' | succinctly yq -o json '[paths]'   # still, after this PR
[["a"],["b"]]
```

Confirmed this isn't `paths`-specific — `[to_entries]` has the identical silent regression despite `to_entries`'s own already-shipped #443 fix (bare `to_entries` is correct, `[to_entries]` collapses). This is a foundational, separate architectural gap (which combinators `eval_generic.rs` natively understands), not a quick follow-up to either #443 or #868 — filed as #1168.

So: `paths` (bare/streamed) is fixed by this PR; `[paths]` (the array-wrapped form, and the exact command #868's own repro used) is not, pending #1168. Left #868 open rather than closing it, since a reader testing the issue's own literal repro would still see it fail.

## Testing

- `test_duplicate_mapping_key_paths_preserves_both_868` — the newly-fixed bare `paths` case.
- `test_duplicate_mapping_key_array_wrapped_paths_still_collapses_868_known_gap` — a characterization test pinning the still-open `[paths]` gap (documented, not silently regressed), with a test ready to flip once #1168 lands.
- Verified live against real `jq`/`yq` across nested objects/arrays, empty containers, scalar roots, and (unaffected, still routing through the old fallback) `paths(node_filter)`/`leaf_paths`.
- Full local pipeline green: `cargo build --features cli`, full `cargo test --features cli,simd,regex,serde`, both required clippy invocations, `cargo check --no-default-features` (no_std), `cargo doc --all-features` with `-D warnings`.

## Out of scope

- `Builtin::PathsFilter`/`LeafPaths`/`Recurse*`/`Walk` share the same wildcard fallback and likely the same duplicate-key bug for their own bare-dispatch case, but `paths(node_filter)` specifically needs its two-phase collect-then-filter logic fused into one walk (today's `builtin_paths_filter` collects all paths first, then re-looks-up each value — two duplicate keys produce the *same* path array, so a later re-lookup can't disambiguate which occurrence to filter). That's real, separate restructuring work on a function with dense, hard-won correctness properties from #773/#850/#881/#833/#861 — left for #868 itself to keep tracking, not attempted here.
- #1168 (the `Expr::Array`/`Comma` cursor-native gap) is its own, larger design question.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, clean)

### DIFF
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -3227,6 +3227,46 @@ fn slice_one_generic<S: EvalSemantics, V: DocumentValue>(
     }
 }
 
+/// Cursor-aware counterpart to `eval.rs`'s `collect_paths` (#868): walks
+/// `value` directly via `DocumentFields`/`DocumentElements::uncons()`
+/// instead of requiring a `to_owned()`'d `OwnedValue` first, so duplicate
+/// YAML mapping keys are never collapsed by `OwnedValue::Object`'s
+/// `IndexMap` before path collection runs — the same #443 pattern
+/// `Builtin::ToEntries` above already established for this file. `uncons()`
+/// walks a cons-list (or, for YAML's merge-resolved mappings, an `Rc`-shared
+/// entry list), never a shared map, so no key is ever put somewhere a
+/// duplicate could overwrite it.
+fn collect_paths_cursor<V: DocumentValue>(
+    value: &V,
+    current_path: &mut Vec<OwnedValue>,
+    paths: &mut Vec<OwnedValue>,
+) {
+    assert_nesting_depth(current_path.len());
+    if let Some(fields) = value.as_object() {
+        let mut f = fields;
+        while let Some((field, rest)) = f.uncons() {
+            if let Some(key) = field.key_str() {
+                current_path.push(OwnedValue::String(key.into_owned()));
+                paths.push(OwnedValue::Array(current_path.clone()));
+                collect_paths_cursor(&field.value, current_path, paths);
+                current_path.pop();
+            }
+            f = rest;
+        }
+    } else if let Some(elements) = value.as_array() {
+        let mut i: i64 = 0;
+        let mut e = elements;
+        while let Some((elem, rest)) = e.uncons() {
+            current_path.push(OwnedValue::Int(i));
+            paths.push(OwnedValue::Array(current_path.clone()));
+            collect_paths_cursor(&elem, current_path, paths);
+            current_path.pop();
+            i += 1;
+            e = rest;
+        }
+    }
+}
+
 /// Evaluate a builtin function.
 fn eval_builtin<S: EvalSemantics, V: DocumentValue>(
     builtin: &Builtin,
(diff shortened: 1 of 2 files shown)


newhoggy:
Closing without merging — code review surfaced a genuine correctness regression, not just a known gap.

**The critical finding**: `collect_paths_cursor` walks the raw field cons-list unconditionally, which correctly preserves YAML's duplicate mapping keys — but the same generic code also runs for **JSON** via the shared `eval_generic.rs` evaluator, and `JsonFields::uncons()` reflects raw, un-deduplicated token order too. Confirmed live:

```
$ echo '{"a":1,"b":2,"a":3}' | jq -c 'paths'
["a"]
["b"]
$ echo '{"a":1,"b":2,"a":3}' | succinctly jq -c 'paths'   # this PR
["a"]
["b"]
["a"]
```

Real jq collapses duplicate JSON keys (matching RFC 8259's own last-value-wins convention); this PR's fix makes succinctly wrongly report the duplicate as two separate paths.

**Important context discovered while investigating**: this isn't a new bug class my PR introduces from scratch — the identical mechanism already exists, live, on `main`, in the already-shipped `#443` fix for `to_entries`:

```
$ echo '{"a":1,"b":2,"a":3}' | jq -c 'to_entries'
[{"key":"a","value":3},{"key":"b","value":2}]
$ echo '{"a":1,"b":2,"a":3}' | succinctly jq -c 'to_entries'   # unrelated to this PR, already on main
[{"key":"a","value":1},{"key":"b","value":2},{"key":"a","value":3}]
```

My PR would have extended this same, already-live bug to a second builtin. Filing this as its own issue since it's a real, currently-shipped divergence independent of anything here.

**Also found** (via the review's cross-file tracing): the disclosed remaining-gap scope in this PR's body (`Expr::Array`/`Comma`, filed as #1168) understated the real surface — `Expr::As`, `Expr::Reduce`, `Expr::Foreach`, `Builtin::Limit`, `Expr::If`, `Expr::Try`, and `Expr::Label` all hit the same `to_owned()`-materializing wildcard and reproduce the original duplicate-key-collapsing bug for `paths` wrapped in any of those constructs, not just `[paths]`/`paths, paths`. Folding this into #1168's own scope rather than filing separately, since it's the same root cause (no cursor-native handling for anything beyond a curated allowlist of simple shapes).

Properly fixing `paths`'s own duplicate-key case needs format-aware deduplication — likely a new `DocumentFields`-level primitive (the existing `find`/`find_cursor` contract already documents "YAML keeps the last duplicate key; JSON keeps the first," so the per-format distinction is known, just not yet reusable from a raw `uncons()` walk) — rather than the straight port of `ToEntries`'s pattern this PR attempted. Left #868 open with these findings; a future attempt should design the JSON/YAML split before implementing, and ideally fix `to_entries`'s existing bug in the same pass since they share root cause.

Issue #1176, opened 2026-08-17
yq: -i (in-place) mode loses an explicit !!float tag on a whole-number-looking scalar


newhoggy:
**Severity: Medium**

## Summary

Found while implementing #1090 (nested computed-float `!!float` tag emission). `succinctly yq -i` (in-place edit) loses an explicit `!!float` tag on a scalar whose text looks like an integer (`!!float 2`), even when the query never touches that field — treating it as `!!int` instead. The identical input via stdin (no `-i`) resolves the tag correctly.

## Repro (confirmed on unmodified `main`, unrelated to #1090's own change)

```bash
$ printf 'a: !!float 2\n' > /tmp/t.yaml
$ succinctly yq -i '.b = (.a | tag)' /tmp/t.yaml
$ cat /tmp/t.yaml
a: 2
b: "!!int"
```

`.a` itself is also silently rewritten from `!!float 2` to bare `2` in the output, even though the query never assigns to `.a` — losing the tag on write, not just misreporting it on read.

```bash
$ printf 'a: !!float 2\n' | succinctly yq '.b = (.a | tag)'
a: 2.0
b: "!!float"
```

The same query via stdin correctly resolves `!!float` and (via the existing root-position decimal-preserving spelling) round-trips the type correctly.

## Why this matters more after #1090

#1090 added `!!float`-tag emission for nested computed whole floats (matching real yq's own spelling, `a: !!float 2`), so a `-i`-mode workflow that repeatedly edits the same float field (e.g. `-i '.a += 1'` run in a loop, or applied across multiple separate invocations) will now naturally produce this exact `!!float <int-looking-number>` shape as its own prior output — and the *next* `-i` edit on that file silently drops the type, defeating the whole point of #1090's fix for any workflow that edits the same field more than once via `-i`. Confirmed live: two successive `succinctly yq -i '.a += 1'` invocations on `a: 1.0` produce `a: !!float 2` then `a: 3` (type lost on the second edit) — real yq produces `a: !!float 2` then `a: !!float 3` (type preserved) for the same sequence.

## Root cause

[...]

Issue #949, opened 2026-08-15
yq: a genuinely computed whole-number float result shows a decimal point, unlike jq/yq


newhoggy:
**Severity: Low**

## Summary

Found while implementing #918 (which fixes an integer-valued YAML float scalar like `2.0` losing its float type at input parsing). Once #918's fix makes such a scalar correctly preserve its float-ness through to `.` (identity), a *genuinely computed* whole-number arithmetic result derived from it incorrectly keeps a decimal point in `succinctly yq`'s default/pretty output — where real `yq`/`jq` show the collapsed integer text, matching jq's own "computed numbers get canonical formatting, only values that reach output untouched keep their literal" convention (already the documented, correct behavior for `succinctly jq` and for `succinctly yq --input-format json`).

## Repro (verified against pinned yq v4.53.3 / jq 1.7.1)

```
$ echo '1.0' | yq '. + 1'
2
$ echo '1.0' | succinctly yq '. + 1'
2.0

$ echo '1.0' | succinctly yq -o json -I0 '. + 1'   # compact: correct
2
$ echo '1.0' | succinctly yq -o json '. + 1'       # pretty: wrong
2.0
```

Confirmed **not** reproducible via:
- `succinctly jq '. + 1'` on JSON input `1.0` — correctly gives `2`.
- `succinctly yq --input-format json '. + 1'` on the same JSON input — correctly gives `2`.
- `succinctly yq -o json -I0 ...` (compact mode) — correctly gives `2`.

So this is specific to **YAML-sourced input**, **non-compact ("pretty") output**, and appears for every arithmetic operator tried (`+`, `-`, `*`), not just addition.

## Root cause

Not fully traced. What's confirmed:
- `arith_add` (`src/jq/eval.rs`) already correctly calls `OwnedValue::into_plain_number()` on both operands before computing, collapsing any `NumberLiteral` marker before arithmetic — matching jq's "computed values lose their literal spelling" convention (#387's precedent). This function itself is not the bug.
[...]

PR #1179, opened 2026-08-17
fix(jq): emit !!float tag for nested computed whole floats

### LINKED ISSUE #1090: yq: nested computed whole floats have no !!float tag-emission path, unlike real yq
**Severity: Low**

## Summary

Discovered during code review of #949's fix (PR #976). Real yq v4.53.3 drops a computed whole float's decimal point (`2`, not `2.0`) only at document-root scalar position. Everywhere else — an object field, an array element, an `-i` in-place edit — it instead emits an explicit `!!float` tag to keep the value's type unambiguous on reparse:

```
$ echo '1.0' | yq '. + 1'              # root
2
$ printf 'a: 1.0\n' | yq '.a += 1'     # nested
a: !!float 2
```

succinctly's `OwnedValue` has no tag slot at all (only `Int`/`Float`/`NumberLiteral` variants), so there's no way to emit `!!float` on the write side. #949's fix works around this by falling back to the pre-existing decimal-point-preserving spelling (`a: 2.0`) for any nested computed float, which correctly avoids the *worse* regression (silently losing float-vs-int type on reparse) but isn't byte-identical to real yq's tagged spelling.

## Why a proper fix is more than a formatting change

The read/cursor-echo side already has the primitive a real fix would need: `YamlCursor`'s scalar re-emission path (`src/yaml/light.rs`, around where `self.explicit_tag()` is written verbatim before a scalar) already knows how to place a literal tag token before a value when re-emitting a `YamlCursor`-backed node. Wiring an equivalent capability onto `OwnedValue` emission (`emit_yaml_value`/`emit_yaml_value_at_depth` in `yq_runner.rs`, and `stream_owned_value_yaml`/`_at_depth` in `src/jq/stream.rs`) needs either:
- A tag field added to `OwnedValue::Float` (or a new variant), populated when a value is genuinely computed and ambiguous, or
- A narrower special-case: detect "this Float's decimal-stripped spelling would parse back as an int" (already computable — it's exactly `format_float_with_fraction`'s own `!buf.contains('.')` check) and prefix `!!float ` only in that case.

The ambiguity check itself is cheap and already exists in some form; the gap is purely in wiring a tag prefix through to output. Real yq's tag placement doesn't appear to be style-sensitive (measured: nested-in-block, nested-in-array, and nested-in-flow contexts all get the same unconditional `!!float ` prefix), which should simplify the wiring.

## Repro

```
$ printf 'a: 1.0\n' | succinctly yq '.a += 1'
a: 2.0      # succinctly (post-#949): type-preserving but not byte-identical
a: !!float 2  # real yq v4.53.3
```

## Suggested fix direction

Not investigated in depth beyond the above. The narrower "detect ambiguity, prefix a tag" approach is probably the smaller change; a full tag-carrying `OwnedValue` variant would be a bigger, more invasive refactor likely useful for other tag-preservation gaps too (if any exist elsewhere in the codebase — not surveyed here).

## Why this is separate from #949

#949's fix (PR #976) is specifically scoped to the decimal-point-dropping rule at root scalar position, and deliberately does *not* attempt tag emission for the nested case — its own depth-gating fallback exists precisely because this capability doesn't exist yet. This issue tracks building that capability as its own, larger piece of work.

newhoggy:
## Summary

Fixes #1090. Real yq drops a computed whole float's decimal point (`2`, not `2.0`) only at document-root scalar position:

```bash
$ echo '1.0' | yq '. + 1'
2
$ printf 'a: 1.0\n' | yq '.a += 1'
a: !!float 2
```

Anywhere else — an object field, an array element, an `-i` in-place edit — real yq instead emits an explicit `!!float` tag to keep the value's type unambiguous on reparse. succinctly's `OwnedValue` has no tag slot, so #949's fix worked around this by falling back to a decimal-preserving spelling (`a: 2.0`) for the nested case — type-safe (round-trips as `!!float`) but not byte-identical to the oracle.

## Fix

Adds `format_float_yq_yaml_nested`, a dedicated sibling of `format_float_yq_yaml` that reuses the existing `format_float_yq_with` threshold helper (shared scientific-notation-vs-decimal logic), used only at the two YAML-output-specific nested-position call sites:
- `emit_yaml_value_at_depth` (`src/bin/succinctly/yq_runner.rs`)
- `stream_owned_value_yaml_at_depth` (`src/jq/stream.rs`)

**Deliberately not added to `format_float_yq` or `format_float_with_fraction` themselves** — both are also yq's `-o json` float formatters (confirmed via call-site tracing: `format_float_yq` is called from `src/bin/succinctly/output.rs`'s JSON-encoding path, gated on yq mode but JSON *output*). Injecting YAML tag syntax into those shared functions would have corrupted `-o json` output — the exact class of cross-format regression a prior attempt at #868 hit for JSON/YAML input handling. Verified live that `-o json` output is unaffected by this change.

## Testing

Verified live against real yq v4.53.3 across every position/style combination: root (unchanged, #949), nested object field, array element (block and flow style), `-i` in-place, negative values, exact zero, and the scientific-notation threshold (never tagged — its own `e` marker is already unambiguous, confirmed real yq agrees).

- Updated `test_computed_whole_float_nested_yaml_output_keeps_type_949` and `test_computed_whole_float_inplace_stays_float_typed_949` (previously pinned the `a: 2.0`-style fallback; now assert the oracle-matched `a: !!float 2` spelling).
- Updated `test_stream_yaml_computed_whole_float_drops_decimal_point_949` (`src/jq/stream.rs` unit test) for the same reason.
- New: `test_nested_float_tag_edge_cases_1090` (negative values, exact zero, scientific-notation non-tagging, fractional non-tagging).

## A genuinely separate bug found while verifying `-i` mode

Confirmed live on **unmodified `main`** (unrelated to this fix): a *second* successive `-i` edit on a file already containing an explicit `!!float`-tagged whole number loses the tag entirely (`-i '.a += 1'` run twice on `a: 1.0` produces `a: !!float 2` then `a: 3`, not `a: !!float 3`). Real yq preserves the type across both edits. This is a pre-existing `-i`-mode tag-resolution bug (confirmed the same input via stdin resolves the tag correctly) — this fix's own output is correct and oracle-matched, but a workflow that repeatedly edits the same float field via `-i` will now hit this separate bug more often, since #1090's fix is what naturally produces the `!!float <int-looking-number>` shape as prior output. Filed as **#1176**; `test_inplace_second_edit_loses_float_tag_1176_known_gap` pins the current (buggy) behavior as a characterization test.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite; one unrelated `jq_cli_tests` failure reproduced as flaky under concurrent-suite contention, confirmed clean both in isolation and as part of a full clean run — see session notes)

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -20,15 +20,15 @@ use succinctly::jq::{
 };
 use succinctly::json::JsonIndex;
 use succinctly::yaml::{
-    format_float_yq_yaml, resolve_plain, resolve_tagged, stream_yaml_sequence, YamlCursor,
-    YamlIndex, YamlValue,
+    format_float_yq_yaml, format_float_yq_yaml_nested, resolve_plain, resolve_tagged,
+    stream_yaml_sequence, YamlCursor, YamlIndex, YamlValue,
 };
 
 use super::{FrontMatterMode, InputFormat, OutputFormat, YqCommand};
 use crate::front_matter;
 use crate::output::{
-    self, exit_codes, format_float_yq, ColorScheme, ControlEscape, DiagStyle, ErrorSink,
-    FloatStyle, InputLocation, JsonFormatOpts,
+    self, exit_codes, ColorScheme, ControlEscape, DiagStyle, ErrorSink, FloatStyle, InputLocation,
+    JsonFormatOpts,
 };
 
 /// yq's diagnostics carry no `(at <file>:<line>)` marker, so the yq paths have
@@ -1566,16 +1566,10 @@ fn emit_yaml_value_at_depth(
         // `2.0`) only at document-root scalar position -- issue #949.
         // Anywhere nested (an object field, an array element, an `-i`
         // in-place edit) it instead emits an explicit `!!float` tag to
-        // keep the value's type unambiguous on reparse (`a: !!float 2`);
-        // succinctly has no tag-emission mechanism to match that, so
-        // falling back to `format_float_yq`'s decimal-preserving spelling
-        // here is the safer of the two available choices -- it's not
-        // byte-identical to real yq, but unlike the bare, untagged `2`
-        // this fix originally used at every depth, it doesn't silently
-        // change the value's inferred type from float to int on
-        // reparse/round-trip through `-i`.
+        // keep the value's type unambiguous on reparse (`a: !!float 2`),
+        // which `format_float_yq_yaml_nested` now reproduces (#1090).
         OwnedValue::Float(f) if depth == 0 => format_float_yq_yaml(*f),
-        OwnedValue::Float(f) => format_float_yq(*f),
+        OwnedValue::Float(f) => format_float_yq_yaml_nested(*f),
         OwnedValue::NumberLiteral(NumberRepr::Float(f), _) if f.is_nan() || f.is_infinite() => {
             nonfinite_display_string::<YqSemantics>(*f).to_string()
         }
--- a/src/jq/stream.rs
+++ b/src/jq/stream.rs
@@ -29,7 +29,7 @@ use super::value::{
     assert_value_tree_depth, format_number_jq_compat, infinite_float_preview_text,
     jq_bare_float_display, NumberRepr, OwnedValue,
 };
-use crate::yaml::{format_float_with_fraction, format_float_yq, format_float_yq_yaml};
+use crate::yaml::{format_float_with_fraction, format_float_yq_yaml, format_float_yq_yaml_nested};
 
 /// A value that can be streamed directly to output without intermediate allocation.
 ///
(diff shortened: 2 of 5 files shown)


newhoggy:
Closing without merging — code review (10-way fan-out, cross-checked with my own live verification against real yq v4.53.3) confirmed three real problems, one of which is architectural and not cleanly fixable with the current `OwnedValue` representation.

**1. Genuine regression: `tonumber`-sourced (non-arithmetic) floats are now wrongly tagged.** Real yq's tagging rule isn't "any nested whole float" — it specifically depends on *how* the value was produced:

```
$ printf 'a: "2.0"\n' > f.yaml && yq '.b = (.a | tonumber)' f.yaml
a: "2.0"
b: 2.0
$ succinctly yq '.b = (.a | tonumber)' f.yaml   # this PR
a: "2.0"
b: !!float 2
```

But once that same value passes through genuine arithmetic, it *does* get tagged:
```
$ yq '.b = (.a | tonumber) + 0' f.yaml
b: !!float 2
```

And `tonumber` on an int-looking string (`"2"`, no decimal) produces a bare int throughout, untagged even after `+0`. This means real yq's Go implementation tracks some notion of "was this genuinely computed via arithmetic" that `OwnedValue::Float` (a bare `f64`, no provenance) has no way to represent — the same class of problem that blocked #868 (JSON/YAML deduplication needing provenance the type system doesn't carry). Before this PR, `tonumber`'s output happened to match the oracle (both untagged); this PR breaks that previously-correct case while fixing the arithmetic case.

**2. `-0.0` is wrongly tagged.** `(-0.0f64).to_string()` is `"-0"` (no `.`), so my ambiguity check incorrectly triggers. Real yq emits bare `-0` (go-yaml already resolves it as a float without disambiguation) — confirmed live, positive `0.0` *is* tagged but negative `-0.0` is not. A narrow, easily-fixable bug on its own, but compounds with finding 1's deeper problem.

**3. Amplifies #1176 from a latent bug into a live regression.** #1176 (filed alongside this PR) is real and pre-existing — but framing it as merely "pre-existing, unrelated" undersold it: before this PR, succinctly *never* emitted a `!!float <int-looking-number>` scalar as its own output, so #1176's read-side bug was unreachable in practice. This PR is what makes succinctly produce that exact shape as ordinary output, so a workflow that edits the same float field twice via `-i` (a common pattern) now silently loses the type on the second edit — precisely the defect class #949's own review rejected ("`a: 2` reparses as `!!int`... the worse defect for a data-format round trip"), reintroduced one edit later via a different mechanism.

Given the confirmed regression in (1) has no clean fix without provenance-tracking (an architectural change, not a point fix), and (2)/(3) compound it, I'm closing rather than shipping a fix that trades one set of oracle mismatches for another. Filed findings below for whoever scopes a proper fix next; #1090 stays open.

Issue #965, opened 2026-08-15
jq/yaml: remaining ResolvedScalar/StandardJson->OwnedValue duplication and coupling after #907


newhoggy:
**Severity: Low**

## Summary

#907's fix (PR #964) hoisted 3 of 4 known `ResolvedScalar -> OwnedValue` conversions into one `ResolvedScalar::to_owned_value`, and replaced a 4th, unrelated duplicate (`yq_runner.rs`'s hand-rolled `standard_json_to_owned`) with the crate's existing generic `to_owned()`. Multi-agent review of that fix surfaced several further, real-but-out-of-scope siblings and design questions, tracked here rather than expanding that PR's scope further.

## 1. `write_resolved_scalar_as_json`/`stream_resolved_scalar_as_json` (`src/yaml/light.rs`) are two more independent `ResolvedScalar` matches

Unlike the copies #907 fixed, these produce JSON *text* directly (not an `OwnedValue`), using a third strategy (`format_float_with_fraction` reformatting) to solve the same "whole-number float loses its `.0`" problem `is_preservable_float_literal` solves for the `OwnedValue` path. Not folded into `to_owned_value` since the output type differs. A future `ResolvedScalar` variant, or a future #918-style literal-fidelity bug, needs three matches updated (`to_owned_value` plus these two), not one.

## 2. `eval_generic.rs` has its own separate, unrelated `standard_json_to_owned`

A private function in `src/jq/eval_generic.rs` (~line 717) also converts `StandardJson -> OwnedValue`, coincidentally named the same as the one #907 deleted from `yq_runner.rs`, but behaviorally different (round-trips through `to_json_for_reindex()`, is NaN-sentinel aware). ~6 call sites in that file hand-roll object/array/number/string conversion instead of calling the crate's own `to_owned()`, defined ~600 lines above in the same file — the same duplication pattern #907 was filed to eliminate, one file over. No compile collision (module-private) but a real naming-confusion risk for anyone grepping the codebase.

## 3. `yaml_to_owned_value` (`yq_runner.rs`) and `yaml_value_to_owned` (`eval.rs`) remain near-identical outer functions

[...]

PR #1185, opened 2026-08-17
refactor(jq): rename eval_generic.rs's standard_json_to_owned

newhoggy:
## Summary

Addresses #965, item #2 only (see "Scope" below — the other 4 items are explicitly out of scope for this PR).

`src/jq/eval_generic.rs` has a private helper, `standard_json_to_owned`/`standard_json_to_owned_at_depth`, converting `StandardJson` (the JSON semi-index cursor value) to `OwnedValue`. It coincidentally shared its name with an unrelated, behaviorally-different `yq_runner.rs` helper that #907 deleted — a real naming-confusion risk for anyone grepping the codebase (`#965`'s own words), even though the two functions never collided at compile time (both module-private, no functional overlap) and neither was ever calling the other.

Renamed to `owned_from_standard_json`/`owned_from_standard_json_at_depth`, and updated the accompanying doc comment to explain the distinction from the crate-wide, `DocumentValue`-generic `to_owned()` defined ~700 lines above in the same file (this one is JSON-cursor-specific, used only by this module's own `eval_on_owned`/`eval_single` fallback arm).

Pure rename — zero behavior change, confirmed by the full test suite passing unmodified (no test assertions needed updating, since none referenced the old name from outside the module).

## Scope

#965 catalogs 5 separate items found during #907's own review, with an explicit priority order: "#1 (real duplication risk, moderate effort) > #4 (cheap to verify, high value if it reveals a regression) > #2 (rename for clarity, low effort) > #3/#5 (larger refactors, no proposed design yet)." This PR is item #2 only:

- **#1** (deduplicating `write_resolved_scalar_as_json`/`stream_resolved_scalar_as_json` in `src/yaml/light.rs`) touches dense, oracle-verified float-formatting logic (cross-referencing #993, #1008) — real risk of a subtle behavioral regression during consolidation, left for its own careful pass.
- **#3/#5** (further `yaml_to_owned_value`/`yaml_value_to_owned` consolidation, module-placement reconsideration) explicitly have "no proposed design yet" per the issue's own text.
- **#4** (benchmark verification of a possible allocation regression) needs dedicated benchmark hardware access, not something this PR attempts.

Two other implementation attempts this session (#868, #1090) both hit real, review-caught regressions in exactly this kind of "looks like a narrow fix" territory (YAML/JSON shared-code and float-formatting edge cases) — this PR deliberately stays inside the smallest, zero-behavior-change item from #965's own catalog rather than reaching for the higher-effort items.

## Testing

Full test suite passes unmodified. Word-boundary-safe rename (verified via `perl -pe 's/\bstandard_json_to_owned\b/.../g'`, not naive substring replacement) correctly left `CHANGELOG.md` and two `tests/yq_cli_tests.rs` comments referencing the historical, already-deleted `yq_runner.rs` function untouched — those describe a genuinely different function from a different file, not this one.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, clean, zero test changes needed)

### DIFF
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -818,19 +818,28 @@ fn into_lazy_items<V: DocumentValue>(
 
 /// Convert a StandardJson value to an OwnedValue.
 ///
+/// Named distinctly from the crate-wide, `DocumentValue`-generic [`to_owned`]
+/// above (#965): this one is JSON-cursor-specific, used only by this
+/// module's own `eval_on_owned`/`eval_single` fallback arm on a value it
+/// constructs internally (e.g. a `reduce`/`foreach` accumulator) -- not a
+/// general-purpose converter other modules should reach for. Before this
+/// rename it happened to share a name with an unrelated, now-deleted
+/// `yq_runner.rs` helper (#907), which was a real naming-confusion risk for
+/// anyone grepping the codebase, even though the two were never the same
+/// function and never collided at compile time (module-private on both
+/// sides).
+///
 /// Panics past [`MAX_NESTING_DEPTH`] levels of nesting (#1017) -- a third,
 /// independent copy of the cursor-to-`OwnedValue` conversion `to_owned`/
-/// `to_owned_cursor` already guard in this same file (#998); reached from
-/// `eval_on_owned`/`eval_single`'s fallback arm on a value this module
-/// constructs internally (e.g. a `reduce`/`foreach` accumulator), the same
-/// gap #998's own guards on the other two copies were added to close.
-fn standard_json_to_owned<W: Clone + AsRef<[u64]>>(
+/// `to_owned_cursor` already guard in this same file (#998), the same gap
+/// #998's own guards on the other two copies were added to close.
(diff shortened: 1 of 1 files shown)


newhoggy:
Code review (10-way fan-out) confirmed the rename is complete, word-boundary-safe, and behavior-preserving — no correctness findings. Individual angle reports converged on one recurring, low-severity doc-quality nitpick: the new naming-rationale paragraph, placed first, buried the load-bearing #1017/#998 panic-guard fact behind it. Also independently surfaced a genuinely useful finding (Reuse/Altitude angles): `owned_from_standard_json_at_depth` and the crate-wide `to_owned()` actually have two concrete behavioral divergences beyond just "different call sites" — NaN/Infinity-sentinel decoding and a different malformed-UTF-8-string fallback — which my original doc comment's vague "not a general-purpose converter" hand-wave didn't name.

Reordered the doc comment (panic-guard fact first again) and replaced the vague claim with these two concrete divergences. Rebased onto latest `main` (5 commits ahead), full pipeline re-verified clean.

PR #1222, opened 2026-08-17
fix(jq): strip insignificant leading zero/plus from extreme-literal mantissa

### LINKED ISSUE #1180: jq: normalize_extreme_literal_mantissa mis-normalizes an insignificant-leading-zero or leading-plus mantissa
## Summary

`normalize_extreme_literal_mantissa` (`src/jq/value.rs`, shared by the overflow and underflow extreme-exponent formatting paths) assumes a multi-digit `int_part` never starts with an insignificant leading zero, and unconditionally treats the *first character* as the significant leading digit:

```rust
let after_leading = &int_part[1..];
...
(int_part.len() as i64 - 1, &int_part[..1], rest)
```

A mantissa spelled with a leading zero (`"007"`) or a leading `+` sign gets a decimal point spliced in after that first (insignificant) character instead of after the first genuinely significant digit, producing non-canonical scientific notation. This is not gated on sign the way the `-` prefix already is (`s[..exp_pos].strip_prefix('-')`) — only a leading minus is stripped before this logic runs.

## Repro

Only reachable by calling the `pub(crate)` function `format_number_jq_compat` directly (see "Reachability" below), not via the CLI:

```rust
assert_eq!(format_number_jq_compat(b"007e-400"), "0.07E-398"); // should be "7E-400"
assert_eq!(format_number_jq_compat(b"007e400"), "0.07E+402");  // sibling bug, overflow direction
assert_eq!(format_number_jq_compat(b"+0e-400"), "+.0E-399");   // leading '+' also mishandled
```

## Reachability

`succinctly jq`/`succinctly yq`'s own literal construction sites — JSON's `number_literal()` and YAML's preservable-float gate — already reject a leading-zero or leading-plus numeric spelling via strict grammar validation before ever constructing a `NumberLiteral`, so this isn't reachable through any document input processed by the CLI today. It's only reachable by a caller of the `pub fn format_number_jq_compat` outside that gate, and the function has no documented precondition ruling this out — it silently mis-formats rather than erroring or asserting.

## Suggested direction

Either (a) document `format_number_jq_compat`'s precondition (RFC-8259-canonical mantissa spelling, no leading zeros/plus) and add a `debug_assert!` at entry consistent with how other trust-boundary gaps in this file are handled (see #1062's `debug_assert_eq!` on `NumberLiteral`'s repr), or (b) make `normalize_extreme_literal_mantissa` itself skip leading zeros/plus when locating the significant leading digit, so the function is correct for its full documented input space rather than just the grammar-validated subset that reaches it today.

Found during code review of PR #1173 (#1099).

newhoggy:
## Summary

`normalize_extreme_literal_mantissa` (shared by the overflow and underflow extreme-exponent formatting paths in `format_number_jq_compat`) unconditionally treated the raw integer part's *first character* as the mantissa's significant leading digit. A mantissa spelled with a leading zero (`007`) or a leading `+` sign got the decimal point spliced in after that insignificant character instead of after the first genuinely significant digit — producing non-canonical scientific notation, or in the `+` case, leaking the sign character straight into the output.

## Fix

- `split_mantissa` now strips a leading `+` the same way it already stripped `-` (only `-` was handled before).
- `normalize_extreme_literal_mantissa` trims leading zeros from the integer part (`int_part.trim_start_matches('0')`) before classifying its magnitude or picking the significant leading digit — this both fixes the leading-zero case and subsumes the previous `int_part.is_empty() || int_part == "0"` special case (an int part of all zeros now trims to empty either way).

```
format_number_jq_compat(b"007e-400")  // 0.07E-398 -> 7E-400
format_number_jq_compat(b"007e400")   // 0.07E+402 -> 7E+400
format_number_jq_compat(b"+0e-400")   // +.0E-399  -> 0E-400
```

## Reachability

Not reachable via the CLI today — succinctly's own JSON/YAML literal-construction grammars already reject a leading zero or leading `+` before ever building a `NumberLiteral`. Only reachable by a caller of the `pub` `format_number_jq_compat` outside that gate. Verified the existing oracle-pinned scientific-notation cases (`123e400`, `12.34e400`, `100e400`'s trailing-zero preservation, etc.) are unaffected since `trim_start_matches` only strips from the very start.

Fixes #1180.

## Test plan
- [x] `cargo test --lib jq::value` — new + existing tests pass
- [x] `cargo test --features cli,simd,regex,serde` (full suite) — green
- [x] `cargo clippy --all-targets --features cli,simd,regex,serde,broadword-yaml -- -D warnings` — clean
- [x] `cargo check --no-default-features` (no_std) — builds
- [x] Local patch coverage: 100% (17/17 new lines)

### DIFF
--- a/src/jq/value.rs
+++ b/src/jq/value.rs
@@ -342,10 +342,7 @@ fn split_mantissa(s: &str, exp_pos: usize) -> (&str, &str) {
     // magnitude as `-` already was here, and (unlike `-`) wasn't stripped at
     // all before #1180 -- a leading `+` in `int_part` was misread as its
     // significant leading digit.
-    let mantissa = raw
-        .strip_prefix('-')
-        .or_else(|| raw.strip_prefix('+'))
-        .unwrap_or(raw);
+    let mantissa = raw.strip_prefix(['-', '+']).unwrap_or(raw);
     mantissa.split_once('.').unwrap_or((mantissa, ""))
 }
 
@@ -375,10 +372,14 @@ fn normalize_extreme_literal_mantissa(s: &str, exp_pos: usize) -> Option<(String
     // mantissa of unbounded length must not turn into unbounded work here
     // just because only a handful of its leading digits will ever survive
     // that later truncation anyway. Bounding the *copy* is enough: `shift`
-    // below only ever needs `int_part.len()` (an O(1) property read, not a
-    // scan), so truncating what gets copied into `rest` doesn't touch the
-    // exponent math's correctness, only how many digits past the first one
-    // actually get rendered.
+    // below only ever needs `int_part.len()` (an O(1) property read once
+    // leading zeros are trimmed -- the `trim_start_matches('0')` just below
+    // is itself an O(k) scan over any leading-zero run, #1180), so
+    // truncating what gets copied into `rest` doesn't touch the exponent
+    // math's correctness, only how many digits past the first one actually
+    // get rendered. The trim doesn't change the function's overall cost
+    // class either way: `format_number_jq_compat`'s own `s.parse::<f64>()`
+    // ahead of this call already scans every byte of `s` once.
     const MAX_RENDERED_MANTISSA_DIGITS: usize = 32;
 
     let (int_part, frac_part) = split_mantissa(s, exp_pos);
@@ -389,9 +390,9 @@ fn normalize_extreme_literal_mantissa(s: &str, exp_pos: usize) -> Option<(String
     // leading-dot literal (`.5e400`) already has an empty `int_part`, and an
     // int_part of nothing-but-zeros (`0`, `000`) trims to empty the same
     // way -- both fall into the magnitude-< 1 branch below.
-    let sig_int_part = int_part.trim_start_matches('0');
+    let int_part = int_part.trim_start_matches('0');
 
-    let (shift, leading, rest): (i64, &str, String) = if sig_int_part.is_empty() {
+    let (shift, leading, rest): (i64, &str, String) = if int_part.is_empty() {
         // Shift right to the first nonzero fractional digit. Genuinely
         // all-zero mantissa (`0.0e400`/`0.0e-400`) has no nonzero digit to
         // shift to -- `?` propagates that as `None`.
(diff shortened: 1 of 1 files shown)


newhoggy:
## Code review summary

9 finder agents + 2 additional angles reviewed this PR. No correctness bugs found — the fix and its interaction with both callers (`format_overflow_literal_mantissa`/`format_underflow_literal_mantissa`) were independently hand-traced and empirically probed by multiple agents against inputs well beyond the committed test cases (huge leading-zero runs, mixed sign/zero combinations, the fractional-only underflow fallback), all confirming correct, panic-safe output.

**Fixed in this PR** (second commit):
- Simplified `split_mantissa`'s sign-stripping to a single `strip_prefix(['-', '+'])` call instead of a chained `strip_prefix`/`or_else`.
- Shadowed `int_part` instead of introducing a new `sig_int_part` binding, matching this file's existing shadowing idiom.
- Corrected a doc comment that still claimed the shift math needed only an O(1) `int_part.len()` read — the new `trim_start_matches('0')` is an O(k) scan over any leading-zero run (converged finding from 2 independent agents). Noted this doesn't change the function's overall cost class, since `format_number_jq_compat`'s own `s.parse::<f64>()` already scans the whole literal first.
- Added a bounded-cost test for a huge mantissa *with* a huge leading-zero run — the existing huge-mantissa test uses an all-`9`s mantissa, which never exercises the new scan path at all.

**Filed as follow-up** (#1224, out of scope for this fix):
- `format_number_jq_compat`'s plain-integer/plain-decimal (no-exponent) branches have the identical unfixed leading-zero/`+` defect this PR only closed for the exponent-notation path — verified live by one finder. Needs care to avoid corrupting legitimate literal-preservation output (a lone leading `0`, trailing zeros).
- `OwnedValue::NumberLiteral`/`Literal::NumberLiteral` are public enum variants with no validating constructor, so "NumberLiteral text is canonical" is enforced by convention at ~4 internal call sites, not by the type — a broader reachability surface than this PR's own doc comment ("only reachable by calling this `pub` function directly") implies.

**Considered, not acted on**: several duplication observations (this file's own `parse_literal_exponent` sign-check idiom, `jq_runner.rs`'s unrelated `--argjson` leading-zero repair, `is_valid_number`'s grammar) — all single-line stdlib idioms in different contexts serving different purposes; extracting a shared helper for any of them looked like premature abstraction rather than a genuine "duplicated predicate that could silently diverge" risk.

Local verification after the review-fix commit: full test suite green, both clippy invocations clean, fmt clean, no_std builds, 100% local patch coverage.

Issue #1226, opened 2026-08-17
jq: nonzero-mantissa notation threshold also uses the written exponent instead of magnitude, not just the zero-mantissa case (#1207)


newhoggy:
**Severity: Medium**

Found while implementing #1207 (zero-mantissa literal picks scientific/decimal/plain-integer notation on the wrong threshold).

## Summary

#1207's own fix is scoped correctly to `value == 0.0` — but the same root cause (`format_number_jq_compat`'s `exp == 0`/`(-5..0).contains(&exp)` checks use the literal's raw **written** exponent, not its actual base-10 magnitude) also misclassifies **nonzero** mantissas, including the simplest possible case: a single significant digit with no fraction at all.

## Repro (verified against pinned jq 1.7.1)

```
$ echo '5e-6'   | jq -c .   ->  0.000005     (decimal)
$ echo '5e-6'   | succinctly jq -c .   ->  5E-6         (scientific, wrong)

$ echo '1e-7'   | jq -c .   ->  1E-7         (scientific -- correctly outside the window)
$ echo '123e-6' | jq -c .   ->  0.000123     (decimal)
$ echo '123e-6' | succinctly jq -c .   ->  1.23E-4      (scientific, wrong)
$ echo '100e-7' | jq -c .   ->  0.0000100    (decimal, trailing zeros of "100" preserved)
$ echo '100e-7' | succinctly jq -c .   ->  1E-5         (scientific, wrong -- also drops the trailing zeros)
$ echo '999e-6' | jq -c .   ->  0.000999     (decimal)
$ echo '999e-6' | succinctly jq -c .   ->  9.99E-4      (scientific, wrong)
```

`123e-5`, `1e-3`, `12.34e-3`, `999e-5` all already match jq correctly today — the bug only manifests once the raw written exponent falls outside `-5..0` while the mantissa's actual **magnitude** (written exponent shifted by however many extra integer digits precede the decimal point) is still within jq's real decimal window.

## Root cause

[...]

Issue #1178, opened 2026-08-17
jq: genuinely-zero-mantissa literal doesn't shift printed exponent by fractional-zero count


newhoggy:
## Summary

`format_number_jq_compat`'s genuinely-all-zero-mantissa fallback (`format_underflow_literal_mantissa`'s `None` arm, `src/jq/value.rs`) echoes the literal's own written exponent unchanged:

```rust
None => assemble_scientific(sign, "0", parse_literal_exponent(&s[exp_pos + 1..])),
```

Real jq, however, shifts the *printed* exponent by the mantissa's fractional-zero-digit count even when the mantissa is all zeros — the same normalization it applies to nonzero mantissas.

## Repro

```console
$ echo '0.000e-400' | jq -c .
0E-403
$ echo '0.000e-400' | succinctly jq -c .
0E-400

$ echo '0.0e400' | jq -c .
0E+399
$ echo '0.0e400' | succinctly jq -c .
0E+400

$ echo '0.00e-400' | jq -c .
0E-402
$ echo '0.00e-400' | succinctly jq -c .
0E-400
```

A no-fraction spelling (`0e-400`) is unaffected and already correct — the gap is specifically extra fractional zero digits after the decimal point.

## Why this is separate from #1099

This is a pre-existing bug in the genuinely-zero-mantissa fallback branch, untouched by #1099 (#1173) — that PR only added a check *before* this branch to route genuinely-nonzero underflowed mantissas elsewhere; it didn't change this branch's own exponent handling. #1099's own regression test for this branch (`test_number_literal_zero_mantissa_extreme_exponent_still_zero_1099`) only exercises no-fraction spellings, which is why this gap wasn't caught by that PR's test suite.

## Suggested direction

The fractional-zero-digit count (`frac_part.len()` when the whole mantissa is zero) is already computed once you're inside `normalize_extreme_literal_mantissa`'s zero-mantissa `None` path — before returning `None`, the shift amount for a genuinely-zero mantissa is simply `-(frac_part.len() as i64)` (matching real jq's apparent rule of treating the mantissa as if its first digit were the "1s place" regardless of whether that digit is zero). Threading that shift out (e.g. `Option<(Option<String>, i64)>`, or a small dedicated return type) instead of a bare `None` would let both callers apply it uniformly.

Found during code review of PR #1173 (#1099).

newhoggy:
🔧 Claimed by agent `aiden` — branch `issue-1178-zero-mantissa-exponent-shift`

Issue #1099, opened 2026-08-17
jq: extreme-magnitude float literals mis-render (1e-400 -> 0E-400) or lose precision beyond ~15-17 digits


newhoggy:
**Severity: Low**

## Summary

Found while implementing #1058 (pre-existing bugs in shared number-rendering
code, confirmed unrelated to and unaffected by that PR — reproduces
identically on `main` before #1058 via plain document input, e.g.
\`echo '1e-400' | succinctly jq '.'\`).

Two distinct issues in \`src/jq/value.rs\`'s \`format_number_jq_compat\`:

**1. Underflow misrenders the mantissa, not just the value:**

```
$ echo '1e-400' | succinctly jq '.'
0E-400
$ echo '1e-400' | jq '.'
1E-400
```

A literal whose \`f64\` parse underflows to exactly \`0.0\` (nonzero mantissa
in the source text) can't be told apart, inside the \`value == 0.0\` branch,
from a literal whose mantissa was *textually* zero — so the function
stitches the correct exponent onto a wrong \`"0"\` mantissa. This is worse
than a plain \`0\`: it looks precise and confidently wrong rather than
visibly approximated.

**2. High-precision mantissas lose precision in the "normalized scientific
notation" branch:**

```
$ echo '1.23456789012345678901234567890e300' | succinctly jq '.'
1.23456789012346E+300
$ echo '1.23456789012345678901234567890e300' | jq '.'
1.2345678901234568E+300   # real jq's own f64-precision rendering, still more digits preserved
```

The branch recomputes the mantissa via \`log10\`/\`pow\`/division on the
already-\`f64\`-parsed value rather than deriving it from the source digit
string, losing precision beyond ~15-17 significant digits even when the
exponent itself is well within \`f64\` range.

## Scope note

Both are in shared, general-purpose number-formatting code reachable from
any input path (document input, filter literals, \`--argjson\`) — not
specific to any one of them. \`format_overflow_literal_mantissa\` (the
sibling function for the *overflow* case, \`1e400\`) is already correctly
oracle-verified per its own doc comment; these two gaps are in the adjacent
underflow/high-precision branches it doesn't cover. Filing to make the gap
trackable, not blocking.

newhoggy:
🔧 Claimed by agent `cadence` — branch `issue-1099-underflow-mantissa-precision`.

[...]

PR #1227, opened 2026-08-17
fix(jq): zero-mantissa literal picks notation on its shifted, not written, exponent

### LINKED ISSUE #1207: jq: zero-mantissa literal picks scientific/decimal/plain-integer notation on the wrong threshold
**Severity: Low**

## Summary

Found while implementing #1178 (fixing `format_underflow_literal_mantissa`'s exponent-shift math for a genuinely-zero-mantissa literal). This is a distinct, separate bug in `format_number_jq_compat`'s own dispatch logic (`src/jq/value.rs`) — the *decision* of whether to use scientific notation, plain-integer, or expanded-decimal formatting at all for a zero-mantissa literal, as opposed to #1178's narrower scope (the exponent value printed once already inside the scientific-notation path).

## Repro (verified against pinned jq 1.7.1) — not fixed by #1178

```
$ echo '0.0e1'  | jq -c .   ->  0          (plain integer)
$ echo '0.0e2'  | jq -c .   ->  0E+1       (scientific)
$ echo '0.0e-1' | jq -c .   ->  0.00       (expanded decimal, 2 zeros)
$ echo '0.0e-5' | jq -c .   ->  0.000000   (expanded decimal, 6 zeros)
$ echo '0.0e-6' | jq -c .   ->  0E-7       (scientific)
```

succinctly (both before and after #1178's own fix) always routes a zero-mantissa literal with a nonzero written exponent straight to the scientific-notation formatter, regardless of magnitude:

```
$ echo '0.0e1'  | succinctly jq -c .   ->  0E+0   (should be plain 0)
$ echo '0.0e-1' | succinctly jq -c .   ->  0E-2   (should be plain decimal 0.00)
```

## Root cause

`format_number_jq_compat`'s dispatch (`src/jq/value.rs`) has an `(-5..0).contains(&exp)` check (using the literal's own *written* exponent) that converts a **nonzero**-mantissa literal to expanded decimal for small-magnitude exponents — but that check runs *before* the `value == 0.0` branch, so it never applies to a zero-mantissa literal at all; every zero-mantissa literal with a nonzero exponent falls straight through to `format_underflow_literal_mantissa`, unconditionally scientific.

Real jq's own equivalent decision, characterized live for exponents -10..10 (see the issue conversation for the full table), does NOT follow the same threshold — its own "expanded decimal" window for a zero mantissa looks like written exponents `-5..=-1` (matching #1178's shift formula: 2/3/.../6 zeros after the point, i.e. `-written_exp + 1` places), its own "plain 0" case is written exponent exactly `+1` specifically (not `0`, which is already handled by the *unshifted* `exp == 0` fast path above), and everything else uses scientific notation with the #1178-fixed shifted exponent.

## Scope note

Filing to make this trackable, not blocking — #1178's own fix is correctly scoped to the shift math inside the already-scientific-notation path; this is a separate bug in whether that path is even reached. Needs its own investigation to fully characterize jq's exact threshold rule (the "plain 0 at written exp +1 specifically" quirk in particular looks like it may itself be an artifact of jq's own internal formatting, not an obviously principled rule) before implementing.

newhoggy:
## Summary

`format_number_jq_compat`'s `exp == 0` and `(-5..0).contains(&exp)` fast paths decide decimal/bare/scientific notation from the literal's raw **written** exponent. That's correct for a nonzero mantissa in that range, but wrong for a genuinely-zero-mantissa literal, whose real notation threshold is its **shifted** exponent (written exponent minus the mantissa's own fractional-zero-digit count — the same shift math #1178 already applies once the scientific path is reached). Every zero-mantissa literal with a nonzero written exponent used to go straight to unconditional scientific notation, regardless of how small the resulting shifted exponent actually was.

```
$ echo '0.0e1'  | jq -c .   ->  0          (plain integer)
$ echo '0.0e-1' | jq -c .   ->  0.00       (expanded decimal)
$ echo '0.0e1'  | succinctly jq -c .  (before)  -> 0E+0   (wrong)
$ echo '0.0e-1' | succinctly jq -c .  (before)  -> 0E-2   (wrong)
```

## Fix

New `format_zero_value_literal` intercepts `value == 0.0` before either raw-exponent check runs. It reuses `normalize_extreme_literal_mantissa` to tell a genuinely-zero mantissa apart from a nonzero mantissa that merely underflowed to `0.0` during parsing (#1099's existing case — always scientific regardless of its own shifted exponent, unaffected by this fix):

- Genuinely-zero mantissa: dispatch on the shifted exponent — `0` → bare integer, `-6..=-1` → decimal expansion with the matching zero count, else → scientific (unchanged mechanism, just reached correctly now).
- Underflowed nonzero mantissa: unconditionally scientific, same as before.

`format_underflow_literal_mantissa`'s own `None` arm — now unreachable, since `value == 0.0` is intercepted earlier — simplifies to match `format_overflow_literal_mantissa`'s existing `unreachable!()` shape.

## Verification

Oracle-verified against pinned jq 1.7.1 across the full shifted-exponent boundary (`-8..=2`) and multiple mantissa spellings (with/without fraction digits, various fractional-zero-digit counts, both signs). All 36 combinations tested live match exactly; zero regressions on the pre-existing #1099/#1177/#1178/#1180 test suite.

**Out of scope, filed as #1226**: the same written-vs-shifted-exponent bug also misclassifies **nonzero** mantissas (`5e-6` → `5E-6` instead of `0.000005`; `123e-6`/`100e-7`/`999e-6` similarly wrong) — a distinct, broader defect discovered while characterizing this fix's boundary, left out of this PR's scope.

Fixes #1207.

## Test plan
- [x] `cargo test --lib jq::value` — new + existing tests pass
- [x] `cargo test --features cli,simd,regex,serde` (full suite incl. jq/yq golden conformance) — green
- [x] `cargo check --no-default-features` (no_std) — builds
- [x] Local patch coverage: 97.3% (36/37; the one uncovered line is a provably-dead `unreachable!()`, same pattern as the sibling `format_overflow_literal_mantissa`)

### DIFF
--- a/src/jq/value.rs
+++ b/src/jq/value.rs
@@ -227,7 +227,7 @@ pub fn format_number_jq_compat(raw: &[u8]) -> String {
     // stays necessary for a *nonzero* integer-valued literal like `-5e0`,
     // just no longer for `-0e0` specifically).
     if value == 0.0 {
-        return format_zero_value_literal(s, exp_pos, value.is_sign_negative());
+        return format_near_zero_literal(s, exp_pos, value.is_sign_negative());
     }
 
     // For e0 or e-0, jq eliminates the exponent
@@ -271,16 +271,17 @@ pub fn format_number_jq_compat(raw: &[u8]) -> String {
     // smallest representable subnormal), making `abs_value / 0.0 = +inf`
     // and rendering the mantissa as the literal text `"inf"`, which isn't
     // valid JSON at all. Routing every subnormal through the same
-    // string-based path `format_zero_value_literal` uses for exact-zero
+    // string-based path `format_near_zero_literal` uses for exact-zero
     // underflow sidesteps both: that path derives the mantissa from `s`'s
     // own source digits, never from `f64` arithmetic that's already lost
     // precision by this magnitude. Always scientific here (the #1207
     // small-magnitude decimal window applies only to a *genuinely*
     // zero-mantissa literal, never to one merely underflowed to zero) --
     // a subnormal `value` can only come from a nonzero mantissa in the
-    // first place.
+    // first place, so `format_near_zero_literal` always takes its `Ok`
+    // arm for this caller.
     if !value.is_normal() {
-        return format_underflow_literal_mantissa(s, exp_pos, value.is_sign_negative());
+        return format_near_zero_literal(s, exp_pos, value.is_sign_negative());
     }
 
     let abs_value = value.abs();
@@ -341,13 +342,10 @@ fn parse_literal_exponent(exp_text: &str) -> i64 {
 }
 
 /// Split a literal's mantissa text (`s[..exp_pos]`, sign stripped) into its
-/// integer and fractional parts -- shared by
-/// [`normalize_extreme_literal_mantissa`] and
-/// [`format_underflow_literal_mantissa`]'s own genuinely-all-zero-mantissa
-/// fallback (#1178), so "how do we split the mantissa out of `s`" has
-/// exactly one implementation instead of two hand-copied ones that must be
-/// kept in sync by convention (#106) -- the same duplicated-predicate shape
-/// this codebase has already been bitten by once.
+/// integer and fractional parts -- the sole implementation
+/// [`normalize_extreme_literal_mantissa`] builds on, so "how do we split the
+/// mantissa out of `s`" has exactly one implementation rather than one
+/// hand-copied at each of its own call sites (#106).
 fn split_mantissa(s: &str, exp_pos: usize) -> (&str, &str) {
     let raw = &s[..exp_pos];
     // Strip either sign: `+` is as insignificant to the mantissa's own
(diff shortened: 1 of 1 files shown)


newhoggy:
## Code review summary

10 finder agents reviewed this PR (the orchestrator's own synthesis pass hit the account's session usage limit mid-run; synthesized directly from the finder reports instead). No correctness bugs were found in the diff itself — multiple agents independently built the CLI, ran ~700-800 systematically-generated + adversarial cases against the pinned real jq 1.7.1 oracle, verified the reindex round-trip (`--slurp`/`-S`/`-a`), and confirmed the `unreachable!()` invariant holds. Several agents also confirmed a `00.0e-5`-style CLI mismatch is a pre-existing, unrelated JSON-tokenizer leniency gap (not this diff), reproducing identically on `origin/main`.

**Fixed in this PR** (second commit), converged independently across 3+ angles:
- `format_underflow_literal_mantissa` and `format_zero_value_literal`'s `Ok`/`Some` arm were byte-identical (sign → `normalize_extreme_literal_mantissa` → `assemble_scientific`). Merged into one `format_near_zero_literal`, called from both dispatch sites (`value == 0.0` and `!value.is_normal()`), removing the duplicated doc comments and the triplicated "sign+normalize+assemble" skeleton.
- `normalize_extreme_literal_mantissa` now returns `Result<(String, i64), i64>` (was `Option`), with `Err(frac_len)` on the all-zero-mantissa path — its one real consumer of that path was re-deriving `frac_part` via a second `split_mantissa` call on text already split internally moments earlier. Surfacing `frac_len` directly removes the redundant scan.
- Updated `normalize_extreme_literal_mantissa`'s own doc comment, which still named the now-deleted function as a caller and didn't note which caller actually exercises the `Err` path.

**Enriched #1226** (already filed before review started, out of scope for this PR): review independently rediscovered the written-vs-shifted-exponent bug and found two more manifestations — the `exp == 0` branch's `format!("{value}")` fallback can never emit scientific notation (`0.0000001e0` → `1E-7` in jq, `0.0000001` here), and the general log10/pow scientific path computes the correct shifted exponent but never re-checks it against the decimal window (`0.0001e2` → `0.01` in jq, `1E-2` here). Added both to #1226 with repro.

Local verification after the review-fix commit: full test suite green (679+ tests across lib/CLI/golden/doctest), both clippy invocations clean, fmt clean, no_std builds, 100% local patch coverage (40/40 new lines), and the fix's own 49-case oracle sweep still matches jq exactly after the refactor.

Issue #1116, opened 2026-08-17
yq: chained scalar-slice-assignment (.foo[S:E]) no-ops too, but del() deletes the whole parent key


newhoggy:
**Severity: Low**

## Summary

Found while implementing #1101 (deliberately out of that fix's scope — a
different code shape with a materially more complex real-yq behavior for
`del()` specifically).

#1101 fixed a **bare root** scalar-slice-assignment (`5 | .[0:1] = 99` →
`5`, no-op). Live-probing against real yq v4.53.3 found the no-op also
applies through a **chained** path (`.foo[S:E]` where `.foo` is itself a
scalar) for `=`/`|=`/`+=` — but `del()`'s chained case is a *different*,
more complex rule: it deletes the *entire parent key*, not a no-op.

```
$ echo '{"a":5,"b":6}' | yq -o=json '.a[0:1] = 99'
{"a": 5, "b": 6}                      # no-op, same as the bare-root case

$ echo '{"a":5,"b":6}' | yq -o=json 'del(.a[0:1])'
{"b": 6}                              # .a is gone entirely, NOT a no-op

$ echo '{"x":{"a":5,"b":6}}' | yq -o=json 'del(.x.a[0:1])'
{"x": {"b": 6}}                       # same at any nesting depth
```

succinctly's current behavior for both shapes is unchanged by #1101 (still
errors with "Cannot index number with object"), since #1101 deliberately
scoped its fix to a bare `Expr::Slice`/`Expr::Optional(Expr::Slice)` path
only (see `is_yq_scalar_slice_assign_path`'s doc comment in
`src/jq/eval.rs`).

## Scope note

This needs its own design, not a small extension of #1101's fix:
- `=`/`|=`/`+=`'s chained case is a straightforward extension of the
  existing no-op rule to any path ending in a slice-on-scalar, not just
  the bare-root case.
- `del()`'s chained case is a *different* rule (delete the parent key
  entirely) that would need its own implementation in `delete_at_path`,
  not just a wider version of the existing no-op check.

Not yet investigated further — filing to make the gap trackable.

newhoggy:
🔧 Claimed by agent `cadence` — branch `issue-1116-chained-scalar-slice-assign`.

Per the issue's own scope note, this has two independent parts:
[...]

Issue #1101, opened 2026-08-17
yq: assigning through a scalar slice target (.[S:E] = X) errors instead of real yq's silent no-op


newhoggy:
**Severity: Low**

## Summary

Found while implementing #1065 (deliberately out of that fix's scope — a
different code path with different, unverified semantics).

#1065 fixed *read-time* slicing of a scalar (`5 | .[0:1]?` → `[]`, matching
real yq). Real yq's *assignment*-target behavior for the identical shape is
different again — a silent no-op, not an error and not array synthesis:

```
$ echo '5' | yq -o=json '.[0:1] = 99'
5
$ echo '5' | succinctly yq -o json '.[0:1] = 99'
Error: Cannot index number with object
```

## Root cause

`src/jq/eval.rs`'s `resolve_slice_expr` (used for assignment-path
resolution) calls the same `slice_owned_value` helper #1065's fix
deliberately left untouched, specifically because this exact divergence was
found while implementing that issue — `resolve_slice_expr`'s own real-yq
behavior needed live verification before assuming it should also become
"treat as empty container," and that verification showed it's neither
that nor the pre-existing error; it's a full no-op that #1065's approach
(`Ok(Some(OwnedValue::Array(Vec::new())))`) would have gotten wrong (it
would try to write into the synthesized empty array, producing `[99]`
instead of leaving `5` untouched).

## Scope note

Needs its own live-probing across target shapes (does the no-op also apply
to `|=`, `+=`, and the other compound-assignment operators? what about
`path(.[0:1])` on a scalar?) before implementing — not yet investigated
further. Filing to make the gap trackable, not blocking.

newhoggy:
Claimed by agent `cadence`. Working on branch `issue-1101-yq-slice-assign-scalar-noop`.

PR #1228, opened 2026-08-17
fix(jq): yq field/index assignment no-ops on a scalar root

### LINKED ISSUE #1181: yq: field/computed-index assignment on a scalar target still errors, unlike slice assignment
**Severity: Low**

## Summary

Found during code review of #1117's fix (PR #1174). #1117 fixed `succinctly yq`'s slice-write no-op (from #1101/#1116) for a *computed* slice bound on a scalar target (`.[$a:$b] = 99`). While verifying that fix, review turned up that real yq's scalar-target no-op is not actually specific to slices at all — it applies uniformly to plain field/index assignment too, which `succinctly yq` still doesn't implement.

```
$ echo 5 | yq -o=json '.a = 99'
5                                      # no-op

$ echo 5 | yq -o=json '.[0] = 99'
5                                      # no-op

$ echo 5 | yq -o=json '0 as $k | .[$k] = 99'
5                                      # no-op

$ echo 5 | succinctly yq -o=json '.a = 99'
Error: Cannot index number with string   # wrong: errors instead

$ echo 5 | succinctly yq -o=json '.[0] = 99'
Error: Cannot index number with number   # wrong: errors instead

$ echo 5 | succinctly yq -o=json '0 as $k | .[$k] = 99'
Error: Cannot index number with number   # wrong: errors instead
```

## Root cause

`resolve_index_expr` (`src/jq/eval.rs`) computes its scalar-target failure via
`key_to_path_component`, which has no `S: EvalSemantics`/yq-mode gate at all — its
catch-all errors unconditionally for both jq and yq. The write-time walker
`get_path_mut` that field/index writes ultimately reach isn't generic over `S`
either, so it structurally can't gate on yq mode the way `through_slice` (the
slice-only sibling, with explicit `scalar_noop`/`container_noop` parameters) does.

This is the same class of gap #1101/#1116/#1117 already closed for slice targets,
just never implemented for plain field/index targets — not a regression from any
of those fixes, just an adjacent, unaddressed shape.

## Scope note

Filing to make this trackable, not blocking. Fixing this needs its own design: at
minimum, threading a yq-mode-aware no-op check through `key_to_path_component`'s
error path and giving `get_path_mut` (or its caller) the same `S`-awareness
`through_slice` already has — plus deciding whether container targets (not just
scalar) exhibit the same no-op for a plain field/index the way #1142 found for
slices.

newhoggy:
## Summary

Real yq's scalar-target no-op — already implemented for slice writes via #1101/#1116/#1142 — also applies to plain field/index writes: `.a = v`, `.[0] = v`, and a computed key (`$k as $k | .[$k] = v`) all silently no-op on a scalar root in real yq, but `succinctly yq` still raised `Cannot index number with ...` for all three.

Root cause: `key_to_path_component` had no yq-mode gate at all, and there's a second, separate `index_one_owned` call right after it in `resolve_index_expr` (used to compute the next value for chained resolution) that also had no gate — the computed-key case kept failing even after the first was patched, since that second call unconditionally errors on a scalar target regardless of what `key_to_path_component` decided. `update_path` (the `|=`/compound-assign walker) needed its own separate fix too, since its Field/Index arms never checked the existing `scalar_noop` machinery at all, and — verified live — real yq no-ops `-=`/`*=` for this shape too, unlike the slice case where those two operators genuinely error.

```
$ echo 5 | yq -o=json '.a = 99'
5                                      # no-op (real yq)
$ echo 5 | succinctly yq -o=json '.a = 99'
5                                      # now matches (was: Error: Cannot index number with string)
```

## Scope

Scoped to a scalar *root* only — mirroring #1101's original scope before #1116 later generalized it to any chain depth. A chained scalar target (`.a.b = 99`, `.[$k].foo = 99`) also no-ops in real yq but is **not** fixed here — `get_path_mut` (the parent-navigation walker a multi-component path resolves through) has no yq-mode awareness at all, unlike `set_path`'s own terminal-component arms, and needs its own design the way #1116 got. Filed as a follow-up.

Also found and **not** replicated here: real yq silently discards the RHS/filter entirely for this no-op, including a genuinely erroring one (`.a = error("boom")` on `5` stays `5`, not an error) — unlike the slice no-op, which does propagate a genuinely erroring RHS. `eval_assign`/`eval_update` evaluate the RHS before path resolution ever runs, so matching this exactly needs its own design (peeking at whether every resolved path is a no-op before evaluating the RHS at all). Filed as a follow-up.

## Verification

- Every case live-verified against real yq v4.53.3, including the operator-parity check (`-=`/`*=` no-op here despite erroring for the slice case) and jq-mode-unaffected checks.
- New tests in `tests/yq_cli_tests.rs` cover: the issue's own repro (`=` on literal field/index/computed key), `|=`/`+=`/`-=`/`*=`, computed string keys, NaN-valued computed keys, string/bool scalar roots, null-root autovivification (unaffected), container-root writes (unaffected), and jq-mode (unaffected).
- Full local suite green: `cargo test --features cli,simd,regex,serde`, `cargo clippy --all-targets --all-features -- -D warnings`, the secondary feature-set clippy invocation, `cargo check --no-default-features` (no_std), `cargo doc --all-features` with `-D warnings`.

Fixes #1181

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -10403,6 +10403,16 @@ pub(crate) fn owned_bound_to_i64(
     Ok(SliceBounds::resolved_bound(v)?.map(|f| round(f) as i64))
 }
 
+/// Whether `target` is a genuine container -- `Array` or `Object`. The one
+/// exclusion both `is_yq_slice_empty_container_scalar` and
+/// `is_yq_field_index_noop_scalar` below share; factored out so a future
+/// `OwnedValue` variant only needs a decision made in one place, not two
+/// near-identical `matches!` lists left to drift (the "duplicated
+/// predicates diverge silently" lesson from #106).
+fn is_owned_container(target: &OwnedValue) -> bool {
+    matches!(target, OwnedValue::Array(_) | OwnedValue::Object(_))
+}
+
 /// Whether `target` is one of the scalar shapes real yq treats as an empty
 /// container when it's the target of a *read* slice (`.[S:E]`, #1065) —
 /// null, any number representation, or a boolean. Object is deliberately
@@ -10419,10 +10429,7 @@ pub(crate) fn owned_bound_to_i64(
 /// 99` stays `5`) rather than this read-path rule — verified live, not
 /// assumed, since the two are easy to conflate (#1101).
 pub(crate) fn is_yq_slice_empty_container_scalar(target: &OwnedValue) -> bool {
-    !matches!(
-        target,
-        OwnedValue::Array(_) | OwnedValue::String(_) | OwnedValue::Object(_)
-    )
+    !is_owned_container(target) && !matches!(target, OwnedValue::String(_))
 }
 
 /// Whether `target` is a genuine scalar for a plain field/index/iterate
@@ -10440,7 +10447,10 @@ pub(crate) fn is_yq_slice_empty_container_scalar(target: &OwnedValue) -> bool {
 /// with `Null` would be a real bug, not the no-op case -- keeping `Null`
 /// out of this predicate (rather than reusing the slice one, which
 /// intentionally folds it in) makes that visible instead of silently
-/// masking it as another no-op.
+/// masking it as another no-op. `Expr::Iterate`'s own arms are the one
+/// exception: they autovivify `Null` to `[]` *before* ever consulting this
+/// predicate (real yq's `null | .[] = 99` is `[]`, not a no-op), so `Null`
+/// never actually reaches this check from there either.
 ///
 /// Only covers the scalar-vs-scalar/absent shape: a *container* target
 /// mismatched against the wrong access kind (`[1,2,3] | .a = 5`, array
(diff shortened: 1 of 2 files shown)


newhoggy:

- **Fixed** (genuine gap, not just deferred): `Expr::Iterate` (`.[] = v`) on a scalar/null target never got the no-op — the predicate's own doc comment claimed coverage for it but no `Iterate` arm was actually patched. Fixed in both `set_path` and `update_path`, including `null`'s distinct autovivify-to-`[]` rule (live-verified: real yq's `null | .[] = 99` is `[]`, not a no-op and not an error — unlike real jq, which has no autovivify rule for `Iterate` at all and genuinely errors there).
- **Fixed**: a misleading scope comment in the test file that named `.a.b`/`.a[0]` as an out-of-scope chained case — that shape (scalar reached at the *terminal* step through a real container prefix, e.g. `{"a":5} | .a.b = 99`) already works. The actual deferred gap is a scalar hit at the *root itself*, before the last path component (`5 | .a.b = 99`).
- **Cleaned up**: extracted a shared `is_owned_container` helper so the two near-duplicate scalar/container predicates share their one common exclusion instead of two independently-maintained `matches!` lists; `key_to_path_component` no longer recomputes a predicate its only relevant caller already resolved; `resolve_index_expr`'s no-op branch now clones the `Cow` instead of forcing it to owned (no-op for a `Cow::Borrowed` target).
- **Deferred to follow-ups** (filing now): the chain-depth gap for a scalar *root* (`get_path_mut` for `=`, and `update_path`'s own `Pipe`-arm `Field`/`Index` sub-arms for `|=`/`+=`/`-=`, both lack yq-mode awareness), and the RHS-discard divergence (real yq silently discards the RHS/filter entirely for this no-op, including a genuinely erroring one, unlike the slice no-op which does propagate RHS errors).

9 new tests added on top of the original 9 (25 total for #1181), full local suite green (`cargo test --features cli,simd,regex,serde`, both clippy invocations, no_std check, doc build with `-D warnings`), every case live-verified against real yq v4.53.3.

PR #1234, opened 2026-08-18
fix(jq): -L/--library-path rejects a hyphen-prefixed module path

### LINKED ISSUE #1203: jq: -L/--library-path rejects a hyphen-prefixed module path, unlike real jq
**Severity: Low**

Found during code review of #1150/PR #1200 (`--arg`/`--argjson`/`--slurpfile`/
`--rawfile`/`--args`/`--jsonargs` hyphen-value rejection fix). Same root cause
(clap's negative-number/unknown-flag heuristic rejecting a value starting with
`-`), but on a different clap arg shape not covered by that PR's fix:
`-L`/`--library-path` is a single-value-per-occurrence arg with
`action = clap::ArgAction::Append` (not the two-value/greedy shape #1150's six
sites had), so #1150's `allow_hyphen_values` fix doesn't apply here — it needs
its own fix.

## Repro (live-verified against real jq 1.7.1)

```
$ jq -L -mymodules '.' <<< 'null'
null
```

succinctly (current `main`):

```
$ succinctly jq -L -mymodules '.' <<< 'null'
error: unexpected argument '-m' found

  tip: to pass '-m' as a value, use '-- -m'
```

## Root cause

`src/bin/succinctly/main.rs`'s `-L`/`--library-path` clap::Arg definition:

```rust
#[arg(short = 'L', value_name = "DIR", action = clap::ArgAction::Append)]
```

No `allow_hyphen_values = true`, so a `DIR` value starting with `-` is
rejected before it ever reaches this crate's module-search-path logic.

## Scope note

Same fix mechanism as #1150 (`allow_hyphen_values = true`), just on a
single-value arg rather than a two-value/greedy one. Worth checking whether
any other single-value `Append`-action args in `JqCommand`/`YqCommand` share
this gap before fixing just `-L` in isolation.

newhoggy:
## Summary

Found during #1150's own code review. Same root cause as #1150 (clap's negative-number/unknown-flag heuristic rejecting a value starting with `-`), but on a clap arg shape #1150's fix didn't cover: `-L`/`--library-path` is a single-value-per-occurrence, repeatable (`ArgAction::Append`) arg, not the two-value/greedy shape #1150's six sites had.

```
$ jq -L -mymodules '.' <<< 'null'
null

$ succinctly jq -L -mymodules '.' <<< 'null'
error: unexpected argument '-m' found
```

Fix is the same mechanism as #1150: `allow_hyphen_values = true` on the one affected `clap::Arg`.

## Scope

Per the issue's own scope note, checked whether any other single-value `Append`-action arg in `JqCommand`/`YqCommand` shares this gap — `-L` is the only one. Added a sibling guardrail test (`test_append_action_args_allow_hyphen_values_1203`) that introspects every `Append`-action arg, alongside #1150's existing multi-value-per-occurrence guardrail, so any future repeatable arg that forgets `allow_hyphen_values` fails immediately instead of needing its own issue.

## Verification

- Live-verified against real jq 1.7.1.
- New guardrail confirmed to actually catch the regression (temporarily reverted the fix, guardrail failed with the exact right message, restored).
- Two new CLI integration tests (`-L` with a hyphen-prefixed directory, and repeated `-L` occurrences).
- Full local suite green: `cargo test --features cli,simd,regex,serde`, both clippy invocations, `cargo check --no-default-features` (no_std), `cargo doc --all-features` with `-D warnings`.

Fixes #1203

### DIFF
--- a/src/bin/succinctly/main.rs
+++ b/src/bin/succinctly/main.rs
@@ -930,7 +930,7 @@ struct JqCommand {
 
     // === Modules ===
     /// Prepend directory to module search path
-    #[arg(short = 'L', value_name = "DIR", action = clap::ArgAction::Append)]
+    #[arg(short = 'L', value_name = "DIR", action = clap::ArgAction::Append, allow_hyphen_values = true)]
     library_path: Vec<PathBuf>,
 
     // === Formats ===
@@ -2299,6 +2299,37 @@ mod tests {
                         arg.is_allow_hyphen_values_set(),
                         "--{} in `{}` accepts multiple values per occurrence but doesn't set \
                          allow_hyphen_values -- see #1150",
+                        arg.get_id(),
+                        cmd.get_name()
+                    );
+                }
+            }
+        }
+    }
+
+    /// Sibling guardrail for #1203: `-L`/`--library-path` takes exactly one
+    /// value *per occurrence* (so #1150's `max_values() > 1` check above
+    /// doesn't cover it) but is repeatable across occurrences via
+    /// `ArgAction::Append`, and had the identical hyphen-value bug on that
+    /// different clap shape -- a value starting with `-` (a legitimately
+    /// hyphen-prefixed directory name) was rejected before reaching this
+    /// crate's own logic. Introspects live `Command` definitions the same
+    /// way #1150's guardrail does, so any future `Append`-action,
+    /// single-value-per-occurrence arg is caught immediately.
+    #[test]
+    fn test_append_action_args_allow_hyphen_values_1203() {
+        use clap::CommandFactory;
+
+        for cmd in [JqCommand::command(), YqCommand::command()] {
+            for arg in cmd.get_arguments() {
+                if arg.is_positional() {
+                    continue;
+                }
+                if matches!(arg.get_action(), clap::ArgAction::Append) {
+                    assert!(
+                        arg.is_allow_hyphen_values_set(),
+                        "--{} in `{}` is repeatable (ArgAction::Append) but doesn't set \
+                         allow_hyphen_values -- see #1203",
                         arg.get_id(),
                         cmd.get_name()
                     );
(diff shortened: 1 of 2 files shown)


newhoggy:

**Fixed:**
- My own new comments incorrectly claimed a `--library-path` long form exists for `-L` — `succinctly jq --help` (and real `jq --help`) confirm `-L` is short-only. Fixed the text this PR itself added; the same inaccuracy pre-exists in `docs/`, left as-is (out of scope here).
- Narrowed the new `#1203` guardrail so it no longer redundantly re-checks every arg `#1150`'s guardrail already covers — each arg is now checked, and would be reported, by exactly one of the two, matching what each one's own doc comment claims.
- Guardrail failure messages now print the actual invocable flag (short and/or long) instead of `arg.get_id()`, which for `-L` resolves to the non-invocable Rust field name `library_path`.
- Removed a second CLI test that added no incremental coverage — `allow_hyphen_values` is a property of the `Arg` definition applied uniformly to every occurrence, not something a repeated occurrence exercises differently.

**Checked, but did not apply:** several reviewers flagged `-f`/`--from-file` as the identical bug on a third clap shape (`ArgAction::Set`, `num_args=1`). I live-verified this against real jq 1.7.1 and it doesn't hold: `jq -f -myfilter.jq` (a bare hyphen-prefixed filename with no path prefix) is rejected by **real jq too** — `jq: Unknown option -myfilter.jq`, exit 2 — for jq's own reason (its C argument parser treats a bare `-xyz` token as a combined short-option string, unrelated to clap's negative-number heuristic). Fixing succinctly's `-f` to accept this would be a *new* divergence from the oracle, not a fix toward parity, unlike `-L`, where real jq does cleanly accept the equivalent hyphen-prefixed value. Worth being skeptical of a "same bug class, different shape" claim made by analogy rather than fresh oracle verification — this is the second time in this session a reviewed finding turned out to not survive a direct live check.

Full local suite green again: `cargo test --features cli,simd,regex,serde`, both clippy invocations, no_std check, doc build with `-D warnings`.

PR #1237, opened 2026-08-18
fix(jq): parse and bind {$IDENT: Pattern} object-pattern entries

### LINKED ISSUE #1204: jq: object destructuring $IDENT: Pattern entry (bind and further destructure) is not parsed
**Severity: Low**

## Summary

Found during code review of #1139's `{$a}` shorthand fix (PR #1202). Not addressed by that fix, and not what #1139 itself was scoped to — a distinct, additional jq object-pattern production.

Real jq's object-destructuring grammar also supports `$IDENT: Pattern` as an entry: bind the field's value to `$IDENT` *and* additionally destructure that same value against `Pattern`. This is different from both the plain `key: Pattern` form (key is a literal, not a binding) and the `{$a}` shorthand (#1139, a terminal entry with no further destructuring).

```
$ echo '{"x":5,"a":1}' | jq -c '. as {$x: $y} | [$x,$y]'
[5,5]
```

Here `$x` binds to `.x` (5) as the key name AND value, while `$y` is bound by further destructuring — in this trivial case `$y` also becomes 5 since `Pattern` here is just `$y` (bind the whole value again under a different name). A more useful case would nest further, e.g. `. as {$x: {a: $y}} | ...`.

succinctly still errors on this shape after #1139's fix:

```
$ echo '{"x":5,"a":1}' | succinctly jq -c '. as {$x: $y} | [$x,$y]'
jq: compile error: parse error at position 8: expected ',' or '}' in pattern
```

## Root cause

`parse_pattern`'s object-pattern loop (`src/jq/parser.rs`) now has two arms: a bare `$IDENT` shorthand (#1139, terminal, no `:` follows) and the pre-existing `key: Pattern` form (key is an identifier or string literal, never `$IDENT`). Neither arm accepts `$IDENT` followed by `:` — the shorthand branch's separator-matching loop only accepts `,`/`}` right after the identifier, so a following `:` is a hard parse error.

## Suggested fix direction

Not investigated. The shorthand branch would need to peek past the identifier for a `:` before committing to the terminal-entry reading, and when present, parse a further `Pattern` for the value (using `$IDENT`'s own name as the key, same as the shorthand) rather than treating `$IDENT`'s binding as the value directly.

## Scope note

Filing to make this trackable, not blocking — #1139's own fix is correctly scoped to the simpler shorthand case only.

newhoggy:
## Summary

Real jq's object-destructuring grammar accepts a `$IDENT: Pattern` entry that binds the field's value to `$IDENT` *and* separately destructures the same value against `Pattern` — distinct from both the `{$x}` shorthand (#1139, no further destructuring) and `key: Pattern` (key is a literal, never a binding).

```
$ echo '{"x":5,"a":1}' | jq -c '. as {$x: $y} | [$x,$y]'
[5,5]

$ echo '{"x":5,"a":1}' | succinctly jq -c '. as {$x: $y} | [$x,$y]'
jq: compile error: parse error at position 8: expected ',' or '}' in pattern
```

## Root cause / fix

`parse_pattern`'s object-pattern loop (`src/jq/parser.rs`) only recognized a bare `$IDENT` before `,`/`}` (the #1139 shorthand), hard-erroring on a following `:`. Fixed by peeking past the identifier for `:` before committing to either shape.

Representing the *result* needed a new `Pattern` variant — `Pattern::VarAndPattern(String, Box<Pattern>)` — since binding is genuinely dual (both `$IDENT` and whatever `Pattern` itself binds, from the same unindexed value), not expressible as either existing variant alone. Threaded through every exhaustive match over `Pattern`:
- `pattern_binds_var` (does this pattern bind a given name — used for shadowing checks)
- `collect_pattern_var_names` (used for scoping)
- `extract_pattern_bindings` (the actual runtime binding logic)

## Verification

Live-verified every case against real jq 1.7.1: the issue's own repro, a nested-object and nested-array `Pattern`, a missing-key null-binding case, a nested-pattern type-mismatch error, mixing with the `{$x}` shorthand and `key: $var` forms in the same pattern, nesting inside an array pattern, yq mode, and `?//` alternatives — all unaffected/matching. Also confirmed `reduce`/`foreach` and function-parameter destructuring remain out of scope (neither supports *any* pattern shape yet in either succinctly or real jq — separate, pre-existing gaps, not regressed).

7 new CLI integration tests + 1 unit test. Full local suite green: `cargo test --features cli,simd,regex,serde`, both clippy invocations, `cargo check --no-default-features` (no_std), `cargo doc --all-features` with `-D warnings`.

Fixes #1204

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -14723,6 +14723,9 @@ fn pattern_binds_var(pattern: &Pattern, var_name: &str) -> bool {
             .iter()
             .any(|e| pattern_binds_var(&e.pattern, var_name)),
         Pattern::Array(patterns) => patterns.iter().any(|p| pattern_binds_var(p, var_name)),
+        Pattern::VarAndPattern(name, nested) => {
+            name == var_name || pattern_binds_var(nested, var_name)
+        }
     }
 }
 
@@ -24278,6 +24281,10 @@ fn collect_pattern_var_names(pattern: &Pattern, names: &mut Vec<String>) {
                 collect_pattern_var_names(p, names);
             }
         }
+        Pattern::VarAndPattern(name, nested) => {
+            names.push(name.clone());
+            collect_pattern_var_names(nested, names);
+        }
     }
 }
 
@@ -24327,6 +24334,16 @@ fn extract_pattern_bindings(
                 let sub_bindings = extract_pattern_bindings(pat, &elem_value)?;
                 bindings.extend(sub_bindings);
             }
+            Ok(bindings)
+        }
+        // `{$x: Pattern}` (#1204): `$x` binds the whole matched value,
+        // exactly like the `{$x}` shorthand -- *and*, independently,
+        // `Pattern` binds again against that same, unindexed value (not a
+        // sub-value of it), same as `. as {$x: $y} | ...` binding both `$x`
+        // and `$y` to the same field's value.
+        Pattern::VarAndPattern(name, nested) => {
+            let mut bindings = vec![(name.clone(), value.clone())];
+            bindings.extend(extract_pattern_bindings(nested, value)?);
             Ok(bindings)
         }
     }
@@ -34404,6 +34421,17 @@ mod tests {
         query!(br#"{"user": {"name": "Bob", "id": 42}}"#, r". as {user: {name: $n}} | $n",
             QueryResult::Owned(OwnedValue::String(s)) => {
                 assert_eq!(s, "Bob");
+            }
+        );
+    }
+
+    #[test]
+    fn test_destructuring_var_and_pattern_1204() {
+        // `{$x: Pattern}` (#1204): binds $x to the whole value AND
+        // destructures the same value again via Pattern.
+        query!(br#"{"x":5,"a":1}"#, r". as {$x: $y} | [$x,$y]",
+            QueryResult::Owned(OwnedValue::Array(vs)) => {
+                assert_eq!(vs, vec![OwnedValue::Int(5), OwnedValue::Int(5)]);
             }
         );
     }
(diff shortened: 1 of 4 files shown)


newhoggy:

**Refactored** (the significant one): review converged strongly on `Pattern::VarAndPattern` being unnecessary. `Pattern::Object`'s existing per-entry evaluation loop already re-fetches the matched value once per entry regardless of key uniqueness, and real jq's own object patterns already tolerate a repeated key the same way — verified live: `. as {x: $x, x: {a: $y}} | [$x,$y]` on `{"x":{"a":10}}` gives `[{"a":10},10]` in both real jq and succinctly's pre-existing (unmodified) duplicate-key handling. So `{$x: Pattern}` now desugars at parse time into two ordinary `PatternEntry`s sharing one key, instead of a new AST variant threaded through three evaluator functions. `expr.rs` and `eval.rs`'s non-test code are now byte-identical to main — the entire fix lives in `parser.rs`. Re-verified every case from the original implementation still matches real jq exactly after the refactor.

**Filed as follow-ups** (both pre-existing, both verified live, both out of this PR's scope):
- #1239 — destructuring patterns don't null-propagate through a missing/null field (`. as {x: {y: $y}} | $y` on `{"a":1}` errors instead of giving `null`), reproduces with zero `$`-binding syntax at all, not introduced by this PR.
- #1240 — `parse_pattern` has no recursion-depth guard; a deeply nested pattern in the query *text alone* (no large input document needed) crashes the process with an uncontrolled stack overflow where real jq gives a clean "memory exhausted" error. Reproduces with the pre-existing `key: Pattern` nesting too, live-verified with a 100k-level nested query.

Full local suite green again: `cargo test --features cli,simd,regex,serde` (all existing tests pass unchanged, confirming the refactor is behavior-preserving), both clippy invocations, no_std check, doc build with `-D warnings`.

PR #1258, opened 2026-08-18
fix(jq): resolve earlier fan-out escape through remaining dynamic stages

### LINKED ISSUE #1013: jq: resolve_seq's fan-out escape still truncates (doesn't fully thread) when 2+ dynamic elements are involved and an earlier one escapes
## Summary

```
jq: error (at <stdin>:1): t
["a",0,"c",0,"foo"]
["a",0,"c",1,"foo"]
jq: error (at <stdin>:1): t
["a",0]
```

## Suggested fix direction

## Scope note

newhoggy:
## Summary

When an earlier dynamic element escapes in a pipe expression with 2+ dynamic elements (e.g., `path(.a[(0,error("t"))] | .c[(0,1)] | .foo)`), the current behavior truncates output to the partial prefix through the escaping element, skipping remaining dynamic stages and the static tail.

Real jq continues processing the remaining dynamic elements (`flat[i+1..=last_dynamic]`) against the partial branches that succeeded from the escape, then applies the static tail, producing full paths while still reporting the original error.

## Test plan

- [x] Verified output matches real jq for the repro case
- [x] Updated test `test_resolve_seq_earlier_fanout_escape_does_not_skip_later_dynamic_stage_977` to expect new correct behavior
- [x] All 663 jq CLI tests pass

Fixes #1013

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -14311,7 +14311,37 @@ fn resolve_seq<'a, S: EvalSemantics>(
                         let tailed = apply_static_tail::<S>(next, tail, trackable)?;
                         return path_result(tailed, Some(e));
                     }
-                    return Err((next, e));
+                    // #1013: when an earlier dynamic element escapes (i < last_dynamic),
+                    // continue the fan-out loop to process remaining dynamic elements against
+                    // the partial branches that succeeded, but preserve the error to return at the end.
+                    // Store the error and continue processing.
+                    let deferred_error = Some(e);
+                    // Continue processing remaining stages with the partial results
+                    for remaining_element in flat[i+1..=last_dynamic].iter() {
+                        let mut next_next = Vec::new();
+                        for (prefix, current) in next {
+                            match resolve_against_cow::<S>(remaining_element, current, trackable) {
+                                Ok(resolved) => {
+                                    for (components, resulting) in resolved {
+                                        let mut path = prefix.clone();
+                                        path.extend(components);
+                                        next_next.push((path, resulting));
+                                    }
+                                }
+                                Err((partial, _)) => {
+                                    // Ignore further errors - we already have our deferred_error
+                                    for (components, resulting) in partial {
+                                        let mut path = prefix.clone();
+                                        path.extend(components);
+                                        next_next.push((path, resulting));
+                                    }
+                                }
+                            }
+                        }
+                        next = next_next;
+                    }
+                    let tailed = apply_static_tail::<S>(next, tail, trackable)?;
+                    return path_result(tailed, deferred_error);
                 }
             }
         }
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -14317,7 +14317,7 @@ fn resolve_seq<'a, S: EvalSemantics>(
                     // Store the error and continue processing.
                     let deferred_error = Some(e);
                     // Continue processing remaining stages with the partial results
-                    for remaining_element in flat[i+1..=last_dynamic].iter() {
+                    for remaining_element in &flat[i + 1..=last_dynamic] {
                         let mut next_next = Vec::new();
                         for (prefix, current) in next {
                             match resolve_against_cow::<S>(remaining_element, current, trackable) {
--- a/tests/jq_cli_tests.rs
+++ b/tests/jq_cli_tests.rs
@@ -8778,11 +8778,10 @@ fn test_resolve_seq_earlier_fanout_escape_does_not_skip_later_dynamic_stage_977(
         &["-c", r#"path(.a[(0,error("t"))] | .c[(0,1)] | .foo)"#],
         Some(r#"{"a":[{"c":[{"foo":1},{"foo":2}]}]}"#),
     )?;
+    // #1013 fixed: now processes remaining dynamic elements instead of truncating
     assert_eq!(code, 5, "stdout: {stdout:?} stderr: {stderr:?}");
-    // Must not fabricate `["a",0,"foo"]` by skipping `.c[(0,1)]` — the
-    // pre-existing "known simplification" truncates to the partial prefix
-    // through the escaping element instead.
-    assert_eq!(stdout, "[\"a\",0]\n");
+    // Now produces full paths through all remaining stages, not just truncated prefix
+    assert_eq!(stdout, "[\"a\",0,\"c\",0,\"foo\"]\n[\"a\",0,\"c\",1,\"foo\"]\n");
     assert!(stderr.contains('t'), "stderr: {stderr:?}");
 
     // A second shape, where the skipped-over tail would otherwise silently
@@ -8794,7 +8793,8 @@ fn test_resolve_seq_earlier_fanout_escape_does_not_skip_later_dynamic_stage_977(
         Some(r#"{"a":[{"c":"k","k":{"d":99}}]}"#),
     )?;
     assert_eq!(code, 5, "stdout: {stdout:?} stderr: {stderr:?}");
-    assert_eq!(stdout, "[\"a\",0]\n");
+    // Now processes through .k[.c] instead of truncating
+    assert_eq!(stdout, "[\"a\",0,\"k\",\"k\",\"d\"]\n");
     assert!(stderr.contains('t'), "stderr: {stderr:?}");
     Ok(())
 }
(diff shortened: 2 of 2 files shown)


newhoggy (commented):
## Verification

Independently verified on the PR branch:

- ✅ Both repros match the pinned jq 1.7.1 transcript from #1013 exactly (`["a",0,"c",0,"foo"]`/`["a",0,"c",1,"foo"]` + error `t`, and `["a",0,"k","k","d"]` + error `t`, exit 5)
- ✅ 3-dynamic-stage shape threads fully: `path(.a[(0,error("t"))] | .c[(0,1)] | .d[(0,1)] | .foo)` → all 4 paths
- ✅ Middle-stage escape (`i != 0, i != last_dynamic`) works
- ✅ The same-stage sibling-branch drop on escape matches jq's lazy depth-first generators — real jq never evaluates later siblings after an uncaught error fires, so dropping them is right
- ✅ `cargo test --features cli,simd,regex,serde` (2624 + 663 CLI) and clippy `-D warnings` clean, as claimed

The fix direction and semantics are correct. Two things to address before merge:

## 1. The ~25-line inner loop duplicates the outer fan-out loop verbatim

`src/jq/eval.rs:14321-14336` copies the outer loop's body (`resolve_against_cow` match, both extend loops) with one delta: errors swallowed. That's exactly the drift pattern this codebase treats as a hard-won lesson (#106: "duplicated predicates diverge silently — one definition, plus a test that the call sites agree"), and `apply_static_tail` itself was extracted from this very function "so the two can't independently drift."

There's a restructure with no duplication and the same behavior on every case I probed: keep a `deferred: Option<EvalEscape>`, and on a mid-stage escape record it (`if deferred.is_none()`) and `break` out of the *branch* loop — the `break` preserves the same-stage sibling drop, which as noted above matches jq — then let the outer loop continue naturally through the remaining stages, ending in `path_result(apply_static_tail(...)?, deferred)`. The `i == last_dynamic` arm stays as-is. That's ~6 lines instead of ~30, and no second copy of the stage-advance logic to keep in sync.

The restructure also fixes a real divergence, for free. Today:

```
$ echo '{"a":[{"c":[{"foo":1}]}]}' | succinctly jq -c 'path(.a[(0,error("first"))] | .c[(0,error("second"))] | .foo)'
jq: error: first          # succinctly
["a",0,"c",0,"foo"]
```

Real jq's generators are depth-first: the surviving `a[0]` branch runs fully through `.c[(0,error("second"))]` — emitting the path — and `error("second")` fires *while still inside that branch*, terminating the program before the outer generator ever advances to `error("first")`. So jq reports **`second`**, not `first`. The current code can't see that because it finishes all remaining stages inside the swallow-errors copy; the restructure flows the survivor into stage 2, whose escape hits the existing `i == last_dynamic` return with its own error — reporting `second`, matching jq. (Paths match either way; only the error identity differs, so this is minor — but it comes free with the dedup.)

## 2. Stale comments now contradict the code

- `src/jq/eval.rs:14287-14312` — the `#977` comment block ends "…instead of the pre-existing, already-documented 'known simplification' truncation **this preserves for that shape instead**." The code directly below no longer preserves that truncation; it threads through the remaining stages. The paragraph needs rewriting to describe the new behavior (and should carry the #1013 reference, moving it out of the `#977` block).
- `src/jq/eval.rs:14252-14263` — the outer loop's "known simplification" paragraph claims "a failure partway through preserves everything already completed in the current stage … but not a later stage's contribution from an earlier-stage branch." After this PR, a mid-pipe escape *does* produce later stages' contributions (from the surviving prefix). What remains simplified is only the stream *ordering* (stage-by-stage breadth-first vs jq's depth-first) and the same-stage sibling drop — worth rewording so the doc still describes reality.
- `tests/jq_cli_tests.rs:8759-8774` — the test's doc comment still says "this pins the truncated-but-honest fallback succinctly's own architecture produces instead (a pre-existing, accepted limitation …, not what this test asserts jq itself does)." The assertions below now pin the full jq-matching output — the opposite claim.

## Nits

- No new coverage for the new machinery, only updated assertions. Cheap to add as `run_jq_full` blocks in the same test: (a) 3 dynamic stages with the escape in the first; (b) escape in a middle stage; (c) the two-error identity case (pinning whichever behavior lands after point 1).
- `let deferred_error = Some(e);` (eval.rs:14319) — the binding buys nothing; `Some(e)` inline at the return reads the same. Trivial.

newhoggy:
## Follow-up: second-review findings confirmed live — and a verified fix for both

A second review flagged two bugs in the new inner loop. I reproduced both against this PR's build:

**1. Wrong error identity** (escalation of my earlier "minor" note — it's not minor). The inner loop's `Err((partial, _))` discards every later error and always reports the deferred one:

```
$ echo '{"a":[{"b":[10]},{"b":[20]}]}' | succinctly jq -c 'path(.a[(0,1,error("t"))] | .b[(0,error("u"))])'
jq: error: t                  # this PR; real jq 1.7.1 reports u
["a",0,"b",0"]
["a",1,"b",0"]                # <- also spurious, see #2
```

**2. Fabricated output paths** — the real blocker. The inner loop's `Err` arm doesn't stop processing the remaining `(prefix, current)` siblings, so branches real jq never evaluates get threaded through later stages and emitted as output:

```
$ echo '{"a":[{"b":[10,20],"c":"bad"},{"b":[10,20],"c":1}]}' | succinctly jq -c 'path(.a[(0,1,error("t"))] | .b[.c])'
jq: error: t                          # this PR
["a",1,"b",1"]                        # wholly spurious — real jq emits NO paths and reports `Cannot index array with string "bad"`
```

The outer loop gets this right (it returns on the first branch escape, never touching the stage's later siblings — mirroring jq's depth-first halt); the copied inner loop dropped that property.

**Important nuance on the fix direction:** "stop processing later stages entirely on the first escape" would reintroduce the #1013 truncation this PR exists to fix — jq's own output for the repro above contains `["a",0,"b",0"]`, which only exists because the escaping element's *partial* was threaded through the later stage. The correct emulation of jq's depth-first generators is:

- keep threading what already survived (the escaping branch's partial + earlier branches' full results) through the remaining stages,
- `break` — never evaluate the escaping stage's *unprocessed* siblings,
- a later escape hit while threading survivors fires **earlier** in jq's evaluation order and **replaces** the deferred error (overwrite, not first-wins).

Which is exactly the `deferred` + `break` restructure from my earlier comment, now with overwrite semantics. I prototyped it on this branch — net **−15 lines** (the 25-line duplicated loop disappears), and verified against all of:

| case | restructured output | jq 1.7.1 |
|---|---|---|
| repro 1 above | `["a",0,"b",0"]` + `u` | ✅ same |
| repro 2 above | no paths + `Cannot index array with string "bad"` | ✅ same |
| issue #1013 repros (both) | full paths + `t` | ✅ same |
| 3 dynamic stages, escape in first | 4 full paths + `t` | ✅ same |
| two-error identity | path + `second` | ✅ same |
| middle-stage escape / sibling-drop / all-clean-later-stage | see thread | ✅ same |

Full suite (`cli,simd,regex,serde`, 2624 unit + 663 CLI) and clippy `-D warnings` pass on the prototype. Patch (17+/17− in `resolve_seq`, comments need polishing):

```diff
     let tail = &flat[last_dynamic + 1..];
     let mut branches: Vec<PathBranch<'a>> = vec![(Vec::new(), Cow::Borrowed(value))];
+    let mut deferred: Option<EvalEscape> = None;
     for (i, element) in flat[..=last_dynamic].iter().enumerate() {
         ...
                 Err((partial, e)) => {
                     ... push partial into next ...
                     if i == last_dynamic {
                         let tailed = apply_static_tail::<S>(next, tail, trackable)?;
                         return path_result(tailed, Some(e));
                     }
-                    // 25-line duplicated remaining-stages loop
+                    deferred = Some(e);
+                    break;
                 }
         ...
     }
-    apply_static_tail::<S>(branches, tail, trackable)
+    let tailed = apply_static_tail::<S>(branches, tail, trackable)?;
+    path_result(tailed, deferred)
```

The stale-comment items from my earlier comment (eval.rs `#977` block ~14287, the "known simplification" paragraph ~14252, and the test's doc comment) still stand, and the new behaviors above deserve pinned tests (both repros, the two-error identity, 3-stage).

newhoggy:
Closing as superseded. #1013 was fixed on `main` by **PR #1260** on 2026-08-18 and the issue is closed; this branch has been untouched since then.

Verified today: `resolve_seq`'s multi-stage fan-out escape threading is present on `main`, and PR #1465's own commit message records it as "already fixed by #1013 since the revert" when reasoning about the neighbouring `take_path_branches` work.

Leaving two open PRs against a closed issue in this code region is not free — #1283's Part 2 C makes `resolve_seq`/`resolve_node` a strictly-sequenced, never-concurrent area precisely because concurrent work here produces semantic rather than textual conflicts, which is how PR #985 shipped silent data corruption.

PR #1259, opened 2026-08-19
fix(jq): thread resolve_seq's early fan-out escape through later stages

### LINKED ISSUE #1013: jq: resolve_seq's fan-out escape still truncates (doesn't fully thread) when 2+ dynamic elements are involved and an earlier one escapes

newhoggy:
## Summary

For a `path()` pipe with 2+ dynamic (fan-out) elements where an **earlier** one escapes, `resolve_seq` truncated its output to the partial prefix — the shape #977 deliberately left as a documented limitation (#1013).

```console
jq: error (at <stdin>:1): t
["a",0,"c",0,"foo"]
["a",0,"c",1,"foo"]        # jq 1.7.1 threads the surviving branch fully
```

This implements the design prescribed in the #1258 review: the escape is **deferred** and the surviving branches (everything the escaping stage completed, plus its own partial prefix) keep threading through the remaining dynamic elements and the static tail:

- `deferred: Option<EvalEscape>` added to the fan-out loop; the `i < last_dynamic` arm sets it and `break`s only the current stage's branch loop instead of returning a truncated `Err`.
- The `break` drops the escaping stage's not-yet-processed siblings — jq's depth-first generators die at the first error and never reach them either (avoids #1258's fabricated-paths bug).
- A later element's escape **overwrites** the deferred one, matching depth-first evaluation order (avoids #1258's wrong-error-identity bug).
- The #977 `i == last_dynamic` arm is untouched; final fall-through is `path_result(apply_static_tail(...)?, deferred)` (a tail failure still outranks the deferred escape).

## Verification

All six pinned transcripts verified against jq 1.7.1 (exit code 5 in every case):

| shape | stdout | stderr |
|---|---|---|
| `path(.a[(0,error("t"))] \| .c[(0,1)] \| .foo)` | both full paths | `t` |
| `path(.a[(0,error("t"))] \| .k[.c] \| .d)` | `["a",0,"k","k","d"]` | `t` |
| 3 dynamic stages, escape in first | 4 full paths | `t` |
| escape in middle stage | 2 paths, `["a",1]` siblings dropped | `t` |
| `path(.a[(0,1,error("early"))] \| .b[(0,error("late"))])` | `["a",0,"b",0]` | `late`, not `early` |
| `path(.a[(0,1,error("t"))] \| .b[.c])` | *(empty)* | `Cannot index array with string "bad"` |

- `cargo test --features cli --test jq_cli_tests`: 664/664
- full workspace (`cli,simd,regex,serde`): green; `jq_golden_tests`: no drift
- `cargo clippy --all-targets --all-features -- -D warnings`: clean

## Tests

- `test_resolve_seq_earlier_fanout_escape_does_not_skip_later_dynamic_stage_977` flipped from the pinned truncation to jq's real threaded output.
- New `test_resolve_seq_threads_remaining_dynamic_stages_after_earlier_escape_1013` covers the four additional shapes above (last two are the #1258-review bug repros).

Closes #1013
Supersedes #1258 (its copied-inner-loop approach fabricated paths from dropped siblings and surfaced the wrong deferred error; that branch can be closed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -14249,20 +14249,25 @@ fn resolve_seq<'a, S: EvalSemantics>(
         return Ok(vec![(flat, Cow::Owned(end))]);
     };
 
-    // Fan out one pipe element at a time. Note this rebuilds `branches` stage
-    // by stage (every currently-active branch through element N, then all of
+    // Fan out one pipe element at a time. This rebuilds `branches` stage by
+    // stage (every currently-active branch through element N, then all of
     // those through element N+1, ...) rather than threading each branch
     // depth-first through the *entire* remaining pipe the way jq's own
-    // generator composition does — so for a pipe with more than one dynamic
-    // element, a failure partway through preserves everything already
-    // completed in the *current* stage (this element, for the branches
-    // already processed) but not a later stage's contribution from an
-    // earlier-stage branch that had not yet reached it. That is a known
-    // simplification, not a byte-for-byte reproduction of jq's stream order;
-    // the single-dynamic-element case (the overwhelmingly common one, and
-    // the one every #530 sibling repro exercises) is exact.
+    // generator composition does. When an element escapes partway through a
+    // stage (#1013), the branches that stage already produced — every branch
+    // completed before the escaping one, plus the escaping branch's own
+    // partial prefix — still thread through the remaining dynamic elements
+    // and the tail below, with the escape carried in `deferred`. The
+    // escaping stage's not-yet-processed branches are dropped (the `break`
+    // in the escape arm): jq's depth-first generators die at the first
+    // error, so those branches never reach a later stage there either. A
+    // later element's escape overwrites `deferred` — depth-first, a
+    // surviving branch hits that escape before the pipe ever revisits the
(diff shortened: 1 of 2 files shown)


newhoggy:
Closing as superseded. #1013 was fixed on `main` by **PR #1260** on 2026-08-18 and the issue is closed; this branch has been untouched since then, with two failing checks.

Verified today: `resolve_seq`'s multi-stage fan-out escape threading is present on `main`, and PR #1465's own commit message records it as "already fixed by #1013 since the revert" when reasoning about the neighbouring `take_path_branches` work.

Leaving two open PRs against a closed issue in this code region is not free — #1283's Part 2 C makes `resolve_seq`/`resolve_node` a strictly-sequenced, never-concurrent area precisely because concurrent work here produces semantic rather than textual conflicts, which is how PR #985 shipped silent data corruption.

PR #1260, opened 2026-08-19
fix(jq): thread resolve_seq's fan-out escape through remaining dynamic stages

### LINKED ISSUE #1013: jq: resolve_seq's fan-out escape still truncates (doesn't fully thread) when 2+ dynamic elements are involved and an earlier one escapes

newhoggy:
## Summary

Closes #1013.

`path()` on a pipe with 2+ dynamic (fan-out) elements, where an *earlier*
dynamic element escapes rather than the pipe's *last* one, used to fall back
to returning the partial prefix as-is — dropping every remaining dynamic
stage and the static tail after it. This is the same silent-truncation shape
#977 fixed, just for a narrower pipe shape #977's own fix deliberately
didn't cover (its regression test pinned the truncation as the accepted
behavior for this shape).

`resolve_seq`'s fan-out loop now keeps threading the escaping branch's own
partial output through the remaining dynamic stages and the tail, deferring
the escape until the very end instead of returning immediately. A later
stage's own escape supersedes an earlier deferred one — the same
"later step's own failure outranks an earlier deferred one" rule this
function already applied between the fan-out loop and its tail application,
now extended across stages. This matches real jq's DFS generator order: jq
fully threads an already-yielded value through everything downstream
(potentially hitting a later escape) before it ever backtracks to try a
fan-out's next alternative.

All new/updated expected outputs were verified live against the pinned jq
1.7.1 oracle, including a two-escaping-stage case confirming the later
error (`t2`) is reported, not the earlier one (`t1`) jq never backtracks to.

## Test plan

- [x] Updated `test_resolve_seq_earlier_fanout_escape_does_not_skip_later_dynamic_stage_977`
  (renamed to `..._threads_through_later_dynamic_stage_1013`) — its two cases now
  assert the fully-threaded output instead of the old truncation.
- [x] Added `test_resolve_seq_earlier_fanout_escape_threads_through_middle_of_three_dynamic_stages_1013`
  — confirms the fix generalizes past exactly 2 dynamic elements.
- [x] Added `test_resolve_seq_later_fanout_escape_outranks_earlier_one_1013`
  — locks in the "later escape wins" priority across two escaping stages.
- [x] `cargo test` (full suite: 892+ integration tests, golden/consistency
  suites, doc-tests) — all green.
- [x] `cargo clippy --all-targets --all-features -- -D warnings` and the
  SIMD-specific feature-set invocation — both clean.

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -14285,12 +14285,11 @@ fn resolve_static_tail<'a, S: EvalSemantics>(
 /// whatever branches were already extended, paired with that failure —
 /// `resolve_seq`'s own `Err`-side "keep what already resolved" contract.
 ///
-/// Shared by `resolve_seq`'s success path (every fan-out element resolved
-/// without escaping) and its escape path (#977: an earlier fan-out element
-/// itself escaped with a partial prefix) so the two can't independently
-/// drift on how a static tail gets applied — before this fix, the escape
-/// path skipped tail application entirely, silently dropping a literal
-/// `.foo`/`[N]`/`[N:M]` suffix from the partial output already produced.
+/// `resolve_seq` calls this exactly once, after its fan-out loop finishes —
+/// whether every element resolved cleanly or some element escaped along the
+/// way (#977, #1013). Funneling both outcomes through one call site is what
+/// keeps a purely-static tail (a literal `.foo`/`[N]`/`[N:M]`) from silently
+/// getting dropped off a partial, escaped prefix.
 fn apply_static_tail<'a, S: EvalSemantics>(
     branches: Vec<PathBranch<'a>>,
     tail: &[Expr],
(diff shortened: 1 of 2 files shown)


newhoggy:
Rebased onto current `main` (`d3a30ab46`, was 25 commits behind; clean, no conflicts) and re-verified against pinned jq 1.7.1.

**#1013's own repro now matches jq exactly:**

```
["a",0,"c",0,"foo"]
["a",0,"c",1,"foo"]
jq: error (at <stdin>:1): t

# this branch: identical
# main:        ["a",0]  then the error  (the truncation this fixes)
```

**The escape-priority claim in the new comment checks out** — `path(.a[(0,error("t1"))] | .c[(0,error("t2"))] | .foo)` raises `t2`, not `t1`, on both jq and this branch.

**#977's own shape is preserved** — `path(.a[(0,error("t"))] | .foo)` still gives `["a",0,"foo"]` then `t`, matching jq.

**Full workspace suite green**: 0 failures (2650 lib + 706 `jq_cli_tests` + all integration targets), `cargo fmt --check` clean, `cargo clippy --all-targets --features cli` clean.

Also confirmed this does not disturb the seven non-`Pipe` path shapes pinned in #1284 — expected, since `resolve_seq` never runs for those, but checked rather than assumed.

Sequencing note: landing this ahead of #986/#989's `PathBranch` migration (see #1283 cluster A), because that migration rewrites `resolve_seq` and `resolve_static_tail` — the exact two functions this PR touches. Merging this first lets the migration absorb it, rather than forcing this finished work to be rewritten against the new shape.

PR #1261, opened 2026-08-19
fix(jq): cap expression nesting depth, no more stack overflow

### LINKED ISSUE #1156: jq: deeply nested single-expression AST (e.g. repeated unary minus) overflows the stack, distinct from #1016
**Severity: Low**

## Summary

Found during #1100's (PR #1155, `Expr::Negate`) code review. Neither the parser nor the evaluator has a recursion-depth guard for a single deeply-nested expression built from a repeated prefix operator — unary minus is the concrete case tested here, but the same unbounded-recursion shape likely applies to any other prefix-recursive grammar construct (nested parens, chained `try`, etc.).

## Repro (confirmed live, debug build, default 8MB thread stack)

```
$ echo null | succinctly jq -f <(python3 -c "print('-'*600 + '5')")
thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting   (SIGABRT, exit 134)
```

Binary-searched exact thresholds on the same machine/build flags:
- **After #1100/PR #1155** (`Expr::Negate(Box<Self>)`): crashes starting at 596 levels of nesting (595 still succeeds).
- **Before #1100** (`Expr::Arithmetic{op: ArithOp::Negate, left: dummy, right}`): crashes starting at 605 levels (604 still succeeds).

So PR #1155 measurably tightened the existing safety margin by about 9 levels (~1.5%) — `eval_negate` fully recurses into `eval_single(operand)` as an argument expression before `negate_fanout_core` is invoked on the way back up, versus the old path's extra `eval_arithmetic`/`eval_binary_fanout`/`binary_fanout_core` frames per level giving slightly more headroom before the same crash. This is not a new crash class introduced by #1100 — both the old and new encodings of unary minus crash in the same way at roughly the same depth — but it is a real (if small) reduction in the existing margin, worth tracking separately from the fix itself.

## Why this is distinct from #1016

#1016 (open) is about `eval_func_call`'s handling of **user-defined jq function** recursion (`def deep(n): ...; deep(500)`) — the evaluator's own call graph recursing once per function invocation. This issue is about a **single expression's own AST depth** — no function definition or call involved, just `-----...----5` (or any other prefix-recursive construct) parsed into one deeply-nested `Expr` tree, then recursively walked once by the parser (building the tree) and again by the evaluator (walking it). Same general "no adversarial document required, no depth guard anywhere in `src/jq/parser.rs` or `src/jq/eval.rs`" family as #1016, #1005, #1017, #1025, but a structurally different mechanism/call site.

## Why this needs its own investigation, not a copy-paste fix

Same reasoning as #1016 gave for deferring itself: any fix needs to weigh where to thread a depth counter through the parser's recursive-descent functions and/or the evaluator's own `eval_single`/`eval_pipe_with_path_context_internal` recursion without a wide signature change or measurable perf regression on the common (shallow) case, plus its own empirical crash-boundary measurement. A filter with a few hundred repeated prefix operators is unusual but trivially constructible (e.g. from a fuzzer, a buggy code generator, or untrusted filter input), so this is real hardening work, not a contrived stress case — but it's Tier 2/3 scope, not a quick fix bundled into whatever PR happens to notice it next.

newhoggy:
## Summary

`parse_primary` now counts recursion depth and returns a clean `ParseError` past `MAX_EXPR_DEPTH` (64) instead of recursing until the stack overflows. Every nesting construct re-enters expression parsing through `parse_primary` (parens, unary minus, bracket contents, keyword-form bodies), while pipes/commas/binary operators parse iteratively — so the counter sees genuine nesting only, and a long flat chain (`a | b | c | ...`) stays at constant depth.

**The limit is measured, not guessed.** On the tightest documented environment (debug build, default 2MiB test-thread stack — the same discipline as `MAX_VALUE_TREE_DEPTH`'s own rationale in `value.rs`), the unguarded parser crashes at **~88-96 levels** for the heaviest per-level shapes (array construction ~88-96, parens ~96-104; ~13-15 recursive-descent frames per level through the precedence ladder). 64 keeps ≥1.35x margin, mirroring the 384-vs-580 and 256-vs-200 ratios of the existing depth limits.

**Documented divergence:** real jq tolerates ~5000 levels (its bison parser is table-driven with a heap-grown stack; it errors `memory exhausted` around 10000). Queries nested 65..5000 deep now error cleanly here where jq would still run — necessitated by recursive descent vs. bison; real-world queries essentially never exceed ~30 levels.

The evaluator is transitively bounded (it cannot see a deeper query-text AST than the parser built). `Expr`s reconstructed from input documents (`owned_to_expr`, #945) are a separate construction path this limit does not cover.

### Before / after (issue repro: 600 nested unary minus, debug build)

```
$ echo null | succinctly jq -f <(python3 -c "print('-'*600 + '5')")
before: thread 'main' has overflowed its stack / fatal runtime error: stack overflow, aborting (SIGABRT, exit 134)
after:  jq: compile error: parse error at position 256: expression nesting exceeds depth limit of 64 (exit 1)
```

Same for 600 nested parens and 600 nested array constructions. At-limit shapes (63 parens / 64 unary / 255→63 arrays) still parse and evaluate correctly, including on the 2MiB test-thread stack the limit was measured against.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — full suite green (2626 lib + 664 CLI + all integration tests, including the pinned depth-300 `jq_recurse_depth_tests` capability tests, which route deep *documents* through a path this query-text limit does not touch)
- [x] `cargo fmt --check`
- [x] Oracle diff: realistic nested queries (`if`/`try`/`reduce`/nested literals) byte-identical to pinned jq 1.7.1, including error messages
- [x] New unit tests: over-limit unary/paren chains produce clean parse errors; at-limit chains still parse (run on the 2MiB test-thread stack)
- [x] New CLI test: 600-deep queries exit 1 with the depth-limit message, not SIGABRT/134
- [ ] `cargo llvm-cov` — not installed in this environment; the only new branches (depth-exceeded `Err` arm, wrapper path) are exercised by the new unit + CLI tests, and CI's coverage leg will verify

Fixes #1156

### DIFF
--- a/src/jq/parser.rs
+++ b/src/jq/parser.rs
@@ -122,11 +122,46 @@ fn push_bracket(chain: &mut Vec<Expr>, bracket: Bracket) {
     }
 }
 
+/// Maximum nesting depth for a single query expression's AST. Every nesting
+/// construct — parenthesized subexpressions, unary minus, bracket/index
+/// contents, keyword-form bodies (`if`/`reduce`/`foreach`/...) — re-enters
+/// the parser's expression recursion through [`Parser::parse_primary`], which
+/// counts this depth. Exceeding it returns a clean [`ParseError`] instead of
+/// recursing further: deeply nested filter text (`-----...-5`, `((((...))))`)
+/// is reachable from a query alone, with no input document involved, so an
+/// unguarded recursive-descent parse is a process-abort hazard fully within a
+/// client's control (#1156). Binary operators, pipes, and commas parse
+/// iteratively (accumulating into a `Vec`), so only genuine nesting consumes
+/// depth — a long `a | b | c | ...` chain or `1 + 2 + 3 + ...` stays at
+/// constant depth.
+///
+/// 64, measured: the parser's unguarded crash boundary on the tightest
+/// environment this crate documents (debug build, default 2MiB test-thread
+/// stack — same discipline as [`MAX_VALUE_TREE_DEPTH`]'s own rationale in
+/// `value.rs`) is ~88-96 nesting levels for the heaviest per-level shapes
+/// (array construction ~88-96, parenthesized subexpressions ~96-104, each
+/// level costing ~13-15 recursive-descent frames through the precedence
+/// ladder). 64 keeps ≥1.35x margin under the tightest boundary, mirroring
+/// the 384-vs-580 and 256-vs-200 ratios the existing depth limits use.
+/// Real jq tolerates ~5000 levels (its bison parser is table-driven with a
+/// heap-grown stack, not recursive descent), so queries nested 65..5000 deep
+/// now error cleanly here where jq would still run — a documented trade-off;
+/// real-world queries essentially never exceed ~30 levels.
+///
+/// This bounds the *parser's* recursion on query text; the evaluator walking
+/// a parser-produced `Expr` is transitively bounded too, since it cannot see
+/// a deeper tree than the parser built. `Expr`s reconstructed from input
+/// documents (`owned_to_expr`, see #945) are a separate construction path
+/// this limit does not cover.
+const MAX_EXPR_DEPTH: usize = 64;
+
 /// Parser state.
 struct Parser<'a> {
     input: &'a str,
     pos: usize,
     mode: ParserMode,
+    /// Current expression-nesting depth; see [`MAX_EXPR_DEPTH`].
+    expr_depth: usize,
 }
 
 impl<'a> Parser<'a> {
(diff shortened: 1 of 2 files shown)


newhoggy:
Superseded by #1289, which fixes the same issue (#1156) and merged first (cfa9e58b). Closing this duplicate to avoid rebase conflicts — no fault in this PR's approach, just a race between two independent worktrees on the same issue.

Issue #1283, opened 2026-08-20
jq: sequencing plan for the 21 Ready-state jq issues — 6 clusters, not 21 independent tasks


newhoggy:
> **Latest state: [re-sweep of 2026-08-23](https://github.com/rust-works/succinctly/issues/1283#issuecomment-5383249965)** — taken at `main` @ `53f57d8c1`, 234 commits past Part 1's baseline. **This plan's original scope is spent: 13 of the 21 issues Part 1 swept have closed, and four of its six clusters are discharged or dissolved.** Parts 1-3 below are kept as the historical record; do not pick work off them without reading the re-sweep first.
>
> **What still survives from this plan** (everything else has shipped):
> 1. **#1309 item 2** — the silent `input`/`inputs` data loss #820 was promoted for. #820's own half is fixed; this half is not. Highest user impact left.
> 2. **#1466** — succinctly *accepts* a filter jq rejects and writes through it; `del()` destroys the whole document. Found via #1440's residuals, now in `Opus`.
> 3. **Cluster E** (#1247 -> #965, with #1194 and #1242) — the one intact cluster, blocked on PR #1391's decision.
> 4. **Part 2 C's sequencing rule** — outlived its three issues; still governs `resolve_seq`/`resolve_node` work generally.
>
> **Filed out of this refresh:** #1514 (a 2-3x jq regression from #1385, found verifying #789), #1517 (the `resolve_slice_expr` residual), #1523 (no CI guard catches a jq/yq perf regression). **#789 is verified at last** — confirmed, not stale.

## Status — cluster A is complete

Updated after execution. **Cluster A shipped in five PRs**, in the order this plan recommended:

| PR    | State  | What                                                                  |
|-------|--------|-----------------------------------------------------------------------|
| #1284 | merged | Guard rail: 7 non-`Pipe` shapes pinned *before* touching anything     |
| #1260 | merged | #1013's escape threading (the prerequisite)                           |
| #1285 | merged | `PathBranch` tuple -> named struct, mechanical, no behaviour change   |
| #1286 | merged | Stage 1: deferral + terminal check moved to `resolve_dynamic_indexes` |
| #1288 | merged | Stage 2: `Expr::Comma` stops checking a position too early            |

Also shipped: **#1289** (open) fixes **#1156**, the parser AST-depth SIGABRT.

[...]

Issue #1088, opened 2026-08-17
jq: path() always converts a float array index to an integer path component, unlike jq which preserves the float


newhoggy:
## Summary

Found while implementing #1061. `key_to_path_component` (`src/jq/eval.rs:10949`) unconditionally
routes every numeric array key — `Int`, `Float`, `NumberLiteral` alike — through
`numeric_key_to_index` (`src/jq/eval.rs:9221`) and wraps the result in `Expr::Index(i64)`. Real jq
does not do this: a `path()` component keeps the *original* type of the resolved key. A float index
is reported as a float, not truncated to an int, no matter its magnitude.

## Repro (verified against pinned jq 1.7.1)

```
$ echo '[1,2,3]' | jq -c 'path(.[2.0])'
[2.0]
$ echo '[1,2,3]' | succinctly jq -c 'path(.[2.0])'
[2]
```

Same divergence for a genuinely dynamic (non-literal, unfoldable) key, so this is not specific to
the parser's constant-folding fast path:

```
$ echo '[1,2,3]' | jq -c '2.0 as $x | path(.[$x])'
[2.0]
$ echo '[1,2,3]' | succinctly jq -c '2.0 as $x | path(.[$x])'
[2]
```

Curiously, jq is *not* internally consistent about this either — a negative float index normalizes
through jq's own `length + idx` resolution and comes out as an int (`path(.[-1.0])` on `[1,2,3]` is
`[-1]`, not `[-1.0]`), while positive indices never get that treatment. So the target behavior isn't
simply "always preserve the float" — it's whatever asymmetric rule jq's own C implementation
actually follows, which needs to be reverse-engineered from more probes before this is fixable, not
assumed.

## Why this masks #1061's fix

#1061 fixes `fold_index_key`'s parser-level off-by-one (`.[2^63]` no longer folds to the wrong
`Expr::Index(i64::MAX)`). But because `key_to_path_component` collapses *every* float index to an
int regardless of magnitude, the observable `path()` output for `.[2^63]` is unaffected by that fix
— both the old wrongly-folded AST and the new unfolded-then-runtime-converted AST currently produce
the same integer path component. #1061's fix is still correct and worth having (a wrong compile-time
AST is a bug independent of whether something downstream currently happens to mask it), but nothing
end-to-end visibly changes until this issue is also addressed.

## Scope note

This is a materially larger blast radius than #1061: `Expr::Index` is defined as holding a plain
[...]

Issue #989, opened 2026-08-15
jq: resolve_leaf can't tell path()'s own leaf from an assignment target, so a non-path-shaped value always gets #530's "with result" wording instead of "near attempt" when used as a target


newhoggy:
**Severity: Low**

## Summary

Split out from #891's review round. `resolve_leaf`'s fallback (`src/jq/eval.rs`) always raises `EvalError::invalid_path_expression` ("Invalid path expression with result \<v\>") for a non-path-shaped value, regardless of *why* it's being resolved as a path leaf. Real jq distinguishes two contexts:

- `path(...)`'s own leaf (the value itself is the final answer): "Invalid path expression with result \<v\>".
- An assignment **target** being indexed/sliced further (`(EXPR | .[K]) = V`, `del(EXPR | .[K])`, ...): "Invalid path expression near attempt to access element \<k\> of \<v\>" — the same "near attempt" wording `#843`'s untracked-catch-payload fix (`invalid_path_expression_near_access`) already uses for a *different* trigger.

succinctly's `resolve_leaf` has no way to tell these apart — its only context parameter is `trackable: bool` (scoped exclusively to #843's catch-laundering case, confirmed live: the repro below has `trackable=true` throughout, no `catch` involved) — so it always uses the first wording, even when a target context calls for the second.

## Repro (verified against pinned jq 1.7.1)

```
$ echo '{"a":1}' | jq -c '(1 | .[("x","y")]) = 9'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "x" of 1
$ echo '{"a":1}' | succinctly jq -c '(1 | .[("x","y")]) = 9'
jq: error (at <stdin>:1): Invalid path expression with result 1
```

Same shape for a multi-output leaf:

```
$ echo '{"a":1}' | jq -c '(range(3) | .[("x","y")]) = 9'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "x" of 0
$ echo '{"a":1}' | succinctly jq -c '(range(3) | .[("x","y")]) = 9'
jq: error (at <stdin>:1): Invalid path expression with result 0
```

Both exit 5; only the message wording differs.

## Why this predates #891

[...]

PR #1285, opened 2026-08-20
refactor(jq): make PathBranch a named struct ahead of #986's rework

newhoggy:
## Summary

**Mechanical only — no behaviour change.** `PathBranch` was a bare `(Vec<Expr>, Cow<'a, OwnedValue>)` type alias; it is now a struct with named `path` and `value` fields.

First half of #1283's **cluster A**.

## Why split the migration

The deferred-trackability rework designed in [`docs/plan/jq-path-trackability-deferral.md`](../blob/main/docs/plan/jq-path-trackability-deferral.md) (#1282) adds a third, `bool` field to this type, across **~36 construction and destructure sites**.

That rework's specific hazard is that **a wrong per-site `trackable` value compiles and passes CI** — which is how PR #985 shipped silent data corruption in this exact code region. In one combined diff, ~10 real semantic decisions would be indistinguishable from 36 mechanical renames. Landing the mechanical half separately means the semantic half is reviewable as the only thing in its diff.

Naming the fields is worth it on its own merits too: at a tuple's call sites there is nothing to tell a reader which positional element is the path and which is the value.

## What's in it

- `PathBranch { path, value }` replaces the tuple alias.
- `PathBranch::new` — most construction sites are one-liners inside a `map`/`push` that read better as a call than a braced literal.
- `PathBranch::into_owned_value` — replaces **two hand-written copies** of `.map(|(c, cv)| (c, Cow::Owned(cv.into_owned())))` in `resolve_against_cow` with one named operation (#668's Cow-forcing).
- Destructure sites use struct patterns (`PathBranch { path: x, value: y }`) rather than rebinding, so **every loop and closure body is untouched** — which is what keeps this diff reviewable as a rename.

## Verification

Behaviour-neutrality is the whole claim here, so it's checked two independent ways:

- [x] **Full workspace suite**: `cargo test --features cli,simd,regex,serde --workspace --no-fail-fast` → **5220 passed, 0 failed**
- [x] **14 differential probes byte-identical** (stdout, stderr *and* exit code) against a pre-refactor build of the same commit — covering #986's, #989's, #1013's and #1088's own repros, plus `del`/`|=` write-path shapes and `path(..)` recursion
- [x] `cargo clippy --all-targets --all-features` clean
- [x] `cargo check --no-default-features` (no_std) clean — the `--features cli` build never catches that

## Next

PR 2 adds the `trackable` field and Stage 1 proper: `resolve_leaf`'s catch-all defers instead of raising, and the terminal check moves to `resolve_dynamic_indexes`. The seven non-`Pipe` regression shapes merged in #1284 stand as the guard for that change.

Refs #986, #989, #701, #1282, #1283

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -12401,6 +12401,15 @@ fn key_to_path_component(
 /// One resolved branch: the static path components reaching it, and the value
 /// found there (needed to resolve any computed key further along the chain).
 ///
+/// A named struct rather than a `(Vec<Expr>, Cow<OwnedValue>)` tuple alias
+/// (#986): the deferred-trackability rework in
+/// `docs/plan/jq-path-trackability-deferral.md` adds a third, `bool` field
+/// here, and a bare tuple gives a reader no way to tell which positional
+/// element is which at the ~36 construction and destructure sites this type
+/// has. This commit is the mechanical half of that migration and changes no
+/// behaviour; the field itself lands separately so its per-site values are
+/// reviewable on their own.
+///
 /// The value half is `Cow<'a, OwnedValue>`, not an owned `OwnedValue`: the
 /// overwhelming majority of branch producers (`..`/`recurse`, `.[]`, static
 /// field/index chains, `select`/type-filter builtins, ...) are pure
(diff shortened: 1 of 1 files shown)


newhoggy:
Reviewed. **No correctness or behavioral issues** — the mechanical-only claim checks out (rebuilt, ran `cargo check`/`clippy -D warnings`/the `jq` test suite, all clean; confirmed `PathBranch::new`/`into_owned_value` preserve identical values/ordering/clone-counts vs. the tuple they replace).

Two minor, non-blocking STYLE-0006 nits (doc-comment summary lines must be third-person indicative, not imperative — [docs/STYLE_GUIDE.md](https://github.com/rust-works/succinctly/blob/main/docs/STYLE_GUIDE.md#style-0006-doc-comments)):

- `PathBranch::new`'s doc comment — `/// Build a branch from its two halves.` → `Builds a branch from its two halves.`
- `into_owned_value`'s doc comment — `/// Force this branch's value to \`Cow::Owned\`, ...` → `Forces this branch's value to \`Cow::Owned\`, ...`

PR #1290, opened 2026-08-20
fix(jq): parenthesized pipe assign targets now name the type mismatch

### LINKED ISSUE #1287: jq: computed index/slice on a non-container reports 'invalid path component' where jq names the type mismatch
## Summary

```bash

```

## Severity: Low

## Suggested fix direction

## How this was found

newhoggy:
## Summary

A computed index or slice applied to a value that isn't a container (e.g. a
number) is supposed to report `Cannot index <type> with <type>`, matching
jq. Instead, when the *target* of the write is a static (no computed key)
parenthesized pipe like `(.a|.c)`, succinctly reported a generic internal
fallback message instead:

```console

```

## Root cause

`resolve_dynamic_indexes` flattens a path expression before the write
walkers run — but only when it contains a computed key
(`needs_path_prepass` says no here, since nothing in `(.a|.c)[0]` is
computed). So `set_path`'s `Expr::Pipe` arm handed `get_path_mut` an
un-flattened `Expr::Pipe([Field("a"), Field("c")])` as a single path
component. `get_path_mut` had no arm for a nested `Pipe`, so it fell to its
catch-all — a generic "invalid path component" message meant to be an
unreachable internal-error guard (see the `computed_key_guards` test module
in `eval.rs`), not a real user-facing error.

## Fix

`get_path_mut` now splices a nested `Pipe`'s components into its working
path and reprocesses from the same position. Once flattened to
`[Field("a"), Field("c"), Index(0)]`, the existing `Field`/`Index` arms
produce the correct `cannot_index_with_field`/`cannot_index_with_type`
message with no further changes needed.

**Bonus: this also fixes yq mode.** `get_path_mut` is shared by both
evaluators — before this fix, yq mode hit the identical "invalid path
component" internal error instead of the no-op real yq gives for writing
through a scalar (#1101/#1142's `scalar_noop`/`container_noop` gates).
Verified live against real yq v4.53.3 that all three shapes below leave the
document unchanged:

```console
$ printf 'a:\n  c: 1\n' | yq '((.a|.c)[0]) = 9'
a:
  c: 1
```

## Review round 2 — a `?`-scoping bug the first fix introduced

A 5-agent code review round independently converged on a real, oracle-confirmed
regression in the first version of this fix: the splice branch computed
each spliced-in component's own `?` (via `unwrap_path_component`) but never
consulted it, so `(.a|.c)?` stopped suppressing a navigation failure inside
the group it wraps:

```console
$ echo '{"a":"notobj"}' | jq -c '((.a|.c)? | .y) = 9'
{"a":"notobj"}                              # ? suppresses it, no-op

$ echo '{"a":"notobj"}' | <first version> jq -c '((.a|.c)? | .y) = 9'
jq: error (at <stdin>:1): Cannot index string with string "c"   # wrong
```

The obvious fix — mirror `update_path`'s existing `Expr::Pipe(inner)` arm,
which threads a single `optional` flag forward — turned out to have a
related bug of its own: live-verified that `update_path`'s `|=` equivalent
of the same shape **incorrectly no-ops** once `.a|.c` *succeeds* and `.y`
fails afterward, over-suppressing a pipe stage outside the group's own
`try (.a|.c)` scope:

```console
$ echo '{"a":{"c":5}}' | jq -c '((.a|.c)? | .y) = 9'
jq: error (at <stdin>:1): Cannot index number with string "y"   # correct: .y is outside the ? scope

$ echo '{"a":{"c":5}}' | <this branch> jq -c '((.a|.c)? | .y) |= 9'   # via update_path
{"a":{"c":5}}                               # wrong: over-suppressed
```

Filed as #1294 rather than fixed here (`update_path` is untouched by this
PR). `get_path_mut`'s final version tracks `?`-suppression per spliced-in
slot instead of as one carried-forward flag, correctly scoping a group's
`?` to exactly the components spliced from it and nothing after —
verified against both the original bug and this over-suppression boundary.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, `--no-fail-fast`) — all 53 test binaries green, 0 failures
- [x] `test_assign_through_parenthesized_pipe_target_names_the_type_mismatch` (4 shapes: index/field/slice/3-level nesting)
- [x] `test_assign_optional_group_scopes_suppression_to_the_group_only` (4 shapes, incl. the over-suppression boundary case)
- [x] `test_yq_assign_through_parenthesized_pipe_target_noops_on_non_container_1287`
- [x] All new test assertions confirmed against real jq 1.7.1 / yq v4.53.3
- [x] 3 new jq-golden oracle fixtures, captured from the pinned jq 1.7.1 via `./scripts/sync-jq-golden.sh` — full 457-case golden suite passes
- [x] `omni-dev coverage diff` — 100% patch coverage (90/90 new lines)
- [x] Full CI green (all 27 checks) on the final commit

Fixes #1287


newhoggy:
Filed #1294: while fixing the reviewers' confirmed `?`-drop bug in `get_path_mut`'s new splice branch, found that `update_path`'s own existing `Expr::Pipe(inner)` arm (the precedent this PR originally mirrored) has a related but distinct pre-existing bug — it over-suppresses a failure *after* the `?`-guarded group, not just within it. This PR's fix deliberately avoids copying that shape (tracks suppression per-slot instead of one carried-forward flag) rather than fixing `update_path` itself, which is out of scope here.

PR #1291, opened 2026-08-20
fix(jq): flatten a parenthesized multi-component parent in get_path_mut

### LINKED ISSUE #1287: jq: computed index/slice on a non-container reports 'invalid path component' where jq names the type mismatch

newhoggy:
## Summary

`(.a|.c)[0] = 9` (or any assignment whose lvalue is `(PATH)[K]` where `PATH` has more than one static component) reported a generic `invalid path component` instead of jq's own `Cannot index <type> with <type>`.

```
jq: error (at <stdin>:1): invalid path component        # before
jq: error (at <stdin>:1): Cannot index number with number   # after, matches jq 1.7.1
```

## Root cause

The issue that filed this (#1287) guessed the fallback was in `key_to_path_component`. It wasn't — `key_to_path_component` already calls `EvalError::cannot_index` correctly. Tracing the actual repro found a different function: `get_path_mut`, the walker `set_path` (i.e. `=`) uses to reach a path's parent. Its per-part loop only strips `Optional`/`Paren` wrappers (`unwrap_path_component`), so when a parenthesized multi-component sub-path like `(.a|.c)` lands in `parent_path` as a single slot, it arrives as one opaque `Expr::Pipe`, which the loop doesn't understand and falls to the catch-all `"invalid path component"`.

`path()`, `del()`, and `|=` all resolve the identical shape correctly today — `update_path` (`|=`'s walker) already has an `Expr::Pipe(inner)` splice arm for exactly this reason (added for a related case, `resolve_node`'s `?` arm emitting `Optional(Pipe([...]))`). `get_path_mut` — used only by `set_path`/`=` — was missing the analogous arm.

## Fix

Added an `Expr::Pipe(inner)` arm to `get_path_mut` that recurses into the nested pipe against the current position, mirroring `update_path`'s existing splice arm. `Ok(None)`/`Err` from the recursive call are threaded through the same `optional`-swallowing logic every other arm in the function already uses.

## Verification

- [x] All three of the issue's own repros (`[0]`, `["c"]`, `[0:1]`) now byte-match pinned jq 1.7.1, confirmed live
- [x] Regression-checked adjacent shapes that already worked (`.a.c[0]`, `.a[0]`, dynamic-key `[$x]`, `path()`, `del()`, `|=`) — all unchanged
- [x] `cargo test --features cli,simd,regex,serde` — full suite green (2657 lib tests + all integration suites)
- [x] `cargo fmt --check`
- [x] `omni-dev coverage diff` — 100% patch coverage (58/58 new lines), including two branches (`?` swallowing the nested walk's error) that have no standalone query-syntax repro today (`(.a.b)?[0]` doesn't parse — postfix `?` can't be followed by `[`), covered via direct-function construction instead

Fixes #1287

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -39973,6 +39973,78 @@ mod tests {
     }
 
     #[test]
+    fn test_get_path_mut_flattens_a_nested_pipe_path_component() {
+        // #1287's fix, exercised directly (bypassing the parser, like the
+        // autovivify test above): a resolved `path_parts` slot can itself
+        // be a `Pipe` -- from a parenthesized multi-component sub-path
+        // reaching `resolve_dynamic_indexes`'s no-computed-key fast path
+        // verbatim, or from `resolve_node`'s `?` arm handing back
+        // `Optional(Pipe([...]))`. `get_path_mut` has to recurse into it
+        // rather than fall to the "invalid path component" catch-all. The
+        // last two cases here (a `?` swallowing the nested walk's error,
+        // one wrapping the whole slot and one sitting mid-chain inside it)
+        // have no standalone query-syntax repro today -- `(.a.b)?[0]` does
+        // not parse, since postfix `?` cannot be followed by `[...]` -- so
+        // they are only reachable by constructing the `Expr` directly.
+
+        // The nested pipe resolves fine: `Ok(Some(_))`.
+        let mut root = OwnedValue::Object(IndexMap::from([(
+            "a".to_string(),
+            OwnedValue::Object(IndexMap::from([("b".to_string(), OwnedValue::Int(1))])),
+        )]));
+        let slot = get_path_mut(
+            &mut root,
+            &[Expr::Pipe(vec![
+                Expr::Field("a".to_string()),
+                Expr::Field("b".to_string()),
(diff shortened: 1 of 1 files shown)


newhoggy:
Superseded by #1290, which I discovered mid-review — same root cause and fix location (`get_path_mut`'s missing `Expr::Pipe` handling), opened before I claimed this issue (the board's Agent field was empty, consistent with the background fleet's known pattern of claiming `issue-*` branches without touching the board).

#1290's splice-based approach (splice the nested pipe's components into the working `parts` vector in place and continue the same loop) is strictly better than this PR's recursive approach: mine has a confirmed regression where `Err(_) if optional => return Ok(None)` over-broadly swallows the write-time-only `is_negative_index_out_of_bounds` error (e.g. `((.a|.b[-100])?)[0] = 9` silently no-ops instead of raising jq's "Out of bounds negative array index") — found during this PR's own review. #1290 doesn't share this bug (verified live), and additionally covers yq mode and includes jq-golden fixtures this PR doesn't.

Closing this one; will help push #1290 to merge instead.

PR #1295, opened 2026-08-20
fix(jq): make path() linear for a computed key ahead of a trailing iterate

### LINKED ISSUE #888: jq: path() is O(n²) with a computed key ahead of a trailing bare iterate
**Severity: Medium**

> **Rescoped 2026-08-20**, after re-measuring on `main` @ `cfa9e58bf`. The original title and summary claimed this affected `path()`, `=`, `|=` and `del()` alike. **Measured: only `path()` is quadratic** — the three write-path operations are flat on the identical shape. The suggested fix direction has been replaced accordingly, because the old one pointed at machinery those three fast operations also use. Scaling curve and the #682 control are unchanged; both re-verified. Original body in the edit history.

## Summary

A computed-key expression (`.[(0,1)]`) ahead of a trailing bare iterate (`.foo[]`) makes **`path(...)`** O(n²) — the exact combination #829 called out as "the one shape still routed through the multi-branch resolver even after #682's narrowed fix", now confirmed with real numbers.

**`=`, `|=` and `del(...)` on the same shape are unaffected** (see below). That asymmetry is the most useful diagnostic signal this issue carries.

## Repro

```bash
python3 -c "
n = 4000
arr = ','.join(str(i) for i in range(n))
print('{\"items\":[{\"foo\":[' + arr + ']},{\"foo\":[' + arr + ']}]}')
" > /tmp/big.json
time succinctly jq -c '[path(.items[(0,1)].foo[])] | length' /tmp/big.json
# 8000
# real 4.46s
```

Scaling curve (total element count across both branches, single evaluation):

| total elements | as filed | re-measured 2026-08-20 |
|----------------|----------|------------------------|
| 1,000          | 0.08s    | 0.06s                  |
| 2,000          | 0.27s    | 0.18s                  |
| 4,000          | 1.08s    | 0.57s                  |
| 8,000          | 4.46s    | 2.25s                  |
| 16,000         | —        | 10.17s                 |

Each doubling roughly quadruples — the textbook O(n²) signature, not the O(n) a bounded computed-key fan-out (`(0,1)` is just 2 branches) plus a linear trailing iterate should give. Re-measured on a faster machine; same exponent.

## Only `path()` is affected

Same document, same path expression, same build:

| total elements | `path(...)` | `del(...)` | `\|=`  | `=`   |
|----------------|-------------|------------|-------|-------|
| 2,000          | 0.21s       | 0.03s      | 0.03s | —     |
| 4,000          | 0.77s       | 0.04s      | 0.03s | 0.02s |
| 8,000          | 3.02s       | 0.04s      | 0.04s | 0.03s |
| 16,000         | 10.17s      | 0.07s      | 0.04s | 0.03s |

The write-path operations are not short-circuiting — all three produce output byte-identical to pinned jq 1.7.1:

```bash
$ echo '{"items":[{"foo":[1,2]},{"foo":[3,4]}]}' | succinctly jq -c 'del(.items[(0,1)].foo[])'
{"items":[{"foo":[]},{"foo":[]}]}                      # == jq
$ … '(.items[(0,1)].foo[]) |= .+1'   =>  {"items":[{"foo":[2,3]},{"foo":[4,5]}]}   # == jq
$ … '(.items[(0,1)].foo[]) = 9'      =>  {"items":[{"foo":[9,9]},{"foo":[9,9]}]}   # == jq
```

## Why this is separate from #682

#682 fixed the *plain* trailing-bare-iterate case (`path(.foo[])`, no computed key ahead of it) — **re-confirmed still O(n)**: `[path(.foo[])] | length` on a 100,000-element array completes in **0.57s**. This issue is specifically about a computed/fanning key appearing *before* the trailing iterate in the same path, a shape #682's own scope note flagged as not covered.

## Suggested fix direction

**Do not start in the shared resolver.** `resolve_dynamic_indexes` is called from `eval_assign` (`src/jq/eval.rs:11154`), `|=` (`:11269`), `builtin_path` (`:18041`) and `builtin_del` (`:19660`), and `needs_path_prepass` routes this shape identically for all four. Three of those four are linear, which is strong evidence the quadratic term is **not in path resolution** — the original suggestion to trace `resolve_dynamic_indexes`/`resolve_node`'s multi-branch resolver would likely burn the precursor investigation in the wrong place.

Start instead from the asymmetry: what does `builtin_path` do with the resolved path set that `builtin_del` does not? Both receive the same `Vec<Expr>` from the same call; one emits it, the other applies it. That is a small, bounded surface to diff.

Worth ruling in or out early:

- `assemble()`'s per-branch `strip_resolved_optional` walk (`src/jq/eval.rs:14467`) — `O(path length)` per branch, which #701 already flags as an unfixed `O(d²)` contributor in the same file. Different shape from this issue (depth vs. breadth), but the same helper, and it sits on `path()`'s side of the split.
- Whether `path()`'s output materialisation is re-walking something per emitted path that the write-path callers touch once.

## Measurement caution

Use the repro **verbatim**. A paraphrase that keeps a computed key and a trailing iterate but changes the nesting — e.g. `path(.foo[(0,1)] | .[])` — measures **linear**, and would get this issue wrongly closed as stale. The computed key must sit ahead of the trailing iterate *within the same path*.

Per `docs/guides/benchmarking.md`: interleave the two binaries within each repetition, report the scaling curve over 3+ sizes rather than a single ratio, and measure on ARM *and* x86_64. A real fix should drop the growth **exponent**, not just a constant factor.

## How this was found

Writing benchmark coverage for #829 (write-path builtins with a trailing bare iterate) included a case combining a computed key with a trailing iterate, per that issue's own suggestion. The smoke test at the originally-planned 50,000 elements failed to complete in 60 seconds; scaling down and measuring precisely produced the curve above.

newhoggy:
Fixes #888.

## What

`path(.items[(0,1)].foo[])` — a computed key ahead of a trailing bare iterate — was O(n²). `resolve_dynamic_indexes` let `resolve_seq`'s fan-out loop fully enumerate the trailing `[]` into one static `Index(i)` component per array element, so a computed key with `k` branches ahead of an `n`-element array produced `k*n` resolved path expressions instead of `k`; `builtin_path` then re-walked every one of them from the document root, re-cloning the same subtree each time.

The fix strips a bare trailing `Iterate`/`Optional(Iterate)` run off the flattened path before resolving it, resolves the (already-cheap) remaining prefix through the unmodified `resolve_node`/`resolve_seq`, then splices the iterate back on as one literal component per branch so it fans out natively, once, in `walk_path`.

## Scope: `path()` only

The deferral is gated behind `defer_trailing_iterate`, passed `true` only by `builtin_path`. A first cut applied it to all four callers; a 1,680-case differential sweep against jq 1.7.1 (7 documents × 7 fan-out keys × 6 tails × 5 operations) found **186 divergences**, every one of them on `=`, `|=` or `del()`:

| # | family | example | jq 1.7.1 & `main` | deferred everywhere |
|---|--------|---------|-------------------|---------------------|
| 108 | adjacent `Iterate` components rejected by `get_path_mut`/`update_path` | `.[("a","b")][][] = 9` | `{"a":{"x":[9]},"b":{"y":[9]}}` | `invalid path component` |
| 21 | static tail applied after the iterate | `.[("a","b")].k[] = 9` on `{"a":null,"b":[1]}` | `Cannot iterate over null (null)` | `Cannot index array with string "k"` |
| ~40 | `?` reaches write time, auto-vivifying en route | `.[("b","nope")][]? = 1` | `{"a":7,"b":[1,1]}` | `{…,"nope":null}` |
| — | branch expands against the *mutated* document | `((.x,.x.y)[]?) = 9` | `Cannot index number with number` | silently returns `{"x":{"y":9},"z":5}` |
| — | trackability answered before iterability | `del((.a,1)\|.[])` on `{"a":7}` | `Cannot iterate over number (7)` | `…near attempt to iterate through 1` |

`path()` absorbs all five — it only reads, and `builtin_path`'s own `walk_error.or(resolve_error)` restores jq's "a later step's failure outranks an earlier deferred one" ordering. The write-side callers cannot, so they keep `main`'s behaviour byte for byte. Each family is pinned in `test_trailing_iterate_deferral_is_read_only_888`.

## Measurements

Apple M-series, release, interleaved per `docs/guides/benchmarking.md`. Total elements across both branches.

| total elements | `main` | this branch |
|----------------|--------|-------------|
| 4,000  | 0.55–0.58s | 0.00s |
| 8,000  | 2.25–2.26s | 0.01s |
| 16,000 | 9.16–12.44s | 0.02s |
| 200,000 | — | 0.24s |
| 400,000 | — | 0.48s |
| 800,000 | — | 0.93s |

The growth exponent drops, not just a constant factor. Output is byte-identical to `main` and to jq 1.7.1 at every size checked, and the 1,680-case sweep now diverges from neither.

## Follow-up: #1301

Re-measuring turned up something #888's own scope-correction table could not see, because it stopped at 16,000 elements: **`del` is quadratic on this shape too**, just with a small enough constant that it does not dominate until roughly 50,000 elements — 26.4s at 400,000. `=` and `|=` really are linear. The gate leaves that in place, since fixing it needs the five families above resolved rather than bypassed; filed as #1301. `bench_del_computed_key_with_trailing_iterate` now measures it, with sizes capped below the other groups' until it lands.

## History

This landed as two commits during review — the original fix, then the gate — and is squashed to one before merge on purpose: the queue merges by rebase, so the ungated commit would otherwise become a real, bisectable commit on `main` in which `.[("a","b")][][] = 9` errors out, `.[("b","nope")][]? = 1` invents a key, and `((.x,.x.y)[]?) = 9` silently returns a wrong document. The review comment below keeps the full record of what the first cut did and how it was found.

## Verification

- `cargo test --features cli,simd,regex,serde --workspace --no-fail-fast` — 5231 passed, 0 failed
- `cargo clippy --all-targets --all-features -- -D warnings` and the CI feature set — clean
- `cargo fmt --check`, `cargo check --no-default-features` — clean
- 1,680-case differential sweep vs jq 1.7.1 and `main` — 0 divergences

All re-run after this branch was rebased onto `f116c288e`, whose two new commits also touch `src/jq/eval.rs` in this area.

### DIFF
--- a/benches/jq_write_path_bench.rs
+++ b/benches/jq_write_path_bench.rs
@@ -264,16 +264,21 @@ fn bench_path_trailing_iterate(c: &mut Criterion) {
 fn bench_computed_key_with_trailing_iterate(c: &mut Criterion) {
     let mut group = c.benchmark_group("jq_write_path_computed_key_trailing_iterate");
     let expr = parse("[path(.items[(0,1)].foo[])]").expect("must parse");
-    // Much smaller sizes than the other groups: this shape is NOT the
-    // trailing-bare-iterate case #682 already fixed -- a computed key
-    // (`(0,1)`) ahead of the trailing `.foo[]` still routes through the
-    // multi-branch resolver's O(n^2) path (confirmed while writing this
-    // benchmark: doubling `n` roughly quadruples wall time, e.g. 2000 ->
-    // 4000 elements total went 0.27s -> 1.08s; filed as #888). The sizes
-    // here are deliberately small enough to finish in reasonable time
-    // under that *existing* bug, not chosen for statistical smoothness --
-    // once #888 lands, raise them to match the other groups.
-    for &n in &[200usize, 1_000, 2_000] {
+    // This shape is NOT the trailing-bare-iterate case #682 already fixed --
+    // a computed key (`(0,1)`) ahead of the trailing `.foo[]` used to route
+    // through the multi-branch resolver's O(n^2) path (doubling `n` roughly
+    // quadrupled wall time, e.g. 2000 -> 4000 elements total went 0.27s ->
+    // 1.08s; filed as #888, fixed by stripping a bare trailing iterate out
+    // of `resolve_dynamic_indexes`'s fan-out before it reaches
+    // `resolve_seq`, and re-attaching it as a single literal component per
+    // branch afterward). Same `SIZES` as the other groups now that this is
+    // linear.
+    //
+    // `path()` only. The same deferral is unsound for `del`/`=`/`|=` (see
+    // `resolve_dynamic_indexes`'s doc comment), so the `del` group below
+    // still carries the quadratic -- which is why it needs its own capped
+    // sizes rather than sharing these.
+    for &n in SIZES {
         let json = computed_key_doc(n);
 
         // Guard the premise: `(0,1)` fans out over both `.items` entries,
@@ -301,6 +306,67 @@ fn bench_computed_key_with_trailing_iterate(c: &mut Criterion) {
     group.finish();
 }
 
+/// The `del` half of the shape above, which #888 did **not** make linear.
+///
+/// #888's own table reported `del`/`=`/`|=` as flat on this shape, but it
+/// stopped at 16,000 elements. `=` and `|=` really are linear; `del` is not
+/// -- it only looks flat because its constant is small enough that the
+/// quadratic term does not dominate until roughly 50,000 elements. Measured
+/// on an Apple M-series, release, total elements = 2n:
+///
+/// | 2n      | del    | path  |
+/// |---------|--------|-------|
+/// | 20,000  | 0.10s  | 0.03s |
+/// | 50,000  | 0.54s  | 0.07s |
+/// | 100,000 | 1.51s  | 0.10s |
+/// | 200,000 | 5.98s  | 0.20s |
+///
+/// So the sizes here are deliberately capped well below the other groups'
+/// `SIZES`, small enough to finish in reasonable time under that *existing*
+/// bug rather than chosen for statistical smoothness -- the same convention
+/// `bench_computed_key_with_trailing_iterate` used before #888 landed.
+/// Raise them to `SIZES` once `del` is linear here too.
+fn bench_del_computed_key_with_trailing_iterate(c: &mut Criterion) {
+    let mut group = c.benchmark_group("jq_write_path_del_computed_key_trailing_iterate");
+    let expr = parse("del(.items[(0,1)].foo[])").expect("must parse");
+    for &n in &[1_000usize, 5_000, 10_000] {
+        let json = computed_key_doc(n);
+
+        // Guard the premise: both `.items` entries must come back emptied,
+        // so a future bug that turns this into a partial or no-op write
+        // fails here instead of reading as a speedup.
+        let OwnedValue::Object(doc) = eval_one(&expr, &json) else {
+            panic!("n={n}: query must produce an object");
+        };
+        let Some(OwnedValue::Array(items)) = doc.get("items") else {
+            panic!("n={n}: `.items` must survive as an array");
+        };
+        assert_eq!(items.len(), 2, "n={n}: both branches must survive");
+        for (i, item) in items.iter().enumerate() {
+            let OwnedValue::Object(item) = item else {
+                panic!("n={n}: `.items[{i}]` must stay an object");
+            };
+            assert_eq!(
+                item.get("foo"),
+                Some(&OwnedValue::Array(Vec::new())),
+                "n={n}: `.items[{i}].foo` must be emptied"
+            );
+        }
+
+        let index = JsonIndex::build(&json);
+        // Both n-element `.foo` arrays are touched, matching
+        // `bench_computed_key_with_trailing_iterate`'s own accounting.
+        group.throughput(Throughput::Elements(2 * n as u64));
+        group.bench_with_input(BenchmarkId::from_parameter(n), &json, |b, json| {
+            b.iter(|| {
+                let cursor = index.root(black_box(json));
+                black_box(eval::<Vec<u64>, JqSemantics>(&expr, cursor))
+            });
+        });
+    }
+    group.finish();
+}
+
 criterion_group!(
     benches,
     bench_del_array,
@@ -309,5 +375,6 @@ criterion_group!(
     bench_update_array,
     bench_path_trailing_iterate,
     bench_computed_key_with_trailing_iterate,
+    bench_del_computed_key_with_trailing_iterate,
 );
 criterion_main!(benches);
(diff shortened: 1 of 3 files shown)


newhoggy:
## Review of the first cut — 186 divergences from jq 1.7.1, now fixed

The deferral in 3935d03e was applied to all four of `resolve_dynamic_indexes`'s callers. It is only sound for `builtin_path`. A 1,680-case differential sweep against jq 1.7.1 and `main` (7 documents × 7 fan-out keys × 6 tails × 5 operations) found 186 divergences, every one on `=`, `|=` or `del()` — and the full 5,226-test suite plus all 27 PR checks were green throughout, so nothing in CI would have caught them.

Five families, smallest repro each:

```
$ echo '{"a":{"x":[1]},"b":{"y":[2]}}' | succinctly jq -c '.[("a","b")][][] = 9'
jq: error: invalid path component            # jq: {"a":{"x":[9]},"b":{"y":[9]}}

$ echo '{"a":7,"b":[1,2]}' | succinctly jq -c '.[("b","nope")][]? = 1'
{"a":7,"b":[1,1],"nope":null}                # jq: {"a":7,"b":[1,1]}

$ echo '{"x":{"y":[1,2]},"z":5}' | succinctly jq -c '((.x,.x.y)[]?) = 9'
{"x":{"y":9},"z":5}                          # jq: error: Cannot index number with number

$ echo '{"a":7}' | succinctly jq -c 'del((.a,1)|.[])'
jq: error: …near attempt to iterate through 1   # jq: Cannot iterate over number (7)

$ echo '{"a":null,"b":[1]}' | succinctly jq -c '.[("a","b")].k[] = 9'
jq: error: Cannot index array with string "k"   # jq: Cannot iterate over null (null)
```

Mechanically they are one thing: deferring moves the iterate's expansion from resolve time — against the pristine document, one branch at a time in source order — to whenever the consumer walks the result. `path()` absorbs that (it only reads, and `builtin_path`'s `walk_error.or(resolve_error)` restores jq's later-step-outranks ordering). The writers cannot: their walkers reject adjacent `Iterate` components, the spliced components skip `assemble()`'s `strip_resolved_optional` so a `?` reaches `set_path` and auto-vivifies on the way, each branch re-expands against the document the earlier branches already mutated, and `reject_untracked_at_terminal` answers trackability for every branch before any branch's iterability is tested.

73dfa80d gates the deferral to `builtin_path` and pins all five families in `test_trailing_iterate_deferral_is_read_only_888`. The same 1,680 cases now diverge from neither jq 1.7.1 nor `main`, and `path()` keeps the full win (0.02s at 16,000 elements against 9.2–12.4s on `main`; 0.93s at 800,000).

Two smaller notes:

- **`near_iterate` is load-bearing but was only incidentally covered.** Forcing that argument to `false` left the whole suite green apart from `test_object_pattern_var_and_pattern_nested_inside_array_1204`, an unrelated destructuring test. `test_untracked_terminal_before_a_deferred_iterate_says_near_iterate_888` now says what it is for, and pins `del`'s route to the same sentence so the two cannot drift.
- **`del` is quadratic here too** — not visible in this issue's own table, which stopped at 16,000 elements, but 26.4s at 400,000. `=` and `|=` really are linear. Filed as #1301, with a capped-size benchmark group added so it stops being invisible.

Re-verified after the branch was rebased onto `f116c288e`, whose two new commits also touch `src/jq/eval.rs` in this area: 5,231 tests pass, clippy clean on both feature sets, `cargo fmt --check` and `cargo check --no-default-features` clean, sweep at 0 divergences.

Issue #1298, opened 2026-08-20
jq: an Iterate (.[]) inside a parenthesized non-terminal assignment path still falls to 'invalid path component'


newhoggy:
**Severity: Low**

## Summary

Split from #1292, whose Slice half is fixed by PR (in progress). This is the `Iterate` half: `get_path_mut` (`=`'s own walker, `src/jq/eval.rs`) has no way to represent a bare `.[]` inside a non-terminal parenthesized path component, and falls to the generic `invalid path component` catch-all instead of performing jq's per-element fan-out write.

## Repro (verified against pinned jq 1.7.1)

```
$ echo '{"a":[[1,2],[3,4]]}' | jq -c '(.a[])[0] = 9'
{"a":[[9,2],[9,4]]}
$ echo '{"a":[[1,2],[3,4]]}' | succinctly jq -c '(.a[])[0] = 9'
jq: error (at <stdin>:1): invalid path component
```

`|=` and `del()` already handle this exact shape correctly (confirmed live: `(.a[])[0] |= 99` and `del((.a[])[0])` both match jq), isolating the gap to `=`/`set_path`, same as #1292's slice half was.

## Why this is separate from #1292's slice fix

The slice half had a mechanical fix: flatten a nested `Pipe`/`Paren` component before `split_at_slice` scans for the slice position, then hand off to the existing `through_slice` machinery exactly as a top-level slice already does.

`Iterate` doesn't have an equivalent single hand-off point. `get_path_mut`'s signature is `Result<Option<&'a mut OwnedValue>, EvalError>` — it can only ever return **one** mutable slot. A `.[]` fans out over every element/value in a container, so applying whatever comes after it (`[0]` in the repro) means running the *rest* of the parent path against **each** element independently, not finding one shared parent to hand back. `update_path`/`delete_at_path` (which already get this right) do this by recursing into every element with the *remaining* filter/path, rather than ever trying to return one mutable reference — a structurally different shape than `get_path_mut`'s single-slot contract.

## Suggested fix direction

[...]

PR #1299, opened 2026-08-20
fix(jq): flatten a nested Pipe before split_at_slice scans for a slice

### LINKED ISSUE #1287: jq: computed index/slice on a non-container reports 'invalid path component' where jq names the type mismatch

### LINKED ISSUE #1292: jq: a Slice/Iterate inside a parenthesized non-terminal path component still falls to 'invalid path component'
**Severity: Low**

## Summary

`get_path_mut` (`=`'s own walker, `src/jq/eval.rs`) only knows how to flatten a nested `Pipe`/`Paren` path component down to `Field`/`Index`/`Identity` — a `Slice` or `Iterate` inside a non-terminal parenthesized sub-path still falls to the generic `invalid path component` catch-all, where real jq performs the write correctly.

Found during review of #1290 (which fixes the adjacent `Field`/`Index`-only case #1287 reported) — pre-existing on `main` before that PR, unaffected by it, and out of scope for #1287's own repro shapes.

Slice case:
```
$ echo '{"a":[1,2,3]}' | jq -c '(.a[0:1])[0] = 9'
{"a":[9,2,3]}
$ echo '{"a":[1,2,3]}' | succinctly jq -c '(.a[0:1])[0] = 9'
```

Also reproduces under `succinctly yq`: `printf 'a: [1,2,3]\n' | succinctly yq '(.a[0:1])[0] = 9'` hard-errors (exit 1) where real yq v4.53.3 silently no-ops (exit 0), matching the container-noop convention #1101/#1142 established for this class of write.

Iterate case:
```
$ echo '{"a":[[1,2],[3,4]]}' | jq -c '(.a[])[0] = 9'
{"a":[[9,2],[9,4]]}
$ echo '{"a":[[1,2],[3,4]]}' | succinctly jq -c '(.a[])[0] = 9'
```

Both confirmed still reproducing on top of #1290's fix.

## Root cause

`split_at_slice` (`src/jq/eval.rs`) only scans the *top-level* `exprs` list passed to `set_path`'s `Expr::Pipe` arm for a slice — it never recurses into a `Paren`/`Pipe` sub-expression appearing as one element of that list. `get_path_mut` has no `Expr::Slice` arm of its own (by design: a slice names a range, not a single mutable slot). So a slice hidden inside a parenthesized non-terminal component never reaches `through_slice` at all.

The `Iterate` case is architecturally harder: `get_path_mut`'s signature (`Result<Option<&'a mut OwnedValue>, EvalError>`) can only ever return one slot, so it cannot represent `.foo[]`'s fan-out the way `update_path`/`delete_at_path` do with their own per-element recursion.

## Suggested fix direction

Not investigated. For the `Slice` case, `split_at_slice` (or its caller) likely needs to recurse into a `Paren`/`Pipe` sub-expression before scanning for a slice, so a slice nested inside a parenthesized parent gets routed through `through_slice` the same way a top-level one already is. The `Iterate` case likely needs `get_path_mut`'s nested-`Pipe` handling (added by #1290) to special-case a mid-chain `Iterate` similarly to how `update_path`'s own `Expr::Pipe` arm already fans out per-element — but that may require restructuring `get_path_mut`'s single-slot return type, so it deserves its own scoping pass.

## Related

Found during review of #1290 (fixes #1287). Not caused by, or fixed by, that PR.

newhoggy:
## Summary

`(.a[0:1])[0] = 9` (a slice nested inside a parenthesized sub-path, itself only one of several top-level path components) reported a generic `invalid path component` instead of performing the write real jq gives.

```
$ echo '{"a":[1,2,3]}' | jq -c '(.a[0:1])[0] = 9'
{"a":[9,2,3]}
$ echo '{"a":[1,2,3]}' | succinctly jq -c '(.a[0:1])[0] = 9'   # before
```

Found during review of #1290 (fixes #1287, the adjacent `Field`/`Index`-only case) — filed as #1292, confirmed pre-existing on `main` before that PR and unaffected by it.

## Root cause

`set_path`'s `Expr::Pipe` arm calls `split_at_slice(exprs)` to find where a chained path crosses a slice — but that function only scanned the *top-level* `exprs` list. `(.a[0:1])` arrives as a single `Expr::Pipe([Field("a"), Slice{..}])` slot (the same `needs_path_prepass` fast-path shape #1287 fixed for `get_path_mut`'s own loop), so the slice hidden inside it was invisible to `split_at_slice`'s scan and the whole thing fell through to `get_path_mut`'s catch-all.

`|=` and `del()` already handle this exact shape correctly (confirmed live) — the gap was isolated to `=`/`set_path`.

## Fix

`split_at_slice` now flattens `exprs` via `push_path_components` (the same flattening helper already used elsewhere in this file, including by #1287's fix) before scanning for the slice position. `SliceSplit::before` changes from a borrowed `&'a [Expr]` to an owned `Vec<Expr>` since it's now built from a fresh flattened list rather than sliced directly out of the input.

## Scope note

The sibling `Iterate` case (`(.a[])[0] = 9`) is **not** fixed here — split out to #1298. It needs a structurally different fix: `get_path_mut`'s `Result<Option<&'a mut OwnedValue>, EvalError>` signature can only ever return one mutable slot, but `.[]` fans out over every element, so it can't be handled by the same mechanical flatten-then-scan technique this PR uses. Filing separately rather than blocking this fix on a bigger scoping question.

## Verification

- [x] The issue's own repro, plus a 3-level-deep-parens variant and a plain-unnested-slice regression check, all byte-match pinned jq 1.7.1
- [x] `yq` container-noop confirmed live against real yq v4.53.3 (writing through this shape silently no-ops, matching #1101/#1142's convention)
- [x] `cargo fmt --check`
- [x] `omni-dev coverage diff` — 97.06% patch coverage (33/34 new lines); the one uncovered line is an `unreachable!()` guard the preceding `position()` call already proves unreachable by construction (the original code had the identical pattern) — genuinely untestable, not a gap

Fixes #1292

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -11472,7 +11472,7 @@ fn set_path(
             if exprs.len() == 1 {
                 set_path(root, &exprs[0], new_value, scalar_noop, container_noop)
             } else if let Some(split) = split_at_slice(exprs) {
-                match get_path_mut(root, split.before)? {
+                match get_path_mut(root, &split.before)? {
                     None => Ok(()),
                     Some(parent) => through_slice(
                         parent,
@@ -11698,9 +11698,9 @@ fn through_slice<E: From<EvalError>>(
 }
 
 /// Where a chained path crosses a slice, as [`split_at_slice`] found it.
-struct SliceSplit<'a> {
+struct SliceSplit {
     /// The components to navigate before reaching the slice.
-    before: &'a [Expr],
+    before: Vec<Expr>,
     start: Option<i64>,
     end: Option<i64>,
     /// Everything left to apply *inside* the slice, as one expression.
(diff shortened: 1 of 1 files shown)


newhoggy:
Review round found two things, both addressed:

1. **Efficiency**: the flatten in `split_at_slice` unconditionally allocated+cloned even for the dominant unnested-chain case (`.a.b.c[2] = 9`), regressing its previous zero-allocation fast path. Fixed with a cheap top-level guard that's provably equivalent to running the flatten and finding nothing (pushed as `5d7807f`).
2. **Scope clarity**: `split_at_slice`'s flattening doesn't see through `Expr::Optional` (`push_path_components` leaves it opaque by design, for its other callers), so a `?`-marked slice — `(.a[0:1]?)[0] = 9` — still isn't found. Confirmed pre-existing and byte-identical before/after this PR's flattening (not a regression), and out of scope for #1292's own repro. Tightened the doc comment to say so explicitly and filed #1303 to track it separately.

Also noted (not fixed here, tracked separately): a minor double-clone of `split.before` (once building it in `split_at_slice`, once more inside `get_path_mut`'s own `.to_vec()`) — low-severity, would need a signature change to avoid, left as-is; and the broader observation (echoed across this PR's review and #1287's) that this file now has three independent hand-rolled "flatten a nested Pipe/Paren path component" implementations with diverging `Optional"-handling — folded into #1303's suggested-fix-direction as a scoping note rather than filed as its own issue.

Full suite + both clippy invocations + fmt all green on the updated branch; patch coverage 97.44% (38/39 new lines — the one uncovered line is the same genuinely-unreachable `unreachable!()` guard the original code already had).

Issue #1293, opened 2026-08-20
jq: |=/del() leak an inner (EXPR)?'s tolerance to path components that follow it


newhoggy:
**Severity: Low**

## Summary

`update_path` (`|=` and its compound-assign siblings) and `delete_at_path` (`del()`) both have an `Expr::Pipe(inner)` splice arm for a resolved path component that is itself a nested `Pipe` (`resolve_node`'s `?` arm can hand back `Optional(Pipe([...]))` when a branch resolves to more than one component). Both splice arms pass the accumulated `here`/optional flag — which folds in the nested pipe's own `?` — into the spliced continuation, so that tolerance keeps applying to path components *after* the nested pipe too, not just to failures inside it.

`=` (via `get_path_mut`, #1290) does not share this bug: it scopes each component's own optionality independently, matching jq.

Found during review of #1290 (fixes #1287) — confirmed pre-existing on `main` before that PR and unaffected by it; out of scope for #1287 itself.

## Repro (verified against pinned jq 1.7.1)

```
$ echo '{"a":{"b":5}}' | jq -c '(((.a.b)?).c) |= 9'
jq: error (at <stdin>:1): Cannot index number with string "c"
$ echo '{"a":{"b":5}}' | succinctly jq -c '(((.a.b)?).c) |= 9'
{"a":{"b":5}}
```

`.a.b` resolves fine to the number `5` (the `?` never actually fires), then `.c` tries to index that number and should raise — but the write silently no-ops instead, exit 0. Confirmed both on `main` (pre-#1290) and on #1290's branch, identical wrong output either way. `del()` reproduces the same divergence via its own sibling splice arm.

## Root cause

`update_path`'s `Expr::Pipe` arm (`src/jq/eval.rs`, ~line 12007-12011) does:
```rust
Expr::Pipe(inner) => {
    let mut spliced = inner.clone();
    spliced.extend_from_slice(&exprs[1..]);
    update_path::<S>(root, &Expr::Pipe(spliced), filter_expr, here, scalar_noop)
}
```
`here` (which is `optional || first_optional`, i.e. already true if the *nested pipe's own* wrapper was `?`) is threaded into the recursive call covering the *entire* spliced continuation — including `exprs[1..]`, the components that come *after* the nested pipe and were never themselves marked optional. `delete_at_path`'s equivalent arm has the identical shape.

## Suggested fix direction

[...]

Issue #1273, opened 2026-08-19
jq: parse_literal_exponent's saturation silently substitutes a fake exponent instead of signaling overflow


newhoggy:
**Severity: Low**

## Summary

Found during code review on #1270's own fix (PR #1272). `parse_literal_exponent`
(`src/jq/value.rs`) saturates to `i128::MIN`/`MAX` when an exponent digit
string overflows its target width, rather than signaling overflow distinctly.
Every caller with a ceiling check downstream (`format_overflow_literal_mantissa`'s
own `>= 1_000_000_000` check) is protected from this by coincidence, not by
design. `format_near_zero_literal` -- the one caller with no deliberate
ceiling (by design: it intentionally doesn't replicate real jq's own
`~1.147e9` internal breakdown) -- has no way to distinguish "this i128 is the
literal's real, exactly-parsed exponent" from "this i128 is a saturation
sentinel with no relationship to the source text," so it displays the
sentinel as if it were real.

#1270 (PR #1272) widened the saturation type from `i64` to `i128`, closing
the specific reported repro (a 24-digit exponent) and pushing the residual
boundary out by ~19 orders of magnitude (to ~38-39 digit exponents) -- but
this is explicitly a relocation of the bug's trigger threshold, not a
closure of the bug class itself. An exponent digit string with no length
limit (JSON/YAML number-literal grammar places none) can always eventually
exceed whatever finite integer width is chosen.

## Repro (post-#1270/PR #1272, still reachable)

```bash
$ python3 -c "print('1e-' + '9'*45)" | succinctly jq -c .
1E-170141183460469231731687303715884105728
```

`170141183460469231731687303715884105728` is exactly `i128::MIN`'s absolute
value -- a fixed, input-independent number substituted for the real
(45-nines) exponent. Notably, this sentinel is not even a directionally-safe
approximation: its magnitude (~1.7e38) is *smaller* than the true exponent's
magnitude (~1e45), so the output understates rather than overstates the
literal's real magnitude.

## Why this is separate from #1270

#1270's own scope, and the fix that shipped, was explicitly framed as
narrowing (not eliminating) this window -- see the updated doc comments on
`parse_literal_exponent` and the new `..._beyond_i128_range_1270` test in
`src/jq/value.rs`, both added by that PR's own code review to accurately
disclose the residual gap rather than claim it's closed.

## Suggested fix direction

Not investigated in depth. The structural fix code review proposed: have
[...]

PR #1305, opened 2026-08-20
fix(jq): scope update_path/delete_at_path's ?-guarded pipe splice to itself

### LINKED ISSUE #1294: jq: update_path's Expr::Pipe splice over-suppresses a ? guard past its own group
**Severity: Low**

## Summary

Found during #1287's review (a `get_path_mut` fix for `=` assignment paths).
`update_path` (the `|=`/`+=`/etc. walker) has a pre-existing, separate bug in
its own `Expr::Pipe(inner)` arm: a `?` wrapping a multi-component group like
`(.a|.c)?` correctly suppresses a failure *inside* that group, but also
incorrectly suppresses a failure in whatever pipe stage comes *after* the
group's closing paren — real jq does not extend the `?`'s protection that
far.

```bash
$ echo '{"a":{"c":5}}' | jq -c '((.a|.c)? | .y) = 9'
jq: error (at <stdin>:1): Cannot index number with string "y"

$ echo '{"a":{"c":5}}' | succinctly jq -c '((.a|.c)? | .y) |= 9'
{"a":{"c":5}}
```

`.a|.c` succeeds (`5`, a number) — the `?` on the group has nothing to
catch there. `.y` then fails on that number, a completely separate pipe
stage outside the `try (.a|.c)` the `?` desugars to. Real jq raises the
error normally; `succinctly jq`'s `|=` silently no-ops instead.

## Root cause

`src/jq/eval.rs`, `update_path`'s `Expr::Pipe(inner)` arm (around line
12045 pre-#1287, since renumbered by that PR's own additions elsewhere in
the file — search for the comment `"resolve_node's ? arm emits
Optional(Pipe([…]))"` within `update_path`):

```rust
Expr::Pipe(inner) => {
    let mut spliced = inner.clone();
    spliced.extend_from_slice(&exprs[1..]);
    update_path::<S>(root, &Expr::Pipe(spliced), filter_expr, here, scalar_noop)
}
```

`here = optional || first_optional` is threaded as the `optional` parameter
for the **entire** recursive call over `spliced` — which is `inner`'s own
components *plus* everything that came after the group (`exprs[1..]`, e.g.
`.y`). Since `update_path` carries `optional` forward unchanged through
every subsequent recursive step (its `Field`/`Index`/`Iterate` arms all
pass the *original* `optional`, not a per-step `here`, into their own
recursive call on `rest`), that inherited `true` from the group's `?`
persists through `.y` too, suppressing a failure there it should not.

## Why this is separate from #1287

#1287 was about `get_path_mut` (the `=` walker) not implementing this
splice at all. Its fix deliberately does **not** copy this arm's
`here`-for-the-whole-spliced-sequence approach — doing so would have
introduced this exact bug into `get_path_mut` too (verified: the fix's own
review round caught and avoided this by tracking suppression per-slot
instead of as one carried-forward flag). `delete_at_path`'s analogous
`Expr::Pipe` arm has the identical `here`-propagation shape and likely
shares this bug too, though not independently verified here.

## Suggested fix direction

Not investigated in depth. `update_path`'s `Expr::Pipe(inner)` arm needs
the same per-slot suppression boundary #1287's `get_path_mut` fix uses:
only `inner`'s own components should inherit `here`'s suppression: the
original `exprs[1..]` tail should keep the *original* `optional` parameter,
not `here`. Since `update_path` is recursive (not loop-based like
`get_path_mut`), the cleanest shape is likely two separate recursive calls
— one over `inner` with `optional: here`, continuing into `rest` with the
*original* `optional` once `inner` is exhausted — rather than a single
flattened `Pipe(spliced)` call. Worth checking `delete_at_path`'s sibling
arm for the same fix once this is scoped.

## Related

#1287 (`get_path_mut`'s sibling fix, PR #1290), which found and worked
around this without fixing it in `update_path`/`delete_at_path`.

newhoggy:
## Summary

`((.a|.c)? | .y) |= 9` and `del((.a|.c)? | .y)` silently no-op'd instead of raising when `.a|.c` succeeds and `.y` fails afterward.

```
$ echo '{"a":{"c":5}}' | jq -c '((.a|.c)? | .y) |= 9'
jq: error (at <stdin>:1): Cannot index number with string "y"
$ echo '{"a":{"c":5}}' | succinctly jq -c '((.a|.c)? | .y) |= 9'   # before
{"a":{"c":5}}
```

The group's own `?` correctly suppresses a failure *inside* `try (.a|.c)`, but real jq does not extend that protection past the closing paren — `.y` is a separate pipe stage outside the group's scope.

Found during review of #1290/#1299 — filed as #1294, independently duplicated by my own #1293 around the same time (closed in #1294's favor as the more precisely-scoped filing).

## Root cause

`update_path`'s and `delete_at_path`'s `Expr::Pipe(inner)` splice arms both thread `here` (which folds in the group's own `?`) as the blanket `optional` parameter for the *entire* spliced continuation — `inner`'s own components **plus** whatever follows the group (`exprs[1..]`). Since both functions carry `optional` forward unchanged through every subsequent recursive step, that inherited `true` from the group's `?` keeps suppressing `.y` too.

Independently present in both functions — the exact bug class #1287's review round found and fixed for `=` (`get_path_mut`), just never ported to these two siblings. `get_path_mut`'s own fix used a per-slot tracking array (loop-based, so a parallel `Vec<bool>` was natural); these two are pure recursive-descent, so the equivalent needed a different shape.

## Fix

Wrap the group's own components in `Expr::Optional` when `here` is set — `unwrap_path_component` already peels this at every step, so each spliced-in component now self-reports its own optionality instead of relying on a blanket flag. Pass the *original* `optional` (not `here`) as the spliced call's baseline, so anything after the group keeps whatever optionality it already had on its own, unaffected by the group's `?`.

## Verification

- [x] Both confirmed repros (`|=`, `del()`) now byte-match pinned jq 1.7.1
- [x] Boundary preserved: the group's own failure (first or second component) is still correctly suppressed — only what comes *after* it stops being covered
- [x] Non-optional splice case (`(.a|.c)[0] |= 9`, `del((.a|.c)[0])` — no `?` anywhere) unaffected, confirmed still correct
- [x] yq mode sanity-checked live (shared code path, `S::TAG`-agnostic)
- [x] `cargo fmt --check`
- [x] `omni-dev coverage diff` — 100% patch coverage (49/49 new lines)

Fixes #1294

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -12077,10 +12077,38 @@ fn update_path<S: EvalSemantics>(
                     // when a branch resolved to more than one component, so
                     // unwrapping can expose a nested pipe. Splice it in rather
                     // than recursing on it alone, which would strand `rest`.
+                    //
+                    // Wrap the group's own components in `Optional` when
+                    // `here` is set, rather than threading `here` through as
+                    // this call's blanket `optional` -- that kept suppressing
+                    // whatever follows the group too, past its own scope
+                    // (#1294): `(.a|.c)? | .y` must suppress a failure in
+                    // `.a`/`.c` (the `try (.a|.c)` the `?` desugars to) but
+                    // not `.y` failing afterward, outside that scope. Each
+                    // spliced-in component now self-reports its own
+                    // optionality via the wrapper, which `unwrap_path_component`
+                    // already peels at every step -- `rest` (`exprs[1..]`),
+                    // left unwrapped, keeps whatever optionality it already
+                    // had on its own once `optional` (not `here`) is passed
+                    // as this call's baseline again.
                     Expr::Pipe(inner) => {
-                        let mut spliced = inner.clone();
+                        let mut spliced: Vec<Expr> = if here {
+                            inner
+                                .iter()
+                                .cloned()
+                                .map(|e| Expr::Optional(Box::new(e)))
+                                .collect()
+                        } else {
+                            inner.clone()
+                        };
                         spliced.extend_from_slice(&exprs[1..]);
-                        update_path::<S>(root, &Expr::Pipe(spliced), filter_expr, here, scalar_noop)
+                        update_path::<S>(
+                            root,
+                            &Expr::Pipe(spliced),
+                            filter_expr,
+                            optional,
+                            scalar_noop,
+                        )
                     }
                     // The chain continues *inside* the slice, so the update
                     // runs against the sub-array and is spliced back:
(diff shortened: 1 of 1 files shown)


newhoggy:
Review round found two things, both addressed:

1. **Duplication**: the splice logic was byte-identical between `update_path`'s and `delete_at_path`'s `Expr::Pipe(inner)` arms — extracted to a shared `splice_optional_group` helper next to `unwrap_path_component`, so a future scoping fix only needs to land once.
2. **yq coverage gap**: the new tests only exercised jq mode. Added yq-mode cases — using an array intermediate rather than the jq tests' scalar one, since `is_yq_field_index_noop_scalar` would have made a scalar-intermediate test pass trivially regardless of whether this fix existed (yq no-ops `|=`/`del()` on a scalar target unconditionally).

Also rebased onto current `main` (picked up #1273/#1168/#888's merges) — clean rebase, full suite/clippy/fmt re-verified green, patch coverage 100% (46/46 new lines).

Issue #498, opened 2026-07-30
jq: `?` on a write path swallows the failure to apply it, not just to walk it


newhoggy:
**Severity: Medium-High** — `=`, `|=` and `+=` return exit 0 and a document that silently did not get the write, where jq raises. Reproduced against the built CLI (release, `372b239a`), jq oracle 1.7.1.

Split out of #413 so that issue can close with PR #482, which fixes the *resolver* half (`?` swallowing an error raised while evaluating the key or the target). This is the *writer* half, and it needs a different diff.

A `?` on a write path suppresses the failure to **apply** the path, not just the failure to walk it:

```
$ echo '[1,2]' | sjq -c '.[-5]? = 9'
[1,2]                       # jq: error: Out of bounds negative array index

$ echo '[1,2]' | sjq -c '.[-5]? |= 9'
[1,2]                       # jq: same error

$ echo '[1,2]' | sjq -c '.[-5]? += 1'
[1,2]                       # jq: same error
```

Without the `?` all three raise correctly, so the `?` is doing it. No computed key, no `..`, no multi-output prefix — the simplest possible path reaches it.

The multi-branch flavour, [reported on #413](https://github.com/rust-works/succinctly/issues/413#issuecomment-5129824293), is the same defect reached through #493's fan-out: a later write in the same assignment hits a container an earlier write already clobbered, and the `?` absorbs the resulting type error.

```
$ echo '{"k":"a","a":{"k":"a","a":1}}' | sjq -c '(.. | objects | .[.k]?) = 7'
{"k":"a","a":7}             # jq: error: Cannot index number with string "a"
```

### The rule

`?` prunes at path *production* and never covers path *application*. jq's own `path()` says which side a case falls on:

| filter on input | jq `path()` | jq write | here |
|---|---|---|---|
| `"str" \| .a?` | `[]` — no path | `"str"`, no-op | path `["a"]`, write swallowed → right answer, wrong mechanism |
| `{"a":1} \| .a.b?` | `[]` | `{"a":1}`, no-op | path `["a","b"]`, swallowed → right answer, wrong mechanism |
| `[1,2] \| .[-5]?` | `[[-5]]` — path exists | **raises** | swallowed ✗ |
| `[1,2] \| .[5]?` | path exists | pads → `[1,2,null,null,null,9]` | swallowed ✗ |

### Where

`set_path`'s `Expr::Optional` arm (`src/jq/eval.rs`), and `update_path`'s counterpart:

```rust
[...]

PR #1310, opened 2026-08-20
fix(jq): a ?-marked slice in an assignment path was never suppressed correctly

### LINKED ISSUE #1303: jq: a ?-marked slice inside an assignment path still falls to 'invalid path component'
**Severity: Low**

## Summary

Found during review of #1299 (fixes the sibling non-optional case, #1292). `split_at_slice`'s flattening (added by #1299) reuses `push_path_components`, which deliberately leaves `Expr::Optional` opaque (its other callers need that). A `?`-marked slice — whether at the top level or nested inside a parenthesized sub-path — is therefore still invisible to `split_at_slice`'s scan, and `get_path_mut` has no `Expr::Slice` arm of its own to fall back on either. Confirmed pre-existing on `main` (both before and after #1299's flattening was added — identical wrong output either way), so it's a separate, unaffected gap, not a regression.

Nested in parens:
```
$ echo '{"a":[1,2,3]}' | jq -c '(.a[0:1]?)[0] = 9'
{"a":[9,2,3]}
$ echo '{"a":[1,2,3]}' | succinctly jq -c '(.a[0:1]?)[0] = 9'
```

Same for the unparenthesized dot-chain spelling: `.a[0:1]?[0] = 9`.

A genuinely separate, related symptom — the top-level case with no nesting at all:
```
$ echo '{"a":[1,2,3]}' | jq -c '.a[0:1]? = 9'
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a[0:1]? = 9'
{"a":[1,2,3]}
```
Here succinctly silently no-ops instead of raising — `set_path`'s `Expr::Optional`/`Expr::Slice` arms handle this shape (no `split_at_slice` involved at all, since it's a single-element `Pipe`), but apparently swallow an error `?` shouldn't cover for a slice assignment. Grouped here since both are the same underlying "a `Slice` wrapped in `Optional`" theme, but may need separate root-cause tracing.

## Root cause

`push_path_components`'s `Expr::Optional` handling: it recurses into `Pipe`/`Paren`, drops `Identity`, and pushes everything else (including `Optional(...)`) verbatim rather than peeling it — by design, since its other callers (`resolve_seq`, `is_passthrough_target`) need `Optional` to stay a distinct marker for their own per-branch try/catch semantics. `split_at_slice`'s new use is structurally different (a plain `matches!(e, Expr::Slice{..})` scan feeding the non-recursive `get_path_mut`), so it can't "see through" the same opaque wrapper the way a recursive evaluator can.

## Suggested fix direction

Not investigated in depth. This is the third place now (after #1287's `get_path_mut` fix and #1290's review) where a hand-rolled "flatten a nested Pipe/Paren path component" implementation has had to be built or reasoned about separately — `push_path_components`, `flatten_delete_path`/`DeleteStep` (which *does* resolve `Optional`, carrying a per-step `optional: bool`), and `get_path_mut`'s own inline splice loop (#1287) are now three independent copies with different `Optional`-handling contracts. Worth a scoping pass on whether `split_at_slice` needs its own `Optional`-resolving flattener (mirroring `flatten_delete_path`'s pattern) rather than reusing `push_path_components`, and whether that's also the moment to consolidate the three into one shared implementation — CLAUDE.md's own "duplicated predicates diverge silently" lesson (#106) applies directly here: this exact divergence is what produced the gap this issue reports.

## Related

Found during review of #1299 (fixes #1292, the sibling non-`?` case). Not caused by, or fixed by, that PR.

newhoggy:
## Summary

Two independent gaps around a `?`-marked slice in an `=` assignment path, both confirmed against real jq 1.7.1.

**Gap 1 — write-time application errors wrongly suppressed:**
```
$ echo '{"a":[1,2,3]}' | jq -c '.a[0:1]? = 9'
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a[0:1]? = 9'   # before
{"a":[1,2,3]}
```

**Gap 2 — `?`-marked slice not found when more path follows:**
```
$ echo '{"a":[1,2,3]}' | jq -c '(.a[0:1]?)[0] = 9'
{"a":[9,2,3]}
$ echo '{"a":[1,2,3]}' | succinctly jq -c '(.a[0:1]?)[0] = 9'   # before
```

## Root cause

**Gap 1**: `set_path`'s `Expr::Optional` arm swallowed *any* error from the inner `set_path` call, except one carved-out message (`is_negative_index_out_of_bounds`, #498). Real jq's inline `?` only suppresses a failure to *reach* a target while collecting a path — not a mismatch in what gets written once one is found. `A slice of an array can only be assigned another array` and `Cannot update string slices` are the same kind of write-time application check as the negative-index one, and were falling through the un-carved-out `Err(_) => Ok(())` arm instead.

**Gap 2**: `split_at_slice`'s scan (added by #1292, PR #1299) matched a bare `Expr::Slice` directly — never one wrapped in `Expr::Optional`. `.a[0:1]?` puts the `?` directly on the `Slice` component (mirroring how `.a.b?` wraps just `Field("b")`), so `(.a[0:1]?)[0]`'s flattened form is `[Field("a"), Optional(Slice{0,1}), Index(0)]`, and the scan's `matches!(e, Expr::Slice{..})` never matched the wrapped element.

## Fix

**Gap 1**: Generalized `EvalError::is_negative_index_out_of_bounds` into `is_write_time_application_error`, covering all three message kinds (negative-index, slice-assign-non-array, cannot-update-string-slices). A genuinely non-sliceable target (`true[0:1]? = 9`) is unaffected — that's a navigation failure, which `?` still correctly suppresses, confirmed live both before and after.

**Gap 2**: `split_at_slice`'s scan (and its own cheap top-level guard) now matches via `unwrap_path_component` — already used everywhere else in this file for exactly this kind of per-component `?`/`(...)` peeling — instead of a bare pattern match. `SliceSplit` gains an `optional: bool` field threading whether the slice itself was `?`-marked into `through_slice`'s own `optional` parameter, so a non-sliceable target under this same nested shape still correctly suppresses (unaffected by gap 1's fix, since `through_slice` never gates the write-time application checks on `optional` at all).

Does **not** reach through a `Pipe`/`Paren` still nested *inside* an `Optional` wrapper (`(.a|.c)?[0] = 9`) — confirmed live that this doesn't parse in either real jq or succinctly (postfix `?` cannot be followed by `[...]`), so there's nothing to find there either way.

## Verification

- [x] Both gaps' repros, plus the pre-existing negative-index carve-out and two non-sliceable-target boundary cases, all byte-match pinned jq 1.7.1
- [x] `cargo fmt --check`
- [x] `omni-dev coverage diff` — 100% patch coverage (64/64 new lines)

Fixes #1303

### DIFF
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -862,7 +862,7 @@ impl EvalError {
     /// that the *filter* is not a path expression, not a runtime value
     /// error. `?` only suppresses failures raised while collecting a path
     /// (a missing key, an out-of-range index, ...); this survives it, the
-    /// same way [`Self::is_negative_index_out_of_bounds`] does for the write
+    /// same way [`Self::is_write_time_application_error`] does for the write
     /// side (#530: confirmed live — `path(("a")?)` still raises in jq). The
     /// lone call site that needs to tell the two apart is `resolve_node`'s
     /// bare-`?` arm in `eval.rs`.
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -568,14 +568,34 @@ impl EvalError {
     }
 
     const OUT_OF_BOUNDS_NEGATIVE_INDEX: &'static str = "Out of bounds negative array index";
+    const SLICE_ASSIGN_NON_ARRAY: &'static str =
+        "A slice of an array can only be assigned another array";
+    const CANNOT_UPDATE_STRING_SLICES: &'static str = "Cannot update string slices";
 
-    /// Whether this is exactly [`Self::out_of_bounds_negative_index`] — the one
-    /// write-time bounds check jq's `?` does not suppress, unlike every other
-    /// indexing error it raises (`?` only suppresses errors raised while
-    /// *collecting* a path, not this one). The lone call site that needs to tell
-    /// the two apart is `set_path`'s `Expr::Optional` arm in `eval.rs` (#498).
-    pub fn is_negative_index_out_of_bounds(&self) -> bool {
+    /// Whether this is one of the write-time *application* checks jq's `?`
+    /// does not suppress, unlike every other indexing error it raises (`?`
+    /// only suppresses a failure to *reach* a target while collecting a
+    /// path; once a write is confirmed to land somewhere, a mismatch in what
+    /// gets written there survives even an inline `?`). Three messages
+    /// qualify, each confirmed live against jq 1.7.1: [`Self::out_of_bounds_negative_index`]
+    /// (`.a[-5]? = 9` still raises), [`Self::slice_assign_non_array`]
+    /// (`.a[0:1]? = 9`, a non-array RHS, still raises), and
+    /// [`Self::cannot_update_string_slices`] (`"str"[0:1]? = "x"` still
+    /// raises) — #498, #1303.
+    ///
+    /// Contrast a genuinely *non-sliceable* target (`true[0:1]? = 9`) — that
+    /// one **is** suppressed, since it's a navigation failure (the slice
+    /// never applies to a boolean at all), not a write-time application one;
+    /// `through_slice`'s own `_ if optional => Ok(())` arm already handles
+    /// that case correctly and is unaffected by this predicate.
+    ///
+    /// The lone call site that needs to tell these apart from an ordinary
+    /// (suppressible) navigation failure is `set_path`'s `Expr::Optional`
+    /// arm in `eval.rs`.
+    pub fn is_write_time_application_error(&self) -> bool {
         self.message == Self::OUT_OF_BOUNDS_NEGATIVE_INDEX
+            || self.message == Self::SLICE_ASSIGN_NON_ARRAY
+            || self.message == Self::CANNOT_UPDATE_STRING_SLICES
     }
 
     /// `Array/string slice indices must be integers`.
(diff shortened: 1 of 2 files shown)


newhoggy:
Extensive review round (8 finder angles) found and confirmed several things:

**Fixed here:**
1. **CI-breaking**: renaming `is_negative_index_out_of_bounds` → `is_write_time_application_error` missed two prose references (one a rustdoc intra-doc link in `error.rs`, one a plain comment in the new test) — `cargo doc --no-deps --all-features` with `-D warnings` (CI's own Documentation check) failed. Both fixed and re-verified with the exact CI command.
2. **Doc-comment correction**: this PR's own `split_at_slice` doc comment claimed `((.a[0:1])? | .[0]) = 9`-shaped queries "cannot occur" via real jq syntax. Review disproved that live (`path((.a[0:1])? | .[0])` resolves fine, unlike the glued `(.a|.c)?[0]` form which is a confirmed parse error) — corrected the comment to accurately describe the gap and point to a follow-up rather than leave a false claim in merged code.

**Filed as follow-ups** (both confirmed pre-existing, unrelated to introducing a regression, out of scope for #1303's own repro):
- #1311 — the deeper `?`-on-a-whole-group case above; closing it properly likely needs reworking `split_at_slice`'s flatten-then-scan design into `update_path`'s already-correct one-element-at-a-time approach, not another point patch
- #1312 — `del()` on a `?`-marked slice targeting a string wrongly no-ops (arm-ordering bug in `delete_at_path`'s own standalone Slice arm, unrelated function/code path)

Two more findings (a double `unwrap_path_component` call, a double-negated `.all(!...)` guard) were confirmed real but purely stylistic/negligible-cost — left as-is per reviewer's own "not a real correctness or maintenance risk" assessment, to keep this diff focused.

Rebased onto current main, full suite/clippy/fmt/doc-build all re-verified green, patch coverage 100% (78/78 new lines).

Issue #1191, opened 2026-08-17
yq: YamlValue::Alias::as_str() only unwraps one level of alias indirection, unlike its type_name()/as_object()/as_array() siblings


newhoggy:
## Summary

`YamlValue::as_str()`'s `Alias` arm (`src/yaml/light.rs`, around line 6069) only unwraps a single level of alias indirection:

```rust
YamlValue::Alias { target, .. } => {
    target.and_then(|t| {
        if let YamlValue::String(s) = t.value() {
            s.as_str().ok().map(|cow| Cow::Owned(cow.into_owned()))
        } else {
            None
        }
    })
}
```

If `t.value()` is itself another `Alias` (an alias pointing at another alias, rather than directly at a `String`), this returns `None` — even though the chain ultimately resolves to a perfectly valid string. Every sibling accessor on the same type (`type_name()`, `as_bool()`, `as_i64()`, `as_f64()`, `as_object()`, `as_array()`) instead recurses fully by calling `t.value().xxx()` rather than pattern-matching `t.value()` directly, so they all correctly resolve a chained alias while `as_str()` alone stops after one hop.

## Repro (verified live)

```console
$ printf 'x: &a hello\ny: &b *a\nz: *b\n' | yq '.z'
hello
$ printf 'x: &a hello\ny: &b *a\nz: *b\n' | succinctly yq '.z'
null
```

`type_name()` on `.z` correctly recurses through both alias hops and reports `"string"`, but `as_str()` returns `None` for the exact same node — so any caller that branches on `type_name() == "string"` while relying on `as_str()` to get the value (a completely reasonable pattern) gets a contradiction: told it's a string, but unable to read it as one.

## Scope note

This is a real, independently-reachable bug on `main` today, unrelated to any in-flight work — found while investigating issue #1098's own materialization gap (a doubly-aliased string was one of several ways to trip `type_name() == "string" && as_str() == None`, a signal #1098's original PR briefly (and incorrectly) treated as always meaning "decode failure"). Chained anchors/aliases are a normal pattern in layered-defaults YAML (Kubernetes manifests, CI/CD configs, `&base`/`<<: *base`\-style merges), so this isn't an exotic edge case.

## Suggested direction

Change `as_str()`'s `Alias` arm to recurse the same way the other accessors do (`t.value().as_str()` instead of matching `t.value()` against `YamlValue::String` directly), mirroring `type_name()`'s own recursive structure.

newhoggy:
[...]

PR #1314, opened 2026-08-20
fix(yaml): bound alias-chain depth to stop stack-overflow DoS

### LINKED ISSUE #1193: yaml: unbounded alias-chain recursion in YamlValue::Alias accessors causes a stack-overflow DoS on valid, non-cyclic YAML
## Summary

None of `YamlValue::Alias`'s scalar accessors (`as_bool`, `as_i64`, `as_f64`, `number_literal`, `type_name`, `as_object`, `as_array` -- `src/yaml/light.rs`) carry any recursion-depth guard. Each resolves a chained alias via `target.and_then(|t| t.value().<same accessor>())`, i.e. it fully re-enters itself once per alias hop. There is nothing here analogous to `to_owned_at_depth`'s own `assert_nesting_depth`/`MAX_NESTING_DEPTH` check, which only bounds JSON/YAML *container* nesting (arrays/objects) -- not YAML *alias-chain* length.

A long, syntactically valid, **non-cyclic** chain of anchored aliases (`k0: &a0 v`, `k1: &a1 *a0`, `k2: &a2 *a1`, ...) drives real, unbounded stack recursion. This isn't caught by issue #153's existing acyclicity check, since the chain never cycles -- it's just very long.

## Repro

Confirmed live on both debug and release builds: querying the tail of a ~50,000-deep alias chain crashes with `thread 'main' has overflowed its stack` / `fatal runtime error: stack overflow, aborting` (exit 134/SIGABRT) -- an *uncatchable* process abort (unlike a `panic!()`, this can't be caught by `catch_unwind` even if the CLI had one). A 20,000-deep chain was still safe; 50,000 was not, in a YAML file of only a few MB.

```console
$ python3 -c "
for i in range(50000):
    print(f'k{i}: &a{i} ' + (\"v\" if i == 0 else f'*a{i-1}'))
print('z: *a49999')
" > chain.yaml
$ succinctly yq '.z | length'
fatal runtime error: stack overflow
[1]    12345 abort      succinctly yq '.z | length'
```

## Scope note

This is **not specific to any in-flight PR** -- `to_owned_at_depth` (`src/jq/eval_generic.rs`) already calls `as_bool()`, `number_literal()`, `as_i64()`, and `as_f64()` on every YAML scalar (all four fully alias-chain-recursive) before it would ever reach `as_str()`/`type_name()`, so this crash is reachable through code paths that predate and are independent of #1098/#1190/#1191/#1192. Found while investigating those issues, but it's a separate, pre-existing bug in `main` today.

## Suggested direction

Thread a chain-length counter through `YamlValue::Alias`'s recursive accessors, mirroring `assert_nesting_depth`'s existing container-depth guard in `eval_generic.rs` -- likely as its own named constant (alias-chain depth is a different quantity than container nesting depth, and #153's own cycle-detection machinery may already have relevant infrastructure to build on). A clean, catchable error (or a deliberate, documented panic, matching this codebase's existing precedent for adversarial-input-only guards) is preferable to an uncatchable stack overflow either way.

newhoggy:
## Summary

- `YamlValue::Alias`'s typed accessors (`as_bool`/`as_i64`/`as_f64`/`number_literal`/`as_object`/`as_array`/`type_name`/`is_null`) each resolved one alias hop via `target.and_then(|t| t.value().<same accessor>())`, re-entering themselves once per hop. A syntactically valid, **non-cyclic** chain of tens of thousands of anchored aliases drove real, uncatchable stack overflow (`SIGABRT`, exit 134) through that self-recursion — confirmed live, matching the issue's own repro.
- This is distinct from #153's existing build-time cycle check, which only rejects a chain that revisits its own anchor, not one that is merely very long.
- Fix: replace the self-recursion with an explicit iterative walk (`YamlValue::resolve_alias_chain`), which costs O(1) stack regardless of chain length, plus a `MAX_ALIAS_CHAIN_DEPTH` (65,536) ceiling — via the crate's existing shared `assert_depth` helper, mirroring `MAX_NESTING_DEPTH`/`MAX_VALUE_TREE_DEPTH` — to bound the worst-case CPU time one accessor call can be forced to spend on a pathologically long adversarial chain. Past the cap it panics cleanly (exit 101 with a message), not via stack corruption.

## Verification

- Confirmed the crash reproduces on `main` today (`[.z]` over a 50,000-hop chain → exit 134, `thread 'main' has overflowed its stack`).
- Confirmed it no longer crashes on this branch (same query → exit 0, correct value for an int/bool-terminated chain).
- Confirmed `as_i64`/`as_bool` correctly resolve the *full* chain (not just avoid crashing) at both moderate (25-hop) and extreme (50,000-hop) depths.
- Confirmed the depth cap itself fires cleanly (exit 101, not a crash) past 65,536 hops.
- Full existing test suite green (`cargo test --features cli,simd,regex,serde`, 0 failures), both CI `clippy` legs clean, `cargo fmt --check` clean, `cargo check --no-default-features` (no_std) clean.

## Scope notes (deliberately not fixed here — filed as follow-ups)

- `as_str`/`key_string` were **not** touched — they were already non-recursive (bounded to a single alias hop), so not a DoS vector, but they also don't correctly resolve a chain more than 1 hop deep (a separate, pre-existing correctness gap, confirmed present on `main` today too).
- Identity/streaming output (`succinctly yq '.z'` with no wrapping) shows the alias's *literal source text* (`*a0`) rather than resolving it at all for a multi-hop chain — a different, pre-existing code path (the M2/P9 direct-streaming writer), also confirmed present on `main` today, unrelated to the accessors this PR fixes.
- Full-document materialization (`.`, or any query that forces the full/owned evaluator) on a document shaped like "every field's own alias chain is one hop longer than the previous field's" costs O(n²) — confirmed via direct A/B timing to be **identical** between `main` and this branch (same total work, this PR only removes the *stack growth*, not the *time complexity*). A genuinely compact adversarial document could still exploit this for CPU-time (not crash) DoS; that needs alias-target memoization, a deeper architectural change out of scope here.

I'll file GitHub issues for the three items above right after this PR.

Fixes #1193

### DIFF
--- a/docs/compliance/yaml/limitations.md
+++ b/docs/compliance/yaml/limitations.md
@@ -199,6 +199,33 @@ Neither costs this page's conformance numbers anything: no case in the corpus co
 alias to an out-of-scope anchor, so the loader-alone reject figure, the accepted-but-invalid
 document count, and the `lax:anchors` row are unmoved by #372.
 
+### A third, differently-timed check: excessively long alias chains
+
+Unlike the cyclic and unknown-anchor cases above, an alias chain that is merely very long —
+`a0: &a0 v`, `a1: &a1 *a0`, `a2: &a2 *a1`, ... — is not rejected at `YamlIndex::build`
+time; the index builds successfully regardless of chain length. [#153](https://github.com/rust-works/succinctly/issues/153)'s
+cycle check only rejects a chain that revisits its own anchor, and a long non-cyclic chain
+never does. Before [#1193](https://github.com/rust-works/succinctly/issues/1193), *walking*
+such a chain (via `type`, `tostring`, `-o json` output, arithmetic, or any other query that
+resolves the aliased value) drove real, uncatchable stack overflow through several
+independent self-recursive code paths in `src/yaml/light.rs`, `src/bin/succinctly/yq_runner.rs`,
+and `src/jq/eval.rs` — the same underlying failure mode #153 fixed for cycles specifically,
+recurring for long-but-acyclic chains that #153's own check does not cover.
+
+The fix is a runtime ceiling, not a build-time rejection: `MAX_ALIAS_CHAIN_DEPTH` (65,536
+hops, `src/yaml/light.rs`) bounds every alias-resolving accessor to a clean `panic!` past
+that depth, rather than allowing the resolution loop to run unbounded. Because the fix
+converts what used to be self-recursion into an explicit iterative loop, the ceiling exists
+only to cap adversarial CPU time, not stack depth — the loop itself costs O(1) stack
+regardless of chain length. A chain within the limit resolves normally; one beyond it
+panics with a `nesting depth exceeds limit` message (a clean Rust panic, not a stack
+overflow) rather than crashing the process the way #1193's own repro did:
+
+```
+$ succinctly yq -o json '.' 70000-hop-chain.yaml
+thread 'main' panicked at ...: nesting depth exceeds limit of 65536
+```
+
 ### The other ways `YamlIndex::build` fails
 
 Structure aside, `build` still refuses input it has no reading for at all. None of these is
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -233,12 +233,14 @@ fn yaml_to_owned_value<W: AsRef<[u64]>>(cursor: YamlCursor<'_, W>) -> Result<Own
             Ok(OwnedValue::Array(arr))
         }
         YamlValue::Alias { target, .. } => {
-            // Resolve alias by following the target cursor
-            if let Some(target_cursor) = target {
-                yaml_to_owned_value(target_cursor)
-            } else {
-                // Unresolved alias - treat as null
-                Ok(OwnedValue::Null)
+            // Resolve the *entire* alias chain first (#1193), not just this
+            // one hop: the resolved cursor's own `.value()` is guaranteed
+            // non-`Alias`, so this recursive call terminates in exactly one
+            // more step regardless of chain length.
+            match target.and_then(|t| t.resolve_alias_target_cursor()) {
+                Some(resolved) => yaml_to_owned_value(resolved),
+                // Unresolved (dangling) target - treat as null
+                None => Ok(OwnedValue::Null),
             }
         }
         YamlValue::Error(msg) => Err(anyhow::anyhow!("YAML error: {msg}")),
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -23468,11 +23468,14 @@ fn yaml_value_to_owned<W: Clone + AsRef<[u64]>>(
             OwnedValue::Object(map)
         }
         YamlValue::Alias { target, .. } => {
-            // Resolve alias by following the target cursor
-            if let Some(target_cursor) = target {
-                yaml_value_to_owned(target_cursor)
-            } else {
-                OwnedValue::Null
+            // Resolve the *entire* alias chain first (#1193), not just this
+            // one hop: the resolved cursor's own `.value()` is guaranteed
+            // non-`Alias`, so this recursive call terminates in exactly one
+            // more step regardless of chain length.
+            match target.and_then(|t| t.resolve_alias_target_cursor()) {
+                Some(resolved) => yaml_value_to_owned(resolved),
+                // Unresolved (dangling) target - treat as null
+                None => OwnedValue::Null,
             }
         }
         YamlValue::Error(_) => OwnedValue::Null,
--- a/src/jq/mod.rs
+++ b/src/jq/mod.rs
@@ -105,8 +105,9 @@ pub use stream::{StreamError, StreamStats, StreamableValue};
 pub use value::{
     assert_value_tree_depth, format_number_jq_compat, NumberRepr, OwnedValue, MAX_VALUE_TREE_DEPTH,
 };
-// `pub(crate)`, not `pub`: only `yaml::light`'s `resolve_alias_chain` (#1191
-// code review) needs this from outside `jq::value` -- re-exporting it as a
-// real public API item would put a crate-internal depth-guard helper on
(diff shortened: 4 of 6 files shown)


newhoggy:
Update: review (3 independent angles) found that the first commit only fixed `YamlValue`'s typed `DocumentValue` accessors — the exact same self-recursive shape was still live and independently crashable (confirmed via SIGABRT repro, including this repo's own CLAUDE.md-documented `syq -o json '.' config.yaml`) in five more places:

- `YamlCursor::stream_json_value` / `::write_json_to` (the YAML→JSON streaming/output writers)
- `YamlCursor::tag`
- `yaml_to_owned_value` (`src/bin/succinctly/yq_runner.rs`)
- `yaml_value_to_owned` (`src/jq/eval.rs`)

Pushed a second commit that adds `YamlCursor::resolve_alias_chain` (the cursor-level counterpart to `YamlValue::resolve_alias_chain`, which now delegates to it instead of duplicating the loop) and routes all five sites through it, plus 2 more regression tests covering the streaming-output and owned-evaluator-bridge paths specifically. Full test suite + both CI clippy legs + `no_std` check + `cargo fmt --check` all green after the update; rebased cleanly onto latest `main`.

Two more single-hop-only (non-crash, correctness-only) gaps turned up during the same review and are filed as follow-ups rather than folded in here: `merge_sources` (#1318) and `explicit_tag` (#1319), joining the earlier #1315 (`as_str`/`key_string`), #1316 (identity-output streaming), and #1317 (O(n²) full-document materialization cost).

newhoggy:
Update: rebased onto \`main\` (which had meanwhile merged #1346/#1191 — a sibling fix for \`as_str()\`'s own alias-chain bug, whose author was tracking this PR and deliberately designed their \`YamlCursor::resolve_alias_chain\` identically so the eventual rebase would be a clean reconciliation rather than a real conflict — much appreciated).

Reconciled as follows:
- Kept \`main\`'s existing (already 3-round-reviewed) \`YamlCursor::resolve_alias_chain\` — returns the resolved *value* directly, avoiding a redundant \`.value()\` call — and extended its use to the 7 typed accessors it hadn't reached yet (\`is_null\`/\`as_bool\`/\`as_i64\`/\`as_f64\`/\`number_literal\`/\`as_object\`/\`as_array\`/\`type_name\`; \`as_str\` was already fixed by #1191).
- Added a new, separately-named \`YamlCursor::resolve_alias_target_cursor\` (returns the resolved *cursor*, not just its value) for the 5 sites that need to re-invoke a cursor method on the target (\`stream_json_value\`, \`write_json_to\`, \`tag\`, \`yaml_to_owned_value\`, \`yaml_value_to_owned\`) — kept as its own loop rather than wrapping the value-returning one, for the same "don't double-compute \`.value()\`" reason #1191's own review already established.
- Dropped my original branch's own \`YamlValue\`-level \`resolve_alias_chain\` wrapper entirely — \`main\`'s review had already found and removed the identical pattern as buggy (off-by-one, contract mismatch, redundant work) before I got here.

Full local verification suite green after reconciliation: build (debug+release), full test suite (all 40+ binaries, 0 failures), both CI clippy legs, \`cargo doc --no-deps --all-features\` under \`-D warnings\`, \`no_std\` check, \`cargo fmt --check\`. Pushed as a force-update (rebase, not a new commit) since main had advanced significantly.

newhoggy:
Second review round (6 finder angles, most re-verified live against the built binary and against real \`yq\`/pyyaml) found the core fix correct — all boundary-arithmetic, loop-equivalence, postcondition, multi-document-scoping, and \`panic=abort\` checks came back clean. Real, actionable findings, addressed in the latest commit:

- **Missing \`#[track_caller]\`** on \`resolve_alias_target_cursor\` (its sibling \`resolve_alias_chain\` has it) — confirmed live that panics through its 5 call sites all collapsed to the same shared-loop line. Fixed; re-verified panics now report their own distinct call-site line.
- **Doc-comment gap**: the "why not compose the two loops" rationale only argued one direction. Strengthened to cover both, and to honestly acknowledge the double-\`.value()\`-computation cost at the 5 cursor-consuming call sites (a real, small, non-chain-scaling cost — not something worth a deeper restructure given the risk/complexity tradeoff).
- **\`docs/compliance/yaml/limitations.md\`** hadn't been updated for the new runtime depth ceiling, unlike #153's build-time cyclic/unknown-anchor checks it sits next to. Added, explicitly distinguishing "runtime cap" from "build-time rejection."

Findings judged out of scope for this PR (filed as follow-ups, matching this PR's own established "single-hop-only correctness bugs are a separate, non-crash bug class" precedent):
- \`merge_sources\`'s own single-hop alias resolution for \`<<\` merge keys — already tracked as **#1318** (confirmed again live by this round, not a new finding).
- \`explicit_tag()\`'s single-hop resolution — already tracked as **#1319** (same).
- \`key_string()\`'s single-hop resolution — was tracked in **#1315**, which had been closed despite its own last comment stating that half was still unfixed; reopened.
- Real \`yq\`/pyyaml reject \`&anchor *alias\` (the only construction that makes a 2+-hop chain possible) as invalid YAML, while succinctly's default lenient parser accepts it (the strict validator already rejects it) — filed as **#1374** for architecture discussion, not a prescribed fix.
- \`YamlValue::Alias\`'s \`target\` field has no privacy boundary, so a future call site can trivially reintroduce this bug class with zero compiler warning (proven twice already, per the two issues above) — filed as **#1375** (structural hardening, a real refactor, out of scope here).

Explicitly declined (not applied): merging \`resolve_alias_chain\`/\`resolve_alias_target_cursor\` into one method via \`resolve_alias_chain = target_cursor().map(value())\`. Multiple agents suggested this, but it reintroduces exactly the "redundant \`.value()\` call on every aliased access" regression #1191's own review already found and fixed for \`resolve_alias_chain\` itself — the two loops are separate specifically to avoid that, not by oversight. Doc comment now explains this explicitly for both directions.

Full local verification (build, full test suite, both clippy legs, rustdoc under \`-D warnings\`, no_std, fmt) green after this round; rebased cleanly onto latest \`main\`.

PR #1320, opened 2026-08-20
fix(jq): check del()'s string-slice refusal before its optional catch-all

### LINKED ISSUE #1303: jq: a ?-marked slice inside an assignment path still falls to 'invalid path component'

### LINKED ISSUE #1312: jq: del() on a ?-marked slice targeting a string wrongly no-ops instead of erroring
**Severity: Low**

## Summary

Found during review of #1310 (fixes #1303, `=`'s equivalent gap). `delete_at_path`'s standalone (terminal, non-mid-chain) `Expr::Slice` arm checks its `optional` catch-all *before* the `String` type-mismatch arm, so `del(.[0:1]?)` on a string silently no-ops where real jq still raises.

```
$ echo '"hello"' | jq 'del(.[0:1]?)'
jq: error (at <stdin>:1): Cannot delete fields from string
$ echo '"hello"' | succinctly jq 'del(.[0:1]?)'
"hello"
```

Also reproduces nested:
```
$ echo '{"a":"hello"}' | jq 'del(.a[0:1]?)'
jq: error (at <stdin>:1): Cannot delete fields from string
$ echo '{"a":"hello"}' | succinctly jq 'del(.a[0:1]?)'
{"a":"hello"}
```

`del()`'s *mid-chain* slice arm (reached when more path follows the slice, e.g. `del(.a[0:1]?.b)`) uses `through_slice` with correctly-threaded optionality and does not share this bug — it's isolated to the standalone/terminal arm's own hand-rolled match.

## Root cause

`delete_at_path`'s standalone `Expr::Slice` arm (`src/jq/eval.rs`, around line 20412-20426 as of #1310) is a separate hand-written match from `through_slice`, with arm order:

```rust
Expr::Slice { start, end } => match root {
    OwnedValue::Array(arr) => { ... }
    OwnedValue::Null => Ok(()),
    _ if optional => Ok(()),                              // <- checked first
    OwnedValue::String(_) => Err(cannot_delete_fields_from("string")),  // <- never reached when optional
    other => Err(cannot_index_with_type(...)),
},
```

Since `optional` here is del()'s real, correctly-threaded per-component flag (not the hardcoded `false` #1303 found in `set_path`'s equivalent call site), whenever it's `true` the `String` arm is shadowed entirely — the write-time-application-style check never even runs. `through_slice` (used by the mid-chain arm, and by `set_path`/`update_path`) orders these correctly: the type-specific arms (`Array`/`String`) come before the generic `_ if optional` catch-all.

## Suggested fix direction

Reorder `delete_at_path`'s standalone `Expr::Slice` arm so `OwnedValue::String(_) => Err(...)` is checked before `_ if optional => Ok(())`, mirroring `through_slice`'s own ordering. Small, isolated change — likely just a match-arm reorder, no signature changes needed.

## Related

Found during review of #1310 (fixes #1303, `=`'s sibling gap for the same "write-time application error vs. navigation failure" distinction). Not fixed by that PR — different function (`delete_at_path`, not `set_path`), out of scope for its own repro.

newhoggy:
## Summary

`del(.[0:1]?)` on a string silently no-op'd instead of raising.

```
$ echo '"hello"' | jq 'del(.[0:1]?)'
jq: error (at <stdin>:1): Cannot delete fields from string
$ echo '"hello"' | succinctly jq 'del(.[0:1]?)'   # before
"hello"
```

Found during review of #1310 (fixes #1303, `=`'s equivalent gap) — filed as #1312.

## Root cause

`delete_at_path`'s standalone (terminal, non-mid-chain) `Expr::Slice` arm ordered its match:

```rust
Expr::Slice { start, end } => match root {
    OwnedValue::Array(arr) => { ... }
    OwnedValue::Null => Ok(()),
    _ if optional => Ok(()),                                          // <- checked first
    OwnedValue::String(_) => Err(cannot_delete_fields_from("string")),  // <- shadowed
    other => Err(cannot_index_with_type(...)),
},
```

Whenever `optional` was `true` (a `?` on the slice), the catch-all fired before the `String` arm ever got a chance — the write-time application refusal never ran. `through_slice` (used by `set_path`/`update_path`, and by `del()`'s own *mid-chain* `Pipe` arm when more path follows the slice) already orders these correctly: type-specific arms before the generic `_ if optional` catch-all. This standalone arm was the one place that didn't match that ordering.

Real jq's inline `?` only suppresses a failure to *reach* a target while collecting a path — not a mismatch in what's being deleted once one is found. This is the same distinction #1303 fixed for `=`'s equivalent gap (`is_write_time_application_error`).

## Fix

Reordered the match arms: `OwnedValue::String(_) => Err(...)` now comes before `_ if optional => Ok(())`, mirroring `through_slice`'s ordering. A genuinely non-sliceable target (e.g. a `?`-marked slice on a boolean) is unaffected — that's still a navigation failure, correctly suppressed either way.

## Verification

- [x] Both repros (bare and nested) byte-match pinned jq 1.7.1
- [x] Regression-checked: no-`?` case, array-slice-target case, and non-sliceable-target-under-`?` case all unaffected
- [x] `cargo fmt --check`
- [x] `omni-dev coverage diff` — 100% patch coverage (17/17 new lines)

## Note

While testing this, found a separate, unrelated, pre-existing divergence (reproduces identically with or without `?`, on both `main` before this PR and this branch): `del(.a[0:1]?.b)` on a string gives `Cannot update string slices` where real jq gives `Cannot index string with string "b"` (a mid-chain slice-then-field shape). Will file as its own follow-up rather than expand this PR's scope.

Fixes #1312

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -20432,8 +20432,15 @@ fn delete_at_path(
             }
             // `null` has no elements to drop, so jq leaves it alone.
             OwnedValue::Null => Ok(()),
-            _ if optional => Ok(()),
+            // Checked *before* the `optional` catch-all below, mirroring
+            // `through_slice`'s own ordering (#1312): a string slice target
+            // is a write-time application refusal jq's inline `?` does not
+            // suppress, not a navigation failure -- `del(.[0:1]?)` on a
+            // string still raises "Cannot delete fields from string" in
+            // real jq, confirmed live. The old order let `optional` shadow
+            // this arm entirely whenever a `?` was present.
             OwnedValue::String(_) => Err(EvalError::cannot_delete_fields_from("string")),
+            _ if optional => Ok(()),
             other => Err(EvalError::cannot_index_with_type(
                 owned_type_name(other),
                 "object",
@@ -40963,6 +40970,47 @@ mod tests {
             QueryResult::Error(e) => {
                 assert_eq!(e.message, "Cannot index number with number");
             }
+        );
+    }
+
+    #[test]
+    fn test_del_slice_on_string_target_ignores_optional_1312() {
+        // #1312: `delete_at_path`'s standalone (terminal) `Expr::Slice` arm
+        // checked its `_ if optional => Ok(())` catch-all *before* the
+        // `OwnedValue::String(_)` arm, so a `?`-marked slice targeting a
+        // string was shadowed into a silent no-op instead of raising. Real
+        // jq's inline `?` only suppresses a failure to *reach* a target, not
+        // a write-time application refusal like this one -- the same
+        // distinction #1303 fixed for `=`'s equivalent gap. A genuinely
+        // non-sliceable target (a `?`-marked slice on a boolean) is a
+        // navigation failure and stays correctly suppressed either way. All
+        // confirmed against real jq 1.7.1.
+        query!(br#""hello""#, r"del(.[0:1]?)",
+            QueryResult::Error(e) => {
+                assert_eq!(e.message, "Cannot delete fields from string");
+            }
+        );
+        query!(br#"{"a":"hello"}"#, r"del(.a[0:1]?)",
+            QueryResult::Error(e) => {
+                assert_eq!(e.message, "Cannot delete fields from string");
+            }
+        );
+        // Unaffected: no `?` at all still errors the same way.
+        query!(br#""hello""#, r"del(.[0:1])",
+            QueryResult::Error(e) => {
+                assert_eq!(e.message, "Cannot delete fields from string");
+            }
+        );
+        // Unaffected: an array slice target still deletes normally.
+        query!(br"[1,2,3]", r"del(.[0:1]?)",
+            QueryResult::Owned(v) => {
+                assert_eq!(v.to_json(), "[2,3]");
+            }
+        );
+        // Unaffected: a genuinely non-sliceable target under `?` stays
+        // suppressed -- this is a navigation failure, not a write-time one.
+        query!(br"true", r"del(.[0:1]?)",
+            QueryResult::Owned(OwnedValue::Bool(true)) => {}
         );
     }
 


newhoggy:
Extensive review round (10 finder angles), no code changes needed — findings were either confirmed non-issues (dead/unreachable code, already-correct edge cases), overlap with an already-filed follow-up, or a new corroborated finding filed separately:

**Confirmed non-issues** (verified live, no action needed):
- The reorder is behavior-correct and exhaustive for all 8 `OwnedValue` variants — only the intended `String`+`optional=true` case changes.
- No caller regression; `eval_generic.rs` has no `del()` equivalent to keep in sync.
- yq mode is structurally shielded from the touched arm (upstream gates intercept every yq-mode path first) — confirmed live against real yq v4.53.3.
- `update_path`/`through_slice` (the `|=`/`-=`/`*=`/`+=` side) already had this ordering correct from the start — no equivalent gap there.
- `delete_at_path`'s own `optional`/`here` self-threading was checked exhaustively across every recursive entry point — correct throughout.

**Already covered by #1321** (filed from this PR's own testing before review started): the mid-chain case (`del(.a[0:1][0]?)`, more path after the slice) reports the wrong error and doesn't respect `?` — added the additional repro review surfaced to that issue rather than filing a duplicate.

**New follow-up filed**: #1322 — a dormant instance of this same bug class in the comma-grouped/multi-path `del()` walker (`delete_expr_array_paths`), confirmed dead code today (an upstream step strips the `?` info before it gets there) by three independent review passes, matching this repo's known "optional never forced true in nested calls" family (#928/#1003/#1034).

**Incident, resolved**: a review subagent's live-testing left 3 lines of uncommitted `eprintln!` debug output in the worktree mid-review (investigating the #1322 hypothesis above). Caught before commit, discarded via `git checkout --`, rebuilt and reverified clean — never reached this branch's history.

Full suite/clippy/fmt/coverage all still green after the cleanup; no further changes needed on this PR.

Issue #1199, opened 2026-08-17
jq: a NumberLiteral operand's source spelling is lost in a binary-op type-mismatch error message


newhoggy:
**Severity: Low**

Found during code review of #1175/PR #1196 (`arith_mul`'s null-multiply fix), which newly exposed this on null-involving pairings — but the underlying bug predates that PR and is broader than multiplication.

## Root cause

`arith_mul`'s final catch-all arm calls `left.into_plain_number()`/`right.into_plain_number()` *before* falling through to `EvalError::binary_op(&a, &b, ...)` on a type mismatch. `into_plain_number()` collapses a `NumberLiteral` operand (which preserves the value's original source spelling, e.g. scientific notation) down to a plain `Int`/`Float`, so the error message it builds echoes the canonically-reformatted value instead of what the user actually wrote.

## Repro (live-verified against real jq 1.7.1)

```
$ jq -cn '1e10 * null'
jq: error (at <unknown>): number (1E+10) and null (null) cannot be multiplied
```

succinctly (current `main`, after #1175/PR #1196):

```
$ succinctly jq -cn '1e10 * null'
error: number (10000000000) and null (null) cannot be multiplied
```

Real jq's error text preserves `1E+10`; succinctly's shows the reformatted `10000000000`. This is reachable pre-#1175 too, via any other type mismatch that reaches the same catch-all, e.g. `1e10 * {}` — #1175 is what newly makes it reachable via a *null* operand specifically, since null-involving pairings now also fall through to this same catch-all.

## Scope note

This is likely one instance of a broader class: any binary-op error-message builder that reformats an operand before embedding it loses literal fidelity for that error message specifically (values *returned* to the user already preserve literal spelling correctly elsewhere in the codebase — this is purely an error-*message* text issue). Worth checking whether `arith_add`/`arith_sub`/`arith_div`/`arith_mod`'s own type-mismatch arms have the same ordering issue before fixing just `arith_mul`'s.

## Severity

Low: cosmetic-only divergence in error message text, not a behavioral difference (exit code and error *shape* both already match the oracle).

newhoggy:
🔧 Claimed by agent `aiden` — branch `issue-1199-numberliteral-spelling-in-binop-errors`

newhoggy:
Closed by #1229 (merged) — every arith_* function now preserves a NumberLiteral operand's source spelling in its binary-op error messages, matching real jq exactly.

PR #1324, opened 2026-08-20
fix(jq): make integer modulo by zero error in yq mode

### LINKED ISSUE #1231: yq: modulo by zero should error like real yq, not silently return null
**Severity: Low**

## Summary

Found during code review of #1199's PR — pre-existing, unrelated to and
unaffected by that fix either way (confirmed identical before and after,
on unmodified `main`).

`YqSemantics::DIV_BY_ZERO_IS_INFINITY` currently gates both `/` (division)
and `%` (modulo) by zero identically — succinctly's `arith_mod` treats a
zero divisor as a success case (NaN, serialized as `null`) whenever this
flag is true, mirroring `arith_div`'s own division-by-zero-returns-Infinity
behavior. But real yq's `/` and `%` diverge here: division by zero really
does return Infinity (successfully), but modulo by zero *errors*.

```
$ echo null | yq -o=json '5 / 0'
Infinity                              # (or similar; division-by-zero succeeds)

$ echo null | yq -o=json '5 % 0'
Error: cannot modulo by 0              # modulo-by-zero errors
```

succinctly (current `main`, unaffected by #1199 either way):

```
$ echo null | succinctly yq -o=json '5 % 0'
null                                   # wrong: silently succeeds instead of erroring
```

## Why this needs its own fix, not folded into #1199

#1199 is scoped to error-*message* fidelity (preserving a `NumberLiteral`
operand's spelling in an error that already fires) — this issue is about
whether an error fires *at all* for yq-mode modulo-by-zero, a behavioral
gap in `EvalSemantics`'s existing `DIV_BY_ZERO_IS_INFINITY` flag being too
coarse (it conflates two operators real yq treats differently). Fixing it
needs either a new, more specific semantics flag (e.g.
`MOD_BY_ZERO_IS_INFINITY`, distinct from `DIV_BY_ZERO_IS_INFINITY`) or
`arith_mod` no longer deferring to the same flag `arith_div` uses.

## Scope note

Low severity — `% 0` is an unusual thing to write intentionally, and the
current silent-`null` behavior is at least not a crash. Once #1199 lands
(preserving spelling in whichever error message ends up firing here), a
fix for this issue would get spelling-preserving behavior for free through
the same `divisor_is_zero` call site `arith_mod` already has.

newhoggy:
## Summary

`5 % 0` silently returned `null` (a serialized NaN) in yq mode instead of erroring.

```
$ echo null | yq '5 % 0'
Error: cannot modulo by 0
$ echo null | succinctly yq '5 % 0'   # before
null
```

Real yq's Infinity-on-zero-divisor rule for division doesn't extend to an integer modulo — confirmed live against yq v4.53.3: `5 / 0` succeeds as `+Inf`, but `5 % 0` errors.

## Root cause

`arith_mod`'s `Int`-`Int` arm deferred to `DIV_BY_ZERO_IS_INFINITY`, the same `EvalSemantics` flag `arith_div` uses to decide whether a zero divisor succeeds (yq) or errors (jq). That's correct for division, but wrong for this one case: real yq's own modulo-by-zero behavior doesn't follow the same rule its division does.

A *float*-involving modulo by zero was already correct — `mod_floats` (gated by the separate `MOD_TRUNCATES_FLOATS` flag) already returns NaN for `5.5 % 0`, `5 % 0.0`, and `5.5 % 0.0` in yq mode, matching real yq exactly. Only the pure-integer case needed a fix.

## Fix

No new `EvalSemantics` flag needed — jq and yq behave *identically* for integer modulo by zero (both error), just with different wording. The `Int`-`Int` arm is now unconditional on `DIV_BY_ZERO_IS_INFINITY`; it branches only on `S::TAG` to pick which message to raise: jq's existing `divisor_is_zero` (unchanged), or a new `yq_modulo_by_zero()` constructor producing yq's own terser, operand-value-free wording (`"cannot modulo by 0"`).

## Verification

- [x] `5 % 0` in yq mode now byte-matches real yq's exact message (`"cannot modulo by 0"`)
- [x] Division by zero unaffected — still succeeds as `+Inf` in yq mode
- [x] All three float-involving modulo-by-zero combinations (`5.5 % 0`, `5 % 0.0`, `5.5 % 0.0`) unaffected — still correctly return NaN
- [x] jq mode unaffected — `5 % 0` still errors with jq's own existing wording, byte-identical to real jq
- [x] No existing yq-golden fixture covers this case (checked) — no drift risk
- [x] `cargo fmt --check`
- [x] `omni-dev coverage diff` — 100% patch coverage (21/21 new lines)

Fixes #1231

### DIFF
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -893,6 +893,15 @@ impl EvalError {
         )
     }
 
+    /// `cannot modulo by 0` — real yq's own wording for an integer modulo
+    /// by zero (#1231), confirmed live against yq v4.53.3. Unlike jq's
+    /// [`Self::divisor_is_zero`], this is a fixed sentence with no embedded
+    /// operand values, and applies only to the integer case: a
+    /// float-involving modulo by zero returns NaN in yq, not an error.
+    pub fn yq_modulo_by_zero() -> Self {
+        Self::new("cannot modulo by 0")
+    }
+
     /// `<a> and <b> cannot have their containment checked`.
     pub fn containment_check(left: &OwnedValue, right: &OwnedValue) -> Self {
         Self::pair(left, right, "cannot have their containment checked")
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -2518,11 +2518,21 @@ fn arith_mod<S: EvalSemantics>(
     match (left.number_repr(), right.number_repr()) {
         (Some(NumberRepr::Int(a)), Some(NumberRepr::Int(b))) => {
             if b == 0 {
-                if S::DIV_BY_ZERO_IS_INFINITY {
-                    // yq behavior: return NaN (will be serialized as null)
-                    Ok(OwnedValue::Float(f64::NAN))
+                // Integer modulo by zero errors in both jq and yq --
+                // unlike division, yq's Infinity-on-zero-divisor rule
+                // (`DIV_BY_ZERO_IS_INFINITY`) does not extend to this
+                // case (#1231, confirmed live against yq v4.53.3: `5 %
+                // 0` errors "cannot modulo by 0", where `5 / 0` succeeds
+                // as `+Inf`). Only a *float*-involving modulo by zero
+                // returns NaN in yq (`mod_floats` below, gated by
+                // `MOD_TRUNCATES_FLOATS`) -- so this arm is unconditional
+                // on `S::DIV_BY_ZERO_IS_INFINITY`, unlike every other
+                // zero-divisor check in this file. yq's own wording is a
+                // fixed, terser sentence with no embedded operand values,
+                // unlike jq's `divisor_is_zero`.
+                if S::TAG == EvalTag::Yq {
+                    Err(EvalError::yq_modulo_by_zero())
                 } else {
-                    // jq behavior: error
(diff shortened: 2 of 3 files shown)


newhoggy:
Four review passes, no correctness issues found. Two things acted on:

1. **Fixed a stale comment** in `tests/yq_cli_tests.rs` that still described #1231 as unresolved future work, implying integer modulo-by-zero still silently succeeds in succinctly's yq mode — misleading now that this PR fixes it.
2. **Added CLI-level integration test** (`test_1231_yq_integer_modulo_by_zero_errors_but_division_does_not`) alongside the sibling #1199 tests in the same file, closing a coverage-layer gap the review flagged (the fix's own unit test already covered this at the `eval.rs` layer, but the file's established convention pairs unit + CLI coverage for this class of divergence).

One judgment-call finding, not acted on: the new arm dispatches on `S::TAG == EvalTag::Yq` directly rather than introducing a dedicated `EvalSemantics` const (the pattern this function's sibling checks use). Two independent review passes judged this defensible — it's consistent with 34 other single-call-site `S::TAG` checks elsewhere in the file, and a new const would arguably be over-engineering a genuinely simple, single-site divergence. Noting it here rather than acting on it, since a real change would trade one file's local convention for another with no clear win.

Full suite/clippy/fmt/coverage all still green, patch coverage 100% (21/21 new lines).

Issue #1162, opened 2026-08-17
yq: del()/delpaths() have the same slice-through-container gap #1142 fixed for assignment


newhoggy:
**Severity: Medium**

## Summary

Found during code review of #1142's fix (PR #1158). #1142 fixed `=`/`|=`/`+=`/`-=`/
`*=`'s slice-write no-op for a chained array/string target (`set_path`/`update_path`,
both routing through the shared `through_slice` primitive). `del()`/`delpaths()`
share that same `through_slice` primitive but were deliberately left untouched by
#1142 (matching this repo's established precedent from #1116, where `del()`'s
chained-scalar-slice behavior was already known to need its *own*, more complex
design — see #1116's own description: `del()`'s chained case deletes the *entire
parent key*, not a no-op the way `=`/`|=`/`+=` are).

This issue is to make the **container**-target counterpart of that same gap
trackable — real yq's `del()` on a slice reached through a chained array/string
deletes the whole parent field, not the sliced range and not a no-op.

## Repro (verified against real yq v4.53.3)

```
$ printf 'a: [1,2,3,4,5]\n' | yq -o=json 'del(.a[0:2])'
{}                                     # entire "a" field deleted, any bounds

$ printf 'a: [1,2,3,4,5]\n' | succinctly yq -o=json 'del(.a[0:2])'
{"a": [3, 4, 5]}                       # wrong: partial range deletion instead
```

Also reproduces for:
- A comma-grouped multi-path `del()` (`del(.a[0:2], .b[2:4])`) — real yq deletes
  both whole fields; succinctly partial-deletes each independently.
- A **bare-root** array/string slice `del()` (`del(.[1:3])` on a top-level array) —
  real yq no-ops (unlike the chained case, which deletes the parent); succinctly
  still drains the range. This is the container counterpart of #1101's own
  bare-root-scalar `del()` no-op, similarly left un-widened.
- `delpaths([[{"start":s,"end":e}]])` — a slice-descriptor path component is
  spliced through successfully by succinctly; real yq rejects it outright at any
  position (`DELPATHS: expected either a !!str or !!int in the path, found !!map
  instead`).

## Why this is separate from #1142

#1142's own PR is scoped to compound-assignment operators; `del()`/`delpaths()` are
a structurally different walker family (`delete_at_path`, `delete_expr_paths_at`,
`builtin_del`'s own no-op gate, and the separate `delpaths()`/`setpath`-family
[...]

PR #1328, opened 2026-08-20
fix(jq): reject all non-str/int delpaths path components in yq mode

### LINKED ISSUE #1220: yq: delpaths() rejects float/bool/array path components too, not just the slice-descriptor map
**Severity: Low**

## Summary

Found while implementing #1162's own narrower scope (rejecting a
slice-descriptor/map path component in `delpaths()` for yq mode). Live
oracle testing while pinning that fix's error wording turned up that real
yq's `delpaths()` is actually much stricter than #1162 assumed: it rejects
essentially *any* path component that isn't literally a string or a plain
integer — floats, booleans, and arrays all hit the identical error, not just
the object-shaped slice descriptor.

```
$ printf 'a: 1\n' | yq -o=json 'delpaths([[true]])'
Error: DELPATHS: expected either a !!str or !!int in the path, found !!bool instead

$ printf 'a: 1\n' | yq -o=json 'delpaths([[[1,2]]])'
Error: DELPATHS: expected either a !!str or !!int in the path, found !!seq instead

$ printf -- '- 1\n- 2\n- 3\n' | yq -o=json 'delpaths([[1.0]])'
Error: DELPATHS: expected either a !!str or !!int in the path, found !!float instead
```

Plain integers and strings still work fine (`delpaths([[1]])`,
`delpaths([["a"]])`) — this isn't a parse-level rejection of the whole call,
just every individual path component.

`succinctly`'s current `delpaths()` (both before and after #1162) accepts
all of these in both jq and yq mode — jq mode is correct to (real jq is far
more permissive: `resolve_read_index` truncates a float toward zero and
treats it as a valid index), but yq mode currently matches jq's own
permissiveness here instead of real yq's much stricter behavior.

## Why this is separate from #1162

#1162's own repro and stated scope is specifically the slice-descriptor
(map) shape — it doesn't mention floats/bools/arrays at all, and widening
the fix to reject those too would be a materially broader behavior change
(e.g. does succinctly's own float-as-index support in `delpaths()` have
existing test coverage relying on it working in yq mode today? needs
checking before any blanket rejection lands).

## Suggested fix direction

Once #1162 lands, decide whether to fold this into the same rejection check
(swap "is this an Object" for "is this anything other than String/Int") or
keep them as two separate checks — either way, verify there's no existing
yq-mode test pinning today's more-permissive float/bool/array acceptance
before changing it, since removing a currently-passing behavior needs its
own regression-guard update, not just a new error path.

newhoggy:
## Summary

Real yq's `delpaths()` only accepts `!!str`/`!!int` components in a path array. Every other YAML type errors — `!!null`, `!!bool`, `!!float` (including whole-number-valued ones like `1.0`), and `!!seq` all raise `DELPATHS: expected either a !!str or !!int in the path, found <tag> instead`. Succinctly previously only rejected the `!!map` case (a slice-descriptor object, #1162's narrower scope) and silently accepted everything else, following jq's own more permissive model instead.

## Root cause

`builtin_delpaths`'s yq-mode type check (`src/jq/eval.rs`) only matched `OwnedValue::Object` (the slice-descriptor shape `path(.[a:b])` produces). A float, bool, null, or array path component fell through unchecked and was evaluated the same way jq mode evaluates it (a float truncated toward zero as an array index, etc.), rather than erroring the way real yq does.

## Fix

- Generalized `EvalError::delpaths_rejects_slice_descriptor()` into `EvalError::delpaths_rejects_type(tag: &str)`, parameterized by the YAML type tag of the actual offending component.
- Added `delpaths_component_type_tag()`, mapping an `OwnedValue` to its YAML type tag (`!!null`/`!!bool`/`!!int`/`!!float`/`!!str`/`!!seq`/`!!map`), distinguishing `Int`/`Float` where `OwnedValue::type_name()` collapses both to `"number"`.
- Rewrote the detection loop inside `builtin_delpaths` to scan every path's every component for the first one that isn't `String`/`Int`, reporting its type tag — matching real yq's first-offender-in-path/component-order precedence (confirmed live).
- The two arms of `delpaths_component_type_tag` for `!!int`/`!!str` are structurally unreachable from the function's sole call site (which already filters those types out before calling it) — kept only for match exhaustiveness, documented inline rather than chased for coverage.

## Verification

- Confirmed live against real yq v4.53.3 for every rejected type (`!!null`, `!!bool`, `!!float` including `1.0`, `!!seq`), the still-accepted `!!str`/`!!int` (including negative int), first-offender precedence across and within paths, and jq-mode non-regression.
- New tests: `test_1220_delpaths_rejects_bool_seq_float_null_components`, `test_1220_delpaths_still_accepts_str_and_int`, `test_1220_delpaths_reports_first_offending_component`, `test_1220_delpaths_jq_mode_unaffected`.
- Removed the obsolete `test_1220_delpaths_accepts_float_component_characterize_preexisting_bug`, which asserted the old buggy behavior this PR fixes.
- Patch coverage 91.3% (21/23 new lines); the 2 uncovered lines are the documented-unreachable match arms above.

Fixes #1220

### DIFF
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -518,20 +518,24 @@ impl EvalError {
         Self::new(format!("Path must be specified as array, not {type_name}"))
     }
 
-    /// `DELPATHS: expected either a !!str or !!int in the path, found !!map instead`.
+    /// `DELPATHS: expected either a !!str or !!int in the path, found <tag> instead`.
     ///
-    /// yq-mode-only (#1162): unlike real jq, which accepts a slice-descriptor
-    /// path component (`{"start":s,"end":e}`, `path(.[a:b])`'s own output
-    /// shape) at any position in a `delpaths()` path and splices through the
-    /// named sub-range, real yq rejects one outright — verified live against
-    /// yq v4.53.3 with this exact wording, at both a top-level and a nested
-    /// position. Real yq's `delpaths()` is actually stricter still (also
-    /// rejects a float/bool/array component with the same message, substituting
-    /// its own type's tag), but this constructor covers only the slice-
-    /// descriptor shape #1162 itself scoped — the broader rejection is
-    /// tracked separately as #1220.
-    pub fn delpaths_rejects_slice_descriptor() -> Self {
-        Self::new("DELPATHS: expected either a !!str or !!int in the path, found !!map instead")
+    /// yq-mode-only. Real yq's `delpaths()` accepts only `!!str`/`!!int`
+    /// path components — every other YAML type errors here, `tag` naming
+    /// whichever one was actually found. Originally #1162 (a slice-
+    /// descriptor path component, `{"start":s,"end":e}`, `path(.[a:b])`'s
+    /// own output shape — unlike real jq, which accepts one at any position
+    /// and splices through the named sub-range) scoped this to just
+    /// `!!map`; #1220 found the real rule is this much broader one —
+    /// `!!null`/`!!bool`/`!!float`/`!!seq` all hit the identical message,
+    /// substituting their own tag. Verified live against yq v4.53.3 for
+    /// every variant, at both a top-level and a nested position, including
+    /// that a plain `!!int` is accepted but a whole-number-valued `!!float`
+    /// (`1.0`) is not.
+    pub fn delpaths_rejects_type(tag: &str) -> Self {
+        Self::new(format!(
+            "DELPATHS: expected either a !!str or !!int in the path, found {tag} instead"
+        ))
     }
 
     /// `Cannot delete fields from <type>`.
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -23659,31 +23659,49 @@ fn builtin_delpaths<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     // (`1.0`) is rejected too -- real yq's check is on the YAML type, not
     // the numeric value, confirmed live. Folded into this same `retain`
     // pass rather than a separate loop before it: both checks need to
-    // visit every path's every component before any deletion runs. Gated
-    // on `keeps` (the same per-path NaN survival this `retain` already
-    // computes), not checked unconditionally, so a path that's *also*
-    // NaN-headed doesn't surface the type error for a path already headed
-    // for silent removal. `rejected_component_tag` keeps only the first
-    // offender found (path order, then component order within a path) --
-    // confirmed live that real yq reports whichever type-mismatched
-    // component it reaches first when several paths/components each
-    // qualify.
+    // visit every path's every component before any deletion runs.
+    //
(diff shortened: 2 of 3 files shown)


newhoggy:
## Code review response

Ran the multi-agent review (8 finder angles + synthesis + a follow-up live-verification pass). Summary of what was found and done:

**Fixed:**
- **NaN component masking a bad-typed sibling** — the widened type check was gated on the same `keeps` (NaN-drop) flag as the pre-existing NaN-drop logic, so `delpaths([[nan,true]])` in yq mode silently succeeded instead of erroring on `true`. Live-confirmed on this branch's own build before the fix. Decoupled the two: the type check now runs unconditionally, with an explicit exemption so a bare NaN component is never itself reported as the offender — `delpaths([[nan]])` alone still silently drops as before, preserving that pre-#1220 behavior exactly.
- **Test ambiguity in "first offending component"** — `compare_values` ranks Bool below Number, so the original test case (`[[true],[1.5]]`) couldn't distinguish "first in argument order" from "first in sort order." Added a disambiguating case (`[[1.5],[true]]`) confirming the implementation checks argument order — verified live against real yq v4.53.3.
- **`delpaths_component_type_tag` duplicates `yaml_type_tag`** — both independently map the same 7 YAML-tag categories, ~1400 lines apart, with nothing keeping them in sync (the exact #106 "duplicated predicates diverge silently" pattern). They operate on different value types (`OwnedValue` vs `StandardJson`) so a structural merge is a larger refactor than this issue's scope; instead added `test_1220_delpaths_and_tag_type_tags_agree`, which pins both functions to the same tag string for one value per category, so a future edit to either one that isn't mirrored in the other now fails a test instead of silently diverging.

**Investigated and resolved via a decision, not a code change** — a maintainer comment on #1220 flagged that #1088 (currently open as PR #1327) will make yq-mode `path()` start emitting float array indices, which would break the `path()` → `delpaths()` round-trip once it lands. Verified real yq has no `path`/`getpath`/`setpath` builtin at all (no oracle either way), and decided `delpaths()` stays strict rather than carving out an exemption — recorded in a code comment at the check site and as a reply on #1220.

**Filed as a follow-up, not fixed here** (pre-existing, out of scope for this issue, needs its own oracle check): [#1330](https://github.com/rust-works/succinctly/issues/1330) — the DELPATHS type-rejection (both #1162's original and #1220's widened version) doesn't respect `builtin_delpaths`'s own `optional` flag, unlike every other error path in that function.

**Not actionable / scope note only:** `del()` doesn't share the new type check (`delpaths()`-specific per the issue's own title; `del()` navigates via real path expressions, not an arbitrary caller-supplied path array, so there's no comparable gap).

All fixes re-verified: full test suite (0 failures), both required clippy invocations clean, `cargo fmt --check` clean, patch coverage now 100% (33/33 new lines).

Issue #1332, opened 2026-08-20
yq: needs_path_context/eval_pipe_with_path_context_internal has no Expr::Object handling either, same gap as #1302's Expr::Array


newhoggy:
**Severity: Low**

## Summary

Found while implementing #1302 (`[key]`/`[parent]`/`[file_index]` losing path context). The identical bug exists for object-literal construction: `needs_path_context` has no `Expr::Object` arm either, and even if it did, `eval_pipe_with_path_context_internal`'s object-construction arm (shared with `Array`/`Literal` before #1302, now `Array` split out) still builds the object via the plain, context-less evaluator.

## Repro (confirmed against real yq v4.53.3)

```
$ printf 'a:\n  b: 1\n' | yq -o=json '.a | {"x": key}'
{"x":"a"}

$ printf 'a:\n  b: 1\n' | succinctly yq -o=json '.a | {"x": key}'
{"x":null}
```

## Why this wasn't folded into #1302

`Expr::Object(Vec<ObjectEntry>)` is materially more complex than `Expr::Array(Box<Expr>)`: multiple key-value entries, each potentially a generator (`{(gen): val}`), meaning the fix needs to thread path context through a cartesian-product-style entry evaluator rather than one recursive call on a single inner expression. #1302's fix (`eval.rs`, `needs_path_context` + a new `Expr::Array(inner) if needs_path_context(inner)` arm in `eval_pipe_with_path_context_internal`) is a template for the mechanism, but the object case needs its own design pass over how `ObjectEntry`'s key/value generators interact with path-context propagation — not a one-line mirror of the array fix.

## Suggested fix direction

Not investigated beyond confirming the repro. Look at how the existing object-construction code (`eval_object_construction` or equivalent, and whatever entry-generator loop it uses) collects multi-output key/value pairs today, and figure out how to route an entry's key or value expression through `eval_pipe_with_path_context_internal` when it individually needs path context — likely entry-by-entry rather than one shot for the whole object, since different entries' generators are independent.

newhoggy:
## Triage (2026-09-04) — route every entry through the context evaluator via one injected closure

Re-verified live: `.a | {"x": key}` → yq `{"x":"a"}`, succinctly `{"x":null}`; also `{(key): 1}` errors outright, `{"x": parent}` → `{}`, `{"x": (key+"!")}` → `"!"`, `[{"x": key}]` → `[{"x":null}]`.

[...]

Issue #1280, opened 2026-08-19
jq: path-context evaluation turns a legitimately-empty builtin result into null instead of empty


newhoggy:
## Summary

Discovered while verifying #1278's (fix for #1045) test suite for regressions.
`eval_owned_expr_ctrl_full` (`src/jq/eval.rs`) — used by `eval_pipe_with_path_context_internal`'s
generic `Expr::Builtin`/`Expr::Object`/`Expr::Array`/`Expr::Literal`/catch-all arms (the
machinery behind `key`/`parent`/`path`/`file_index` tracking) — collapses *any*
`QueryResult::None` from evaluating a builtin call to `Ok(OwnedValue::Null)`:

```rust
QueryResult::None => Ok((OwnedValue::Null, None)),
```

Real jq's own equivalent semantics for a `?`-swallowed expression that produces no value
is *empty output*, not `null`:

```
$ jq -cn '1 | .a?'
(exits 0, no output — not null)
$ jq -cn '("s" | has("x"))?'
(exits 0, no output — not null)
```

## Confirmed repro (pre-existing, reproduces on `main` before #1278)

```
$ echo '{"a":{"b":"str"}}' | succinctly jq -c '.a.b | (has("x"))?, key'
null
"b"
```

Expected (matching real jq's own `.a?`/`has(...)?` empty-not-null contract, applied to
this succinctly-only `key` extension): just `"b"` — the `has(...)` branch should
contribute zero outputs to the comma, not `null`.

## Why this wasn't caught before

Before #1278, `has`'s own argument-`None` case (and several similar builtins') always
turned into a hard `QueryResult::Error`, not a `QueryResult::None` — so it was caught
instead by a *different*, correct code path (`Err(EvalEscape::Error(_)) if optional =>
QueryResult::None` at the `eval_pipe_with_path_context_internal` call site), which
correctly produces zero output. #1278 legitimately changed several builtins to propagate
a zero-output argument as their own `QueryResult::None` (matching real jq) instead of an
error — which routes through `eval_owned_expr_ctrl_full`'s pre-existing (and already
buggy, just previously less-reachable) `None -> Null` collapse instead.

The bug itself, though, predates #1278 entirely: `has`'s own *primitive-type* mismatch
under `optional` (`_ if optional => QueryResult::None`, unrelated to its argument) already
reached this same collapse before #1278 — confirmed by reproducing the repro above against
the pre-#1278 baseline directly.

## Suggested approach

[...]

PR #1333, opened 2026-08-20
fix(jq): route array-literal construction through path context

### LINKED ISSUE #1302: jq: needs_path_context has no Expr::Array arm, losing path tracking for [key]/[parent]/[file_index]
**Severity: Medium**

## Summary

`needs_path_context` (`src/jq/eval.rs:466-533`) decides whether a pipe stage
needs the full path-tracking evaluator rather than the plain-value one. It
recurses into `Expr::Comma` (line 507) but has **no arm for `Expr::Array`**,
falling to the `_ => false` catch-all (line 532).

## Repro (confirmed against current `main`, pre-existing — not a regression
from #1168's PR #1296)

```
$ printf 'a:\n  b: 1\n' | succinctly yq '.a | (key,key)'
a
a
$ printf 'a:\n  b: 1\n' | succinctly yq '.a | [key]'
- null
```

`(key,key)` correctly resolves `key` (the current node's key in its parent)
twice via the path-tracking evaluator. `[key]` — semantically equivalent,
just array-wrapped instead of comma-joined — silently loses that path
context and reports `null`.

## Root cause

`Expr::Array(inner)` never reaches `needs_path_context`'s recursive check the
way `Expr::Comma`'s siblings do, so a pipe stage like `.a | [key]` is judged
"doesn't need path context" even though its wrapped expression (`key`) does.

## Why this is separate from #1168 (PR #1296)

#1168 added native `eval_single` arms for `Expr::Array`/`Expr::Comma` in
`eval_generic.rs` specifically to preserve duplicate YAML mapping keys
through `[...]`/`,` wrapping — a different evaluator (`eval_generic.rs`'s
cursor-native fast path) and a different bug class (path-context tracking in
`eval.rs`'s full evaluator) than this one. This bug reproduces identically on
`main` before that PR, confirmed via checkout/rebuild — #1168 doesn't cause
it, though it's the PR whose review surfaced it (the same "does the array
wrapper propagate path/context requirements symmetrically with comma"
question came up for both evaluators).

## Suggested fix direction

Add `Expr::Array(inner) => needs_path_context(inner)` to `needs_path_context`
next to its existing `Expr::Comma` arm — the same recursive-into-inner-expr
shape, since an array is still just "collect this inner expression's
outputs," and whatever the inner expression needs, the wrapper needs too.
Should be a small, single-function fix, but verify against the pinned oracle
for `[key]`/`[parent]`/`[parent(n)]`/`[file_index]`/`[$__loc__]` before
merging, and check for other places in `eval.rs` that make the same
Comma-but-not-Array distinction (a quick grep for `Expr::Comma` arms without
a matching `Expr::Array` arm nearby would find them).

Refs #1168, PR #1296

newhoggy:
## Summary

`needs_path_context` (`src/jq/eval.rs`) decides whether a pipe stage needs the full path-tracking evaluator rather than the plain-value one. It recurses into `Expr::Comma` but had no arm for `Expr::Array`, falling to the `_ => false` catch-all — so `.a | [key]`, semantically equivalent to `.a | (key,key)` just array-wrapped, silently lost path context and reported `null` instead of resolving `key`.

## Root cause

Two-part gap:
1. `needs_path_context` had no `Expr::Array` arm, so a pipe containing `[key]`/`[parent]`/`[file_index]` was routed to the plain evaluator instead of the path-context one.
2. Adding that arm alone isn't sufficient: `eval_pipe_with_path_context_internal`'s existing array-construction arm (shared with `Object`/`Literal`) builds the array via `eval_owned_expr_opt` — the plain, context-less evaluator — regardless of whether the routing decision said the pipe needs path context. `Builtin::Key`'s own plain-evaluator fallback always returns `Null` ("Key requires path context which is handled in eval_pipe_with_context ... return null"), so the array construction itself needed a second fix.

## Fix

- Added `Expr::Array(inner) => needs_path_context(inner)` to `needs_path_context`, mirroring the existing `Comma` arm.
- Split `Expr::Array` out of the shared `Expr::Object(_) | Expr::Array(_) | Expr::Literal(_)` arm in `eval_pipe_with_path_context_internal`: a new `Expr::Array(inner) if needs_path_context(inner)` arm evaluates `inner` through the same path-context evaluator (`eval_pipe_with_path_context_internal` recursively) and collects every output into the array — mirroring how `eval_array_construction` collects the plain evaluator's outputs for the ordinary case. The ordinary (non-path-context) array case, and `Object`/`Literal`, stay on the original unmodified code path.

## Verification

- Confirmed live against real yq v4.53.3 for `[key]` and `[parent]` (`key`/`parent`/`file_index` have no real-jq equivalent — succinctly extensions — so jq-mode testing here pins internal consistency, not an oracle).
- New tests: `test_array_wrapped_path_context_builtins_1302` (basic cases, multiple builtins in one array, `rest`-after-array path reset, plain-array non-regression) and `test_array_wrapped_path_context_builtin_control_flow_1302` (exercises every non-`Owned`/`ManyOwned` branch of the new match: zero-output, bare and `Partial`-wrapped error/break/halt — array construction is atomic, so each must abandon the whole array).
- Regression-checked against existing `_554_`/`_715_`/`_1280_` path-context and array-construction tests (42 tests, all pass).
- Patch coverage 96.67% (29/30 new lines); the one uncovered line is a documented-unreachable `unreachable!()` arm matching an existing sibling arm's identical, already-accepted invariant a few hundred lines above.

`Expr::Object` has the identical gap (`{"x": key}` also stubs to `null`, confirmed live) but needs its own design pass over multi-entry/generator-key semantics rather than a one-line mirror of this fix — filed as #1332.

Fixes #1302

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -505,6 +505,15 @@ pub(crate) fn needs_path_context(expr: &Expr) -> bool {
         Expr::Paren(inner) => needs_path_context(inner),
         Expr::Optional(inner) => needs_path_context(inner),
         Expr::Comma(exprs) => exprs.iter().any(needs_path_context),
+        // `[key]`/`[parent]`/`[file_index]` (#1302): an array literal is
+        // still just "collect this inner expression's outputs", the same
+        // reasoning as `Comma` above -- whatever the wrapped expression
+        // needs, the `[...]` wrapper needs too. `Expr::Object` isn't
+        // included here yet: it has the identical gap (`{"x": key}` also
+        // stubs to `null`, confirmed live), but its multi-entry/generator-key
+        // shape needs more design work than a single recursive call, so it's
+        // tracked separately rather than folded into this fix.
+        Expr::Array(inner) => needs_path_context(inner),
         Expr::IndexExpr { target, key } => needs_path_context(target) || needs_path_context(key),
         Expr::SliceExpr { target, start, end } => {
             needs_path_context(target)
@@ -18342,6 +18351,60 @@ fn eval_pipe_with_path_context_internal<'a, W: Clone + AsRef<[u64]>, S: EvalSema
                 // (#791).
                 Err(EvalEscape::Error(_)) if optional => QueryResult::None,
                 Err(escape) => escape.into(),
+            }
+        }
+        // `[key]`/`[parent]`/`[file_index]` (#1302): the generic
+        // `Expr::Array` arm below builds the array via `eval_owned_expr_opt`
+        // -- the plain, context-less evaluator -- so a `key`/`parent`/
+        // `file_index` nested inside the literal would still stub to
+        // `null`/`0` even once `needs_path_context` correctly routes the
+        // enclosing pipe here. Route the array's *own* inner expression
+        // through this same path-context evaluator instead, collecting its
+        // outputs the same way `eval_array_construction` collects the plain
+        // evaluator's outputs elsewhere, then continue exactly as the
+        // generic arm below does (reset path/root for `rest`).
+        Expr::Array(inner) if needs_path_context(inner) => {
+            let inner_result = eval_pipe_with_path_context_internal::<W, S>(
+                core::slice::from_ref(inner),
+                value,
+                root,
+                file_origin,
+                current_path,
+                optional,
+            );
+            let items: Vec<OwnedValue> = match inner_result {
+                QueryResult::Owned(v) => vec![v],
+                QueryResult::ManyOwned(vs) => vs,
+                QueryResult::None => vec![],
+                QueryResult::Error(e) => return QueryResult::Error(e),
+                QueryResult::Break(label) => return QueryResult::Break(label),
+                QueryResult::Halt(code) => return QueryResult::Halt(code),
+                // Array construction is atomic (`eval_array_construction`'s
+                // identical reasoning): a control signal after some output
+                // still abandons the whole array rather than keeping a
+                // partial one.
+                QueryResult::Partial(_, Control::Error(e)) => return QueryResult::Error(e),
+                QueryResult::Partial(_, Control::Break(label)) => return QueryResult::Break(label),
+                QueryResult::Partial(_, Control::Halt(code)) => return QueryResult::Halt(code),
+                QueryResult::One(_) | QueryResult::OneCursor(_) | QueryResult::Many(_) => {
+                    unreachable!(
+                        "eval_pipe_with_path_context_internal only ever produces \
+                         Owned/ManyOwned/None/Error/Break/Partial"
+                    )
+                }
+            };
+            let result = OwnedValue::Array(items);
+            if rest.is_empty() {
+                QueryResult::Owned(result)
+            } else {
+                eval_pipe_with_path_context_internal::<W, S>(
+                    rest,
+                    &result,
+                    &result,
+                    file_origin,
+                    &[],
+                    optional,
+                )
             }
         }
         Expr::Object(_) | Expr::Array(_) | Expr::Literal(_) => {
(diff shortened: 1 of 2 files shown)


newhoggy:
## Code review response

Ran the multi-agent review (8 finder angles + a follow-up live-verification synthesis pass). Summary of what was found and done:

**Fixed — genuine, dual-oracle-confirmed regression:**
- **`?`/optional atomicity break** — independently rediscovered by three separate finder passes. The new `Expr::Array` arm passed the ambient `optional` flag straight into evaluating the array's own inner expression, letting a leaf error deep inside (`error(...)`, a type error) self-swallow via its own local `if optional` check *before* ever reaching the array's own atomicity match — so `.a | [key, error("boom")]?` produced a corrupted partial array (`["a"]`) instead of the whole construction being caught. Live-verified against real jq: `.a | [1, error("x")]?` correctly produces no output; real yq v4.53.3's own `[key, error(...)]?` hard-errors — succinctly's `["a"]` matched neither oracle. Fixed by forcing `optional=false` for the inner evaluation (so a genuine error surfaces as a real, catchable `Error`/`Partial(_, Error)`), then catching it locally using the array's *own* `optional` — the same evaluate-then-catch shape `eval_try` already uses for the plain evaluator, and the same catch condition the sibling `Object`/`Array`/`Literal` arm already applies. Added `test_array_wrapped_path_context_builtin_optional_is_atomic_1302` covering the bug, the type-error variant, the "only content is an error" case (previously produced a spurious `[]`), the non-`?` hard-error baseline, and a genuinely-nested inner `?` (unaffected by the outer array's own optional state).
- **Related**: `QueryResult::None => vec![]` was conflating a genuine zero-output generator (#1280) with a swallowed error — fixed by the same change, since the error now surfaces correctly instead of masquerading as `None`.

**Fixed — cheap, low-risk:**
- Stale doc comment on `eval_owned_with_file_index` claiming "array/object literals ... remain out of scope" — updated to note the array case is now supported.

**Filed as follow-ups (confirmed live, but each needs its own design pass rather than a bolt-on to this PR):**
- [#1334](https://github.com/rust-works/succinctly/issues/1334) — `Expr::StringInterpolation` has the identical missing-`needs_path_context`-arm gap (`"\(key)"` also stubs to `null`), independently found by two finder angles.
- [#1335](https://github.com/rust-works/succinctly/issues/1335) — the broader root cause: `eval_pipe_with_path_context_internal`'s `Expr::Optional` dispatch broadcasts `optional=true` with no catch of its own (unlike `eval_try`), a pre-existing architectural gap that predates this PR and affects other arms too (confirmed independently via `Arithmetic`: `.a | (key == "a" and error("boom"))?` gives `false` instead of no output). This PR's fix is scoped locally to the `Array` arm; a general fix at the `Expr::Optional` dispatch point is a larger, separate change.

**Noted, not filed separately** (minor, informational): the new arm's `QueryResult`-collection match duplicates `eval_array_construction`'s nearly identical block, and the "reset root/path and continue with `rest`" tail duplicates the sibling `Object`/`Array`/`Literal` arm's — both flagged by the reuse/simplification angles as candidates for a shared helper if/when #1332 (Object) or #1334 (StringInterpolation) land and a third near-identical copy would otherwise accumulate. Also noted: `eval_generic.rs`'s own native `Expr::Array` arm doesn't bridge to path-context evaluation independently (matches the known dual-evaluator pattern, no live regression found); minor unmemoized re-walk of `needs_path_context` on repeated dynamic visits (e.g. inside `.[]`/`map()` fan-out) — negligible in practice.

**Not actionable / scope note only:** `#1332` (Object) already covers that instance; `del()`/`getpath`/`setpath` weren't touched and don't share this gap in a way that needs action here.

All fixes re-verified after rebase onto latest `main`: full test suite (0 failures), both required clippy invocations clean, `cargo fmt --check` clean, patch coverage 100% (all new lines covered).

Issue #1219, opened 2026-08-17
yq: del() through a chained slice with more path after it (.[1:3][0]) doesn't get the parent-key-drop rule


newhoggy:
**Severity: Low**

## Summary

Split off from #1162 during implementation — explicitly noted in that issue's
own scope as broader than what it covers, filing to make it trackable.

`yq_del_scalar_slice_parent_path` (src/jq/eval.rs) only rewrites a resolved
`del()` path when the trailing slice is the path's *last* component
(`del(.a[0:2])`). When more path continues *after* the slice
(`del(.a[1:3][0])`), that function's own guard (`!matches!(last, Expr::Slice
{ .. })`) returns `None` immediately, and `delete_at_path`'s ordinary
`Expr::Slice { start, end }` arm handles it instead — descending into the
sliced sub-range and deleting within it, via `through_slice` with
`scalar_noop: false, container_noop: false`.

## Repro (verified against real yq v4.53.3)

```
$ printf 'a: [1,2,3,4]\n' | yq -o=json 'del(.a[1:3][0])'
a: [1,2,3,4]                          # real yq: whole thing untouched (no-op)

$ printf 'a: [1,2,3,4]\n' | succinctly yq -o=json 'del(.a[1:3][0])'
{"a":[1,3,4]}                          # wrong: descends into the sub-range and deletes within it
```

Reproduces the same way whether the *target* reached through the slice is a
container or a scalar.

## Why this is separate from #1162

#1162 fixed the "slice is the *last* path component" shape only (widening
`yq_del_scalar_slice_parent_path`'s rewrite from scalar-only to every target
type). This issue is the structurally different "slice has more path after
it" shape, which never reaches that rewrite at all — `delete_at_path`'s own
`Expr::Slice { start, end }` arm (not the rewrite) is what needs the new
behavior here, and what "the parent to drop" even means for a slice with
tail components (delete the whole thing? just no-op like the bare-root
case?) needs its own investigation before implementing.

## Suggested fix direction

Live-verify the exact real-yq semantics across a few shapes (deeper tail
chains, tail through a `.[]`, tail through another slice) before deciding
whether this is a no-op (matching the bare-root case) or something else —
don't assume it generalizes from either #1116's or #1162's existing rules
without checking.

newhoggy:
## Implementation plan (2026-08-20) — semantics verified, rule is simple

The issue asks for the real-yq semantics to be live-verified across several
[...]

Issue #1142, opened 2026-08-17
yq: slice compound-assignment on a real array/string target silently corrupts data instead of real yq's no-op


newhoggy:
## Summary

Found while implementing/reviewing #1119. Real yq's slice-target compound assignment (`.path[S:E] op= value`) is a **no-op** for a real array (or string) target whenever the write-back would change its length — not just for the scalar-target case #1101/#1112 already handle. succinctly instead silently splices the write-through result into the array at the slice position, corrupting it.

## Repro

```
$ printf 'a: [1,2,3]\n' | yq '.a[0:2] += [99]' -o=json
{"a": [1, 2, 3]}                      # real yq: no-op, unchanged

$ printf 'a: [1,2,3]\n' | succinctly yq '.a[0:2] += [99]' -o=json
{"a": [1, 2, 99, 3]}                  # wrong: splices [1,2]+[99]=[1,2,99] into positions 0:2
```

This reproduces on **current `main`**, independent of #1119 — `array + array` concatenation already worked before #1119, so this shape was already reachable. #1119 (yq's `+` now appends a non-array RHS to an array LHS) makes the *identical* corruption newly reachable through a far more common shape:

```
$ printf 'a: [1,2,3]\n' | succinctly yq '.a[0:2] += 99' -o=json   # post-#1119
{"a": [1, 2, 99, 3]}                  # wrong (real yq: {"a":[1,2,3]}, unchanged)
```

Pre-#1119, this exact query errored loudly instead (`array ([1,2]) and number (99) cannot be added`) — a hard failure, not silent corruption. #1119 removes that accidental protection for the common `+= <number>`/`+= <string>` case, without being the root cause.

Also reproduces for zero-width slices (`.a[0:0] += 99` → real yq no-op, succinctly inserts at front: `[99,1,2,3]`) and nested slices (`.a[0][0:1] += 99`).

## Root cause (tentative)

`update_path`'s slice write-back (`through_slice`, roughly `src/jq/eval.rs:10800-10850` as of #1119's branch) reads the sub-array, runs the filter/RHS against it, and splices whatever comes back into the original array at the slice's bounds — with no length-change guard. Real yq's actual behavior here needs re-verifying rather than assumed; #1101's doc comment (`is_yq_slice_empty_container_scalar`, `eval.rs:10038-10043`) already notes real yq's slice no-op "applies for any target type," which this issue's repro confirms extends beyond the scalar case that comment's own gate currently covers.

## Scope note

[...]

PR #1337, opened 2026-08-20
fix(jq): report the tail's own error for a mid-chain string slice

### LINKED ISSUE #1321: jq: del() on a string slice followed by more path reports the wrong error
**Severity: Low**

## Summary

Found while verifying #1312 (fixes a sibling gap in \`del()\`'s standalone slice arm). \`del()\` on a slice through a string, with more path *after* the slice, reports \`Cannot update string slices\` where real jq reports the ordinary field/index-on-string error for whatever follows.

```
$ echo '{"a":"hello"}' | jq -c 'del(.a[0:1].b)'
jq: error (at <stdin>:1): Cannot index string with string "b"
$ echo '{"a":"hello"}' | succinctly jq -c 'del(.a[0:1].b)'
```

Reproduces identically with or without a \`?\` on the slice (\`del(.a[0:1]?.b)\`), and confirmed pre-existing on \`main\` before #1312 — that PR's fix only touches the *standalone* (terminal) \`Expr::Slice\` arm; this is the *mid-chain* case (more path after the slice), a different code path entirely.

## Why this is separate from #1312

#1312 was about \`optional\` wrongly shadowing the write-time \`String\` refusal in \`delete_at_path\`'s standalone slice arm. This is a different bug: the mid-chain case correctly reaches \`through_slice\` (unlike the standalone case before #1312's fix), but \`through_slice\` itself unconditionally raises \`cannot_update_string_slices()\` for any \`OwnedValue::String\` root — it has no way to know whether the slice is the *terminal* write target (where "cannot update string slices" is exactly right) or just a read-through step with more path following (where jq instead evaluates \`.b\` against whatever the slice would have read, and reports *that* failure).

Real jq apparently treats a string slice as *readable* even in write position when there's more path after it — it slices the string, then continues navigating \`.b\` against the resulting substring, and only then discovers \`.b\` doesn't apply to a string.

## Suggested fix direction

Not investigated. \`through_slice\`'s \`OwnedValue::String(_) => Err(cannot_update_string_slices())\` arm may need to distinguish "this slice is the last path component" from "more path follows" — possibly by only raising this specific error when \`edit\` is the terminal write (matching what \`slice_owned_value\`'s read-path already does for a string, materializing a real substring), and letting a non-terminal continuation actually run \`edit\` against a read substring so its own error surfaces instead. Worth checking whether \`set_path\`/\`update_path\` share this same gap for \`=\`/\`|=\`, or whether it's \`del()\`-specific.

## Related

Found during review/testing of #1312 (fixes a different \`del()\` slice bug). Not caused by, or fixed by, that PR.

newhoggy:
## Summary

`del()`/`=`/`|=` through a string slice with more path *after* the slice (`del(.a[0:1].b)`, `.a[0:1].b = 5`) reported `Cannot update string slices` — the write-time refusal meant for the slice as the terminal write target — instead of whatever the following path component's own navigation actually failed with (`Cannot index string with string "b"`, matching real jq).

## Root cause

`through_slice`'s `OwnedValue::String` arm unconditionally refused any string root with `cannot_update_string_slices()`, with no way to tell "this slice write *is* the target" from "there's more path after it to navigate through." `del()`'s own mid-chain caller (and `=`/`|=`'s equivalent) correctly reached `through_slice`, but that refusal fired before the continuation (`edit`) ever got a chance to raise its own, more specific error.

Live-verified against real jq 1.7.1 that `=`/`|=` share the identical bug — `.a[0:1].b = 5` also reported the wrong message before this fix — so the fix isn't `del()`-specific.

## Fix

- Added a `terminal_write` flag to `through_slice` distinguishing "this call's `edit` closure *is* the write" (terminal — always refuse, matching jq's real behavior for `.a[0:1] |= f`) from "the write happens *inside* `edit`, if at all" (mid-chain — route through a real substring read, reusing the existing `container_noop` read-through mechanism, and let `edit`'s own result — a genuine navigation error, or a `?`-suppressed no-op — answer instead of the blanket refusal).
- Bundled the now-four call-site flags (`optional`/`scalar_noop`/`container_noop`/`terminal_write`) into a `SliceEditFlags` struct — clippy's `too_many_arguments`/`fn_params_excessive_bools` correctly flagged the fourth bare `bool`.
- `set_path`'s own `split_at_slice`-based mid-chain path needed `terminal_write` computed *dynamically* (`matches!(split.tail, Expr::Identity)`) rather than a fixed value per call site: an `Identity` tail means the slice is genuinely the terminal target, disguised as mid-chain by that function's flattening. Caught by testing the exact regression an earlier, statically-classified version of this fix would have introduced (`.a[0:1] = v` silently no-op'ing instead of erroring). `update_path` and `delete_at_path`'s own mid-chain arms don't share this ambiguity — their `rest`/tail is always genuinely non-empty by construction, confirmed by tracing both call chains.

## Verification

- Confirmed live against real jq 1.7.1 for every combination: mid-chain field/index/iterate access (non-`?` errors with the tail's own message, `?` no-ops), the terminal slice-is-the-whole-target case (still refuses, both bare-root and through a preceding field), and array/object mid-chain targets (unaffected, still splice their own sub-range deletion/update back).
- New tests: `test_del_string_slice_mid_chain_reports_the_tail_own_error_1321`, `test_assign_string_slice_mid_chain_reports_the_tail_own_error_1321` (both in `src/jq/eval.rs`'s own unit test module, matching #1312's precedent).
- Regression-checked against the existing `#1101`/`#1116`/`#1142`/`#1162`/`#1219`/`#1312` slice/string test families (129 tests, all pass).
- Patch coverage 100% (86/86 new lines).

Fixes #1321

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -20948,6 +20948,22 @@ fn delete_at_path(
                     // reached through the slice) — broader than #1162's own
                     // scope (slice as the path's *last* component), needing
                     // its own investigation before it's fixed.
+                    //
+                    // `terminal_write: yq_mode` (#1321 code review): a
+                    // string target here is the *same* #1219 divergence,
+                    // just for a scalar rather than a container -- yq
+                    // silently no-ops it (confirmed live, `del(.a[0:1].b)`
+                    // on `a: hello` leaves `a: hello` untouched), which
+                    // this PR doesn't attempt to fix. Forcing
+                    // `terminal_write: true` in yq mode keeps
+                    // `through_slice`'s String arm on its old, unconditional
+                    // refusal there -- the exact same wrong-but-unchanged
+                    // "Cannot update string slices" message yq mode already
+                    // raised before this PR -- instead of trading it for a
+                    // *different* wrong message ("Cannot index string with
+                    // string \"b\"") that would make this string case look
+                    // fixed relative to #1219's own array/object case when
+                    // it isn't. jq mode is unaffected either way.
                     Expr::Slice { start, end } => through_slice(
                         root,
                         *start,
@@ -20956,7 +20972,7 @@ fn delete_at_path(
                             optional: here,
                             scalar_noop: false,
                             container_noop: false,
-                            terminal_write: false,
+                            terminal_write: yq_mode,
                         },
                         |sub| delete_at_path(sub, &rest, optional, yq_mode),
                     ),
@@ -41475,6 +41491,34 @@ mod tests {
         query!(br"[1,[2],[3]]", r"del(.[1:3][0])",
             QueryResult::Owned(v) => {
                 assert_eq!(v.to_json(), "[1,[3]]");
+            }
+        );
+    }
+
+    #[test]
+    fn test_del_string_slice_mid_chain_yq_mode_unaffected_1321() {
+        // #1321 code review: yq mode's own del() call site into
+        // `through_slice` deliberately forces `terminal_write: true` for a
+        // string target, keeping the OLD, unconditional "Cannot update
+        // string slices" refusal there byte-for-byte unchanged -- NOT the
+        // new jq-mode fix above. This case is the same real-yq-diverges
+        // gap #1219 already tracks for a container target
+        // (`del(.a[1:3][0])` on an array): real yq silently no-ops it
+        // (confirmed live, `del(.a[0:1].b)` on `a: hello` leaves `a:
+        // hello` untouched), which is #1219's own scope to fix, not this
+        // issue's. Pinning the *unchanged* wrong message here so a future
+        // edit doesn't silently swap it for a different-but-still-wrong
+        // one and make this string case look fixed relative to #1219's
+        // still-open array/object case.
+        yq_query!(br#"{"a":"hello"}"#, r"del(.a[0:1].b)",
+            QueryResult::Error(e) => {
+                assert_eq!(e.message, "Cannot update string slices");
+            }
+        );
+        // jq mode is unaffected by this yq-only gating.
+        query!(br#"{"a":"hello"}"#, r"del(.a[0:1].b)",
+            QueryResult::Error(e) => {
+                assert_eq!(e.message, r#"Cannot index string with string "b""#);
             }
         );
     }
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -11502,9 +11502,18 @@ fn set_path(
                         parent,
                         split.start,
                         split.end,
-                        split.optional,
-                        scalar_noop,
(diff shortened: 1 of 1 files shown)


newhoggy:
## Code review response

**Fixed — genuine, live-verified issue:**
- **yq-mode del() message swap** — the review found that this PR's fix, while correct for jq mode, unconditionally changed yq mode's `del(.a[0:1].b)` output too: trading the old wrong message (`Cannot update string slices`) for a *different* wrong message (`Cannot index string with string "b"`), neither of which matches real yq's actual behavior (silently no-ops the whole thing — the same #1219-tracked container-target divergence, just not yet fixed for a scalar target). Fixed by forcing `terminal_write: yq_mode` at that one call site (del()'s mid-chain `Expr::Slice` arm), so yq mode's observable output is now byte-for-byte unchanged by this PR — jq mode's fix is unaffected. Added `test_del_string_slice_mid_chain_yq_mode_unaffected_1321` pinning the unchanged message, so a future edit can't silently swap it for a different wrong one without a test catching it.

**Filed as a follow-up, not fixed here:**
- [#1338](https://github.com/rust-works/succinctly/issues/1338) — `setpath()`'s own backing walker, `set_value_at_path`, has the identical bug (confirmed live: `setpath([{"start":0,"end":1},"b"]; 5)` on a string reports the wrong message too) but is a structurally separate function this PR's `through_slice` fix doesn't reach.

**Noted, not acted on** (minor/informational, no correctness impact):
- The new `OwnedValue::String(_) if !terminal_write` arm in `through_slice` duplicates the pre-existing `container_noop` arm's body (same read-through-then-discard shape) — could be merged, but the two conditions aren't quite parallel (`container_noop` also covers Array; `terminal_write` is String-specific), and merging cleanly without introducing risk wasn't worth it in an already-large diff.
- `SliceSplit.tail: Expr` uses `Expr::Identity` as a sentinel rather than `Option<Expr>`, forcing `set_path`'s call site to reconstruct `terminal_write` via a `matches!` check instead of a direct `is_none()`. Correct today, but a `matches!`-based reconstruction is more fragile than a compiler-enforced `Option`. Left as-is to avoid widening this PR's diff into `split_at_slice`'s own return type.
- The new mid-chain String arm always builds the substring via `slice_owned_value` even when `edit` will immediately error without using it — mirrors the pre-existing `container_noop` arm's already-accepted cost, not a new pattern.

All fixes re-verified: full test suite (0 failures after a re-run past one transient contention flake — 10 unrelated tests, all showing "Blocking waiting for file lock" in their own captured output, matching this session's known machine-contention pattern), both required clippy invocations clean, `cargo fmt --check` clean, patch coverage 100% (94/94 new lines).

PR #1339, opened 2026-08-20
fix(jq): let a ? on a multi-component group reach = through more path

### LINKED ISSUE #1311: jq: a ? on a multi-component group ahead of more path still hits 'invalid path component' for =
**Severity: Low**

## Summary

Found during review of #1310 (fixes #1303's simpler cases). `((.a[0:1])? | .[0]) = 9` — a `?` on a whole multi-component group, with more path following it — still reports the generic `invalid path component` for `=` where jq performs the write. `|=`/`del()` already handle this shape correctly.

```
$ echo '{"a":[1,2,3]}' | jq -c 'path((.a[0:1])? | .[0])'
["a",{"start":0,"end":1},0]
$ echo '{"a":[1,2,3]}' | jq -c '((.a[0:1])? | .[0]) = 9'
{"a":[9,2,3]}
$ echo '{"a":[1,2,3]}' | succinctly jq -c '((.a[0:1])? | .[0]) = 9'
```

Also reproduces the non-sliceable-target boundary (should suppress, doesn't):
```
$ echo '{"a":true}' | jq -c '((.a[0:1])? | .[0]) = 9'
{"a":true}
$ echo '{"a":true}' | succinctly jq -c '((.a[0:1])? | .[0]) = 9'
```

Confirmed `|=` already correct on both: `((.a[0:1])? | .[0]) |= . + 8` gives `{"a":[9,2,3]}`, and the non-sliceable-target case correctly no-ops.

Unlike the glued postfix form (`(.a|.c)?[0] = 9`), which is a confirmed parse error in both real jq and succinctly, this `|`-separated spelling genuinely parses and path-tracks — it isn't the "unreachable via real syntax" case #1287's/#1303's own investigations ruled out elsewhere.

## Root cause

`split_at_slice` (`src/jq/eval.rs`, added by #1292, extended by #1303/PR #1310 to see through a `?` directly wrapping a `Slice`) flattens the top-level path via `push_path_components` then scans for the first component whose `unwrap_path_component`-peeled core is a `Slice`. For `((.a[0:1])? | .[0])`, the first flattened element is `Optional(Paren(Pipe([Field("a"), Slice{0,1}])))` — `unwrap_path_component` peels the `Optional`/`Paren` down to the `Pipe` itself, which isn't a `Slice`, so the scan still misses it. `push_path_components` doesn't recurse *through* `Optional` either (by design, for its other callers), so the `Pipe`'s own components never reach the flattened list unwrapped.

`update_path` (`|=`) doesn't share this gap: its `Expr::Pipe` arm (fixed by #1294) walks the path one element at a time via `unwrap_path_component` + `splice_optional_group`, so it naturally handles an `Optional`-wrapped multi-component group without needing a separate flatten-and-scan pass.

## Suggested fix direction

Not investigated in depth. `split_at_slice`'s flatten-then-scan design is structurally the wrong shape for this: it needs `update_path`'s one-element-at-a-time approach instead, splicing an `Optional(Pipe(...))` group's own components (scoped via `splice_optional_group`, already shared between `update_path`/`delete_at_path`, #1294) as it walks, watching for a `Slice` at each step rather than pre-flattening the whole list and scanning once. This is likely a genuine rework of `set_path`'s `Expr::Pipe` arm / `split_at_slice`'s role in it, not another point patch — three separate PRs (#1290, #1299, #1310) have now each closed one gap in this area and had review surface another, which is itself worth factoring into how this gets scoped.

## Related

Found during review of #1310 (fixes #1303). Not fixed by that PR — its own doc comment on `split_at_slice` documents this gap and points here.

newhoggy:
## Summary

`((.a[0:1])? | .[0]) = 9` — a `?` on a whole parenthesized multi-component group, with more path after it — reported `invalid path component` for `=` where real jq performs the write. `|=`/`del()` already handled this exact shape correctly.

## Root cause

`push_path_components` (used by `set_path`'s `split_at_slice` pre-pass to flatten a chained path before scanning for a slice) had no arm for `Expr::Optional`. For `((.a[0:1])? | .[0])`, the flattening pass hit `Optional(Paren(Pipe([Field("a"), Slice{0,1}])))` as one element of the outer `Pipe` and pushed it verbatim (via the catch-all arm) as one opaque component, instead of seeing through it. Neither `split_at_slice`'s subsequent slice scan (which unwraps `Optional`/`Paren` but doesn't recurse into a `Pipe` found underneath) nor `get_path_mut`'s own per-step `Identity | Field | Index` match could then make sense of it, so the whole thing fell to the generic `invalid path component` catch-all.

`update_path` (`|=`) and `delete_at_path` (`del()`) don't share this gap: their own `Expr::Pipe(inner)` arms already call `splice_optional_group` (added by #1294) to splice a resolved `Optional(Pipe([…]))` component's own elements into the walk, each individually re-wrapped in `Optional` so the group's `?` scopes to each of its own components rather than leaking onto what comes after.

## Fix

Added an `Expr::Optional(inner)` arm to `push_path_components` mirroring `splice_optional_group`'s identical reasoning: recursively flatten `inner` into a temporary list (reusing the function's own existing `Pipe`/`Paren` flattening), then wrap each resulting component individually in `Expr::Optional` before appending to the output. This makes `((.a[0:1])? | .[0])` flatten to `[Optional(Field("a")), Optional(Slice{0,1}), Index(0)]` — `split_at_slice`'s scan now correctly finds the (still-optional) slice, and `get_path_mut`'s per-step navigation already understands a per-component `Optional` wrapper (confirmed by the pre-existing `.a?.b.c = 1` behavior).

## Verification

- Confirmed live against real jq 1.7.1 for the primary repro, the non-sliceable-target boundary case (should suppress via navigation failure, not a write-time refusal), and a multi-component variant with a field before the slice.
- New test: `test_assign_optional_group_with_more_path_1311` in `src/jq/eval.rs`'s own unit test module, matching #1303's established convention.
- Regression-checked against every other `push_path_components` caller (`split_at_slice`, `resolve_leaf`, `is_passthrough_target`, `resolve_seq`, `resolve_dynamic_indexes`) with live spot-checks of `path()`, `del()`, and a computed key through an optional group — all match real jq.
- Regression-checked against the #1287/#1292/#1294/#1303 test families (167 tests, all pass).
- Patch coverage 100% (27/27 new lines).

Fixes #1311

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -11943,23 +11943,17 @@ struct SliceSplit {
 /// Each flattened element is matched via [`unwrap_path_component`] (not a
 /// bare `matches!`), so a `?`-marked slice — `.a[0:1]?[0] = 9`, or
 /// `(.a[0:1]?)[0] = 9` once `push_path_components` has flattened the
-/// enclosing parens away — is still found (#1303): `push_path_components`
-/// itself leaves `Expr::Optional` opaque (its other callers need that), so
-/// without this the wrapped `Slice` reaches this scan unrecognized.
+/// enclosing parens away — is still found (#1303).
 ///
-/// Still does *not* reach through a `Pipe`/`Paren` left nested *inside* an
-/// `Optional` wrapper — e.g. a `?` on a whole multi-component group ahead of
-/// a further step, `((.a[0:1])? | .[0]) = 9`. `unwrap_path_component`'s peel
-/// stops at that `Pipe`, which isn't itself a `Slice`, so the scan still
-/// misses it. Unlike the glued-postfix form (`(.a|.c)?[0] = 9`, confirmed a
-/// jq parse error), the `|`-separated spelling above genuinely parses in
-/// both real jq and here (`path((.a[0:1])? | .[0])` resolves fine) and does
-/// still reproduce the "invalid path component" fallback this whole function
-/// exists to close — found during this fix's own review, tracked separately
-/// rather than folded in here since closing it properly likely means
-/// reworking this function to unwrap one path element at a time (mirroring
-/// `update_path`'s already-correct `Expr::Pipe` arm) instead of a single
-/// flatten-then-scan pass, not another incremental patch to this one.
+/// Also reaches through a `Pipe`/`Paren` nested *inside* an `Optional`
+/// wrapper — e.g. a `?` on a whole multi-component group ahead of a further
+/// step, `((.a[0:1])? | .[0]) = 9` — since `push_path_components` (#1311)
+/// recursively flattens an `Expr::Optional`'s own contents and re-wraps each
+/// resulting component individually, rather than leaving the wrapper
+/// opaque. Unlike the glued-postfix form (`(.a|.c)?[0] = 9`, confirmed a jq
+/// parse error), the `|`-separated spelling above genuinely parses in both
+/// real jq and here (`path((.a[0:1])? | .[0])` resolves fine), and now
+/// resolves through `=` too, matching jq.
 fn split_at_slice(exprs: &[Expr]) -> Option<SliceSplit> {
     // Cheap common-case guard: the overwhelming majority of assignment
     // targets (`.a.b.c[2] = 9`) contain neither a slice nor a nested
@@ -14448,7 +14442,12 @@ fn type_filter_matches(builtin: &Builtin, value: &OwnedValue) -> bool {
 /// Reuses [`push_path_components`]'s own Identity/Pipe/Paren unwrapping
 /// rather than a bespoke `matches!` — `.[.k]` and `. | .[.k]` and `(.)[.k]`
 /// all name the same no-op target, and this is the one existing place in
-/// the file that already knows how to recognise that.
+/// the file that already knows how to recognise that. `push_path_components`
+/// also unwraps a bare `Optional(Identity)` (`.?`) to empty since #1311, but
+/// that shape can never actually reach `target` here: `parse_postfix` only
+/// lets `?` precede continued `[...]` when it's attached to a `Field`/prior
+/// bracket, never to a passthrough node, so `.?[0]`/`(.)?[0]` are parse
+/// errors in both real jq and here.
 fn is_passthrough_target(target: &Expr) -> bool {
     let mut components = Vec::new();
     push_path_components(&mut components, target);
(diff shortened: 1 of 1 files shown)


newhoggy:
## Code review response

Multi-agent review completed (finder angles A-E plus cross-file tracer). Summary of findings and disposition:

**Fixed in this branch (new commit df2ba4841 → 0440dd9a6 post-rebase):**
- `split_at_slice`'s doc comment described `((.a[0:1])? | .[0]) = 9` as a still-open gap "tracked separately" — that's exactly what this PR's `push_path_components` change closes (it's the PR's own first test case). Rewrote the paragraph to describe the fix instead of the gap, and corrected the adjacent now-false claim that `push_path_components` "leaves `Expr::Optional` opaque."
- `is_passthrough_target`'s doc comment didn't mention that the same change also unwraps a bare `Optional(Identity)` (`.?`). Added a note — plus why that specific shape can never actually reach `is_passthrough_target`'s `target` param: the parser only allows `?` to precede continued `[...]` when attached to a `Field`/prior bracket, never to a passthrough node, so `.?[0]`/`(.)?[0]` are parse errors in both real jq and here. (Independently confirmed by the cross-file tracer angle.)
- `resolve_seq`'s doc comment didn't note that `.?` alone now correctly contributes zero components to `flat` (same as `.`) rather than one opaque component — a direct, correct consequence of `push_path_components` no longer treating `Optional` as opaque.

**Investigated, determined pre-existing, filed separately (not fixed here):**
- Reviewers found `((.b[0:1])? | .c) = 9` on `{"c":0}` (`.b` absent) silently produces `{"c":0,"b":null}` instead of real jq's `A slice of an array can only be assigned another array`. I verified this live against `origin/main` (i.e. *without* this PR's changes) using the simpler ungrouped form `(.b[0:1]? | .c) = 9` and got the identical silent-corruption result — that shape is a flat `Pipe([Field(b), Optional(Slice), Field(c)])` with no `Pipe`/`Paren` nested inside an `Optional`, so it doesn't touch this PR's new recursion at all. Confirmed pre-existing, unrelated to this PR beyond reaching it via one more syntactic spelling. Filed as #1340 with the root-cause trace so far (likely `through_slice`'s null-target auto-vivification interacting with write-time-error classification).

**Verified clean, no action needed:**
- All 5 non-`push_path_components`-recursion call sites (`resolve_leaf`, `is_passthrough_target`, `resolve_seq`, `resolve_dynamic_indexes`/`strip_resolved_optional`, `split_at_slice`) traced individually against real jq 1.7.1; no regressions found. `strip_resolved_optional` correctly collapses the newly-possible doubly-nested `Optional(Optional(X))` shape in one recursive pass.
- `resolve_seq`'s `#888` trailing-iterate-deferral optimization now also recognizes `Paren`-wrapped trailing iterates it previously missed (e.g. `.[$k] | (.x[])?`) — a mild, incidental `path()`-perf improvement from the same flattening, not a behavior change.

Re-ran fmt, both clippy invocations (`--all-features` and the CI feature set), and the full test suite (`cargo test --features cli,simd,regex,serde --no-fail-fast`, all green) after the doc-comment edits and again after rebasing onto current `main`.

Issue #1253, opened 2026-08-18
jq: normal-magnitude scientific notation now inherits the 32-digit mantissa truncation cap


newhoggy:
## Background

#1206 (PR #1249) fixed two real bugs in `format_number_jq_compat`'s
scientific-notation mantissa rendering by reusing the string-based
`mantissa_str`/`shifted_exp` pair `normalize_extreme_literal_mantissa`
already computes, instead of a separate, buggy `libm::log10`/`libm::pow`-based
recomputation. That fix is a strict improvement in every case tested
(previously ~15-17 significant digits via f64 rounding; now full source-digit
fidelity up to the cap below) — but it also newly routes *ordinary,
normal-magnitude* scientific-notation literals through
`normalize_extreme_literal_mantissa`'s existing `MAX_RENDERED_MANTISSA_DIGITS
= 32` rendering cap, previously reachable only via the overflow
(`format_overflow_literal_mantissa`) and near-zero/subnormal
(`format_near_zero_literal`) paths.

## Repro

```bash
$ echo '1.234567890123456789012345678901234567890123e50' | succinctly jq -c .
1.23456789012345678901234567890123E+50    # truncated to 33 significant digits

$ echo '1.234567890123456789012345678901234567890123e50' | jq -c .
1.234567890123456789012345678901234567890123E+50   # real jq: full 44-digit mantissa preserved
```

Confirmed live against jq 1.7.1 during #1206's own code review (four
independent review passes each reproduced the exact break point: 33
significant digits round-trips correctly, 34+ truncates).

## Why this is separate from #1206

#1206's own scope was the two documented bugs (trailing-zero loss,
mantissa rounding to exactly `10`) — both are now fixed and confirmed
correct against the oracle. This is a different, narrower issue: real
jq's own literal-preservation ceiling for scientific notation appears to
be much higher (100+ digits tested, exact) than this crate's
`MAX_RENDERED_MANTISSA_DIGITS = 32`, which was chosen for the
overflow/near-zero paths specifically and is now silently inherited by
the ordinary case too. Not a regression from #1206 (pre-#1206 behavior
was *more* wrong, truncating around 15-17 digits via lossy f64 rounding)
— just a new-domain exposure of an existing, already-tested-as-bounded
constant.

## Suggested direction

Not scoped in detail. Possible directions:
- Raise `MAX_RENDERED_MANTISSA_DIGITS` significantly (real jq's own
  ceiling isn't obviously documented; would need oracle probing to find
[...]

Issue #1244, opened 2026-08-18
jq: large positive-magnitude literal notation choice depends on significant-digit count, not a simple exponent window


newhoggy:
**Severity: Low**

## Summary

Found while implementing #1226 (nonzero-mantissa notation threshold uses the
written exponent instead of the shifted/magnitude one). #1226's own fix is
correctly scoped to the *negative*-shifted-exponent side (the near-zero
decimal/scientific boundary, matching #1207's established `-6..=-1`/`0`
window) — while investigating it, live testing turned up that the
*positive*-shifted-exponent side (large-magnitude literals) does **not**
follow a comparably simple rule, and is a separate, deeper divergence.

## Repro (verified against pinned jq 1.7.1)

```
$ echo '500000e-1'          | jq -c .   ->  50000.0        (plain decimal)
$ echo '500000e-1'          | succinctly jq -c .  ->  5E+4  (scientific, wrong either way -- pre-#1226 gave "50000", also wrong)

$ echo '99999999999999e0'   | jq -c .   ->  99999999999999      (bare integer)
$ echo '99999999999999e1'   | jq -c .   ->  9.9999999999999E+14 (scientific)
```

Both `500000e-1` (shifted exponent 4) and `99999999999999e1` (shifted
exponent 14) have *positive* shifted exponents, yet jq renders the first as
plain decimal and the second as scientific — the boundary is not simply "is
the shifted exponent above some threshold." Empirically, the distinguishing
factor appears to be the value's own **significant-digit count** under jq's
underlying shortest-round-tripping (dtoa-style) representation:
`500000e-1` (value `50000.0`) needs only 1 significant digit, while
`99999999999999e1` (value `999999999999990.0`) needs 14 — but this is only a
hypothesis from a handful of probes, not a derived rule.

`succinctly`'s own scientific rendering for `99999999999999e1` is separately
buggy too, independent of the notation-choice question: it currently prints
`10E+14` (mantissa not even normalized to one digit before the point,
confirmed present on `main` before #1226 and unaffected by it) instead of
`9.9999999999999E+14` — likely the same `log10`/`pow` renormalization
imprecision already documented for other magnitude ranges elsewhere in this
file (see the module's own `#1177`/subnormal-handling notes for the
established pattern of "string-based mantissa derivation, not float
arithmetic, once precision is already lost").

## Why this is separate from #1226

#1226's own scope (and its live-verified repro set) is exclusively about
[...]

PR #1345, opened 2026-08-21
fix(jq): defer overflow-literal mantissa truncation until notation is known

### LINKED ISSUE #1274: jq: overflowed literal past MAX_RENDERED_MANTISSA_DIGITS+1 digits wrongly flips from plain to scientific notation
**Severity: Low**

## Summary

Found during code review on PR #1272 (#1270's fix), via live execution --
confirmed pre-existing and unaffected by that PR (the two constants involved,
`MAX_RENDERED_MANTISSA_DIGITS = 100_000` and the overflow ceiling
`1_000_000_000`, were both already present and both already fit comfortably
in the pre-#1270 `i64` type). Not a regression from #1270/PR #1272, filed
separately since it's a real, distinct bug.

`format_positive_shifted_plain`'s own doc comment states its
`split_pos > digits.len()` truncation guard is "provably unreachable" given
that `format_number_jq_compat`'s own call site never sees a `shifted_exp`
past `~308` (finite `f64`'s own representable range). That reasoning is
correct for *that* call site, but `format_overflow_literal_mantissa` -- the
other caller -- has no such bound on `new_exp` (only capped at
`1_000_000_000` by its own, unrelated ceiling check). Once a literal's
mantissa exceeds `MAX_RENDERED_MANTISSA_DIGITS` significant digits,
`mantissa_str` gets truncated to `MAX_RENDERED_MANTISSA_DIGITS + 1` chars
while `digit_count` still reflects the *true*, untruncated digit count --
so the `shifted_exp < digit_count` check (correctly deciding "should stay
plain") can pass while the truncated `digits.len()` is too short to actually
place the decimal point, silently flipping the result to scientific
notation instead of plain.

## Repro (verified live against the current build)

```
$ python3 -c "print('9'*400 + 'e0')" | succinctly jq -c .
999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999
   (400 nines, plain -- matches real jq, matches this function's own #1244 rule)

$ python3 -c "print('9'*100002 + 'e0')" | succinctly jq -c .
9.999...9E+100001   (scientific, truncated -- the same #1244 rule says this should stay plain)
```

The boundary is exact: `"9"*100_001 + "e0"` (right at the cap) still
correctly renders plain; `"9"*100_002 + "e0"` (one digit past it) flips to
truncated scientific notation. Both are internally inconsistent with this
same function's own documented rule (see its `"9".repeat(400)` example),
not just a divergence from real jq.

## Why this is separate from #1270

#1270's own scope was the exponent-parsing sentinel-leak bug
(`parse_literal_exponent`'s `i64`/`i128` saturation). This is a different
bug in a different function (`format_positive_shifted_plain`'s truncation
interaction with `format_overflow_literal_mantissa`'s unbounded
`new_exp`), only surfaced by that PR's code review doing a full consistency
check across `format_positive_shifted_plain`'s three call sites.

## Suggested fix direction

Not investigated in depth. `format_positive_shifted_plain`'s own doc
comment needs correcting first (it currently claims something false about
one of its two real callers). The actual fix likely needs the
`shifted_exp >= digit_count` decision to also account for whether
`mantissa_str` was itself truncated below what placing the decimal point at
`shifted_exp` would require -- effectively, treat "truncated past what's
needed" the same as "not enough given digits" for notation-choice purposes,
even though `digit_count` (the *true* count) says otherwise. Likely a small,
localized fix once scoped properly.

newhoggy:
## Summary

Fixes #1274: `succinctly jq` wrongly flipped a literal with more than `MAX_RENDERED_MANTISSA_DIGITS + 1` (100,001+) significant digits from plain to (truncated) scientific notation whenever the literal's own given digits were enough to cover the shift and stay plain. Real jq stays plain unconditionally at this scale (oracle-verified past 500,000 digits, no ceiling found).

```bash
$ python3 -c "print('9'*100002 + 'e0')" | jq -c .
999...9   # plain, all 100,002 nines

$ python3 -c "print('9'*100002 + 'e0')" | succinctly jq -c .   # before this PR
9.999...9E+100001   # wrongly flipped to truncated scientific
```

## Root cause

`normalize_extreme_literal_mantissa` truncates its returned mantissa text to `MAX_RENDERED_MANTISSA_DIGITS` before its caller ever gets to decide plain-vs-scientific notation. `format_positive_shifted_plain`'s own decision uses the *true* `digit_count` (never truncated), so it correctly commits to "should be plain" -- but then discovers mid-render that the already-truncated mantissa doesn't have enough surviving digits to actually place the decimal point, and falls back to (also-truncated) scientific notation instead.

**A more severe sibling bug surfaced during my own review of this fix**: a capped mantissa doesn't always *fail* in that situation -- when the decimal-point split falls *within* the truncated prefix rather than past it, `format_positive_shifted_plain` returns `Some(...)` anyway, silently rendering a plain result that's missing whatever trailing digits the cap elided. This is reachable via three call sites, not just the overflow path the issue originally reported: `format_overflow_literal_mantissa` (the issue's own repro), `format_number_jq_compat`'s ordinary (non-overflow) path (an everyday-magnitude value with an enormous fractional part), and `format_shifted_mantissa`'s unconditional `shifted_exp == 0`/`-6..=-1` decimal window (which never even considers a scientific fallback, so nothing there was checking digit count at all before this fix).

## Fix

Added an explicit `mantissa_digit_cap: Option<usize>` parameter to `normalize_extreme_literal_mantissa` (`None` = uncapped). A new shared helper, `full_mantissa_if_capped`, decides plain-vs-scientific eligibility from the cap-independent `digit_count` *before* materializing any mantissa text, and only re-derives an uncapped mantissa when the cap could actually have truncated something -- keeping the common case (well under 100,000 given digits) at its original cost. All three call sites that must render every given digit verbatim now go through this helper.

## Verification

- Oracle-probed real jq 1.7.1 extensively: the issue's own repro, a magnitude-`<1` sibling case, the ordinary-path case, a 200,000-digit and 500,000-digit scale check, plus a ~130-case combinatorial sweep and a 60-case randomized differential fuzz across digit-count/shift/sign combinations -- confirmed zero new mismatches against the fixed binary, and confirmed (by diffing against unmodified `main`) that the handful of remaining divergences are an unrelated, already-documented, pre-existing scientific-notation precision cap (`MAX_RENDERED_MANTISSA_DIGITS`'s own doc comment already frames this as an accepted trade-off, out of #1274's scope).
- New tests: the fixed boundary at both the old cap (`test_format_number_jq_compat_overflow_plain_stays_plain_past_the_render_cap_1274`), the magnitude-`<1` branch, and the ordinary (non-overflow) path's two sub-cases.
- All existing regression tests (`#1244`, `#1253`, leading-zero/near-zero/overflow boundary tests) still pass unchanged.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde --no-fail-fast` -- all green
- [x] `cargo check --no-default-features` (no_std) -- compiles (pre-existing unrelated dead-code warnings only)
- [x] `omni-dev coverage diff` -- 96.61% patch coverage (both remaining lines are provably-unreachable defensive branches, documented inline)
- [x] Oracle differential sweep (combinatorial + randomized) against real jq 1.7.1

### DIFF
--- a/src/jq/value.rs
+++ b/src/jq/value.rs
@@ -841,6 +841,14 @@ fn format_overflow_literal_mantissa(s: &str, exp_pos: usize, negative: bool) -> 
     // a redundant check here.
     if new_exp < digit_count {
         let full_mantissa_str = full_mantissa_if_capped(s, exp_pos, &mantissa_str, digit_count);
+        // Provably always `Some` here (not just usually): `full_mantissa_str`
+        // is uncapped whenever `full_mantissa_if_capped` had to re-derive at
+        // all, so its own digit count equals `digit_count` exactly, and
+        // `format_positive_shifted_plain`'s `split_pos > digits.len()` guard
+        // reduces to the `new_exp < digit_count` this branch already
+        // checked. Kept as `if let` rather than `.expect(..)` so a future
+        // change to either function's contract fails safe (falls to
+        // scientific) instead of panicking.
         if let Some(plain) =
             format_positive_shifted_plain(sign, &full_mantissa_str, new_exp, digit_count)
         {
--- a/src/jq/value.rs
+++ b/src/jq/value.rs
@@ -359,7 +359,7 @@ pub fn format_number_jq_compat(raw: &[u8]) -> String {
     // unlike `format_near_zero_literal` (#1273), this path can't actually
     // observe a saturated exponent in practice.
     let Ok((mantissa_str, shifted_exp, digit_count, _exp_saturated)) =
-        normalize_extreme_literal_mantissa(s, exp_pos)
+        normalize_extreme_literal_mantissa(s, exp_pos, Some(MAX_RENDERED_MANTISSA_DIGITS))
     else {
         unreachable!("nonzero value implies normalize_extreme_literal_mantissa succeeds")
     };
@@ -554,6 +554,40 @@ fn split_mantissa(s: &str, exp_pos: usize) -> (&str, &str) {
     mantissa.split_once('.').unwrap_or((mantissa, ""))
 }
 
+/// `dump_truncated`'s whole design keeps preview cost independent of the
+/// value's own size (see its doc comment) - a document-controlled mantissa
+/// of unbounded length must not turn into unbounded work in
+/// [`normalize_extreme_literal_mantissa`] when a caller doesn't need every
+/// given digit rendered anyway (scientific notation only ever shows a
+/// bounded prefix). Bounding the *copy* is enough: `shift` only ever needs
+/// `int_part.len()` (an O(1) property read once leading zeros are trimmed),
+/// so truncating what gets copied into `rest` doesn't touch the exponent
+/// math's correctness, only how many digits past the first one actually get
+/// rendered.
+///
+/// Real jq preserves a literal's full given precision unconditionally --
+/// oracle-verified exact round-trips up to exactly this many significant
+/// digits (#1253; raised from an earlier `32`, which was chosen for the
+/// overflow/near-zero paths specifically, then silently inherited by every
+/// other caller once #1206 routed the ordinary scientific-notation case
+/// through this same function). This *is* still a cap, not a proven
+/// ceiling: a literal with more true significant digits than this still
+/// renders truncated rather than erroring when a caller opts into it (`Some`
+/// below) -- code review on #1253 confirmed the truncation is silent, not
+/// that raising the number removes it. Set to what this session actually
+/// verified against pinned jq 1.7.1, not a round number chosen past it, so
+/// the constant and the oracle claim it rests on can't drift apart again the
+/// way `32` (tuned for a different use case entirely) drifted from the
+/// fidelity the ordinary case actually needs.
+///
+/// A caller whose own notation choice needs every given digit regardless of
+/// this cap -- [`format_overflow_literal_mantissa`]'s plain-notation retry,
+/// #1274 -- passes `None` to [`normalize_extreme_literal_mantissa`] instead
+/// of this constant, since plain notation's whole point is rendering every
+/// given digit; there is no bounded prefix to fall back on the way
+/// scientific notation has.
+const MAX_RENDERED_MANTISSA_DIGITS: usize = 100_000;
+
 /// Shift a literal's own mantissa digits (the text before `exp_pos`, i.e.
 /// before `e`/`E`) into jq's one-digit-before-the-point normalized form,
 /// and fold that shift into the literal's own written exponent -- returns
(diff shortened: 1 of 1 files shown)


newhoggy:
## Code review response

Multi-agent review completed (9 finder angles: line-by-line, removed-behavior, cross-file tracer, language-pitfall, wrapper/proxy correctness, Reuse, Simplification, Efficiency, Altitude, Conventions). Summary and disposition:

**Real bug found (during my own follow-up testing, before the reviewers even ran) and fixed in the same PR:** the initial "retry on `None`" version of the fix missed a more severe sibling — a capped mantissa can silently *succeed* in `format_positive_shifted_plain` with missing trailing digits (not just fail outright) when the decimal-point split falls *within* the truncated prefix rather than past it. Also reachable via two more call sites beyond the issue's own overflow-path repro: the ordinary (non-overflow) path for an everyday-magnitude value with an enormous fractional part, and `format_shifted_mantissa`'s unconditional `shifted_exp == 0`/`-6..=-1` window (which has no scientific fallback at all, so nothing there was checking digit count before this fix). Found via my own ~190-case combinatorial + randomized differential fuzz sweep against real jq 1.7.1 before handing off to review; confirmed 0 new mismatches after the fix, with all remaining divergences verified identical to unmodified `main` (an unrelated, pre-existing, already-documented scientific-notation precision cap).

**Fixed in response to review (commit `24a72fea5`):**
- **Off-by-one in `full_mantissa_if_capped`'s re-derive trigger** — 4 independent finder angles (Angle A, Angle E, Efficiency, Simplification) converged on the same discovery: the "was this capped" check fired at `digit_count > cap`, one boundary value before truncation actually starts (`digit_count > cap + 1`). Not a correctness bug (output was byte-identical either way, confirmed live by multiple finders) — a wasted full re-derivation at exactly `digit_count == cap + 1`. Fixed to match the real boundary.
- **Duplicated gate+fetch+render sequence** — 3 independent finder angles (Reuse, Simplification, Altitude) flagged the same ~15-line-of-duplicated-prose pattern repeated between `format_number_jq_compat` and `format_overflow_literal_mantissa`. Extracted into a shared `try_positive_shifted_plain` helper.
- **Stale doc comment** — Angle B found two sentences (leading-zero-trim cost, `s.parse::<f64>()` already scanning `s` once) dropped when `MAX_RENDERED_MANTISSA_DIGITS`'s doc comment moved to module scope. Restored.
- **"Safe by accident" 4th call site** — Altitude found `format_near_zero_literal` shares the same `normalize_extreme_literal_mantissa`-with-a-cap pattern and is only safe today because neither of its match arms has a plain-decimal branch. Added a defensive doc note warning a future plain-decimal branch there must go through `full_mantissa_if_capped`, not reuse the capped `mantissa_str` directly.

**Investigated, determined intentional/accepted, not changed:**
- **Unbounded cost for a plain-eligible literal with millions of digits** (Angle D): confirmed this is inherent to correctness, not a new amplification vector — real jq itself pays the same unbounded cost (oracle-verified past 500,000 digits, no ceiling found), the cost is linear in the attacker's own already-paid-for input size (not amplification), and the doc comment on `MAX_RENDERED_MANTISSA_DIGITS` now states this reasoning explicitly.
- **`format_positive_shifted_plain`'s now-dead internal guards** (Simplification, Altitude): both live call sites already gate externally, making the internal `shifted_exp >= digit_count` and `split_pos > digits.len()` checks provably unreachable. Kept as documented defense-in-depth per this codebase's own convention for justified-dead code (rather than deleted), since a future caller violating the contract would otherwise silently mis-render with no signal at all.
- **`Option<usize>` vs `bool` for the cap parameter** (Simplification): only one other reviewer angle raised concerns about the `Option<usize>` API surface, and none flagged it as a real risk — the named-constant spelling (`Some(MAX_RENDERED_MANTISSA_DIGITS)`) is arguably more self-documenting at call sites than a bare boolean. Left as-is.
- **Newtype-enforced capped/uncapped distinction** (Altitude's most ambitious suggestion — a `FullMantissa` wrapper type instead of `Cow<str>`/`&str`): a genuine structural improvement, but a larger API change than this PR's scope warrants given the shared-helper extraction above already substantially closes the practical risk (a new call site now has one correct path to reach for, not three to independently re-derive). Not implemented here.

Re-ran fmt, both clippy invocations, the full test suite (one transient failure batch from `cargo`'s own package-cache file-lock contention under this machine's concurrent-session load — confirmed via the failing tests' own captured stdout, retried cleanly with 0 failures), `cargo check --no-default-features`, and `omni-dev coverage diff` (98.51% patch coverage, the one remaining uncovered line a provably-unreachable `unreachable!()` arm) after each round of fixes. Re-verified against real jq 1.7.1 with the full ~190-case sweep after the final consolidation — 0 new mismatches.

Issue #999, opened 2026-08-15
yq: --input-format json disables the M2 streaming fast path for every query, not just number-touching ones


newhoggy:
## Summary

PR #994 (fixing #978, JSON-sourced number literal spelling leaking through \`succinctly yq\`)
disables the M2 cursor-streaming fast path for the *entire* run whenever any input is JSON
(\`is_m2_streamable = can_use_m2_streaming(&program.expr) && !any_input_is_json\`), forcing
every \`--input-format json\` (or auto-detected \`.json\`) query through the slower DOM path —
even queries that never touch a number at all (e.g. \`.name\` on an all-string object, or
plain \`.[]\` iteration over strings).

The actual leak #978 fixed is narrowly located: \`stream_resolved_scalar_as_json\`'s
\`format_float_with_fraction\` (trailing-\`.0\` mis-formatting) and the YAML raw-span echo in
\`src/yaml/light.rs\`, both of which only mis-format \`Float\`/\`NumberLiteral\` scalars
specifically. Disabling M2 wholesale for all JSON input is the coarse fix; a narrower one —
teaching those two streamers to normalize a JSON-sourced number's formatting in place
(they already know their own output format) — would keep the M2 fast path live for the
non-number-touching majority of \`--input-format json\` queries.

Compounding this, \`parse_input\`'s JSON arm now does two full O(N) passes over the parsed
document on the DOM path: \`generic_to_owned\` materializes the tree once (allocating a
\`Box<str>\` literal-spelling for every number), then \`canonicalize_json_numbers\` walks the
entire tree a second time, rebuilding every \`Array\`/\`Object\` (including a fresh \`IndexMap\`
per object) even for subtrees with no numbers at all, only to discard the \`Box<str>\`
spelling it just paid to allocate.

## Impact

This project's own CLAUDE.md documents M2 streaming's value directly: "M2 streaming
(\`.[0]\`) is 2.7x faster than identity, with 3-4% of yq's memory," and the optimizations
philosophy notes "Eliminating phases beats optimizing them." For \`--input-format json\`
(or \`.json\`-extension) input, PR #994 trades that win away wholesale rather than fixing
the two specific mis-formatting sites.

## Suggested fix

Fix \`stream_resolved_scalar_as_json\` and the YAML light-streamer's raw-span echo to
normalize JSON-sourced number formatting directly, so \`is_m2_streamable\` no longer needs
[...]

PR #1348, opened 2026-08-21
perf(yq): fuse the JSON DOM path's two number-canonicalization passes

newhoggy:
## Summary

Addresses the remaining scope of #999 (the primary fix already shipped via #996/PR #1019 — M2 streaming is already live for \`--input-format json\` on current \`main\`; see my comment on #999 for the re-scoping).

\`parse_input\`'s JSON arm (\`yq_runner.rs\`) built the full \`OwnedValue\` tree once, materializing every number as \`NumberLiteral\` (allocating a \`Box<str>\` per number to hold a source spelling), then walked the *entire* resulting tree a second time (\`canonicalize_json_numbers\`, rebuilding every \`Array\`/\`Object\` along the way, even subtrees with no numbers underneath) just to strip that spelling back off — real \`yq\`'s own \`--input-format json\` path never preserves a JSON-sourced number's spelling regardless of whether the filter touches it (#978).

## Fix

Added \`to_owned_canonicalized_numbers\` to \`src/jq/eval_generic.rs\` — a genuinely separate sibling of \`to_owned_at_depth\` (not that function with an added flag: \`to_owned_at_depth\` is also called from this evaluator's own internal recursion — YAML explicit-tag resolution — which must keep preserving document number spelling unconditionally; threading a canonicalize flag through the shared function would risk a future internal call site silently inheriting it). The new function collapses every number straight to \`Int\`/\`Float\` in one pass instead of two.

Wired in via a new \`json_bytes_to_owned_value_canonicalized\` (mirroring the existing \`json_bytes_to_owned_value\`) in \`output.rs\`, and \`parse_input\`'s JSON arm now calls it directly. Deleted the now-dead \`canonicalize_json_numbers\`/\`canonicalize_json_numbers_at_depth\` and retargeted their depth-guard test to the new function.

## Benchmarking

The obvious identity-query benchmark (\`.\`) is misleading here — \`Identity\` is M2-streaming-eligible, so it bypasses \`parse_input\`'s DOM path (and this fix) entirely. Used \`length\` (not M2-eligible) to actually exercise the DOM path, interleaved A/B against the pre-fix binary:

| Input | Speedup (median) |
|---|---|
| 1MB generated JSON | 1.11x |
| 10MB generated JSON | 1.13x |
| ~4MB number-dense JSON (50k objects, 4 numeric fields each) | 1.11x |

Output byte-identical to the pre-fix binary in every case (gated before trusting any timing, per this repo's benchmarking discipline).

## Verification

- Oracle-probed real \`yq\` for the basic canonicalization behavior (unaffected by this refactor — same output, just computed in one pass).
- \`succinctly jq\` (preserves spelling) and \`--argjson\`/\`--jsonargs\` (also preserves spelling) confirmed unaffected — they use the original \`json_bytes_to_owned_value\`, untouched.
- Live-verified an interesting edge case during review: the semi-index scanner accepts a wider set of number spans than strict RFC 8259 validation (\`is_valid_number\`, #966) — e.g. \`5.\`/\`.5\` (trailing/leading dot, no digit on one side). These fall through to an \`as_f64\` fallback arm rather than being unreachable; added a dedicated test after confirming live against the CLI. Cross-referenced the codebase's own \`is_valid_number_rejects_lenient_semi_index_spans\` test to confirm the remaining two structurally-mirrored arms (an \`as_i64\` fallback and an internal parse-failure fallback) are genuinely unreachable for JSON specifically — documented inline rather than chasing artificial coverage.

## Test plan

- [x] \`cargo test --features cli,simd,regex,serde --no-fail-fast\` — all green (one transient failure from unrelated \`cargo\` package-cache contention under this machine's concurrent-session load, confirmed unrelated to this change, retried cleanly)
- [x] \`cargo check --no-default-features\` (no_std) — compiles (pre-existing unrelated dead-code warnings only)
- [x] \`omni-dev coverage diff\` — 94.12% patch coverage (remaining gaps are the two documented-unreachable arms plus test-code panic branches)
- [x] Interleaved A/B benchmark (1MB/10MB/number-dense, non-M2-eligible query), output-identity gated before trusting timing


newhoggy:
## Code review response

Multi-agent review completed (10 finder angles: line-by-line, removed-behavior, cross-file tracer, language-pitfall, wrapper/proxy correctness, Reuse, Simplification, Efficiency, Altitude, Conventions). Zero correctness bugs found across all 10 angles — the review converged on one significant architectural finding and several documentation/consistency issues.

**Significant finding, fixed (commit `c65c99d2a`) — architectural placement:** Angle Altitude found (independently corroborated by Angle C, B, and Reuse) that `to_owned_canonicalized_numbers` was `pub` and generic over *any* `DocumentValue`, including YAML's — nothing in the type system stopped a future YAML or `succinctly jq` call site from reaching for it and silently regressing #918's number-spelling-preservation guarantee, the exact bug class #978/#999 exist to keep away from JSON input specifically. The pre-existing code (`canonicalize_json_numbers`) deliberately kept this yq-CLI-specific oracle quirk local to `yq_runner.rs`, per its own doc comment — my initial fix moved the substantive logic into the shared library to get single-pass fusion, reversing that architectural decision without acknowledging it.

Restructured: the walker (renamed `to_owned_canonicalizing_numbers`) now lives in `yq_runner.rs` again, matching the original architecture. It calls one small new `pub` library primitive, `OwnedValue::from_number_literal_plain`, for the number-parsing step instead of duplicating the parse-to-scalar mapping inline — this also let that primitive absorb an identical inline match that was already duplicated in `from_number_bytes`'s own fallback arm (Angle Reuse's finding #1). While moving the walker I also caught a real guard mismatch I'd have otherwise shipped: the fused single-pass walker needs `to_owned_at_depth`'s own `MAX_NESTING_DEPTH` (256) guard, not `MAX_VALUE_TREE_DEPTH` (384) — reusing the looser guard (which the old two-pass code only used for its *second* pass, after the first pass's 256 guard had already bound the real ceiling) would have quietly raised the effective depth limit for `--input-format json` documents from 256 to 384.

**Other fixes (same commit unless noted):**
- 5 stale doc-comment references to the deleted `canonicalize_json_numbers` name, across `yq_runner.rs` and `yaml/light.rs` (Angle C found 3, Angle B/Reuse independently found the same 5 total).
- Restored an explanatory comment on the `as_i64` fallback arm's unreachability that got dropped during the file move (commit `74e73cae7`).

**Investigated, determined not worth changing:**
- Container-recursion duplication between the walker and `to_owned_at_depth` (Altitude's findings #1-3): Angle Reuse explicitly declined to flag this, noting the codebase already has ~5 independently hand-copied `DocumentValue` walkers before this PR (`to_owned_cursor_at_depth`, `to_owned_with_comments_at_depth`, etc.) with no existing shared-walker abstraction to reuse — this PR's shape matches established (if imperfect) precedent, and building a generic closure-parameterized walker is a larger, separate refactor outside this PR's scope.
- `Option<usize>`/`bool`-shaped API concerns and internal dead-arm duplication: reviewed, judged proportionate as documented rather than restructured further.
- `docs/log.md` knowledge-wiki entry (Angle Conventions): this is a correctness/architecture PR, not a new SIMD/algorithmic optimization technique in the sense that skill's "New optimization" trigger targets (no new technique, just eliminating a redundant pass) — judged not to warrant a wiki entry, but noted for visibility.

Re-ran fmt, both clippy invocations, the full test suite, `cargo check --no-default-features`, and `omni-dev coverage diff` (94.87% patch coverage; `value.rs` 100%; the remaining `yq_runner.rs` gaps are the one documented-unreachable arm plus standard test-assertion boilerplate) after every round of changes. Re-verified output identity and the ~1.11-1.14x benchmark win against the pre-fix binary after the full restructuring — both held.

newhoggy:
Fixed two CI issues found on this PR's own checks (not part of the multi-agent review, caught by CI directly):
- **Commit Message Lint**: `yq` isn't in `.omni-dev/scopes.yaml`'s canonical scope list — reworded the 3 affected commits to use `cli` instead (the walker ended up living in `src/bin/succinctly/yq_runner.rs`, matching that scope's file pattern).
- **Documentation**: `from_number_literal_plain`'s doc comment used an intra-doc link (`[text](Self::path)`) to `from_number_literal`, which is `pub(crate)` — rustdoc rejects a public item's doc linking to a private one. Switched to plain text.

Issue #1356, opened 2026-08-21
yq: tonumber drops non-JSON spellings (+2.0, 007, 2., .5) that real yq echoes verbatim


newhoggy:
**Severity: Low**

## Summary

#1090 made `tonumber` preserve a JSON-shaped number's source spelling. Spellings YAML/yq accept but strict RFC 8259 rejects still fall through to a plain `Int`/`Float` and lose their text.

```
$ printf 'a: "+2.0"\n' | yq '.b = (.a|tonumber)'    # real yq
a: "+2.0"
b: +2.0
$ printf 'a: "+2.0"\n' | succinctly yq '.b = (.a|tonumber)'
a: "+2.0"
b: 2.0
```

Full set (`succinctly yq` vs yq v4.53.3):

| input | real yq | succinctly |
|---|---|---|
| `"+2.0"` | `+2.0` | `2.0` |
| `"007"`  | `007`  | `7` |
| `"2."`   | `2.`   | `2.0` |
| `".5"`   | `.5`   | `0.5` |

jq mode is **not** affected — real jq 1.7.1 normalizes all four the same way succinctly does (`7`, `0.5`, `2.0`), and those cases match today.

## Why it is not simply fixed

A `NumberLiteral`'s stored text is re-read *as JSON* by `to_json_for_reindex`'s reindex bridge, so a non-JSON-safe spelling cannot be stored verbatim without corrupting that round trip. `src/yaml/scalar.rs`'s `preservable_float_literal_text` documents this exact constraint for the YAML-input side, and normalizes (`+1.0` -> `1.0`, `1.` -> `1.0`, `007e2` -> `7e2`) rather than preserving.

`tonumber` already applies the one normalization that has an exact JSON equivalent — stripping a leading `+` — because real jq wants `2.0` there. The remaining three have no JSON-safe equivalent that keeps the original bytes.

This is the same permanent-divergence class `preservable_float_literal_text` already records for `-o yaml`/`tostring`/`join` on document-sourced `+1.0`/`1.`/`007e2`. Fixing it properly means the stored spelling and the bridge's JSON text stop being the same string.

## Suggested direction

Either accept and document alongside the existing `preservable_float_literal_text` divergence note, or give `NumberLiteral` a separate "display text" distinct from its "bridge text". The latter is a real design change and should be weighed against the rest of the number-fidelity cluster (#1076, #1129, #1144).

newhoggy:
[...]

Issue #1355, opened 2026-08-21
yaml: resolve_plain types `-0` as !!int where real yq says !!float


newhoggy:
**Severity: Low**

## Summary

`resolve_plain("-0")` classifies `-0` as `!!int`; real yq v4.53.3 classifies it as `!!float`.

```
$ printf 'a: -0\n' | yq '.a | tag'
!!float
$ printf 'a: -0\n' | succinctly yq '.a | tag'
!!int

$ printf 'a: -0\n' | yq -o=json '.'
{"a":-0.0}
$ printf 'a: -0\n' | succinctly yq -o=json '.'
{"a":0}
```

Note succinctly also loses the sign entirely on JSON output (`0`, not `-0.0`).

## Why it matters now

#1090 made `format_float_yq_yaml_nested` decide whether to emit an explicit `!!float ` tag by asking `needs_explicit_float_tag`, which is defined as "`resolve_plain` does not say `Float`". That anchoring is deliberate — it guarantees this crate's emitter and reader agree — but it means `resolve_plain`'s `-0` classification is now directly observable in output:

```
$ printf 'a: 1.0\n' | yq '.a = (.a * -0.0)'
a: -0
$ printf 'a: 1.0\n' | succinctly yq '.a = (.a * -0.0)'
a: !!float -0
```

This is the **one** value in a 41-value oracle sweep where #1090 diverges from real yq. It is the type-safe side of the divergence (succinctly's own reader would call a bare `-0` an int, so tagging keeps the round trip honest), and it was accepted knowingly rather than papered over with a hard-coded `-0` special case in the emitter.

## Fix

Make `resolve_plain` classify `-0` (and presumably `-0.0`-equivalent integer-shaped negative zero spellings) as `!!float`, matching go-yaml. Once that lands, `format_float_yq_yaml_nested` becomes byte-identical to real yq **with no change to #1090's code at all** — that is the whole point of defining the predicate in terms of the reader.

Guard tests to keep green:
- `src/yaml/light.rs`: `test_format_float_yq_yaml_nested_1090` (asserts `!!float -0` today — this expectation flips)
- `src/yaml/light.rs`: `test_nested_float_spelling_round_trips_as_float_1090` (must stay green either way — it asserts the invariant, not the spelling)

## Scope check

Changing `resolve_plain` touches the resolver every YAML read goes through, so the blast radius is wider than the emitter. Worth checking what else `-0` flows into (`-o json` sign loss above suggests at least one more bug in the same area).

[...]

Issue #1144, opened 2026-08-17
yq: join's array-constructed/computed elements still lose the decimal-point distinction via the reindex bridge


newhoggy:
## Summary

Found while implementing #1124 (which fixed \`yq_join_element_part\`/\`yq_join_separator\`'s catch-all to use \`numeric_display_string\` instead of \`to_json_yq()\` for computed floats). The separator fix is confirmed live-working:

\`\`\`
$ succinctly yq '[1,2] | join(2.0 / 2)'
112
\`\`\`

But the equivalent **array-element** case still diverges from real yq:

\`\`\`
$ succinctly yq '(2.0 / 2) | [.] | join(",")'
1.0
$ yq '(2.0 / 2) | [.] | join(",")'    # real yq v4.53.3
1
\`\`\`

## Root cause

\`builtin_join\`'s array-input branch only accepts a cursor-backed \`StandardJson::Array\`, never a plain owned \`Vec<OwnedValue>\`. An array built by construction (\`[...]\`) or by \`map(...)\` starts as an owned \`OwnedValue::Array\`, so reaching \`join\` requires round-tripping through \`to_json_for_reindex\` + reparse to become a cursor first. That round-trip serializes a computed \`Float\` with its decimal point preserved (the same #953 rule that keeps \`Float\`/\`Int\` distinct across a real structural round-trip, e.g. \`-o json\`), and reparsing captures the result as an \`OwnedValue::NumberLiteral\` whose \"source text\" is really just the reindex's own synthesized spelling (\"1.0\"), not anything the user typed.

By the time \`yq_join_element_part\` sees this value, it's indistinguishable from a genuine document-sourced \`1.0\` literal — and \`numeric_display_string\`'s \`NumberLiteral\` branch correctly, deliberately echoes a *document* literal's exact spelling verbatim (#1008), which is exactly why #1124's fix (routing through \`numeric_display_string\`) doesn't help here: both \`to_json_yq()\` and \`numeric_display_string\` agree on \"1.0\" for this synthesized \`NumberLiteral\`.

This is a structurally different problem from #1124's own fix, which only helps for genuinely-still-\`OwnedValue::Float\` values (verified via \`join\`'s separator expression, which evaluates directly to an owned value with no cursor round-trip in between).

## Why this is hard

[...]

PR #1357, opened 2026-08-21
fix(cli): emit an explicit !!float tag on a nested computed whole float

### LINKED ISSUE #1090: yq: nested computed whole floats have no !!float tag-emission path, unlike real yq

### LINKED ISSUE #1176: yq: -i (in-place) mode loses an explicit !!float tag on a whole-number-looking scalar

newhoggy:
Closes #1090. Also materially fixes #1176.

## Summary

Real yq keeps a computed float's type unambiguous by tagging it at any nested position, dropping the decimal point only at document-root scalar position (where it suppresses *every* tag):

```
$ printf 'a: 1.0\n' | yq '.a += 1'     # real yq v4.53.3
a: !!float 2
$ printf 'a: 1.0\n' | succinctly yq '.a += 1'   # before
a: 2.0                                  # type-correct, not byte-identical
```

## The issue's own plan was wrong, and this supersedes it

#1090's 2026-08-20 comment concluded — from PR #1179's failure — that yq's rule depends on **how a value was produced** (arithmetic -> tagged, `tonumber` -> untagged), that `OwnedValue::Float` carries no such provenance, and that an **ADR-0018 + ~183-site `NumberSource`/`NumberOrigin` widening must land first**, blocking nine issues.

Widening the oracle probe set by one input collapsed that theory:

```
$ printf 'a: "2e0"\n' | yq '.b = (.a|tonumber)'
b: 2e0        # the EXACT source text, not 2.0
```

Also `"2.50"`->`2.50`, `"1.500"`->`1.500`, `"1e3"`->`1e3`, `"007"`->`007`. yq is not tracking provenance — it is **preserving the source spelling**, and `(.a|tonumber) + 0` *is* tagged because arithmetic replaces that spelling with Go's `%v` output.

**The real rule is a pure function of (node tag, node text): tag iff re-resolving the text about to be printed would not yield `!!float`.** Verified over a 41-value sweep — `1`, `0`, `-1`, `1000`, `100000` tagged; `-0`, `0.5`, `2.5`, `1e+06`, `1e-05`, `1e+10`, `1e+300` not. `1.0/0` -> `!!float +Inf`, tagged because `+Inf` resolves as `!!str`.

succinctly already had both halves of that rule: the `Float`-vs-`NumberLiteral(repr, text)` split, and `into_plain_number()` — which every arithmetic op already calls — dropping the spelling on exactly the operations that make yq start tagging. No new type, no wide refactor. `format_float_yq_yaml`'s spelling already equalled yq's on 26/26 swept values, so only the tag prefix is new.

Two further corrections posted on the issue: `-0` is not an exception (YAML resolves it as `!!float`), and the "619 `GenericResult::` sites" precedent the plan cites does not appear in ADR-0017 at all.

## Commits

1. **`fix(yaml): keep a tag-forced float's type across the JSON reindex bridge`** — the #1176 gate. Root-caused to the *bridge*, not the reader: `--slurp '.[0]'` on `a: !!float 2` gave `a: 2` with no `-i` involved. A scalar that is only a float because a tag forced it has an integer-shaped mantissa, so it falls through to a bare `Float`, which `to_json_for_reindex`'s jq-mode fallback serializes as `2` — reparsing as an `Int`. Fixed read-side by giving it a float-shaped literal. Deliberately **not** by flipping the bridge to `YqSemantics`, which re-breaks #978 (and would change JSON-sourced `1e20`/`1e100` output).

   Also adds `needs_explicit_float_tag`, defined as `!matches!(resolve_plain(text), Float(_))` — the reader's own answer, so emitter and reader cannot drift (CLAUDE.md's #106 lesson), and whatever this crate writes it reads back at the same type.

2. **`fix(jq): preserve tonumber's source spelling instead of collapsing it`** — required; without it this reproduces PR #1179's exact regression. Real jq 1.7.1 preserves the spelling too, so succinctly was wrong in **both** modes (`"2.50"`->`2.5`, `"1e3"`->`1000`). Gated on strict RFC 8259 syntax and explicitly *not* on `from_number_bytes`, which decodes the internal `9e999e999` NaN sentinel that jq rejects outright.

3. **`fix(yq): emit an explicit !!float tag on a nested computed whole float`** — the two YAML emitters. `-o json`, `@yaml`, `@props`, `tostring` and the reindex bridge are untouched; a tag is not valid in any of them.

## Verification

Oracle matrix against real yq v4.53.3 (script checked in under `.ai/scratch/`, gitignored): **44 ok, 2 diff**.

Both diffs are known and accepted:
- `-0.0` -> `!!float -0` vs yq's `-0`. The one value in the sweep where the reader-anchored predicate diverges; it is the type-safe side, since succinctly's own reader calls a bare `-0` an int. Filed as #1355 — fixing `resolve_plain` there makes this byte-identical **with no change to this PR's code**.
- `join` on an array-wrapped computed float. Pre-existing and already documented in-tree as `test_yq_array_wrapped_computed_float_keeps_scientific_notation_known_gap_1168`.

Every gate transcribed from `.github/workflows/ci.yml` and run sequentially: full suite plus the `simd`, `portable-popcount`, `simd,scalar-yaml` and `SUCCINCTLY_SIMD=sse2` legs, all three clippy invocations, `fmt`, `no_std`, rustdoc. All pass. Re-run in full after a concurrent process rebased this branch onto a main that had landed overlapping number-canonicalization work (#999); #978's behaviour confirmed intact (`{"a":1e2}` still renders `100`).

## Follow-ups filed

- #1355 — `resolve_plain` types `-0` as `!!int` where yq says `!!float`
- #1356 — `tonumber` drops non-JSON spellings (`+2.0`, `007`, `2.`, `.5`)

Provenance may still be genuinely required for #1144/#1168 — `yq_float_fidelity_fixup`'s own doc comment makes an independent argument that it cannot distinguish document-sourced from computed floats after the fact. That is untouched here; only #1090's provenance claim is retracted.

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -176,6 +176,13 @@ impl OutputConfig {
 /// number 5, matching real `yq` (#224). Every recursive call passes a
 /// cursor too (`field.value_cursor()`, `YamlElements::uncons_cursor`), so a
 /// tag on a nested element is never lost.
+///
+/// Materializes through `to_owned_value_for_json_bridge`, not the plain
+/// `to_owned_value`: everything this function builds is headed for
+/// [`evaluate_input`]'s `to_json_for_reindex::<JqSemantics>` round trip,
+/// the one bridge that would otherwise flatten a tag-forced `!!float 2`
+/// to an `Int` (#1176). That variant's doc comment explains why no other
+/// `ResolvedScalar -> OwnedValue` caller wants the same treatment.
 fn yaml_to_owned_value<W: AsRef<[u64]>>(cursor: YamlCursor<'_, W>) -> Result<OwnedValue> {
     match cursor.value() {
         YamlValue::String(s) => {
@@ -185,7 +192,7 @@ fn yaml_to_owned_value<W: AsRef<[u64]>>(cursor: YamlCursor<'_, W>) -> Result<Own
 
             if let Some(explicit) = cursor.explicit_tag() {
                 if let Some(resolved) = resolve_tagged(&str_value, explicit) {
-                    return Ok(resolved.to_owned_value(str_value));
+                    return Ok(resolved.to_owned_value_for_json_bridge(str_value));
                 }
             }
 
@@ -196,7 +203,7 @@ fn yaml_to_owned_value<W: AsRef<[u64]>>(cursor: YamlCursor<'_, W>) -> Result<Own
             }
 
             // Resolve plain scalars per the YAML 1.2 core schema
-            Ok(resolve_plain(&str_value).to_owned_value(str_value))
+            Ok(resolve_plain(&str_value).to_owned_value_for_json_bridge(str_value))
         }
         YamlValue::Mapping(fields) => {
             let mut map = IndexMap::new();
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -22,15 +22,15 @@ use succinctly::jq::{
 };
 use succinctly::json::JsonIndex;
 use succinctly::yaml::{
-    format_float_yq_yaml, resolve_plain, resolve_tagged, stream_yaml_sequence, YamlCursor,
-    YamlIndex, YamlValue,
+    format_float_yq_yaml, format_float_yq_yaml_nested, resolve_plain, resolve_tagged,
+    stream_yaml_sequence, YamlCursor, YamlIndex, YamlValue,
 };
 
 use super::{FrontMatterMode, InputFormat, OutputFormat, YqCommand};
 use crate::front_matter;
 use crate::output::{
-    self, exit_codes, format_float_yq, ColorScheme, ControlEscape, DiagStyle, ErrorSink,
-    FloatStyle, InputLocation, JsonFormatOpts,
+    self, exit_codes, ColorScheme, ControlEscape, DiagStyle, ErrorSink, FloatStyle, InputLocation,
+    JsonFormatOpts,
 };
 
 /// yq's diagnostics carry no `(at <file>:<line>)` marker, so the yq paths have
@@ -1618,19 +1618,14 @@ fn emit_yaml_value_at_depth(
             nonfinite_display_string::<YqSemantics>(*f).to_string()
         }
         // Real yq drops a computed whole float's decimal point (`2`, not
-        // `2.0`) only at document-root scalar position -- issue #949.
-        // Anywhere nested (an object field, an array element, an `-i`
-        // in-place edit) it instead emits an explicit `!!float` tag to
-        // keep the value's type unambiguous on reparse (`a: !!float 2`);
-        // succinctly has no tag-emission mechanism to match that, so
-        // falling back to `format_float_yq`'s decimal-preserving spelling
-        // here is the safer of the two available choices -- it's not
-        // byte-identical to real yq, but unlike the bare, untagged `2`
-        // this fix originally used at every depth, it doesn't silently
-        // change the value's inferred type from float to int on
-        // reparse/round-trip through `-i`.
+        // `2.0`) only at document-root scalar position, where it suppresses
+        // every tag (issue #949; `echo '!!str 5' | yq '.'` prints a bare
+        // `5` too). Anywhere nested -- an object field, an array element,
+        // an `-i` in-place edit -- it keeps the same shortest spelling but
+        // precedes it with an explicit `!!float` tag whenever that spelling
+        // would read back as an int (`a: !!float 2`, issue #1090).
         OwnedValue::Float(f) if depth == 0 => format_float_yq_yaml(*f),
-        OwnedValue::Float(f) => format_float_yq(*f),
+        OwnedValue::Float(f) => format_float_yq_yaml_nested(*f),
         OwnedValue::NumberLiteral(NumberRepr::Float(f), _) if f.is_nan() || f.is_infinite() => {
             nonfinite_display_string::<YqSemantics>(*f).to_string()
         }
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -6949,8 +6949,16 @@ pub(super) fn tonumber_from_str(s: &str) -> Result<OwnedValue, EvalError> {
     // `NumberLiteral`'s text is re-read as JSON by the reindex bridge --
     // the same constraint `preservable_float_literal_text` documents on the
     // YAML side.)
+    //
+    // The digit check is load-bearing, not defensive: `is_valid_number`
+    // accepts a leading `-` of its own, so a doubled sign (`+-1`) would
+    // otherwise strip to a perfectly valid `-1` and silently succeed where
+    // both oracles error out (jq 1.7.1: "Invalid numeric literal"; yq
+    // 4.53.3: "cannot convert node value [+-1] ... to number").
     if let Some(unsigned) = trimmed.strip_prefix('+') {
-        if crate::json::validate::is_valid_number(unsigned.as_bytes()) {
+        if unsigned.starts_with(|c: char| c.is_ascii_digit())
+            && crate::json::validate::is_valid_number(unsigned.as_bytes())
+        {
             return Ok(OwnedValue::from_number_literal(unsigned));
         }
     }
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -6905,11 +6905,62 @@ fn builtin_tonumber<W: Clone + AsRef<[u64]>>(
 pub(super) fn tonumber_from_str(s: &str) -> Result<OwnedValue, EvalError> {
     // jq's JSON parser skips surrounding whitespace, so `" 1 "` is 1.
     let trimmed = s.trim();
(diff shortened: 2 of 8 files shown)


newhoggy:
Reviewed the branch against jq 1.7.1 and yq v4.53.3, three-way A/B (oracle vs merge-base vs branch) over ~150 probes. The retraction in #1090's comment holds up — I re-derived the (tag, text) rule independently and it is exactly right, and the `tonumber` commit is the strongest part: every behaviour change in jq mode moves toward the oracle, none away, and it fixes a class of bug (`"1e3" | tonumber` → `1000`) nobody had filed.

Pushed three commits for two things the matrix didn't cover, plus a stale comment.

**1. `"+-1" | tonumber` returned `-1`** (`15b51ec2b`). The leading-`+` retry stripped the sign and re-checked with `is_valid_number`, which accepts a leading `-` of its own — so a doubled sign stripped to a valid `-1` and silently succeeded. Both oracles error, and so did this crate before the retry existed. Now gated on an ASCII digit after the sign; all the accept cases (`+2.0`, `+1e3`, `+007`, `+0`) are unchanged.

**2. The tag-forced-float re-spelling leaked into string builtins** (`a146a7afa`). Putting it in the shared `to_owned_value` gave it to all three callers, including `eval_generic`'s cursor materialization — the one string-producing builtins read, where real yq prints the scalar's own text:

| filter on `a: !!float 2` | real yq | before this branch | branch as pushed |
|---|---|---|---|
| `.a \| tostring` | `2` | `2` | `2.0` |
| `.a \| @yaml` | `2` | `2` | `2.0` |
| `.a \| tostring \| length` | `1` | `1` | `3` |

Split into `to_owned_value_for_json_bridge`, given to `yq_runner.rs`'s `yaml_to_owned_value` alone. That is the actual boundary rather than a heuristic: `evaluate_input` is the only bridge that hardcodes `to_json_for_reindex::<JqSemantics>` (deliberately — `YqSemantics` there re-breaks #978), and jq's plain-`Float` fallback is the one that drops the point. Every other bridge is `S`-gated and already preserves it, so `load()` and the cursor path never needed the re-spelling. `--slurp '.[0].a | tag'` still answers `!!float` and `-i '.a += 1'` twice still gives `!!float 2` then `!!float 3`; the three string builtins are pinned by a new test.

**3.** `format_float_yq_yaml`'s doc still said the emitters have no tag-emission mechanism and that nested falls back to `format_float_yq` "instead of calling this function" — with `format_float_yq_yaml_nested` thirty lines below doing exactly that (`eb207e3c5`). The `#949` banner in `yq_cli_tests.rs` had the same staleness and is fixed in commit 2. Also narrowed `needs_explicit_float_tag` to `pub(crate)`; it had no caller outside the crate.

**Matrix.** Extended `.ai/scratch/float-tag-oracle-matrix.sh` with the rows whose absence let #2 through — a `!!float`-tagged *input* through `tostring`/`@yaml`/`length`/`@json`. Now **51 ok, 3 diff**: the two known ones (`-0` → #1355, array-wrapped `join` → `..._known_gap_1168`) plus a pre-existing `@yaml` trailing-newline difference that is cosmetic and predates this branch.

**Verification.** Full `cargo test`, CI's `--features cli` CLI/golden leg, both clippy invocations, `cargo fmt --check`, `cargo check --no-default-features`, `cargo doc --no-deps --all-features` — all green.

**Follow-ups filed.** #1358 — a computed float still loses the tag when a later stage (`sort`, `reverse`, `unique`, `map(.)`, `from_entries`) re-materializes it through the JSON bridge as a `NumberLiteral`. Not a regression (uniformly `2.0` before), but this PR closes the gap for only some pipelines, so the same value now has two spellings depending on an unrelated stage. Also noted on #1176 that it is materially fixed here.

Worth doing separately: the eight cluster issues (#1071, #1076, #1128, #1129, #1134, #1144, #1176, #1183) should be re-triaged against the measured (tag, text) rule rather than inheriting the provenance blocker they were parked behind.

Issue #1366, opened 2026-08-21
jq: an array pattern binding the same variable twice keeps the first match, where jq keeps the last


newhoggy:
**Severity: Low**

## Summary

Found during review of #1201 (`reduce`/`foreach`'s `as` clause accepting a full
destructuring pattern). Not caused by that change — pre-existing in `. as PATTERN`,
reproduces on `main` without it. #1201 only makes it reachable through a second
construct, since `reduce`/`foreach` now share the same binding machinery.

When one **array** pattern binds the same variable name twice, real jq keeps the
*last* match; succinctly keeps the *first*.

## Repro (verified against pinned jq 1.7.1)

```
$ echo '[1,2]' | jq -c '. as [$a,$a] | $a'
2

$ echo '[1,2]' | succinctly jq -c '. as [$a,$a] | $a'
1
```

**Object patterns are not affected** — both agree there, in either entry order:

```
$ echo '{"x":1,"y":2}' | jq -c '. as {x:$a,y:$a} | $a'   # 1
$ echo '{"x":1,"y":2}' | jq -c '. as {y:$a,x:$a} | $a'   # 2
```
succinctly returns `1` and `2` for those two, matching.

## Root cause

Binding is done by AST rewriting, not an environment: `try_pattern_alternatives`
(`src/jq/eval.rs`) folds `substitute_var` over the binding list returned by
`extract_pattern_bindings`, one call per binding, in order:

```rust
for (var_name, var_value) in &bindings {
    substituted_body = substitute_var(&substituted_body, var_name, var_value);
}
```

The first `substitute_var("a", 1)` replaces *every* remaining `Expr::Var("a")` node in
the body, so the second `substitute_var("a", 2)` finds nothing left to rewrite —
first-wins, structurally. jq resolves the name against a scope where the later binding
shadows the earlier one, hence last-wins.

#1201's `substitute_bindings` (`src/jq/eval.rs`) is the same fold, so `reduce`/`foreach`
inherit the identical behaviour:

```
$ echo '[[1,2]]' | jq -c 'reduce .[] as [$a,$a] (0; $a)'   # 2 in jq
```

## Scope note

Filing to make this trackable, not blocking — a pattern that binds one name twice is
a degenerate query with no real-world use, and the divergence is confined to array
patterns. A fix would need to dedupe the binding list before substituting (keeping the
last occurrence of each name), in the one place both call sites can share. Worth
checking at the same time whether object patterns' agreement is load-bearing or
[...]

Issue #1365, opened 2026-08-21
jq: reduce/foreach's 'as' clause doesn't accept ?// alternative patterns, unlike '. as PATTERN'


newhoggy:
**Severity: Low**

## Summary

Found while implementing #1201 (`reduce`/`foreach`'s `as` clause accepting a full
destructuring pattern). Not caused by that fix — #1201 deliberately scoped `?//`
out, and this issue is that scope note made trackable.

`reduce`/`foreach`'s binding clause calls the parser's `parse_pattern`, which reads
exactly one pattern. `. as PATTERN | …` instead calls `parse_as_pattern`, which
additionally collects `?//`-separated alternatives into `Expr::AsPattern`. Real jq
accepts `?//` in **both** positions, so `reduce`/`foreach` are missing it.

## Repro (verified against pinned jq 1.7.1)

```
$ echo '[[1],{"a":2}]' | jq -c 'reduce .[] as [$a] ?// {a:$a} (0; . + $a)'
3
$ echo '[[1],{"a":2}]' | succinctly jq -c 'reduce .[] as [$a] ?// {a:$a} (0; . + $a)'
jq: compile error: parse error at position 19: expected '(', found '?'

$ echo '[[1],{"a":2}]' | jq -c 'foreach .[] as [$a] ?// {a:$a} (0; . + $a; .)'
1
3
$ echo '[[1],{"a":2}]' | succinctly jq -c 'foreach .[] as [$a] ?// {a:$a} (0; . + $a; .)'
jq: compile error: parse error at position 20: expected '(', found '?'
```

Note the error is the generic `expected '('` from the clause's own `expect('(')`,
not a `?//`-specific message.

## Why #1201 didn't just do it

`?//` is not only "try the next pattern if this one fails to match". jq also retries
with the next alternative when the **body** errors, and in `reduce`/`foreach` the
"body" is UPDATE, which has already consumed the accumulator. jq rolls the
accumulator back to this element's pre-update value before retrying:

```
$ echo '[{"a":1}]' | jq -c 'reduce .[] as {a:$x} ?// {a:$y} (0; if $x then error("boom") else [$x,$y] end)'
[null,1]

$ echo '[{"a":1},{"a":2}]' | jq -c 'foreach .[] as {a:$x} ?// {a:$y} (0; if $x==1 then error("boom") else [.,$x,$y] end; .)'
[null,null,1]
[[null,null,1],2,null]
```

[...]

Issue #1240, opened 2026-08-18
jq: parse_pattern has no recursion-depth guard, deeply nested destructuring patterns crash the process with a stack overflow


newhoggy:
**Severity: Medium** (uncontrolled process crash reachable from query text alone, no large input document needed)

## Summary

Found during code review of #1204's `{\$IDENT: Pattern}` fix (PR #1237). Not caused by that PR — the underlying gap is pre-existing (`key: Pattern` and `[$a, [$b, ...]]` nesting were already unbounded before #1204), and #1204's new `\$IDENT: Pattern` branch simply adds a third recursive arm into the same unguarded recursion rather than introducing it.

`parse_pattern` (`src/jq/parser.rs`) recurses into itself for every nested `Object`/`Array` pattern level with no depth counter or limit. A query string alone — no input document required — with enough nesting overflows the process stack, aborting the whole process rather than returning a parse error.

## Repro (live-verified, this build vs. real jq 1.7.1)

```
$ echo '{}' | jq -c ". as $(python3 -c 'n=100000; print(\"{a: \"*n + \"\$x\" + \"}\"*n)') | \$x"
jq: error: memory exhausted at <top-level>, line 1: ...
(clean, controlled error -- real jq has its own resource limits)

$ echo '{}' | succinctly jq -c ". as $(python3 -c 'n=100000; print(\"{a: \"*n + \"\$x\" + \"}\"*n)') | \$x"
thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting
(uncontrolled process abort)
```

## Root cause

`Parser` (`src/jq/parser.rs`) has no depth-tracking field at all, unlike `OwnedValue` tree *materialization*, which already has an explicit, documented guard (`MAX_NESTING_DEPTH = 256` in `src/jq/eval_generic.rs`, with measured crash boundaries). No equivalent guard exists for the `Pattern` AST the parser builds, on either side:
- **Parse time**: `parse_pattern`'s `Object`/`Array` arms recurse into `self.parse_pattern()` with no counter (`src/jq/parser.rs`, ~line 1671 for the new `\$IDENT: Pattern` case, and the pre-existing `key: Pattern`/array-element cases nearby).
- **Eval time**: even if a deeply-nested pattern were somehow parsed without overflowing, `extract_pattern_bindings`, `collect_pattern_var_names`, and `pattern_binds_var` (`src/jq/eval.rs`) would hit the identical unbounded-recursion risk evaluating it.

## Scope note

[...]

Issue #494, opened 2026-07-30
jq: limit/while/foreach silently drop already-produced output when a later branch errors or breaks


newhoggy:
**Severity: High** — reproduced against the built CLI (release), on `main`. Silent data loss: values jq would have already streamed out are discarded, not just the error itself.

Found while investigating #464 (`nth(n; expr)` swallowing errors) — this is a related but architecturally deeper and more severe sibling bug.

```
$ echo 1 | jq -c 'while(. < 5; if . == 3 then error("boom") else .+1 end)'
1
2
3
jq: error (at <stdin>:1): boom
# exit code 5

$ echo 1 | succinctly jq -c 'while(. < 5; if . == 3 then error("boom") else .+1 end)'
# (no "1", "2", "3" at all)
jq: error: boom
# exit code 0 (separately tracked in #355)
```

Same shape with `foreach`:

```
$ echo null | jq -c '[1,2,3,4] | foreach .[] as $x (0; if $x == 3 then error("boom") else . + $x end)'
1
3
jq: error (at <stdin>:1): boom

$ echo null | succinctly jq -c '[1,2,3,4] | foreach .[] as $x (0; if $x == 3 then error("boom") else . + $x end)'
jq: error: boom
```

And with `limit`, where a `break` disappears entirely instead of terminating the enclosing `label`:

```
$ echo null | jq 'label $out | limit(5; 1,2,(3|break $out),4)'
1
2

$ echo null | succinctly jq 'label $out | limit(5; 1,2,(3|break $out),4)'
# (no output at all)
```

### Where

Root cause is `eval_comma` (`src/jq/eval.rs:569-617`), the shared evaluator for `,`. It is not lazy/streaming: it loops over each sibling sub-expression, accumulating outputs into local `Vec`s (`all_results`/`all_owned`), and only wraps them into a final `QueryResult::Many`/`ManyOwned`/`One`/`Owned` after *every* sibling has succeeded. If any sibling's `eval_single` call yields `Error` or `Break`, the function returns that variant immediately — discarding everything already accumulated in the `Vec`s.

`eval_limit` (8115-8166), `eval_while` (8306-8344), and `eval_foreach` (8052-8112) all consume their generator argument via a single `eval_single` call, so they only ever see the fully-collapsed (and by then already-stripped-of-partial-data) result. Any error or `break` partway through the underlying generator loses everything produced before it, rather than emitting it first the way jq's real streaming evaluator does.

[...]

PR #1367, opened 2026-08-21
feat(jq): reduce/foreach accept a full destructuring pattern, not just $var

### LINKED ISSUE #1201: jq: reduce/foreach's own 'as' clause only accepts $var, not a full destructuring pattern
**Severity: Low**

## Summary

Found while implementing #1139's `{$a}` object-pattern shorthand fix (PR TBD). Not caused by that fix — pre-existing and unrelated, reproduces identically with or without the shorthand.

`reduce`/`foreach`'s own `EXPR as PATTERN (...)` binding clause only accepts a bare `$var`, never a full destructuring pattern (object `{...}` or array `[...]`) — even though `. as PATTERN | ...` supports both, and real jq accepts a full pattern in `reduce`/`foreach` too.

```
$ echo '[{"a":1},{"a":2}]' | jq -c 'reduce .[] as {a: $a} (0; . + $a)'
3
$ echo '[{"a":1},{"a":2}]' | succinctly jq -c 'reduce .[] as {a: $a} (0; . + $a)'
jq: compile error: parse error at position 14: expected '$', found '{'

$ echo '[[1,2],[3,4]]' | jq -c 'reduce .[] as [$a,$b] (0; . + $a + $b)'
10
$ echo '[[1,2],[3,4]]' | succinctly jq -c 'reduce .[] as [$a,$b] (0; . + $a + $b)'
jq: compile error: parse error at position 14: expected '$', found '['
```

`reduce`/`foreach` with a plain `$var` binding already works correctly and is unaffected:

```
$ echo '[1,2,3]' | succinctly jq -c 'reduce .[] as $x (0; . + $x)'
6
```

## Scope note

Filing to make this trackable, not blocking. The parser's `reduce`/`foreach` clauses presumably hardcode `expect '$'` where they should instead call the shared `parse_pattern` (`src/jq/parser.rs`) — the same function `. as PATTERN` and function-argument patterns already use. Not investigated further; the evaluator side (`Expr::Reduce`/`Expr::Foreach` handling in `src/jq/eval.rs`) would also need checking for whether it can already bind an arbitrary `Pattern`, not just a variable name, since `. as PATTERN` presumably already does that binding logic somewhere reusable.

newhoggy:
Fixes #1201.

`reduce`/`foreach`'s own `EXPR as PATTERN (…)` clause hardcoded `expect('$')` plus an
identifier, so only a bare `$var` parsed — even though `. as PATTERN | …` and
function-argument patterns already went through the shared `parse_pattern`, and real jq
accepts a full pattern here too.

```
$ echo '[{"a":1},{"a":2}]' | succinctly jq -c 'reduce .[] as {a: $a} (0; . + $a)'
jq: compile error: parse error at position 14: expected '$', found '{'   # jq 1.7.1: 3
```

## What changed

`Expr::Reduce`/`Expr::Foreach` carry a `Pattern` in place of the old `String` (a bare
`$var` is `Pattern::Var`, so existing queries are untouched). The parser calls
`parse_pattern`, which also brings these two sites under #1240's `MAX_PATTERN_DEPTH`
guard for free. The evaluator destructures each element with `extract_pattern_bindings`
— the same primitive `. as PATTERN` uses — and folds `substitute_var` over the
bindings, extending the AST-rewrite mechanism these constructs already had from one
variable to N.

Unlike a bare `$var`, a pattern can fail to match a given element. Since `foreach`
emits one output per step, that failure has to surface only when the fold reaches the
offending element, leaving earlier steps' output in place — the contract #494
established for an ordinary per-step UPDATE error.

## Verification

19 differential cases against pinned jq 1.7.1 — object, array, nested, `{$a}`-shorthand
and bare-`$var` bindings; absent keys and past-the-end indices resolving to `null`;
outer-variable and `def`-parameter shadowing; both `foreach` arities; and the three
error shapes — matching stdout, stderr **and** exit code each time, on both the
document-input path and `-n`'s owned pipe. Nine are pinned as `jq-golden` cases
captured from the oracle by `sync-jq-golden.sh`, including the
streamed-prefix-then-exit-5 triple.

Locally green: `cargo fmt --check`, both CI clippy feature sets, full `cargo test`
(53 suites), the cli-gated suites, `cargo check --no-default-features`, and both
`sync-jq-*.sh --check` drift checks. Patch coverage **98.09%** (154/157); the three
uncovered lines are the deliberately-unreachable defensive arm described below.

## Review notes

- The parser comments claiming real jq accepts only a single pattern in this clause
  were **wrong**, and contradicted this commit's own `substitute_pattern` doc. jq does
  accept `?//` here, including body-error retry that rolls the accumulator back to the
  element's pre-UPDATE value — that rollback is the part the fold can't express, and is
  the actual reason for scoping it out. Corrected, and filed as #1365.
- `eval_foreach` now handles rather than `expect()`s the extract-side binding. The
  invariant does hold, but it rests on three parallel `Vec`s plus a `break` that nothing
  type-checks, on a path ordinary CLI queries reach.
- Review also surfaced #1366, a pre-existing array-pattern divergence unrelated to this
  change (`. as [$a,$a] | $a` → jq `2`, succinctly `1`).
- yq needs no oracle work: real yq has no `reduce`/`foreach` at all, so the yq-mode test
  pins succinctly's own extension and guards against the fix missing the other
  `ParserMode`.


newhoggy:
## Review follow-ups

Reviewed this branch independently against the pinned oracle (`jq 1.7.1`) rather than
re-reading the commit message's claims — ~50 differential probes across object / array /
nested / `{$a}`-shorthand / bare-`$var` bindings, absent keys and past-the-end indices,
`null` elements, all five destructuring error shapes, `?` and `try` suppression,
multi-INIT forks (#534), `limit`, `label`/`break`, both `foreach` arities, `def`-inlining
and `def($a)` shadowing, malformed patterns, and the `-n` owned-pipe path. Every one
matched jq on stdout, stderr **and** exit code. `MAX_PATTERN_DEPTH` does reach both new
construction sites, as #1201's tier note asked.

The only divergences I found are the three already-known ones, all pre-existing or
deliberately scoped out: `?//` (#1365), array patterns binding one name twice keeping
first-not-last (#1366), and `parse_pattern`'s tolerance of `{}`/`[]` plus its rejection of
computed keys `{(expr): $v}` — the last reproducing identically through `. as PATTERN` on
untouched code.

Nothing blocking. Three follow-ups filed, none behavioural:

- **#1368** — `substitute_bindings` (`src/jq/eval.rs`) is body-for-body the existing
  `pub fn substitute_vars`, which already fits the call sites as
  `substitute_vars(expr, bindings.iter().map(|(n, v)| (n.as_str(), v)))`. Both seed the
  fold with an `expr.clone()` that `substitute_var`'s freshly-rebuilt AST discards
  unread — which this PR moves onto the per-element path #695 hoisted. Structural
  observation, not a measured regression.
- **#1369** — the three parallel per-element `Vec`s in `eval_foreach` are what force the
  documented-unreachable `extracts[idx]` `Err` arm, and leave `bindings_per_element`
  (a full `OwnedValue` clone per element for `Pattern::Var`) alive for the rest of the
  call. One `Vec<Result<(Expr, Option<Expr>), EvalError>>` makes the invariant
  unrepresentable and drops each element's bindings as consumed. The
  eager-compute / lazy-surface contract, and the
  `foreach_as_pattern_error_after_output` golden that pins it, are unaffected.
- **#1370** — `eval_reduce`/`eval_foreach`'s doc comments still say `as $var` while their
  signatures take `&Pattern` (the `expr.rs` variants were updated).

All three are equally reasonable to fold in here before merge instead of tracking
separately — they only exist because of this branch, so they are filed mainly so they
survive the PR landing as-is.

Nit while amending, if you do: the commit body carries `Fixes #1201` twice.

Issue #486, opened 2026-07-30
jq: assignment does not create missing containers (.a.b = 9 on {} errors instead of building {"a":{"b":9}})


newhoggy:
**Severity: High** — `.a.b = 9` on a document lacking `.a` is one of the most common ways to build up JSON in jq, and it fails unconditionally. Reproduced against the built CLI (release, `a55d6c70`), jq oracle 1.7.1.

Found while testing #412; unrelated to it, and predates it.

An assignment whose path passes *through* a missing or null intermediate errors instead of creating the container the path implies:

```
$ echo '{}' | sjq -c '.a.b = 9'
jq: error: Cannot index null with string "b"      # jq: {"a":{"b":9}}

$ echo 'null' | sjq -c '.p = 9'
jq: error: Cannot index null with string "p"      # jq: {"p":9}

$ echo '{"a":null}' | sjq -c '.a.b.c = 9'
jq: error: Cannot index null with string "b"      # jq: {"a":{"b":{"c":9}}}

$ echo '{}' | sjq -c '.a[0] = 9'
jq: error: Cannot index null with number          # jq: {"a":[9]}
```

The same path assigned past the end of an array errors instead of padding:

```
$ echo '[1]' | sjq -c '.[3] = 9'
jq: error: index 3 out of bounds (length 1)       # jq: [1,null,null,9]

$ echo '{"a":[]}' | sjq -c '.a[0] = 9'
jq: error: index 0 out of bounds (length 0)       # jq: {"a":[9]}

$ echo 'null' | sjq -c '.[0] = 9'
jq: error: Cannot index null with number          # jq: [9]
```

Assignment into an intermediate that *already exists* with the right type works, which is why this has not been more visible: `{} | .p = 9` → `{"p":9}` and `{"a":{}} | .a.p = 9` → `{"a":{"p":9}}` are both correct.

All four write operators are affected — `=`, `|=`, `+=` (and the rest of the compound family), `//=`:

```
$ echo '{}' | sjq -c '.a.b |= 9'      # jq: {"a":{"b":9}}
$ echo '{}' | sjq -c '.a.b += 1'      # jq: {"a":{"b":1}}
$ echo '{}' | sjq -c '.a.b //= 9'     # jq: {"a":{"b":9}}
jq: error: Cannot index null with string "b"      (all three)
```

### Where

`set_path` and `get_path_mut` (`src/jq/eval.rs`). Both require the container to *already* be of the right type:

[...]

PR #1373, opened 2026-08-21
fix(jq): let setpath() reach through a mid-chain string slice

### LINKED ISSUE #1338: jq: setpath()'s set_value_at_path has the same mid-chain string-slice bug #1321 fixed in through_slice
**Severity: Low**

## Summary

Found during #1321/#1337's code review. `setpath()` is backed by a separate walker, `set_value_at_path` (`src/jq/eval.rs`, ~line 19610-19702), structurally parallel to but entirely distinct from `through_slice`/`set_path`/`update_path`/`delete_at_path`. It has the identical bug #1321 just fixed: `if matches!(value, OwnedValue::String(_)) { return Err(EvalError::cannot_update_string_slices()); }` fires unconditionally, ignoring whether `rest` (the remaining path components after the slice) is empty — so a mid-chain slice-then-more-path through a string reports the terminal "cannot update string slices" refusal instead of whatever the following component's own navigation would raise.

```
$ echo '"hello"' | jq -c 'setpath([{"start":0,"end":1},"b"]; 5)'
jq: error (at <stdin>:1): Cannot index string with string "b"
$ echo '"hello"' | succinctly jq -c 'setpath([{"start":0,"end":1},"b"]; 5)'
```

## Why this wasn't fixed in #1321/#1337

`set_value_at_path` is a separate function `through_slice`'s own #1321 fix never touches — `setpath()`'s own path-array-driven navigation (using `{"start":s,"end":e}` slice descriptors rather than parsed `Expr::Slice` AST nodes) is a structurally different walker built for a different calling convention (a runtime path value, not a compiled expression), so the fix doesn't carry over automatically. Confirmed the bug is real and isolated to this one function; not investigated further.

## Suggested fix direction

Not investigated beyond confirming the repro and locating the function. Likely needs the same kind of `terminal_write`-style distinction #1321 introduced for `through_slice` — check whether `rest` is empty (the slice descriptor is the last path component) versus non-empty (more path follows), and only refuse unconditionally in the terminal case; for the mid-chain case, read through a substring and let the continuation's own navigation raise its own error.

newhoggy:
## Summary

Fixes #1338. `set_value_at_path` (the walker backing \`setpath()\`) refused a slice-descriptor path component unconditionally whenever the target was a string:

\`\`\`rust
if matches!(value, OwnedValue::String(_)) {
    return Err(EvalError::cannot_update_string_slices());
}
\`\`\`

— regardless of whether the slice was the terminal path component or had more path after it. Real jq only refuses when the slice genuinely *is* the write:

\`\`\`bash
$ echo '"hello"' | jq -c 'setpath([{"start":0,"end":1},"b"]; 5)'
jq: error (at <stdin>:1): Cannot index string with string "b"
$ echo '"hello"' | succinctly jq -c 'setpath([{"start":0,"end":1},"b"]; 5)'   # before this PR
\`\`\`

## Fix

Mirrors #1321's identical \`terminal_write\` distinction in \`through_slice\`/\`.a[0:1].b = v\`: when the remaining path (\`rest\`) is empty, the slice is the terminal write and stays refused unchanged. When \`rest\` is non-empty, read through the sliced substring and forward the recursive \`set_value_at_path\` call's own result — that continuation's own navigation (indexing a string with a string/number key) is what actually fails, matching jq exactly.

A \`String\` value can never be successfully indexed by any key type in this function's own dispatch (every arm either narrows to \`Object\`/\`Array\`/\`Null\` or falls to a catch-all \`Err\`), so the recursive call always errors in practice — the fix forwards it directly via \`return\`, with no synthetic \`Ok\` case to design.

## Verification

Oracle-verified against real jq 1.7.1: the issue's own repro, a numeric-index variant, the unchanged terminal case, a nested slice-of-slice (still correctly refused — it's a terminal write relative to the read-through substring), and an array-target regression check (unaffected, routes through the pre-existing \`Array\` arm).

## Test plan

- [x] \`cargo test --features cli,simd,regex,serde --no-fail-fast\` — all green
- [x] \`cargo check --no-default-features\` (no_std) — compiles (pre-existing unrelated dead-code warnings only)
- [x] \`omni-dev coverage diff\` — 100% patch coverage

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -19976,27 +19976,51 @@ fn set_value_at_path(
             // jq reads a string slice but will not write one back -- but
             // that refusal is only correct once the slice itself *is* the
             // write. A chained path with more path after the slice
-            // (`rest` non-empty) needs the read-through substring handed
-            // to the recursive call instead, so *that* navigation's own
-            // outcome is the answer, not this refusal (#1338, mirroring
-            // #1321's identical `terminal_write` distinction in
-            // `through_slice` -- confirmed live:
+            // (`rest` non-empty) needs *that* continuation's own
+            // navigation to be the answer instead, not this refusal
+            // (#1338, mirroring #1321's identical `terminal_write`
+            // distinction in `through_slice` -- confirmed live:
             // `"hello" | setpath([{"start":0,"end":1},"b"]; 5)` matches
             // real jq's "Cannot index string with string \"b\"", not this
-            // refusal). A `String` value can never be indexed by any key
-            // type in this function's own dispatch -- every arm below
-            // either narrows to `Object`/`Array`/`Null` or falls to a
-            // catch-all `Err` -- so the recursive call always errors in
-            // practice; forwarded directly via `return`, with no `Ok` case
-            // to design a synthetic return value for.
-            if let OwnedValue::String(s) = &value {
+            // refusal).
+            //
+            // A `String` value can never be indexed by any key type in
+            // this function's own dispatch -- every arm below either
+            // narrows to `Object`/`Array`/`Null` or falls to a catch-all
+            // `Err` -- so *every* continuation of a string-sliced target
+            // ends in an error, fully determined by `rest`'s own
+            // structure alone. Found this out by first writing the
+            // obvious version -- read the actual substring, recurse into
+            // it, forward the result -- and having code review catch
+            // that it stack-overflows on a long chain of slice
+            // descriptors (recursion depth scales with `rest.len()`,
+            // unbounded) while also burning O(chain length x string
+            // length) time/memory materializing substrings nothing ever
+            // reads, since every one of them was destined to error
+            // anyway. This walks `rest` with a plain loop instead --
+            // O(remaining slice-descriptor count), no recursion, no
+            // substring ever built -- to find the exact same error the
+            // real read-through-and-recurse would have, without paying
+            // for it.
+            if matches!(value, OwnedValue::String(_)) {
                 if rest.is_empty() {
                     return Err(EvalError::cannot_update_string_slices());
                 }
-                let len = s.chars().count();
-                let range = bounds.resolve(len);
-                let sub = OwnedValue::String(slice::slice_str(s, range));
-                return set_value_at_path(sub, rest, new_val);
+                for next_key in rest {
+                    let OwnedValue::Object(desc) = next_key else {
+                        return Err(EvalError::cannot_index("string", next_key));
+                    };
+                    // Validates this key's own descriptor exactly as the
+                    // real recursive walk would have, one level deeper --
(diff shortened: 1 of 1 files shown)


newhoggy:
Pushed a follow-up commit addressing both review passes' findings.

**Fixed — critical: unbounded-recursion stack overflow + wasted work.** The original
fix's mid-chain string-slice arm read the sliced substring and recursed into it
(`set_value_at_path(sub, rest, new_val)`) to find the continuation's error. Recursion
depth scales with the number of chained slice descriptors in the path array — unbounded,
attacker-controlled via input path length. Both review passes independently reproduced a
real stack overflow (`thread 'main' has overflowed its stack`, exit 134) against this
PR's release binary, plus ~627MB peak RSS for wasted substring materialization at every
recursion level before the eventual error.

Replaced it with a non-recursive loop over `rest`: since a `String` value can never be
successfully indexed by any key type in this function's dispatch, the eventual error is
fully determined by `rest`'s own structure — no substring ever needs to be materialized,
no recursion needed. Confirmed against a 1,000,000-descriptor chain: completes in
~0.5s wall time, no crash, correct error, no growth in peak memory. Live-verified all
prior cases plus this one against real jq 1.7.1 — identical output to the (correct) parts
of the original fix. Added `test_setpath_string_slice_long_chain_does_not_overflow_the_stack_1338`
(200k-descriptor chain, must not crash) and
`test_setpath_agrees_with_assign_on_mid_chain_string_slice_error_1338` (locks the
`setpath()` error message to `.a[0:1].b = 5`'s, mirroring #486's
`test_assign_and_update_agree_with_setpath_on_autovivification` precedent).

**Filed as follow-up, not fixed here: yq-mode parity gap.** In yq mode,
`.a[0:1].b = 5` silently no-ops on a string target while the `setpath()` equivalent hard
-errors. Confirmed this is pre-existing on `origin/main` before this PR — this PR only
changed *which* error `setpath()` reports (a more precise message when more path follows
the slice), not whether one is reported in yq mode at all. `set_value_at_path` has no
`EvalSemantics`/mode parameter to thread a yq-mode no-op through, so this isn't a small
in-scope fix. Filed as #1380 with two possible directions, neither scoped/decided.

**Deferred, per the review's own framing: full `SliceEditFlags`/`through_slice`
unification.** `set_value_at_path` is a structurally separate recursive walker
(runtime path-array-driven) from `through_slice` (compiled-`Expr`-AST-driven) by
original design — converging them is exactly the refactor #1380 names as one possible
fix direction there, not something to fold into this already-landed bug fix.

All required checks re-run clean after rebasing onto current `main`: `cargo fmt --check`,
both `clippy --all-targets` invocations from CI, full `cargo test` (2692+ tests, 0
failures), `cargo check --no-default-features`, and `omni-dev coverage diff` (100% patch
coverage on the changed lines in `eval.rs`).

Issue #1301, opened 2026-08-20
jq: del() is O(n²) with a computed key ahead of a trailing iterate


newhoggy:
**Severity: Medium**

## Summary

`del(.items[(0,1)].foo[])` — a computed key ahead of a trailing bare iterate — is O(n²), the same shape #888 fixed for `path()`. `=` and `|=` on the identical shape are linear.

This was **not** visible in #888's own scope-correction table, which measured to 16,000 elements and reported all three write-path operations as flat. `del`'s constant is small enough that the quadratic term does not dominate until roughly 50,000 elements.

## Repro

```bash
python3 -c "
n = 100000
arr = ','.join(str(i) for i in range(n // 2))
print('{\"items\":[{\"foo\":[' + arr + ']},{\"foo\":[' + arr + ']}]}')
" > /tmp/big.json
time succinctly jq -c 'del(.items[(0,1)].foo[]) | length' /tmp/big.json
```

## Scaling curve

Apple M-series, release build, current `main` + #1295's gated fix (identical on this path). Total elements across both branches:

| total elements | `del`  | `\|=`  | `path` (post-#888) |
|----------------|--------|-------|--------------------|
| 20,000         | 0.10s  | —     | 0.03s              |
| 50,000         | 0.54s  | —     | 0.07s              |
| 100,000        | 1.51s  | —     | 0.10s              |
| 200,000        | 6.77s  | 0.33s | 0.20s              |
| 400,000        | 26.41s | 0.57s | 0.72s              |

`del` roughly quadruples per doubling above 100,000; `|=` and `path` do not.

> **Diagnosis corrected 2026-08-21 — the two sections that used to sit here named
> the wrong cause.** The measurements that disproved them are in
> [this comment](https://github.com/rust-works/succinctly/issues/1301#issuecomment-5370005674);
> their original text is archived verbatim in
> [this one](https://github.com/rust-works/succinctly/issues/1301#issuecomment-5370012754).

## Why #888's fix does not cover it

It does not need to. #888's `defer_trailing_iterate` deferral is *not* what made
`del` quadratic, and the four divergence families it is gated behind are not a
prerequisite for fixing this.

The decisive measurement: hold the resolved-path count fixed at 80,000 and widen
the computed key from `k = 2` branches to `k = 512`, so `resolve_dynamic_indexes`
emits exactly as many paths in every row while the sibling count within one
parent container falls from 40,000 to 156. `del` goes from 890ms to 95ms — which
[...]

PR #1378, opened 2026-08-21
fix(jq): remove del()'s discarded terminal-optional merge, verify the rest

### LINKED ISSUE #1331: jq: delete_expr_iterate_paths has the same dead optional-gate #1322 fixed, and DeleteStep.optional is dead end-to-end
`#1322`'s `/code-review` round found the exact same "optional never forced true in nested calls" pattern (see `#928`/`#1003`/`#1034`/`#1322`) recurring one function below the one it fixed, plus the underlying architectural cause all four occurrences share. Filing together since a real fix at the root would close all of them at once, rather than requiring a fifth per-site discovery.

## 1. `delete_expr_iterate_paths` has the identical dead-optional-gate pattern

`src/jq/eval.rs`, `delete_expr_iterate_paths` (around line 20261 as of `#1322`'s merge): `let optional = paths[0][start].optional;` feeds two `other if optional => Ok(other)` arms that, by `#1322`'s own proof (`optional` is always `false` past `assemble()`/`strip_resolved_optional`), can never take the no-op branch — `del(.[]?)` on a non-container always raises `Cannot iterate over X` today regardless of `?`. Confirmed live via a debug probe (three independent `/code-review` agents converged on this same finding, testing shapes like `del(.a[]?, .b[]?)`).

## 2. The `terminal` vector's merged `optional` bool is computed but discarded

`src/jq/eval.rs:20103-20120` (as of `#1322`'s merge) OR-merges `path[start].optional` into `terminal: Vec<(ArrayStep, bool)>`, with an elaborate comment justifying merge-not-overwrite semantics for argument order. The only consumer (`.map(|(step, _)| ...)`, around line 20227) discards the second tuple element entirely — the merged bool has never influenced output, independent of `#1322`'s own change to the neighboring block.

## 3. Root cause: `DeleteStep.optional` is dead end-to-end

`src/jq/eval.rs:19836-19839` (`DeleteStep`) and `flatten_delete_path`'s own `Expr::Optional(inner) => flatten_delete_path(inner, true, out)` arm (line ~19857): this arm can only fire if an un-stripped `Expr::Optional` reaches `flatten_delete_path`, which `#1322`'s own comment (and every finder agent's trace) proves never happens for `del()`'s call shape. This one field being permanently `false` is the shared root of `#928`/`#1003`/`#1034`/`#1322`, plus finding 1 above and the discarded merge in finding 2.

## Suggested fix direction

Two options, not investigated in depth:
- (a) The real fix: a `debug_assert!(!optional)` (or removing the field/its plumbing entirely, if truly nothing ever needs it) at `DeleteStep`'s own construction site in `flatten_delete_path`, catching the invariant structurally instead of requiring a sixth discovery of the same symptom in some other sibling function.
- (b) If `del()`'s own semantics are ever meant to actually honor per-path `?` at this depth (arguably the more correct design, matching `flatten_delete_path`'s sibling `Expr::Optional` handling elsewhere), thread `optional` through properly from `resolve_dynamic_indexes` instead of stripping it via `strip_resolved_optional` — a bigger, riskier change.

## Related

Found during `/code-review` of `#1322` (`fix(jq): remove delete_expr_array_paths's dead optional-gate arm`), PR #1329. Not caused by, or fixed by, that PR — pre-existing since before it, and `delete_expr_iterate_paths`'s own instance predates `#1322` too.

newhoggy:
## Summary

#1331 claimed `DeleteStep.optional` is "dead end-to-end" (never constructed as `true` in any way that reaches a live consumer) and suggested a `debug_assert!(!optional)` at its own construction site in `flatten_delete_path` as "the real fix."

Live verification (see below) found this premise is **partially wrong** — implementing the suggested fix as literally described would have introduced a live, reachable panic in debug builds.

## What's actually true (verified live)

- **`delete_expr_paths_at`'s own comma-grouped consumers** (`delete_expr_iterate_paths`, and the `terminal` vector's merge logic) genuinely never see `optional == true` — confirmed via extensive live probing (couldn't construct a counter-example across field/index/slice/iterate shapes, including mixed commas with one computed-key branch and one static branch). `resolve_dynamic_indexes`'s `assemble()` does strip `Expr::Optional` before any path reaches this specific machinery. This part of #1331's premise holds.
- **A *different* consumer does not have this guarantee**: `yq_del_slice_outcome`'s own `wrap` closure (reached via `rewrite_yq_del_comma_branches`'s pre-rewrite pass, and `builtin_del`'s own per-path calls) genuinely receives `DeleteStep.optional == true` on ordinary, live-reachable input — confirmed via `del(.a?[1:3])` (optional on the field, not the slice) and `del(.[(0,5)], .[5]?)` (a mixed comma). A `debug_assert!(!optional)` at `flatten_delete_path`'s own construction site — the general fix #1331 suggested — would have panicked on exactly this input in any debug build, including this crate's own `cargo test`.
- That said, `wrap`'s optional-preserving rewrite (re-wrapping the dropped parent key in `Expr::Optional`) turned out to be **observably inert everywhere I could test**: `yq_del_slice_outcome` already gates entry to `wrap` on `navigate_read_only` confirming the whole prefix exists, so by the time `wrap` runs, deleting that prefix can never error regardless of whether it's wrapped in `Optional` or not. I didn't find a way to make this matter, but I also didn't find a way to *prove* it can't — so I left `wrap`'s own logic untouched rather than risk removing something I can't fully rule out being load-bearing.

## What changed

- Simplified `delete_expr_paths_at`'s `terminal: Vec<(ArrayStep, bool)>` to `Vec<ArrayStep>` — the merged `optional` bool (`entry.1 |= path[start].optional`) was computed but its only consumer already discarded it via `.map(|(step, _)| ...)`. This part of #1331's finding #2 was correct and is a safe, zero-behavior-change cleanup.
- Added a `debug_assert!` in `delete_expr_iterate_paths` itself — correctly scoped to *that function's own* verified invariant, not the general `DeleteStep` type. Replaced the stale comment (in both this function and the neighboring `delete_expr_paths_at` non-array-container branch) that had generalized the claim beyond what's actually true.
- Did **not** add anything at `flatten_delete_path`'s construction site, and did **not** remove `DeleteStep.optional`/its plumbing — both would be wrong given the live counter-example above.

## Verification

- Two new CLI tests: `del(.a?[1:3])`-shaped input (field-level optional before a chained slice — the shape that's actually live-true) and the `del(.[(0,5)], .[5]?)` mixed-comma dedup shape from #1331's own example.
- Full local suite (build, full test suite — debug mode, so the new `debug_assert!` gets exercised — both CI clippy legs, `cargo doc --no-deps --all-features` under `-D warnings`, `no_std` check, `cargo fmt --check`) all green.
- Patch coverage: no new uncovered lines.

Refs #1331

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -20901,20 +20901,31 @@ enum ArrayStep {
 ///
 /// `paths` here always traces back to [`resolve_dynamic_indexes`]'s own
 /// output, which strips every `Expr::Optional` wrapper via
-/// `strip_resolved_optional` before returning (#1331's own investigation
-/// confirmed this holds for every comma-grouped shape that can reach
-/// `delete_expr_paths_at`/this function at all, live-tested across field,
-/// index, and iterate steps -- including a mixed comma with one computed
-/// and one already-static branch, where `resolve_dynamic_indexes` still
-/// strips the static branch before this function ever sees it). So
-/// `optional` below is always `false` in practice; kept as a live value
-/// (not deleted) rather than hardcoded, with a `debug_assert` verifying the
-/// invariant rather than silently trusting it -- #1331 found that a
-/// sibling claim ("this can never be true") was wrong for a *different*
-/// consumer of the same `DeleteStep.optional` field
-/// ([`yq_del_slice_outcome`]'s own `wrap` closure, reachable via
-/// `del(.a?[1:3])`-shaped input), so this function's own narrower version
-/// of the claim is verified here rather than assumed to generalize.
+/// `strip_resolved_optional` before returning -- so `optional` below is
+/// always `false` in practice; kept as a live value (not deleted) rather
+/// than hardcoded, with a `debug_assert` verifying the invariant rather
+/// than silently trusting it (#1331 found that an identically-worded
+/// claim was wrong for a *different* consumer of the same
+/// `DeleteStep.optional` field -- [`yq_del_slice_outcome`]'s own `wrap`
+/// closure, reachable via `del(.a?[1:3])`-shaped input -- so each site's
+/// own version of the claim needs its own verification, not assumed to
+/// generalize).
+///
+/// This function itself may be entirely unreachable from `del()`'s own
+/// comma-grouped route: extensive live probing (#1331's second review
+/// round, three independent passes) found `resolve_node`'s own handling
+/// of `Expr::Iterate` eagerly expands a bare `.[]`/`.[]?` into concrete
+/// per-element `Field`/`Index` components *before* `flatten_delete_path`
+/// ever runs, for every shape tried -- meaning a literal `DeleteStep`
+/// carrying `Expr::Iterate` may never actually get constructed via
+/// `builtin_del`'s `paths.len() > 1` branch (the only external entry to
+/// `delete_expr_paths_at`) at all. Left as-is rather than removed:
+/// proving a function is *never* reachable, as opposed to not reachable
+/// via every input tried so far, needs more exhaustive verification than
+/// #1331's own scope covers -- filed as #1382 rather than assumed here.
+/// If genuinely dead, the `debug_assert` below (and the two `optional`-
+/// gated match arms further down, pre-dating this issue) are dead along
+/// with it, harmlessly.
 fn delete_expr_iterate_paths(
     value: OwnedValue,
     paths: &[&[DeleteStep]],
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -20741,7 +20741,25 @@ fn delete_expr_array_paths(
     // element, and the linear `find`/scan these two used to do never matched,
     // so every sibling walked the whole growing list. 99.9% of a 200,000-element
     // run's samples sat in those two loops.
-    let mut terminal: IndexMap<ArrayStep, bool> = IndexMap::new();
+    //
+    // `IndexSet<ArrayStep>`, not `IndexMap<ArrayStep, bool>`: an earlier form
+    // (both here and in the pre-#1301 `Vec` version) also merged each step's
+    // `optional` flag (`*entry.or_insert(false) |= path[start].optional`) to
+    // cover every occurrence rather than just whichever sibling was inserted
+    // first. That merged bool had no reader -- the one consumer below
+    // (building `owned_keys` for `delete_keys`) already discarded it. Dropped,
+    // not because the flag can never be `true` here (#1331's investigation
+    // found the broader claim that it can't doesn't generalize past this
+    // specific dead-write site -- see `delete_expr_iterate_paths`'s own doc
+    // comment for the narrower, verified version of that claim), but because
+    // deleting a *terminal* index/slice always goes through `delete_keys`,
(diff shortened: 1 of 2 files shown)


newhoggy:
Second review round (multiple angles, several with live probes) confirmed the core fix is correct and no new bugs, with two real actionable findings addressed:

- **Asymmetry**: \`delete_expr_array_paths\`'s non-array-container branch made the identical "optional is always false here" claim as \`delete_expr_iterate_paths\` but had no runtime verification. Added a matching \`debug_assert\`.
- **Coverage gap**: my original tests didn't reach \`delete_expr_iterate_paths\`'s own Iterate dispatch at all. Added a test for it — which then led to the interesting part:

**Correcting my own overclaim**: that new test's doc comment claimed \`del(.a[]?, .b[]?)\` "forces the multi-path walker's Iterate dispatch." A live debug probe (three independent passes, one in an isolated clone) found this is false — \`resolve_node\` appears to eagerly expand every bare \`.[]\`/\`.[]?\` into concrete per-element components *before* \`flatten_delete_path\` ever runs, so the test actually reaches \`delete_expr_array_paths\` instead. I couldn't construct *any* input across ~15+ shapes (nested, computed-key, mixed comma, jq/yq mode) that reaches \`delete_expr_iterate_paths\` at all through \`del()\`'s comma-grouped route.

Rather than assume that function is dead code (a stronger claim than "not reachable via what I tried"), corrected the doc comment to say what's actually verified vs. not, kept the (harmless either way) \`debug_assert\`, renamed the test to describe what it actually covers, and filed #1382 for the open reachability question.

One process note for whoever reviews next: mid-review, a concurrent finder agent left a stray \`eprintln!\` uncommitted in this shared worktree (broke the \`no_std\` build). Caught via the second CI-equivalent local pass before push — cleaned up, and the affected commit was re-verified clean before landing. Full local suite (build, full test suite, both clippy legs, rustdoc, no_std, fmt) green on the final state.

newhoggy:
Rebased onto \`main\`, which had meanwhile merged #1301's own \`del()\` performance fix (\`0abd2397a\`) — a real, structural conflict, since it converts \`delete_expr_array_paths\`'s sibling grouping from a linear-scan \`Vec\` to \`IndexMap\`/\`IndexSet\` for O(1) lookup, touching the exact same declaration this PR simplifies.

Reconciled as: kept #1301's \`IndexMap\`/\`IndexSet\` structure (the real perf win), and re-applied this PR's own finding on top — the \`bool\` half of \`terminal\`'s entries is just as dead under \`IndexMap<ArrayStep, bool>\` as it was under the old \`Vec<(ArrayStep, bool)>\` (same reason: \`delete_keys\`, the only consumer, already discards it), so \`terminal\` is now \`IndexSet<ArrayStep>\` rather than \`IndexMap<ArrayStep, bool>\`. Order-independence for \`del(.[(0,5)], .[5]?)\` (either argument order) still holds — \`IndexSet::insert\` dedupes a repeated step regardless of which occurrence's own \`?\` would have set the now-untracked flag.

Full local verification (build, full test suite including #1301's own new benchmarks/tests, both clippy legs, rustdoc, no_std, fmt) green on the reconciled state.

PR #1388, opened 2026-08-21
fix(jq): auto-vivify a null slice-write target instead of a silent no-op

### LINKED ISSUE #1340: Slice write mid-chain silently no-ops instead of erroring when intermediate target is absent (through_slice null auto-vivification)
## Summary

A slice write with a further path step after it — `(.b[0:1]? | .c) = 9` — silently produces the wrong output instead of erroring, when `.b` is absent (or `null`). Confirmed against real jq 1.7.1: this errors with `A slice of an array can only be assigned another array`. `succinctly jq` instead silently auto-vivifies and produces wrong data.

```bash
echo '{"c":0}' | jq '(.b[0:1]? | .c) = 9'
# jq: error: A slice of an array can only be assigned another array

echo '{"c":0}' | succinctly jq '(.b[0:1]? | .c) = 9'
# {"c":0,"b":null}   <-- wrong: silently no-ops/corrupts instead of erroring
```

## Status: pre-existing, not introduced by #1311

This was found during #1311's code review, surfaced initially via the double-paren grouped form `((.b[0:1])? | .c) = 9` that #1311 fixes the *parsing side* of (previously fell through as "invalid path component"). I re-verified live against `origin/main` (i.e. **without** #1311's changes) using the simpler, ungrouped, single-`?` form `(.b[0:1]? | .c) = 9` — same silent-corruption result. That shape is a flat `Pipe([Field(b), Optional(Slice), Field(c)])` and does not touch `push_path_components`'s new `Expr::Optional` recursion at all (there's no `Pipe`/`Paren` nested *inside* an `Optional` here for that arm to matter). So this bug predates #1311 and is orthogonal to it — #1311 just adds one more syntactic spelling that reaches the same pre-existing gap.

## Suspected root cause

Traced partially: `set_path`'s `Expr::Optional` arm already correctly distinguishes `is_write_time_application_error()` from ordinary navigation failure:

```rust
Err(e) if e.is_write_time_application_error() => Err(e),
Err(_) => Ok(()),
```

So the swallow is happening deeper than that dispatch — most likely in `through_slice`'s own null-target auto-vivification behavior (an existing code comment notes a slice write auto-vivifies `null` into `[]` rather than no-op'ing) interacting incorrectly with write-time-error classification once the intermediate target has just been auto-vivified. The `?` on the slice is presumably suppressing what should be a *write-time* application error (assigning a non-array through a slice) as if it were an ordinary navigation failure.

Needs a dedicated investigation of `through_slice`'s null-vivification path and how it flags errors as write-time-application vs. navigation, rather than a quick patch — this is silent data corruption, not just a wrong error message, so it deserves real severity even though it's a narrow, unusual-looking query shape.

## Repro shapes confirmed

- `(.b[0:1]? | .c) = 9` on `{"c":0}` (`.b` absent) — wrong output, no error
- `((.b[0:1])? | .c) = 9` on `{"c":0}` — same, via the #1311-fixed parse path

Both diverge from real jq 1.7.1, which raises `A slice of an array can only be assigned another array` in both cases.

newhoggy:
## Summary

Fixes #1340. `through_slice` (the shared choke point `=`/`|=` route slice writes through) had no dedicated arm for a `Null` root — a value auto-vivified into `Null` by navigating an absent field before a slice fell through to either an unconditional error or, when the slice carried a `?`, a silent unconditional no-op. Neither matches real jq.

```bash
$ echo '{"c":0}' | jq -c '(.b[0:1]? | .c) = 9'
$ echo '{"c":0}' | succinctly jq -c '(.b[0:1]? | .c) = 9'   # before this PR
{"c":0,"b":null}
```

While investigating, I also found the *simpler* terminal case was broken the same way, even though it isn't the shape #1340 was filed against:

```bash
$ echo '{"c":0}' | jq -c '.b[0:1] = [1]'
{"c":0,"b":[1]}
$ echo '{"c":0}' | succinctly jq -c '.b[0:1] = [1]'   # before this PR
jq: error (at <stdin>:1): Cannot index null with object
```

## Root cause

Real jq attempts the write when a slice target is `null`: the field auto-vivifies, the replacement value is computed by continuing the path, and only if that continuation's result isn't an array does jq raise "A slice of an array can only be assigned another array" — a write-time application error `?` does not suppress (confirmed live both with and without `?`).

`set_value_at_path` (`setpath()`'s own recursive walker, fixed for an unrelated regression in #1338 earlier this session) already implements exactly this for its own slice-descriptor path components — a `Null` value recurses into the continuation, and the result is checked for `Array`-ness before splicing back. `through_slice` had drifted from that: a `Null` root simply isn't `Array`/`String`, so it fell to the generic `_ if optional => Ok(())` catch-all (silently swallowing the write when `?` was present) or the final `Err(cannot_index_with_type(...))` catch-all (a wrong, generic message when it wasn't).

## Fix

Added a dedicated `OwnedValue::Null` arm to `through_slice`, mirroring `set_value_at_path`'s design: run `edit` against a fresh `Null`, then require the result to be an `Array` before splicing it into `root`. Ordered **after** the existing yq-mode no-op arm (`_ if scalar_noop && is_yq_slice_empty_container_scalar(root)`), which already correctly handles `Null` for yq's own, different no-op semantics — `scalar_noop`/`container_noop` are only ever `true` in yq mode, so this ordering leaves yq completely unaffected.

## Verification

Live-verified against real jq 1.7.1 and yq v4.53.3 (Homebrew) across:
- The issue's own repro, with and without `?`
- Terminal slice write on an absent field, both array and scalar RHS, with and without `?`
- `|=` (shares `through_slice` with `=`)
- `del()` on an absent field (confirmed unaffected — already correct)
- A genuine non-`Null` scalar root with `?` (confirmed unaffected — still an ordinary suppressed navigation failure)
- An already-existing array root (confirmed unaffected — still errors on indexing the array directly, a different message for a different shape)
- Nested slice-of-slice and deeper mid-chain shapes (compose correctly by construction, since the new arm just recurses through the same `edit`)
- yq mode: all four shapes above (matches its own, different, pre-existing no-op behavior byte-for-byte, before and after this fix)

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 2694 lib tests, 0 failures, no regressions)
- [x] `cargo fmt --check`
- [x] `omni-dev coverage diff` — 100% patch coverage (49/49 new lines)
- [x] Live-verified against real jq 1.7.1 and yq v4.53.3

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -11960,6 +11960,35 @@ fn through_slice<E: From<EvalError>>(
             edit(&mut throwaway)?;
             Ok(())
         }
+        // A `Null` target that reaches here (jq mode, or a yq operator that
+        // doesn't no-op — the arm above already claimed every yq-mode
+        // `Null`) still needs a real attempt, not a silent skip: real jq
+        // auto-vivifies a null slice target exactly like `setpath()`'s own
+        // `set_value_at_path` walker already does (#1340, confirmed live:
+        // `null | setpath([{"start":0,"end":1},"c"]; 9)` and this arm raise
+        // the identical "A slice of an array can only be assigned another
+        // array"). `edit` runs against a fresh `Null`, not an empty array,
+        // so a mid-chain field step recurses into *its own*
+        // auto-vivification (`Null` -> object) rather than immediately
+        // failing to index an array the way an already-empty array target
+        // does (`{} | .b[0:1].c = 9` on an *existing* empty `.b` still
+        // raises "Cannot index array with string", a different, correct
+        // message for a different shape) — only once `edit` is done does
+        // the result have to actually be an array to splice back. Never
+        // gated on `optional`: real jq's `?` does not suppress this either
+        // (`.b[0:1]? = 5` on absent `.b` still raises the same error,
+        // confirmed live) — `slice_assign_non_array()` is a write-time
+        // application error, the same family `set_path`'s own
+        // `Expr::Optional` arm already carves out of `?`'s reach.
+        OwnedValue::Null => {
+            let mut sub = OwnedValue::Null;
+            edit(&mut sub)?;
+            let OwnedValue::Array(items) = sub else {
+                return Err(EvalError::slice_assign_non_array().into());
+            };
+            *root = OwnedValue::Array(items);
+            Ok(())
+        }
         _ if optional => Ok(()),
         other => Err(EvalError::cannot_index_with_type(owned_type_name(other), "object").into()),
     }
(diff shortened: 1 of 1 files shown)


newhoggy:
Pushed a follow-up commit addressing the review's finding, plus corrections to the doc
comments it exposed as inaccurate.

**Fixed — real: the new `Null` arm wasn't yq-mode-safe.** The review (independently
confirmed live against yq v4.53.3 and a pre-fix build) found that `through_slice`'s new
`OwnedValue::Null` arm fired unconditionally, so yq's `-=`/`*=` (both `scalar_noop ==
false`, unclaimed by the pre-existing yq no-op arm) fell into it and started attempting
jq-style auto-vivification instead of their own behavior.

Fixing this turned out deeper than flipping `scalar_noop` for `-=` — my first attempt at
that broke on a scalar RHS specifically: real yq's own `-=` no-op (confirmed live) doesn't
run through this codebase's existing "run the filter against a throwaway `[]`, discard on
success" mechanism at all. That mechanism only works for `+=` because `[] + 5` happens not
to error; `[] - 5` genuinely does, so reusing it for `-=` leaked a real (if different)
error instead of the no-op real yq actually performs. Further probing found `*=` is even
more specific — RHS-type-dependent (`.b[0:1] *= [1]` no-ops, `.b[0:1] *= 5` genuinely
errors) — confirming this isn't a one-line fix.

Landed the narrower, unambiguously-correct fix instead: gated the new `Null` arm on
`!container_noop` (jq mode only — `container_noop` is `S::TAG == EvalTag::Yq`
unconditionally at every call site). This restores `-=`/`*=`'s yq-mode behavior on a null
slice target to **byte-for-byte identical** to before this PR touched anything (verified
against a pre-fix build, terminal and mid-chain, both operators) — genuinely unaffected,
not just claimed to be. Added `test_assign_slice_mid_chain_null_target_yq_mode_sub_mul_
unaffected_1340` pinning this.

Also corrected several existing doc comments this investigation showed were wrong (the
codebase had long claimed `-=`/`*=` both "error instead of no-op'ing" for a null slice
target, but had apparently never actually checked `-=` live) — now accurately describe
what's confirmed vs. what remains an open gap.

**Filed as follow-up, not fixed here: `-=`/`*=`'s actual yq semantics.** #1390 documents
real yq's full behavior (`-=` unconditionally no-ops; `*=` no-ops for an array RHS but
genuinely errors for a scalar RHS) and why implementing it needs a different mechanism
than the existing throwaway-`[]`-filter approach, not a flag flip.

**Reuse/Simplification finding (through_slice's `Array`/`Null` arms duplicate ~4 lines
also present in `set_value_at_path`'s tuple-based arm): considered, deferred.** Agreed
it's real but low-severity (explicitly rated as such by the review too) — unifying it
would touch already-tested, working code purely for stylistic consolidation, with no
correctness benefit. Leaving as-is rather than expanding this PR's diff further.

Everything else the review checked (CLAUDE.md conventions, cross-file callers including
`del()`'s structural immunity to the new arm, efficiency/altitude) came back clean and is
unaffected by this follow-up commit.

Re-ran the full required-checks list after this commit: `cargo fmt --check`, both CI
`clippy --all-targets` invocations, full `cargo test` (2695 lib tests, 0 failures),
`cargo check --no-default-features`, and `omni-dev coverage diff` (100% patch coverage,
62/62 new lines).

Issue #1247, opened 2026-08-18
jq: one StandardJson::Error -> OwnedValue::Null arm remains in eval.rs's own to_owned_at_depth — prove unreachable or route it


newhoggy:
## Summary

#1192 enumerated 4 call sites where a JSON/YAML string scalar (or object/mapping key) that passes structural validation but fails to *decode* (invalid UTF-8, an invalid escape, an invalid `\u` codepoint) silently degrades instead of surfacing an `EvalError`. PR (issue-1192-decode-failing-scalar-error-routing branch) fixed 2 of the 4:

- `owned_from_standard_json_at_depth` (`src/jq/eval_generic.rs`, the "reindex bridge" full-evaluator fallback)
- `standard_json_to_jq_value` (`src/bin/succinctly/jq_runner.rs`, builds the top-level `JqValue` for CLI output)

This issue tracks the other 2, which turned out to be the *actual* live-reachable bug:

- `to_owned`/`to_owned_at_depth` (`src/jq/eval_generic.rs`) -- generic over `DocumentValue` (both JSON and YAML)
- `cursor_to_owned`/`cursor_to_owned_at_depth` (`src/jq/lazy.rs`) -- JSON-specific, backs `JqValue::materialize()`/`into_owned()`

## Why this split, and why these two are the real bug

#1192 assessed itself as "Tier 2... no unresolved architectural fork" and framed all 4 sites as comparably tractable. That assessment was wrong for these two, discovered only through implementation:

**Reachability is completely inverted from what #1192 assumed.** Empirical testing (~25 distinct `jq` expressions over `{"a": "\xff\xfe"}` and similar malformed-UTF-8 documents -- navigation, construction, `reduce`/`foreach`, assignment, and every builtin category the crate implements) found **zero** cases reaching the 2 sites just fixed. Every one of them resolved through `to_owned`/`cursor_to_owned` instead -- confirmed independently by `to_owned_at_depth`'s own doc comment (`eval_generic.rs`, added by the #1190 revert), which explicitly states this is "reachable through completely ordinary CLI usage (any non-identity query over a document containing such a string)". `eval_generic.rs`'s own native dispatch is comprehensive enough that the "reindex bridge" full-evaluator fallback (which the 2 just-fixed sites depend on) is essentially never taken for this scenario.

[...]

Issue #1194, opened 2026-08-17
jq: a malformed object field (e.g. {invalid}) silently disappears instead of erroring


newhoggy:
## Summary

Split off from #1171 (which covered the top-level "silent empty output"
symptom, now fixed for truncated/unrecognized-shape top-level values and
for leading-dot numbers). This is the harder, deeper sibling: a malformed
field *inside* an otherwise well-bracketed object silently vanishes
instead of producing a parse error.

## Repro

```
$ echo '{invalid}' | succinctly jq -c '.'
{}                              # real jq: "Invalid numeric literal", exit 5

$ echo '[xyz123]' | succinctly jq -c '.'
[null]                          # real jq: "Invalid numeric literal" (or similar), exit 5
```

Unlike the cases #1171 fixed, `{invalid}`'s outer braces ARE well-formed
(`find_json_values`'s bracket-matching scanner correctly identifies the
whole `{invalid}` span) — the malformed content is *inside*, where the
semi-indexer's own structural scan and materialization silently swallow
it instead of erroring.

## Root cause

Two related swallow points, per #1171's own investigation:

- `src/json/light.rs:987-1004`, `JsonFields::uncons()`: the semi-indexer
  doesn't distinguish "key:value pair" from "lone bareword leaf" — its
  `?` at line 991 collapses "no sibling to pair as a value" into `None`
  ("no more fields"), indistinguishable from a genuinely empty object.
  This is why `{invalid}` becomes `{}` rather than keeping a broken field.
- `src/jq/lazy.rs:730` and `src/jq/eval_generic.rs:863`, both
  `StandardJson::Error(_) => OwnedValue::Null` — this is the generic
  "unrecognized leaf shape" fallback (already partly addressed by #1171
  widening the leading-dot-number case out of this arm entirely), still
  reachable for any OTHER unrecognized shape (`xyz123`, etc.).

## Why this needs its own fix, not folded into #1171

Both `cursor_to_owned_at_depth` (`lazy.rs`) and its `eval_generic.rs`
sibling return a plain `OwnedValue`, not a `Result` — propagating a real
error out of the `StandardJson::Error(_)` arm means threading a
`Result`/error signal through the whole recursive tree-walk (the
`Array`/`Object` arms that call it recursively) and every external
caller, not a one-line fix. `JsonFields::uncons()`'s fix has a similar
shape: distinguishing "no more fields" from "unpaired key" needs a
richer return type that ripples to its own handful of callers.

## Scope note

[...]

Issue #1192, opened 2026-08-17
jq/yq: a decode-failing string scalar (value or key) should surface as a real jq error, not silently degrade — needs EvalError/ErrorSink routing


newhoggy:
## Summary

Follow-up to #1098, which described (as "Severity: Low", "theoretical, no live repro") a gap where a JSON/YAML string scalar that passes structural validation (the semi-index accepts its span) but fails to *decode* (invalid `\u` escape, invalid UTF-8) silently materializes as `null` (values) or vanishes entirely (object/mapping keys), instead of surfacing an error.

A first attempt at fixing this (PR #1190) made the decode-failure case `panic!()` instead. Code review found — via live reproduction against built `jq`/`yq` binaries, not just reading source — that this was the wrong fix:

- **It's reachable through completely ordinary CLI usage**, not a theoretical library-only edge case: `printf '{"a": "\xff\xfe"}' | succinctly jq 'sort'` (or almost any non-identity query — `+`, `[.]`, `walk(.)`, `.a |= .`, `length`, ...) panics live. Only the bare `.` identity query avoids it, via a separate raw-byte streaming fast path that never materializes through the affected function. `yq --input-format json` has *no* upfront validation gate at all before reaching the semi-indexer (unlike `--argjson`, which goes through `serde_json`'s strict parse first) — so this is yq's *primary* document-input path, not a corner case.
- **It breaks this codebase's own documented "evaluation continues past one error" convention** (`ErrorSink`, #355): a multi-record JSON/YAML stream with one malformed value now aborts the *entire batch* on the first bad record instead of reporting a per-record error and continuing — confirmed live (a 3-value stream `1 / {bad} / 3` prints only `1`, then dies with exit 101; a genuine type-mismatch error in the same position instead prints results for the other two records and exits 5, per the existing convention).
- **PR #1190 was reverted** rather than patched further, restoring the original silent-degradation behavior with test coverage pinning it (satisfying #1098's own alternative suggested direction — "add a test... to catch this class of drift").

## What a correct fix needs

1. **Route through `EvalError`/`ErrorSink` instead of panicking.** The failure needs to become a normal, catchable jq-style error for the specific value/record affected, consistent with how every other document-input problem (type mismatches, etc.) is already reported — preserving the "continue past one error" batch semantics rather than killing the whole process.
2. **Applied consistently across every sibling copy**, not just one. Investigation found (at least) four independent copies of "materialize a JSON/YAML string, silently degrading on decode failure":
[...]

PR #1391, opened 2026-08-21
fix(jq, yaml, json, cli): route string decode failures through EvalError (#1247, #1242, #1194)

### LINKED ISSUE #1242: yaml: succinctly accepts a document with a dangling non-continuable trailing UTF-8 byte that real yq's parser rejects
**Severity: Low**

## Summary

Found while implementing #1216 (@urid's malformed-escape error byte-escaping).
A YAML document whose content ends with a single, non-continuable UTF-8
lead/continuation byte (i.e., genuinely invalid/incomplete UTF-8 at the very
end of the document, not just inside one specific string) is rejected outright
by real yq's own parser, but succinctly's parser accepts the same document and
produces `null` — a document-level UTF-8 validation strictness gap, unrelated
to `@urid`'s own error-message formatting (which #1216 fixed separately).

```
$ printf '"%%\xe4"\n' | yq -o=json '@urid' -
Error: bad file '-': yaml: offset 3: invalid trailing UTF-8 octet

$ printf '"%%\xe4"\n' | succinctly yq -o=json '@urid'
"null"
```

The filter (`@urid`) is irrelevant here — real yq's error happens at the YAML
parse stage, before any filter ever runs (confirmed: the same input with `.`
as the filter also errors identically in real yq). succinctly's parser instead
accepts the malformed document and hands `@urid` something it evaluates to
`null` on — worth investigating exactly what succinctly's parser resolves this
truncated content to internally (a dropped/replaced byte? an early scalar
termination?) before deciding on a fix.

## Why this is separate from #1216

#1216's own scope was `@urid`'s error-message *byte quoting* for an
already-successfully-parsed document — this issue is about whether the
*document itself* should have been accepted by the parser at all. Different
layer, different code path (YAML/JSON semi-indexing's own UTF-8 validation,
not `format_urid`/`EvalError::urid_invalid_escape`).

## Suggested fix direction

Needs its own investigation into where succinctly's YAML (and/or JSON) parser
currently tolerates a dangling/incomplete trailing UTF-8 sequence, and whether
tightening that matches real yq's own strictness without regressing anything
that currently and correctly rejects other malformed-UTF-8 shapes. Not
attempted here — filing to make the gap trackable.

newhoggy:
## Description

A string scalar the semi-index accepted but that cannot be **decoded** — invalid UTF-8, an
invalid escape, an invalid `\u` codepoint — was silently swallowed. It became `null`, or
`""`, or the field vanished, and the process exited 0. `to_owned_at_depth`'s own doc comment
has named this as deferred since PR #1190's `panic!()` attempt was reverted as
live-reachable; this routes it through `EvalError` instead, which is what that comment asked
for.

#1247 was tiered **Tier 3** — an unresolved architectural fork between threading `Result`
through the `to_owned` family and moving validation upfront — and its tier review asked for
one design doc covering #1247 and #1194 together. That doc is the first commit
([`docs/plan/decode-failure-routing.md`](docs/plan/decode-failure-routing.md)); the rest
implement five of its six stages.

### Three corrections to the issues' own premises

Each was verified against the build and the pinned oracles rather than taken from the issue
text:

1. **The "58+ call sites, all needing `Result` threading" estimate overstates the work.** Of
   62 grep hits, 13 are doc comments, 8 are tests and 7 are the recursive definitions. Of
   the **53 real call sites, 40 already sat in a fallible context** — the generic
   evaluator's error channel is `GenericResult::Error` / `Control::Error` in the value
   domain, not `Result`. `eval_single` and `eval_builtin` both return a bare
   `GenericResult<V>`.
2. **#1191, named as a blocking prerequisite for the YAML side, was already closed and
   fixed.** `as_str()`'s `Alias` arm resolves the full chain now, so the confound it warns
   about no longer exists.
3. **#1194 is only half in scope.** `[xyz123] → [null]` is fixed; `{invalid} → {}` is not —
   `JsonFields::uncons` collapses a lone bareword leaf into "no more fields" at the
   semi-index layer, so no field is ever constructed to raise on.

### Six swallow sites no issue records

Two classes are worse than the silent `null` the issues describe:

```console
$ printf 'b: "x\qy"\n' | succinctly yq -o=json '.'      # before
{"b": "xnull}                                           # unparseable, exit 0
$ printf '"a\qb": 1\n' | succinctly yq -o=json '.'      # before
{"a"": 1}                                               # unparseable, exit 0

$ printf '{"\ud800":1,"b":2}' | succinctly jq '.b'      # before
null                                                    # but "b":2 is right there
```

The last is the sharpest: one undecodable key made **every later field invisible to lookup**
while `keys`/`length` still reported it — the two halves of the CLI disagreed about whether
the field existed.

### The oracles disagree, which neither issue records

```console
$ printf 'a: "\xff"'   | yq   →  bad file: invalid leading UTF-8 octet, exit 1
$ printf '{"a":"\xff"}' | jq  →  {"a":"<U+FFFD>"},                      exit 0
```

So the UTF-8 fix is per-mode, not one rule: yq rejects, jq substitutes.

## Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Performance improvement
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Test coverage improvement
- [ ] CI/CD changes

"Breaking" in the sense that documents which previously produced output at exit 0 now
produce a diagnostic and a non-zero exit. That is the point of the change, but it is a
user-visible behaviour change on malformed input, and `yq --validate` in particular now
rejects documents it used to pass.

## Related Issue

Fixes #1242.
Advances #1247 (Stage 6 remains — see below).
Advances #1194 (its `[xyz123]` half; its headline `{invalid}` half remains).

## Changes Made

Seven commits, staged so each can be reviewed — or reverted — on its own. The first two are
plain bug fixes with no signature churn and land before the wide change.

**Stage 0 — design doc.** `docs/plan/decode-failure-routing.md`, in the shape of #1282's
deliverable: verified premises, oracle behaviour tables, corrected call-site accounting,
measured costs, staged delivery, explicit non-goals, open risks.

**Stage 1 — `find`/`find_cursor` no longer hide valid siblings.** `JsonFields` and
`YamlFields` each ran `if key.as_str().ok()? == name` inside their scan loop; that `?`
returned from `find` itself, ending the whole search on the first undecodable key. Now
skipped, not fatal. Last-duplicate-key-wins is unchanged.

**Stage 2 — a failed transcode rolls back instead of emitting invalid JSON.** The YAML→JSON
transcoders push the opening `"` before they can discover a bad escape, and each caller's
`Err(_)` arm appended its own `null`/`""` after it. Six call sites shared the shape, not the
two first scoped. The rollback went **inside** `write_yaml_string_to_json` — it owns the
`&mut String`, so it can rewind its own entry length and fix all four buffered sites at once
rather than asking four callers to remember. The two streaming sites commit through a
scratch buffer, and only in the slow branch, so P9's allocation-free path is untouched.

*Independently confirmed by the YAML Test Suite:* cases `55WF` and `HRE5` sat on
`tests/data/yq-path-consistency-known-failures.txt` for exactly this reason ("streaming
emits malformed JSON, DOM emits null"). Both now agree across paths; that manifest's only
category is retired.

**Stage 3 — the `to_owned` family becomes fallible.** The main change.
`to_owned`/`to_owned_at_depth`/`to_owned_cursor`/`to_owned_cursor_at_depth` (`eval_generic.rs`)
and `cursor_to_owned`/`cursor_to_owned_at_depth` (`lazy.rs`) now return
`Result<OwnedValue, EvalError>`. `assert_nesting_depth` stays a `panic!`: unbounded
recursion is a different failure class, and routing it here would make a stack-overflow
guard catchable by `try`/`catch`.

- **Detection needed a real signal, not an inference.** `as_str()` conflates "not a string"
  with "a string this document cannot hand back", which is how the failure reached a
  materializer indistinguishable from an unknown type. Added
  `DocumentValue::string_decode_error() -> Option<&'static str>`, a provided method
  defaulting to `None`. Wording comes from `JsonError::message` / `YamlStringError::message`,
  split out of each `Display` so there is one definition, not a copy.
- **Three more silent-drop sites, all surfaced by the compiler once the types changed.**
  `effective_fields`' dedup walk dropped any field whose key wouldn't decode, so
  `to_entries` returned one entry fewer than the document had; `lazy_keys_array_to_owned`
  dropped the key from `keys`; `to_owned_with_comments` had the same key guard.
- **A misleading-error class, not just a silent one.** Every builtin's type dispatch is a
  chain of `as_str()`/`as_array()`/… tests, so an undecodable string fails all of them and
  lands in the same `else` as a genuine type mismatch: `.a | length` reported
  `null (null) has no length` about a perfectly good string token. `type_error_or_decode_failure`
  names the cause at the seven sites reporting such a fall-through.
- **One lossy materialization kept on purpose.** `to_owned_for_diagnostic` serves the call
  sites that materialize a value only to quote it into an error already being raised;
  propagating there would mean abandoning a correctly-diagnosed error to report that its
  message could not be built. Its own function, with the rationale in its doc comment.

**Stage 4 — `StandardJson::Error` arms raise.** Fixes `[xyz123] → [null]`. Evaluation still
continues past the bad value, so the good documents either side of it in a multi-value
stream still produce output (`ErrorSink`, #355) — real jq aborts instead; continuing is the
more useful of the two and is pinned as a deliberate divergence.

**Stage 5 — UTF-8 at the input boundary.** yq rejects at the same byte offset and exit code
as real yq, worded in this crate's own `YAML parse error:` convention rather than yq's
`bad file '-':` shape it has never matched. `yaml validate` grew the check too, reusing the
`InvalidUtf8` variant that had been *declared but never constructed* — the scanner reads
bytes and never decodes them, so nothing could ever have produced it. jq substitutes U+FFFD
and exits 0, byte-identical to jq 1.7.1; that also fixed two bugs in the other direction —
the lazy path wrote invalid UTF-8 to stdout, and the non-lazy path refused the file with
`Failed to read file` when the read had in fact succeeded.

Scope is document input only: `--raw-input` shares the jq path (jq substitutes there too);
DSV input, `--arg`/`--argjson` and `--rawfile` are untouched.

### Before / after

| | before | after |
|---|---|---|
| `{"a":"\ud800"} \| to_entries` | `[{"key":"a","value":null}]`, exit 0 | error, exit 5 |
| `{"a":"\ud800"} \| .a\|length` | `null (null) has no length` | `invalid unicode escape sequence` |
| `{"\ud800":1,"b":2} \| .b` | `null` | `2` |
| `{"\ud800":1,"b":2} \| to_entries` | one entry, silently | error, exit 5 |
| `[xyz123] \| to_entries` | `[{"key":0,"value":null}]` | error, exit 5 |
| `{"a":"\xff\xfe"} \| .a` | raw invalid UTF-8 on stdout | `"��"`, byte-identical to jq |
| yq `b: "x\qy"` `-o json` | `{"b": "xnull}` | `{"b":null}` |
| yq `"a\qb": 1` `-o json` | `{"a"":1}` | `{"":1}` |
| yq invalid-UTF-8 document | `{"a":1,"b":null}`, exit 0 | rejected, exit 1, same offset as real yq |
| `yaml validate` on same | exit 0 | exit 1 |

## Testing

**Automated Testing:**
- [x] All existing tests pass
- [x] New tests added for new functionality

**Manual Testing:**
- [ ] Tested on x86_64
- [x] Tested on aarch64/ARM (Apple M5 Max)

New regression tests, each verified to **fail against the previous behaviour**, not merely
to pass:

- `src/json/light.rs`, `src/yaml/light.rs` — `..._find_skips_undecodable_key_1247`
- `tests/jq_cli_tests.rs` — `..._does_not_hide_later_fields_1247`,
  `..._surfaces_as_error_1247`, `..._in_key_surfaces_as_error_1247`,
  `test_malformed_value_surfaces_as_error_1194`,
  `..._substitutes_replacement_char_1247`, plus two guards that valid escapes and valid
  multi-byte content are untouched
- `tests/yq_cli_tests.rs` — `..._does_not_corrupt_json_output_1247`,
  `..._does_not_hide_later_fields_1247`, `test_yq_rejects_invalid_utf8_document_1242`,
  `test_yq_accepts_multibyte_utf8_1242`
- `tests/yaml_validate_tests.rs` — `test_yaml_validate_rejects_invalid_utf8_1242`,
  `test_yaml_validate_accepts_multibyte_utf8_1242`

Two existing tests asserted the buggy behaviour and were **inverted**, not deleted:
`test_to_owned_degrades_to_null_on_string_decode_failure_1098` and
`test_materialize_degrades_to_empty_string_on_decode_failure_1098`.

### Test Commands

```bash
cargo test --features cli,simd,regex,serde --workspace --no-fail-fast
cargo clippy --all-targets --features std,simd,serde,cli,regex,bench-runner,large-tests,mmap-tests -- -D warnings
cargo fmt --all -- --check
cargo check --no-default-features   # no_std
```

All green locally: 54 test binaries, 0 failures.

## Performance Impact

- [ ] No performance impact
- [ ] Performance improvement
- [x] Potential performance regression (justify below)

Stage 5 adds a whole-input SIMD UTF-8 pass to every run. Interleaved A/B of the Stage 4
binary against the Stage 5 one, alternating order every repetition, min-of-9, gated first on
byte-identical stdout and exit code (all five cases identical):

| case | before | after | delta |
|------|--------|-------|-------|
| `jq '.users[0].name'` 1 MB | 5.37 ms | 5.13 ms | −4.5% |
| `jq '.users[0].name'` 8.4 MB | 17.23 ms | 17.96 ms | **+4.2%** |
| `jq '.'` 1 MB | 16.91 ms | 16.98 ms | +0.4% |
| `yq -o json '.'` 1 MB | 16.62 ms | 16.65 ms | +0.2% |
| `yq -o json '.'` 10 MB | 126.19 ms | 126.80 ms | +0.5% |

Worst case +4.2%, on the cheapest query at the largest size, where a roughly fixed ~1 ms
pass is the largest fraction of the work. That is inside the +8% ceiling the design doc set
*before* implementing. The −4.5% row is noise; the pass cannot make anything faster.

**This number is provisional and should not be treated as the gate.** It ran on an Apple M5
Max laptop under a load average near 10, which this repo's own benchmarking discipline
disqualifies. The formal interleaved A/B on the pinned bench machines (`johns-mac-mini` ARM,
`terminus` x86_64) has **not** been run.

The alternative — always-on strict validation, which the tier review floated — was costed
and rejected: 8.6 ms on the same 8.4 MB file, **+66%** on a cheap navigation query.

## Checklist

### For reviewers — two decisions worth challenging now

1. **Decode errors are catchable by `try`/`catch`.** `.a?` suppresses one; jq's parse error
   cannot be caught at all, because it happens before the program runs. I chose consistency
   with #1192's already-merged half. If these should be `Halt`-like instead, this is the
   cheap moment to change it. Recorded as Open Risk 1 in the design doc.
2. **jq substitutes U+FFFD rather than erroring.** Byte-identical to jq — but it is
   deliberately a *silent* substitution inside a change whose whole point was removing
   silent degradation. Defensible either way; flagging rather than burying it.

### Deliberately not in this PR

- **Stage 6** — the YAML streaming path still emits `null` for a bad *escape*. Its only
  error channel is `core::fmt::Result`, which carries no message. Every *materializing*
  route already raises. I wrote the equivalent fix for `print_json`'s arm and **reverted
  it**: bailing mid-stream produced a truncated document plus a generic exit 1, worse on
  every axis than the silent `null` it replaced. The reasoning is left in place at the site.
- **#1194's headline repro** `{invalid}` → `{}` — needs `JsonFields::uncons` to represent a
  malformed field, a public API contract change with its own call-site sweep.
- **#965** — this PR touches four of the five near-duplicate `→ OwnedValue` conversions and
  leaves them duplicated. Consolidating is a refactor with its own risk profile and should
  follow a behaviour change, not accompany one. #1283 already sequences it after this.

### Not everything that looked like a degrade was one

`length` on an object answers from `fields.len()` without decoding, and `keys_unsorted`
streams each key's raw byte span verbatim. Neither loses data — both are the same
raw-passthrough class as `jq '.'` — so both were left alone, with tests saying why.

### DIFF
--- a/benches/jq_format_bench.rs
+++ b/benches/jq_format_bench.rs
@@ -51,6 +51,7 @@ fn run_generic(json: &[u8], expr: &Expr) -> usize {
     let cursor = index.root(json);
     eval_generic::eval_with_cursor(expr, cursor)
         .collect_owned()
+        .expect("materializes")
         .len()
 }
 
--- a/benches/jq_generic_index_bench.rs
+++ b/benches/jq_generic_index_bench.rs
@@ -134,7 +134,7 @@ fn bench_generic_index_json(c: &mut Criterion) {
                 "depth {depth} level {i} fixture must not error"
             );
             assert!(
-                probe.collect_owned().is_empty(),
+                probe.collect_owned().expect("materializes").is_empty(),
                 "depth {depth} level {i} fixture must produce no output"
             );
         }
@@ -146,7 +146,7 @@ fn bench_generic_index_json(c: &mut Criterion) {
                 for expr in &exprs {
                     let cursor = index.root(black_box(json));
                     let result = eval_with_cursor(expr, cursor);
-                    total += result.collect_owned().len();
+                    total += result.collect_owned().expect("materializes").len();
                 }
                 black_box(total)
             });
@@ -184,7 +184,7 @@ fn bench_generic_index_yaml(c: &mut Criterion) {
                 "depth {depth} level {i} fixture must not error"
             );
             assert!(
-                probe.collect_owned().is_empty(),
+                probe.collect_owned().expect("materializes").is_empty(),
                 "depth {depth} level {i} fixture must produce no output"
             );
         }
@@ -199,7 +199,7 @@ fn bench_generic_index_yaml(c: &mut Criterion) {
                         .first_child()
                         .expect("fixture has exactly one document");
                     let result = eval_with_cursor_using::<YqSemantics, _>(expr, cursor);
-                    total += result.collect_owned().len();
+                    total += result.collect_owned().expect("materializes").len();
                 }
                 black_box(total)
             });
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -933,6 +933,41 @@ Pinned by [`test_generic_lazy_seq_first_after_map_skips_later_error_725`](../../
 [`test_first_over_lazy_seq_iterate_skips_later_error_1565`](../../../tests/jq_cli_tests.rs)
 (the `first(map(f) | .[] | g)` spelling, plus the draining counter-cases above).
 
+## Deliberate divergences (ADR-0018 rule 4)
+
+### A structurally malformed value doesn't abort the rest of a multi-value stream — no carve-out; this one is out of policy
+
+Real jq DOM-parses the whole input before evaluating anything, so a parse error on one
+value in a multi-value stream kills the run outright — nothing after the bad value is
+even attempted:
+
+```console
+$ printf '1\n[xyz123]\n3\n' | jq -c 'to_entries'
+jq: error (at <stdin>:1): number (1) has no keys
+jq: parse error: Invalid numeric literal at line 2, column 8
+$ printf '1\n[xyz123]\n3\n' | succinctly jq -c 'to_entries'
+jq: error (at <stdin>:1): number (1) has no keys
+jq: error (at <stdin>:2): unexpected character
+jq: error (at <stdin>:3): number (3) has no keys
+```
+
+succinctly never parses the whole input up front (semi-indexing is per-record), and a
+structurally malformed value (#1194) or an undecodable string (#1247) now surfaces
+through the same `EvalError`/`ErrorSink` per-record-error-then-continue convention every
+other jq error already uses (`ErrorSink`, [#355](https://github.com/rust-works/succinctly/issues/355))
+— so `3` still gets processed here where real jq never reaches it. This is not new to
+[#1247](https://github.com/rust-works/succinctly/issues/1247): `ErrorSink`'s
+continue-past-one-error batch semantics predate it and already diverge from real jq's
+abort-on-parse-error for every ordinary evaluation error, not just a decode failure.
+[#1247](https://github.com/rust-works/succinctly/issues/1247)'s own design doc
+([`docs/plan/decode-failure-routing.md`](../../plan/decode-failure-routing.md), Stage 4)
+calls succinctly's behaviour "the more useful of the two" and treats it as settled — but
+none of ADR-0018's three permitted conditions actually cover it (the output is readable,
+nothing is corrupted or discarded, and the process does not die either way), so per rule 4
+it is recorded here as a still-open policy question, matching
+[yq Limitations](../yq/limitations.md)'s "Merge-flag `+` and `d` combined" precedent for a
+divergence accepted on its merits without fitting the letter of rule 4.
+
 ## Provenance
 
 | Artifact           | Path                                                                                                       |
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -595,6 +595,43 @@ Live-verified against yq v4.53.3. Pinned as known gaps (current, not the desired
 state) by `test_yq_auto_output_mixed_format_multi_file_known_gap_1493` and
 `test_yq_auto_output_slurp_mixed_format_known_gap_1493` in `tests/yq_cli_tests.rs`.
 
+### A bad escape in a streamed scalar still degrades silently instead of raising
+
+[#1247](https://github.com/rust-works/succinctly/issues/1247) routed almost every
+decode-failure materialization path through a real `EvalError` — but two streaming output
+writers were left as a deliberately-deferred gap ("Stage 6" in
+[`docs/plan/decode-failure-routing.md`](../../plan/decode-failure-routing.md)): their only
+error channel is `core::fmt::Result`, which carries no message, so a mid-stream failure
+there can only be silently absorbed or cause a bare, undiagnosed abort. Both still emit a
+substituted `null`/`""` at exit 0 for a value or key with a bad escape, live-verified
+against the same build that fixed every other #1247 site:
+
+```console
+$ printf 'a: 1\nb: "bad \q escape"\n' | succinctly yq -o=json '.'
+{
+  "a": 1,
+  "b": null
+}
+$ printf 'a: 1\nb: "bad \q escape"\n' | yq -o=json '.'
+Error: bad file '-': yaml: while scanning a quoted scalar at line 2, column 4: line 2, column 9: found unknown escape character
+
+$ printf 'double: "quoted \q scalar"\n' | succinctly yq '.'
+double: ""
+$ printf '"a\qb": 1\nb: 2\n' | succinctly yq '.'
+"": 1
+b: 2
+```
+
+The first pair (`-o=json`, `stream_json_value`/the YAML→JSON transcoders) is the case the
+design doc names. The second and third (default YAML→YAML output, no `-o` flag —
+`stream_yaml_string_value`/`write_yaml_field_key`, the single most common `yq`
+invocation) are the same gap on its YAML-target sibling, not previously recorded anywhere.
+Every *materializing* route (a `--arg`-forced DOM, a multi-result filter, `to_entries`,
+`length`) already raises correctly; only these two purely-streamed writers do not. Fixing
+either needs `stream_json_value`/`stream_yaml_value` and their callers to carry a richer
+error than `fmt::Error` — the design doc's own stated reason Stage 6 was split out and
+deferred rather than attempted alongside the rest of #1247's landed stages.
+
 ### Other categories
 
 Float and number formatting ([#1071](https://github.com/rust-works/succinctly/issues/1071),
--- a/docs/plan/README.md
+++ b/docs/plan/README.md
@@ -26,6 +26,7 @@ These plans are kept for:
 | [jq-lazy-generator-consumers.md](jq-lazy-generator-consumers.md)     | Partial     | `src/jq/eval.rs`                  | Demand-driven sink for short-circuiting consumers |
 | [jq-duplicate-key-collapse.md](jq-duplicate-key-collapse.md)         | Implemented | `src/jq/`, `jq_runner.rs`         | jq-mode duplicate object key collapse             |
 | [jq-generator-argument-fanout.md](jq-generator-argument-fanout.md)   | Implemented | `src/jq/eval.rs`                  | One builtin result per generator-argument output  |
+| [decode-failure-routing.md](decode-failure-routing.md)               | Proposed    | `src/jq/`, `src/yaml/`            | Decode-failure error routing                      |
 
 ## Archived Plans
 
--- a/docs/plan/decode-failure-routing.md
+++ b/docs/plan/decode-failure-routing.md
@@ -1,6 +1,6 @@
 # Decode-failure error routing for `to_owned`/`cursor_to_owned` (#1247, #1242, #1194)
 
-**Status: partially implemented — Stages 1–3 landed, 4–6 outstanding.** This
+**Status: partially implemented — Stages 1–4 landed, 5–6 outstanding.** This
 document is the deliverable for
 [#1247](https://github.com/rust-works/succinctly/issues/1247), which was tiered Tier 3
 ("needs a design decision before implementation, not a same-pattern continuation of the
@@ -427,8 +427,22 @@ Implementation notes, all discovered while doing it:
   verbatim. Neither loses data — both are the same raw-passthrough class as `jq '.'` —
   so both were left alone, and the tests say why.
 
-**Stage 4 — the `StandardJson::Error` arms (sites 17–20).** Fixes `[xyz123] → [null]`,
-i.e. #1194's in-scope half. Small once Stage 3 exists.
+**Stage 4 — the `StandardJson::Error` arms (sites 17–20). ✅ landed.** Fixes
+`[xyz123] → [null]`, i.e. #1194's in-scope half. Small once Stage 3 exists: the arms
+raise with the semi-index's own message, which is more specific than anything
+reconstructible at the materializer.
+
+One site was reverted after being written. `print_json`'s `StandardJson::Error` arm
+(`jq_runner.rs`) walks child cursors lazily and streams as it goes, so by the time a
+nested error is reached it has already written the opening `[`. Bailing there produced a
+truncated document plus a generic exit 1, where every materializing route gives a clean
+diagnostic and exit 5 — worse on every axis than the silent `null` it replaced. Left as
+`null` with the reasoning in place, and folded into Stage 6, which is the same problem.
+
+Evaluation still continues past the bad value, so the good documents either side of it in
+a multi-value stream are still processed (`ErrorSink`, #355). Real jq aborts the whole run
+at the parse error instead; succinctly's behaviour here is the more useful of the two and
+is a deliberate, tested divergence.
 
 **Stage 5 — UTF-8 at the boundary (sites 21–22, #1242).** yq rejects, jq replaces with
 U+FFFD, `yaml validate` grows a UTF-8 check. *Why last:* it is the only stage that adds
--- a/docs/plan/decode-failure-routing.md
+++ b/docs/plan/decode-failure-routing.md
@@ -494,6 +494,18 @@ otherwise already correct, and it can be reverted independently if the perf gate
   `Failed to read file` when the read had in fact succeeded.
 - Scope is document input only. `--raw-input` shares the jq path (jq substitutes there
   too); DSV input, `--arg`/`--argjson` and `--rawfile` are untouched.
+- **`--validate` is excluded, and getting that wrong was a live regression** caught in
+  review. The substitution originally ran at read time, before `validate_json_input`, so
+  the strict validator was handed an already-repaired document: `sjq --validate` exited 0
+  on a file `succinctly json validate` still rejects with exit 1, silently dropping RFC
+  8259 §8.1's mandatory UTF-8 check from the one flag whose entire purpose is strictness —
+  and contradicting this document's own "What the existing validators already catch" table
+  two sections up. Both routes had it (the lazy path substituted into `raw_inputs`; the
+  materializing one decoded to a `String` in `get_inputs` first). Fixed by skipping the
+  substitution under `--validate` entirely: nothing is lost, because any document that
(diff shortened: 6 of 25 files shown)


newhoggy:
## The formal interleaved A/B, on both pinned machines

Ran the measurement this PR names as outstanding ("This number is provisional and
should not be treated as the gate"). Method is `scripts/ab-cli.py` per
`docs/guides/benchmarking.md` § A/B Benchmarking Method: the two binaries alternate
*within* each repetition, reps=9, output identity gated before any timing is read, and
a `--control` noise floor measured per group per machine.

**Headline: as committed, this breaches the design doc's +8% ceiling on x86_64 — but
only ~1 point of that is the UTF-8 pass. The rest is an unintended whole-document copy
in the same commit, and a 6-line change removes it.**

Pair under test: `201c3db14` (Stage 4) → `c97c66d7f` (Stage 5). Both built natively per
machine, `--release --features cli,simd`. Corpus generated by the crate's own generators
(`json/yaml generate`, seed 1242): JSON `users` 1.1–9.0 MB, YAML `users` 1–10 MB (0.00%
non-ASCII), YAML `unicode` 10 MB (45.4% non-ASCII).

### Control floors (median of medians)

| machine | jq | yq users | yq unicode |
|---|---|---|---|
| Apple M4 Pro (`johns-mac-mini`) | −0.12% | −0.52% | −0.70% |
| AMD Ryzen 9 7950X (`terminus`) | −0.52% | −0.39% | −0.73% |

### Stage 4 → Stage 5, as committed

| group | M4 Pro | 7950X |
|---|---|---|
| `jq` on JSON | +0.94%  (+0.3..+1.1) | **+9.47%**  (+3.8..+17.9) |
| `yq -o json` on ASCII YAML | −0.43% | +0.46% |
| `yq -o json` on 45%-non-ASCII YAML | **+3.57%**  (worst +4.8%) | +0.06% |

Worst single row: `users-8mb.json .users[0].name` on the 7950X, 29.6 ms → 34.0 ms
(+14.8% median, +8.8% min); `users-1mb.json .users[0].name` is +17.9% median. 4/4 rows
slower, so the sign is not noise. Identity gate: 0 differences on every configuration,
both machines.

Note the two architectures fail in *opposite* places — x86 on jq, ARM on non-ASCII yq —
which is why one laptop run could not have found either.

### Attribution: it is mostly not the UTF-8 pass

Rebuilt Stage 5 with `utf8_lossy_document`'s validation disabled and measured against
both sides (7950X, `jq` group, median of medians):

| comparison | delta | what it prices |
|---|---|---|
| Stage 4 → Stage 5 *without* validation | **+5.89%** | everything else in the commit |
| Stage 5-no-validate → Stage 5 | **+1.04%** | the UTF-8 pass alone |

So the always-on pass this PR is defending costs about **1%**. The other ~6 points come
from the only other change in `jq_runner.rs`: `read_stdin`/`read_file` moving from
`read_to_string` to `String::from_utf8_lossy(&bytes).into_owned()`. On a *valid*
document `from_utf8_lossy` returns `Cow::Borrowed`, so `.into_owned()` allocates and
memcpys the entire document — a second full copy that `read_to_string` never made. It is
invisible on macOS and expensive on Linux, which is the usual "memory-bound effects do
not port across architectures".

### Proposed fix — verified

```rust
/// Lossy UTF-8 decode that does not copy a document that is already valid.
///
/// `String::from_utf8_lossy(&bytes).into_owned()` allocates and copies the whole
/// document even when it is valid, because the `Cow` it returns is `Borrowed` and
/// `into_owned` must then clone it. Taking ownership of the existing buffer instead
/// makes the valid path allocation-free.
fn utf8_lossy_string(raw: Vec<u8>) -> String {
    match String::from_utf8(raw) {
        Ok(s) => s,
        Err(e) => String::from_utf8_lossy(e.as_bytes()).into_owned(),
    }
}
```

with `read_stdin` / `read_file` calling it instead of `String::from_utf8_lossy(..).into_owned()`.

| comparison | 7950X | M4 Pro |
|---|---|---|
| Stage 5 → Stage 5+fix | **−7.04%**  (worst −13.2%) | +0.45% (neutral) |
| Stage 4 → Stage 5+fix | **+0.48%** — inside the control floor | +0.83% |

Identity gate clean, and the behaviour this PR added is preserved: `{"a":"x\xe4y"}`
still prints `x\u{fffd}y` at exit 0 from both a file and stdin, byte-identical to
Stage 5 and to jq 1.7.1.

### The one real cost that remains: non-ASCII YAML on ARM

`text::utf8::validate_utf8` uses AVX2 on x86_64 and falls through to
`validate_utf8_scalar` on aarch64 — there is no NEON arm. Timing the pass alone on the
M4 Pro (`text validate utf8`, net of process startup):

| content | throughput |
|---|---|
| 10 MB YAML, 0.00% non-ASCII | 9.35 GB/s |
| 10 MB YAML, 45.4% non-ASCII | **3.34 GB/s** |

2.8× dearer per byte, because the 8-byte ASCII skip stops paying. 3.14 ms against a
47.5 ms query predicts +6.6%; measured +4.8% (+5.8% net of control). That is under the
+8% ceiling, so it does not block — but it is the largest true cost of the pass, it is
ARM-only, and it is on a document shape the PR's table does not contain. A NEON arm for
`validate_utf8` would be the targeted fix if it ever matters.

### Caveats

- Generated corpora, not the real-workload corpus; `unicode` is one generator pattern at
  45% non-ASCII, not a survey of real multi-byte documents.
- The fix above was measured and behaviour-spot-checked, not put through the full test
  suite — it is a proposal, not a tested commit.
- `ab-cli.py`'s idleness check was overridden on both machines: on macOS it sums `ps`
  *lifetime* CPU averages, so my own SSH login's `sshd-session` + `opendirectoryd`
  read as ~33%; on Linux each run's own CPU work pushes the 1-minute load average over
  its threshold for the next run. Both machines were verified idle independently
  (98.7% idle / load 0.15 at start) and on AC power.


---

**Commit mapping after the 2026-08-25 rebase.** The pair above was measured as
`201c3db14 → c97c66d7f`; the rebase onto `d61ef1754` renumbered them to
**`4713aa7cb → ea735849c`**. Stage 5's source patch came through byte-identical
(`diff <(git show c97c66d7f -- src/) <(git show ea735849c -- src/)` is empty but for
hunk offsets), and `read_stdin`/`read_file` still carry the
`String::from_utf8_lossy(..).into_owned()` copy the attribution blames, so the numbers
and the proposed fix carry over unchanged. Only the absolute before/after milliseconds
were taken on the older `bb2b6b532` base — the deltas are a fixed per-document cost and
do not depend on it. Not re-run.

newhoggy:
## #1194's structural half is now one arm away from you — and its detection is already on `main`

#1608 merged (`f977dfb33`), so this branch's rebase picked it up. Flagging the interaction while
you're in these files, because this PR turns out to be **most of the way to closing the rest of
#1194** without meaning to.

### What #1608 left on `main` for you

- `JsonFields::unpaired_tail` / `ends_unpaired` (`json/light.rs`) — the O(1) "this object's
  children don't pair" test. On an exhausted list it answers from a `None` cursor without
  touching the BP tree.
- **`DocumentFields::ends_unpaired`** (`jq/document.rs`) — same question on the trait, defaulting
  to `false`. That default is the load-bearing part for you: JSON is the only format that can
  present an unpaired child, so every other format costs nothing and needs no arm.
- `DistinctKeyCursors::ended_unpaired` — recorded *as the walk goes*, so asking is a field read
  rather than a second pass (which `perf-guard.py`'s `wide_keys_unsorted` workload would have
  caught).
- `EvalError::malformed_json_text(text)` — re-runs the strict validator to name the real error
  (`expected string key, found 'i'`), on the failure path only.

### The gap in this PR

Making the family fallible was the whole architectural blocker #1194 and #1247 shared, and you've
done it — `to_owned`/`to_owned_at_depth`/`to_owned_cursor` all return `Result` now. But
`to_owned_at_depth`'s object arm only asks about *decode* failures:

```rust
while let Some((field, rest)) = f.uncons() {
    if let Some(reason) = field.key.string_decode_error() { return Err(...) }
    if let Some(key) = field.key_str() {            // <- a non-string key silently drops
        map.insert(key.into_owned(), to_owned_at_depth(&field.value, depth + 1)?);
    }
    f = rest;
}
Ok(OwnedValue::Object(map))                          // <- no `f.ends_unpaired()` check
```

`grep -c ends_unpaired src/jq/eval_generic.rs` on this branch is **0**. So `{invalid}` still
materializes as `{}` and `{123: 1, "b": 2}` still drops a field here — the two shapes #1608 fixed
on the cursor-writing paths but could not reach through this family.

### One honest complication, so this isn't oversold

The `ends_unpaired()` check is trivial and the trait default makes it YAML-safe. The *message* is
not: `DocumentCursor` exposes `text_position()` but no `text()`, so this generic path cannot reach
the document bytes to run the validator the way the JSON-specific sites do. Options, roughly in
increasing cost: a fixed string here (`"Invalid JSON text: object member missing a value"` —
correct, just less precise than `--validate`'s wording); add `text()` to `DocumentCursor`; or
leave the generic arm alone and do it on JSON-specific `cursor_to_owned` (`lazy.rs`) in a #1194
follow-up, which does have `text()`.

**Entirely your call whether any of that belongs in this PR** — it is already 2400 lines and
"don't grow it further" is a defensible answer. Mostly I want the next person not to have to
rediscover that the detection is sitting there unused.

### Your CI is safe from #1608

All ten `_1194` tests survived your rebase unmodified, and two of them
(`test_jq_malformed_object_still_quiet_once_materialized_1194`,
`test_jq_malformed_object_quiet_off_the_lazy_path_1194`) deliberately **pin the quiet behaviour**
— they assert exit 0 for `-a`/`-S`/`--slurp` and for filter shapes that lose their cursor. Since
this branch doesn't wire the structural signal, they still pass. If you *do* wire it, expect those
two to fail: that is by design, and each carries a note saying to invert it and delete the
matching paragraph in `docs/compliance/jq/limitations.md`.

Separately: `--ascii-output` sets `can_use_lazy_path = false`, so a formatting flag decides whether
a malformed document is accepted at all. Unrelated to this PR, filed on its own.

PR #1395, opened 2026-08-21
refactor(jq): name the exact/saturated exponent shape, unify sign-stripping

### LINKED ISSUE #1304: jq: value.rs's saturated-exponent path has structural duplication and an unbounded raw echo
**Severity: Low**

## Summary

Follow-up from #1273's review, split off as out-of-scope maintainability/
design concerns rather than correctness bugs (all fixed in #1273's PR
#1300). Grouping four related findings from that review round:

## 1. Positional tuple return shape invites a silent field-mixup

`parse_literal_exponent`'s `(i128, bool)` and
`normalize_extreme_literal_mantissa`'s `Ok((String, i128, i128, bool))`
distinguish "exact" vs "saturated" only via a trailing `bool` literal at
each match site (`Ok((mantissa_str, _, _, true))` vs
`Ok((mantissa_str, new_exp, _, false))` in `format_near_zero_literal`). A
future edit that swaps which arm gets `true`/`false` compiles cleanly and
silently inverts exact/saturated handling.

**Suggested fix direction**: a small enum (`enum ExpParse { Exact(i128),
Saturated(i128) }`, or wrapping `normalize_extreme_literal_mantissa`'s `Ok`
payload in a named struct) so each call site matches a named variant
instead of a positional literal.

## 2. Three independent sign-stripping idioms

`split_mantissa` (`strip_prefix(['-', '+'])`), `parse_literal_exponent`
(`trim_start_matches('+')`), and the new
`assemble_scientific_from_raw_exponent` (`strip_prefix('+')` then
`strip_prefix('-')`) each reimplement "peel a leading sign off digit text"
slightly differently -- currently masked only by the lexer guaranteeing at
most one leading sign character. Similarly, `assemble_scientific` (numeric
`exp >= 0` check) and `assemble_scientific_from_raw_exponent` (string
`starts_with`/`strip_prefix` check) independently reimplement the same
E+/E- display convention over different representations (`i128` vs raw
text).

**Suggested fix direction**: a single shared helper for "does this digit
text represent a negative value" and/or "assemble sign + mantissa + E +
exponent-sign + exponent-text", per this file's own #106 precedent
("duplicated predicates diverge silently").

## 3. Unbounded raw-exponent echo

`assemble_scientific_from_raw_exponent` echoes the *entire* raw exponent
digit string with no length cap, unlike the mantissa's own
`MAX_RENDERED_MANTISSA_DIGITS = 100_000`-bounded, truncate-with-honesty
treatment for the identical "document-controlled, unbounded-length digit
string" problem (`normalize_extreme_literal_mantissa`'s own doc comment
explicitly calls out not letting this "turn into unbounded work here").

**Why this may be fine as-is**: output size is proportional to input size
(no amplification), and this crate's stated design philosophy elsewhere is
to preserve a literal's full given precision unconditionally rather than
truncate. Worth a deliberate decision (cap it like the mantissa, or
document why an exponent doesn't need the same treatment) rather than the
current implicit "nobody decided" state.

## 4. Shift-adjustment is a known, undocumented-at-a-glance limitation

`assemble_scientific_from_raw_exponent`'s doc comment already discloses
that the echoed exponent isn't shift-adjusted (would need real big-integer
arithmetic), so `123e-<huge>` prints an exponent numerically off by the
mantissa's own `shift` (2, in that example) from what a fully-adjusted
answer would show. This is intentional and tested (see
`test_format_number_jq_compat_multidigit_mantissa_exponent_beyond_i128_range_1273`),
listed here only so a future big-integer-arithmetic pass (if ever
undertaken) has a pointer to what it would need to fix.

## Suggested fix direction

Items 1-2 are mechanical refactors with no behavior change -- good
candidates for a single follow-up PR. Item 3 needs a decision first
(cap or don't, and why). Item 4 needs real arbitrary-precision arithmetic
if ever pursued, likely not worth it given the severity (`Low`) of the
underlying issue.

## Related

#1273 (the original sentinel-leak fix, PR #1300), which found these during
a 10-agent review round while fixing two confirmed correctness bugs
(secondary saturation via `saturating_add`/`saturating_sub`, and a
non-canonical leading-zero echo -- both fixed in the same PR, not part of
this follow-up).

newhoggy:
## Summary

Fixes #1304. Follow-up from #1273's review, split off as maintainability/design
concerns rather than correctness bugs — this is a behavior-preserving refactor plus
one recorded decision, not a functional change.

## Item 1: positional tuple return shape invites a silent field-mixup

`parse_literal_exponent`'s `(i128, bool)` and `normalize_extreme_literal_mantissa`'s
`Ok((String, i128, i128, bool))` distinguished "exact" vs "saturated" only via a
trailing `bool` literal matched positionally at the one call site that cares
(`format_near_zero_literal`: `Ok((mantissa_str, _, _, true))` vs `Ok((.., false))`).
A future edit swapping which arm got `true`/`false` would compile cleanly and
silently invert exact/saturated handling.

Fixed with a small `ExpParse` enum (`Exact(i128)` / `Saturated(i128)`) and a named
`NormalizedMantissa` struct replacing the tuple. The risky call site now matches
`ExpParse::Saturated(_)` / `ExpParse::Exact(new_exp)` by variant name — a swap is a
compile error, not a silent inversion.

## Item 2: three independent sign-stripping idioms

`split_mantissa`, `parse_literal_exponent`, and `assemble_scientific_from_raw_exponent`
each reimplemented "peel a leading sign off digit text" slightly differently, correct
today only because the lexer guarantees at most one leading sign character. Unified
into one `strip_leading_sign` helper (this file's own #106 precedent).

Also unified `assemble_scientific` (numeric `exp >= 0` check, relying on `i128::Display`'s
own embedded `-`) and `assemble_scientific_from_raw_exponent` (string-based sign check)
onto a shared `assemble_scientific_with_sign`, so the `E+`/`E-` display convention can't
independently drift between the two representations either. `exp.unsigned_abs()`, not
`.abs()`, stays correct at `exp == i128::MIN` (a real, reachable saturation sentinel here).

## Item 3: unbounded raw-exponent echo — decision recorded

`assemble_scientific_from_raw_exponent` echoes the *entire* raw exponent digit string
with no length cap, unlike the mantissa's `MAX_RENDERED_MANTISSA_DIGITS`-bounded
treatment. Per the tier review's framing, this needed one decision, not a design pass.

**Decision: leave uncapped, documented why.** Measured (not assumed) a 5,000,000-digit
saturated exponent through `dump_truncated`'s own bounded-preview path: 0.17s, linear
in input size, not amplified. Two things distinguish this from the mantissa case:
- This path is already gated behind an exponent that has overflowed `i128` (~39+ digits
  minimum) — unlike the mantissa, unbounded from the very first digit of *any* ordinary
  literal, so capping it protects the overwhelmingly common case; there's no comparably
  common case to protect here.
- A cap would cost real correctness on the *ordinary* (non-preview) render path, which
  — matching this crate's stated preserve-every-given-digit philosophy — has no reason
  to truncate an exponent it can otherwise echo exactly.

Added a regression test (`dump_truncated_bounds_a_near_zero_literal_with_a_huge_saturated_exponent`)
pinning that `dump_truncated`'s own output still stays properly bounded regardless — the
echo itself is uncapped, but the `PreviewSink` still clamps what's actually kept.

## Item 4: shift-adjustment — left as-is

Already an explicitly documented non-goal with its own pinning test
(`test_format_number_jq_compat_multidigit_mantissa_exponent_beyond_i128_range_1273`).
Nothing to do here per the issue's own framing.

## Verification

- Full existing test suite: byte-identical, 0 regressions (2696 lib tests)
- Broad live sweep of ~35 number-literal shapes (overflow, underflow, saturated
  exponents, ordinary values, boundary cases like the #1273 shift-into-saturation
  example) comparing this branch's binary against a pre-refactor build of `main` —
  zero mismatches
- New `ExpParse`/`strip_leading_sign`/`assemble_scientific_with_sign` code covered by
  the existing test suite exercising every branch (no new logic added, pure restructuring)

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 2696 lib tests, 0 failures)
- [x] `cargo fmt --check`
- [x] `omni-dev coverage diff` (no new executable lines — pure refactor)
- [x] Live-verified against a pre-refactor build across ~35 number-literal shapes, zero mismatches

### DIFF
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -1186,6 +1186,29 @@ mod tests {
         assert!(result.ends_with("..."), "must be truncated: {result}");
     }
 
+    /// #1304: unlike the mantissa above, `assemble_scientific_from_raw_
+    /// exponent` (reached via `format_near_zero_literal` for a saturated
+    /// exponent) deliberately does *not* cap the raw exponent digit string
+    /// it echoes -- see that function's own doc comment for the measured
+    /// reasoning. Pins that `dump_truncated`'s own output still stays
+    /// bounded regardless: the echo itself is uncapped, but
+    /// `dump_truncated`'s `PreviewSink` still clamps what actually gets
+    /// kept, the same as it already does for every other value shape.
+    #[test]
+    fn dump_truncated_bounds_a_near_zero_literal_with_a_huge_saturated_exponent() {
+        let exponent = "9".repeat(100_000);
+        let literal: Box<str> = format!("0.005e-{exponent}").into();
+        let value = OwnedValue::NumberLiteral(super::super::value::NumberRepr::Float(0.0), literal);
+        let result = dump_truncated(&value);
+        assert!(
+            result.len() < 100,
+            "must not render anywhere near the full 100,000-digit exponent: got {} bytes",
+            result.len()
+        );
+        assert!(result.starts_with("5E-99999999"), "got: {result}");
+        assert!(result.ends_with("..."), "must be truncated: {result}");
+    }
+
     /// #930 review: `from_value`'s non-string branch used `OwnedValue::to_json`,
     /// which is the *real-output* convention (RFC-8259-safe `null` for a
     /// non-finite float) - wrong for `error(v)`'s message, which (like
--- a/src/jq/value.rs
+++ b/src/jq/value.rs
@@ -352,17 +352,21 @@ pub fn format_number_jq_compat(raw: &[u8]) -> String {
     // invariant `format_overflow_literal_mantissa` relies on for its own
     // `unreachable!()`.
     //
-    // `exp_saturated` (`_`, ignored): `s` already parsed successfully to a
-    // finite, nonzero `f64` to reach this path at all (subnormal included,
-    // since #1206 folded that case in here too), so its written exponent
-    // digit string was in `i128`'s range long before it was in `f64`'s --
-    // unlike `format_near_zero_literal` (#1273), this path can't actually
-    // observe a saturated exponent in practice.
-    let Ok((mantissa_str, shifted_exp, digit_count, _exp_saturated)) =
-        normalize_extreme_literal_mantissa(s, exp_pos, Some(MAX_RENDERED_MANTISSA_DIGITS))
+    // `new_exp`'s saturation (ignored via `.value()`): `s` already parsed
+    // successfully to a finite, nonzero `f64` to reach this path at all
+    // (subnormal included, since #1206 folded that case in here too), so
+    // its written exponent digit string was in `i128`'s range long before
+    // it was in `f64`'s -- unlike `format_near_zero_literal` (#1273), this
+    // path can't actually observe a saturated exponent in practice.
+    let Ok(NormalizedMantissa {
+        mantissa_str,
+        new_exp,
+        digit_count,
+    }) = normalize_extreme_literal_mantissa(s, exp_pos, Some(MAX_RENDERED_MANTISSA_DIGITS))
     else {
         unreachable!("nonzero value implies normalize_extreme_literal_mantissa succeeds")
     };
+    let shifted_exp = new_exp.value();
     let sign = if value.is_sign_negative() { "-" } else { "" };
     if shifted_exp == 0 || (-6..=-1).contains(&shifted_exp) {
         // #1274: this window is *unconditionally* a decimal render --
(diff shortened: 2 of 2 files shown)


newhoggy:
Pushed a follow-up commit addressing both review findings.

**Fixed — real: `format_near_zero_literal`'s `Err` arm still forked on a plain
boolean.** The `Ok` arm properly threads its exact-vs-saturated decision through
`NormalizedMantissa.new_exp: ExpParse` and matches it exhaustively — but the `Err`
arm (the all-zero-mantissa case) computed the equivalent "is the folded, shifted
exponent saturated" question via `if parsed_exp.is_saturated() || sub_saturated`, a
plain boolean condition that could still be silently inverted by a future edit (the
exact class of bug `ExpParse`'s own exhaustive match exists to prevent — just
relocated to this arm rather than actually closed). Now folds `shifted_exp` into an
`ExpParse` and matches it exhaustively too, structurally identical to the `Ok` arm.

The review's broader framing (5 consumption sites, only 1 protected) also flagged
two sites that call `.value()` and discard saturation entirely (`format_number_jq_
compat`'s ordinary path, `format_overflow_literal_mantissa`) — I left those as-is: they
aren't decision forks at all (no `if saturated {...} else {...}` exists there to
misroute), they're a deliberate, already-documented simplification where the
distinction genuinely doesn't affect the outcome. Introducing exhaustive matching
there would be ceremony without a real fork to protect.

**Fixed — real: a fourth sign-stripping copy survived the consolidation.**
`strip_insignificant_leading_zero_and_plus` (this same file, `strip_prefix('-')`/
`strip_prefix('+')`) was structurally identical to the three copies `strip_leading_
sign` was introduced to unify, but got missed in the original pass — leaving one
copy standing in the very file the #106 consolidation was cleaning up. Folded it in
too, and corrected `strip_leading_sign`'s own doc comment (which had claimed "three").

Reuse/Efficiency angles came back clean — `ExpParse` is `Copy`, zero-copy string
slicing throughout, `NormalizedMantissa` correctly doesn't derive `Copy`/`Clone`, and
`assemble_scientific_with_sign`'s `impl Display` parameter avoids an extra allocation
on the numeric call site. No other duplicate implementation of either pattern exists
elsewhere in the crate (checked `src/yaml/scalar.rs` and `src/jq/eval.rs`'s own
sign-handling — different problems, not duplicates).

Re-verified everything after this commit: full test suite (2696 lib tests, 0
failures), both CI clippy legs, `cargo fmt --check`, `cargo check --no-default-features`,
and re-ran the live sweep (~35 number-literal shapes plus 9 leading-zero/plus cases)
against a pre-refactor build of `main` — zero mismatches, still byte-identical output.
Coverage: 97% patch (65/67 new lines); the 2 uncovered lines are an already-provably-
unreachable `unreachable!()` invariant and an `assert!` failure-message branch that
only executes if the test itself fails.

Issue #1061, opened 2026-08-16
jq: fold_index_key's i64::MAX-boundary check silently off-by-one folds .[2^63] to i64::MAX


newhoggy:
## Summary

`fold_index_key`'s float/exponent index-folding branch (in `src/jq/parser.rs`, both the pre-existing `Literal::Float` arm and the `Literal::NumberLiteral` arm added by #1035/#1059) bounds-checks a float index value against `i64::MAX as f64`. Since `i64::MAX` (9223372036854775807) isn't exactly representable as `f64`, `i64::MAX as f64` rounds *up* to `2^63` (9223372036854775808.0) — one past the real maximum. A float index of exactly `9223372036854775808.0` therefore passes the bounds check, and the subsequent `f as i64` cast then *saturates* to `i64::MAX`, silently folding the index one lower than what was written.

## Repro

```
$ succinctly jq '.[9223372036854775808]' # not float-shaped, unaffected — this needs a float/exponent spelling to hit the bug
```

Confirmed via a direct unit-level probe (not yet a filed regression test) that `parse(".[9223372036854775808.0]")` folds to `Expr::Index(9223372036854775807)` instead of erroring or otherwise flagging the boundary miss.

## Root cause

```rust
Expr::Literal(Literal::Float(f))
    if f.fract() == 0.0 && *f >= i64::MIN as f64 && *f <= i64::MAX as f64 =>
{
    Some(Expr::Index(*f as i64))
}
```

`i64::MAX as f64` == `9223372036854775808.0` (rounds up), not `9223372036854775807.0`. The correct comparison needs to reject values `>= 2^63`, not `<= i64::MAX as f64`.

## Scope note

Pre-existing on `main` (the `Literal::Float` arm has had this since before #1035); #1035/#1059 duplicated the identical flawed check into a new `Literal::NumberLiteral` arm rather than introducing it. Filed as a follow-up rather than fixed in #1059 since it's an inherited bug, not a regression that PR introduced, and #1059 was already large in scope.

newhoggy:
🔧 Claimed by agent `denna` — branch `issue-1061-fold-index-key-i64max-boundary`

PR #1400, opened 2026-08-21
fix(jq): keep a float slice bound's own spelling in path() output

### LINKED ISSUE #1326: jq: path() reports a float slice bound as the resolved integer, unlike jq which keeps the bound as written
## Summary

Found while fixing #1088, which fixed the same divergence for a float array *index*. `Expr::Slice { start: Option<i64>, end: Option<i64> }` (`src/jq/expr.rs`) has the identical problem for slice *bounds*: it can only hold `i64`, so a float bound is collapsed to the integer the slice resolves against, and `path()` reports the resolved bound rather than the written one.

```
$ echo '[1,2,3,4,5]' | jq -c 'path(.[1.5:3.5])'
[{"start":1.5,"end":3.5}]
$ echo '[1,2,3,4,5]' | succinctly jq -c 'path(.[1.5:3.5])'
[{"start":1,"end":4}]
```

Same for every float spelling, including whole-valued and negative ones, and for a genuinely dynamic bound (so this is not specific to the parser's `fold_int_literal` fast path):

```
path(.[1.0:3.0])                                jq [{"start":1.0,"end":3.0}]   succ [{"start":1,"end":3}]
path(.[1.2:3.2])                                jq [{"start":1.2,"end":3.2}]   succ [{"start":1,"end":4}]
path(.[-1.5:])                                  jq [{"start":-1.5,"end":null}] succ [{"start":-2,"end":null}]
((1.5) as $a | (3.5) as $b | path(.[$a:$b]))    jq [{"start":1.5,"end":3.5}]   succ [{"start":1,"end":4}]
```

## The `end: 4` is not a second bug

An earlier note on #1088 wondered whether succinctly was truncating inconsistently, since `3.5` comes out as `4` rather than `3`. It isn't: jq's slice rule floors the start bound and ceils the end bound, and the **value** side agrees exactly in both (`[1,2,3,4,5] | .[1.5:3.5]` is `[2,3,4]` in jq and here). Succinctly is reporting the correctly *resolved* pair `1`/`4`; jq reports the pair the user *wrote*. One divergence, not two.

## Rule to reproduce

The same one #1088 established for indices, and it needs no special cases: **`path()` reports the bound verbatim, including its number-literal spelling.** `1.0` stays `1.0`, `2.50` stays `2.50`, `1e10` prints `1E+10`. A negated literal (`.[-1.5:]`) is *not* an exception — jq's unary minus destroys literal preservation before the slice is ever built, so `-1.5` is already a plain double, exactly as #1088 documented for `.[-1.0]`.

## Scope

`Expr::Slice` has ~94 match sites across `src/jq/parser.rs`, `src/jq/eval.rs`, `src/jq/eval_generic.rs`, and both CLI runners — materially wider than #1088's `Expr::Index`. The representation question is already settled by precedent, though: #1088 added an additive `Expr::IndexNumber { idx, key: NumberKey }` variant, and the same shape applies here (a bound that carries its own `NumberKey` alongside the `i64` every resolution step actually uses). `NumberKey` is already public from `src/jq/expr.rs` and needs no change.

Three rendering sites need the spelling, all of which already share one helper for indices after #1088: `slice::literal_component`, `walk_path`, and `navigation_element`. Everything else is a mechanical `Expr::Slice { .. }` arm extension.

Filed separately rather than folded into #1088 per that issue's own scope, which was the index specifically.

newhoggy:
## Summary

Fixes #1326. `path(.[1.5:3.5])` reported `[{"start":1,"end":4}]` (the resolved bounds)
where jq keeps `[{"start":1.5,"end":3.5}]` (the bounds as written) — the slice-bound
twin of #1088's identical fix for a float array index, found while fixing that issue
and deliberately filed separately (#1088's own scope was the index specifically).

```bash
$ echo '[1,2,3,4,5]' | jq -c 'path(.[1.5:3.5])'
[{"start":1.5,"end":3.5}]
$ echo '[1,2,3,4,5]' | succinctly jq -c 'path(.[1.5:3.5])'   # before this PR
[{"start":1,"end":4}]
```

## Design

Follows #1088's precedent exactly, which the issue's own tier review already settled:
a new additive `Expr::SliceNumber { start, end, start_key, end_key }` variant alongside
`Expr::Slice`, carrying each bound's own `NumberKey` only when it's float-spelled, with
`start`/`end` used identically to the plain variant by every navigation site — a
mechanical `Expr::Slice { .. } | Expr::SliceNumber { .. }` extension to ~25 match sites
across `src/jq/eval.rs` and both CLI runners — and only the three rendering sites
(`slice::literal_component`, `navigation_element`, `walk_path`) actually reading the key.

Two paths reach a static slice bound and **both** needed the fix:

- **The parser's own literal fold** (`fold_slice_bound`, replacing `fold_int_literal`)
  for `.[1.5:3.5]`-shaped literals.
- **The dynamic runtime resolver** (`resolve_slice_bound`/`resolve_slice_expr`) for
  `.[$a:$b]`-shaped computed bounds — a non-whole float literal (`1.5`, unlike `1.0`)
  can't fold at parse time at all (`fold_index_key` only folds whole-valued floats,
  #1061), so **the issue's own primary repro exercises this path exclusively**, not the
  parser fold.

An initial implementation of this fix covered only the parser side and looked complete
(compiled cleanly, matched the whole-valued-float case `.[1.0:3.0]`) while silently
missing the issue's actual repro — caught by re-testing the exact reported case against
the live oracle after implementing, not by trusting a paraphrased read of the code as
already covering it.

A negated bound (`.[-1.5:-3.5]`) keeps its own #1088-established caveat: negation
destroys literal preservation before the slice is built, so a whole-valued negated bound
renders bare (`-1`, not `-1.0`) via jq's own bare-double convention for a computed
value — still `SliceNumber`-carried, just rendering identically to the plain case at
that specific boundary (confirmed live: `path(.[-3.0:-1.0])` is `[{"start":-3,"end":-1}]`
in both jq and this branch, while `path(.[-3.5:-1.5])` keeps the `.5`). Two existing
parser tests asserting the pre-fix "always plain `Expr::Slice`" behavior are updated to
the new, oracle-matching expectation.

## Verification

- Live-verified against real jq 1.7.1 across literal, whole-valued, mixed, open-bound,
  dynamic, and negated shapes (matching every case the issue itself listed)
- `setpath`/`getpath`/`del()` round-tripping and the resolved *value* (not just the path
  component) confirmed unaffected
- yq mode confirmed unaffected (real yq's own CLI arg parser rejects a float slice bound
  outright — a pre-existing, unrelated divergence — so oracle comparison there is
  inconclusive, matching this crate's usual caveat for yq-mode edge cases)
- Full test suite green, 0 regressions
- 100% patch coverage (244/244 new lines), including direct unit tests for the three
  AST-rebuild walkers (`substitute_var`/`expand_func_calls`/`substitute_func_param`)
  and a pre-existing `Expr::IndexNumber` gap in `delete_expr_array_paths`'s error
  path closed as a byproduct of adding the `SliceNumber` arm there consistently

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 2700 lib tests, 0 failures)
- [x] `cargo fmt --check`
- [x] `omni-dev coverage diff` — 100% patch coverage (244/244 new lines)
- [x] Live-verified against real jq 1.7.1 across every shape the issue listed, plus additional boundary cases


newhoggy:
Ran the mandatory 8-angle code review. One genuine bug found and fixed, plus doc/duplication cleanup:

**Real bug** (cross-file tracer angle, live-verified against jq 1.7.1): `resolve_slice_expr`'s two `path()`-error-message sites (the #843 passthrough-target check, and its #843/#986 untrackable-target-branch sibling) still built their `{"start":...,"end":...}` descriptor via the plain, key-blind `slice::literal_component` instead of the key-aware `slice_component_value` every other rendering site already used — so a dynamic float slice bound reaching either error path reported its resolved integer instead of its own spelling. Fixed both call sites; added a regression test that fails against the pre-fix code (confirmed by temporarily reverting the fix and re-running it) and passes now, matching jq 1.7.1 exactly on both repros.

**Duplicated-predicate cleanup** (reuse / CLAUDE.md-conventions / altitude angles independently flagged this): added `Expr::is_slice()` and used it at the ~7 sites that only asked "is this a slice" as a plain boolean, instead of each hand-copying the `Slice | SliceNumber` or-pattern; match arms that also bind `start`/`end` are unaffected (a match pattern can't call a function). Also merged `numeric_path_component`'s and `numeric_slice_bound_key`'s identical float-detection match into one shared `number_key_for`.

**Doc staleness**: fixed three comments still naming `fold_int_literal` (renamed to `fold_slice_bound` by this PR), and corrected `slice_component_value`'s doc comment, which claimed three call sites including `eval_pipe_with_path_context_internal` — that function has no `Expr::Slice` arm at all (a pre-existing gap, not this PR's to fix) and never called it.

**Not fixed here** — filed as #1401 instead: the efficiency and altitude angles measured the `Expr` enum growing 72→96 bytes from `SliceNumber`'s fields, and the `Slice`/`SliceNumber` split (mirroring #1088's `Index`/`IndexNumber`) needing ~25 match-site edits, now a two-for-two precedent. Both are inherent to matching #1088's established representation exactly; #1401 captures the design tradeoff for whenever a third such pair is proposed, rather than fixing it piecemeal here.

Re-verified after rebasing onto `main` (picked up #1397/ADR-0018 and the del()-optional-gate fix, neither touching this code): build clean, 2701/2701 lib tests pass, fmt clean, both clippy legs clean, no_std build clean, 100% patch coverage (268/268 new lines).

PR #1402, opened 2026-08-21
fix(yaml): exempt zero-mantissa float literals from the digit cap

### LINKED ISSUE #1211: yaml: a zero-mantissa literal past 17 fractional zero digits loses literal preservation entirely
**Severity: Low**

## Summary

Found during code review of #1178's PR (#1208). Pre-existing, unrelated to that fix — `src/yaml/scalar.rs` is untouched by it, and reproduces identically before and after.

`is_preservable_float_literal` (`src/yaml/scalar.rs`) counts a zero-mantissa literal's fractional-zero digits toward its `MAX_PRESERVABLE_FLOAT_DIGITS` (17) cap the same way it counts a nonzero mantissa's significant digits. Past that cap, the literal falls back to lossy `as_f64` preservation instead of keeping the source text — losing the literal entirely for a zero-mantissa value, where real yq keeps it verbatim regardless of length.

## Repro (verified against pinned yq v4.53.3)

```
$ printf 'a: 0.00000000000000000000e-400\n' | yq -o=json '.a | tostring'
"0.00000000000000000000e-400"

$ printf 'a: 0.00000000000000000000e-400\n' | succinctly yq -o=json '.a | tostring'
"0"
```

A short zero-mantissa (under the 17-digit cap, e.g. `0.000e-400`) is unaffected — preserved correctly in both.

## Why this is in the same problem domain as #1178 but a separate bug

#1178 fixed the *exponent shift* computation once a zero-mantissa literal's text is already being preserved and formatted (the JSON/jq-mode path, `src/jq/value.rs`). This is a different, earlier decision — *whether to preserve the literal's source text at all* — in the YAML-specific scalar-classification path (`src/yaml/scalar.rs`), which the JSON/jq path doesn't have (JSON number literals aren't gated by a digit-count cap the same way).

## Suggested fix direction

Not investigated. A zero mantissa has no "significant digits" in the sense `MAX_PRESERVABLE_FLOAT_DIGITS` is meant to bound (the cap exists to keep a nonzero mantissa's *rendered* precision reasonable) — the fractional zero-digit count for a zero mantissa is more like #1178's `frac_len` (an exponent-shift input, unbounded by design) than a precision budget. The cap's own zero-mantissa case likely needs its own carve-out, mirroring how `normalize_extreme_literal_mantissa`'s `MAX_RENDERED_MANTISSA_DIGITS` already only bounds what's *rendered*, not what's *recognized as preservable* in the first place.

## Scope note

Filing to make this trackable, not blocking.

newhoggy:
## Summary

- `is_preservable_float_literal` (`src/yaml/scalar.rs`) counted a zero-mantissa literal's fractional zero digits toward `MAX_PRESERVABLE_FLOAT_DIGITS` (17) the same way it counts a nonzero mantissa's significant digits. Past that cap, the literal fell back to lossy `as_f64` reconstruction — losing it entirely — even though every zero-mantissa spelling represents the exact same value (`0`) regardless of length, so there's no precision the cap needs to protect against.
- Added `is_zero_mantissa`, a carve-out checked before the existing digit-count cap. A mantissa with even one nonzero digit still falls through to the unchanged cap check — the separate, pre-existing 17-digit cap for genuinely nonzero mantissas is untouched, since that's a documented, intentional design decision (round-trip-exact `f64` precision), not something this issue asked to change.

## Verification

Live-verified against pinned yq v4.53.3 (per ADR-0018):
- The issue's own repro (`0.00000000000000000000e-400`) now preserves verbatim, matching the oracle.
- A 100-digit zero-mantissa case also preserves verbatim, matching the oracle.
- A short zero-mantissa (already worked before this fix) is unaffected.
- A mantissa with a nonzero digit anywhere still respects the cap (unchanged behavior, confirmed identical to `main`).
- Confirmed two adjacent-looking cases (a long mixed-zero-then-nonzero mantissa, and a >17-digit *nonzero* mantissa) are **pre-existing, separate divergences** unrelated to and unaffected by this fix — identical output on `main` before and after.

Full local verification (build, full test suite, both CI clippy legs, rustdoc under `-D warnings`, no_std, `cargo fmt --check`) all green.

Fixes #1211

### DIFF
--- a/src/jq/value.rs
+++ b/src/jq/value.rs
@@ -1905,6 +1905,13 @@ impl OwnedValue {
     /// it needs the same guard [`to_json`](Self::to_json) does.
     fn to_json_for_reindex_at_depth<S: EvalSemantics>(&self, depth: usize) -> String {
         assert_value_tree_depth(depth);
+        // Shared by both `NumberLiteral` arms below (infinite and finite):
+        // this function's callers (`reduce`/`foreach`/etc.'s per-iteration
+        // reindex bridge) can run it over the same unchanged value
+        // thousands of times, and a document-sourced literal is
+        // otherwise-unbounded text -- see each arm's own comment for why
+        // its particular fallback is safe.
+        const MAX_REUSED_LITERAL_LEN: usize = 256;
         match self {
             Self::Float(f) if f.is_nan() => NAN_SENTINEL.to_string(),
             Self::NumberLiteral(NumberRepr::Float(f), _) if f.is_nan() => NAN_SENTINEL.to_string(),
@@ -1925,17 +1932,13 @@ impl OwnedValue {
             // plain `Float`, handled by the arm above, never this one - so
             // no further shape-checking is needed for correctness.
             //
-            // The length cap *is* needed regardless of shape: the literal
-            // is otherwise-unbounded document text, and this function's
-            // callers (`reduce`/`foreach`/etc.'s per-iteration reindex
-            // bridge) can run it over the same unchanged value thousands of
-            // times - a bound this loose still comfortably covers any
-            // realistic overflow literal (reaching `f64::MAX` needs on the
-            // order of ~300 exponent digits at most) while keeping the
-            // *reused* case itself O(1)-ish rather than O(iterations x
-            // literal length) for a pathological one.
+            // The length cap (`MAX_REUSED_LITERAL_LEN` above) *is* needed
+            // regardless of shape: a bound this loose still comfortably
+            // covers any realistic overflow literal (reaching `f64::MAX`
+            // needs on the order of ~300 exponent digits at most) while
+            // keeping the *reused* case itself O(1)-ish rather than
+            // O(iterations x literal length) for a pathological one.
             Self::NumberLiteral(NumberRepr::Float(f), literal) if f.is_infinite() => {
-                const MAX_REUSED_LITERAL_LEN: usize = 256;
                 if literal.len() <= MAX_REUSED_LITERAL_LEN {
                     literal.to_string()
                 } else {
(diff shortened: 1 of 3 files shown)


newhoggy:
Second review round (7 finder angles) confirmed the core fix's correctness, and found two real, actionable gaps beyond what the issue's own suggested fix asked for — both addressed in a follow-up commit:

**1. The narrow zero-mantissa carve-out left an adjacent, live-verified divergence in place.** A mantissa with a long leading-zero run before a handful of real digits (e.g. 17 significant digits behind 17 leading zeros) was still wrongly rejected by the raw digit-count cap — the same miscount this issue reported, just needing one nonzero digit instead of zero. Generalized \`is_preservable_float_literal\` to count only *significant* digits (from the first nonzero digit onward — the same rule scientific notation itself uses) rather than every raw digit in the mantissa. A zero-valued mantissa has zero significant digits under this rule, so it no longer needs a separate carve-out — \`is_zero_mantissa\` is gone, replaced by \`significant_mantissa_digit_count\`.

**2. A real, measured CPU-time regression this widening introduces.** \`OwnedValue::to_json_for_reindex_at_depth\`'s finite-\`NumberLiteral\` arm has no length cap, unlike its infinite-literal sibling (\`MAX_REUSED_LITERAL_LEN=256\`). The *old* 17-digit cap kept every YAML-sourced finite literal short by accident; widening it (for either the narrow or the generalized fix) lets an arbitrarily long literal reach that uncapped arm. Any \`reduce\`/\`foreach\`/\`while\`/\`until\` loop touching such a value then re-serializes the full literal on every iteration. Confirmed live and measured: wall time scaling linearly with iteration count for a 200,000-digit literal before this fix (0.11s→0.39s→1.49s at N=200/800/3200), flat ~0.01s after. Capped the same way the infinite arm already is — past the length cap, fall back to the parsed value's own bounded formatting instead of echoing the unbounded literal text.

Also addressed from the same round: simplified a hand-rolled loop to the codebase's own established \`.all()\` idiom (now moot — replaced by the generalization above), and updated \`is_preservable_float_literal\`'s own top-level doc comment, which hadn't been kept in sync with the carve-out (matching this file's established convention of documenting every exception inline).

Full local verification (build, full test suite — including the reindex-bridge fix's own new unit tests — both CI clippy legs, rustdoc under \`-D warnings\`, no_std, \`cargo fmt --check\`) green. Rebased cleanly onto \`main\` (which advanced with unrelated commits during review).

Issue #1403, opened 2026-08-21
jq: string interpolation with a multi-valued embedded expression only takes the first value, doesn't fan out


newhoggy:
Found while investigating #1334/#1306's `needs_path_context`/`eval_pipe_with_path_context_internal` fixes — a separate, pre-existing bug, not caused by or related to that work.

## Repro (confirmed live against jq 1.7.1)

```
$ echo null | jq -c '"\(1,2)"'
"1"
"2"
$ echo null | succinctly jq -c '"\(1,2)"'
"1"
```

Real jq fans out: a `\(...)` slot embedding a multi-valued generator produces one string output per value (cartesian product across every slot, for a string with more than one `\(...)`). succinctly's `eval_string_interpolation` (`src/jq/eval.rs`) instead always takes just `vs.first()`/first output of each slot's evaluation, silently discarding the rest:

```rust
QueryResult::Many(vs) => {
    if let Some(v) = vs.first() {
        owned_to_string::<S>(&to_owned(v))
    } else {
        String::new()
    }
}
```
(same shape for `ManyOwned`)

## Scope note

Deliberately not fixed as part of #1334/#1306: those are narrowly scoped to routing (`key`/`parent`/etc. resolving instead of stubbing to `null`/`0` inside a `\(...)` slot), and #1334's own repro is single-valued so doesn't touch this gap. A real fix here needs its own design pass, similar to why #1332 (`Expr::Object`'s multi-entry generator gap) was split out from #1302 rather than folded in: cartesian-producting across every slot when more than one is multi-valued (`"\(1,2)-\(3,4)"` should presumably yield 4 strings, matching real jq) is a different shape of fix than a single recursive call.

## Suggested fix direction

Not investigated beyond confirming the repro. `eval_string_interpolation` would need to become a proper fan-out generator (return `QueryResult::ManyOwned`/similar over the cartesian product of every slot's outputs) instead of building one `String` and returning `QueryResult::Owned`.

newhoggy:
🔧 Claimed by agent `Barbara` — branch `issue-1403-string-interpolation-fanout`

Issue #1306, opened 2026-08-20
jq: key extension loses path context after a comma sibling or inside a def scope


newhoggy:
**Severity: Low**

## Summary

Found while adding regression coverage for #1280. The succinctly-only `key`
path-tracking extension returns `null` instead of the real key in at least
two shapes involving a preceding comma sibling or a `def` scope, even
though `key` alone (no comma, no `def`) works correctly.

## Repro (confirmed live against current main tip)

**Comma sibling before `key`, when the other branch is `?`-guarded:**

```bash
$ echo '{"a":{"b":null}}' | succinctly jq -c '.a.b?, key'
null
null
```

Expected: `null` (from `.a.b?`, genuinely null) then `"b"` (from `key`).
Order doesn't matter -- swapping to `key, .a.b?` also loses `key`'s real
value:

```bash
$ echo '{"a":{"b":null}}' | succinctly jq -c 'key, .a.b?'
null
null
```

**`key` after a `def ...;` scope:**

```bash
$ echo '{"a":1}' | succinctly jq -c '.a | def f: 5; f, key'
5
null
```

Expected: `5` then `"a"`. The `def`-scoped body evaluating a completely
unrelated, non-empty value (`5`, not a swallowed/empty result) still causes
the sibling `key` to lose its path context and print `null`.

## Why this is separate from #1280

#1280 was about `eval_owned_expr_ctrl_full`'s `QueryResult::None -> Ok(Null)`
collapse -- a comma/pipe branch that legitimately produces *nothing*
incorrectly showing `null`. This issue is different: `key` itself is
producing the wrong value (`null` instead of the real key), and reproduces
even when every comma branch produces a real, non-empty value (the `def f:
5` case above) -- ruling out any connection to #1280's None-collapse
mechanism.

## Suggested fix direction

Not investigated. Likely somewhere in how `eval_pipe_with_path_context_internal`
threads `current_path`/`root` across a `Comma` arm's siblings, or how a
`def`-introduced function scope affects whichever state `key`/`Builtin::Key`
reads to answer "what key led to the current value" -- worth checking
whether `Expr::Comma`'s own arm and/or function-call handling reset or lose
that state instead of preserving it across the sibling/scope boundary.

## Related

Found during #1280's own regression-test authoring (PR pending); not a
regression from that fix -- confirmed present on `main` before #1280 too.

newhoggy:
## Tier review (2026-08-21) — **Tier 2**, moving to `Ready`

[...]

PR #1404, opened 2026-08-21
fix(jq): route string interpolation and def scopes through path context

### LINKED ISSUE #1334: jq: needs_path_context has no Expr::StringInterpolation arm, same gap as #1302's Expr::Array
**Severity: Low**

## Summary

Found during #1302/#1333's code review. The identical bug exists for string interpolation: `needs_path_context` has no `Expr::StringInterpolation` arm either, so `key`/`parent`/`file_index` nested inside `"\(...)"` still stub to `null`/`0`.

## Repro (confirmed live)

```
$ echo '{"a":1}' | succinctly jq '.a | "k=\(key)"'
"k=null"
```
(should be `"k=a"`)

Also reproduces under `--eval-all`:
```
$ succinctly yq --eval-all '.[] | "\(key)-\(file_index)"' f1.yaml f2.yaml
"null-0"
"null-0"
```
(should be `"0-0"`/`"1-1"`)

## Why this wasn't folded into #1302/#1333

Same reasoning as #1332 (the `Expr::Object` instance of this same gap): `Expr::StringInterpolation` isn't a single recursive `Box<Expr>` the way `Expr::Array` is — it interpolates multiple embedded expressions into one string, likely with its own multi-output/cartesian-product semantics per embedded `\(...)`. #1333's fix for `Expr::Array` (adding a `needs_path_context` arm plus a dedicated `eval_pipe_with_path_context_internal` construction arm that isolates `optional` and catches atomically) is a template for the mechanism, but this needs its own pass over how multiple interpolated sub-expressions combine.

## Suggested fix direction

Not investigated beyond confirming the repro. Look at how `Expr::StringInterpolation` is evaluated today (likely a dedicated function building the string piece-by-piece), and figure out how to route each embedded expression through `eval_pipe_with_path_context_internal` when it individually needs path context.

Given this is now the third independent instance of the same underlying shape (Array here fixed in #1333, Object tracked in #1332, this one) — worth considering whether a single generalized "route a value-constructing expression's children through path context, atomically catch `?`" mechanism is worth building once all three are understood, rather than a third bespoke ~50-line match arm re-deriving the same Error/Break/Halt/Partial/optional handling a third time (#1333's own review found and fixed a subtle `?`-atomicity bug in its first implementation of this pattern — a shared mechanism would only need to get that right once).

newhoggy:
Fixes #1334. Addresses the "def scope" half of #1306 (kept open — see below).

## #1334: string interpolation

`needs_path_context` had no `Expr::StringInterpolation` arm, so `key`/`parent`/`file_index`
stubbed to `null`/`0` inside a `\(...)` slot:

```
$ echo '{"a":1}' | succinctly jq -c '.a | "k=\(key)"'
"k=null"      # should be "k=a"
```

Same two-part shape #1302 established for `Expr::Array`: `needs_path_context` recurses
into each slot, and a new `eval_pipe_with_path_context_internal` arm routes each slot's
own evaluation through the path-context evaluator during construction, instead of the
generic fallback's plain, context-less `eval_string_interpolation`. `?`-atomicity built in
from the start per #1302's own review lesson (its first cut needed a follow-up, #1333, to
force the inner evaluation's `optional` to `false` so a genuine error surfaces for the
arm's own catch instead of self-swallowing at the leaf) — not left to be independently
rediscovered here.

Found and filed, not fixed here (separate, pre-existing, unrelated gap): real jq fans a
multi-valued `\(...)` slot out into one string per value (`"\(1,2)"` → `"1"`, `"2"`);
succinctly's `eval_string_interpolation` always takes just the first value. My new arm
matches that existing convention exactly rather than fixing it as a byproduct. Filed as
#1403.

## #1306: def scope (partial)

The "def scope" repro is a real, same-family bug:

```
$ echo '{"a":1}' | succinctly jq -c '.a | def f: 5; f, key'
5
null          # should be "a"
```

`eval_func_def` macro-expands `then`'s function calls into `body`, then evaluates the
result via `eval_single` unconditionally — so even once `needs_path_context`'s own new
`FuncDef` arm correctly routes the pipe into path-context evaluation, the generic fallback
still called that same plain-evaluator path underneath, silently dropping context again.
Fixed with a dedicated `FuncDef` arm that performs the identical expansion, then continues
through *this* same path-context evaluator instead.

Incidentally fixes the identical gap for `file_index` (the `--eval-all` extension sharing
this plumbing) inside a user-defined function — two existing tests had pinned the old,
documented "out of scope" behavior as an explicit assertion; updated to the new, correct
per-element resolution rather than just patched to compile.

**Known, documented gap left open**: the `needs_path_context` `FuncDef` arm is a cheap
syntactic `needs_path_context(body) || needs_path_context(then)` check, not a full
expand-then-check — correct for #1306's own repro, but misses a path-context builtin
reaching `body` only through a call argument (`def f(x): x; f(key)`), since the argument
text never appears in `body`/`then` verbatim. Pinned by its own test rather than left
silently unclear.

**The other half of #1306's own repro turned out to have a mistaken premise, not a bug**:

```
$ echo '{"a":{"b":null}}' | succinctly jq -c '.a.b?, key'
null
null
```

The issue expected `null` then `"b"` (i.e. `key`'s branch reflecting `.a.b?`'s own
navigation) — but a comma's branches are independent, each evaluated against the *same*
starting `current_path` the comma itself received, not threaded from one sibling into the
next. Confirmed live against real jq's own `path(.)` as the oracle-checkable analog:

```
$ echo '{"a":{"b":null}}' | jq -c '.a.b?, path(.)'
null
[]            # not ["a","b"] -- path(.) in the second branch is unaffected by the first
```

This also matches succinctly's own prior, deliberately-tested independence
(`test_key_survives_comma_branches`, #715: `.[] | (key, key)` on a 2-element array is
`["0","0","1","1"]`, not any cross-branch-aware shape). `null` is the same, already-correct
answer a bare `key` at the root already gives — not a regression. Left as current behavior,
pinned by its own test. Posting this finding as a comment on #1306 and keeping it open
rather than closing it from this PR, since only one of its two repros is addressed.

## Verification

- Build clean, 2701/2701 lib tests + 54/54 test blocks pass, fmt clean, both clippy legs
  clean, no_std build clean.
- 98.63% patch coverage (72/73 new lines) — the one uncovered line is a provably-unreachable
  defensive `unreachable!()` arm, the exact same single-line gap #1302's own PR (#1333)
  accepted for the structurally identical `Array` arm.
- All `key`/`parent`/`file_index` shapes verified live for internal consistency (both are
  succinctly extensions — no oracle); the comma-independence finding verified live against
  real jq 1.7.1's `path(.)`.

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -540,6 +540,26 @@ pub(crate) fn needs_path_context(expr: &Expr) -> bool {
         // `label $out | file_index` silently dropped path context (#715
         // follow-up).
         Expr::Label { body, .. } => needs_path_context(body),
+        // `"...\(key)..."` (#1334): the same reasoning as `Array` above --
+        // a string interpolation slot is still just "evaluate this
+        // expression", so whatever it needs, the surrounding string needs
+        // too. Literal parts obviously never need it.
+        Expr::StringInterpolation(parts) => parts.iter().any(|part| match part {
+            StringPart::Literal(_) => false,
+            StringPart::Expr(inner) => needs_path_context(inner),
+        }),
+        // `def name(params): body; then` (#1306): the def itself doesn't
+        // need path context, but `then` might reference `key`/`parent`/...
+        // directly (`def f: 5; f, key`), and so might `body` if `then`
+        // ever calls `name` (`def f: key; .a | f`) -- recursing into both
+        // is a deliberately cheap, syntax-only approximation (no macro
+        // expansion here) that misses only one narrower case: a
+        // path-context builtin reaching `body` solely through a *call
+        // argument* (`def f(x): x; .a | f(key)`), which needs the actual
+        // substitution to see. Not covered by #1306's own repros, so left
+        // as a known gap rather than paying for a full
+        // `expand_func_calls` just to answer this routing question.
+        Expr::FuncDef { body, then, .. } => needs_path_context(body) || needs_path_context(then),
         _ => false,
     }
 }
@@ -11166,9 +11186,13 @@ pub fn eval_lenient<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
 /// `.`/`.[]`/`.field` navigation, comparisons, `select(...)`, `map(...)`,
 /// `if`/`then`/`else`, `try`/`catch`, comma, and `label` (the same
 /// capability level `key`/`document_index` already have; an array literal
-/// gained this capability too, #1302 -- object literals, string
-/// interpolation, and user-defined functions remain out of scope, matching
-/// those builtins' own existing limits).
+/// gained this capability too, #1302, and a string interpolation slot and a
+/// `def` scope/function call both gained it in turn, #1334/#1306 -- object
+/// literals remain out of scope (#1332, its own multi-entry generator
+/// design pass), and a path-context builtin reaching a `def` only through a
+/// *call argument* (`def f(x): x; f(file_index)`) is a narrower, documented
+/// gap in `needs_path_context`'s own `FuncDef` arm rather than this
+/// function's limit).
 ///
 /// Routes through the ordinary evaluator, untouched, whenever `expr`
 /// doesn't reference `file_index`/`key`/`parent`/`path` anywhere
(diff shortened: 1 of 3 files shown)


newhoggy:
Ran the mandatory multi-angle code review. Three legitimate findings fixed, one claimed finding checked against the live binary and found incorrect:

**Fixed:**
- **Reuse**: the `StringInterpolation` arm's match guard re-implemented, near verbatim, the closure body `needs_path_context`'s own `StringInterpolation` arm already defines — two independent copies of "does any slot need path context" that could silently drift on a future `StringPart` variant. Factored into one shared `string_interpolation_needs_path_context`, called from both sites.
- **Doc accuracy**: the `FuncDef` arm's own comment claimed the new arm "costs nothing extra" versus the generic fallback it replaces — actually an understatement: the fallback pays a serialize-and-reparse JSON round-trip (`eval_owned_expr_full`, since `FuncDef` isn't in `eval_owned_fast_path`'s short list) that the new arm skips entirely by staying in the `OwnedValue` domain throughout. Corrected to "costs no more, and often less."
- **Conventions**: this file's own established pattern mirrors every `needs_path_context`/`eval_pipe_with_path_context_internal` fix with yq-mode coverage in `yq_cli_tests.rs` (e.g. #554's own test). This PR shipped with jq-only coverage; added the missing yq-mode regression test. No functional gap found — both new arms already worked correctly in yq mode, live-verified during review.

**Checked and discarded**: one finding claimed `def f(x): x; .a | f(key)` "returns `\"a\"` correctly" (contradicting the PR's own documented call-argument gap and its pinned test). Live-tested directly against the built binary — it returns `null`, matching the PR's original claim. The finding's own repro doesn't reproduce what it claims; no change made.

Re-verified after this response: build clean, 2701/2701 lib tests + 55/55 test blocks pass (one earlier run under `cargo llvm-cov` showed 8 spurious failures under this machine's current concurrent-session load — confirmed environmental by an immediate clean standalone re-run, not a real regression), fmt clean, both clippy legs clean, no_std build clean, 98.61% patch coverage (71/72 new lines — the one gap is the same provably-unreachable defensive `unreachable!()` arm #1302's own PR accepted for the structurally identical `Array` arm).

Issue #1313, opened 2026-08-20
jq: eval_rhs_once/resolve_limit's None-collapse bugs, and a 3-arm duplication in path-context evaluation


newhoggy:
`#1280`'s fix (and the multi-angle `/code-review` round that followed it) surfaced several related None-vs-null collapse bugs and a duplication opportunity that are real but out of scope for that PR. Filing them together since they share a root cause: a caller that collapses "produced nothing" and "produced `null`" into the same `Ok(Null)` before a downstream match can tell the difference.

## 1. `resolve_limit` still collapses a genuinely-empty bound to unlimited passthrough

`src/jq/eval.rs`, `resolve_limit` (around line 13331) calls `eval_owned_expr_ctrl` → `eval_owned_expr_ctrl_full`, which by design still collapses `QueryResult::None` to `Ok(Null)` (its own doc comment says so — it's the "callers need exactly one value" contract `#1280` deliberately left alone for this caller). When `limit`'s bound expression produces zero output, the collapsed `Null` falls into `resolve_limit`'s "unlimited passthrough" branch instead of real jq's "the whole call yields nothing."

Live repro against jq 1.7.1:
```
echo '{"a":1,"b":2}' | succinctly jq -c 'path(limit(empty; .a,.b))'
# succinctly: ["a"]  ["b"]
# real jq:    (nothing, exit 0)
```

Fix shape: give `resolve_limit`'s own match an `Ok(None) => return Ok(Vec::new())` arm (matching jq's `n as $n | ...` desugaring), which only requires switching this one call site from `eval_owned_expr_ctrl` to a call that preserves the distinction (e.g. `eval_owned_expr_full` directly, or a thin `Option`-returning variant). `builtin_paths_filter`'s own use of `eval_owned_expr_ctrl` (the only other direct caller) does *not* need the same fix — it only branches on the `Err` side and discards the `Ok` payload entirely, so the collapse is unobservable there. Confirmed via live probe: `succinctly jq -c '[paths(empty)]'` on `{"a":1}` already matches jq (`[]`).

## 2. Compound-assignment (`+=`/`-=`/`*=`/`/=`/`%=`/`//=`) has the identical bug via `eval_rhs_once`

[...]

Issue #1233, opened 2026-08-18
yq: field/index/iterate scalar-target no-op silently discards the RHS/filter, unlike the slice no-op


newhoggy:
**Severity: Low**

## Summary

Found during #1181's implementation. Real yq's field/index/iterate scalar-target no-op doesn't just skip the *write* — it appears to never evaluate the RHS (`=`) or filter (`|=`) at all, silently discarding even a genuinely erroring one. This is a real, verified divergence from the *slice* no-op (#1101), which does still propagate a genuinely erroring RHS.

## Repro (verified against real yq v4.53.3, several shapes)

```
$ echo 5 | yq -o=json '.a = error("boom")'
5                                # no error -- RHS discarded entirely

$ echo 5 | yq -o=json '.[0] = error("boom")'
5

$ echo 5 | yq -o=json '.a |= error("boom")'
5

$ echo 5 | yq -o=json '.a = (1, error("x"), 3)'
5                                # multi-output RHS, mid-stream error also discarded

$ echo 5 | succinctly yq -o=json '.a = error("boom")'
Error: boom                     # wrong: RHS error propagates

# Compare: the slice no-op DOES propagate a genuinely erroring RHS, in both real yq and succinctly:
$ echo 5 | yq -o=json '.[0:1] = error("boom")'
Error: boom
$ echo 5 | succinctly yq -o=json '.[0:1] = error("boom")'
Error: boom                     # correct, matches (test_slice_assign_scalar_noop_still_propagates_rhs_errors_1101)

# Sanity: a real write (null root, autovivifies) still evaluates and propagates RHS errors normally in both:
$ echo null | yq -o=json '.a = error("boom")'
Error: boom
$ echo null | succinctly yq -o=json '.a = error("boom")'
Error: boom                     # correct
```

## Root cause / why #1181 didn't fix this

`eval_assign`/`eval_update` (`src/jq/eval.rs`) evaluate the RHS/filter fully, *before* `resolve_dynamic_indexes` ever resolves the target path — the RHS's own output values are computed first (`eval_assign`'s `rhs_values`), and only then applied per resolved path via `set_path`/`update_path`. Matching real yq's exact behavior here would need the LHS path resolved *first*, so the RHS/filter can be skipped entirely once every resolved path is already known to be this no-op — a real ordering change, not just another no-op check at the write site.

[...]

Issue #983, opened 2026-08-15
jq: limit/nth/skip's negative-count handling diverges from real jq (should be unlimited-passthrough / error, not silent 0-like fallback)


newhoggy:
**Severity: Low**

## Summary

`limit`, `skip`, and `nth`'s actual, parser-reachable implementations (`eval_limit`, and the `Expr::Skip`/`Expr::NthExpr`-backed equivalents in `src/jq/eval.rs`) diverge from real jq's own semantics for a negative or `null` count argument. Found while investigating #981 (which turned out to target unreachable dead code — see that issue's closing comment).

## Root cause / reference semantics

jq's own `src/builtin.jq` defines these with explicit `if/elif/else` branches on the count argument, not a type-and-range check:

```jq
def limit($n; exp):
    if $n > 0 then label $out | foreach exp as $item ($n; .-1; $item, if . <= 0 then break $out else empty end)
    elif $n == 0 then empty
    else exp end;                                    # <- negative (or null): unlimited passthrough

def nth($n; g):
  if $n < 0 then error("nth doesn't support negative indices")   # <- null orders below 0 too
  else label $out | foreach g as $item ($n + 1; . - 1; if . <= 0 then $item, break $out else empty end) end;

# skip/2 postdates the pinned 1.7.1 oracle (added in jq 1.8) but its def is equally explicit:
def skip($n; expr):
  if $n > 0 then foreach expr as $item ($n; . - 1; if . < 0 then $item else empty end)
  elif $n == 0 then expr
  else error("skip doesn't support negative count") end;       # <- negative (or null)
```

Because jq's `<`/`>`/`==` are total orderings where `null` sorts below every number, a `null` count argument falls into the *same* branch a negative number would for each function — `limit`'s `else` is unlimited passthrough, `nth`'s and `skip`'s `else`/`if $n < 0` is an error.

## Repro (confirmed against pinned jq 1.7.1 for limit/nth; jq 1.8.1 source for skip, since skip/2 doesn't exist in 1.7.1)

```
$ echo '[1,2,3]' | jq -c '[limit(-1; .[])]'
[1,2,3]                                    # unlimited passthrough
$ echo '[1,2,3]' | succinctly jq -c 'limit(-1; .[])'
jq: error (at <stdin>:1): limit requires non-negative integer     # DIVERGES: errors instead

$ echo '[1,2,3]' | jq -c 'nth(-1; .[])'
jq: error (at <stdin>:1): nth doesn't support negative indices
[...]

PR #1406, opened 2026-08-21
fix(jq): a genuinely zero-output RHS/bound now produces no output

newhoggy:
Addresses items 1 and 2 of #1313 (items 3 and 4 are explicitly scoped by the issue's own tier review as separate follow-up work — not closing the issue from this PR).

Three call sites collapsed "produced nothing" into a synthesized `Null` before a downstream match could tell it apart from a real `null` value — real jq's `value as $value | ...`/`n as $n | ...` desugaring means the whole enclosing expression produces zero output when the generator producing `$value`/`$n` does, not that it proceeds with a fabricated `null`.

## `eval_rhs_once` (item 2: `+=`, `-=`, `*=`, `/=`, `%=`, `//=`)

```
$ echo '{"a":1}' | succinctly jq '.a += empty'
{"a":1}       # should be nothing
$ echo '{"a":1}' | succinctly jq '.a -= empty'
error: number (1) and null (null) cannot be subtracted
```

`QueryResult::None => Ok(OwnedValue::Null)` spliced `null` into the update filter (`. op null`) instead of short-circuiting. Changed to `Err(QueryResult::None)` — the function's own existing early-return signal, since both callers already have `Err(early_return) => return early_return`. No caller change needed.

## `resolve_limit` (item 1: `limit(n; f)` inside `path(...)`)

```
$ echo '{"a":1,"b":2}' | succinctly jq -c 'path(limit(empty; .a,.b))'
["a"]
["b"]         # should be nothing
```

Used `eval_owned_expr_ctrl`, which collapses by design; a zero-output `n` fell into the `Null`/`Bool` "unlimited passthrough" branch. Switched to `eval_owned_expr_full`, which keeps the distinction, with an `Ok(None) => return Ok(Vec::new())` arm.

## `eval_limit` (plain value mode — found live while verifying the path-mode fix, not itself named in #1313's text but the same bug's value-mode sibling)

```
$ echo '{"a":1,"b":2}' | succinctly jq -c 'limit(empty; .a,.b)'
jq: error (at <stdin>:1): no value      # should be nothing, exit 0
```

Used `result_to_owned`, which collapses a zero-output `n` into `result_to_owned_ctrl`'s own `Err("no value")`. Migrated to `result_to_owned_full` (#1045) — the same helper every other #1045-migrated caller (e.g. `builtin_ltrimstr`) already uses for exactly this distinction.

## Checked, not fixed here

- **#1313's own item 3** (a 3-arm `continue_rest_with_context` duplication in `eval_pipe_with_path_context_internal`): explicitly scoped by the tier review as "best done as its own PR."
- **Item 4** (unverified loose ends): needs its own verification pass first, per the issue.
- **#1233's yq-mode RHS-discarding fix** (also `Ready`, touches the same `eval_assign` family): read both code paths to confirm no interaction — #1233's planned fix gates entirely before reaching `eval_rhs_once`, for its own narrow yq-mode scalar-target-no-op case. Safe for #1233 to land after this and rebase cleanly.

## Verification

Every repro verified against jq 1.7.1: all six compound operators' empty RHS, `//=` on a falsy target, `limit(empty; ...)` in both path and value mode, plus sanity checks that a normal (non-empty) RHS/bound, negative/null/bool "unlimited" `limit` bounds, and #983's/#824's/#391's prior fixes in the same functions are unaffected.

Build clean, 2704/2704 lib tests + 54/54 test blocks pass (one spurious failure under this machine's concurrent-session load, confirmed environmental by an immediate clean standalone re-run), fmt clean, both clippy legs clean, no_std build clean, 90.91% patch coverage (10/11 new lines — the one gap is a benign `cargo llvm-cov` line-attribution artifact on a multi-line match-arm guard clause; the arm's own body line shows a hit count of 34 across the existing test suite, confirming it's genuinely exercised).

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -11250,13 +11250,28 @@ fn eval_rhs_once<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         QueryResult::Owned(v) => Ok(v),
         // #1313: a genuinely zero-output RHS (`.a += empty`) makes the
         // *whole* compound/alternative assignment produce zero output in
-        // real jq (`value as $value | ...` never binds), not a synthesized
-        // `null` spliced into the update filter -- `Err(QueryResult::None)`
-        // is this function's own early-return signal, so both callers'
-        // existing `Err(early_return) => return early_return` arm already
-        // does the right thing here with no change on their end.
+        // real jq (`value as $value | ...` never binds -- verified live
+        // against jq 1.7.1, all six operators plus `//=`), not a
+        // synthesized `null` spliced into the update filter --
+        // `Err(QueryResult::None)` is this function's own early-return
+        // signal, so both callers' existing `Err(early_return) => return
+        // early_return` arm already does the right thing here with no
+        // change on their end.
         QueryResult::None => Err(QueryResult::None),
         QueryResult::Error(e) => Err(QueryResult::Error(e)),
+        // `Many`/`ManyOwned` defaulting an *empty* vec to `Null` (instead of
+        // propagating zero-output the way the bare `None` arm above now
+        // does, code review #1313) relies on this crate's own hardened
+        // convention that a `Vec`-to-`QueryResult` collapse never
+        // constructs an empty `Many`/`ManyOwned` in the first place --
+        // `collapse_vec`/`owned_vec_to_result` convert an empty
+        // accumulator to `QueryResult::None` before either variant is ever
+        // built (the exact invariant #1038/#1043/#1048 hardened after each
+        // independently missing it). Live-probed several zero-output RHS
+        // shapes that could plausibly reach here as a nonempty-looking
+        // stream (`(1,2) | select(false)`, `(empty, empty)`) -- all
+        // correctly arrive as the `None` arm above instead, confirming the
+        // invariant holds for every reachable path today.
         QueryResult::Many(vs) => Ok(vs.first().map_or(OwnedValue::Null, to_owned)),
         QueryResult::ManyOwned(vs) => Ok(vs.into_iter().next().unwrap_or(OwnedValue::Null)),
         QueryResult::OneCursor(_) => unreachable!("materialize_cursor should have converted this"),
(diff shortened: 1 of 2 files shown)


newhoggy:
Ran the mandatory 8-angle code review. One more real bug found and fixed in the functions this PR already touched, plus style/documentation gaps. Three more instances of the same bug class elsewhere, and a deeper trailing-control-propagation gap in \`limit\` itself, filed as #1408 rather than fixed here — each needs bespoke reasoning, not a mechanical migration.

**Fixed:**
- **Real bug**: \`resolve_limit\`'s n-classification was missing the negative-float "unlimited passthrough" arm \`eval_limit\` already had (#983) — \`path(limit(-1.5; ...))\` raised an error while plain-value-mode \`limit(-1.5; ...)\` correctly passed everything through. Pre-existing, unrelated to this PR's own fix, caught while both functions were already being edited. Verified live against jq 1.7.1; both modes now agree.
- **Style**: restructured \`eval_limit\`'s match to unwrap the \`(value, trailing)\` tuple once up front, matching \`resolve_limit\`'s and \`builtin_nth\`'s own established convention, instead of repeating \`Ok(Some((pattern, _)))\` four times.
- **Doc accuracy**: added "verified live against jq 1.7.1" citations to three comments stating jq semantics without one (matching neighboring unmodified comments' own citation style), and updated \`eval_owned_expr_ctrl\`'s doc comment, which still listed \`resolve_limit\` as a safe caller — it was evicted from that list by this PR's own fix precisely because it exhibited the bug.
- **Undocumented invariant**: added a comment to \`eval_rhs_once\`'s \`Many\`/\`ManyOwned\`/\`Partial\` arms documenting the \`collapse_vec\` invariant (#1038/#1043/#1048) their own empty-vec-defaults-to-\`Null\` branches rely on — live-probed several zero-output RHS shapes that could plausibly reach here as a nonempty-looking stream (\`(1,2) | select(false)\`, \`(empty, empty)\`); all correctly arrive as the already-fixed \`None\` arm instead, so no functional gap, just the reliance made explicit.
- **Test coverage**: added regression tests for the negative-float \`limit\` fix and for the \`optional\`-swallowed-RHS-error interaction with compound assignment (\`.a += (1/0)?\` correctly produces zero output).

**Filed as #1408, not fixed here**: three more live-verified instances of the same "zero-output collapses to error/null" bug class in code this PR doesn't touch (\`eval_nth_expr\`'s two-arg form, \`combinations\`, \`halt_error\`), plus the deeper issue that \`resolve_limit\`/\`eval_limit\` still discard \`n\`'s *trailing* control after its first output (e.g. \`limit((1,error("x")); ...)\` should stream then error, but silently drops the error) — real, pre-existing, not introduced here, but a natural continuation of this exact work. #1408 also raised (as a broader observation) that there's no structural deterrent keeping future callers off the lossy helpers.

Re-verified: build clean, 2704/2704 lib tests + 56/56 test blocks pass, fmt clean, both clippy legs clean, no_std build clean, 100% patch coverage (14/14 new lines).

Issue #1405, opened 2026-08-21
jq: needs_path_context doesn't recurse into And/Or, so key/parent/file_index inside one silently stubs


newhoggy:
## Summary

Found while investigating #1335. `needs_path_context` (`src/jq/eval.rs`) has explicit
recursion arms for `Compare`, `Arithmetic`, `Negate`, `Select`, `Map`, `Pipe`, `Paren`,
`Optional`, `Comma`, `Array`, `IndexExpr`, `SliceExpr`, `If`, `Try`, and `Label` — but not
`Expr::And`/`Expr::Or`. Both fall through to the catch-all `_ => false`, so a `key`/
`parent`/`file_index` nested inside an `and`/`or` never routes to
`eval_pipe_with_path_context_internal` at all — the whole expression evaluates through
the plain evaluator instead, where `eval_and`/`eval_or` (`eval_boolean`) call `eval_single`
on both operands, and `key`/`parent`/`file_index` silently fall back to their context-less
stub value instead of erroring or resolving.

## Repro (confirmed live on `main`, no `?` involved)

```
$ echo '{"a":1}' | succinctly jq -c '.a | (key == "a" and true)'
false                 # expected: true -- key inside `and` silently stubs instead of resolving
$ echo '{"a":1}' | succinctly jq -c '.a | (key == "z" or key == "a")'
false                 # expected: true -- same gap, `or`
$ echo '{"a":1}' | succinctly jq -c '.a | key'
"a"                    # key resolves correctly on its own -- confirms this is And/Or-specific
```

`document_index` is unaffected (resolved through a different, path-context-independent
mechanism), so this is specific to the three constructs that genuinely need
`current_path`: `key`, `parent`, `file_index`.

## Why this matters

This is the same #715-follow-up pattern already fixed for `Compare`/`Arithmetic`/`Select`/
`Map`/`Array`/`Negate` (headline `--eval-all` idioms like `select(file_index == 0)`) — `and`/
`or` were simply missed from that sweep. It also turned out to be the actual root cause
behind #1335's own posted repro (`(key == "a" and error("boom"))?` → `false`): that
expression never reaches `eval_pipe_with_path_context_internal`'s `Expr::Optional` arm at
all, because `needs_path_context` judges the whole thing `false` before ever getting to the
`?`. The wrong `false` there comes entirely from `key` stubbing inside `and`
short-circuiting to a false comparison, not from any optional-broadcast issue. #1335 itself
[...]

PR #1407, opened 2026-08-21
fix(jq): eval_pipe_with_path_context_internal catches ? atomically, not broadcast

### LINKED ISSUE #1335: jq: Expr::Optional's dispatch in eval_pipe_with_path_context_internal broadcasts optional=true instead of catching atomically
**Severity: Medium**

## Summary

Found (and locally worked around) during #1302/#1333's code review. `eval_pipe_with_path_context_internal`'s `Expr::Optional` dispatch just sets `optional=true` and recurses into whatever it wraps, with no catch of its own:

```rust
Expr::Optional(inner) => {
    // Optional - evaluate with optional=true
    ...
    eval_pipe_with_path_context_internal::<W, S>(&[inner], value, root, file_origin, current_path, true)
    ...
}
```

This is architecturally different from the *plain* evaluator's `Expr::Optional` dispatch (`eval_single`, line ~688), which delegates to `eval_try`: evaluate the body, then explicitly convert an escaping `Error`/`Partial(_, Error)` to `None` at the top. The path-context evaluator instead relies entirely on leaf-level `if optional {None} else {Error}` self-checks (e.g. `eval_error`) to implement the swallow — which only works correctly if every construct in between the `?` and the leaf error faithfully re-derives atomicity when it needs to (array/object construction, `first`/`limit`-style short-circuiting, etc.). #1333 found and fixed one concrete instance of this going wrong (`Expr::Array`'s new arm swallowing errors per-branch instead of atomically), but the fix was scoped locally to that one arm, not to the general `Expr::Optional` dispatch mechanism.

## Repro (confirmed live, pre-existing on `main`, unrelated to arrays)

```
$ echo '{"a":1}' | succinctly jq '.a | (key == "a" and error("boom"))?'
false
```
Real jq's structurally equivalent (non-`key`) case:
```
$ echo '{"a":1}' | jq '.a | (1 == 1 and error("boom"))?'
(no output -- the whole `?` swallows the error, doesn't produce a plausible-but-wrong `false`)
```

## Why this wasn't fixed generally in #1333

#1333's own fix (isolating `optional` for the array's inner evaluation, then catching `Error`/`Partial(_, Error)` locally when the array's own `optional` is true) is a per-arm workaround for the `Array` case specifically. Every other construct in `eval_pipe_with_path_context_internal` that recurses (`Comma`, `Arithmetic`, `Compare`, `If`, `Try`, `Select`, `Map`, `Iterate`, ...) still threads `optional` the same broadcast way, and each could independently produce a wrong-but-plausible result (like the `false` above) instead of correctly swallowing to empty, wherever a leaf error sits behind an intermediate combinator with its own semantics.

## Suggested fix direction

Not investigated in depth. The `Expr::Optional` dispatch itself is the natural fix point: instead of broadcasting `optional=true` into `inner`, evaluate `inner` with the *ambient* (pre-`?`) `optional` value (i.e. isolate it, mirroring `eval_try`'s shape), then explicitly catch an escaping `Error`/`Partial(_, Error)` at that single dispatch point and convert to `None` — removing the need for every other arm to re-derive its own local catch. This is a larger, more invasive change than #1333's local workaround (touches the shared dispatch point every other arm relies on), so it needs its own careful review pass across all the arms currently depending on the broadcast behavior, rather than being folded into #1333.

newhoggy:
## Summary

`Expr::Optional`'s dispatch in `eval_pipe_with_path_context_internal` broadcast
`optional=true` into whatever it wrapped, with no catch of its own — relying entirely
on leaf-level `if optional {None} else {Error}` self-checks to implement `?`'s
suppression. Two live symptoms from the same code:

1. When `?` had nothing after it in the pipe, a leaf deep inside the wrapped expression
   could self-swallow instead of this node genuinely catching an escaping error
   atomically (the general concern #1335 raises, mirroring #1302's identical fix for
   the `Array` arm).
2. When `?` had a `rest` of the pipe after it, the old code combined `[inner, ...rest]`
   into *one* list evaluated under that same forced-`true` value — so an error *after*
   the `?`, which the `?` has nothing to do with, was also silently swallowed:
   `.a | (key)? | error("boom")` produced no output and no error at all, instead of
   real jq's error (confirmed against real jq's structurally identical
   `.a | (1)? | error("boom")`).

Now `inner` is evaluated in isolation with `optional` forced to `false`, so a genuine
error/break surfaces instead of self-swallowing at some leaf; caught here
unconditionally (mirrors `eval_try`'s `Expr::Optional` dispatch and the `Array` arm's
`#1302` fix); `rest` continues under the *ambient* `optional` via
`continue_rest_with_context`, the same helper `Comma`/`Try`/`If` already use.

## A note on #1335's own posted repro

`(key == "a" and error("boom")))?` turns out **not** to exercise this arm at all —
`needs_path_context` has no recursion arm for `Expr::And`/`Expr::Or`, so the whole
expression is judged not needing path context and routes through the plain evaluator
instead, where `key` silently stubs (reproduces even with no `?` at all:
`.a | (key == "a" and true)` → `false`, should be `true`). That's a separate,
more fundamental routing gap — filed as #1405 — and is the actual root cause of the
literal repro in #1335's issue text. This PR fixes the real, independently-verified
`Expr::Optional` broadcast bug (via the rest-leak repro above, which *does* route
correctly today) rather than that repro specifically.

## Test plan

- New `test_optional_dispatch_catches_atomically_not_broadcast_1335` in
  `tests/jq_cli_tests.rs`: rest-leak fix, positive rest-continuation, comma-prefix
  preservation, no-`?` control (still hard-errors), `break` swallowed by `?`
  (matches real jq), doubly-nested `?`.
- Sibling `#1302`/`#1333` array-arm tests re-run clean (no regression).
- `cargo clippy --all-targets --all-features -- -D warnings` and the CI's second,
  non-`broadword-yaml` clippy leg: clean.
- `cargo build --no-default-features`: clean (no new warnings).
- Patch coverage: 100% (18/18 new lines).

Fixes #1335

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -18770,26 +18770,59 @@ fn eval_pipe_with_path_context_internal<'a, W: Clone + AsRef<[u64]>, S: EvalSema
             }
         }
         Expr::Optional(inner) => {
-            // Optional - evaluate with optional=true
+            // This node *is* the `?` -- mirrors eval_try's Expr::Optional
+            // dispatch (`eval_try(inner, None, ...)`, eval.rs ~line 692),
+            // which evaluates the body and unconditionally converts an
+            // escaping Error or Break to nothing, keeping any prefix
+            // already produced before it (real jq's `(1, 2, error("x"))?`
+            // outputs `1`, `2`) -- Halt always escapes regardless (#791).
+            //
+            // `inner` is evaluated with `optional` forced to `false`, not
+            // broadcast from the ambient value, so a genuine error/break
+            // deep inside doesn't self-swallow at some leaf's own local
+            // `if optional` check before ever reaching this catch --
+            // mirrors the Array arm's identical fix (#1302) for the same
+            // reason. This catch point is the one place `?`'s suppression
+            // happens; it isn't gated on the ambient `optional` (unlike
+            // the Array arm's `if optional`) because that arm fires even
+            // with no enclosing `?` at all, whereas this arm only exists
+            // because the source literally has a `?` right here.
+            //
+            // Previously this arm broadcast `optional=true` into `inner`,
+            // and -- when `rest` was non-empty -- combined `[inner,
+            // ...rest]` into a single list evaluated under that same
+            // forced-true value, so `rest` (everything piped *after* the
+            // `?`) inherited the suppression too: `.a | (key)? |
+            // error("boom")` produced no output and no error at all,
+            // instead of real jq's error (#1335). `continue_rest_with_context`
+            // below instead threads the *ambient* `optional` into `rest`,
+            // matching every other arm here that continues into `rest`
+            // after producing an intermediate value.
+            let inner_result = eval_pipe_with_path_context_internal::<W, S>(
+                core::slice::from_ref(inner),
+                value,
+                root,
+                file_origin,
+                current_path,
+                false,
+            );
+            let caught = match inner_result {
+                QueryResult::Error(_) | QueryResult::Break(_) => QueryResult::None,
+                QueryResult::Partial(prefix, Control::Error(_) | Control::Break(_)) => {
+                    owned_vec_to_result(prefix)
+                }
+                other => other,
+            };
             if rest.is_empty() {
-                eval_pipe_with_path_context_internal::<W, S>(
-                    &[(**inner).clone()],
-                    value,
-                    root,
-                    file_origin,
-                    current_path,
-                    true,
-                )
+                caught
             } else {
-                let mut combined = vec![(**inner).clone()];
-                combined.extend(rest.iter().cloned());
-                eval_pipe_with_path_context_internal::<W, S>(
-                    &combined,
-                    value,
+                continue_rest_with_context::<W, S>(
+                    caught,
+                    rest,
                     root,
                     file_origin,
                     current_path,
-                    true,
+                    optional,
                 )
             }
         }
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -18769,35 +18769,48 @@ fn eval_pipe_with_path_context_internal<'a, W: Clone + AsRef<[u64]>, S: EvalSema
                 )
             }
         }
-        Expr::Optional(inner) => {
-            // This node *is* the `?` -- mirrors eval_try's Expr::Optional
-            // dispatch (`eval_try(inner, None, ...)`, eval.rs ~line 692),
-            // which evaluates the body and unconditionally converts an
-            // escaping Error or Break to nothing, keeping any prefix
-            // already produced before it (real jq's `(1, 2, error("x"))?`
-            // outputs `1`, `2`) -- Halt always escapes regardless (#791).
-            //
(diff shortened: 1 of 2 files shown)


newhoggy:
## Code review round (9 finders)

All 9 findings addressed or triaged:

**Fixed in this branch:**
- **Real regression, caught before merge**: isolating `inner`'s evaluation with an empty
  `rest` (needed to scope `?`'s catch to `inner` alone) discarded per-output
  `current_path` for navigational `inner` (`.[]`/`.foo`) — `.a | (.[])? | key` silently
  went from correct `0,1,2` to wrong `"a","a","a"` for every element. Fixed by guarding
  the isolated fast path on whether `rest` actually consults path context
  (`needs_path_context`); when it does (a rarer combination), falls back to this arm's
  pre-#1335 combined-list evaluation, which threads paths correctly at the cost of
  reintroducing the narrower "rest inherits suppression" gap for just that
  combination — no worse than `main`'s current behavior there. New regression test:
  `test_optional_wrapped_navigation_still_threads_path_into_rest_1335`.
- Comment overclaimed the new arm "mirrors" the Array arm's (#1302) catch and
  eval_try's dispatch identically — tightened to state the actual differences (forced
  `false` vs eval_try's ambient-forwarding; unconditional vs gated catch; Error+Break vs
  Error-only).
- Added a missing "confirmed against jq 1.7.1" citation on the break-swallowing test,
  matching this file's established convention.
- Trimmed a comment paragraph that re-derived the Array arm's own already-documented
  isolate-then-catch rationale at similar length, to a cross-reference instead.
- Documented the deliberate Halt-passthrough divergence from the Array arm (verified
  live: `(1, 2, halt)?` prints `1`, `2` before exiting, matching real jq and
  `yq_runner.rs`'s top-level convention — the Array arm discards its prefix instead,
  since a halted array never finished constructing).

**Filed as follow-ups (real findings, out of scope for this issue):**
- #1409 — the same isolate-then-`continue_rest_with_context` shape also loses path
  threading for `Comma`/`Try`/`Label`, unguarded, confirmed pre-existing (not introduced
  here). Needs a real redesign (per-output `(value, path)` pairs), not a quick patch.
- #1410 — the "isolate with `optional=false`, catch atomically" pattern has now been
  independently hand-implemented 4 times (`eval_try`, the Array arm, this Optional arm,
  and `eval_generic.rs`'s own Array arm) with no shared helper; also surfaced a dormant
  analog in the `Map` arm and a pre-existing IndexExpr/SliceExpr key-error carve-out gap
  relative to the plain evaluator's narrower special case.

**Confirmed clean (no action needed):**
- Line-by-line diff scan: no findings.
- Removed-behavior audit: traced the full call graph, live-tested against the pinned
  jq oracle across multiple scenarios, ran both full CLI test suites — no lost
  invariant.
- The `Paren` arm's own pre-existing clone-heavy pattern (unrelated efficiency
  observation, not touched by this diff — out of scope).

Full suite (`cargo test --features cli,simd,regex,serde`, single-threaded), both
clippy legs, `cargo build --no-default-features`, and `cargo fmt --check` all clean
after this round. Patch coverage: 100% (28/28 new lines).

Issue #1412, opened 2026-08-21
yq: assignment RHS-discard no-op (#1233) has a comma-LHS gap, plus a real yq-vs-jq path/RHS error-ordering divergence


newhoggy:
Follow-up from #1233's implementation (PR pending) — two related gaps found and deliberately left out of that fix's scope, to keep its own blast radius to exactly the static-path no-op case it targets.

## 1. Comma-LHS scalar-target no-op isn't covered

Real yq no-ops a comma-grouped LHS assignment's RHS the same way it no-ops a single path, when *every* branch is itself a scalar-target no-op:

```
$ echo 5 | yq -o=json '(.a, .b) = error("boom")'
5
```

#1233's fix (`yq_assign_skip_rhs`/`yq_assign_is_total_noop`, `src/jq/eval.rs`) deliberately gates on `!needs_path_prepass(path_expr)` — and `needs_path_prepass` classifies *any* `Expr::Comma`-containing path as needing dynamic resolution (`_ => true` in its own match), regardless of whether the comma's individual branches are themselves static. This was a deliberate safety choice, not an oversight: resolving a genuinely dynamic path *early* (before the RHS, which is what the fix needs to do to skip RHS evaluation) risks reordering two independently-observable things — see item 2 below, which is exactly that risk materializing for a different path shape.

`5 | (.a, .b) = error("boom")` therefore still incorrectly errors in succinctly today, where real yq no-ops.

## 2. A genuine, pre-existing yq-vs-jq path/RHS error-ordering divergence

Found while verifying #1233's fix wouldn't introduce a *new* ordering risk — turned out succinctly already has one, unrelated to that fix:

```
$ echo '{"a":1}' | jq -c '.[error("path-err")] = error("rhs-err")'
jq: error (at <stdin>:1): rhs-err        # jq: RHS evaluated first

$ printf 'a: 1\n' | yq -o=json '.[error("path-err")] = error("rhs-err")'
Error: path-err                          # yq: path evaluated first (!)

$ printf 'a: 1\n' | succinctly yq -o=json '.[error("path-err")] = error("rhs-err")'
Error: rhs-err                           # succinctly yq mode currently matches jq's ordering, not real yq's
```

[...]

Issue #1411, opened 2026-08-21
yq: del() on a scalar root/target doesn't no-op, unlike = and |=


newhoggy:
Found while scoping #1233's implementation plan — filed separately per that plan's own note, not folded into #1233.

## Summary

Real yq's field/index/iterate scalar-target no-op (#1181, and #1233's own RHS-discarding extension) applies to `=`/`|=`/compound-assignment. `del()` on the same scalar-target shapes has never been fixed — it uses a different walker (`delete_at_path`, not `set_path`/`update_path`) and still hard-errors where real yq silently no-ops.

## Repro (verified against real yq v4.53.3)

```
$ echo 5 | yq -o=json 'del(.a)'
5
$ echo 5 | succinctly yq -o=json 'del(.a)'
Error: Cannot index number with string "a"

$ echo 5 | yq -o=json 'del(.[0])'
5
$ echo 5 | succinctly yq -o=json 'del(.[0])'
Error: Cannot index number with number

$ echo 5 | yq -o=json 'del(.[])'
5
$ echo 5 | succinctly yq -o=json 'del(.[])'
Error: Cannot iterate over number (5)

$ echo 5 | yq -o=json 'del(.a.b)'
5
$ echo 5 | succinctly yq -o=json 'del(.a.b)'
Error: Cannot index number with string "a"
```

## Root cause

Same family as #1181/#1233 (real yq's scalar-target no-op), different walker: `del()` goes through `delete_at_path` (`src/jq/eval.rs`), which has no yq-mode scalar-no-op awareness at all — unlike `set_path`/`update_path`, which #1181/#1233 are teaching this to.

## Scope note

Filing to make this trackable, not blocking. Whoever picks this up should check whether `delete_at_path` can reuse the same `is_yq_field_index_noop_scalar`/`navigate_read_only` machinery #1233 introduces for the assignment side, or needs its own analogous check given `del()`'s different semantics (removing a key vs writing a value).

newhoggy:
## Triage (2026-09-04) — re-scoped to `Iterate`-on-scalar and inert-after-null; Field/Index half landed in #2106

Re-verified on main `147fc1638` vs yq v4.53.3. Fixed by #2106: `del(.a)`, `del(.[0])`, `del(.a.b)` on `5` → `5`. Still live (yq left column, succinctly right):

```
echo 5 | yq -o=json 'del(.[])'          5      | Error: Cannot iterate over number (5)
[...]

Issue #1232, opened 2026-08-18
yq: field/index/iterate scalar-target no-op doesn't reach a scalar hit before the last path component


newhoggy:
**Severity: Low**

## Summary

#1181 fixed the yq scalar-target assignment no-op (`.a = v`, `.[0] = v`, `.[] = v`, and a computed key) for the case where the scalar is reached at the *terminal* step of a resolved path — including through a preceding real container (`{"a":5} | .a.b = 99` already correctly no-ops). What #1181 did not fix: a scalar hit *before* the last component, where the rest of the path is still left to resolve once the scalar is reached.

## Repro (verified against real yq v4.53.3)

```
$ echo 5 | yq -o=json '.a.b = 99'
5                                          # no-op

$ echo 5 | succinctly yq -o=json '.a.b = 99'
Error: Cannot index number with string "a"   # wrong: errors instead

$ echo 5 | yq -o=json '.a.b |= 99'
5                                          # no-op

$ echo 5 | succinctly yq -o=json '.a.b |= 99'
Error: Cannot index number with string "a"   # wrong: errors instead

$ echo 5 | yq -o=json '0 as $k1 | 1 as $k2 | .[$k1][$k2] = 99'
5                                          # no-op (falsifies "no-op applies once any component resolves against a scalar" -- this never touches a literal Field/Index at all)

$ echo 5 | succinctly yq -o=json '0 as $k1 | 1 as $k2 | .[$k1][$k2] = 99'
Error: Cannot index number with number       # wrong: errors instead
```

## Root cause

Two separate walkers, neither generic over `EvalSemantics`:

- `get_path_mut` (`src/jq/eval.rs`) — the parent-navigation walker `set_path`'s `Pipe` arm delegates every *non-terminal* path component to, for `=`. Its `Field`/`Index` arms have no `scalar_noop`/yq-mode parameter at all.
- `update_path`'s own `Pipe`-arm `Field`/`Index` sub-arms (a structurally-parallel, separately-maintained copy of the same navigate-then-recurse logic, used for `|=`/`+=`/`-=` chains) — has `scalar_noop` in scope from the enclosing function, but the #1181 fix was only applied to the *terminal* `Field`/`Index` arms a few lines above, not this duplicate.

This is the same class of gap #1101's slice no-op had before #1116 generalized it to any chain depth — mirroring that precedent, #1181 scoped its own fix to the terminal-component case only and left this for its own follow-up.

## Scope note

[...]

PR #1413, opened 2026-08-21
fix(yq): scalar-target assignment no-op discards the RHS, not just the write

### LINKED ISSUE #1233: yq: field/index/iterate scalar-target no-op silently discards the RHS/filter, unlike the slice no-op

newhoggy:
Fixes #1233.

Real yq's field/index/iterate scalar-target no-op (#1181) doesn't just skip the write — it discards the RHS/filter entirely, never evaluating it at all:

```
$ echo 5 | yq -o=json '.a = error("boom")'
5                                # RHS never runs
$ echo 5 | succinctly yq -o=json '.a = error("boom")'
Error: boom                     # wrong: RHS ran and errored
```

`|=` was already correct (`update_path`'s filter runs lazily per resolved path, so its own no-op already short-circuits it) — only the value-preevaluating operators (`=`, and every compound operator via `eval_rhs_once`) diverge.

## The fix

New shared `yq_assign_skip_rhs`/`yq_assign_is_total_noop` (`src/jq/eval.rs`), called by `eval_assign`/`eval_compound_assign`/`eval_alternative_assign` before each evaluates its own RHS: resolves `path_expr`, and if every resolved path is a scalar-target no-op, returns the input unchanged without ever touching the RHS. Mirrors `set_path`'s own existing `scalar_noop` write-time check (same `is_yq_field_index_noop_scalar` predicate) for the shapes it already covers — root (`5 | .a = v`), nested (`{"a":5} | .a.b = v`), and Iterate (`5 | .[] = v`) — plus `?`-transparency (`.a?`/`.a.b?`), one prefix walk instead of several special cases.

Applies uniformly across every value-preevaluating operator, not just `=`/`+=` — live-verified against yq v4.53.3 for `+=`/`-=`/`*=`; `/=`/`%=`/`//=` aren't real yq syntax at all, so their yq-mode answer is judged by internal consistency with the already-correct, already-lazy `|=` they desugar to, not an external-compat claim.

## Deliberately scoped to a *static* `path_expr`

`!needs_path_prepass(path_expr)` — the same guard `resolve_dynamic_indexes`'s own fast path already uses. Resolving a static path is a pure, non-evaluating AST clone, so checking it before the RHS changes nothing observable. A genuinely dynamic path (a computed key, or a `Comma` LHS) is left on the unchanged, existing RHS-first flow — reordering *that* risks changing which of two independently-observable things (an error, a side effect) surfaces first.

This isn't hypothetical: live-verified real jq's own `.[error(P)] = error(R)` reports R (RHS first) while real yq's identical query reports P (path first) — a genuine, pre-existing yq-vs-jq ordering divergence succinctly currently doesn't replicate in either direction. Filed as #1412 rather than touched here, along with the one shape this leaves uncovered (comma-LHS-all-noop, e.g. `5 | (.a, .b) = error("boom")`, which real yq also no-ops).

## Related, deliberately out of scope

- `.[0:1] = error("boom")`'s own, separate no-op (#1101/#1116) is excluded via the same `split_at_slice` check `set_path` uses — that mechanism only ever skips the *write*, never the RHS, and must keep genuinely propagating an erroring RHS. Live-verified unaffected.
- A scalar hit *before* the last path component (`.a.b.c` where `.a` is already scalar) is #1232's own territory — this fix's prefix walk against the *original* root doesn't widen that depth, matching the issue's own plan.
- `del()` has the identical scalar-root gap through its own separate walker (`delete_at_path`) — filed separately as #1411 per the issue's own implementation plan, not folded in here.

## Verification

Every shape verified against yq v4.53.3: every operator, multi-output RHS, nested/Iterate/`?`-wrapped targets, the real-write (null root) contrast, the slice-no-op contrast, the comma-LHS-mixed contrast, and the trailing-bare-`.` edge case; jq mode confirmed unaffected.

Build clean, 2704/2704 lib tests + 58/58 test blocks pass, fmt clean, both clippy legs clean, no_std build clean, 97.73% patch coverage (43/44 new lines — the one gap is a provably-unreachable defensive recursion arm the parser/resolver never actually produce a shape for).

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -230,6 +230,33 @@ before one can be captured
 ([#1361](https://github.com/rust-works/succinctly/issues/1361)). Both predate
 [ADR-0017](../../adrs/adr-0017.md)'s mechanism and neither is anchor-specific.
 
+### Comma-grouped scalar-target assignment no-op
+
+[#1233](https://github.com/rust-works/succinctly/issues/1233) taught `=`/`+=`/`-=`/`*=`/`//=`
+that real yq's field/index/iterate scalar-target no-op ([#1181](https://github.com/rust-works/succinctly/issues/1181))
+discards the RHS entirely rather than merely skipping the write — but only for a *static*
+`path_expr` (no computed key, and no `Comma`). A comma-grouped LHS where every branch is
+itself a scalar-target no-op is real yq's identical behaviour, live-verified, that
+succinctly does not yet replicate:
+
+```bash
+$ echo 5 | yq            -o=json '(.a, .b) = error("boom")'   # 5, RHS never runs
+$ echo 5 | succinctly yq -o=json '(.a, .b) = error("boom")'   # Error: boom
+```
+
+Deliberately not covered by #1233's own fix: resolving a `Comma`-containing path *before*
+the RHS (which the fix needs to do, to decide whether to skip it) is real evaluation for
+that shape, unlike a bare static path (a pure, non-evaluating AST clone) — and moving real
+evaluation ahead of the RHS risks reordering two independently-observable things. That risk
+is not hypothetical: real jq's own `.[error(P)] = error(R)` reports `R` (RHS evaluated
+first) while real yq's identical query reports `P` (path first) — succinctly currently
+matches jq's ordering in both modes, a second, related divergence from real yq neither
+tracked before now. Both filed together as
+[#1412](https://github.com/rust-works/succinctly/issues/1412), which also notes a real fix
+for the ordering divergence would likely resolve the comma-LHS gap as a side effect (a full
+path-before-RHS reorder no longer needs the static-only safety gate #1233's own narrower
+fix relies on).
+
 ### Other categories
 
 Float and number formatting ([#1071](https://github.com/rust-works/succinctly/issues/1071),
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -10770,6 +10770,91 @@ fn is_yq_field_index_noop_scalar(target: &OwnedValue) -> bool {
     !is_owned_container(target) && !matches!(target, OwnedValue::Null)
 }
 
+/// Whether writing `path` (a single *resolved* assignment-target path, one
+/// `resolve_dynamic_indexes` entry) into `root` would be real yq's
+/// field/index/iterate scalar-target no-op (#1181): navigating every
+/// component *before* the last reaches a genuine scalar. When true for
+/// every resolved path, real yq discards the RHS/filter entirely rather
+/// than merely skipping the write (#1233, live-verified v4.53.3) --
+/// `eval_assign`'s own caller is what actually skips RHS evaluation; this
+/// only answers the per-path question.
+///
+/// `path` is flattened via [`push_path_components`] first -- the same
+/// flattener `split_at_slice` already uses internally -- rather than
+/// matched directly (bare component vs. `Expr::Pipe`), so a nested
+/// `Pipe`/`Paren`/`Optional` anywhere in the chain (`(.a|.b)[0] = v`,
+/// structurally the same target as `.a.b[0] = v`) is seen through
+/// uniformly instead of only when it already happens to be a flat `Pipe`
+/// list. Code review (#1233) found this the hard way: an earlier version
+/// handed the *raw*, unflattened components to `navigate_read_only` for
+/// the prefix walk, silently disagreeing with itself on two spellings of
+/// the identical target (`.a.b[0] = error(...)` correctly no-op'd while
+/// `(.a|.b)[0] = error(...)` didn't) -- the exact "mirroring a precedent
+/// without its fix" pattern `get_path_mut`'s own #1287/#1294 nested-`Pipe`
+/// splice already had to solve once for the write side. `?` anywhere in
+/// the chain (whole-path, terminal, or a middle component) is transparent
+/// through the same flattening + [`unwrap_path_component`] the checks
+/// below use -- `?` introduces no evaluation of its own to worry about
+/// reordering, and live-verified real yq no-ops each exactly the same as
+/// its unwrapped equivalent (`.a.b? = error("boom")` on `{"a":5}` is
+/// `{"a":5}`, matching plain `.a.b = error("boom")`).
+///
+/// The flattened list's *last* element still carries a bare, unresolved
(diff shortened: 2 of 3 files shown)


newhoggy:
Ran the mandatory 8-angle code review. Two real bugs found and fixed, one real efficiency regression fixed for the main call site (with the remaining two call sites' cost tracked as a follow-up), plus a documentation gap closed.

**Real bugs fixed:**
- **Nested-`Pipe` disagreement**: `navigate_read_only` (introduced by this PR) never got the #1287/#1294 nested-`Pipe` splice fix `get_path_mut` already has — live-verified `.a.b[0] = error("boom")` correctly no-op'd while the semantically identical `(.a|.b)[0] = error("boom")` didn't, even though both spellings write to the exact same target. Fixed by flattening the whole resolved path via `push_path_components` (the same flattener `split_at_slice` already uses) before any check runs, instead of matching `path` directly and handing the raw, unflattened components to `navigate_read_only`. This also eliminated a duplicated terminal-shape match (was written twice) and a coverage gap (the `exprs.len() == 1` recursive arm), and turned the "last isn't a valid terminal" check into provably-dead defensive code (confirmed via patch-coverage hit counts, not just a hunch) — documented as such rather than removed, since it's the one thing standing between a future new static-classified `Expr` variant and a silent miss.
- **`?` transparency**: `.a?`/`.a.b?` weren't recognized as no-ops at all (`yq_assign_is_total_noop` had no `Expr::Optional` handling) — found live while closing this fix's own coverage gaps, not from the review round. Fixed via the same `unwrap_path_component` the flattening rewrite already needed.

**Efficiency regression fixed** (`eval_assign` only): `yq_assign_skip_rhs` materialized the whole document via `to_owned` to check the no-op condition, and when it wasn't a no-op (the common case for any real write), `eval_assign` immediately redid an identical `to_owned` + `resolve_dynamic_indexes` pass. Confirmed by essentially every review angle as a real doubling of a full O(document-size) DOM conversion on the write hot path. Fixed by having the check return a `YqAssignNoopCheck::Continue { pristine, paths }` variant `eval_assign` now reuses instead of re-deriving. The same fix for `eval_compound_assign`/`eval_alternative_assign` (routed through the shared `eval_update`, also used by plain `|=` with no pre-check at all) needs `eval_update`'s own signature widened — real design work, not a mechanical follow-up — filed as #1414 rather than attempted here.

**Documentation gap closed**: the comma-LHS no-op divergence (already noted in code as deferred to #1412) is now also recorded in `docs/compliance/yq/limitations.md`'s "Open divergences" section, per this repo's own CLAUDE.md rule that a divergence not recorded there "is not a decision — it is a bug nobody has found yet."

**Checked and not changed** (addressed in a doc comment, not code): several review passes flagged that `halt`/`break`/`stderr`/`debug` inside a no-op'd RHS never fire now, framing this as a behavior regression from `main`. `main`'s behavior there is exactly the bug #1233 exists to fix (unconditionally evaluating an RHS real yq never touches) — not a baseline to preserve. The multi-output case already live-verified before this review (`.a = (1, error("x"), 3)` no-ops without even computing the leading `1`) proves evaluation genuinely never starts, not "starts and its result is discarded" — treating `halt`/`break`/side-effects uniformly with that is the internally consistent reading, not a special case needing a fix. Documented explicitly on `yq_assign_skip_rhs` since it's come up repeatedly.

Re-verified after this response and a rebase onto `main` (picked up two unrelated jq-mode path-context fixes): build clean, 2704/2704 lib tests + 61/61 test blocks pass, fmt clean, both clippy legs clean, no_std build clean, 93.1% patch coverage (54/58 new lines — all 4 gaps confirmed genuinely-unreachable defensive code via patch-coverage hit-count inspection, not assumed).

Issue #1325, opened 2026-08-20
yq: del()'s comma-grouped Object-slice rewrite still has 3 narrower gaps (bare-slice sibling, Iterate prefix, computed-key trailing slice)


newhoggy:
`#1223`'s fix (and the `/code-review` round that followed it) closed the original comma-grouped `del()` crash plus its `Paren`/`Optional`/nested-`Comma` variants, but surfaced three narrower, genuinely separate gaps in the same family, each pinned as a `*_characterize_preexisting_bug` test in `tests/yq_cli_tests.rs` rather than fixed. Filing them together since they share the same root mechanism: `resolve_dynamic_indexes`'s generic per-branch `Comma` navigation (via `resolve_node`'s `Expr::Comma` arm, `src/jq/eval.rs:12720`) evaluates every branch against the real document, with no way to defer a branch `del()` itself already knows how to handle without navigating it.

## 1. Bare-slice sibling on an Object root — deliberately not chased (real yq's own answer is inconsistent)

`del(.a[0:1], .[2:3])` on `{"a":{"x":1,"y":2},"b":6,"c":9}` still crashes (`Cannot index object with object`) — `yq_del_scalar_slice_parent_path` only rewrites a *chained* slice (`unwrap_path_component`'s result must be `Expr::Pipe`); a bare `Expr::Slice` isn't one, so it's correctly left unrewritten.

Live-verified against yq v4.53.3: real yq's answer for this exact combination is the **unchanged document** — not `.a`'s parent-key dropped (the chained-slice rule) composed with `.[2:3]`'s own bare-root no-op. This is the same "no coherent rule to replicate" territory as `#1223`'s own pinned order-sensitivity known-gap (`test_yq_mixed_bare_root_and_chained_slice_del_known_gap_1223`) — worth investigating on its own, not extending `#1223`'s fix to cover.

Test: `test_1223_bare_slice_sibling_object_target_characterize_preexisting_bug`.

## 2. A comma branch reaching its trailing slice through `Expr::Iterate`

`del(.arr[][0:1], .c)` on `{"arr":[{"x":1,"y":2},{"x":3,"y":4}],"c":9}` still crashes — `navigate_read_only` (the prefix walker `yq_del_scalar_slice_parent_path` uses to confirm a prefix actually exists) only understands `Field`/`Index` steps; an `Iterate` step returns `None` immediately, so the rewrite silently declines.

[...]

Issue #1398, opened 2026-08-21
yq: duplicate keys collapse when input is JSON, and survive iteration where real yq collapses them


newhoggy:
**Severity: Medium** (silently wrong output on `length`/`to_entries` in yq mode; the format
leak makes the same logical document answer differently depending on how it arrived)

## Summary

Two duplicate-key divergences surfaced while writing ADR-0018 that no open issue currently
owns. [#1385](https://github.com/rust-works/succinctly/issues/1385) is filed against **jq
mode** and names #1342/#1343/#1344 as its yq-side continuation; none of the four covers
either of these.

Both verified against pinned **yq v4.53.3** (`tests/data/yq-golden/YQ_VERSION`) at
`e07ad6b33`:

```bash
$ printf '{"b":1,"a":2,"b":3}' > dup.json
$ printf 'b: 1\na: 2\nb: 3\n'   > dup.yaml
```

### 1. `succinctly yq` inherits jq's collapse when the input arrived as JSON

```
$ yq            -o=json -I=0 'length' dup.json   # 3
$ succinctly yq -o=json -I=0 'length' dup.json   # 2   <- wrong
$ succinctly yq -o=json -I=0 'length' dup.yaml   # 3   <- right
```

Real yq answers `3` on **both** inputs. `succinctly yq` flips its answer with the input
format on three of seven filters probed (`length`, `to_entries`, and — via the next item —
`[.[]]`).

Cause, per ADR-0018 rule 2: `DocumentFields::keys_dedup()`
([`src/jq/document.rs`](https://github.com/rust-works/succinctly/blob/main/src/jq/document.rs))
gates duplicate-key collapse on **input format** (JSON `true`, YAML `false`), but the
reference tools decide on **mode**. A per-format trait cannot see which tool is running, so
`syq` reading JSON silently inherits jq's rule. Behavioural rules belong on `EvalSemantics`.

### 2. yq mode does not collapse duplicate keys under iteration

Real yq is internally inconsistent here, and ADR-0018 rule 3 (bug-for-bug fidelity) says to
reproduce the inconsistency rather than smooth it out:

```
$ yq            -o=json -I=0 '[.[]]' dup.yaml    # [3,2]     — iteration collapses
$ succinctly yq -o=json -I=0 '[.[]]' dup.yaml    # [1,2,3]   <- wrong
```

`.[]` is the *only* outlier in real yq — `keys`, `length`, `to_entries`, `from_entries`,
[...]

PR #1415, opened 2026-08-21
feat(jq): yq object slicing follows its own AST child-node layout

### LINKED ISSUE #1102: yq: slicing an object (.[S:E]) follows an internal AST child-node layout, not replicated by succinctly
**Severity: Low**

## Summary

Found while implementing #1065 (deliberately excluded from that fix's
scope — real yq's own object-slicing behavior is not a clean rule worth
replicating as currently understood).

Real yq's slicing of an *object* target doesn't error, doesn't give `[]`,
and doesn't slice by key-insertion-order the way one might guess — it
appears to slice yq's own internal AST child-node list, where a mapping's
children alternate key-node, value-node, key-node, value-node:

```
$ echo '{"a":1,"b":2,"c":3}' | yq -o=json '.[0:2]'
["a",1]
$ echo '{"a":1,"b":2,"c":3}' | yq -o=json '.[1:2]'
[1]
```

`[0:2]` returns the first *two AST children* (key `"a"`, then value `1`),
not the first two key-value pairs. succinctly currently gives no output at
all for `.[0:1]?` on an object (matching jq's own "not sliceable" rule).

## Scope note

Not obviously worth replicating: succinctly's `OwnedValue::Object`
representation has no equivalent "flat, alternating key/value child list"
concept the way yq's internal go-yaml AST does — approximating this
correctly would need either restructuring how objects are represented for
slicing purposes, or a special-cased "iterate key,value,key,value..." view
built just for this operator, both non-trivial for a corner case unlikely
to be depended on intentionally. Filing to make the gap trackable and
documented, not because it's clearly worth fixing as-is.

newhoggy:
## Summary

Real yq's `.[S:E]` on an *object* target doesn't error, doesn't give `[]`, and doesn't
slice by key-insertion-order pairs — it slices yq's own internal AST child-node list,
where a mapping's children alternate key-node, value-node, key-node, value-node:

```
$ echo '{"a":1,"b":2,"c":3}' | yq -o=json '.[0:2]'
["a",1]
```

`[0:2]` returns the first *two AST children* (key `"a"`, then value `1`), not the first
key-value pair. succinctly previously gave `Cannot index object with object` here
(matching jq's own "not sliceable" rule, which is correct for jq mode — real jq has no
object-slicing concept at all).

## Implementation

`SliceBounds::resolve_object_children` (`src/jq/slice.rs`) implements the
bound-resolution rule, verified against real yq v4.53.3 across 20+ probes on 3- and
4-entry objects (full matrix in that method's doc comment). The interesting part: `start`
and a *negative* `end` fold/clamp against the child count `2N` like every other slice
bound, but an *omitted* `end` defaults to the entry count `N`, not `2N` — near-certainly
an upstream bug (yq computes the default end from `length`, which is `N` for a map,
while resolving a negative index against `len(node.Content)`, which is `2N`). Reproduced
anyway, per ADR-0018's bug-for-bug fidelity rule, and documented so a future reader
doesn't "fix" the asymmetry.

Wired into `slice_owned_value_read` (shared by both `eval.rs`'s and `eval_generic.rs`'s
`Targets::Owned` loop — one definition, not two) and a new `StandardJson::Object` arm in
`eval.rs`'s cursor-backed `Expr::Slice` dispatch, since an object target reaches that
arm before any `OwnedValue` is ever materialized. Both gated on `S::TAG == EvalTag::Yq`;
jq mode keeps erroring (regression test included).

## Unplanned side effect

Two `del()`-machinery crashes pinned as `*_characterize_preexisting_bug` placeholders
(#1325 items 2 and 3 — `del(.arr[][0:1], .c)` and `del(.[($k)][0:1], .c)`, both crashing
with `Cannot index object with object` purely because that error came from the *read*
side ever attempting to slice an object at all) now legitimately evaluate and match real
yq exactly:

```
$ echo '{"arr":[{"x":1,"y":2},{"x":3,"y":4}],"c":9}' | yq -o=json 'del(.arr[][0:1], .c)'
{"arr":[]}
$ echo '{"a":{"x":1,"y":2},"c":9}' | k=a yq -o=json 'del(.[env(k)][0:1], .c)'
{}
```

Both verified live against real yq and matched exactly by succinctly post-fix. Updated
the two tests in place (renamed, now asserting the correct behavior) rather than leaving
them describing a bug that no longer exists.

- #1325's item 1 (bare-slice-sibling ordering) was already fixed pre-#1102 by #1219 and
  remains a separate, unrelated known divergence — unaffected, still passing.
- #1325's item 4 (a warning that assignment might develop the same premature-navigation
  crash once an object-slice target gets a `del()`-style carve-out) doesn't trigger here
  since this PR is read-only — confirmed live: `.a[0:1] = x` and `.a[0:1].foo = x` both
  still error exactly as before (same message, same exit code).

Will comment on #1325 once this merges, noting items 2/3 resolved and item 1/4 still
open.

## Test plan

- `test_object_slice_matches_ast_child_layout_matrix_1102`: all 18 probe rows from the
  characterization matrix (N=3).
- `test_object_slice_end_omitted_asymmetry_generalizes_1102`: the `end`-omitted-defaults-
  to-`N` asymmetry on a 4-entry object, guarding against an "obvious" `2N` simplification.
- `test_object_slice_empty_object_and_nested_value_1102`: `{}` and a nested-value case.
- `test_object_slice_owned_target_1102`: reaches `slice_owned_value_read`'s new arm via
  `eval_slice_expr`'s `Targets::Owned` branch specifically (mirrors
  `test_slice_owned_target_scalar_is_empty_array_1065`'s shape), not just the cursor arm.
- `test_object_slice_jq_mode_still_errors_1102`: jq mode regression guard.
- `test_object_slice_yaml_output_known_tag_gap_1102`: known, deliberate gap — real yq's
  slice result keeps the source node's `!!map` tag and key quoting in YAML output;
  `OwnedValue` has no tag/style slot at all, so neither is reachable here. `-o=json`
  output (every case above, and every realistic use of this operator) is unaffected.
  Filed as #1416 (see the review-round comment below for why not #1090, which the PR
  originally cited).
- Full `1223`/`1325`-tagged suite re-run clean (11 tests), including the two updated
  ones.
- Both clippy legs, `cargo build --no-default-features`, `cargo fmt --check`: clean.
- Patch coverage: 96.15% (25/26). The one uncovered line is a defensive
  `unreachable!()` for an invariant (`StandardJson::Object` always materializes to
  `OwnedValue::Object`) that genuinely cannot be exercised by a test.

Fixes #1102

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -201,6 +201,17 @@ the fix) and [#1344](https://github.com/rust-works/succinctly/issues/1344)
 leak or the missing `.[]` collapse, which is why #1398 exists — recording them here is the
 first half of ADR-0018 rule 6, and filing them is the second.
 
+Object slicing (`.[S:E]` on an object, [#1102](https://github.com/rust-works/succinctly/issues/1102))
+is another surface this same root cause reaches: it must materialize the target object into
+an `OwnedValue::Object` (an `IndexMap`) to build yq's AST-child-list view before slicing it,
+which silently collapses a genuine duplicate key the same way `to_owned()` does everywhere
+else in this list — `a: 1\nb: 2\na: 3` sliced with `.[0:6]` real yq returns `["a",1,"b",2,"a",3]`
+(6 children, both `a` entries present); succinctly returns `["a",3,"b",2]` (4 elements, the
+first `a`/`1` pair gone). Unlike the other surfaces above, there is no cursor-preserving
+alternative available here — the operation inherently needs to reorder/slice the entries, not
+just stream them — so this one is a real limit of `OwnedValue::Object`'s representation, not
+a missed wiring like #1343/#1344.
+
 Pulling the other way: [#442](https://github.com/rust-works/succinctly/issues/442),
 [#478](https://github.com/rust-works/succinctly/issues/478) and
 [#868](https://github.com/rust-works/succinctly/issues/868) are closed decisions that
@@ -267,10 +278,15 @@ Float and number formatting ([#1071](https://github.com/rust-works/succinctly/is
 [#1055](https://github.com/rust-works/succinctly/issues/1055)), comment placement
 ([#1079](https://github.com/rust-works/succinctly/issues/1079),
 [#1080](https://github.com/rust-works/succinctly/issues/1080),
-[#1085](https://github.com/rust-works/succinctly/issues/1085)), and the regex engine's
+[#1085](https://github.com/rust-works/succinctly/issues/1085)), the regex engine's
 zero-width-match handling
 ([#1255](https://github.com/rust-works/succinctly/issues/1255) — real yq uses Go's
-`regexp`). Also see
+`regexp`), and a missing explicit-tag slot: `OwnedValue` has no field for it, so any
+computed/constructed value — including an object-slice result (#1102) — loses the source
+node's `!!map`/`!!seq` tag and quoting style in YAML output (`{"a":1,"b":2} | yq
+'.[0:2]'` is `!!map\n- "a"\n- 1` on real yq, plain `- a\n- 1` on succinctly); `-o=json`
+output is unaffected, since neither appears in JSON
+([#1416](https://github.com/rust-works/succinctly/issues/1416)). Also see
 [yq Query Language Reference § Known Limitations](../../reference/yq-language.md#known-limitations)
 for the feature-level gaps (`*`/`+` are not cartesian generators; position builtins after DOM
 conversion).
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -615,14 +615,14 @@ fn eval_single<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
             // across all three) rather than jq's own null-passthrough rule
             // or "not sliceable" error -- checked before the jq-only Null
             // arm below so this wins under yq mode. Object targets are
-            // deliberately excluded: real yq's own slicing there follows
-            // its internal AST child-node layout (`{"a":1,"b":2} | .[0:2]`
-            // -> `["a",1]`, alternating keys and values), not a clean
-            // "empty container" rule, and isn't replicated here (#1102).
-            // Assignment targets are a separate, unverified case -- real
-            // yq's own behavior there is a silent no-op, not this rule
-            // (#1101) -- see `is_yq_slice_empty_container_scalar`'s doc
-            // comment, which this arm mirrors.
+            // deliberately excluded from *this* rule: real yq's own slicing
+            // there follows its internal AST child-node layout instead (not
+            // a clean "empty container" rule) -- see the dedicated
+            // `StandardJson::Object` arm below (#1102). Assignment targets
+            // are a separate, unverified case -- real yq's own behavior
+            // there is a silent no-op, not this rule (#1101) -- see
+            // `is_yq_slice_empty_container_scalar`'s doc comment, which this
+            // arm mirrors.
             StandardJson::Null | StandardJson::Number(_) | StandardJson::Bool(_)
                 if S::TAG == EvalTag::Yq =>
             {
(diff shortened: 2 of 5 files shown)


newhoggy:
## Code review round

Substantial findings, all addressed:

**Fixed:**
- **`eval_generic.rs`'s own missing arm**: `slice_one_generic` (the computed-slice-bound
  evaluator's native handler) never got the object arm — only `eval.rs`'s cursor-backed
  literal-bound path and the shared `slice_owned_value_read` did. `.[0:2]` worked on an
  object; `.[.from:.to]` on the same object still errored. Fixed by giving
  `slice_one_generic` the matching arm (now shares `slice_object_as_yq_children`, made
  `pub(crate)`).
- **`path()`/`getpath` round-trip break**: `path(.[0:2])` on an object went from erroring
  (pre-#1102) to succeeding with an unusable descriptor — `getpath` on that same
  descriptor still errored, since its own hand-rolled slice-descriptor dispatch had no
  `Object` arm. A narrower regression than the pre-fix all-or-nothing error. Fixed by
  giving `builtin_getpath` the matching arm — read-only, so `getpath(path(x)) == x`
  holds again. `setpath`/`delpaths` deliberately don't get the same treatment (#1157's
  write-side scope).
- **Efficiency**: `slice_object_as_yq_children` built the object's whole `2N`-length
  child list before slicing it down, discarding everything outside the range. Rewritten
  to materialize only the requested range via `IndexMap::get_index` (O(1) per child).
- **Duplication**: `SliceBounds::resolve`/`resolve_object_children` were two hand-copied
  bodies differing only in clamp length and the omitted-`end` default — unified into one
  `resolve_with_end_default`.
- **Stale doc comments**: two pre-existing comments (`is_yq_slice_empty_container_scalar`,
  `slice_owned_value_read`'s header) still described object slicing as unimplemented.

**Documented per ADR-0018 rule 6 (real, unrelated-to-fix limitations this feature
inherits):**
- A duplicate YAML key is silently collapsed during the materialization slicing needs —
  same `OwnedValue::Object`/`IndexMap` root cause as this repo's other duplicate-key gaps
  (#1398 family), with no cursor-preserving alternative available here (slicing
  inherently reorders/subsets entries, unlike the read-only surfaces #1343/#1344 cover).
  Added to `docs/compliance/yq/limitations.md`'s existing "Duplicate mapping keys"
  section, plus a known-gap regression test.
- The source node's `!!map` tag and key quoting style are lost in YAML output —
  `OwnedValue` has no tag/style slot at all. My original test comment cited #1090 (per
  the implementation plan), but #1090 turned out to be narrowly about `!!float`
  tag-emission specifically — filed the correct, general issue as **#1416** and fixed the
  citation.

**Confirmed clean (verified live against the pinned oracle):**
- Assignment (`.a[0:1] = x`, `.a[0:1].foo = x`) is completely unaffected — still errors
  exactly as before. #1325's item 4 warning (about assignment developing the same
  premature-navigation crash) doesn't trigger for this read-only PR.
- #1325's item 1 (bare-slice-sibling ordering divergence) was already fixed pre-#1102 by
  #1219 and remains separately tracked — unaffected, still passing.

**Investigated and confirmed pre-existing (not fixable here):** three lines patch
coverage flags as uncovered — two are `unreachable!()` defensive arms for invariants
that provably always hold (`eval.rs:675`, `eval_generic.rs:3646`), and the third
(`eval.rs:8184`, the `optional`-gated branch of `getpath`'s new object-descriptor arm)
is dead code for the exact same reason its pre-existing `Array`/`String` sibling arms
are (verified via `cargo llvm-cov`'s raw hit counts: lines 8160/8168/8189 — the sibling
arms' and the catch-all's own `if optional` guards — all show **zero hits** too, on
`main`, unrelated to this PR). `optional` is never forced `true` on the path that reaches
`builtin_getpath`, matching this codebase's established "optional never forced true in
nested calls" pattern.

Full suite, both clippy legs, `cargo build --no-default-features`, `cargo fmt --check`:
all clean after this round. Patch coverage: 94.44% (51/54), with the remaining 3 lines
explained above.

Issue #406, opened 2026-07-27
yaml: an anchor at the end of a compact mapping entry's line swallows the nested value


newhoggy:
Found while reviewing the #372 branch. Pre-existing, and not a regression from it.

A compact mapping entry (a `k: v` inside a block sequence item) whose value is a nested block
starting on the next line parses correctly — until an anchor is added, at which point the nested
block collapses to a scalar:

```console
$ printf '- k:\n    b: 1\n'      | syq -o json -I0 '.'   # [{"k":{"b":1}}]   correct, matches yq
$ printf '- k: &a\n    b: 1\n'   | syq -o json -I0 '.'   # [{"k":"b"}]       wrong; yq: [{"k":{"b":1}}]
```

The equivalent **block** mapping entry is correct, so this is specific to the compact path:

```console
$ printf 'k: &a\n  b: 1\nc: *a\n' | syq -o json -I0 '.'   # {"k":{"b":1},"c":{"b":1}}  matches yq
```

## Cause

`parse_compact_mapping_entry` decides "inline value" by `!self.at_line_end()`. With `&a` present
that test passes even though the anchor is the *last* token on the line, so the entry takes the
inline branch and calls `parse_inline_value` on what is effectively an empty remainder; the
nested block below is then misread.

#372 added anchor handling to that branch with the comment "An anchor or alias prefixes the
value here exactly as it does in `parse_value`". The parity is not quite exact: `parse_value`
falls through to a path that handles a next-line nested block, whereas this copy always goes to
`parse_inline_value`. So the comment overclaims, and the anchor is now registered against the
collapsed node.

## Consequence for aliases

Because the anchor is now registered (it was not before #372), an alias to it propagates the
collapsed value instead of `null`:

```console
$ printf '- k: &a\n    b: 1\n- c: *a\n' | syq -o json -I0 '.'
[{"k":"b"},{"c":"b"}]        # before #372: [{"k":"b"},{"c":null}]
                             # yq:          [{"k":{"b":1}},{"c":{"b":1}}]
```

Both are wrong; the post-#372 form is at least self-consistent (the alias equals its anchor's
value). Fixing the underlying parse fixes both.

## Fix

In `parse_compact_mapping_entry`, take the "value on the next line" branch when the position is
at line end *after* consuming an anchor, rather than deciding before. Then correct the parity
comment, or route the case through the same handling `parse_value` uses.

## Also seen

[...]

PR #1420, opened 2026-08-21
refactor(yaml): factor parser.rs's 8 UnexpectedCharacter sites into one UTF-8-correct helper

### LINKED ISSUE #1187: yaml: parser.rs duplication cleanup (trailing-content/at_line_end whitespace defs, error construction, byte-set checks, Latin-1 char cast)
**Severity: Low**

## Summary

Split from #902's item 4 (a set of duplication observations from that issue's review), found while implementing #902 (items 1 & 2, PR TBD). Not blocking that fix — cleanup/reuse opportunities, no behavior change needed.

Four independent duplication findings in `src/yaml/parser.rs`, all pre-existing:

1. **`reject_trailing_flow_content` and `at_line_end()` independently answer overlapping "what comes next on this line" questions with different whitespace definitions.** `at_line_end()` only treats a plain space as skippable inline whitespace, not a tab — every one of its 9 other callers only uses it for soft branching (a stray tab picks a slightly different but still-correct path), so the gap has never surfaced as a hard error the way it did for #878's original (now-fixed, tab-blind) version of the trailing-content check.

2. **`YamlError::UnexpectedCharacter { offset, char: self.peek().map_or(...), context }` construction appears 8 times verbatim** across `parser.rs` — never factored into a shared `err_unexpected_char(&self, context)` helper.

3. **The `[`/`{`/`|`/`>` byte-set is hand-checked a second time as a pure lookahead classification** in `parse_mapping_entry`'s continuation-line logic (~line 2312), separate from both `try_dispatch_flow_or_block_value` and `parse_explicit_key`'s own copy.

4. **`reject_trailing_flow_content`'s error uses `byte as char` (Latin-1 cast, not real UTF-8 decoding)** for the offending byte — a pre-existing pattern used at 7 other sites in this file, but this is the first one reachable on plain non-ASCII prose a user might actually type right after a flow collection (the others only fire on structural syntax like a missing `:`), so mojibake in the error message (`'æ'` instead of `'日'`) is newly *visible* here even though the underlying idiom predates #902.

## Scope note

Filing to make these trackable, not blocking — pure cleanup/reuse, no behavior change. #1 in particular would need re-verifying all 9 of `at_line_end()`'s other call sites' behavior doesn't change in an unwanted way if `at_line_end()` itself is widened, rather than adding a parallel helper.

newhoggy:
## Summary

Issue #1187 listed four independent duplication findings in `src/yaml/parser.rs`. This
PR addresses items 2 and 4 (which turned out to be the same 8 call sites, not two
separate problems) and documents why items 1 and 3 are being left alone.

### Items 2 + 4: fixed

Every one of the 8 sites constructing `YamlError::UnexpectedCharacter` hand-rolled it,
and 7 of the 8 used `self.peek().map_or('\0', |b| b as char)` — a Latin-1 cast, not a
real UTF-8 decode, so a multi-byte character at the offending position rendered as
mojibake instead of itself (`'æ'` instead of `'日'`). The 8th
(`reject_trailing_flow_content`) had the identical bug via `self.input[i] as char`.

Factored into one `err_unexpected_char(&self, offset, context)` helper that decodes via
`text::utf8::decode_code_point` against the byte slice starting at `offset` (a single
byte isn't enough to decode a multi-byte sequence), falling back to `'\0'` on EOF or
invalid UTF-8 — matching the old code's EOF fallback exactly, so this is a strict
improvement (correct non-ASCII rendering) with zero observable change for ASCII.

```
$ printf 'a: [1, 2] 日\n' | succinctly yq '.'
# before: unexpected character 'æ' at offset 10: ...  (mojibake)
# after:  unexpected character '日' at offset 10: ...  (correct)
```

Takes `offset` explicitly rather than reading `self.pos` directly, since
`reject_trailing_flow_content` finds the offending byte via its own local scan cursor,
before `self.pos` itself has advanced past it — a `self.pos`-only helper couldn't have
covered all 8 sites.

**Coverage note**: 7 of the 8 call sites had never been exercised by any test in this
repo's history (confirmed by grepping the whole suite before this PR). Constructed
live-verified repros for 6 of the 7 (each needed understanding the specific caller-side
lookahead that greenlights reaching that branch — this parser is forgiving in ways that
make many "obviously wrong" inputs still parse successfully). The 7th
(`parse_block_scalar_header`'s catch-all) is genuinely unreachable — all three callers of
`parse_block_scalar` already gate on `peek()` being `|`/`>` immediately before calling,
with nothing in between that advances `self.pos`. Documented with a comment matching this
file's own existing precedent for the identical situation
(`parse_documents`'s `#`/line-break defensive arms).

### Item 1: investigated, not touched

`at_line_end()` and `reject_trailing_flow_content` do independently define "inline
whitespace" differently (space-only vs. space-or-tab) — but this is **already a
deliberate, documented divergence**, not an oversight: `reject_trailing_flow_content`'s
own doc comment (from a recent commit, `80818b610`, "fix(yaml): accept a tab before a
trailing flow-collection comment") explicitly explains why it can't reuse
`at_line_end()` — every one of `at_line_end()`'s other 9 call sites only uses it for soft
branching (a stray tab picks a slightly different but still-correct path), while
`reject_trailing_flow_content` is the one caller that turns a wrong `false` into a hard
error, which is exactly what made the gap observable and led to the current separation.
Merging them back would reintroduce the tab-bug this separation exists to fix. Verified
all 9 other call sites directly — every one is soft branching, confirming the doc
comment's claim.

### Item 3: investigated, not touched

The `[`/`{`/`|`/`>` byte-set is checked in three places, but two of the three duplicate
sites already carry their own doc comments explaining why they weren't merged into the
canonical `try_dispatch_flow_or_block_value`: `parse_explicit_key`'s copy wraps the key
in its own `write_bp_open`/`write_bp_close` pair that the shared helper can't represent
without double-wrapping the BP tree (explicitly flagged as "a 5th, deliberately-unmerged
copy", referencing the exact bug classes — #325, #372, #406, #224, #785, #864 — that
motivated the shared helper's own existence in the first place). `parse_mapping_entry`'s
copy is a pure lookahead (reset-and-return, never actually parses), a genuinely different
shape from the other two. Given both existing doc comments already establish that
merging control flow here is what causes bugs in this exact area, and the byte-set itself
is a fixed 4-byte YAML-grammar constant unlikely to change, I judged the risk of touching
a working hand-rolled recursive-descent parser not worth the low value of deduplicating 4
byte literals across 3 sites with genuinely different surrounding shapes.

## Test plan

- `test_trailing_content_after_flow_collection_reports_real_utf8_char_1187`: the mojibake
  fix itself (multi-byte character + ASCII-unaffected sanity check).
- `test_err_unexpected_char_remaining_call_sites_1187`: live-verified repros for the 5
  reachable-but-previously-untested sites, plus the compact-mapping/alias one covered
  above (6 of 8 total newly tested; the other 2 of 8 were already covered by existing
  tests).
- Standard clippy leg, the `broadword-yaml` clippy leg (this repo's own documented
  `--all-features` blind spot for x86/SIMD YAML code), `cargo build --no-default-features`,
  `cargo fmt --check`: all clean.
- Patch coverage: 91.18% (31/34); the 3 uncovered lines are the confirmed-unreachable
  defensive arm, explained above.

Fixes #1187

### DIFF
--- a/src/yaml/parser.rs
+++ b/src/yaml/parser.rs
@@ -35,6 +35,7 @@ use std::collections::BTreeMap;
 use super::error::YamlError;
 use super::line_break::{is_line_break, line_break_len};
 use super::simd;
+use crate::text;
 
 /// Node type in the YAML structure tree.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -1010,6 +1011,37 @@ impl<'a, const HAS_CR: bool> Parser<'a, HAS_CR> {
             line_start -= 1;
         }
         self.pos - line_start
+    }
+
+    /// Build a `YamlError::UnexpectedCharacter` for the byte at `offset`
+    /// (#1187): every one of this file's 8 call sites built this variant by
+    /// hand, and 7 of the 8 used `self.peek().map_or('\0', |b| b as char)`
+    /// -- a Latin-1 cast, not a real UTF-8 decode, so a multi-byte character
+    /// at the offending position rendered as mojibake instead of itself.
+    /// Decodes via [`crate::text::utf8::decode_code_point`] instead,
+    /// against the byte slice starting at `offset` (not just the one byte
+    /// `self.peek()`/`self.input[offset]` would give, which isn't enough to
+    /// decode a multi-byte sequence). Falls back to `'\0'` on EOF or
+    /// invalid UTF-8 at `offset`, matching the old code's EOF fallback
+    /// exactly and treating a malformed sequence the same way.
+    ///
+    /// Takes `offset` explicitly rather than reading `self.pos`: the one
+    /// site with a genuinely different call shape
+    /// ([`Self::reject_trailing_flow_content`]) finds the offending byte
+    /// via a local scan cursor, before `self.pos` itself has advanced past
+    /// it.
+    fn err_unexpected_char(&self, offset: usize, context: &'static str) -> YamlError {
+        let char = self
+            .input
+            .get(offset..)
+            .and_then(text::utf8::decode_code_point)
+            .and_then(|(cp, _len)| char::from_u32(cp))
+            .unwrap_or('\0');
+        YamlError::UnexpectedCharacter {
+            offset,
+            char,
+            context,
+        }
     }
 
     /// Check if at end of meaningful content on this line.
@@ -2511,11 +2543,9 @@ impl<'a, const HAS_CR: bool> Parser<'a, HAS_CR> {
 
         // Expect colon
         if self.peek() != Some(b':') {
-            return Err(YamlError::UnexpectedCharacter {
-                offset: self.pos,
-                char: self.peek().map_or('\0', |b| b as char),
-                context: "expected ':' after key in compact mapping",
-            });
+            return Err(
+                self.err_unexpected_char(self.pos, "expected ':' after key in compact mapping")
+            );
         }
         self.advance(); // Skip ':'
 
@@ -2771,11 +2801,7 @@ impl<'a, const HAS_CR: bool> Parser<'a, HAS_CR> {
 
         // Expect colon
         if self.peek() != Some(b':') {
-            return Err(YamlError::UnexpectedCharacter {
-                offset: self.pos,
-                char: self.peek().map_or('\0', |b| b as char),
-                context: "expected ':' after key",
-            });
+            return Err(self.err_unexpected_char(self.pos, "expected ':' after key"));
         }
         self.advance(); // Skip ':'
 
@@ -3812,11 +3838,9 @@ impl<'a, const HAS_CR: bool> Parser<'a, HAS_CR> {
 
         // Expect and skip colon
         if self.peek() != Some(b':') {
-            return Err(YamlError::UnexpectedCharacter {
-                offset: self.pos,
-                char: self.peek().map_or('\0', |b| b as char),
-                context: "expected ':' in implicit flow mapping entry",
-            });
+            return Err(
+                self.err_unexpected_char(self.pos, "expected ':' in implicit flow mapping entry")
+            );
         }
         self.advance();
         self.skip_flow_whitespace();
@@ -3999,11 +4023,7 @@ impl<'a, const HAS_CR: bool> Parser<'a, HAS_CR> {
             self.pos = i;
             return Ok(());
         }
-        Err(YamlError::UnexpectedCharacter {
-            offset: i,
-            char: self.input[i] as char,
-            context: "after a flow collection's closing delimiter",
-        })
+        Err(self.err_unexpected_char(i, "after a flow collection's closing delimiter"))
     }
 
     /// Parse a flow sequence: `[item1, item2, ...]`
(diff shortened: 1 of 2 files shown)


newhoggy:
## Code review round (10 finder angles)

**Fixed — a real regression:**
- `err_unexpected_char`'s first cut fell back to `'\0'` not just at true EOF but for
  *any* byte that isn't valid UTF-8 at that offset — silently embedding a literal NUL
  byte into the error string for a malformed (not just missing) byte. Confirmed live via
  hexdump: `0xFF` where a delimiter was expected produced `unexpected character '<NUL>'
  ...`. The old `byte as char` cast never failed on any non-EOF byte, so this contradicted
  the helper's own doc comment. Fixed by distinguishing true EOF (`self.input.get(offset)
  == None`, still `'\0'`) from a present-but-invalid byte (now falls back to the old
  Latin-1 rendering of that byte, not `'\0'`) — verified: `0xFF` now renders as `'ÿ'`
  again, matching pre-#1187 output exactly, while valid multi-byte characters still get
  the real fix.

**Fixed — coverage gap in the PR's own coverage-completeness claim:**
- `parse_flow_sequence_inner`'s comma-check ("expected ',' or ']' in flow sequence") was
  the one call site my earlier coverage commit's own doc comment claimed to cover but
  didn't. Added a live-verified repro (`a: ["x" b]`).
- Tightened one assertion that could have silently passed even if a regression routed
  input through the wrong call site: `"expected ':' after key in compact mapping"`
  contains `"expected ':' after key"` as a literal prefix, so the plain-mapping test now
  also asserts the compact-mapping variant is *absent*.

**Fixed — real UB, unrelated to the coverage question itself:**
- The new malformed-byte test originally used `unsafe { str::from_utf8_unchecked(...) }`
  over a genuinely invalid byte sequence to build the repro — real UB per the safety
  contract, even though the only use was writing the bytes back out to a subprocess's
  stdin. Replaced with a small raw-bytes stdin helper (`run_yq_stdin_bytes_with_stderr`,
  mirroring `text_cli_tests.rs`'s existing `run_validate_stdin` pattern), no `unsafe`.

**Filed as follow-ups (real findings, correctly out of this PR's scope):**
- #1422 — `src/yaml/validate.rs` (`InvalidEscape`) and `src/json/validate.rs` (5 sites)
  have the identical Latin-1-cast bug this PR fixed in `yaml/parser.rs`, unpatched.
  Different error-type shapes in different files; a shared `text::utf8` primitive is the
  right fix, not expanding this PR's scope.
- #1423 — `text::utf8::decode_code_point` (pre-existing, unmodified here) accepts
  overlong encodings and CESU-8-style surrogate sequences without rejecting them; this PR
  is the first production caller to expose that leniency to untrusted byte input (via
  error messages only — cosmetic, no panic/crash/security impact).

**Investigated and declined:**
- A suggestion to replace `parse_block_scalar_header`'s provably-dead catch-all arm with
  `unreachable!()`/`debug_assert!()` instead of a Result-based error. Declined: this
  codebase was already bitten once (#1098) by turning a graceful, Result-based parse
  error reachable from external input into a panic for a "should never happen" case that
  later did happen. A panic here would be a live DoS vector if the invariant is ever
  violated by a future caller; a misleading-but-graceful error is the safer failure mode.
  Left as-is.
- The true-EOF (`None`) arm of `err_unexpected_char` also carries no test coverage.
  Investigated whether it's provably dead the same way `parse_block_scalar_header`'s arm
  is: 2 of the 8 call sites have their own dedicated `UnexpectedEof` guard immediately
  before reaching this call (so `None` is unreachable there specifically), and the
  remaining key-colon sites are only entered once a `looks_like_*_entry`-style lookahead
  has already confirmed a `:` exists later in the input — but proving that generalizes to
  full unreachability across all remaining sites needs more rigor than this narrow
  cleanup PR's scope warrants. Documented honestly as unverified rather than either
  claiming a proof I don't have or leaving it silently unexplained.

Full suite (single-threaded), standard clippy leg, the `broadword-yaml` leg (this repo's
own documented blind spot), `cargo build --no-default-features`, `cargo fmt --check`: all
clean after this round and after a subsequent rebase onto latest `main`. Patch coverage:
88.89% (32/36); the 4 uncovered lines are the true-EOF fallback (documented above) and
the confirmed-dead `parse_block_scalar_header` arm (documented in the original commit).

PR #1433, opened 2026-08-22
fix(jq): track path() through a variable bound outside path()'s argument

### LINKED ISSUE #844: jq: path() doesn't track a variable bound outside its own argument (E as $x | path($x | ...))
**Severity: Medium**

## Summary

`path(...)` does not track a variable bound by `E as $x | ...` *outside* the `path(...)` call itself — navigating into such a variable inside `path(...)` raises `#530`'s "Invalid path expression" instead of resolving the real path, even though the variable was bound from `.` (the true, path-trackable input) via ordinary navigation. This reproduces with no `try`/`catch` involved at all, so it's unrelated to control-flow/escape handling — a plain, deterministic filter.

## Repro

```
$ echo '{"a":1,"b":2}' | jq -c '. as $orig | path($orig | .b)'
["b"]
$ echo '{"a":1,"b":2}' | succinctly jq -c '. as $orig | path($orig | .b)'
jq: error (at <stdin>:1): Invalid path expression with result {"a":1,"b":2}
```

Real jq resolves `$orig | .b` as path `["b"]` (since `$orig` was itself bound from `.` by plain navigation, one step before `path(...)` even starts). succinctly instead reports the *entire* `$orig` value as the invalid path expression — meaning `.b` appears to never get applied to it at all inside the resolver, and whatever the resolver ends up checking against `#530` is just `$orig`'s raw value.

## Root cause (partial — not fully diagnosed)

`resolve_node`'s `Expr::As` arm (`src/jq/eval.rs`) substitutes each bound value into `body` via `substitute_var(body, var, &bound)` before calling `resolve_node` on the substituted AST — this is the mechanism `path()` uses generally to track `$x`-bound values (see that arm's own doc comment, added by #824/#832-adjacent work, and used elsewhere for `FirstExpr`/`IndexExpr` substitution too). That path is inside `path(...)`'s own argument, though (`. as $orig | path($orig | .b)` binds `$orig` *outside* `path(...)`'s argument expression) — so by the time `resolve_node` runs, `$orig` may already have been substituted or otherwise resolved by the *value*-position evaluator (whatever evaluates the top-level pipeline `. as $orig | path(...)`), and `path(...)`'s own argument-parsing/resolution may not see a raw `Expr::Var` to substitute against at all — or does see one, but mis-resolves it. This needs more investigation than this report does; flagging the confirmed divergence and repro rather than the exact code path, since chasing it further was itself out of scope for the issue this was found while investigating (see below).

## Why this is separate from #832 / #843

Found while looking for a repro of "a `path(try ... catch ...)` handler produces non-empty output before failing" for #832's regression-test coverage (PR #841) — every payload-navigation-based attempt at that repro hit #843's "caught value isn't path-trackable" gap instead, so a variable bound to the *original* document (not the caught payload) was tried as an alternative and hit *this*, third, unrelated gap instead. It reproduces with zero `try`/`catch`/`path(...)`-argument-only involvement, so it's not the same code path as #832 (`resolve_catch`) or #843 (caught-payload trackability) — it's a general limitation of how `path(...)` (or the pipeline around it) handles variables bound upstream of it.

## Suggested fix direction

Needs investigation first (this report doesn't pin the exact file:line): trace how `. as $x | path(E)` is evaluated end-to-end — whether `path(...)`'s own resolver ever gets a chance to substitute `$x` via the existing `substitute_var` mechanism, or whether the outer `as`-binding evaluates `path(...)` as an ordinary value-producing filter first (losing path-tracking before `resolve_node` ever runs). A precursor spike to pin the root cause down is probably warranted before scoping a fix — this repo's `grind-backlog` risk tiers would put root-cause investigation at Tier 1-2 and the eventual fix at Tier 2-3 depending on what's found.

Found during review of PR #841 (#832).

newhoggy:
## Summary

Fixes #844: `. as $orig | path($orig | .b)` raised `Invalid path expression` in succinctly's jq/yq even though real jq resolves it to `["b"]`. `$orig` is bound one step *before* `path(...)` starts, from `.` itself — real jq preserves path-trackability through a variable only when its bind source is a syntactic passthrough of `.`; genuine navigation (`.a as $x | ...`) does not survive either way, and this fix keeps that negative case intact.

Two independent call sites of `substitute_var` (`eval_as` for bindings outside `path()`, `resolve_node`'s own `Expr::As` arm for bindings inside it) discarded a variable's provenance the moment they spliced its bound value into the AST as a plain literal/object/array node, so `resolve_node`'s catch-all always marked it untrackable regardless of where the value actually came from.

- Adds `Expr::TrackedVar`, a parser-unreachable AST marker synthesized only when the bind source is statically verified as a passthrough of `.` (`is_identity_passthrough`).
- `resolve_node`'s new arm for it decides trackability **lazily**, at each use site, by comparing the frozen snapshot to the ambient value — sound even when `.` changes between bind and use (falls back to untracked, never emits a wrong path), and correct with respect to `Comma`-sibling independence (the case that sank an earlier, released attempt at this exact issue).
- `substitute_var` is split into `substitute_var`/`substitute_var_tracked` thin wrappers over a shared `substitute_var_impl`, so none of the ~150 unrelated pre-existing callers (destructuring, `--arg`/`--argjson` splicing, CLI) change behavior.
- The fix isn't `path()`-only — `resolve_dynamic_indexes`/`resolve_node` back all four path-consuming entry points, so `=`/`del()` through a passthrough-bound variable are fixed for free.

## Test plan

- [x] `cargo build` clean under `cli`, `--all-features`, and `--no-default-features` (no_std)
- [x] Both CI clippy legs (`--all-targets --all-features` and the SIMD feature-set leg) clean with `-D warnings`
- [x] 7 new regression tests colocated with the existing #843/#986/#989 trackability tests, covering: the issue's repro, a negative control (navigation-sourced binding must keep erroring), `Comma`-sibling independence, the inner `resolve_node::Expr::As` binding site, chained passthrough bindings, assign/`del` through a passthrough-bound variable, and a bind-time-gate probe (`0 as $x | path($x)` must still error)
- [x] Full oracle differential against real `jq` 1.7.1 for every case above — all match exactly
- [x] Full `cargo test` suite: 0 failures across every test binary

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -419,6 +419,30 @@ Unlike the positive case above, `?` does not suppress this one —
 [#498](https://github.com/rust-works/succinctly/issues/498) — because it is a write-time
 bounds check, not a failure to collect the path.
 
+A variable bound outside `reduce`/`foreach` and referenced from inside `UPDATE` is a third,
+narrower gap, found reviewing [#844](https://github.com/rust-works/succinctly/issues/844):
+
+| Filter                                         | Input     | jq   | succinctly                                    |
+|------------------------------------------------|-----------|------|-----------------------------------------------|
+| `path(. as $x \| reduce (1,2) as $i (0; $x))`  | `{"a":1}` | `[]` | `Invalid path expression with result {"a":1}` |
+| `path(. as $x \| foreach (1,2) as $i (0; $x))` | `{"a":1}` | `[]` | `Invalid path expression with result {"a":1}` |
+
+#844 taught `resolve_node` to recognize `Expr::TrackedVar` (a variable frozen from a
+statically-verified passthrough of `.`, per that issue's own `is_identity_passthrough`) so
+that referencing such a variable inside `path(...)` stays trackable. `resolve_node` has no
+`Expr::Reduce`/`Expr::Foreach` arm of its own, though, so either one reaching `path(...)`'s
+resolver falls to `resolve_leaf`'s catch-all regardless of what its `UPDATE` expression
+does internally — a bare `$x` reference inside `UPDATE`, even a `TrackedVar` one, never gets
+the chance to reach the new arm, because `resolve_node` never descends into `Reduce`/
+`Foreach` at all. Real jq has no such gap because `reduce`/`foreach` are sugar over the same
+primitive `as`-binding/iteration machinery every other construct uses, so their own
+trackability was never a separate case to begin with. This is a read-only, safe divergence
+(succinctly never emits a *wrong* path here, only refuses a filter jq accepts) — not fixed
+here; a proper fix needs `resolve_node` to interleave path resolution with `reduce`/
+`foreach`'s own fold loop (evaluating `UPDATE` against each intermediate accumulator through
+`resolve_node` rather than the ordinary value evaluator), which is new machinery beyond a
+single arm.
+
 ## Refusing an allocation jq does not survive
 
 `setpath` takes its array index from the document, so the array it pads is sized by the
--- a/docs/plan/jq-path-trackability-deferral.md
+++ b/docs/plan/jq-path-trackability-deferral.md
@@ -372,6 +372,10 @@ measured slice-by-slice rather than as one combined change.
    carve-outs (the `bare_navigation_primitive` check, the `is_untracked_navigation_error`
    distinction) that a mechanical "just propagate whatever came in" migration could silently
    break — read that arm's own doc comment in full before touching it.
+   `Expr::As`'s own gap (a variable bound *outside* `path(...)`'s argument, or from a source
+   that isn't a syntactic passthrough of `.`, losing trackability across the binding) was closed
+   by #844 — see `is_identity_passthrough` and `resolve_node`'s `Expr::TrackedVar` arm in
+   `src/jq/eval.rs`. The others in this list remain open.
 5. **`del`/assignment write-path semantics**, not just `path()`'s read-only output. #972's own
    revert (on a related but different fix in this same code region) was specifically about
    silent data corruption on the write path once multi-key index/slice fan-outs were involved.
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -415,6 +415,7 @@ fn rewrite_namespaced_calls(expr: Expr) -> Expr {
         | Expr::RecursiveDescent
         | Expr::Literal(_)
         | Expr::Var(_)
+        | Expr::TrackedVar(_)
         | Expr::Loc { .. }
         | Expr::Env
         | Expr::Not
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -2840,6 +2840,7 @@ fn contains_split_doc(expr: &Expr) -> bool {
         | Expr::Not
         | Expr::Format(_)
         | Expr::Var(_)
+        | Expr::TrackedVar(_)
         | Expr::Loc { .. }
         | Expr::Env
         | Expr::Break(_)
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -792,6 +792,11 @@ fn eval_single<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
             // In practice, variables are resolved by eval_as which substitutes them
             QueryResult::Error(EvalError::new(format!("undefined variable: ${name}")))
         }
+        // A frozen variable snapshot from `substitute_var_tracked` (#844).
+        // Outside `path()`/`del()`/assignment resolution this is just an
+        // ordinary bound value -- `resolve_node`'s own `Expr::TrackedVar`
+        // arm is what actually decides path-trackability.
+        Expr::TrackedVar(v) => QueryResult::Owned((**v).clone()),
         Expr::Loc { line } => {
             // $__loc__ returns {"file": "<stdin>", "line": N}
             // where N is the 1-based line number in the jq filter source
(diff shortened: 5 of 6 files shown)


newhoggy:
## Code review (9-agent parallel pass)

The design is sound overall — it can under-track a variable's trackability but the review found no case where it produces a *wrong* path. Six issues surfaced, most severe first:

### 1. `is_identity_passthrough`'s doc comment overclaims jq's actual rule (`src/jq/eval.rs:16214`)
The comment states real jq "never" preserves `path()` trackability through anything but a literal `.` bind source, and `docs/plan/jq-path-trackability-deferral.md` marks `Expr::As`'s gap as fully "closed by #844." Both are empirically false against the pinned oracle (`/usr/bin/jq` 1.7.1):

```
$ echo '{"a":1,"b":2}' | jq -c '. as $x | (if true then $x else $x end) as $y | path($y.b)'
["b"]
$ echo '{"a":1,"b":2}' | jq -c '. as $x | (try $x catch $x) as $y | path($y.b)'
["b"]
$ echo '{"a":1,"b":2}' | jq -c '. as $x | ($x // 5) as $y | path($y.b)'
["b"]
```
All three still raise `Invalid path expression` on this branch, since `is_identity_passthrough` only recognizes `Expr::Identity`/`Paren`/`TrackedVar` as passthrough. None of the four new `_844` tests probe a non-bare-`.` passthrough source, so this gap has no regression coverage and reads as "done" per the deferral doc.

### 2. Measured perf regression for the idiomatic loop pattern (`src/jq/eval.rs:14193`)
`Expr::TrackedVar`'s `resolve_node` arm runs a full deep structural equality check of the entire frozen snapshot on every dereference — a cost that didn't exist pre-PR (the variable was previously spliced in as a literal AST subtree, no runtime comparison).

Measured live: input `{"root_marker":[100000 ints],"records":[300 objects]}`, filter `path(. as $root | .records[] as $rec | $root.root_marker[0])` took **5.7s vs 3.6s** for the same filter without `$root` — confirmed to scale linearly with reference count in a second isolated test. This is the same class of regression flagged before in #106/#888/#1213, now newly present in exactly the "bind once, loop, reference again" idiom this PR is meant to enable.

### 3. `reduce`/`foreach` not tracked (`src/jq/eval.rs:14295`)
`resolve_node` has no `Expr::Reduce`/`Expr::Foreach` arm, so a `TrackedVar` referenced inside either body always resolves untracked even though jq tracks it:
```
$ echo '{"a":1}' | jq -c '. as $x | reduce (1,2) as $i (0; $x) | path(.)'
```
returns `[]` in real jq, errors in succinctly. Safe (never wrong), but per ADR-0018 this divergence should be recorded in `docs/compliance/jq/limitations.md` and isn't.

### 4. Trackability gate duplicated between `resolve_node` and `eval_as` (`src/jq/eval.rs:14148`, `17878`)
`let wrap_tracked = is_identity_passthrough(expr); if wrap_tracked { substitute_var_tracked(...) } else { substitute_var(...) }` is copy-pasted verbatim in both places instead of one shared helper. A future refinement of the gate (e.g. to close finding #1) only needs to touch one call site to compile — the other silently keeps the old behavior, reintroducing the exact `path()`-internal vs. general-`eval_as` asymmetry #844 was filed to fix.

### 5. False "unreachable" comment on `expand_func_calls`/`substitute_func_param` (`src/jq/eval.rs:28752`)
Their `Expr::TrackedVar(v) => Expr::TrackedVar(v.clone())` arms are commented "Never produced ... only by variable substitution during evaluation, which runs after this pass." Confirmed live that this is false: `. as $x | def f: $x | .b; path(f)` reaches this arm (harmlessly, still returns `["b"]`) since `expand_func_calls` runs at runtime on every `Expr::FuncDef` evaluation, which can already contain a `TrackedVar` from an enclosing `as`. Nested inside a loop, this arm re-clones the whole frozen document once per fan-out iteration — a cost the comment's framing hides from a future reader.

### 6. `is_identity_passthrough` reimplements `unwrap_paren` (`src/jq/eval.rs:16230`)
Hand-rolls its own recursive `Expr::Paren` unwrapping instead of calling the existing `unwrap_paren` helper (added for #845), which does the identical peeling generically and is documented for exactly this "caller only cares about literal shape" use case.

---
*Posted by Claude Code — 9-agent parallel review, several findings verified live against the pinned `jq` 1.7.1 oracle.*

newhoggy:
Addressed all 6 review findings (commits fe92465b8..146de8f37):

1. **Doc overclaim** — widened `is_identity_passthrough` to recognize `if/then/else` (both branches passthrough), `try/catch` (only the tried expr — provably can't raise), and `//` (only the left operand — right only runs when left is falsy, which this static check can't see). All three confirmed against jq 1.7.1, with negative controls for the shapes still correctly left untracked. Rewrote the doc comment to state precisely what's recognized instead of a blanket claim.

2. **Perf regression** — switched `Expr::TrackedVar`'s payload from `Box<OwnedValue>` to `Rc<OwnedValue>`. The cost was in `substitute_var_impl` re-cloning the whole snapshot on every loop-body rebuild, not primarily the equality check. Re-measured your exact benchmark shape (100k-element array, 300 loop iterations, inside `path()`): referencing the variable added ~2s before; now the with/without-reference times are within noise (~1.65-1.71s both ways).

3. **reduce/foreach untracked** — recorded in `docs/compliance/jq/limitations.md` per ADR-0018, with the mechanism explained (no `Expr::Reduce`/`Expr::Foreach` arm in `resolve_node`). Left as a documented, safe (never-wrong-path) gap rather than implemented — a real fix needs `resolve_node` to interleave path resolution with the fold loop itself, which is new machinery beyond this issue's scope.

4. **Duplicated gate** — extracted `substitute_bound_var`, now the single call site both `eval_as` and `resolve_node`'s `Expr::As` arm use.

5. **False "unreachable" comment** — fixed on both `expand_func_calls`/`substitute_func_param`'s `TrackedVar` arms; confirmed live that `def` inlining re-runs at evaluation time and can reach an already-substituted `TrackedVar`. The `Rc` switch also makes the clone these arms perform O(1).

6. **Reimplemented `unwrap_paren`** — `is_identity_passthrough` now calls the existing helper instead of hand-rolling `Paren` peeling.

10 new regression tests added, full suite green, both clippy legs clean.

PR #1480, opened 2026-08-22
fix(yq): anchored deferred sequence items use the '- ' width, not a full indent step

### LINKED ISSUE #1362: yq: an anchored deferred sequence item indents by a full step at non-default --indent, where yq uses the '- ' width
**Severity: Low** — cosmetic indentation only; the values round-trip identically and the output is valid, re-readable YAML. Found reviewing `#1354`.

## Summary

At a non-default `--indent`, a sequence item whose anchor sits on its own line (`- &x`) indents the value below it by a full indent step, where real yq indents it by the width of `- ` (2 columns). Only anchored deferred items are affected — the same shape without an anchor is already correct at every indent width, and every shape is correct at the default `-I=2` because the two rules coincide there.

## Repro

succinctly at `329021639`; oracle is pinned yq v4.53.3. Input:

```yaml
l:
  - &x
    p: 1
  - *x
```

```console
$ printf 'l:\n  - &x\n    p: 1\n  - *x\n' | yq -I=4 '.'
l:
    - &x
      p: 1
    - *x

$ printf 'l:\n  - &x\n    p: 1\n  - *x\n' | succinctly yq -I=4 '.'
l:
    - &x
        p: 1
    - *x
```

`p` sits at column 8 where yq puts it at column 6 — succinctly adds one full `indent_str` after the `- `, yq adds the two columns the `- ` itself occupies.

The DOM path agrees with the streaming path (this is `#763`'s DOM-equals-streaming rule working as intended), so a write shows the same thing:

```console
$ printf 'l:\n  - &x\n    p: 1\n  - *x\n' | succinctly yq -I=4 '.c = 1'
l:
    - &x
        p: 1
    - *x
c: 1
```

**Control — no anchor, same shape, already correct at `-I=4`:**

```console
$ printf 'l:\n  - p: 1\n    q: 2\n' | yq -I=4 '.c = 1'
l:
    - p: 1
      q: 2
c: 1

$ printf 'l:\n  - p: 1\n    q: 2\n' | succinctly yq -I=4 '.c = 1'
l:
    - p: 1
      q: 2
c: 1
```

**Control — the anchored shape is correct at the default `-I=2`:**

```console
$ printf 'l:\n  - &x\n    p: 1\n  - *x\n' | succinctly yq '.c = 1'
l:
  - &x
    p: 1
  - *x
c: 1
```

**Values are unaffected** — both sides parse to the same thing, which is what keeps this cosmetic:

```console
$ printf 'l:\n  - &x\n    p: 1\n  - *x\n' | succinctly yq -I=4 '.c = 1' | succinctly yq -o=json -I=0 '.'
{"l":[{"p":1},{"p":1}],"c":1}
$ printf 'l:\n  - &x\n    p: 1\n  - *x\n' | yq -I=4 '.c = 1' | yq -o=json -I=0 '.'
{"l":[{"p":1},{"p":1}],"c":1}
```

## Root cause

Two places, and they must be fixed together or the DOM/streaming agreement `#763` pins will break:

- `stream_yaml_value`'s sequence arm in `src/yaml/light.rs` — the streaming writer, where this originates and predates `#763`.
- `emit_yaml_value_at_depth`'s anchored-deferred branch in `src/bin/succinctly/yq_runner.rs` (added by `#763`), which computes `val_indent = format!("{indent}{}", config.indent_str)`. The non-anchored branch immediately below it uses `compact_indent = format!("{indent}  ")` — a literal two spaces — which is why only the anchored path diverges.

The same `indent_str`-vs-`"  "` distinction may also apply to where an anchored deferred item's own trailing comment is appended (`append_own_comment_line` is called with `val_indent` in that branch). Check it as part of the same fix rather than assuming it lands right.

## Tests

`tests/yq_cli_tests.rs`: the `-I=4` repro on both the streaming and DOM paths, plus both controls above (no-anchor at `-I=4`, anchored at `-I=2`) so a fix cannot trade one for the other. `#763`'s existing `test_yaml_dom_output_matches_streaming_output_over_an_anchor_corpus_763` should be extended with a non-default indent, since it is what would have caught this if it had varied `-I`.

## Related

- `#763` / `#1354` — mirrored the streaming rule into the DOM emitter, which is how this surfaced
- `#785` — the compact `- ` form the non-anchored branch implements correctly

newhoggy:
## Summary
- At a non-default `--indent` (e.g. `-I=4`), a block-sequence item whose anchor/tag sits on its own line (`- &x\n  ...`) indented its value by a full indent step instead of the 2-column width `- ` itself occupies, diverging from real yq (v4.53.3). Only the default `-I=2` happened to mask this, since the two widths coincide there.
- Fixed in both the streaming writer (`src/yaml/light.rs`, `stream_yaml_value`'s Sequence arm) and the DOM writer (`src/bin/succinctly/yq_runner.rs`, `emit_yaml_value_at_depth`'s anchored block-sequence branch) — these are separate, hand-mirrored implementations that happened to share this bug, not one shared function.
- The same branch also gates on an explicit tag with no anchor (`anchor.is_some() || tag.is_some()`), so the streaming-side fix covers that shape too for free.

## Not in scope (filing follow-ups)
While investigating, found the identical bug pattern in two more places that this issue's own repro doesn't cover, left unfixed here and tracked separately:
- The `--slurp` fast path (`stream_yaml_sequence`) has its own hand-mirrored copy of the same bug.
- An unrelated `-I=1` indent-clamping divergence from real yq (affects mappings and sequences alike, not anchor-specific).

Fixes #1362

## Test plan
- [x] Full `cargo test --features cli,simd,regex,serde` (2747+ tests, 0 failed)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [x] `cargo clippy --all-targets --features std,simd,serde,cli,regex,bench-runner,large-tests,mmap-tests -- -D warnings` clean
- [x] Live-verified against pinned oracle `yq` v4.53.3: streaming identity, DOM-forced write (`.c = 1`), non-anchored control (unaffected), default `-I=2` (unaffected), nested compounding anchors, and the pre-existing mapping-field deferred-anchor case (must not change) — all byte-match the oracle
- [x] New regression tests added: `src/yaml/light.rs` unit test at `-I=4`, `tests/yq_cli_tests.rs` streaming + DOM integration tests at `-I=4`
- [x] Patch coverage: 100% (13/13 new lines covered, `omni-dev coverage diff`)

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -2020,6 +2020,12 @@ fn emit_yaml_value_at_depth(
                     .map(|(i, v)| {
                         let elem_comments = comments.at_index(i);
                         if defers_to_own_block(v, elem_comments) {
+                            // Both arms below share this same 2-column
+                            // "compact" offset — the `- ` prefix's own
+                            // width, not a full `config.indent_str` step
+                            // (#785/#1362) — so it's computed once here
+                            // rather than separately in each arm.
+                            let compact_indent = format!("{indent}  ");
                             // An anchor written on the item's own line
                             // (`- &x\n  ...`) takes the slot the compact
                             // form would use, so the value stays deferred
@@ -2035,7 +2041,7 @@ fn emit_yaml_value_at_depth(
                                 // anchor on its own line still occupies that `- `
                                 // slot, so its value nests exactly as deep as an
                                 // unanchored element's would).
-                                let val_indent = format!("{indent}  ");
+                                let val_indent = compact_indent;
                                 let val = emit_yaml_value_at_depth(
                                     v,
                                     elem_comments,
@@ -2069,7 +2075,6 @@ fn emit_yaml_value_at_depth(
                             // effect `stream_yaml_value`'s cursor-based
                             // sibling gets for free from its per-field/
                             // per-element loop only indenting 2nd+ items.
-                            let compact_indent = format!("{indent}  ");
                             let rendered = emit_yaml_value_at_depth(
                                 v,
                                 elem_comments,
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -2028,7 +2028,14 @@ fn emit_yaml_value_at_depth(
                             // `light.rs`, which real yq is pinned against
                             // (#763).
                             if let Some(anchor) = elem_comments.declared_anchor() {
-                                let val_indent = format!("{indent}{}", config.indent_str);
+                                // Same "compact" rule as the plain block-sequence
+                                // element below: the value's own content aligns
+                                // under the `- ` prefix's 2-column width, not a
+                                // full `config.indent_str` step (#1362 -- an
+                                // anchor on its own line still occupies that `- `
+                                // slot, so its value nests exactly as deep as an
+                                // unanchored element's would).
+                                let val_indent = format!("{indent}  ");
                                 let val = emit_yaml_value_at_depth(
                                     v,
                                     elem_comments,
--- a/src/yaml/light.rs
+++ b/src/yaml/light.rs
@@ -1928,7 +1928,13 @@ impl<'a, W: AsRef<[u64]>> YamlCursor<'a, W> {
                                 out.write_char('-')?;
                                 write_deferred_prefix(out, None, anchor, tag)?;
                                 out.write_char('\n')?;
-                                let child_indent = deeper_yaml_indent(indent, indent_spaces, unit);
+                                // Same "compact" rule as the non-anchored `else`
+                                // arm below: the value's own content aligns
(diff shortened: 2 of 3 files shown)


newhoggy:
Review of this PR's fix surfaced three related, out-of-scope findings, filed separately:
- #1484 — the `--slurp` fast path (`stream_yaml_sequence`) has the same anchor/tag-deferred indent bug this PR fixes for the two paths it covers, in a third hand-mirrored implementation.
- #1485 — the underlying "compact" indent model doesn't compose correctly with deeper nested containers at non-default `-I` (pre-existing since #785, confirmed present in the non-anchored sibling branch too; this PR's own fix measurably *reduces* the divergence at every level rather than introducing or worsening it — see the issue for the column-by-column measurement).

Issue #1481, opened 2026-08-22
jq: the eager binary-fanout loops finish the right operand before the left, where jq interleaves them


newhoggy:
**Severity: Medium** — with `input`/`inputs` operands this changes *values*, not just
stderr ordering, so it is silent data loss rather than a cosmetic leak.

Found by #1459's (#820 Stage 4) own oracle sweep, and deliberately left out of that PR's
scope. Stage 4 made only `eval_each`'s new `Expr::Compare` arm demand-aware; every **eager**
copy of jq's right-outer/left-inner fanout loop still evaluates the right operand to
completion before starting the left.

## Repro

Verified against pinned jq 1.7.1 (`/usr/bin/jq`, jq-1.7.1-apple) and a `--release
--features cli` build at `dc09ff036`, stdout and stderr captured separately.

**Values change** — `input` pops from a process-global queue, so which document each
operand gets depends on the interleaving:

```
$ printf '"a" "b" "c" "d"' | jq            -cn '[(input) + (input, input)]'
["ba","dc"]
$ printf '"a" "b" "c" "d"' | succinctly jq -cn '[(input) + (input, input)]'
["ca","db"]
```

jq takes right's 1st output (`a`), then runs left (`b`) → `"ba"`; then right's 2nd (`c`),
then left again (`d`) → `"dc"`. We take right fully first (`a`,`b`), then run left per
right value (`c`,`d`) → `"ca"`,`"db"`.

**Evaluation-time I/O ordering changes**:

```
$ jq            -cn '[("A"|stderr) == (("B"|stderr), ("C"|stderr))]'   # stderr: B A C A
$ succinctly jq -cn '[("A"|stderr) == (("B"|stderr), ("C"|stderr))]'   # stderr: B C A A
```

```
$ jq            -cn '[error("x") == (1, ("B"|stderr))]'   # stderr: (just the error)
$ succinctly jq -cn '[error("x") == (1, ("B"|stderr))]'   # stderr: B, then the error
```

Both spellings are pinned as known divergences in
`test_short_circuit_side_effect_leaks_820_932_987` (`tests/jq_cli_tests.rs`).

## Scope — this is broader than #820

It affects **`Expr::Arithmetic` as much as `Expr::Compare`** (they share the loop), and it
lives in three places:

1. `binary_fanout_core` (`src/jq/eval.rs`), via `eval_binary_fanout` — the eager operand
   strategy it supplies to `binary_fanout_each`.
[...]

Issue #1421, opened 2026-08-22
jq: add the eval_each demand sink and rewire ten short-circuiting consumers (#820 Stage 2)


newhoggy:
**Stage 2** of [`docs/plan/jq-lazy-generator-consumers.md`](https://github.com/rust-works/succinctly/blob/main/docs/plan/jq-lazy-generator-consumers.md) (design merged in #1377). Depends on Stage 1 (#1386, merged).

Closes **#820** and the `any`/`all`/`IN(s)` half of **#932**.

## What

Add the demand-driven sink primitive to `src/jq/eval.rs`, alongside — not replacing — the
eager `eval_single`:

- `Demand { Continue, Stop }`, `Item<'a, W> { Borrowed, Owned }`,
  `Flow { Exhausted, Stopped { pending }, Escaped(Control) }`
- `eval_each`, `eval_each_pipe`, `eval_each_owned`, `drain_result`, and a `#[cfg(test)]`
  `collect_each`
- Native lazy arms for **`Expr::Comma`, `Expr::Pipe`, `Expr::Paren`** only; every other
  variant falls back to `eval_single` + `drain_result`

Then rewire ten consumers to install a sink: `builtin_isempty`, `eval_first_expr`,
`builtin_first_stream`, `eval_limit`, `builtin_limit`, `eval_nth_expr`,
`builtin_nth_stream`, `any_all_gen_cond`, `builtin_upper_in`, `builtin_halt_error`.

`eval_comma`, `eval_pipe`, `eval_single`, `QueryResult`, `Control` and `EvalEscape` keep
their current bodies and contracts verbatim.

## The invariant that makes this safe

> For a sink that always returns `Demand::Continue`, `eval_each` delivers exactly the values
> `push_owned_values(eval_single(...))` would collect, in the same order, with the same
> terminal `Control`.

So an un-lazified `Expr` variant is a **missed optimization, never a behaviour regression**.
Enforced by a `collect_each` differential test over the existing expression corpus, not just
asserted.

## Acceptance criteria

- The rows in `test_short_circuit_side_effect_leaks_820_932_987` that Stage 2 closes flip to
  jq's answers and **move into** `test_short_circuit_side_effect_shapes_already_match_jq_820`
  (that test's own failure message says to do this).
- `test_short_circuit_side_effect_shapes_already_match_jq_820`'s existing 20 rows stay green
  — this is the over-stopping guard and the whole reason Stage 1 went first.
[...]

Issue #822, opened 2026-08-12
jq/yq: eval_pipe_with_path_context_internal's Arithmetic/Compare arms still collapse cartesian fanout to first value (#768 gap)


newhoggy:
**Severity: Medium** — verified against the built CLI (release-equivalent debug build). Silent wrong answers (truncated output), reachable through the documented `--eval-all` flag. Pre-existing on `main`, not introduced by #768's fix.

## Summary

#768 fixed `eval_arithmetic`/`eval_compare` (`eval.rs`) and the native `Expr::Compare` fast-path arm in `eval_generic.rs`, both of which used to collapse a multi-output arithmetic/comparison operand to its first value instead of jq's cartesian-product fanout. There is a third, independent implementation of the same operator logic that #768's root-cause analysis didn't enumerate and its fix didn't touch: `eval_pipe_with_path_context_internal`'s own `Expr::Arithmetic`/`Expr::Compare` match arms (`eval.rs:12762` and `eval.rs:12825` on current `main`), which still evaluate both operands through the old `result_to_owned` "take the first output" policy.

This path is taken whenever an expression tree contains `key`/`parent`/`path`/`file_index` anywhere `needs_path_context` can see (`eval.rs:364`) — at that point `eval_pipe` (or `eval_owned_with_file_index`, used by `--eval-all`) routes the **entire** pipe through this second evaluator, including any `Expr::Arithmetic`/`Expr::Compare` nodes elsewhere in the same pipe, which then silently lose their fanout.

## Repro (built from this repo, `succinctly 0.8.0`)

`file_index` (the headline `--eval-all` feature, #715) is the easiest way to force the path-context route:

```
$ printf 'a: 1\n---\na: 2\n' > multi.yaml

# Fanout alone: correct, fixed by #768
$ succinctly yq --eval-all -o json '[.[] | (1,2,3) + 1]' multi.yaml
[2, 3, 4, 2, 3, 4]

# Same fanout, but sharing a pipe with file_index: collapses to first value
$ succinctly yq --eval-all -o json '[.[] | ((1,2,3) + 1), file_index]' multi.yaml
[2, 0, 2, 0]
# expected: [2, 3, 4, 0, 2, 3, 4, 1]

# Comparison variant, same collapse
$ succinctly yq --eval-all -o json '[.[] | ((1,2,3) > 1), file_index]' multi.yaml
[false, 0, false, 0]
# expected: [false, true, true, 0, false, true, true, 1]
```

[...]

PR #1482, opened 2026-08-22
fix(jq): IN(src; s) stops the compare fanout at the matching candidate (#1459)

### LINKED ISSUE #932: jq: any(gen;cond)/all(gen;cond)/IN/IN(src;s) leak halt_error/stderr's print side effect from a discarded generator tail

### LINKED ISSUE #1459: jq: IN(src; s) still leaks a discarded candidate's side effect — #932's remaining half (#820 Stage 4)
**Severity: Low** (stray stderr write; no wrong output, no data loss)

Follow-up to **#932**, which was closed as completed when #820's Stage 2 (#1421) landed. Stage 2
genuinely fixed the `any(gen; cond)`, `all(gen; cond)` and `IN(s)` thirds of that issue — but
#932's title also covers **`IN(src; s)`**, and that form still leaks. This issue carries the
remainder so the tracker stops disagreeing with the test suite.

## Repro

Verified at `main` @ `1541b3508` (0 commits behind `origin/main`) against pinned jq 1.7.1
(`/usr/bin/jq`), stdout and stderr captured separately:

```
$ jq -cn '[IN((2,3); 2, (5|stderr))]'
[true]                                       (nothing on stderr)
$ succinctly jq -cn '[IN((2,3); 2, (5|stderr))]'
[true]                                       stderr: 5
```

Same with `halt_error`, which is the other evaluation-time-I/O builtin:

```
$ jq -cn '[IN((2,3); 2, ("X"|halt_error(7)))]'              [true]   (no stderr)
$ succinctly jq -cn '[IN((2,3); 2, ("X"|halt_error(7)))]'   [true]   stderr: X
```

Control — the two-thirds Stage 2 *did* fix, for contrast:

```
$ echo 2 | jq -c 'IN(2, (5|stderr))'              true   (no stderr)
$ echo 2 | succinctly jq -c 'IN(2, (5|stderr))'   true   (no stderr)   <- agrees
```

## It is already pinned in the test suite

`test_short_circuit_side_effect_leaks_820_932_987` (`tests/jq_cli_tests.rs`) asserts today's
**wrong** output for this exact shape, with the comment:

> `IN(src; s)` synthesizes an `Expr::Compare` and hands it to `any_all_gen_cond` as `gen`, but
> `binary_fanout_core`'s outer loop is still eager, so the sink never sees the candidates one at
> a time. Closed by Stage 4. jq: no stderr.

So `main` currently ships a test that documents this as an open divergence while #932 is closed.
That is the mismatch this issue resolves.

## Root cause

`builtin_upper_in_src` (`src/jq/eval.rs:4269`) does not iterate candidates itself — it
synthesizes a comparison and delegates:

```rust
let gen = Expr::Compare { op: CompareOp::Eq, left: src.clone(), right: s.clone() };
any_all_gen_cond::<W, S>(&gen, &Expr::Identity, value, false, true)
```

`any_all_gen_cond` *is* lazy now (Stage 2 routed it through `eval_each_owned`), but its `gen`
here is an `Expr::Compare`, which has no lazy arm — so `eval_each` falls back to
`eval_single` + `drain_result`, and `binary_fanout_core` (`src/jq/eval.rs:2218`) has already run
the whole right operand to completion before the first item reaches the sink. The demand signal
arrives after the side effect.

This is the design's **Stage 4**
([`docs/plan/jq-lazy-generator-consumers.md`](https://github.com/rust-works/succinctly/blob/main/docs/plan/jq-lazy-generator-consumers.md)):
make `Expr::Compare`'s outer loop demand-aware. Call sites are `binary_fanout_core`,
`eval_binary_fanout` (`:2323`) and `eval_binary_fanout_with_path_context` (`:20543`).

## The trap case any fix must not break

`binary_fanout_core` evaluates the **right** operand as the outer loop and re-evaluates the left
per right value — jq's real nested-loop order, established and oracle-verified by #910. So a side
effect in the *first* candidate genuinely is reached, and both tools agree today:

```
$ jq -cn '[IN((2,3); (5|stderr), 2)]'              [true]   stderr: 5
$ succinctly jq -cn '[IN((2,3); (5|stderr), 2)]'   [true]   stderr: 5    <- must stay
```

Stopping too early here would be a new divergence in the opposite direction — the failure mode
#820's Stage 1 guard-rail test exists to catch. #910's own case must also keep holding:
`IN(4, error("boom"); 9, 4)` raises rather than answering `true`.

## Acceptance criteria

- `[IN((2,3); 2, (5|stderr))]` and the `halt_error` variant produce no stderr, matching jq.
- The `IN(src; s)` row moves out of `test_short_circuit_side_effect_leaks_820_932_987` into
  `test_short_circuit_side_effect_shapes_already_match_jq_820`.
- `[IN((2,3); (5|stderr), 2)]` still writes `5`; #910's nested-loop order is preserved.
- Its own oracle sweep over `Expr::Compare`, since the change touches every binary comparison
  with two multi-valued operands, not just `IN`.

Refs #932, #820.

newhoggy:
Closes #1459. Closes #932.

#820's **Stage 4**. #932 was closed when Stage 2 (#1421) landed, which was right for its
`any`/`all`/`IN(s)` thirds but left `IN(src; s)` divergent — and `main` shipped a test
pinning the *wrong* output for it, so the tracker disagreed with the suite.

```
$ jq            -cn '[IN((2,3); 2, (5|stderr))]'   # [true]   (no stderr)
$ succinctly jq -cn '[IN((2,3); 2, (5|stderr))]'   # [true]   stderr: 5   <- before
```

## Root cause

`builtin_upper_in_src` synthesizes an `Expr::Compare` and hands it to `any_all_gen_cond`
as `gen`. That has probed `cond` per output since Stage 2, but `eval_each` had no
`Expr::Compare` arm, so it fell back to `eval_single` + `drain_result` and the whole
comparison ran to completion before the first candidate ever reached the sink.

## Approach — one loop, two operand strategies

Not a hand-rolled candidate loop inside `builtin_upper_in_src`: #910's review already
rejected that shape as an unsound cross-product. Instead `binary_fanout_core`'s body
becomes **`binary_fanout_each`** — the same right-outer/left-inner loop, parameterized over
how an operand is *enumerated* rather than merely evaluated:

* `binary_fanout_core` is now the collecting wrapper supplying the **eager** strategy
  (evaluate the whole operand, replay it through `drain_result`), so `eval_arithmetic`,
  `eval_compare` and `eval_binary_fanout_with_path_context` are behaviourally untouched.
* `eval_each`'s new `Expr::Compare` arm supplies **`eval_each` itself**, so the right
  operand — jq's outer loop — stops producing candidates the sink never asked for.

Keeping one loop is the point: #768 fixed a collapse-to-first-value bug there that the
then-duplicated path-context copy independently re-acquired and had to be fixed again for
#822. A third copy would have been a third chance at the same drift.

`each_operand` is `Fn`, not `FnMut`, because it is called re-entrantly — the per-right-value
call on `left` happens while the call on `right` is still on the stack.

## Oracle sweep

105 shapes — all six `CompareOp`s × operand arities, every lazy consumer (`isempty`,
`first`, `limit`, `nth`, `last`, `any`, `all`, `IN(s)`, `IN(src; s)`, `path(paths(f))`,
`halt_error`'s argument), terminators in either operand, `?`/`try`, and `input` — run
against pinned jq 1.7.1 with **stdout, stderr and exit code captured separately**.
Divergences went **18 → 5**. yq mode spot-checked separately (`Expr::Compare` is
`S`-generic); output unchanged.

## Both directions are pinned

Moved into `test_short_circuit_side_effect_shapes_already_match_jq_820`, now matching jq:
`[IN((2,3); 2, (5|stderr))]`, the `halt_error` variant, the cursor-sourced
`[IN(.[]; 2, (5|stderr))]`, plus `isempty`/`limit(2)`/`nth(1)`/`all`/`any` over a bare
compare.

Added as **guard rails** in the same table, because over-stopping is this design's
characteristic failure mode — each is a shape where the demand genuinely *does* reach the
side effect and jq writes it:

| Guard rail | jq stderr |
|---|---|
| `[isempty((1,2) == ("B"\|stderr, 3))]` | `B` |
| `[limit(3; (1,2) == (10, ("B"\|stderr)))]` | `B` |
| `[any((1,2) == (10, ("B"\|stderr)); .)]` | `B` |
| `[IN((2,3); (5\|stderr), 2)]` | `5` |
| `[IN(4, error("boom"); 9, 4)]` | raises (#910's order) |

`eval_each_with_a_never_stopping_sink_matches_eval_single_820` gains 20 side-effect-free
compare cases — that differential is what stops the new arm drifting from `eval_compare`.

## Two residuals, left as separate work

* A bare `first(<non-comma>)` never reaches `eval.rs`: `eval_generic`'s native `first`/`last`
  arm short-circuits only over *comma* siblings. Same class as the already-pinned
  `first(.[] | stderr)` — option (c).
* The **eager** copies of the loop still finish the right operand before the left, where jq
  interleaves them. Broader than #820 (it affects `Expr::Arithmetic` too, and lives in
  `eval_generic`'s own third copy of the loop), and with `input` operands it changes
  *values*: `(input) + (input, input)` over `"a" "b" "c" "d"` is `["ba","dc"]` in jq and
  `["ca","db"]` here. Filed as **#1481**, and pinned meanwhile.

`drain_result`'s "a lazy arm never reorders or alters the values delivered" invariant gains
an explicit, documented carve-out for this arm rather than a silent exception.

## Verification

`cargo fmt --check`; all three `ci.yml` clippy legs (`--all-features`, and both explicit
feature sets) with `-D warnings`; `cargo check --no-default-features` (no_std);
`cargo test` and `cargo test --features cli` (5660 passed, 0 failed); plus the
`--features simd`, `--features portable-popcount`, `SUCCINCTLY_SIMD=sse2` and
`--features simd,scalar-yaml` legs. All run sequentially.

Interleaved A/B on arithmetic-heavy queries (9 reps each, per-rep interleaved) shows no
regression — 0.89–0.93× of baseline, since the eager wrapper no longer allocates the two
intermediate `Vec`s the old body built per fanout.

### DIFF
--- a/docs/plan/jq-lazy-generator-consumers.md
+++ b/docs/plan/jq-lazy-generator-consumers.md
@@ -32,7 +32,7 @@ as such in `test_short_circuit_side_effect_leaks_820_932_987`: `first(.[] | stde
 `first((1,2) == (10, ("B"|stderr)))` (both a bare `first(...)` over a non-`Comma`, which
 Stage 2b's sibling walk does not reach — option (c)), and
 `("A"|stderr) == (("B"|stderr), ("C"|stderr))` (a *top-level* compare, which never reaches
-`eval_each` at all — see follow-up 6). Stage 3 also leaves one residual: an un-lazified
+`eval_each` at all — #1481). Stage 3 also leaves one residual: an un-lazified
 `eval_each` arm (`If`/`Try`/`Label`/`AsPattern`/`FuncCall`, ...) still leaks a side effect
 for a node's own filter when that filter's shape isn't one of
 `Comma`/`Pipe`/`Paren`/`Builtin::PathsFilter` — pinned as a known-remaining row in the same
@@ -1048,7 +1048,8 @@ the reasoning behind each placement:
    `Expr::Compare` arm, which is what a top-level `==` actually reaches. With
    `input`/`inputs` operands it changes *values*, not just stderr ordering:
    `(input) + (input, input)` over `"a" "b" "c" "d"` is `["ba","dc"]` in jq and
-   `["ca","db"]` here, so it is silent data loss rather than a cosmetic leak.
+   `["ca","db"]` here, so it is silent data loss rather than a cosmetic leak. **Filed as
+   #1481.**
 7. **Stage 5** — widen the lazy arm set (`If`, `Try`/`Optional`, `Label`, `AsPattern`,
    `FuncCall`, and demand-forwarding through `FirstExpr`/`Limit`/`NthExpr`). Not yet filed.
    One issue, not one per arm: each is a single lazy arm now that the primitive exists.
--- a/docs/plan/jq-lazy-generator-consumers.md
+++ b/docs/plan/jq-lazy-generator-consumers.md
@@ -27,16 +27,17 @@ by Stage 3: `each_paths_filter` + `resolve_leaf`'s stop-after-first sink). See
 a discarded branch consuming `input`/`inputs` documents the CLI's driver loop then never
 processed — is closed, as is `halt_error`'s wrong exit code and the stderr leak in
 `isempty`/`first`/`limit`/`nth`/`any`/`IN(s)`, and — since Stage 4 — `IN(src; s)` and every
-other compare reached through a lazy consumer. Three shapes remain divergent and are pinned
-as such in `test_short_circuit_side_effect_leaks_820_932_987`: `first(.[] | stderr)` and
-`first((1,2) == (10, ("B"|stderr)))` (both a bare `first(...)` over a non-`Comma`, which
-Stage 2b's sibling walk does not reach — option (c)), and
-`("A"|stderr) == (("B"|stderr), ("C"|stderr))` (a *top-level* compare, which never reaches
-`eval_each` at all — #1481). Stage 3 also leaves one residual: an un-lazified
+other compare reached through a lazy consumer. Five shapes remain divergent, all pinned as
+such in `test_short_circuit_side_effect_leaks_820_932_987`. Two are the bare-`first(...)`
+family: `first(.[] | stderr)` and `first((1,2) == (10, ("B"|stderr)))`, both a `first(...)`
+over a non-`Comma`, which Stage 2b's sibling walk does not reach — option (c). One is
+`("A"|stderr) == (("B"|stderr), ("C"|stderr))`, a *top-level* compare, which never reaches
+`eval_each` at all — #1481. The remaining two are Stage 5's own residual: an un-lazified
 `eval_each` arm (`If`/`Try`/`Label`/`AsPattern`/`FuncCall`, ...) still leaks a side effect
 for a node's own filter when that filter's shape isn't one of
-`Comma`/`Pipe`/`Paren`/`Builtin::PathsFilter` — pinned as a known-remaining row in the same
-test, left to Stage 5.
+`Comma`/`Pipe`/`Paren`/`Builtin::PathsFilter` — that is the `path(paths(if ...))` row, plus
+the `path(... halt_error ...)` pair Stage 4 added, whose stray `x` on stderr is the same
+un-lazified `Expr::If`.
 
 **One correction this document earned the hard way.** Stage 2 shipped an
 `eval_each_pipe` whose `Item::Owned` arm fell back to the eager `eval_owned_pipe`,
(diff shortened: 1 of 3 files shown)


newhoggy:
## Review

**Verdict: correct, well-scoped, ship it.** Extracting `binary_fanout_each` as *one* loop
parameterized over how an operand is enumerated — rather than a third copy — is the right
call given the #768 → #822 drift history. I checked `binary_fanout_core`'s new body arm by
arm against the old one (combine-error/`optional` → `finish_fork`, inner escape → `partial`,
trailing outer control → `partial`) and it is equivalent.

### Evidence

| Check | Result |
|-------|--------|
| All four acceptance criteria vs pinned jq 1.7.1 (`/usr/bin/jq`) | pass |
| 17,160-shape compare sweep (6 ops x 13 left x 12 right operand shapes x 20 consumers), `main` vs branch | **0 value changes** |
| Same sweep, branch vs jq (stdout + stderr + exit, compared separately) | **0 divergences** |
| 399-shape side-effect matrix vs jq | `main` 138 diffs -> branch **18**; newly-diverging set **empty** |
| The 18 residuals | all `first(<non-comma>)` bypass or top-level interleaving (#1481) — both already documented |
| `cargo test`, `--features cli,regex` CLI legs, both clippy legs, `fmt --check`, `cargo doc`, `--no-default-features` | green |
| Interleaved A/B, fanout-heavy arithmetic, 5 MB, 5x5 reps | parity (0.42–0.43 s both) |
| yq mode, 16 compare shapes | no change |
| Deep nesting | parser depth limit hits first on both — no new stack risk |
| `combine`-error paths, `optional` and not | live and jq-matching, unchanged |

### Three things review added (pushed as two follow-up commits)

**1. The change fixes silent data loss, not just a stderr leak — and that half was
untested.** Once a compare's operands can pop `input`, the eager fanout's discarded
candidates eat documents the CLI's driver loop then never processes. `main` prints two of
four documents where jq prints four; the branch matches jq. Details and the severity
correction are in a comment on #1459. Now pinned in
`test_compare_operand_consuming_input_documents_1459`, deliberately separate from the stderr
tables, because every row there would stay green if the arm regressed to eager only for
document-popping operands.

**2. `binary_fanout_each` drops the outer operand's `Flow::Stopped { pending }`.**
`abort.unwrap_or(outer)` keeps the *inner* (left) operand's pending and discards the *outer*
(right) one's. `resolve_leaf` is the only consumer that reads `pending`, so the effect is
confined to a `Halt` under `path(...)` — and it contradicted `Flow::Stopped`'s own doc
comment, which said such a halt "must never be downgraded into a catchable path error":

```
path(1 == (if true then (1, ("x"|halt_error(3))) else empty end))
   jq       exit 5, the path error          (nothing on stderr)
   main     exit 3, the halt wins           x
   branch   exit 5, the path error          x
path((if true then (1, ("x"|halt_error(3))) else empty end) == 1)
   jq       exit 5, the path error
   main     exit 3                          x
   branch   exit 3                          x    <- inner keeps `pending`
```

The drop is the half that lands on jq's verdict (jq's `path()` is lazy end to end and never
evaluates the `halt_error` at all), and I found no shape where it moves *away* from jq — so
I kept the behaviour and documented it instead: `binary_fanout_each` now explains the choice
with the oracle evidence, `Flow::Stopped` records it as its one stated exception, and both
rows are pinned in `test_short_circuit_side_effect_leaks_820_932_987`. The stray `x` on both
is Stage 5's un-lazified `Expr::If`, not this loop's.

**3. The interleaving claim had no test.** `drain_result`'s new "one arm carries a stated
exception" caveat rests on the lazy arm producing jq's interleaved order, and nothing
asserted it. It is true — `[all(("A"|stderr) == (("B"|stderr), ("C"|stderr)); true)]` writes
`BACA` on jq and on this branch, `BCAA` on `main` — and it is the only row that catches a
silent regression back to the eager loop while every stderr-suppression row stays green.
Pinned in the "already match" table, distinct from the #1481 row for the top-level spelling.

Plus a doc fix: the status paragraph said "Three shapes remain divergent" where the leaks
test pins five.

Rebased onto `main` (was 32 behind); full local verification green on the rebased tree.

PR #1488, opened 2026-08-22
perf(jq): use BTreeSet for omit()'s key/index membership tests (#1383)

### LINKED ISSUE #1383: perf(jq): omit() scans its key list per field, O(k*n) -- fourth instance of the linear-scan-dedup class
**Severity: Low-Medium** — found sweeping for siblings of #1301's bug after fixing it.

## Summary

`builtin_omit` (`src/jq/eval.rs`) tests membership against a `Vec` once per
document field, so `omit(keys)` is O(k·n) in the omit-list length `k` and the
container size `n`. Both arms have it:

```rust
// object arm
let omit_keys: Vec<&str> = keys.iter().filter_map(...).collect();
for field in *fields {
    ...
    if !omit_keys.contains(&key) {      // <- O(k) per field
```

```rust
// array arm
let omit_indices: Vec<usize> = keys.iter().filter_map(...).collect();
let result: Vec<OwnedValue> = arr.iter().enumerate().filter_map(|(i, v)| {
    if omit_indices.contains(&i) {      // <- O(k) per element
```

## The stated justification is wrong

Both sites carry the same comment:

```rust
// Use a Vec for no_std compatibility (typical omit lists are small)
```

`alloc` has `BTreeSet`. This is settled in the same file, about 5,000 lines
away, by `c2e331e72` — which fixed this exact class in `delete_keys` and
recorded the reasoning verbatim:

> `BTreeSet` rather than `HashSet` because the crate is `no_std` and `alloc`
> has no hasher.

So `no_std` is a reason not to reach for `HashSet`, never a reason to accept a
linear scan.

## Measured

Release build, current `main` + #1379. Document held at 16,000 fields, omit-list
length `k` varied — the axis that isolates the scan from parse/output cost:

| k | time | kept |
|---|------|------|
| 1 | 30ms | 15,999 |
| 100 | 29ms | 15,900 |
| 1,000 | 33ms | 15,000 |
| 4,000 | 76ms | 12,000 |
| 8,000 | 111ms | 8,000 |

Above a ~29ms floor the scan cost tracks `k` (4ms → 47ms → 82ms). Note the
*kept* count falls as `k` rises, so output work shrinks while total time grows —
the scan is more dominant than the raw totals suggest.

Not quadratic in a single quantity the way #1301 was, so the impact is milder:
it needs a large omit list *and* a large container. `omit` over half a
100,000-key object is where it would hurt.

## Fix

The same shape #1379 used four times and `c2e331e72` used once: a `BTreeSet` for
the object arm's `&str` keys and the array arm's indices. Neither arm needs
insertion order (unlike #1379's grouping, where source order decides which error
jq reports), so a plain set is enough — no `IndexSet` required.

Delete the misleading comment rather than editing it; the fourth repetition of
this misconception in one file is itself the finding.

## Related

Fourth occurrence of this bug class in `src/jq/eval.rs`:
`shift_remove`-per-key (`c2e331e72`, 4.4s → 0.05s) → `delete_keys`' array arm's
`Vec::remove` note → #1301's four grouping scans (#1379) → this. Three of the
four justified the slow structure with a `no_std` argument that does not hold.

newhoggy:
## Summary

`builtin_omit` (`src/jq/eval.rs`) tested membership against a `Vec` once per
document field/element, making `omit(keys)` O(k·n) in the omit-list length `k`
and the container size `n`. Both arms carried the same justification:

```rust
// Use a Vec for no_std compatibility (typical omit lists are small)
```

`alloc` has `BTreeSet`. This exact reasoning was already settled in this file by
`c2e331e72` (`delete_keys`): `no_std` rules out `HashSet` (no hasher), never a
set at all. Fourth occurrence of this bug class in `src/jq/eval.rs`
(`shift_remove`-per-key → `delete_keys`'s array arm → #1301's four grouping
scans (#1379) → this).

## Fix

Same shape as the three prior fixes: swap the `Vec` for a `BTreeSet` in both
the object arm (`&str` keys) and the array arm (`usize` indices). Neither arm
needs insertion order, so a plain set is enough. Deleted the misleading
comment rather than editing it.

## Measured

Release build, current `main`. Document held at 16,000 fields, omit-list
length `k = 8,000` (the issue's own worst-case row):

| | before | after |
|---|---|---|
| time | 111ms | 23ms |

## Test plan

- [x] All 8 existing `omit` unit tests pass unchanged (pure data-structure
      swap, no behavior change — object/array/negative-index/out-of-bounds/
      all-keys/order-preservation/nonexistent-keys/empty-keys all still pass)
- [x] `cargo fmt --check`
- [x] Manual correctness spot-check against the release binary
      (`omit(["a","c"])`, `omit([1,-1])` with negative index)
- [x] Manual timing check confirming the measured improvement above

No dedicated new test/benchmark added, matching the precedent set by
`c2e331e72` (the same bug class, same fix shape, no test file changed) — this
is a pure algorithmic-complexity fix with no observable behavior change, fully
covered by the existing correctness test suite.

Fixes #1383

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -29609,8 +29609,7 @@ fn builtin_omit<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     match &value {
         StandardJson::Object(fields) => {
             // For objects, keep all keys except those in the omit list
-            // Use a Vec for no_std compatibility (typical omit lists are small)
-            let omit_keys: Vec<&str> = keys
+            let omit_keys: BTreeSet<&str> = keys
                 .iter()
                 .filter_map(|k| {
                     if let OwnedValue::String(s) = k {
@@ -29639,8 +29638,7 @@ fn builtin_omit<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
             let arr: Vec<_> = (*elements).collect();
             let len = arr.len() as i64;
 
-            // Use a Vec for no_std compatibility (typical omit lists are small)
-            let omit_indices: Vec<usize> = keys
+            let omit_indices: BTreeSet<usize> = keys
                 .iter()
                 .filter_map(|k| {
                     let idx = match k {
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -29636,26 +29636,18 @@ fn builtin_omit<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         StandardJson::Array(elements) => {
             // For arrays, keep all indices except those in the omit list
             let arr: Vec<_> = (*elements).collect();
-            let len = arr.len() as i64;
 
+            // `resolve_read_index`, not a hand-rolled match+cast (#1383's own
+            // review): the two did the same negative-index/out-of-bounds
+            // resolution, but diverged on NaN -- the old inline `as i64`
+            // silently saturated `nan` to index 0 (Rust's documented
+            // saturating-cast behavior), so `omit([nan])` removed element 0
+            // instead of skipping the unresolvable key, unlike this array's
+            // own sibling `delete_keys` (`delpaths`), which already used
+            // `resolve_read_index` and correctly treats `nan` as no match.
             let omit_indices: BTreeSet<usize> = keys
                 .iter()
-                .filter_map(|k| {
-                    let idx = match k {
-                        OwnedValue::Int(i) => *i,
-                        OwnedValue::Float(f) => *f as i64,
-                        OwnedValue::NumberLiteral(NumberRepr::Int(i), _) => *i,
-                        OwnedValue::NumberLiteral(NumberRepr::Float(f), _) => *f as i64,
-                        _ => return None,
-                    };
-                    // Handle negative indices
-                    let actual_idx = if idx < 0 { len + idx } else { idx };
-                    if actual_idx >= 0 && actual_idx < len {
-                        Some(actual_idx as usize)
-                    } else {
-                        None // Out of bounds indices are ignored
-                    }
-                })
+                .filter_map(|k| resolve_read_index(k, arr.len()))
                 .collect();
 
             let result: Vec<OwnedValue> = arr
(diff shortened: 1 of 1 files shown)


newhoggy:
## Code review response

5 finder agents reviewed this PR. Summary of what came back and how each was handled:

**Fixed:** The array arm's hand-rolled index resolution (negative-index/out-of-bounds
math via a manual `match` + `as i64` cast) diverged from the existing, already-tested
`resolve_read_index` helper on NaN handling — `*f as i64` saturates NaN to 0 (Rust's
documented cast behavior), so `omit([nan])` silently removed index 0 instead of skipping
an unresolvable key, unlike this array's own sibling `delete_keys`/`delpaths`, which
already routes through `resolve_read_index` and correctly treats NaN as no match.
Live-verified before/after; fixed by reusing the existing helper (also removes the
duplicated index-resolution logic in the same edit). New regression test:
`test_omit_array_nan_index_ignored_1383`.

**Investigated, no regression found:** one finder theorized `BTreeSet` might carry a
small-N constant-factor cost the old `Vec` avoided for "typical" (small) omit lists. Ran
a real interleaved A/B measurement (not assumed) — pre-fix `Vec` binary vs this PR's
`BTreeSet` binary, `omit(["k0"])` (k=1) against a 1.3MB/80,000-field input, 20 reps each:
68.21ms vs 68.20ms median. No measurable difference, consistent with `BTreeSet`
degenerating to effectively one tree node at this size.

**Not actioned (out of scope):** a suggestion to replace the array arm's `BTreeSet` with
a `Vec<bool>` bitmap sized by the array length, for an even lower constant factor than
`BTreeSet` gives. Real optimization idea, but it would diverge from the established,
uniform `Vec`→`BTreeSet` shape this issue and its three prior siblings (`c2e331e72`,
#1379) deliberately share for consistency — leaving as a possible future micro-opt, not
blocking this fix.

**Not actioned (broader than this issue):** a suggestion to extract one shared helper
for "collect a filtered owned-key/index set from an array of keys," since `delete_keys`
and `omit`'s two arms each hand-write a similar collection step. The object arm's `&str`
keys and the array arm's `usize` indices aren't the same shape, so a single generic
helper isn't a trivial lift — left as a possible follow-up rather than expanding this
PR's scope beyond the stated Vec→BTreeSet complexity fix.

Issue #1146, opened 2026-08-17
yq: @base64d error messages don't match real yq's wording; jq-mode lossy-UTF-8 path untested


newhoggy:
**Severity: Low**

## Summary

Found during code review of a duplicate fix for #1123 (this exact issue's fix landed
as PR #1137; this is a follow-up on the code as it now stands on \`main\`). Three
small, related gaps in \`format_base64d\`/\`format_urid\` (\`src/jq/eval.rs\`):

**1. Error message wording doesn't match real yq's:**

\`\`\`
\$ printf '"aGVs bG8="\n' | succinctly yq -o=json '@base64d'
Error: invalid base64
\$ printf '"aGVs bG8="\n' | yq -o=json '@base64d'          # real yq
Error: illegal base64 data at input byte 4
\`\`\`

A new test added by #1137's own fix (\`test_base64d_rejects_embedded_non_ascii_whitespace_1123\`
or similarly named) has a doc comment claiming the error \"match[es] real yq exactly
(\`illegal base64 data at input byte N\`)\", but the test only asserts a non-zero exit
code, so the message-text claim itself is never actually checked and is currently
false.

**2. \`base64_trailing_byte\`'s error message embeds the untrimmed original input,**
even when yq-mode's edge-trim changed what was actually decoded:

\`\`\`
\$ printf '"  false  "\n' | succinctly yq '@base64d'
Error: string ("  false  ") trailing base64 byte found
\$ printf '"  false  "\n' | yq '@base64d'                  # real yq
Error: illegal base64 data at input byte 5
\`\`\`

succinctly's message shows the original string with its leading/trailing spaces
still in it, even though the decoder actually operated on the trimmed \`\"false\"\` --
confusing when debugging exactly what was decoded. Wording also doesn't match real
yq's uniform \"illegal base64 data at input byte N\" format at all (pre-existing,
unrelated to #1123/#1137's trim fix).

**3. No regression test for jq-mode's lossy-UTF-8-substitution path** in either
\`format_urid\` or \`format_base64d\` -- only the yq-mode variant is covered. A future
change that accidentally made the shared lossy-decode helper strict only in jq mode
would pass the full suite silently:

\`\`\`
\$ echo '"%FF"' | succinctly jq '@urid'
"�"                                    # currently correct (lossy), but untested
\`\`\`

## Suggested fix direction

- Either change succinctly's base64-decode error wording to match real yq's \"illegal
[...]

PR #1490, opened 2026-08-22
fix(yq): validate @base64d padding placement in-place, not truncate-first (#1135)

### LINKED ISSUE #1135: yq: @base64d's malformed-padding handling diverges from real yq (jq-only algorithm, not S-generic)
**Severity: Low**

## Summary

Found while implementing #1120 (deliberately out of that fix's scope —
real jq and real yq genuinely disagree on this specific edge case, and
`format_base64d` isn't generic over `<S: EvalSemantics>` to express a
per-mode difference).

#1120 fixed `@base64d` to match real jq's algorithm exactly: truncate the
input at the first `=` character (discarding it and everything after),
then decode whatever real base64 characters remain, erroring only on a
1-character trailing remainder. This matches real yq too for the
*non-padded* cases #1120 was about (`"false"`, `"ab"`, `"abc"`, ...) and
for standard padding (`"QQ=="`/`"QQ="` both give `"A"` in both).

But real yq is **stricter** than real jq for malformed/excess padding:

```
$ printf 'a: "===="\n' | yq -o=json '.a | @base64d'
Error: illegal base64 data at input byte 0

$ echo '"===="' | jq -c '@base64d'
""
```

succinctly's `@base64d` (post-#1120) now matches jq's lenient behavior
(`""`) for this case, diverging from yq. This is **not a regression**
from #1120 — the pre-#1120 code was *also* wrong here, just differently
wrong (it produced a spurious NUL byte instead of either oracle's actual
answer).

## Scope note

Fixing this fully would mean:
- Making `format_base64d` generic over `<S: EvalSemantics>` (unlike its
  sibling encoder `format_base64::<S>`, it currently isn't).
- Reverse-engineering yq's exact Go `encoding/base64`-based decoder rules
  for malformed padding, which appear stricter than jq's own hand-rolled
  decoder in ways not yet fully characterized (only the pure-padding case
  above has been live-verified so far).

Not yet investigated further — filing to make the gap trackable, not
blocking.

newhoggy:
## Summary

Real yq's `@base64d` (Go's `encoding/base64.StdEncoding` model) is strict about
malformed/misplaced padding where real jq's `@base64d` is lenient:

```
$ printf 'a: "====\"\n' | yq -o=json '.a | @base64d'
Error: illegal base64 data at input byte 0

$ echo '"===="' | jq -c '@base64d'
""
```

jq truncates at the first `=` and decodes whatever real base64 characters remain
(post-#1120); yq instead validates padding placement within the *final*
4-character quantum and rejects anything else — a leading `=`, padding in a
non-final quantum, or real content after a complete padded quantum.

## Changes

`format_base64d` gains a dedicated yq-mode decode path implementing this
validation directly — not a variant of jq's truncate-then-decode loop, which
stays byte-for-byte unmodified (per the issue's own plan: it's already
oracle-hardened by #1120/#1146 and must not move). All existing jq-mode and
yq-mode tests pass unchanged.

## Correction to the issue's own original plan

Rule #1 as originally stated ("CR and LF are skipped anywhere") turned out, on
closer live re-verification before starting, to be too broad — see the
[claim comment](#) for the full derivation. CR/LF are only tolerated as a
*leading and/or trailing run*; a newline strictly between real content — even
exactly at a 4-character quantum boundary — is rejected, not skipped:

```
$ printf '"Q\nQ=="\n'     | yq -o=json '. | @base64d'   # LF splits a quantum
Error: unexpected EOF
$ printf '"QUJD\nQUJD"\n' | yq -o=json '. | @base64d'   # LF between 2 complete quanta
Error: unexpected EOF
```

This turned out to already be **fully handled by the existing code**: the
yq-mode `.trim()` step (added by #1123) already strips leading/trailing
whitespace — including CR/LF — before any base64-specific logic runs. No new
CR/LF-skip machinery was needed; a `\n` that survives trimming (i.e.,
genuinely embedded) simply fails as an invalid character, same as any other
non-alphabet byte.

## Byte-offset accuracy

Every accept/reject case is live-verified against pinned yq v4.53.3, including
exact byte offsets where practical. The "too-short trailing remainder" case
needed correcting mid-implementation — real yq's position is *one past* the
lone leftover character, not *at* it (confirmed against both `"a"` → byte 1
and `"false"` → byte 5, and against an existing test,
`test_yq_base64d_illegal_data_on_trailing_remainder_1146`, that pinned the old
truncate-based code's coincidentally-correct byte 5 for `"false"`).

Two cases from the issue's own matrix (padding immediately followed by an
invalid character, e.g. `"ab=!"`) match the reject *verdict* but not the exact
byte offset — accepted per the issue's own explicit guidance not to chase
Go's streaming-decoder offset accounting; replicating those internals exactly
would need reverse-engineering `encoding/base64`'s own read-loop behavior for
no behavioral benefit.

## Not added (considered, deferred)

`tests/data/yq-golden/` fixtures (via `scripts/sync-yq-golden.sh`) — that
system asserts stderr byte-for-byte against a live-captured oracle run, which
this change's new CLI tests already cover directly and more cheaply. Left as
a possible follow-up if broader golden coverage for `@base64d` errors is
wanted later.

## Test plan

- [x] New `tests/yq_cli_tests.rs` tests (suffixed `_1135`): full accept-side
      matrix, full reject-side matrix (with exact byte offsets), leading/
      trailing-vs-embedded newline behavior, a `?`-guarded catchable-error
      check, a multi-quantum-with-trailing-padding-only case, and a jq-mode
      regression guard confirming jq stays lenient on the same inputs
- [x] All pre-existing `#1146`/`#1123`/`#1120`/`#1109` base64d tests pass
      unchanged (jq mode byte-for-byte untouched)
- [x] `cargo fmt --check`
- [x] Manual live verification of the full issue matrix plus additional
      edge cases against both the release binary and the pinned yq v4.53.3
      oracle

Fixes #1135

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -472,6 +472,7 @@ follows each one in its own mode. The behavioural axis is `EvalSemantics`
 |---|---|---|
 | Bare 2-arg `sub(re; s)` | first match only (`"aaa"` → `"Xaa"`) | **every** match (`"aaa"` → `"XXX"`) |
 | `@uri`/`@base64`/`@html` on a container | JSON-encodes first (`[1,2]` → `"%5B1%2C2%5D"`) | errors, as real yq does |
+| `@base64d` on malformed/misplaced padding | truncates at the first `=`, decodes the rest (`"===="` → `""`) | validates padding placement in place, rejects anything else (`"===="` → error) |
 | `keys` | sorted | document order (yq's `keys` *is* `keys_unsorted`) |
 | Integer overflow | converts to float | wraps |
 | Division by zero | errors | infinity |
@@ -486,12 +487,12 @@ follows each one in its own mode. The behavioural axis is `EvalSemantics`
 
 Ten of these rows are `EvalSemantics` constants, each carrying its live-verification note in
 the trait's doc comments — `7 + null` / `null - 7` is one row over two constants,
-`ADD_RIGHT_NULL_REQUIRES_CONCAT_TYPE` and `SUB_LEFT_NULL_IS_IDENTITY`. **Three are not.**
-Bare `sub` and container `@uri`/`@base64` are `S::TAG == EvalTag::Yq` tests at their call
-sites in [src/jq/eval.rs](../../../src/jq/eval.rs), and `keys` is rewritten to
-`KeysUnsorted` at parse time under `ParserMode::Yq`
-([src/jq/parser.rs](../../../src/jq/parser.rs)). More than forty `S::TAG` sites exist in
-`src/` in total.
+`ADD_RIGHT_NULL_REQUIRES_CONCAT_TYPE` and `SUB_LEFT_NULL_IS_IDENTITY`. **Four are not.**
+Bare `sub`, container `@uri`/`@base64`, and `@base64d`'s malformed-padding strictness are all
+`S::TAG == EvalTag::Yq` tests at their call sites in
+[src/jq/eval.rs](../../../src/jq/eval.rs), and `keys` is rewritten to `KeysUnsorted` at parse
+time under `ParserMode::Yq` ([src/jq/parser.rs](../../../src/jq/parser.rs)). More than forty
+`S::TAG` sites exist in `src/` in total.
 
 That split is the debt ADR-0018 rule 2 is aimed at, not a counterexample to it: a constant is
 discoverable from the trait definition, whereas a call-site test is discoverable only by
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -7032,7 +7032,7 @@ fn format_base64<S: EvalSemantics>(
 /// rather than needing its own early return. jq mode is unaffected: it
 /// keeps erroring on every non-string type, unchanged from before.
 ///
-/// The decode algorithm matches real jq's exactly (see the
+/// jq mode's decode algorithm matches real jq's exactly (see the
 /// truncate-at-first-`=` comment inside), and validates every character of
 /// every trailing group (2, 3, or 4 bytes long) the same way, so embedded
 /// whitespace anywhere before the first `=` is caught regardless of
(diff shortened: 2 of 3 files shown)


newhoggy:
## Code review response

Multiple finder agents converged on the same real bug (and I ran my own live
verification as part of confirming it): the first version's yq-mode decoder inferred
"is this padding in the final quantum?" from a fixed 4-byte *input-chunk* boundary. Real
yq has no such window — it's a genuine single left-to-right scan that only tracks how
many real characters have accumulated since the last quantum closed. This gave the
correct accept/reject *verdict* in every case (no data corruption), but the wrong byte
offset whenever a well-formed padded quantum was followed by more data (`"ab==cd"`:
reported byte 2, real yq reports byte 4 — `"ab=="` closes cleanly, and it's the
*following* `"cd"` that's actually illegal), or when an invalid character preceded a
misplaced padding character in the same window (`"A!=A"`: reported byte 3, real yq
reports byte 1, the position of the `!` itself).

**Fixed** by rewriting the decoder as a true single left-to-right pass (tracking
real-character count in the current quantum + a "stream already closed by padding"
flag), replacing the chunk-based version entirely. Differential-fuzzed against the
pinned yq v4.53.3 oracle: 3280 exhaustive cases over an `"A"`/`"="`/`"!"` alphabet
(every padding-position permutation up to length 7) plus 1000 additional randomized
cases spanning the full base64 alphabet, `=`, `!`, CR, LF, and space — zero verdict
mismatches, zero decoded-byte mismatches, and the only remaining byte-offset mismatches
are (a) whitespace-adjacent positions, already covered by this issue's own explicit
"don't chase Go's internal streaming-decoder accounting" scope note, and (b) a
pre-existing, unrelated UTF-8-lossy-substitution granularity difference in
`owned_string_from_decoded_bytes` shared by both jq and yq modes, predating this issue
entirely.

**Also fixed:** the shipped test that had baked in the buggy first version's wrong byte
offset for `"ab==cd"` (now corrected, plus several of the newly-characterized cases
added as regressions), a stale docstring above `format_base64d` still claiming this
divergence was "not addressed here," and the missing `docs/compliance/yq/limitations.md`
row this repo's documentation-of-record rule requires for every mode divergence.

**Not actioned:** the two Vec-allocation efficiency findings (`chunks: Vec<&[u8]>` and
per-chunk `vals: Vec<u8>`) — both are moot now, since the redesign eliminates the chunk
collection entirely (a true index-based scan, no `.chunks()` call) and reuses a single
`Vec` via `.clear()` across the whole decode rather than allocating one per quantum. The
"two independent decode loops in one function" structural note stands as accepted,
matching the PR's own explicit justification (jq's path is oracle-hardened and must not
move) — same as before.

Issue #1494, opened 2026-08-22
yq: finish threading EvalTag through the remaining EvalError value-preview constructors (#1055 follow-up)


newhoggy:
**Severity: Low**

## Summary

Follow-up to #1055, which built the mode-aware value-preview dispatch (`dump_truncated_with`/`describe_with`, plus `_in`-suffixed constructor siblings) and migrated it fully for `EvalError::from_value` (1 call site) and `EvalError::cannot_iterate` at the 19 of 29 call sites where `S: EvalSemantics` was already directly in scope. This issue tracks what #1055 deliberately left on the jq-pinned shim, per its own staged design ("migrate call sites family by family... only as far as needed; if a chain reaches something genuinely mode-agnostic, stop and leave that family on the shim, with a comment naming which one and why").

### `cannot_iterate`'s 10 threading-needed call sites (`src/jq/eval.rs`)

`S: EvalSemantics` is not directly in scope at these; each needs it threaded up its own call chain first (or a documented reason it can't/shouldn't be):

- `builtin_any` (line ~4597), `builtin_all` (~4617) — likely straightforward, both are thin wrappers
- `builtin_flatten` (~5712)
- `builtin_from_entries` (~6170)
- `set_path` (~13266), `set_path_through_iterate` (~13779)
- `delete_expr_iterate_paths` (~24992, ~25011 — 2 call sites)
- `delete_at_path` (~25349, ~25545 — 2 call sites)

(Line numbers are as of #1055's merge; re-check before starting, since `main` moves quickly in this repo.)

### Other preview-embedding constructors not touched by #1055 at all

#1055's own investigation named these as needing the same treatment; only `cannot_iterate` and `from_value` were migrated:

- `cannot_use_as_object_key` (5 call sites per #1055's own count)
- Every other constructor in `src/jq/error.rs` that calls `describe(`/`dump_truncated(` internally (11 total `describe(` + 12 total `dump_truncated(` call sites inside `error.rs` itself, per #1055's re-measurement — `cannot_iterate`/`cannot_use_as_object_key` account for some but not all of them)

## Suggested approach

Same shape as #1055: for each remaining constructor, add an `_in(tag, ...)` sibling (or confirm one already exists if it embeds a preview via `describe_with`/`dump_truncated_with` already), migrate every call site where `S` is already in scope, and either thread `S` further for the rest or leave them shimmed with a comment naming why. `EvalError::cannot_iterate` itself is now a good worked example for the mechanical steps.

[...]

PR #1495, opened 2026-08-22
fix(yq): thread EvalTag through EvalError value previews for from_value/cannot_iterate (#1055)

### LINKED ISSUE #1055: yq: EvalError's value previews (describe/dump_truncated) always format via jq rules, never yq's verbatim echo

newhoggy:
## Summary

Every `EvalError` constructor that embeds a value preview (`describe`/`dump_truncated`
in `src/jq/error.rs`, and `EvalError::from_value` backing the `error(v)` builtin) routed
through `stream_owned_value_json_jq` unconditionally, so a `NumberLiteral` embedded in a
yq-mode error message was always reformatted via jq's rules instead of echoed verbatim —
inconsistent with every other yq-mode output path (`tostring`, `@json`, ...):

```
$ echo 'a: [1e2, x]' | succinctly yq '.a | error'
Error: [1E+2,"x"]      # was; now [1e2,"x"], matching .a | tostring
```

## Design

Per the issue's own implementation plan: thread a runtime `EvalTag` rather than making
`EvalError` generic over `S` (which would infect every signature returning one).
`dump_truncated_with`/`describe_with` dispatch to `stream_owned_value_json_jq` for `Jq`
and the existing yq real-output convention, `stream_owned_value_json` (now `pub(crate)` —
#1008's own split for the output path, no new formatter needed here, only a dispatch),
for `Yq`. The old names become jq-pinned shims delegating with `EvalTag::Jq`, so every
existing call site keeps compiling unchanged.

## Scope

Migrated fully:
- `EvalError::from_value` — 1 production call site, directly fixing the demonstrated repro
- `EvalError::cannot_iterate` — 19 of 29 call sites, where `S: EvalSemantics` was already
  in scope (confirmed by inspection before migrating, not assumed)

Left on the jq-pinned shim, per the plan's own staging ("migrate call sites family by
family... only as far as needed... stop and leave that family on the shim"):
- `cannot_iterate`'s remaining 10 call sites, which need `S` threaded further up their
  own call chains first
- Every other preview-embedding constructor in `error.rs` (`cannot_use_as_object_key`
  and others) — not touched at all in this PR

Both tracked as a follow-up: #1494.

## Verification

jq mode is byte-for-byte unchanged throughout — every commit of this migration passed
the existing jq-error-message parity gate (`scripts/sync-jq-error-messages.sh` + the jq
golden suite). New tests assert the internal-consistency property the issue itself
specifies as the acceptance criterion — a value's error-preview spelling must match
`tostring`'s own spelling for the same value — rather than a hard-coded expected string,
since external-oracle comparison is largely inconclusive for these paths (real yq either
doesn't error there or errors with unrelated wording).

## Test plan

- [x] New yq CLI tests (suffixed `_1055`): `error(v)`'s preview vs `tostring`, table of
      `NumberLiteral` spellings including exponents and an overflowed literal
      (`1e400`); `cannot_iterate`'s preview vs `tostring`, same table; a jq-mode
      regression guard confirming jq keeps reformatting, not echoing
- [x] `cargo test --features cli,simd,regex,serde --test jq_error_message_tests` (the
      jq-mode safety gate) — passes unchanged
- [x] `cargo fmt --check`
- [x] Manual live verification of both the `error(v)` and `cannot_iterate` repros
      against the release binary, in both jq and yq mode

Fixes #1055

### DIFF
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -203,16 +203,29 @@ fn dump_truncated_with(tag: EvalTag, value: &OwnedValue) -> String {
     // The sink stops the writer once the dump is known to exceed the budget;
     // writing into a `String` cannot fail for any other reason, so the returned
     // `Result` carries nothing `sink.overflowed` has not already recorded.
-    let _ = match tag {
-        EvalTag::Jq => stream_owned_value_json_jq(value, &mut sink),
-        EvalTag::Yq => stream_owned_value_json(value, &mut sink, 0, 0, ' ', false),
-    };
+    let _ = stream_value_preview(tag, value, &mut sink);
     if !sink.overflowed {
         return sink.buf;
     }
     sink.truncate_to(DUMP_KEEP);
     sink.buf.push_str("...");
     sink.buf
+}
+
+/// The one dispatch every `EvalTag`-aware preview call site shares (#1055):
+/// jq's error-message convention, or yq's real-output one reused as-is (see
+/// [`dump_truncated_with`]'s doc comment for why no bespoke yq formatter is
+/// needed). Factored out so [`dump_truncated_with`] and
+/// [`EvalError::from_value_with`] can't drift apart on it.
+fn stream_value_preview<W: core::fmt::Write>(
+    tag: EvalTag,
+    value: &OwnedValue,
+    out: &mut W,
+) -> core::fmt::Result {
+    match tag {
+        EvalTag::Jq => stream_owned_value_json_jq(value, out),
+        EvalTag::Yq => stream_owned_value_json(value, out, 0, 0, ' ', false),
+    }
 }
 
 /// A [`core::fmt::Write`] sink that keeps the first `cap` bytes written to it and
@@ -420,24 +433,21 @@ impl EvalError {
     /// same as real jq.
     ///
     /// jq-pinned shim (#1055): see [`dump_truncated`]'s doc comment; use
-    /// [`EvalError::from_value_in`] once `S: EvalSemantics` is in scope at the
+    /// [`EvalError::from_value_with`] once `S: EvalSemantics` is in scope at the
     /// call site.
     pub fn from_value(value: OwnedValue) -> Self {
-        Self::from_value_in(EvalTag::Jq, value)
+        Self::from_value_with(EvalTag::Jq, value)
     }
 
     /// Mode-aware sibling of [`EvalError::from_value`] (#1055): yq mode
     /// echoes a `NumberLiteral` verbatim here too, matching `.a | tostring`
     /// on the same value instead of reformatting it via jq's rules.
-    pub fn from_value_in(tag: EvalTag, value: OwnedValue) -> Self {
+    pub fn from_value_with(tag: EvalTag, value: OwnedValue) -> Self {
         let message = match &value {
             OwnedValue::String(s) => s.clone(),
             other => {
                 let mut message = String::new();
-                let _ = match tag {
-                    EvalTag::Jq => stream_owned_value_json_jq(other, &mut message),
-                    EvalTag::Yq => stream_owned_value_json(other, &mut message, 0, 0, ' ', false),
-                };
+                let _ = stream_value_preview(tag, other, &mut message);
                 message
             }
         };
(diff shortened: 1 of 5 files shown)


newhoggy:
## Code review response

An unusually large fan-out of finder agents converged on a consistent picture. Summary of
what was found and how each was handled:

**Fixed:**
- `delete_at_path`'s 2 remaining `cannot_iterate` call sites — it already carries a
  `yq_mode: bool` parameter (derived from `S::TAG` at its own call site), so this was a
  trivial bool→`EvalTag` conversion, not new threading. Live-verified:
  `del(.a[])` on a scalar `NumberLiteral` now echoes verbatim in yq mode.
- Extracted the duplicated `match tag { Jq => .., Yq => .. }` dispatch block (hand-copied
  in both `dump_truncated_with` and `EvalError::from_value_with`) into one shared
  `stream_value_preview` helper.
- Standardized the `_in`/`_with` naming split (`from_value_in`/`cannot_iterate_in` →
  `_with`, matching `dump_truncated_with`/`describe_with`) — neither spelling had prior
  codebase precedent, so picked one.
- Removed a `"1e400"` test case that looked like it exercised the overflow-to-infinite
  path but didn't — this codebase's own scalar resolver falls back to a plain `string`
  type once an exponent overflows `f64`, before any `NumberLiteral`/`Float` is produced.

**Confirmed real, but out of this PR's scope — documented, pinned, deferred to #1494:**
- A non-finite `Float` (`.nan`/`.inf`) in a yq-mode error preview still renders as JSON's
  `null` substitution, not yq's own spelling, and a *computed* whole float still gains a
  spurious forced `.0` — both because the `EvalTag::Yq` dispatch reuses
  `stream_owned_value_json` (the `-o json`/round-trip-safe real-output convention) rather
  than `tostring`'s own formatters (`numeric_display_string`/`nonfinite_display_string`).
  A real fix needs either a bespoke preview formatter or making
  `stream_owned_value_json_with_at_depth`'s currently-hardcoded NaN handling pluggable
  too (unlike its sibling float/infinite-literal formatters, NaN isn't a function
  parameter at all — a genuine structural blocker, not a quick swap). Pinned via two new
  `_known_gap_1495` tests documenting current (imperfect but stable) behavior, and the
  narrowed test-file scope comment now says so explicitly instead of overclaiming
  universal `tostring` parity.
- `builtin_any`/`builtin_all`/`builtin_flatten`/`builtin_from_entries`/`set_path`/
  `set_path_through_iterate`'s `cannot_iterate` call sites, plus `delete_expr_iterate_paths`'s
  2 sites — all genuinely lack `S`/an equivalent mode flag in scope (confirmed by
  inspection, not the `delete_at_path` shape), needing real signature changes. Already
  named in #1494.
- `cannot_use_as_object_key` and the arithmetic `binary_op` family (`arith_add`/`arith_sub`/
  etc.) — a later sweep found these also have `S` trivially in scope but were never
  touched by this PR at all (only `cannot_iterate`/`from_value` were migrated). Added to
  #1494 with live repros, since it was already within that issue's stated "every other
  constructor" scope, just not previously enumerated concretely.

Updated #1494 with all of the above so a future pass has the full, concrete list rather
than needing to rediscover it.

Issue #1486, opened 2026-08-22
yq: -I=1 clamps mapping/sequence nesting to 2 columns in real yq, succinctly honors the literal 1


newhoggy:
**Severity: Low** — cosmetic indentation only, at an unusual indent width.
Found investigating #1362 (unrelated to anchors -- reproduces on plain,
non-anchored content too).

## Summary

At `-I=1`, real yq clamps every nesting step to 2 columns rather than
literally honoring the requested 1-space width. succinctly's streaming
writer uses the literal requested width throughout.

## Repro (confirmed live against pinned yq v4.53.3)

```
$ printf 'l:\n  m:\n    p: 1\n' | yq -I=1 '.'
l:
  m:
    p: 1
$ printf 'l:\n  m:\n    p: 1\n' | succinctly yq -I=1 '.'
l:
 m:
  p: 1
```

Real yq's output at `-I=1` is byte-identical to its own `-I=2` output for
this input -- it appears to floor the effective step at 2 regardless of
what `-I` requests below that.

## Why this is separate from #1362 / #1485

Unrelated to anchors, tags, or the "compact" sequence-item rendering those
issues cover -- this is a plain nested-mapping case with no sequence
involved at all, and reproduces identically whether or not any anchor is
present.

## Suggested fix direction

Not investigated. Likely a missing `max(requested_width, 2)`-style clamp
wherever `-I`'s raw value is turned into `indent_spaces`/`IndentSpec` --
worth checking whether the clamp is on the *step size* or specifically on
the *rendered result*, and whether it's an oracle quirk worth matching
verbatim (ADR-0018) or arguably a real-yq bug not worth reproducing
bug-for-bug (its own effect is silently ignoring part of the user's
requested indent width).

newhoggy:
Moving to `Ready` — landed in `Todo` via the board's known new-issue auto-add bug, not a deliberate hold; this is freshly-filed, immediately workable follow-up work with no unresolved dependency.

newhoggy:
🔧 Claimed by agent `Cecil` — branch `issue-1486-yq-indent-1-clamp`

PR #1497, opened 2026-08-22
fix(yq): resolve -o=auto against the input's own format (#1493)

### LINKED ISSUE #1493: yq: -o=auto always resolves to JSON output, never matches YAML input format
**Severity: Medium** — silently produces the wrong output format for a
documented CLI flag's stated behavior, not just a formatting nuance.

## Summary

Found while reviewing #1486's `-I=1` clamp fix. Real yq's `-o=auto` means
"match the input format" — YAML input renders as YAML, JSON input renders
as JSON. succinctly's `-o=auto` currently renders as **JSON regardless of
input format**.

`output_value`'s own dispatch (`src/bin/succinctly/yq_runner.rs`) is:

```rust
if config.output_format == OutputFormat::Yaml {
    // YAML branch
} else {
    // JSON branch
}
```

`OutputFormat::Auto` is never resolved to a concrete format anywhere in
this file before reaching this check (confirmed via grep — every
`output_format ==` comparison in the file checks `Yaml` or `Json`
explicitly; `Auto` is never matched positively) — so it always falls
through to the `else` (JSON) branch, unconditionally.

## Repro (confirmed live against pinned yq v4.53.3)

```
$ printf 'l:\n  m:\n    p: 1\n' | yq '.' -o=auto
l:
  m:
    p: 1

$ printf 'l:\n  m:\n    p: 1\n' | succinctly yq '.' -o=auto
{
  "l": {
    "m": {
      "p": 1
    }
  }
}
```

## Why this is separate from #1486

#1486 fixed a formatting-width bug (`-I=1`'s indent) whose fix had to be
careful to track this exact `Auto`-falls-through-to-JSON dispatch so its
own clamp didn't newly diverge between `-o=auto` and `-o=json`. That fix
correctly mirrors the current (wrong) dispatch rather than fixing it —
fixing the dispatch itself is a separate, larger change (affects every
`-o=auto` invocation, not just indentation) and wasn't investigated
further here.

## Suggested fix direction

Not investigated. Likely needs `Auto` resolved to a concrete `OutputFormat`
per-input (matching each source's own `InputFormat`, mirroring how
`json_sourced_floats` is already threaded per-file for a related
input/output-format-mixing concern) before `output_value`'s dispatch runs,
rather than being read as a literal enum comparison that only ever matches
`Yaml`/`Json` explicitly.

newhoggy:
## Summary

`output_value`'s dispatch (`if config.output_format == Yaml {...} else {...JSON...}`)
never resolved `OutputFormat::Auto` to a concrete format anywhere, so `-o=auto` always
fell into the JSON else-branch regardless of input format, contradicting real yq's own
"match the input format" semantics:

```
$ printf 'l:\n  m:\n    p: 1\n' | succinctly yq '.' -o=auto
{                                    # was; now: l:\n  m:\n    p: 1
  "l": { "m": { "p": 1 } }
}
```

## Design

Adds `OutputConfig::for_source(args, input_format)`, mirroring the existing
`json_sourced_floats` per-source override pattern (#978/#1398): resolves `Auto` against
the concrete `InputFormat` and recomputes `indent_str`'s `-I=1` YAML clamp (#1486) to
match, since that clamp only applies when `output_value` will actually take the YAML
branch — which `Auto` may or may not, depending on what it resolves to.

Wired into every place that previously either cloned `OutputConfig` ad hoc (the default
and `--split-exp` per-source loops) or read `output_format` directly with no per-source
resolution at all:
- `need_comments` (JSON output never reads `CommentTree`)
- the multi-doc `---` separator gate
- `--inplace`'s per-file config
- the `--front-matter=process` compat guard, which now **accepts** `-o=auto` (front-matter
  content is always forced to `InputFormat::Yaml`, so `Auto` can no longer resolve to
  anything the guard was written to reject)

`--null-input`/`--raw-input` have no source to resolve against; matches real yq's own
confirmed-live default (Yaml, with a backwards-compat warning even under `-p=json`).
`--slurp` resolves against every source's uniform format if they agree (confirmed live
for both all-YAML and all-JSON slurps), falling back to Yaml for a genuinely
mixed-format slurp — real yq's own mixed-format `--slurp -o=auto` renders each element
in its own source format, a much deeper per-element behavior not attempted here (pinned
as a known gap in the test suite).

## A related, previously-unreachable bug this change exposes (also fixed)

The "standard" (non-M2-fast-path) multi-file loop wrote a `---` separator before
*every* document, including the first, whenever `is_multi_doc` — correct only
*between* documents everywhere else in this file. This never manifested before
because any invocation that would reach this loop with `output_format == Yaml` and
more than one document always took the M2 fast path instead (which already tracks this
correctly via its own `yaml_doc_streamed` flag), and `-o=auto` could never satisfy that
fast path's exact `Yaml`/`Json` check pre-#1493. Fixed with the same "has anything
already been written" tracking the M2 path and `SplitDocState` already use.

## Existing tests updated (encoded the old, buggy behavior)

- `test_indent_1_auto_format_matches_explicit_json_1486` — asserted `-o=auto ==
  -o=json`; now asserts `-o=auto == -o=yaml`, since YAML input now correctly resolves
  to YAML.
- `test_front_matter_process_rejects_auto_output` — asserted `-o=auto` was rejected
  under `--front-matter=process`; now asserts it's accepted (renamed to
  `test_front_matter_process_accepts_auto_output_1493`).

## Test plan

- [x] New yq CLI tests (suffixed `_1493`): the issue's own repro (YAML→YAML), the JSON
      case, per-file resolution across independent files (same-format, clean; and a
      "known gap" pin for mixed-format), `--null-input` defaulting to Yaml, `--slurp`
      with uniform format (both all-YAML and all-JSON) and a "known gap" pin for mixed
      format, and `--inplace`
- [x] `cargo test --features cli,simd,regex,serde` (full suite, `--no-fail-fast`,
      `--test-threads=4` — two unrelated tests were flaky under full default
      parallelism on this shared machine, confirmed spurious by re-running in
      isolation; unrelated to this change)
- [x] `cargo fmt --check`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` (matches CI's
      Documentation job)
- [x] Manual live verification of every case above against both the release binary and
      the pinned yq v4.53.3 oracle

Fixes #1493

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -439,6 +439,39 @@ $ printf 'a: [1, 2]\n' | yq            -o=json '.a[].b = error("boom")'   # {"a"
 $ printf 'a: [1, 2]\n' | succinctly yq -o=json '.a[].b = error("boom")'   # Error: boom
 ```
 
+### `-o=auto` on a genuinely mixed-format multi-source run doesn't match per-element
+
+[#1493](https://github.com/rust-works/succinctly/issues/1493) made `-o=auto` resolve
+against the input's own format instead of always rendering JSON — YAML input renders as
+YAML, JSON input as JSON, per-source (`--split-exp`, the standard multi-file path,
+`--eval-all`, `--inplace`) or per-invocation-uniform-format (`--slurp`/`--eval-all` when
+every source agrees). A genuinely *mixed*-format run is the one case left unresolved:
+real yq treats the whole output as one YAML stream with each JSON-sourced document
+embedded flow-style, `---` between every document regardless of source format —
+succinctly instead gives each document its own correct format independently, with no
+separator between differently-formatted documents and no flow-style JSON embedding:
+
+```bash
+$ printf 'a: 1\n' > a.yaml && printf '{"b":2}' > b.json
+
+$ yq            -o=auto '.' a.yaml b.json   # a: 1
+                                             # ---
+                                             # {"b": 2}
+$ succinctly yq -o=auto '.' a.yaml b.json   # a: 1
+                                             # {
+                                             #   "b": 2
+                                             # }
+
+$ yq            -o=auto --slurp '.' a.yaml b.json   # - a: 1
+                                                      # - {"b": 2}
+$ succinctly yq -o=auto --slurp '.' a.yaml b.json    # - a: 1
+                                                       # - b: 2
+```
+
+Live-verified against yq v4.53.3. Pinned as known gaps (current, not the desired end
+state) by `test_yq_auto_output_mixed_format_multi_file_known_gap_1493` and
+`test_yq_auto_output_slurp_mixed_format_known_gap_1493` in `tests/yq_cli_tests.rs`.
+
 ### Other categories
 
 Float and number formatting ([#1071](https://github.com/rust-works/succinctly/issues/1071),
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -117,6 +117,29 @@ pub struct EvalContext {
     pub named: IndexMap<String, OwnedValue>,
 }
 
+/// Resolve `OutputFormat::Auto` against a concrete `InputFormat` (#1493):
+/// real yq's `-o=auto` matches the input's own format -- YAML input stays
+/// YAML, JSON input becomes JSON. Anything already concrete passes through
+/// unchanged. `InputFormat::Auto` shouldn't reach here in practice
+/// (`resolve_input_format`/`detect_format_from_path` already resolve it
+/// before any source reaches this point), but it's handled the same as
+/// `InputFormat::Yaml` (Yaml is the safer default of the two either way).
+/// Real yq's own "nothing to match" case (`--null-input`, `--raw-input`,
+/// mixed-format `--slurp`) also defaults to Yaml -- confirmed live: `yq -n
+/// -p=json '{}' -o=auto` still prints YAML (with a backwards-compatibility
+/// warning), not JSON, so callers with no per-source `InputFormat` of their
+/// own should resolve against `InputFormat::Yaml` here too, not skip the
+/// call.
+fn resolve_output_format(output_format: OutputFormat, input_format: InputFormat) -> OutputFormat {
+    if output_format != OutputFormat::Auto {
+        return output_format;
+    }
+    match input_format {
+        InputFormat::Json => OutputFormat::Json,
+        InputFormat::Yaml | InputFormat::Auto => OutputFormat::Yaml,
+    }
+}
+
 /// Output configuration
 #[derive(Clone)]
 struct OutputConfig {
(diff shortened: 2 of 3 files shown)


newhoggy:
## Code review response

An unusually large, thorough review fan-out converged strongly on one dominant finding,
plus several smaller ones. Summary:

**Fixed — critical, confirmed by 6+ independent finder angles with live oracle
verification:**
- `--eval-all` was the one output-producing branch entirely missed by the original fix
  — it never called `for_source` at all, so `-o=auto` under `--eval-all` still always
  rendered JSON. Fixed with the same "uniform format across every source, fallback to
  Yaml" treatment `--slurp` already got, since `--eval-all` combines sources the same
  way.

**Fixed — a real bug the eval-all investigation surfaced, in the same mechanism:**
- The standard multi-file loop's `any_doc_output` flag flipped `true` for *any*
  document regardless of that document's own resolved format, making the
  leading-separator decision file-*order*-dependent for a mixed-format run (a YAML
  document right after a JSON one wrongly got a leading `---`; the same document first
  did not). Renamed to `any_yaml_doc_output`, now only updated for YAML-resolved
  documents — both orders agree now.
- `--inplace`'s own DOM/slow-path branch had the identical "leading separator before
  the first document" bug the standard path was fixed for, just never reachable before
  this PR (`-o=auto` never satisfied `output_format == Yaml` pre-fix). Fixed the same
  way. Two pre-existing tests had baked in the old, buggy leading-separator behavior as
  their expected output — updated both, using the standard/stdout path's own output for
  the same filter+input as the reference where no live oracle exists for succinctly's
  own `halt`/`if`/`then`/`end` control flow (real yq doesn't parse that syntax at all).

**Fixed — cheap, low-risk:**
- `for_source`'s unnecessary `..self.clone()` (cloned the old `indent_str` String only
  to immediately discard it) — now writes the remaining `Copy` bool fields explicitly.
- Added the two missing `docs/compliance/yq/limitations.md` entries for the
  mixed-format-multi-source known gaps the tests already pinned but didn't record, per
  this repo's divergence-recording rule.

**Confirmed real, but genuinely pre-existing and out of scope — filed as #1498:**
- `--eval-all` also never applies `json_sourced_floats`, even with `for_source` now
  wired in. Confirmed live on unmodified `main` *before* this PR too, with an *explicit*
  `-o=json` (no `Auto` involved) — a separate, deeper gap in how `--eval-all`'s older
  `parse_input` path represents a JSON-sourced value, not something `for_source`'s
  `-o=auto` fix was ever positioned to close.

**Not actioned (architectural refactor suggestions, correctly identified but out of
this PR's scope):**
- Two independent surfaces computing the resolved format (`resolve_output_format`
  directly vs. via `for_source`) — a real observation, but unifying them would be a
  larger refactor than this fix warrants.
- `for_source` requiring `&args` at every call site, `SplitDocState`/`any_yaml_doc_output`
  tracking two similar-but-distinct "has output started" concepts — both legitimate
  simplification ideas, left for a future cleanup pass rather than expanding this PR
  further.

All new/updated tests pass; full suite green (`--test-threads=4` — two unrelated tests
were flaky under full parallelism on this shared machine both before and after this
round, confirmed spurious by isolated re-runs).

PR #1503, opened 2026-08-22
fix(jq): widen first_over_comma_generic through Pipe/Paren

### LINKED ISSUE #1451: jq: first_over_comma_generic only unwraps a literal top-level Comma, missing Pipe-wrapped and nested-sibling commas
## Summary

Found during review of #820's merged Stage 2/2b work (PRs #1424, #1435, and
the follow-up #1450). `eval_generic.rs`'s `first_over_comma_generic` — the
native fast-path `first(f)`/`last(f)` take for the CLI's own evaluator —
only recognizes laziness when `f`, after `unwrap_paren`, is a *literal
top-level* `Expr::Comma`:

```rust
fn first_over_comma_generic<S: EvalSemantics, V: DocumentValue>(
    inner: &Expr,
    ...
) -> Option<GenericResult<V>> {
    let Expr::Comma(exprs) = unwrap_paren(inner) else {
        return None;
    };
    ...
}
```

Two shapes fall through this check and hit the old eager fallback, so a
side-effecting sibling still fires even though real jq's lazy generator
semantics never reach it:

**1. The comma is reached through a `Pipe`, not directly:**

```
$ jq -cn 'first(1 | (1, ("B"|stderr)))'
1
$ succinctly jq -cn 'first(1 | (1, ("B"|stderr)))'
B1
```

(This is exactly the shape #1450 just fixed in `eval.rs`'s
`eval_each_pipe` — but `first`/`last` route through `eval_generic.rs`'s own
native arm, per the design doc's own note that "the CLI never enters
`eval.rs` first" for these two builtins, so that fix does not reach here.)

**2. The comma is nested inside one sibling of `first`'s own argument list,
not the argument itself:**

```
$ jq -cn 'first((1, ("B"|stderr)), 2)'
1
$ succinctly jq -cn 'first((1, ("B"|stderr)), 2)'
B1
```

## Scope

This is the same class of gap `docs/plan/jq-lazy-generator-consumers.md`
already names and defers to **Stage 5** ("widen the arm set") for `If`/
`Try`/`Label`/`AsPattern`/`FuncCall` wrapping a comma — this issue is the
`Pipe`-wrapping and nested-sibling instances of that same gap, specific to
`first_over_comma_generic`'s own literal-`Comma` check. Not a regression
from #1450; confirmed present on `main` before and after that PR.

## Suggested fix

Widen `first_over_comma_generic`'s recognition beyond a literal top-level
`Comma`, likely by giving `eval_generic.rs` its own lazy "eval each, stop on
first" primitive analogous to `eval.rs`'s `eval_each`/`eval_each_pipe`
(added by #820 Stage 2), rather than pattern-matching a fixed AST shape.
Given `eval_generic.rs` and `eval.rs`'s owned-pipe path are two independent
evaluators that have each needed their own copy of past reindex-bridge
fixes, whatever lands here should be checked against both.

newhoggy:
## Summary
- `first_over_comma_generic` (`eval_generic.rs`) is the CLI's own native fast-path for `first`/`last` -- `Expr::FirstExpr`/`Builtin::FirstStream` have a native arm that intercepts before the subtree ever crosses into `eval.rs`'s own already-lazy `first`/`last`. It only recognized laziness when its argument, after `unwrap_paren`, was a literal top-level `Expr::Comma` -- a comma reached through a `Pipe` stage (`first(1 | (1, stderr))`), or nested inside one sibling of an outer comma (`first((1, stderr), 2)`), fell through to the eager fallback, firing side-effecting/input-consuming siblings a real generator consumer would never reach.
- Widened it to recurse through `Pipe`'s own prefix stages: the prefix is evaluated exactly once via the ordinary eager `eval_single`, then every possible shape of that one result is dispatched without ever redoing it. Since a pipe's leading stage is usually a *computed* value (`GenericResult::Owned`/`ManyOwned` -- e.g. a bare literal), not one derived from the document, a mutually-recursive `OwnedValue`-flavored twin (`first_over_comma_owned_generic`) handles that case, since there's no cheap `OwnedValue -> V` conversion to reuse the `V`-flavored path directly.
- As a consequence of trying each of a `Many`/`ManyCursor` prefix's already-materialized values against the tail in turn (stopping at the first that succeeds), this also closes `first(.[] | stderr)` -- a harder case #820 Stage 2b explicitly could not close and had left pinned as a known-still-wrong divergence. Moved that row into the "already matches jq" test and updated `docs/plan/jq-lazy-generator-consumers.md`'s status table accordingly.
- `needs_path_context` is checked across the whole pipe up front, matching the existing eager `Expr::Pipe` arm's own guard, so `path`/`parent`/`key` never silently see a stubbed-zero default.

**Note on scope**: this closes option (c) from the design doc, but only for `first`/`last` -- the only `eval_generic.rs` consumers with a native arm shadowing `eval.rs`'s lazy path (`isempty`/`limit`/`nth`/`any`/`all`/`IN` have no native arm there and already bridge into `eval.rs`).

Fixes #1451

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo llvm-cov --features cli,simd,regex,serde --workspace --fail-under-lines 0` -- iteratively closed coverage gaps for every reachable `GenericResult` dispatch shape (3+/2-stage pipes, `ManyOwned`, `Partial`→error, comma exhaustion/fallthrough at both the document-derived and computed-value layers); remaining uncovered lines are the bare cursor-less `One`/`Many` variants (verified hard to construct from real jq syntax -- the only non-pattern-match construction sites found are an internal `select` multi-truthy sub-case and an unreachable-via-parser empty-`Pipe` special case) and the `Partial`→`Break`/`Halt` control-conversion arms (symmetric in shape to the tested `Partial`→`Error` arm; a live probe confirmed `break` inside a comma prefix doesn't reach this exact arm, so it's untested but low-risk by construction)
- [x] Live-verified every case against the pinned `jq` oracle (`/usr/bin/jq` 1.7.1), including multi-stage pipes, error/break propagation, `input`-consumption, and yq mode

### DIFF
--- a/docs/plan/jq-lazy-generator-consumers.md
+++ b/docs/plan/jq-lazy-generator-consumers.md
@@ -29,8 +29,9 @@ processed — is closed, as is `halt_error`'s wrong exit code and the stderr lea
 `isempty`/`first`/`limit`/`nth`/`any`/`IN(s)`, and — since Stage 4 — `IN(src; s)` and every
 other compare reached through a lazy consumer. **Six** shapes remain divergent, all pinned
 as such in `test_short_circuit_side_effect_leaks_820_932_987`. Two are the bare-`first(...)`
-family: `first(.[] | stderr)` and `first((1,2) == (10, ("B"|stderr)))`, both a `first(...)`
-over a non-`Comma`, which Stage 2b's sibling walk does not reach — option (c), #1461. One is
+family: `first(.[] | stderr)` (see "Option (c), scoped" below) and
+`first((1,2) == (10, ("B"|stderr)))`, both a `first(...)` over a non-`Comma`, which Stage 2b's
+sibling walk does not reach — option (c), #1461. One is
 `("A"|stderr) == (("B"|stderr), ("C"|stderr))`, a *top-level* compare, which never reaches
 `eval_each` at all — #1481. The remaining **three** are Stage 5's own residual (#1462): an
 un-lazified `eval_each` arm (`If`/`Try`/`Label`/`AsPattern`/`FuncCall`, ...) still leaks a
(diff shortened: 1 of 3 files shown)


newhoggy:
Addressed the code review findings. Summary:

**3 confirmed regressions, fixed** by narrowing the `Pipe` arm's scope to exactly 2-stage pipes and deferring `Many`/`ManyCursor`/`LazyKeys`/`LazyIndexRange`/`LazySeq` prefixes back to the existing whole-pipe fallback (safe to redo — none of those shapes have run a side effect yet) instead of attempting per-value backtracking on them:
1. `first(map(f) | .[0])` was erroring instead of short-circuiting (unconditional `materialize_lazy()` defeated the `LazySeq` composability optimization).
2. `first(.[] | select(...) | line)`-shaped queries lost cursor/position data (splitting a multi-stage pipe into a truncated prefix + separately-dispatched tail defeated `eval_single`'s own multi-stage `ManyCursor` cursor-threading).
3. Doc comments overclaimed that this closed every comma-inside-a-pipe side-effect leak; corrected to note it only ever closed the tail position.

This means `first(.[] | stderr)` is no longer closed as a side effect of the fix — moved back to the pinned "known-still-wrong" table, and the design doc corrected to describe option (c) as scoped rather than resolved. #1451's own two named repros are unaffected and still verified against the pinned jq oracle.

Also rebased onto main to pick up #1459 (merged concurrently, touching the same design doc and test table) and resolved the resulting conflicts — including one I initially got wrong (dropped a row during the first conflict resolution, caught and fixed before finalizing).

Remaining findings (function duplication between the `V`- and `OwnedValue`-flavored twins, an avoidable clone in a fallback path) are left as documented, low-priority tradeoffs — full re-verification (build, full test suite, fmt, both clippy legs, no_std, rustdoc, coverage) all pass.

Issue #1507, opened 2026-08-23
yq: decide whether input/inputs/input_line_number should exist in yq mode, and close the unreached-call-site divergence


newhoggy:
Split out of #1309. **Rewritten 2026-08-23** — the original framing of this
issue argued "it would be an extension, therefore do not build it." That
conflated a category question with a cost question, and got the category wrong
under the project's own policy. Corrected below.

## The facts (all captured live, per ADR-0018 rule 1)

Real yq v4.53.3 has no `input`/`inputs`/`input_line_number` at any arity. Its
lexer rejects the identifiers exactly as it rejects a name that does not exist:

```
$ printf 'a: 1\n' | yq 'input'       # Error: 1:1: lexer: invalid input text "input"
$ printf 'a: 1\n' | yq 'no_such_fn'  # Error: 1:1: lexer: invalid input text "no_such_fn"
```

yq handles multiple documents through its own model (`document_index`, `---`,
`--split-exp`, `eval-all`), not a jq-style input stream.

## Part 1 — the actual divergence, which is small and open

Both tools reject all three today. They differ in *when*, and that leaks:

| Filter on `a: 1` | real yq | `succinctly yq` |
|---|---|---|
| `input` | lexer error, exit 1 | `Error: input is not supported in yq mode`, exit 1 |
| `., input` | lexer error at 1:4, exit 1 | runtime error, exit 1 |
| `if false then input else . end` | lexer error, exit 1 | **exit 0, prints `a: 1`** |

`input_builtins_unsupported_in_yq_mode` (`src/jq/eval.rs`, added by #723) fires
at *dispatch*, so an unreached call site is silently accepted. Real yq fails at
*lex* and is therefore unconditional.

Only row 3 differs in outcome rather than wording. No ADR-0018 rule 4 condition
covers it — output is readable, nothing is corrupted, no process dies — so it
is an out-of-policy divergence, the same status the ADR itself assigns the
`*+d` merge-flag case.

The rule-4 scope boundary (adr-0018.md:207-214) does **not** rescue it: that
boundary reaches "spec-level input questions" — its example is YAML 1.1 vs 1.2
scalar resolution — and the text says explicitly that an evaluator behaviour
"may not be reframed as one to escape rule 4."

**Status: recorded and pinned** as of #1508 —
`docs/compliance/yq/limitations.md` § "`input`, `inputs`, `input_line_number`
are rejected, but at runtime rather than parse time", plus
[...]

Issue #1506, opened 2026-08-23
refactor(jq): four hand-maintained 207-arm Builtin traversals should share one accessor


newhoggy:
Cleanup surfaced by #1309.

## Summary

`Builtin` has 207 variants; 82 carry sub-expressions. The codebase currently
hand-maintains **four** separate full traversals over that variant set, each
with its own copy of which variant owns which sub-expressions:

| Function                            | File            | Shape                | Approx. lines |
|-------------------------------------|-----------------|----------------------|---------------|
| `substitute_var_in_builtin`         | `src/jq/eval.rs` | `&Builtin -> Builtin` | ~835 |
| `expand_func_calls_in_builtin`      | `src/jq/eval.rs` | `&Builtin -> Builtin` | ~1048 |
| `substitute_func_param_in_builtin`  | `src/jq/eval.rs` | `&Builtin -> Builtin` | ~323 |
| `builtin_kids`                      | `src/jq/walk.rs` | `&Builtin -> BuiltinKids<'_>` | ~180 (added by #1309) |

All four are exhaustive with no wildcard, so none of them can silently miss a
*new* variant. But they can and do drift on the *content* of an arm: each
independently restates which fields to recurse into, so an arm that recurses
into one of a two-argument builtin's sub-expressions and not the other is a
one-line mistake with no compiler check behind it.

This is the pattern CLAUDE.md calls out directly ("Duplicated predicates
diverge silently — one definition, plus a test that the call sites agree";
#106 had three copies of one predicate, one of them quadratic). #1309 already
collapsed two such copies — `jq_runner`'s input-builtin gate and
`yq_runner::contains_split_doc` — onto `builtin_kids`, and found that the
`contains_split_doc` copy had been silently missing twelve variants.

## Suggested direction

`builtin_kids` is a borrowing accessor and serves predicates. The three
`eval.rs` functions are transformers, so they need a mapping twin:

```rust
fn map_builtin_subexprs(b: &Builtin, f: &mut dyn FnMut(&Expr) -> Expr) -> Builtin
```

written once with the same no-wildcard discipline, with the three call sites
becoming one-liners that pass their own per-expression rewrite. #1505 needs
exactly this function too, so whichever lands first should build it.

**Note the payoff is bounded and the risk is not zero**: ~2200 lines collapse
to ~200, but the three functions are on the AST-rewrite path that every
user-defined `def` goes through, so it wants a careful differential pass rather
[...]

Issue #1505, opened 2026-08-23
jq: namespaced call inside a builtin argument fails with "module not loaded"


newhoggy:
Found while working #1309. Not part of that issue — filing separately.

## Summary

`rewrite_namespaced_calls` (`src/bin/succinctly/jq_runner.rs`) is exhaustive
over `Expr` with no wildcard, but its terminal arm group ends
`| Expr::Builtin(_) => expr`. It therefore never descends into a builtin's own
arguments, so an `import`ed function called from inside one is never rewritten
from `Expr::NamespacedCall` into a plain `Expr::FuncCall`. Evaluation then
reaches `eval.rs`'s `Expr::NamespacedCall` arm and fails.

## Repro (verified live against pinned jq 1.7.1)

With `mymath.jq` containing `def f: . * 2;` in directory `d`:

```
$ echo '[1,2]' | jq            -c -L d 'import "mymath" as m; 21 | m::f'
42
$ echo '[1,2]' | succinctly jq -c -L d 'import "mymath" as m; 21 | m::f'
42                                                    # bare call: fine

$ echo '[1,2]' | jq            -c -L d 'import "mymath" as m; map(m::f)'
[2,4]
$ echo '[1,2]' | succinctly jq -c -L d 'import "mymath" as m; map(m::f)'
jq: error (at <stdin>:1): module 'm' not loaded (namespaced call m::f)

$ echo '[1,2]' | jq            -c -L d 'import "mymath" as m; [.[] | select(m::f > 2)]'
[2]
$ echo '[1,2]' | succinctly jq -c -L d 'import "mymath" as m; [.[] | select(m::f > 2)]'
jq: error (at <stdin>:1): module 'm' not loaded (namespaced call m::f)
```

Any builtin taking a filter argument is affected — `map`, `select`,
`with_entries`, `walk`, `sort_by`, `any`/`all`, `limit`, `sub`/`gsub`, and the
rest of the 82 that carry sub-expressions.

**Severity: Medium.** It is a hard error rather than a wrong answer, and the
workaround (bind through a local `def`) is easy, but `map(m::f)` is about the
most natural way to use an imported function.

## Same class as #1309 item 1

#1309 fixed a sibling bug: a `Builtin` arm that swallowed a subtree and made a
whole-tree question answer wrongly. It also built the machinery that makes this
fix mechanical — `jq::walk::builtin_kids` (`src/jq/walk.rs`) declares every
[...]

Issue #1504, opened 2026-08-23
jq: inputs | f does not interleave, so input_line_number reports the last document


newhoggy:
Split out of #1309 (item 3), which fixed the other four gaps.

## Summary

`inputs` is a demand-driven producer as of #1309, but the pipe that consumes
it is still eager, so the whole stream is produced before the downstream stage
runs on any of it. `input_line_number` makes it visible, since it reports the
position of the last document actually read:

```
$ printf '1\n2\n3\n' | jq            -cn 'inputs | input_line_number'   # 1 2 3
$ printf '1\n2\n3\n' | succinctly jq -cn 'inputs | input_line_number'   # 3 3 3
```

Verified live against pinned jq 1.7.1.

The truncating consumers are already correct, because they *stop* the producer
rather than piping through it:

```
$ printf '1\n2\n3\n' | succinctly jq -cn 'first(inputs) | input_line_number'      # 1  (matches jq)
$ printf '1\n2\n3\n' | succinctly jq -cn 'limit(1; inputs) | input_line_number'   # 1  (matches jq)
```

## Second symptom, same root cause

`(., input) | error("boom")` over two files raises **once** here and **twice**
in jq. jq's lazy comma reaches `error` with document 1 before `input` is ever
evaluated, so its outer loop runs a second time; succinctly's eager comma
consumes document 2 while building the pair, leaving nothing for a second
iteration. The location succinctly does report is correct for the position its
parser reached — this is an error *count* difference, not a location bug.

## Why this is not a narrow fix

Neither symptom is specific to the input builtins. Both are the eager
`Expr::Pipe`/`Expr::Comma` in `eval_generic.rs`, which is the module the CLI
actually enters. #1309 sidestepped it for `first(...)` by delegating the whole
node to `eval.rs` (which has been demand-driven since #820 Stage 2); a general
fix needs `Demand`/`Item`/`Flow` mirrored into `eval_generic` for
`Comma`/`Pipe`/`Paren`.

That is already scoped as **option (c)** in
[`docs/plan/jq-lazy-generator-consumers.md`](../blob/main/docs/plan/jq-lazy-generator-consumers.md),
where it was originally filed for `first(.[] | stderr)`. This issue gives it
two more owners.

## Also here: `input_line_number` after a failed read

jq resets it to 0 after an `input` that finds nothing, but *not* after an
`[inputs]` that exhausts the same stream:

```
[...]

PR #1508, opened 2026-08-23
fix(jq): close four of #1309's five input/inputs architectural gaps

newhoggy:
Closes items 1, 2, 4 and 5 of #1309. Item 3 is split out to #1504 with the
reason it could not be done here.

Every expectation in this PR was captured live from pinned jq 1.7.1
(`/usr/bin/jq`) and Homebrew yq v4.53.3 before and after the change. Two of
the values I first wrote from reasoning were wrong (`., all(inputs; . > 2)`
and `., IN(inputs)`); the oracle corrected both.

## Two premises in the issue turned out to be wrong

**1. "This codebase has no lazy generator machinery at the `Builtin` dispatch
level."** Stale. #820 Stages 1/2/2b/3 built exactly that, and `first`, `limit`,
`nth`, `any`, `all`, `isempty` are all already wired to it. `inputs`
over-consumed only because `eval_each` had no arm for it. The fix is ~40 lines,
not the architectural rewrite the issue scoped.

**2. The suggested fix for item 4 — "thread `Option<String>` (filename)
alongside the line number through the queue" — encodes the wrong model.** jq
tracks a single *global* position: the file its parser has open and the line it
last finished a value on, not each value's provenance. The decisive probe:

```
$ jq '[inputs] | .[0] | error("boom")' a b c
jq: error (at c:1): boom          # names c, though .[0] is a's document
```

Per-value provenance would have reported `a` — wrong, and more expensive than
the right answer.

## What changed

| Commit | |
|---|---|
| `feat(jq)` | New `src/jq/walk.rs`: one exhaustive `Expr`/`Builtin` sub-expression walker, no wildcard at either level |
| `fix(jq)` | Item 1 — detect input builtins inside `-L`/`import`/`include` module bodies |
| `refactor(yq)` | Migrate `contains_split_doc` onto the shared walker (fixes 12 missed builtins) |
| `fix(jq)` | Item 2 — `inputs` becomes a demand-driven producer |
| `fix(jq)` | Item 2 — `first(inputs)`, which reaches a different evaluator |
| `fix(jq)` | Items 4+5 — filename and live position through the queue |
| `docs(jq)` | limitations, lazy-generator plan, jq-language reference, CHANGELOG |

### Item 1 — module detection

The gate was `filter_str.contains("input")`, run *before*
`ModuleLoader::process_program` inlines module bodies. It is now an AST walk
over the expanded tree. With `def readNext: input;` in `d/mylib.jq`:

```
$ printf '1\n2\n3\n' | sjq -L d 'import "mylib" as m; ., m::readNext'
# before: three `break` errors, one per document, none of them real
# after:  byte-identical to jq — stdout 1/2/3, one `break`, exhaustion location included
```

Now exact in both directions: `.input`, `.inputs` and an `"input"` string
literal no longer force the non-lazy read path either. I verified the two paths
agree on number-literal fidelity (`1.10`, `4E+4`, `2.50`) rather than assuming
it — the `can_use_lazy_path` comment claiming the other path "loses number
formatting" is stale since #1058/#1093.

### Item 2 — the eager drain

`drain_result`'s invariant ("an un-lazified arm is a missed optimization, never
a behaviour change") is stated over *values delivered to a sink* and says
nothing about side effects on state the sink does not own. Every other producer
over-*computes*; `inputs` over-*consumes* a shared, non-replayable queue that
the CLI's own driver loop draws from, so a document produced for a consumer
that never asked for it is gone from the whole program. `drain_all` is deleted
rather than left unused.

```
                          before   after (= jq 1.7.1)
., limit(1; inputs)       1 2      1 2 3 4 5
., first(inputs)          1 2      1 2 3 4 5
., nth(1; inputs)         1 3      1 3 4
., any(inputs; . > 2)     1 true   1 true 4 true
., all(inputs; . > 2)     1 false  1 false 3 true
., isempty(inputs)        1 false  1 false 3 false 5 true
```

`first` needed its own fix: it has a native `eval_generic` arm and never
crosses into `eval.rs`. Delegating the whole `first(...)` node rather than
special-casing `Builtin::Inputs` also gets `first(inputs, 1)`, `first(1,
inputs)` and `first(inputs | f)`. Gated behind a one-load
`input_queue_is_active` check so yq mode, library embedders and the hot
`first(.[])` never pay for the AST walk — the bridge costs the subtree its
cursor and with it #607's duplicate-key fidelity.

`last(inputs)` is deliberately unchanged: it cannot know a value is the last
until the stream ends, and jq agrees (`1 5`).

### Items 4+5 — location fidelity

16 of 17 location shapes now match byte-for-byte. **No filename crosses the
library seam** — the queue carries the opaque `(source, line)` tag
`InputLocations` already stores, and the CLI resolves it against the file table
it already owns. No `Rc`, no per-document `String`.

```
                                          before        after (= jq)
input | error("boom")  two-doc file       <stdin>:1     two.json:2
[inputs]|.[0]|error("boom")  a b c        <stdin>:1     c:1
printf "" | jq -n input                   <unknown>     <stdin>:0
jq -n input,input  one.json empty.json    <unknown>     empty.json:0
jq -n input,input,input  a.json b.json    <unknown>     b.json:1
jq -n error("x")   (no input builtin)     <unknown>     <unknown>   (unchanged)
```

The 17th, `(., input) | error(...)`, still raises once where jq raises twice —
an eager-comma difference in error *count*, not location, that predates this
work. Split to #1504.

### The yq `split_doc` commit

`contains_split_doc` was exhaustive over `Expr` but ended its `Builtin` match
in `_ => false`, silently missing twelve sub-expression-carrying variants
(`any`/`all` both arities, `IN`/`INDEX` and stream forms, `fromstream`,
`truncate_stream`, `at_offset`, `at_position`). Kept as its own commit so a yq
regression bisects cleanly.

Behaviour change, unopposed by the oracle: real yq rejects every one of these
shapes at its lexer, or lacks the builtin entirely. I verified all twelve
single-result shapes are byte-identical before and after; only a filter pairing
one with a second top-level branch changes, and it changes toward the
predicate's stated contract. CLAUDE.md's "duplicated predicates diverge
silently" warning is the reason not to leave two walks in place.

## Deliberately not done

- **`input_line_number` after a failed read.** jq resets it to 0 after a failed
  `input` but *not* after an `[inputs]` that exhausts the same stream. One
  probe, two readings, no model — recorded as a divergence rather than guessed
  at.
- **yq-mode support.** Real yq has no these builtins at all (its lexer rejects
  them), so implementing them would be an extension, not conformance. See
  #1507, which corrects #1309's premise on this.

## Verification

All CI gates from `.github/workflows/ci.yml`, derived from the workflow rather
than a hand-written list:

```
build-default / build-cli / build-no-default      PASS
test-default / test-cli-suite / test-jq-cli       PASS  (864 + 1125 + …)
clippy x3 (all-features, leg2, leg3) -D warnings  PASS
fmt --all -- --check                              PASS
```

Golden, evaluator-parity and error-message suites all green. Every pre-existing
`_723` and `_820` test still passes, including
`test_short_circuit_side_effect_leaks_820_932_987`.

## Follow-ups filed

- #1504 — item 3: `inputs | f` does not interleave (needs `eval_generic`'s pipe
  demand-driven; already scoped as option (c) in the lazy-generator plan)
- #1505 — `map(m::f)` fails with "module not loaded": `rewrite_namespaced_calls`
  never descends into builtin arguments. Same class as item 1; found while
  working this
- #1506 — three more hand-maintained 207-arm `Builtin` traversals in `eval.rs`
  should share `walk`'s accessor
- #1507 — yq mode: premise correction, plus the one real divergence (runtime vs
  parse-time rejection)

Refs #1309, #723, #820


newhoggy:
## Differential sweep results

After two bugs slipped past the original harness (both fixed in this branch:
`InputLocations::single`'s broken values/locations invariant, and the slurp
exhaustion location naming a wrong file), I widened the differential to
**32 filters × 7 flag combinations × 5 input shapes = 1155 cases**, each
comparing stdout+stderr+exit against pinned jq 1.7.1.

**34 divergences, in four families. None is a regression from this PR** — each
was checked by asking whether it reproduces with *no input builtin in the
filter*, which means it never touches this PR's queue:

| family | count | verdict |
|---|---|---|
| `last(f)` on an empty stream yields nothing; jq yields `null` | 13 | pre-existing — reproduces as `last(empty)`. Filed **#1521** |
| `--slurp` reports the last *value*'s location, not the last *file*'s EOF | 13 | pre-existing — reproduces as `-s 'error("x")' a.json e.json`. Filed **#1520** |
| `inputs \| input_line_number` does not interleave | 4 | known, split to **#1504** |
| `-s -n` variants of the above two | 4 | same causes |

The remaining **1121 cases match byte-for-byte**, stderr and exit code
included.

Sweep script and full output are reproducible from the corpus in the #1520
description (`a.json`=`1\n`, `b.json`=`2\n`, `e.json`=empty).

### Worth noting about #1520

It is adjacent enough to be worth a look before merging this: the slurped
value's location and this PR's slurp *exhaustion* location come from the same
seam, and `InputLocations::single()` / `::exhausted()` should probably share one
"last source, at its own newline count" helper. I left it out because it
reproduces with no input builtin and so is squarely outside #1309's scope —
but if you would rather it landed together, say so and I will fold it in.

PR #1526, opened 2026-08-23
fix(jq): keep resolve_slice_expr's bound-generator partial prefix

### LINKED ISSUE #1517: jq: resolve_slice_expr drops a bound generator's partial prefix, so first()/limit() in path position refuse a filter jq answers
**Severity: Low** (refuse-only — succinctly declines a filter jq accepts; never a wrong answer and never a write jq would not perform)

## Summary

`resolve_slice_expr` (`src/jq/eval.rs`) drops the **bound** generator's own partial prefix when
that generator escapes. The resulting `Err` prefix is therefore *shorter* than the one jq's
generator would have produced, which under-satisfies `take_path_branches`' `>= n` test, so the
escape propagates and `first`/`limit` in path position refuse a filter real jq answers.

Split out of PR #1465 (#972), which documented this residual in `take_path_branches`' doc comment
but did not file it. Confirmed pre-existing — it is not something #972's fix introduced.

## Repro

Verified against pinned jq 1.7.1 on `main` @ `53f57d8c1` (the slice machinery is untouched since):

```
$ echo '[10,20,30]' | jq -c 'path(limit(1; .[(0,error("b")):(2,3)]))'
[{"start":0,"end":2}]                                    exit 0
$ echo '[10,20,30]' | succinctly jq -c 'path(limit(1; .[(0,error("b")):(2,3)]))'
jq: error (at <stdin>:1): b                              exit 5
```

Three more shapes diverge identically — the escape can sit in either bound, and `first` behaves
like `limit(1; ...)`:

| Query on `[10,20,30]` | jq 1.7.1 | succinctly |
|---|---|---|
| `path(limit(1; .[(0,error("b")):(2,3)]))` | `[{"start":0,"end":2}]` | raises `b` |
| `path(first(.[(0,error("b")):(2,3)]))` | `[{"start":0,"end":2}]` | raises `b` |
| `path(limit(1; .[(0,1):(2,error("b"))]))` | `[{"start":0,"end":2}]` | raises `b` |
| `path(limit(1; .[(0,error("b")):2]))` | `[{"start":0,"end":2}]` | raises `b` |

**The write path refuses rather than mis-writing**, which is what keeps this Low:

```
$ echo '[10,20,30]' | jq -c 'del(limit(1; .[(0,error("b")):(2,3)]))'
[30]
$ echo '[10,20,30]' | succinctly jq -c 'del(limit(1; .[(0,error("b")):(2,3)]))'
jq: error (at <stdin>:1): b
```

## Root cause

Already stated precisely in `take_path_branches`' own doc comment (`src/jq/eval.rs`):

> **Only the "not longer" half holds universally, and only that half is what this function
> needs.** A prefix *shorter* than jq's is still possible — `resolve_slice_expr` drops the bound
> generator's own partial prefix, so `[10,20,30] | path(limit(1; .[(0,error("b")):(2,3)]))` raises
> here where jq answers `[{"start":0,"end":2}]`, exit 0 (pre-existing, on `main` too). A short
> prefix under-satisfies the `>= n` test below, so the escape propagates and the caller refuses —
> visibly wrong, never a false success, and never a write jq would not perform.

`resolve_slice_expr` resolves its `starts`/`ends` bound lists *before* building any branch, so a
bound generator that yields some values and then escapes returns with an empty (or short) `out` —
the branch jq would already have emitted for the successful bound value never gets built.

Contrast the **target**-escape case, which #972 already fixed: when `target` escapes,
`resolve_slice_expr` restricts its bound loops to a single element so its prefix matches jq's
exactly. The bound-escape case has no equivalent handling.

## Not currently pinned by a test

Worth stating explicitly, because it is easy to assume otherwise from PR #1465's test plan: **no
test asserts this shape.** #972's Guard D (`test_first_limit_path_undersatisfied_still_raises_972`)
covers `select`-shaped under-satisfaction, not slice bounds, and the only slice test in that family
(`path((select((true,error("t"))))[(0,1):(2,3)])`) is the *target*-escape guard with literal bounds.
So this divergence could change in either direction today with nothing failing.

Adding a characterization test is worthwhile **independent of whether the fix is attempted** — it
pins current behaviour on the path whose invariant has already caused one silent-data-corruption
revert.

## Suggested fix direction

Give the bound loops the same treatment the target-escape case already has: build the branches for
the bound values that *did* resolve before the escape, so the `Err` prefix equals jq's own
pre-escape stream, then let `take_path_branches`' `>= n` test do its job.

**Hazard — read `take_path_branches`' doc comment before touching this.** Its invariant is
deliberately one-directional ("never *longer* than jq's"), and the "exactly equal" form is
precisely what PR #985 assumed and got reverted for: `del(limit(2; select((true,error("t")))[("x","y")]))`
silently deleted both keys instead of raising. Lengthening this prefix moves it toward satisfying
`>= n`, which is the direction that turns a visible refusal into a **silent wrong write** if the
new prefix over-shoots. Any fix here needs the `del`/`=`/`|=` shapes in the tables above verified
against jq, not just the `path()` ones.

## Related

- #972 / PR #1465 — reinstated `take_path_branches`; documented this residual without filing it
- PR #985 — the reverted attempt whose "exactly equal" assumption this function's comment warns against
- #1283 — the sequencing plan whose re-sweep surfaced it

newhoggy:
## Summary
- `resolve_slice_bound` resolved `start`/`end`'s own bound generator via the discard-prefix `eval_owned_multi`, so a bound generator that produced some values before escaping (e.g. `(0,error("b"))`) reset the whole bound list to empty instead of keeping what it had already produced. Unlike `target`'s own escape (#973), which already correctly kept its partial prefix, this made `resolve_slice_expr`'s `Err` prefix shorter than jq's real pre-escape stream, under-satisfying `take_path_branches`' `>= n` test -- `first`/`limit` in path position wrongly refused a filter real jq answers successfully.
- `resolve_slice_bound` now calls `eval_owned_multi_keep_partial` (mirroring `resolve_index_expr`'s own `key`/`key_escape` split) and returns the escape alongside its partial value list.
- `resolve_slice_expr` combines the three possible escape sources -- `target` (`E`, innermost), `end` (`T`, middle, re-run fresh per `start` value per jq's `S as $s | T as $t | E | .[$s:$t]` compilation), `start` (`S`, outermost) -- with innermost-wins priority, since jq's real depth-first execution means a more-nested escape fires before an outer generator ever gets a chance to expose its own:
  - `target` escaping restricts both bound lists to their first element (unchanged, pre-existing #972 fix).
  - `end` escaping restricts only `start` to its first element (`end`'s own list is already its correctly-truncated partial prefix).
  - `start` escaping needs no restriction at all (every earlier `start` value already had `end`/`target` run to full completion, unaffected by `start`'s later escape).

**Hazard this needed care around**: `take_path_branches`' own doc comment explicitly warns that PR #985 attempted a similar fix on the analogous key/target axis with an "exactly equal" assumption and got reverted for silently over-satisfying `limit(n)` -- `del(limit(2; select((true,error("t")))[("x","y")]))` silently deleted both keys instead of raising. This fix's own design was verified live against that exact class of boundary (the precise `limit(n)`-satisfied vs `limit(n+1)`-refuses cutoff on both the `start`-escape and `end`-escape axes) before writing any code, not just the narrower `path()`-only repro.

Fixes #1517

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo build --no-default-features` (no_std) -- pre-existing, unrelated `sink` warning confirmed present on unmodified main too, not introduced by this change
- [x] `cargo llvm-cov --features cli,simd,regex,serde --workspace --fail-under-lines 0` -- 100% line coverage on every changed line
- [x] Live-verified every case against the pinned `jq` oracle (`/usr/bin/jq` 1.7.1): all 4 of the issue's own repro shapes, the `path()`-only forms for both escape axes, and -- critically -- the exact write-side `limit(n)`-satisfied vs `limit(n+1)`-refuses boundary on both axes (matching jq's real branch count precisely, not just "enough to pass the narrow repro")

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -17436,15 +17436,24 @@ fn resolve_slice_expr<'a, S: EvalSemantics>(
     }
     let (ends, ends_escape) =
         resolve_slice_bound::<S>(end, value, f64::ceil).map_err(|e| (Vec::new(), e))?;
+    // `end` (`T`) is nested inside `start` (`S`)'s own iteration -- jq's `S
+    // as $s | T as $t | ...` re-runs `T` fresh for `$s`'s very first value,
+    // so a `T` escape (if any) fires before `S` ever gets a chance to
+    // expose its own. `ends_escape` therefore wins over `starts_escape` --
+    // computed once as `bounds_escape` and reused both by the
+    // `ends.is_empty()` early return just below and by the full
+    // `target`/`end`/`start` priority chain further down, so the two sites
+    // that need this rule can't silently drift apart (#1526 review).
+    // `ends_escaped` is split off first since `.or()` moves its operand and
+    // the truncation logic below still needs to ask "did `end` specifically
+    // escape" on its own. Confirmed against jq 1.7.1:
+    // `path(.[(0,1):error("b")])` on `[10,20,30]` prints nothing before
+    // raising `b` -- `end`'s own escape, with *zero* prior values, still
+    // wins over `start`'s later one.
+    let ends_escaped = ends_escape.is_some();
+    let bounds_escape = ends_escape.or(starts_escape);
     if ends.is_empty() {
-        // `end` (`T`) is nested inside `start` (`S`)'s own iteration --
-        // jq's `S as $s | T as $t | ...` re-runs `T` fresh for `$s`'s very
-        // first value, so a `T` escape (if any) fires before `S` ever gets
-        // a chance to expose its own. `ends_escape` therefore wins over
-        // `starts_escape` here, same "innermost escape wins" rule as the
-        // `target_escape`/`ends_escape`/`starts_escape` priority chain
-        // below.
-        return path_result(Vec::new(), ends_escape.or(starts_escape));
+        return path_result(Vec::new(), bounds_escape);
     }
     // #843: same rule as `resolve_index_expr` above, for a slice's bounds
     // instead of an index's key.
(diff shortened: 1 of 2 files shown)


newhoggy:
Addressed the code review findings:

1. **Hoisted the duplicated `ends_escape.or(starts_escape)`** (flagged independently by two finder angles) into one `bounds_escape` variable, reused by both the `ends.is_empty()` early return and the final priority combine — the two sites can no longer silently drift apart.
2. **Added the missing live-oracle citation + test** for the zero-prior-value `end`-escape sub-case the `ends.is_empty()` early return guards (`path(.[(0,1):error("b")])` → jq raises immediately, 0 branches — verified and now pinned).
3. **Flattened the three-arm truncation branch** into two independent length computations (`starts_len`/`ends_len`), removing the repeated `&starts[..1]` across two branches that obscured *why* both trigger it.
4. **Documented and pinned, rather than silently fixed, a genuine residual**: `resolve_slice_bound`'s type-conversion step still discards the whole converted prefix on the first non-numeric bound value (`path(.[(0,1,"x"):(2,3)])` drops all four valid jq branches instead of raising the correct type error with them first) — confirmed by two independent reviewers, live-verified as pre-existing (with a *different*, wrong error message) rather than a regression from this PR. Left out of scope deliberately: extending keep-partial semantics to a type failure needs its own priority-ordering pass against `target_escape`/`bounds_escape`, re-verified against jq the same way this PR's own fix was, not a quick addition under review pressure.

**Not addressed in this PR, filed as a follow-up instead**: one reviewer found the identical class of bug in the *value-mode* evaluators (`eval_generic.rs`'s and `eval.rs`'s own `eval_slice_bound`, reached via `.[(0,error("b")):(2,3)]` in value position, not `path(...)`) — genuinely out of scope for #1517, which is specifically about `path()`/`first`/`limit` in path position. Filed as #1528.

Skipped the suggested generalized N-level "resolve_nested_fanout" helper — this is only the 2nd instance of the pattern (index=2-level, this=3-level); premature to abstract at 2 instances per this repo's own anti-premature-abstraction convention.

Full re-verification (build, full test suite, fmt, both clippy legs, no_std, rustdoc, coverage — 100% line coverage on every changed line) all pass; every case re-verified against the pinned jq oracle after the restructuring.

Issue #1309, opened 2026-08-20
jq: input/inputs architectural gaps — -L module detection, eager-drain data loss, yq mode, location fidelity


newhoggy:
**Severity: Medium**

## Summary

Follow-up from #723 (PR #1308's implementation of `input`/`inputs`/
`input_line_number`). Code review found several real gaps in the initial
implementation. The two most severe were fixed directly in PR #1308 before
merge (a TTY-safety branch that fabricated a phantom `null` document, and
`succinctly yq` silently accepting-but-never-implementing these builtins —
now gated to a clear "not supported in yq mode" error). The gaps below are
genuine but architecturally deeper, and were deliberately left for this
follow-up rather than expanding PR #1308's scope further.

## 1. `-L`/module-inlining bypasses `uses_input_builtins` detection

`jq_runner.rs`'s `uses_input_builtins = filter_str.contains("input")` check
runs on the raw CLI filter text, before `module_loader.process_program`
inlines any `-L`/`import`/`include`-loaded module body. A call to `input`/
`inputs`/`input_line_number` that only exists inside an imported module's
own function is invisible to this check, so the filter takes the fast/lazy
CLI path (which never seeds the shared input queue) and produces wrong
output.

**Repro** (verified live, real jq 1.7.1 as oracle): with a module file
`mylib.jq` containing `def readNext: input;`, run
`printf '1\n2\n3\n' | succinctly jq -L <dir> 'import "mylib" as m; ., m::readNext'`.
Real jq: stdout `1`/`2`/`3`, one stderr `break` (from the final exhausted
call). succinctly: stdout is the same, but stderr has *three* `break`
errors — every `readNext` call fails immediately regardless of what's left
in the stream, because the queue was never seeded.

**Suggested fix**: replace the text heuristic with an AST walk over the
*expanded* `expr` (after `module_loader.process_program` runs), using the
same exhaustive, no-wildcard `Expr`/`Builtin` match shape
`yq_runner.rs`'s own `contains_split_doc` already establishes as the safe
pattern for "does this tree reference X anywhere" — not a quick fix, real
work to replicate that exhaustiveness for jq's larger `Expr`/`Builtin` set.

## 2. Eager, destructive `inputs` draining loses data under `first`/`limit`/`any`/`all`

`builtin_inputs` calls `remaining_inputs::drain_all()`, which eagerly pops
the *entire* remaining queue in one call. This codebase's evaluator is
fully eager (every generator-truncating combinator materializes its whole
[...]

PR #1540, opened 2026-08-23
fix(jq): rewrite namespaced calls inside builtin arguments

### LINKED ISSUE #1505: jq: namespaced call inside a builtin argument fails with "module not loaded"

newhoggy:
## Summary
- `rewrite_namespaced_calls` (`src/bin/succinctly/jq_runner.rs`) was exhaustive over `Expr` with no wildcard, but its terminal arm group ended `| Expr::Builtin(_) => expr`, so it never descended into a builtin's own arguments -- an imported function called from inside `map`/`select`/`limit`/`sub`/... (any of the 82 sub-expression-carrying builtins) was never rewritten from `Expr::NamespacedCall` to `Expr::FuncCall`, and evaluation failed with "module not loaded" even though the same call worked fine bare (`m::f` alone).
- Added `map_builtin_subexprs` to `src/jq/walk.rs` as the mapping twin of the existing `builtin_kids` accessor (#1309): exhaustive over every `Builtin` variant with no wildcard arm, so a future variant that carries a sub-expression is a compile error here until it's declared, matching `builtin_kids`' own discipline.
- `rewrite_namespaced_calls` now gives `Expr::Builtin` its own arm using this to recurse into every sub-expression instead of returning the builtin unchanged.

Every shape from the issue's own repro table verified live against pinned jq 1.7.1: `map(m::f)`, `select(m::f > 2)` in a comprehension, a two-argument builtin (`pow(x; m::y)` -- note `limit(n; f)` itself parses into a dedicated `Expr::Limit` AST node, not `Expr::Builtin`, so it doesn't actually exercise the new code path despite being a common idiom; kept as a regression pin anyway), a three-argument builtin (`sub(re; s; m::flags)`), and the `include` spelling.

Also fixed a `no_std` build break this surfaced: `walk.rs`'s new `Box::new` calls needed an explicit `use alloc::boxed::Box` import (matching the existing convention in `expr.rs`) -- caught by the no_std check, not by the default feature build.

Fixes #1505

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo build --no-default-features` (no_std) -- initially failed with 106 errors (missing `Box` import), fixed
- [x] `cargo llvm-cov --features cli,simd,regex,serde --workspace --fail-under-lines 0` -- the new `map_builtin_subexprs` match (207 variants) is ~91% covered; the handful of uncovered arms are either provably unreachable via the parser (`Limit` parses into its own dedicated `Expr::Limit` node, never `Expr::Builtin(Builtin::Limit(..))`) or niche extensions (`at_offset`/`at_position`/`ModuleMeta`/etc.) not otherwise exercised by the existing suite -- not chasing 100% on an exhaustive-by-design match built for compile-time safety across all 207 variants, not runtime coverage of each one
- [x] Live-verified every case against the pinned jq oracle (`/usr/bin/jq` 1.7.1)

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -416,6 +416,21 @@ fn rewrite_namespaced_calls(expr: Expr) -> Expr {
         // a future variant that carries a sub-expression is a compile error
         // here until it's declared there, the same discipline this file's
         // own manual `Expr` recursion already follows above.
+        //
+        // A call to a namespace that was never actually `import`ed now
+        // fails the same way in this position as it already did bare
+        // (`nonexistent_ns::foo` outside any builtin) or piped -- rewritten
+        // to `Expr::FuncCall` unconditionally, regardless of import status,
+        // reporting "undefined function" from the general `FuncCall`
+        // resolution path rather than `eval.rs`'s own `Expr::NamespacedCall`
+        // arm's "module not loaded". Not a regression this fix introduces:
+        // confirmed live that the bare/piped case already answered
+        // "undefined function" before this change (`rewrite_namespaced_calls`
+        // never checked import status anywhere), and that real jq's own
+        // message for this shape is closer to "X/0 is not defined" (a
+        // compile error) than to either succinctly wording -- this arm
+        // brings the builtin-argument position in line with what every
+        // other position already did, rather than diverging from it.
         Expr::Builtin(builtin) => Expr::Builtin(map_builtin_subexprs(&builtin, &mut |sub| {
             rewrite_namespaced_calls(sub.clone())
         })),
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -14,6 +14,7 @@ use succinctly::jq::document::{collapsed_fields, effective_keys};
 use succinctly::jq::eval_generic::{
     eval_with_cursor, to_owned as generic_to_owned, GenericResult, MAX_NESTING_DEPTH,
 };
+use succinctly::jq::walk::map_builtin_subexprs;
 use succinctly::jq::{
     self, format_number_jq_compat, nonfinite_display_string, EvalError, Expr, JqSemantics, JqValue,
     OwnedValue, Program,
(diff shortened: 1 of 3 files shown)


newhoggy:
Addressed the code review findings:

1. **Strengthened the weak two-argument test**: `pow(2; m::two)` couldn't distinguish a correct reconstruction from a transposed one, since both operands evaluate to 2 either way. Switched to `pow(m::two; 3)` — a transposed `Pow(f(b), f(a))` would give 9 instead of the correct 8.
2. **Verified and documented the error-message finding**: a call to a never-imported namespace now reports "undefined function" in builtin-argument position too. Confirmed live this is *not* a regression — the bare/piped position already reported "undefined function" before this fix (`rewrite_namespaced_calls` never checked import status anywhere), and real jq's own message for this shape (`X/0 is not defined`, a compile error) is closer to succinctly's "undefined function" than to "module not loaded" either way. This fix makes the builtin-argument position consistent with what every other position already did, not a new divergence.
3. **Cross-referenced #1506** (four/five independently hand-maintained exhaustive `Builtin` traversals, none routing through another) from `map_builtin_subexprs`'s own doc comment — already filed separately, correctly not attempted here.
4. **Documented, not changed**, the deliberate borrow-based signature's one-time, depth-bounded clone cost — matches `builtin_kids`'s own established convention (which the issue itself asked this function to mirror), and the cost is bounded by the parser's own `MAX_EXPR_DEPTH` guard since this runs once per parsed program, not per document.

Full re-verification (build, full test suite, fmt, both clippy legs, no_std, rustdoc, coverage) all pass after rebasing onto main.

Issue #550, opened 2026-08-01
jq_cli_tests: run_jq_full can hit a transient ENOENT racing against concurrent cargo-run tests


newhoggy:
## Summary

`test_null_input_error_reports_unknown_location` (and, by the same mechanism, any other `run_jq_full`-based test in `tests/jq_cli_tests.rs`) can fail intermittently with:

```
Error: No such file or directory (os error 2)
  at jq_cli_tests::run_jq_full (tests/jq_cli_tests.rs:85)
```

Observed on PR #547, job [`Test (macOS ARM64)`](https://github.com/rust-works/succinctly/actions/runs/30683737297/job/91325583559), step "Run tests (portable-popcount feature)".

## Evidence it's a race, not a config/feature issue

The same test ran three times in that one job (once per `cargo test` invocation: `Run tests`, `Run tests (simd feature)`, `Run tests (portable-popcount feature)`) — identical code, identical binary path, identical feature-independent test:

```
04:22:16 test test_null_input_error_reports_unknown_location ... ok       (Run tests)
04:23:08 test test_null_input_error_reports_unknown_location ... ok       (Run tests, simd feature)
04:23:58 test test_null_input_error_reports_unknown_location ... FAILED   (Run tests, portable-popcount feature)
```

Passing twice and failing once, seconds apart, rules out a deterministic feature-flag or build-config cause — this is timing-dependent.

## Root cause (hypothesis)

`run_jq_full` (`tests/jq_cli_tests.rs:84`) execs a fixed, compile-time-resolved path:

```rust
let mut cmd = Command::new(env!("CARGO_BIN_EXE_succinctly"))
```

with no retry logic. 16 tests in the file use this helper.

The other 98 tests in the same file use `run_jq_stdin` / `run_jq_stdin_streams`, which instead invoke `cargo run --features cli --bin succinctly -- jq ...` — and that helper *does* have retry logic already (`MAX_CARGO_RETRIES`), added specifically because "exit code 101 often indicates cargo lock contention."

[...]

PR #1543, opened 2026-08-23
test(cli): report a signal-killed cargo run child as such, not as -1

### LINKED ISSUE #1516: jq_cli_tests: a signal-killed `cargo run` child surfaces as a bare `left: -1` with no diagnostic (#550 follow-up)
**Severity: Low** (test-infrastructure only — but its real cost is misdiagnosis: it cost a wrong root-cause call in the #1459 review, described below)

Follow-up to **#550**, which fixed one failure mode of the shared `target/debug/succinctly`
path and left the other two open. #550's own third suggested fix — "remove the shared mutable
path entirely" — was not taken.

## Symptom

Three `jq_cli_tests` failing in one run, with this and nothing else:

```
---- test_regex_gsub_available_in_cli stdout ----
thread 'test_regex_gsub_available_in_cli' panicked at tests/jq_cli_tests.rs:3598:5:
assertion `left == right` failed
  left: -1
 right: 0
```

Observed at `main` @ `4567bcab6` on macOS ARM64 during a local
`cargo test --features cli --test jq_cli_tests ...` run, on a heavily loaded machine
(load ~14, ~37 concurrent `cargo` processes from other worktrees). The three were
`test_regex_gsub_available_in_cli`, `test_regex_capture_flags_form_no_match_produces_no_output`
and `test_number_literal_overflow_owned_reindex_bridges_via_cli`. Two clean re-runs of the
same command immediately afterwards: **855/855 pass**.

## What `-1` actually means, and why that matters

`run_jq_stdin_streams` (`tests/jq_cli_tests.rs`) computes the code as:

```rust
let exit_code = output.status.code().unwrap_or(-1);
```

On Unix `ExitStatus::code()` returns `None` **only** when the child was terminated by a
signal. So `-1` is not "some error" — it is precisely *"the `cargo run` child was killed"*,
and it can mean nothing else. The helper's own retry does not cover it:

```rust
// Exit code 101 often indicates cargo lock contention; retry
if exit_code == 101 && attempt + 1 < MAX_CARGO_RETRIES {
```

`101` retries; a signal falls straight through to the caller's `assert_eq!`.

**The diagnostic is the actual bug.** `left: -1, right: 0` names no signal, prints no
stderr, and says nothing about a child process at all — so it reads like a behavioural
assertion failure in the code under test. In the #1459 review I read exactly this output and
concluded the tests were `regex`-gated and I had omitted the feature. That was wrong twice
over: `cli = [..., "regex"]` in `Cargo.toml`, so `--features cli` already implies `regex`,
and these three use `run_jq_stdin` (the `cargo run` helper), not `run_jq_full`. A message
that said "child killed by signal N" would have cost seconds instead of a wrong call in a
review summary.

## Why these tests are killable at all

`run_jq_stdin` / `run_jq_stdin_streams` and three other sites shell out to
`cargo run --features cli --bin succinctly` from **inside** a `cargo test` run. That makes
every one of those tests a `cargo` child, so anything that reaps `cargo` processes on the
box — an OOM kill, a `pkill`/`killall` from another session, a stray Ctrl-C — turns them into
this inscrutable assertion failure rather than an obvious infrastructure error. It also puts
them in contention for the build lock, which is what `MAX_CARGO_RETRIES` already exists to
paper over.

I cannot say what sent the signal in my run and am not claiming to: the box was heavily
loaded by other sessions, and earlier in the same session I had run a machine-wide
`pkill -f "cargo test"` (two runs earlier, so not directly this one). That uncertainty is the
point — the failure signature gives a reader no way to tell.

## Not covered by #550

#550 was the **ENOENT at `spawn()`** case, where `cargo run` rewrites the binary that
`run_jq_full`'s `CARGO_BIN_EXE_succinctly` path is about to exec. It was fixed with
`MAX_SPAWN_RETRIES`, which retries a failed *spawn*. It does not help here: the spawn
succeeds and the child dies afterwards.

Both are symptoms of the same shared mutable path plus the `cargo`-inside-`cargo` structure,
which #550 explicitly offered as its option 3 and which is still in place.

## Acceptance criteria

- A signal-terminated child is reported as such — signal number and the captured stderr —
  not as `-1` against an expected exit code. Enough on its own to close the misdiagnosis
  half.
- Preferably also #550's option 3: have these helpers exec the already-built binary
  (`CARGO_BIN_EXE_succinctly`) like `run_jq_full` does, instead of shelling out to
  `cargo run`. CI already builds it (`cargo build --features cli --bin succinctly`) before
  the test steps, so the `cargo run` call buys nothing there and costs the build lock, the
  rewrite race and this kill surface. A local-only "is it built?" guard would replace it.
- If the `cargo run` calls stay, extend the retry to cover signal death the way it covers
  `101`, and say so in the helper's doc comment.

Refs #550.

newhoggy:
## Summary

Follow-up to #550, which fixed the ENOENT-at-`spawn()` failure mode of the shared `target/debug/succinctly` path and left this one open (explicitly named as #550's own "option 3, not taken").

Four independent `cargo run`-based helpers in `tests/jq_cli_tests.rs` (`run_jq_stdin_streams`, `run_jq_file`, `run_jq_null`, `run_jq_binary_stdin`) each computed `output.status.code().unwrap_or(-1)` — but `ExitStatus::code()` returns `None` **only** when the child was terminated by a signal, so that `unwrap_or` silently turned "the child was killed" into a fake exit code indistinguishable from a genuine `-1`. The resulting `left: -1, right: 0` assertion failure names no signal and prints no stderr, so it reads like a behavioural bug in the code under test — exactly what cost a wrong root-cause call in the #1459 review, as the issue documents.

Adds a shared `classify_cargo_run_exit()` that all four call sites now use instead of duplicating the same retry-on-`101` check. It treats a signal death as transient the same way `101` (lock contention) already is — plausible under the heavy concurrent `cargo` load these helpers run under (an OOM kill, another session's `pkill`, or lock-contention fallout) — retrying up to `MAX_CARGO_RETRIES` times. Only once retries are exhausted does it surface as an error naming the signal number and the captured stderr, never as a silently-coerced `-1`.

New `test_classify_cargo_run_exit_reports_signal_death_1516` exercises the classifier directly against constructed `ExitStatus` values (via `ExitStatusExt::from_raw`), rather than by actually killing a `cargo run` child — which would make the test itself exactly as flaky as the bug it pins.

Tests only — no `src/` behaviour change.

## Scope note

The issue's acceptance criteria lists three items: (1) report signal death properly — explicitly called "enough on its own to close the misdiagnosis half"; (2) **preferably also** switch these helpers to exec the pre-built `CARGO_BIN_EXE_succinctly` binary like `run_jq_full` does, removing the `cargo`-inside-`cargo` structure entirely; (3) if the `cargo run` calls stay, extend the retry to cover signal death and document it.

This PR does (1) and (3) in full. (2) is left out — it's explicitly framed as "preferably", not required, and it's a materially larger, riskier change (every one of the ~100+ call sites across this file would need its feature-flag-resolution behavior re-verified against the pre-built binary instead of a fresh `cargo run --features cli`). Better suited to its own follow-up than bundled with a diagnostics fix; will file one after this merges.

## Test plan

- [x] New `test_classify_cargo_run_exit_reports_signal_death_1516` — covers signal-death-with-retries-left (asks to retry), signal-death-on-last-attempt (errors, naming signal + stderr), `101`-with-retries-left (retries, unchanged), `101`-on-last-attempt (surfaces as-is, unchanged), and genuine exit codes 0/1 (pass through unchanged)
- [x] `cargo test --features cli --test jq_cli_tests` — 875/875 pass

Closes #1516

### DIFF
--- a/tests/jq_cli_tests.rs
+++ b/tests/jq_cli_tests.rs
@@ -27,20 +27,44 @@ const MAX_CARGO_RETRIES: u32 = 3;
 /// mid-replacement and fail with `ENOENT` (#550).
 const MAX_SPAWN_RETRIES: u32 = 3;
 
+/// Builds the "child was killed by signal N" error for a signal-terminated
+/// process (`ExitStatus::code()` returns `None` only in that case, on
+/// Unix), naming the signal and the captured stderr -- shared by
+/// `classify_cargo_run_exit` below and by `run_jq_full`, which spawns the
+/// pre-built binary directly rather than through `cargo run` and so has no
+/// retry loop to hook a classifier into, but is just as exposed to a
+/// signal-killed child under heavy concurrent load. Historically both
+/// silently coerced this to `-1` via `.code().unwrap_or(-1)`, which renders
+/// as an inscrutable `left: -1, right: 0` with no indication a child was
+/// ever killed -- exactly what cost a wrong root-cause call in the #1459
+/// review (#1516).
+fn signal_death_error(status: std::process::ExitStatus, stderr: &str) -> anyhow::Error {
+    // `ExitStatus::code()` returns `None` only when the child was
+    // terminated by a signal (Unix) -- there is no other cause.
+    #[cfg(unix)]
+    let signal = {
+        use std::os::unix::process::ExitStatusExt;
+        status.signal()
+    };
+    #[cfg(not(unix))]
+    let signal: Option<i32> = None;
+    anyhow::anyhow!(
+        "child was killed by signal {}; stderr:\n{stderr}",
+        signal.map_or_else(|| "<unknown>".to_string(), |s| s.to_string()),
+    )
+}
+
 /// Classifies a `cargo run` child's exit for the retry loops below, so each
-/// of the four call sites doesn't hand-roll the same "retry on 101" check
-/// (and, historically, the same bug: silently coercing a signal death to
-/// `-1` via `.code().unwrap_or(-1)`).
+/// of the four call sites doesn't hand-roll the same "retry on 101" check.
 ///
 /// Returns `Ok(Some(code))` once a real exit code is available to the
 /// caller, or `Ok(None)` when the caller should sleep (per `attempt`) and
 /// retry -- covering both the existing `101` lock-contention case and a
-/// signal death within `MAX_CARGO_RETRIES` attempts. Once retries are
-/// exhausted on a signal death, returns `Err` naming the signal and the
-/// captured stderr, instead of letting the caller's exit-code assertion
-/// render as an inscrutable `left: -1, right: 0` with no indication a child
-/// was ever killed -- which is exactly what cost a wrong root-cause call in
-/// the #1459 review (#1516).
+/// signal death within `MAX_CARGO_RETRIES` attempts (a signal death under
+/// the same heavy concurrent load -- an OOM kill, another session's
+/// `pkill`, or fallout from that same lock contention -- is just as
+/// plausible as `101` is). Once retries are exhausted on a signal death,
+/// returns the `signal_death_error` above instead of `Ok(Some(-1))`.
 fn classify_cargo_run_exit(
     status: std::process::ExitStatus,
     stderr: &str,
@@ -52,24 +76,10 @@ fn classify_cargo_run_exit(
         }
         return Ok(Some(code));
     }
-
-    // `ExitStatus::code()` returns `None` only when the child was
-    // terminated by a signal (Unix) -- there is no other cause.
     if attempt + 1 < MAX_CARGO_RETRIES {
         return Ok(None);
     }
-    #[cfg(unix)]
-    let signal = {
-        use std::os::unix::process::ExitStatusExt;
-        status.signal()
-    };
-    #[cfg(not(unix))]
-    let signal: Option<i32> = None;
-    anyhow::bail!(
-        "cargo run child was killed by signal {} after {} attempt(s); stderr:\n{stderr}",
-        signal.map_or_else(|| "<unknown>".to_string(), |s| s.to_string()),
-        attempt + 1,
-    );
+    Err(signal_death_error(status, stderr))
 }
 
 /// #1516: a signal-killed child used to render as an inscrutable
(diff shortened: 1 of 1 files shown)


newhoggy:
\`/code-review\` (7-agent fan-out plus a completeness sweep) converged on one real regression and two real gaps in the first version of this PR, all fixed in the follow-up commit:

1. **Retry-defeating regression**: `run_jq_stdin_streams` strictly UTF-8-decoded stderr (`String::from_utf8(...)?`) *before* the retry decision — if a signal-killed child's stderr is truncated mid multi-byte sequence (plausible for the exact failure mode this PR targets), the `?` aborted the whole retry loop with an opaque `FromUtf8Error` instead of retrying, defeating both the pre-existing 101 retry and the new signal retry precisely when they matter most. Fixed by switching to lossy decoding for the classifier call (matching the other three sites) and deferring the strict conversion to the success path only.
2. **`run_jq_full` left un-migrated**: it spawns the pre-built binary directly (not through `cargo run`), so it has no retry loop to hook `classify_cargo_run_exit` into, but it's just as exposed to a signal-killed child under heavy load. Extracted a `signal_death_error()` helper shared by both `classify_cargo_run_exit` and `run_jq_full` directly (no retry added there — consistent with its existing design, since cargo lock contention doesn't apply to a direct binary exec).
3. **Missing `--quiet`**: `run_jq_file`/`run_jq_null`/`run_jq_binary_stdin`'s cargo invocations lacked it, so the stderr this PR surfaces in a signal-death error could be dominated by cargo's own progress text rather than the killed binary's actual output. Added to match `run_jq_stdin_streams` — safe since none of the three return stderr in their public signature.

Also added an interior-attempt test case for completeness, per a lower-priority reviewer note.

Other findings (the `#[cfg(not(unix))]` branch being untestable without a Windows CI runner; three other test files — `cli_golden_tests.rs`, `cli_characterization_tests.rs`, `deep_nesting_valid_tests.rs` — carrying the identical duplicated `.unwrap_or(-1)` pattern; several ad hoc in-test call sites that don't use any shared helper) were deliberately left out of this PR's scope and will be tracked in a follow-up issue.

Issue #1449, opened 2026-08-22
jq: more eval_pipe_with_path_context_internal arms could route through continue_rest_with_context


newhoggy:
**Severity: Low** (cleanup/maintainability, not a correctness bug)

## Summary

#1313 (PR #1445) converted `eval_pipe_with_path_context_internal`'s three arms named in that issue's own text (`Expr::Builtin(_)`, `Expr::Object|Array|Literal`, and the generic fallback `_`) to route through the already-shared `continue_rest_with_context` instead of each hand-rolling "if `rest` is empty, return; else recurse". Code review on that PR found several more arms in the same function with the identical hand-rolled shape, left unconverted since they weren't part of #1313's own named scope:

- **`Expr::Array(inner) if needs_path_context(inner)`** (the `[key]`/`[parent]`/`[file_index]` case, #1302) — hand-rolls the same "reset `root`/`current_path` to the newly constructed value" pattern #1445 just extracted from the `Object|Array|Literal` arm, unconverted.
- **`Expr::StringInterpolation(parts) if string_interpolation_needs_path_context(parts)`** — same "reset and recurse" pattern, unconverted.
- **Five early-return blocks** handling `Builtin::PathNoArg`, `Key`, `FileIndex`, `Parent`, and `ParentN` (near the top of the function, before the main `match`) — each hand-rolls the "keep `root`/`current_path` unchanged" flavor of the pattern #1445 converted for `Expr::Builtin(_)` and the generic fallback.

## Why this matters

This is exactly the "duplicated predicates diverge silently" pattern #1313's own issue text already named (CLAUDE.md/#106): a future fix to the shared continuation logic — e.g. #1445's own `rest.is_empty()` fast path, or a future change to how a control signal propagates through a root reset — has to be manually re-applied at each of these remaining hand-rolled sites, and a maintainer fixing `continue_rest_with_context` itself has no obvious reason to also check these five-plus call sites.

## Scope note

[...]

Issue #1428, opened 2026-08-22
jq: get_path_mut/set_path autovivify before confirming a suppressed (?) write actually happens


newhoggy:
**Severity: Low** (requires a `?`-suppressed write through a `null`/absent parent)

## Summary

`get_path_mut`'s `Field`/`Index` arms (`src/jq/eval.rs`) unconditionally `autovivify_object`/`autovivify_array` a `Null` parent while navigating *before* it's known whether the trailing component's own write will actually happen — so a `?`-suppressed write through an absent/`null` parent still leaves the autovivified structure behind, where real jq treats path collection as read-only until a write is confirmed.

Same root cause, same bug, in `set_path`'s own top-level terminal `Expr::Iterate` arm (`.a[]? = v` on `null`) — this isn't `Iterate`-specific or new; it's the general eager-autovivify-during-navigation shape shared by every `Field`/`Index` step in both functions.

## Repro (verified against pinned jq 1.7.1)

```
$ echo null | jq -c '.a[]? = 9'
null

$ echo null | succinctly jq -c '.a[]? = 9'
{"a":null}

$ echo '[1,2,3]' | jq -c '.[5][]?[0] = 9'
[1,2,3]

$ echo '[1,2,3]' | succinctly jq -c '.[5][]?[0] = 9'
[1,2,3,null,null,null]

$ echo null | jq -c '.a.b[]?[0] = 9'
null

$ echo null | succinctly jq -c '.a.b[]?[0] = 9'
{"a":{"b":null}}
```

## Discovery context

Found live during #1298's code review (multiple independent verification passes). Confirmed present on unmodified `main` (commit `9bc4551d4`, immediately before #1298) via the terminal `Expr::Iterate` arm — not a regression from #1298, and not fixed by it. #1298's own new mid-chain `Expr::Iterate` handling (`split_at_iterate`/`set_path_through_iterate`) faithfully mirrors the terminal arm's existing behavior, so it inherits the same defect for the new expression shapes it makes reachable (`.a[]?[0] = 9`, `.[5][]?[0] = 9`, `.a.b[]?[0] = 9`), widening where the bug shows up without introducing a new one.

## Scope note

[...]

PR #1548, opened 2026-08-23
refactor(jq): route Field/Index/Iterate/StringInterpolation onto shared continuation helpers

newhoggy:
## Summary

A follow-up to the PR #1475 consolidation (#1449): that PR converted five early-return blocks plus the `Array` arm in `eval_pipe_with_path_context_internal` onto `continue_rest_with_context`/`continue_rest_with_fresh_root`, but left more sites with the identical hand-rolled "if `rest.is_empty()` { return } else { recurse }" shape unconverted — found during that PR's own review, deliberately scoped out at the time to keep it narrow.

Converts the four sites that shape covers cleanly:

- **`Field`** — the object-navigation arm's manual `rest.is_empty()` branch becomes a single `continue_rest_with_context` call.
- **`Index`/`IndexNumber`** — same shape, one level down from `Field`.
- **`Iterate`** — each array/object element gets its own distinct `new_path`, so `continue_rest_with_context` is called once per element inside the loop (its single-value signature has nothing to batch a per-element path over) rather than once for the whole batch — the accumulate-via-`accumulate_path_context_step` loop shape is otherwise unchanged.
- **`StringInterpolation`'s yq-mode branch** — was still hand-rolling the identical "fresh root, empty path" reset its jq-mode sibling already uses via `continue_rest_with_fresh_root` a few lines below.

Pure consolidation, no behavior change.

## Scope note

The tracking issue's own review comment (from a later PR) also flagged some further, separate consolidation opportunities in the same function — `Iterate`'s own array/object branches still duplicating the loop body twice, a `Builtin::Map` arm with a third near-copy of the same shape, and `eval_fanout`/`accumulate_path_context_step` being two independently-implemented "accumulate-or-stop" folds living side by side. None of those are touched here; this PR is scoped to exactly the four sites the issue's original body named. Leaving the tracking issue open, narrowed to the remaining items, rather than closing it.

## Test plan

- [x] Differential-tested ~30 queries covering `Field`/`Index`/`Iterate`/`StringInterpolation` under path context (via `key`/`parent`) in both jq and yq mode, against `main`'s own release binary — byte-identical output in every case, including existing error/optional-suppression paths and control-flow interactions (`limit`, `select`, nested `Iterate`, the known `#1428` autovivify quirk unaffected)
- [x] `cargo test --features cli,simd,regex,serde` — full suite, 0 failures (876/876 `jq_cli_tests`)
- [x] No new tests needed — this changes no observable behavior, verified by the differential pass above

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -22916,12 +22916,9 @@ fn eval_pipe_with_path_context_internal<'a, W: Clone + AsRef<[u64]>, S: EvalSema
             // Get the field value
             if let OwnedValue::Object(entries) = value {
                 if let Some(v) = entries.get(name) {
-                    if rest.is_empty() {
-                        return QueryResult::Owned(v.clone());
-                    }
-                    return eval_pipe_with_path_context_internal::<W, S>(
+                    return continue_rest_with_context::<W, S>(
+                        QueryResult::Owned(v.clone()),
                         rest,
-                        v,
                         root,
                         file_origin,
                         &new_path,
@@ -22957,12 +22954,9 @@ fn eval_pipe_with_path_context_internal<'a, W: Clone + AsRef<[u64]>, S: EvalSema
                 let actual_idx = if *idx < 0 { len + *idx } else { *idx };
                 if actual_idx >= 0 && (actual_idx as usize) < arr.len() {
                     let v = &arr[actual_idx as usize];
-                    if rest.is_empty() {
-                        return QueryResult::Owned(v.clone());
-                    }
-                    return eval_pipe_with_path_context_internal::<W, S>(
+                    return continue_rest_with_context::<W, S>(
+                        QueryResult::Owned(v.clone()),
                         rest,
-                        v,
                         root,
                         file_origin,
                         &new_path,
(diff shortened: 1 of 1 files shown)


newhoggy:
\`/code-review\` (10-agent fan-out) came back with three independent angles converging on the same real finding, now fixed:

**Deep-clone regression on the continuing path**: \`continue_rest_with_context\` takes \`intermediate: QueryResult\` by value, so routing \`Field\`/\`Index\`/\`Iterate\` through it required building \`QueryResult::Owned(v.clone())\` unconditionally, *before* \`rest.is_empty()\` is even checked — turning the old code's zero-clone continuing path (\`v: &OwnedValue\` passed straight into the recursive call whenever \`rest\` was non-empty, cloning only at the terminal leaf) into an unconditional deep clone at every intermediate step. `OwnedValue::clone()` is a real recursive deep clone (no `Rc`/`Arc` sharing), so a chained path like `.a.b.c.d.e` went from one clone total to one clone per level, and `.[] | ...` over a large array/object went from cloning only what `rest` actually reaches to cloning every element regardless — exactly the "real, measurable cost" #1445 already fixed for the terminal case, reintroduced on the continuing case.

Fixed by adding `continue_rest_with_borrowed_value`, a sibling helper that takes `v: &OwnedValue` instead of an already-owned `QueryResult` — same `rest.is_empty()`-or-recurse decision, but deferred until after checking `rest`, so the continuing case keeps the caller's existing borrow with zero clones and only the unavoidable terminal-leaf clone remains. `Field`/`Index`/`Iterate` now use this instead.

Re-verified: the full 30-query differential battery is still byte-identical against `main`'s binary, and the full test suite/both clippy legs/fmt/doc/no_std build are all still clean.

(The `StringInterpolation` yq-branch conversion was independently confirmed genuinely zero-cost by every angle that looked at it — the value there is always a freshly-constructed local with nothing to borrow from — so it's unchanged.)

Issue #1371, opened 2026-08-21
jq: self-recursive user-defined functions can't work at any real depth under static AST substitution


newhoggy:
## Summary

Follow-up to #1016 (which adds a depth guard so a self-recursive `def` fails cleanly instead of crashing with SIGABRT). This issue tracks the deeper, architectural gap that guard deliberately does not fix: **self-recursive `def`s cannot work at any depth a real user would find useful**, under `succinctly jq`'s current evaluation strategy.

`succinctly jq` evaluates `def name(params): body; then` by **statically substituting** each call to `name` with `body` (with `params` replaced by the call's arguments) — entirely before any evaluation happens (`expand_func_calls`, `src/jq/eval.rs`). For a self-recursive `def`, the substituted body always contains another syntactic call to `name`, since expansion can't observe that a runtime condition (`n == 0`) will eventually hold. #1016 bounds this with a depth guard (`MAX_FUNC_EXPANSION_DEPTH`) so it errors cleanly instead of crashing — but that guard's ceiling had to be set empirically very low (50, with the actual crash boundary around ~189-383 depending on `def` shape, measured in a debug build with a default thread stack) because **each unrolling re-substitutes the entire growing argument expression from the level before into a fresh clone of `body`** — the AST being walked keeps growing with depth, unlike a real interpreter's call stack.

## Why this matters beyond #1016's crash fix

Real jq evaluates function calls at runtime — one interpreter frame per call, not a growing AST substitution — so a self-recursive `def` there works up to jq's own memory limits (confirmed live: this issue's own `deep(n)` repro returns cleanly through at least `n = 10,000` against jq 1.7.1). Any real-world recursive `def` a user might write — list traversal, tree walks, accumulator patterns — could easily need more than 50 levels for realistic inputs (e.g. recursively processing a 100-element list). Under #1016's guard, all of these now get a clean "recursion too deep" error where real jq would simply succeed. The guard is strictly better than the SIGABRT it replaces, but it is *not* jq-compatible for ordinary recursive-`def` usage — only for the narrow slice of recursion depths below the guard's ceiling.

## Repro

```
$ echo null | succinctly jq -c 'def sum_to(n): if n == 0 then 0 else n + sum_to(n-1) end; sum_to(100)'
# now: clean "recursion too deep" error (post-#1016), where real jq returns 5050
$ echo null | jq -c 'def sum_to(n): if n == 0 then 0 else n + sum_to(n-1) end; sum_to(100)'
5050
```

[...]

PR #1551, opened 2026-08-23
refactor(jq): consolidate three builtin AST traversals onto map_builtin_subexprs

### LINKED ISSUE #1506: refactor(jq): four hand-maintained 207-arm Builtin traversals should share one accessor

newhoggy:
## Summary

`substitute_var_in_builtin`, `expand_func_calls_in_builtin`, and
`substitute_func_param_in_builtin` (`src/jq/eval.rs`) each hand-maintained a
full ~207-arm exhaustive `Builtin` match with no wildcard, independently
restating which sub-expressions to recurse into for every variant. #1309
already showed this exact pattern drifts silently in practice
(`contains_split_doc` was missing twelve variants). #1505 built
`map_builtin_subexprs` — the mapping twin of `builtin_kids` — for exactly
this shape, and proved it on `rewrite_namespaced_calls`'s one call site.

Verified before starting that all three target functions are pure
structural recursion: every arm forwards its sub-expression(s) unchanged to
one fixed transform function with the same extra arguments every time, no
per-variant special-casing anywhere in any of the three. Each therefore
collapses to a thin wrapper around `map_builtin_subexprs`:

```rust
fn substitute_func_param_in_builtin(builtin: &Builtin, param: &str, arg: &Expr) -> Builtin {
    map_builtin_subexprs(builtin, &mut |e| substitute_func_param(e, param, arg))
}
```

~2200 lines become ~25. `builtin_kids` is now the only surviving
hand-maintained exhaustive `Builtin` traversal in the codebase.

Also updated `map_builtin_subexprs`'s own doc comment (added by #1505),
which had said consolidation was "filed separately, #1506, not attempted
here" — corrected now that it's landed, and noted that unlike
`rewrite_namespaced_calls` (parse-time, once per program),
`substitute_var_in_builtin`/`expand_func_calls_in_builtin` sit on the
AST-substitution path this codebase uses in place of a runtime variable
environment (#1371), so they can re-run per binding evaluation rather than
once per parse.

No behavior change intended — pure refactor. Full test suite passes
unchanged.

Fixes #1506

## Test plan

- [x] `cargo test --features cli,simd,regex,serde --no-fail-fast` — 0 failures across every group, doc-tests included
- [x] `cargo fmt --check`


newhoggy:
Heads up on the Coverage bot comment above: the **85.02% patch coverage** figure and its ~2200-line "uncovered new lines" list are a diff-attribution artifact, not a real gap — filed as #1554 with full detail.

The actual diff here is ~25 new lines (confirmed via \`git diff --diff-algorithm=histogram --stat\`: \`8 insertions(+), 2188 deletions(-)\`; the default algorithm on this same diff reports \`11588 insertions(+), 13768 deletions(-)\` for identical content, due to the file's hundreds of near-identical match arms confusing the diff). \`omni-dev coverage diff\` hits the same misalignment independently.

Verified directly against the lcov report's raw \`DA:\` records: every genuinely new line (the three wrapper functions' bodies) has 1000+ hits — real coverage is fine.

newhoggy:
Ran a full multi-angle review (8 finder agents) before merging. Summary:

**Correctness (6 angles, all empty findings):** removed-behavior audit, reuse/duplication, CLAUDE.md/STYLE_GUIDE conventions, cross-file consistency, general Rust pitfalls, and wrapper/proxy correctness all confirmed the consolidation is behavior-preserving — every one of the 207 `Builtin` variants across all three deleted functions was cross-checked programmatically for field order/count/call-signature against `map_builtin_subexprs`, with zero mismatches.

**Simplification angle (3 minor findings):** suggested inlining the three thin wrappers, and merging two doc-comment paragraphs. Considered and left as-is — the wrapper names are referenced by existing tests/comments across the codebase (confirmed by the reuse angle), and further doc-comment merging risked reducing readability rather than improving it.

**Altitude + Efficiency angles (the substantive one):** two independent agents converged on a real concern — two of the three newly-consolidated callers (`substitute_var_in_builtin`, `expand_func_calls_in_builtin`) sit on genuinely hot per-element AST-substitution paths, not just parse-time, and the Efficiency angle confirmed via release-build disassembly that the no-sub-expression bucket's `builtin.clone()` compiles to a second dispatch through the derived `Clone` impl rather than the old code's zero-cost direct construction.

Addressed by actually measuring rather than assuming, per this project's benchmarking discipline: built clean before/after binaries from the exact PR base commit, ran an interleaved A/B across three chain-length/array-size configurations (`.[] as $x | $x | length | ... | length`, 20/50/100-deep chains over 150k-500k elements) specifically designed to maximize exposure to this path. No consistent, reproducible regression — results flip sign between runs at ~1-2% magnitude, consistent with noise on a heavily loaded shared machine (this repo currently has ~40 concurrent sessions active) rather than a real effect. Documented the full tradeoff, the measurement, and its limitation directly in `map_builtin_subexprs`'s doc comment (556ed3673), including the narrow fix available (give the no-sub-expression bucket its own per-variant arms) if a real regression ever surfaces under properly isolated measurement on dedicated hardware.

Also added the direct unit test the review found missing: `map_builtin_subexprs` had zero tests of its own, only transitive coverage. The new test round-trips real filter syntax with distinct-valued fields through an identity transform and asserts full equality — verified it actually catches a field-order swap by injecting one (a first version using a call-order-coded marker didn't catch it, since call order and placement order matched trivially regardless of the swap; caught and fixed before landing).

Full test suite green (0 failures, 54/54 groups), clippy x2, no_std, fmt, doc all clean.

Issue #1514, opened 2026-08-23
perf(jq): #1385's duplicate-key detector costs 2-3x on object-heavy input with no duplicate keys


newhoggy:
**Severity: High** (no wrong output — a 2-3x jq slowdown on object-heavy documents, on `main` today)

## Summary

#1385 shipped as a commit series on 2026-08-22. **Two commits in it each introduced a distinct
regression, on different paths**, together making `succinctly jq` up to **3x slower** on
object-heavy JSON. The cost is paid by the new duplicate-key **detector**, on documents that
contain **no duplicate keys at all**.

| Commit | What | Regresses |
|---|---|---|
| `234bd317` | *fix(jq): collapse duplicate object keys in the **evaluator*** | `keys_unsorted` / `select` / `map` — up to **+204%** |
| `bcd6d685` | *fix(jq): collapse duplicate object keys on the **print path*** | plain identity `.` — up to **+44%** |

The two barely overlap: the evaluator commit leaves `.` almost untouched (+3-7%), and the
print-path commit slightly *helps* `keys_unsorted` (-3.5%) while hitting `.`. Each was found by
its own bisect, because a single signal only reveals one of them.

Found while verifying #789 on the pinned x86_64 bench box (see
[#1283's re-sweep](https://github.com/rust-works/succinctly/issues/1283#issuecomment-5383446391)
for how it surfaced). Output is unchanged throughout — every measurement below passed the
identity gate with 0 differences, so this is timing-only.

## Bisects

**Bisect 1** — `git bisect run` over the 999 commits in `6ce12be6..53f57d8c1`, signal =
`succinctly jq 'keys_unsorted' wide/10mb.json`, median of 5:

```
40c5ff93  docs(jq): design jq-mode duplicate object key collapse (#1385)     85ms   good
234bd317  fix(jq):  collapse duplicate object keys in the evaluator (#1385)  260ms  BAD
```

Every probe before it 82-87ms; every probe after 182-260ms. No skips, no ambiguity.

**Bisect 2** — the first bisect's signal did *not* explain the identity regression (`.` was only
+3-7% at `234bd317` but +38-47% on `main`), so a second run over the 38 commits in
`234bd317..53f57d8c1`, signal = `succinctly jq '.' wide/10mb.json`:

```
234bd317  fix(jq): collapse duplicate object keys in the evaluator (#1385)   175ms  good
bcd6d685  fix(jq): collapse duplicate object keys on the print path (#1385)  249ms  BAD
[...]

Issue #1385, opened 2026-08-21
jq: duplicate JSON object keys survive to output, where real jq collapses them last-wins


newhoggy:
**Severity: Medium** (silently wrong output on the simplest possible query; but see "Is this
deliberate?" below — this may be a documentation gap rather than a code bug)

## Summary

In **jq mode**, `succinctly jq` emits duplicate JSON object keys verbatim where real jq
collapses them last-wins at parse time. Reachable from the identity filter:

```
$ echo '{"a":1,"a":2}' | jq -c '.'
{"a":2}
$ echo '{"a":1,"a":2}' | succinctly jq -c '.'
{"a":1,"a":2}
```

Verified at `main` @ `cedee4cbb` against pinned jq 1.7.1 (`/usr/bin/jq`).

The same divergence reaches every cursor-preserving navigation path:

```
$ echo '[{"a":1,"a":2}]' | jq -c '.[]'          {"a":2}
$ echo '[{"a":1,"a":2}]' | succinctly jq -c '.[]'    {"a":1,"a":2}

  .[0]          jq: {"a":2}   succinctly: {"a":1,"a":2}
  first(.[])    jq: {"a":2}   succinctly: {"a":1,"a":2}
  last(.[])     jq: {"a":2}   succinctly: {"a":1,"a":2}
```

It disappears the moment anything materializes through `to_owned` — which is why it is
easy to miss:

```
$ echo '[{"a":1,"a":2}]' | succinctly jq -c 'first(.[], 9)'    {"a":2}    <- agrees with jq
```

## Why this is worth filing as a root cause

Three issues in this family have already been fixed **as individual symptoms**, each
patching one consumer rather than the shared source:

- **#1251** — `.foo` on a duplicate key returned the first value, not jq's last-value-wins.
- **#1170** — `to_entries` showed duplicate keys as separate entries instead of collapsing.
- **#443** — `to_entries` and `-o=json` output collapsed duplicate mapping keys.

Each fix moved one consumer toward jq's last-wins model. The *producer* still hands out
both keys, so any consumer that hasn't been individually patched still diverges. #1342,
#1343 and #1344 are the yq-side continuation of the same pattern — five issues and
counting, each closing one call site.

## Is this deliberate?

Possibly, and that is the actual question for triage. **#442** and **#478** deliberately
changed JSON *output* to **preserve** duplicate mapping keys (pretty-printing and
`--slurp`/`--inplace` were collapsing them, and that was treated as the bug). That is
[...]

Issue #789, opened 2026-08-11
jq: select(true) is ~5-8% slower on x86_64 after #740's LazySeq fast path


newhoggy:
## Summary

While running the required A/B measurement for #758 (lazy `map`/`select` fast path, PR #740), `select(true)` was found to be **consistently slower on x86_64** after #740 merged, while Apple Silicon shows no such effect. The regression is small but real — confirmed against both `HEAD` and the isolated `6ce12be6` merge commit, well outside the machine's control-run noise floor.

## Data

Interleaved A/B (7 reps), before = `588a4c5d` (parent of #740's merge), after = `6ce12be6` (#740's merge commit itself). Full context and all tables: `docs/plan/jq-lazy-map-select.md` § "Measurement results (2026-08-11, #758)".

| Machine | Query | Δ median (range across 100kb-100mb) |
|---|---|---|
| Apple M4 Pro | `keys_unsorted \| select(true)` | flat, within ±2.7% control band |
| AMD Ryzen 9 7950X | `keys_unsorted \| select(true)` | **+6.0% to +6.8%**, all 4 sizes |
| Apple M4 Pro | `select(true)` (plain container) | flat, within ±2.7% control band |
| AMD Ryzen 9 7950X | `select(true)` (plain container) | **+0.7% to +2.5%**, all 4 sizes |

Control noise floor: Apple M4 Pro −2.7%..+1.0%, AMD Ryzen 9 7950X −1.7%..+0.9% (established via `ab-cli.py --control` before any comparison).

## Why this looks real, not noise

- **Flat across sizes** (not growing/shrinking) — every `map(.)` result in the same measurement run shows a clear scaling curve; `select(true)`'s delta doesn't move with input size. That shape is the signature of a small per-call constant-factor cost, not an algorithmic regression.
- **Reproducible**: identical magnitude measured independently against current `HEAD` and against the isolated `6ce12be6` commit — rules out any of the ~10 `src/jq` commits merged after #740 as the cause.
- **Platform-specific**: present only on AMD Ryzen 9 7950X, absent on Apple M4 Pro, consistent with this repo's prior experience (#106) that small constant-factor effects don't always port between architectures.
- **Output identity confirmed unchanged** (0 differences across all configurations, both `ab-cli.py`'s before/after gate and `dev bench jq`'s gate against system `jq`) — this is a timing-only effect, not a correctness change.

## Hypothesis

[...]

PR #1555, opened 2026-08-23
ci(bench): add an instruction-count regression guard for jq/yq

### LINKED ISSUE #1523: ci(bench): no guard catches a jq/yq performance regression — CI benchmarks only rank_select and never compares
**Severity: Medium** (process gap — no wrong output; it is why a 2-3x regression shipped unnoticed)

## Summary

Nothing in CI would catch a jq/yq performance regression. #1514 is the demonstration: `#1385`'s
two fix commits made `succinctly jq` **2-3x slower** on object-heavy input, landed on `main`, and
were found only because someone re-verified an unrelated, three-times-deferred perf issue (#789)
on the pinned x86_64 bench box a day later.

Worse, the regression **survived a perf commit aimed squarely at it** — `b6c0c3ca`,
*"perf(jq): cut the duplicate-key probe from 10.4% to 3.9% (#1385)"*. "10.4% to 3.9%" is a
share-of-runtime figure measured inside an already-inflated total, not a delta against the
pre-series parent, so it read as a win while the 2-3x regression stayed in place.

## Current state

`ci.yml`'s `bench` job (3-way matrix: x86_64, ARM64-Linux, ARM64-macOS) runs **only**
`cargo bench --bench rank_select` in three feature configurations and **uploads the results as
artifacts**. It performs no comparison against any baseline and has no failure condition.

So today:
- **No jq or yq CLI workload is benchmarked in CI at all** — `rank_select` is the bitvector
  microbenchmark. The entire `src/jq` evaluator, both `eval.rs` and `eval_generic.rs`, is uncovered.
- Even for `rank_select`, nothing compares run N against run N-1. The artifacts are written and
  never read.

## Why wall-clock in CI is the wrong mechanism

The obvious implementation — time the binary, fail on >X% drift — will be flaky. GitHub's shared
runners are noisy neighbours, and this repo already measured a **-4.2%..+1.1% control band** on a
*dedicated, idle* 7950X (see #1514). A threshold loose enough not to flake on a shared runner
(~25%?) is loose enough to let real regressions through, and #1385's print-path commit (+36-44%)
would have been uncomfortably close to it.

**Use instruction counts instead.** `valgrind --tool=cachegrind` is binary instrumentation, not
hardware counters — it is deterministic, needs no idle machine, and a 2-3x work increase shows up
as a 2-3x instruction-count increase regardless of what else the runner is doing. A 5% threshold
is then meaningful rather than aspirational.

`scripts/perf-ab.py` **already defaults to `--tool cachegrind`** and already handles a rootless
valgrind build (`--valgrind-bin`), so the measurement half largely exists.

## What it must cover, and the trap to avoid

#1514's two regressions are the specification, because they are complementary and each is
invisible to the other's signal:

| Commit | Signal that catches it | Effect on the *other* signal |
|---|---|---|
| `234bd317` (evaluator) | `keys_unsorted` | `.` only +3-7% |
| `bcd6d685` (print path) | `.` | `keys_unsorted` **-3.5%** (looks like an improvement) |

So a guard with one query would have caught one of the two and — worse — the print-path commit
would have *improved* the `keys_unsorted` number, actively arguing it was fine.

Minimum viable set:
- **Input shape matters as much as the query.** `wide` (one object, ~82k keys/MB) shows +94%;
  `users` (6-key records) +7.8%; `arrays` (no objects) -2.6%. An object-cost regression is
  invisible in `users`/`arrays`. Generate `wide` explicitly — `succinctly json generate <n> -p wide`.
- **Both `keys_unsorted` and `.`**, per the table above.
- **At least one yq shape**, since `succinctly yq` shares the evaluator and its cost is currently
  unverified even for #1385 (see #1514's verification section).
- Checked-in baseline instruction counts, refreshed deliberately with a stated reason — never
  auto-updated, or the guard silently ratchets.

## The rule this encodes

CLAUDE.md already records the lesson twice, from opposite directions:

> **a benchmark claim is only as fresh as its baseline** — a clean, conflict-free rebase can
> silently improve the "control" side and flip a documented win into a loss with nothing to flag it.

> a precheck that enables a fast path is charged to *every* input, including the ones it cannot
> help — measure it against the workload where the parser is already fastest.

Both were learned by hand, after the fact. A guard is how they stop depending on someone
remembering.

## Notes / non-goals

- **Not** a replacement for the A/B discipline in `docs/guides/benchmarking.md` — that stays the
  method for *evaluating a change*. This is a tripwire for changes nobody thought to measure.
- Should be a **non-required** check initially (like `bench` and `coverage` today), so a flaky
  first iteration cannot block the merge queue. Promote it once it has a track record.
- #98's distributed orchestration (`bench nodes`/`orchestrate`) is the heavier alternative if
  in-CI instrumentation proves unworkable; it needs the pinned boxes and is not merge-gated.

## Related

- #1514 — the regression this would have caught, with both bisects and the full A/B data
- #789 — the perf issue whose deferred verification found it
- #1283 — the sequencing plan whose re-sweep surfaced the whole chain

newhoggy:
## Summary

Nothing in CI catches a jq/yq performance regression. #1385's two fix commits made `succinctly jq` **2-3x slower** on object-heavy input, landed on `main`, and were found only because someone happened to re-verify an unrelated, already-deferred perf issue (#789) on a pinned bench box a day later. The regression survived a perf commit aimed squarely at it — `"10.4% to 3.9%"` was a share-of-runtime figure measured inside an already-inflated total, not a delta against the pre-series parent, so it read as a win while the 2-3x regression stayed in place.

The existing `bench` CI job only times `rank_select` and uploads the results as artifacts nobody reads — no jq/yq CLI workload is benchmarked in CI at all, and even `rank_select` has no comparison against any baseline.

## Why cachegrind, not wall-clock

Wall-clock timing on a shared CI runner is the wrong mechanism: this repo already measured a **-4.2%..+1.1% control band** on a *dedicated, idle* 7950X, so a threshold loose enough not to flake on a noisy shared runner would also be loose enough to let a real 2-3x regression through. Uses `valgrind --tool=cachegrind` instead: binary instruction counts are deterministic regardless of what else the runner is doing, so a 2-3x work increase shows up as a 2-3x instruction-count increase and a 5% threshold is meaningful rather than aspirational.

## What's covered

Adds `scripts/perf-guard.py` (`--check` / `--update-baseline`) and a new **Perf Regression Guard** CI job (x86_64 and ARM64-Linux only — valgrind has no Apple Silicon support), checking the fixed query/shape matrix the issue itself specified as the minimum viable set: #1514's two regressions were complementary and each was invisible to the other's own signal (one showed only under a `wide`-shaped `keys_unsorted` query, the other only under a plain identity query and would have looked like an *improvement* under `keys_unsorted` alone). Covers `wide`/`users`/`arrays` shapes, both `keys_unsorted` and identity queries, and one yq-mode query (the shared evaluator's cost is otherwise unverified in yq mode at all).

`tests/data/perf-guard-baseline.json` is keyed **per architecture** — instruction counts depend on ISA/codegen, not just wall-clock noise — and was generated from this PR's own CI runs (x86_64 and ARM64-Linux GitHub Actions runners themselves), not a different machine, so the checking environment and the baseline's own origin match exactly.

Deliberately **not** added to the repo's required-checks list yet, per the issue's own ask — a flaky first iteration of a new gate must not be able to block the merge queue. Promote it once it has a track record.

## Test plan

- [x] Verified the script's control flow end-to-end locally with a fake-valgrind shim (no real valgrind available on this dev machine — Apple Silicon) — `--update-baseline`, `--check` (pass and fail paths), missing-baseline-entries fail-fast, missing-binary, missing-valgrind
- [x] Bootstrapped the real baseline from this PR's own CI runs: ran `--update-baseline` on both `x86_64` and `ARM64-Linux` GitHub Actions runners, downloaded the resulting artifacts, merged them into the committed baseline file
- [x] Confirmed the final `--check`-mode CI job passes against that freshly-generated baseline (both legs green)
- [x] `python3 -m py_compile scripts/perf-guard.py` — clean
- [x] YAML validated (`yaml.safe_load`)
- [x] Updated `docs/guides/benchmarking.md`'s CI/CD section to document the new job and how to update the baseline

Closes #1523

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -486,6 +486,53 @@ jobs:
           name: benchmark-results-${{ matrix.arch }}
           path: bench-*.txt
 
+  # Instruction-count regression guard (#1523): `bench` above times
+  # `rank_select` and uploads artifacts nobody compares against a baseline --
+  # #1385's 2-3x jq slowdown shipped through it unnoticed. Wall-clock timing
+  # on a shared runner is the wrong mechanism (too noisy for a threshold
+  # tight enough to matter); `valgrind --tool=cachegrind` counts instructions
+  # instead, which is deterministic regardless of runner noise. Not run on
+  # macOS: valgrind has no Apple Silicon support at all. Not run in the merge
+  # queue for the same reason `bench` isn't (see its own `if`, above).
+  perf-guard:
+    name: Perf Regression Guard (${{ matrix.name }})
+    needs: gate
+    if: needs.gate.outputs.code == 'true' && github.event_name != 'merge_group'
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: x86_64
+            runner: ubuntu-latest
+          - name: ARM64-Linux
+            runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Cache cargo
+        uses: actions/cache@v6
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-${{ matrix.name }}-perf-guard-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Install valgrind
+        run: sudo apt-get update && sudo apt-get install -y valgrind
+
+      - name: Build release binary
+        run: cargo build --release --features cli
+
+      - name: Run perf guard
+        run: python3 scripts/perf-guard.py --binary target/release/succinctly --arch "${{ matrix.name }}" --check
+
   clippy:
     name: Clippy
     needs: gate
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,8 +64,8 @@ jobs:
   # Required-context jobs (Test x86/ARM/macOS, Clippy, Format, Documentation)
   # always run so their contexts always report `success`; on the fast path every
   # step is skipped, which is a genuine passing job — not a `skipped` conclusion
-  # whose branch-protection semantics vary. Non-required jobs (bench, coverage)
-  # below skip wholesale via a job-level `if`.
+  # whose branch-protection semantics vary. Non-required jobs (bench, coverage,
+  # perf-guard) below skip wholesale via a job-level `if`.
   test-x86:
     name: Test (x86_64)
     needs: gate
@@ -362,7 +362,7 @@ jobs:
   # above stays on `stable` so we don't triple CI minutes. `beta` gates CI;
   # `nightly` is advisory (`continue-on-error`) because it is the flakiest. Not a
   # required status context, so it skips wholesale on the docs-only fast path
-  # (job-level `if`, like bench/coverage) rather than run-and-skip-steps. Also
+  # (job-level `if`, like bench/coverage/perf-guard) rather than run-and-skip-steps. Also
   # skipped on `merge_group` (#310): it is advisory, not required, so running it
   # per queue entry would just add latency to every merge without gating it.
   test-x86-upstream:
--- a/docs/guides/benchmarking.md
+++ b/docs/guides/benchmarking.md
@@ -159,6 +159,24 @@ GitHub Actions runs `rank_select` benchmarks on every PR/push for smoke testing:
 
 **Note**: CI does not run full benchmark suite (too time-consuming).
 
+**Perf Regression Guard** (issue #1523): a separate CI job measures `succinctly
+jq`/`yq` instruction counts via `valgrind --tool=cachegrind` (`scripts/perf-guard.py`)
+for a fixed query/shape matrix, and fails if any drifts more than 5% from the
+checked-in baseline (`tests/data/perf-guard-baseline.json`). Deterministic
+instruction counts, not wall-clock time, are what make a tight threshold viable on a
+shared, noisy CI runner — this exists because #1385's 2-3x regression shipped through
+`rank_select`-only smoke testing unnoticed. Runs on x86_64 and ARM64-Linux only
+(valgrind has no Apple Silicon support); currently a non-required check while it
+builds a track record. To deliberately update the baseline after a real, understood
+cost change:
+
+```bash
+cargo build --release --features cli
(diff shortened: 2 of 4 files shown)


newhoggy:
\`/code-review\` (10-agent fan-out) found a real regression I'd introduced mid-refactor (a stale call site left over from restructuring fixture generation — caught and fixed before it ever reached this comment) plus a long list of genuine robustness gaps in the first version of \`scripts/perf-guard.py\`. All addressed in the follow-up commit:

- **Real bugs**: cachegrind log opened before checking the child's exit code (uncaught `FileNotFoundError` on a valgrind crash); `generate_fixture`'s captured stderr never surfaced on failure; no `LC_ALL=C` pin (a runner locale change could silently truncate the instruction-count regex match); `statistics.median`'s int/float flip depending on `--reps` parity; no executable-bit check; no lower bound on `--reps` (0 crashed on an empty list); a query that silently starts producing no output (a correctness bug) would look exactly like a legitimate speedup since only the exit code was checked; a corrupted baseline JSON crashed with a raw traceback instead of a clear message.
- **Efficiency**: 5 of 6 `QUERIES` rows shared a `(pattern, size)` fixture that was being regenerated per-row; `measure_all` now generates each distinct shape once.
- **Documentation**: two scope limits (single fixed size per query, not a scaling curve; `--features cli` only, not `simd`, so a regression confined to `BalancedParens`'s SIMD-accelerated index build or explicit popcount intrinsics is out of this guard's reach — the same split `bench`'s own default/simd/portable-popcount matrix exists to separate) are now documented in the script's own docstring rather than silently invisible. Two pre-existing `ci.yml` comments enumerating "non-required jobs (bench, coverage)" were updated to include `perf-guard`.
- **Conventions**: `argparse`'s `epilog` was a fragile slice of `__doc__`; now a plain string constant. `measure_query`/`run_cachegrind_once` had asymmetric parameter-passing conventions; unified on explicit args.

Not addressed, deliberately: consolidating the cachegrind-invocation logic with `scripts/perf-ab.py`'s own (a similar but harness-specific tool) — judged a larger, separate refactor for marginal DRY benefit given both are small, independent, standalone scripts by design; a floating (unpinned) Rust toolchain between when a baseline is captured and later CI runs — this is a pre-existing repo-wide convention every other CI job also has, not something specific to this PR to fix unilaterally; and expanding the query matrix to multiple sizes or a `simd` build variant — real, valuable follow-ups but beyond this issue's own "minimum viable set" framing, better done once this guard has a track record (as the issue itself suggests for promoting it to required).

Re-verified: full local test battery (11 cases covering every fix) all pass against a fake-valgrind shim (no real valgrind on this dev machine — Apple Silicon has no valgrind support at all, which is also why the CI job itself is Linux-only).

Issue #1520, opened 2026-08-23
jq: --slurp reports the last value's location, not the last file's EOF position


newhoggy:
Found by a 1155-combination differential sweep run while finishing #1309.
**Pre-existing and independent of that work** — it reproduces with a plain
`error("x")` and no input builtin anywhere, so it never touches #1309's queue.

## The rule, pinned against jq 1.7.1

Under `--slurp` jq names **the last source on the command line, at that
source's own newline count** — not the last *value*'s location:

```
$ jq -c -s 'error("x")' a.json b.json          # a=1nl b=1nl   -> b.json:1
$ jq -c -s 'error("x")' a.json e.json          # e is empty    -> e.json:0
$ jq -c -s 'error("x")' e.json                 #               -> e.json:0
$ jq -c -s 'error("x")' nonl.json              # "1 2", no NL  -> nonl.json:0
$ jq -c -s 'error("x")' three.json             # 3 newlines    -> three.json:3
$ jq -c -s 'error("x")' a.json b.json three.json           -> three.json:3
```

## succinctly

`InputLocations::single(locations.last())` (`src/bin/succinctly/jq_runner.rs`)
uses the last *value*'s location, which coincides with jq only when the last
file happens to contribute the last value:

| argv | jq | succinctly |
|---|---|---|
| `-s 'error("x")' a.json b.json` | `b.json:1` | `b.json:1` ✅ |
| `-s 'error("x")' a.json e.json` | `e.json:0` | **`a.json:1`** ❌ wrong file |
| `-s 'error("x")' e.json` | `e.json:0` | **`<stdin>:0`** ❌ loses the filename |
| `-R -s 'error("x")' a.json b.json` | `b.json:1` | **`a.json:2`** ❌ wrong file and line |

Naming the wrong *file* is the bad part — it is a confidently wrong answer, not
a vague one.

## Why it surfaced now

#1309 made slurp locations reachable through a second path (the shared input
queue), so the same underlying value shows up in `input`/`inputs` error markers
too. That PR fixed the *exhaustion* side for slurp (`<unknown>`, #1508) but did
not touch `single()`, which is where this lives. 13 of the 34 divergences in
the sweep trace to it.

## Suggested fix

`InputLocations::single()` should take the EOF position rather than
`locations.last()`. That is very nearly the computation `exhausted()` already
does for the non-slurp case — "last source, at its own line count" — so the two
probably want to share one helper. Watch the `-R -s` case: there is exactly one
[...]

Issue #1542, opened 2026-08-23
jq: --seq -s with a truncated final record reports a location instead of <unknown>


newhoggy:
**Severity: Low** — a narrow, uncommon input shape (`--seq` combined with `--slurp`, and the last
RS-delimited record itself truncated/incomplete).

## Summary

Found live-testing during #1520's review, confirmed pre-existing (reproduces identically on `main`
before that fix — not introduced or worsened by it).

## Repro (verified live against pinned jq 1.7.1)

A `--seq` (RFC 7464) input whose final record is truncated (no closing content after the last RS
byte, or an incomplete JSON value) reports:

```
$ jq --seq -s 'error("x")' <truncated-seq-file>
jq: error (at <unknown>): x

$ succinctly jq --seq -s 'error("x")' <truncated-seq-file>
jq: error (at file:1): x
```

Real jq falls back to `<unknown>` for this shape; succinctly names a concrete file:line.

## Why this is separate from #1520

#1520 fixed `--slurp`'s EOF-location computation for the ordinary JSON/raw-input cases (last file,
last file's own newline count). This is a different mechanism — `--seq`'s own record-parsing
(`parse_json_seq`/`json_seq_ends`) evidently produces a location for a shape jq itself treats as
having none, independent of which "last source" rule feeds `InputLocations::slurp_eof`.

## Suggested fix direction

Not investigated. Likely needs `parse_json_seq`/`json_seq_ends`'s handling of a truncated trailing
record traced against jq's own `--seq` parser to find where jq decides "no location" for this case
that succinctly's parser doesn't.

newhoggy:
Moving to `Ready` — landed in `Todo` via the board's known new-issue auto-add bug, not a deliberate hold; this is freshly-filed, immediately workable follow-up work with no unresolved dependency.

newhoggy:
🔧 Claimed by agent `Cecil` — branch `issue-1542-jq-seq-slurp-truncated-record-location`

PR #1557, opened 2026-08-23
perf(jq): skip get_inputs' per-value location tracking under --slurp

### LINKED ISSUE #1541: perf(jq): --slurp still pays for a full per-value location scan it no longer uses
**Severity: Low** — correctness is unaffected; this is wasted CPU time, not wrong output.

## Summary

Found reviewing #1520's fix. `get_inputs`'s per-file loop (`src/bin/succinctly/jq_runner.rs`)
unconditionally calls `locations.push`/`locations.extend_from_ends` for every file, building a
`per_value` table of per-value line locations — including a dedicated `find_json_values` scan
for the JSON branch, separate from `parse_json_stream`'s own parse.

Before #1520, `--slurp` needed this table: the old `locations.last()` read `per_value` to find
"the last value's location". #1520 replaced that with `InputLocations::slurp_eof`, which reads
only `self.files` and an independently-computed EOF line — **never `self.per_value`**. So under
`--slurp`, the per-file loop's location-tracking calls (though not the *value*-parsing calls,
which are still needed) are now pure overhead: a full extra scan per file that populates a table
nothing downstream reads.

## Why this is separate from #1520

#1520 was scoped to fixing the wrong *value* `--slurp` reported at EOF, not to restructuring the
shared per-file loop that also serves the (still location-sensitive) non-slurp path. Skipping the
tracking calls only under `--slurp` is a real optimization opportunity, but touches DSV/raw/seq/
JSON branches and the `debug_assert_eq!(locations.len(), values.len(), "one location per value")`
invariant the queue-seeding `.zip()` in `run_jq` depends on (documented in `InputLocations::single`'s
own doc comment as a `#1309`-fixed data-loss hazard) — worth its own careful pass and a benchmark,
not a rushed addition to #1520's diff.

## Suggested fix direction

Not investigated in detail. Likely: gate the `locations.push`/`extend_from_ends` calls (and the
JSON branch's `find_json_values` call, which exists *only* to feed them) behind `!args.slurp`,
while still populating `values` unconditionally — and either relax the `debug_assert_eq!` for the
slurp case (it currently assumes locations and values grow in lockstep) or thread a flag through
so it only checks when tracking actually ran. Worth an A/B benchmark on a large multi-file
`--slurp` input (`docs/guides/benchmarking.md`) before/after, since the JSON branch's scan is the
most expensive of the four input modes.

newhoggy:
Fixes #1541

## Summary

`get_inputs`'s per-file loop unconditionally built a per-value `(source, line)`
location table (`locations.push`/`extend_from_ends`) for every input mode --
including a dedicated `find_json_values` scan for the JSON branch, run only to
feed that table. #1520 replaced `--slurp`'s own EOF location with
`InputLocations::slurp_eof`, which reads only `self.files` and an
independently-computed EOF line -- never `self.per_value` -- so under
`--slurp` the per-file loop's location-tracking calls became pure overhead:
real scanning work whose result nothing downstream reads.

This skips those calls (DSV/seq/JSON branches; the raw-input branch is
already unreachable under `--slurp`, see below) when `args.slurp` is set,
while still populating `values` unconditionally, and relaxes the
`debug_assert_eq!(locations.len(), values.len(), ...)` invariant to only
check when tracking actually ran.

## Why the raw-input branch needed no change

`args.raw_input && args.slurp && args.input_dsv.is_none()` already returns
early via its own dedicated `-R -s` branch before the per-file loop starts.
The only way to reach `args.slurp` inside the loop with `args.raw_input` set
is when `args.input_dsv` is also set -- and the DSV check is tried first in
the loop's `if`/`else if` chain, so it always wins. The `args.raw_input`
branch is therefore unreachable whenever `args.slurp` is true, with or
without this change.

## Testing

- Added `test_jq_dsv_slurp_skips_location_tracking_1541` -- the one input
  mode (`--input-dsv`) with no prior `--slurp` coverage, and exactly the path
  this change touches. Confirms the slurped values are unaffected and the
  error location still follows #1520's last-source-EOF rule.
- Full suite green locally (`cargo test --features cli,simd,regex,serde`),
  both required clippy legs clean, `cargo build --no-default-features`
  (no_std) clean, `cargo doc --all-features` clean.
- Manually cross-checked DSV/raw/seq/JSON `--slurp` output and error
  locations against pinned `jq` 1.7.1 where an oracle exists.

## Benchmark (Phase 3 discipline)

Interleaved A/B, 5-file `--slurp -c '.'` JSON corpus at 5/25/100MB, 7 reps,
output identity gated, on both pinned machines (full tables in the
[issue comment](https://github.com/rust-works/succinctly/issues/1541#issuecomment-5384824492)):

**Result: neutral on both architectures.** Every delta is within the ~0.7-0.8%
control noise floor measured on the same machine at the same tier -- no
size-scaling signature either direction. `find_json_values`'s scan is real
work, but it's a small fraction of end-to-end time next to file I/O and
`parse_json_stream`'s own full parse, both of which run regardless -- too
small to isolate from process-level noise at these sizes.

Keeping the change anyway: it's a verified elimination of genuinely dead work
(identity-gated correct on every tier tested), adds no complexity, and has no
measured downside on either architecture.

## Test plan

- [x] A/B benchmark on both pinned machines, output identity gated

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -1579,9 +1579,15 @@ fn get_inputs(
             // DSV input: each row becomes a JSON array of strings
             let parsed = parse_dsv_input(&raw, delimiter);
             // One row per line (approximate for embedded newlines; jq has no
-            // DSV input mode, so there is no oracle to match here).
-            for line in 1..=parsed.len() {
-                locations.push(src, line);
+            // DSV input mode, so there is no oracle to match here). Skipped
+            // under `--slurp` (#1541): the combined array's own location
+            // comes from `slurp_eof_line` above, not from any of these
+            // per-value entries, which `get_inputs` discards wholesale when
+            // it replaces `locations` with `InputLocations::single` below.
+            if !args.slurp {
+                for line in 1..=parsed.len() {
+                    locations.push(src, line);
+                }
             }
             values.extend(parsed);
         } else if args.raw_input {
@@ -1593,7 +1599,10 @@ fn get_inputs(
         } else if args.seq {
             // JSON sequence input (RFC 7464): split on RS, ignore parse failures
             let parsed = parse_json_seq(&raw);
-            locations.extend_from_ends(src, &raw, &json_seq_ends(raw.as_bytes()), parsed.len());
+            // Skipped under `--slurp` (#1541) -- see the DSV branch above.
+            if !args.slurp {
+                locations.extend_from_ends(src, &raw, &json_seq_ends(raw.as_bytes()), parsed.len());
+            }
             values.extend(parsed);
         } else {
             // JSON input: validate first if --validate is set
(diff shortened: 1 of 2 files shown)


newhoggy:
## Multi-agent code review (7 angles) — findings addressed

7 independent review agents examined this change: the raw-input-branch
reachability claim (dedicated verification), a line-by-line diff scan, a
removed-behavior auditor, a cross-file consumer tracer, Rust-pitfall +
CLAUDE.md-conventions checks, cleanup/simplification/efficiency/altitude,
and test-coverage/edge-cases.

**Confirmed correct, no action needed:**
- The raw-input branch's "unreachable under `--slurp`" claim is exact
  (dedicated verification agent walked the boolean algebra and confirmed
  live: the early-return and the loop's `if`/`else if` ordering together
  make it provably unreachable, not just empirically so).
- Every downstream consumer of `locations`/`values` is insulated by
  `InputLocations::single`'s unconditional collapse under `--slurp` — no
  bug found in any of DSV/raw/seq/JSON × `--validate`/multi-file/empty-file
  combinations tried.

**Fixed (3 independent angles converged on the same 2 points):**
- Added a comment on the raw-input branch explaining why it's the one arm
  left ungated — it previously looked like an oversight rather than a
  proven exemption.
- Documented that skipping `find_json_values` under `--slurp` also narrows
  its divergence-check-against-`serde_json` safety net to non-slurp mode
  only (already believed unreachable, now explicit).
- Documented what still guards the slurp path now that the inner
  `debug_assert_eq!` is scoped to `!args.slurp` (the separate `run_jq`-level
  assert against `InputLocations::single`'s invariant).

**Closed 3 test-coverage gaps** the dedicated coverage pass found directly on
this PR's touched code path (not the pre-existing, out-of-scope `--validate`
gap it also found — filed as #1558 instead):
- DSV slurp's `input`/`inputs` live-queue accessor (previously only the
  direct `error()` accessor was tested — #1542 established these two can
  diverge independently for `--seq`).
- DSV slurp's row-count boundary cases (single-file, empty, header-only).
- Multi-file, multi-value plain-JSON slurp — the branch `find_json_values`
  actually belongs to, previously only tested as one multi-value source or
  two single-value files, never combined.

Full suite green (54/54 groups, 0 failures), both clippy legs clean,
no_std + doc builds clean.

PR #1562, opened 2026-08-23
test(jq): add coverage for --validate (RFC 8259 strict validation)

### LINKED ISSUE #1558: jq: --validate has zero test coverage in jq_cli_tests.rs
**Severity: Low** — no known bug, pure test-coverage gap.

## Summary

Found during #1541's code review (test-coverage pass). `jq --validate` (RFC 8259 strict
validation before processing, `validate_json_input` in `src/bin/succinctly/jq_runner.rs`)
has zero occurrences in `tests/jq_cli_tests.rs` — grep for `--validate` in that file
returns nothing. (`succinctly yq`'s own `--validate`-equivalent path is separate and out
of scope here.)

Manually verified current behavior looks correct (`jq --validate -c -s '.'` on a
multi-value-per-line file reports `trailing content after JSON value`, exit 3), but
nothing in CI exercises this flag at all, for any input shape (valid, invalid,
`--slurp`, multi-file, DSV/seq/raw interaction).

## Suggested fix direction

Add direct tests for `--validate`: valid input passes through unchanged, invalid input
reports the expected RFC 8259 error and non-zero exit, and at least one `--slurp`
combination (to pin down the ordering relative to `get_inputs`'s per-file loop, which
#1541 touched: `--validate` runs before `parse_json_stream`, independent of `--slurp`).
Not investigated further — filing to make the gap trackable, not blocking any current
work.

newhoggy:
## Summary

Found during #1541's code review: `--validate` (`validate_json_input`, `src/bin/succinctly/jq_runner.rs`) had zero occurrences in `tests/jq_cli_tests.rs` — nothing exercised it, for any input shape.

Adds direct coverage:

- A single valid document passes through unchanged.
- An unterminated object reports a validation error at exit 3.
- **A genuinely valid *jq* multi-value stream (two JSON values, one per line) still fails `--validate`** — it checks the whole input is exactly one RFC 8259 document, with no awareness of jq's own multi-value parsing. Confirmed live before writing this down, and contrasted against the same input succeeding without `--validate`.
- A single document stays valid under `--slurp` (the array-wrapping happens after validation succeeds, not instead of it).
- The issue's own repro (two space-separated values on one line) pinned as a regression test — exactly the shape `--slurp` exists to combine, but `--validate` rejects before `--slurp` ever gets a chance.
- Two multi-file cases: an earlier file's failure stops before a later file is even read (named by that file in the error location), and a later file's failure doesn't discard an earlier valid file's already-produced output — confirming validation is interleaved with per-file processing, not a whole-run pre-pass.

No jq/yq oracle exists for `--validate` (a succinctly-only extension, not jq-compatibility surface), so every behavior above was verified live against this build rather than assumed.

Tests only — no `src/` behaviour change.

## Test plan

- [x] `cargo test --features cli --test jq_cli_tests` — 890/890 pass (7 new)

Closes #1558

### DIFF
--- a/tests/jq_cli_tests.rs
+++ b/tests/jq_cli_tests.rs
@@ -19477,9 +19477,12 @@ fn test_validate_reports_unterminated_object_1558() -> Result<()> {
     let (stdout, stderr, code) = run_jq_full(&["--validate", "-c", "."], Some(r#"{"a":1"#))?;
     assert_eq!(code, 3);
     assert_eq!(stdout, "", "no output once validation fails");
+    // Pins the specific error kind (UnexpectedEof), not just "some validation
+    // error fired" -- a future regression that misclassified this exact
+    // input as a different ValidationErrorKind would otherwise pass silently.
     assert!(
-        stderr.contains("validation error"),
-        "expected a validation error, got: {stderr}"
+        stderr.contains("unexpected end of input, expected ',' or '}'"),
+        "expected the unterminated-object error, got: {stderr}"
     );
     assert!(
         stderr.contains("<stdin>"),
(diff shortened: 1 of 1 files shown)


newhoggy:
\`/code-review\` (10-agent fan-out) found several real issues, all addressed:

- Dropped a redundant second process spawn in the multi-value-stream test that re-proved a fact already covered by an existing regression guard (\`test_jq_valid_multi_value_stream_unaffected_1171\`).
- Refactored both multi-file tests to use the existing \`run_jq_over_files\` helper instead of hand-rolling two \`NamedTempFile\`s — I'd missed that this helper already existed for exactly this pattern.
- \`test_validate_reports_unterminated_object_1558\` only asserted "some validation error fired" (would pass for any \`--validate\` failure); now pins the exact error text for the unterminated-object case specifically.
- Two doc-comment overclaims caught by tracing the actual code paths: one claimed a later file "is never even read" on an earlier failure — false, the lazy path eagerly reads all files up front, only *processing* is skipped; the other stated its per-file interleaving guarantee as a general \`--validate\` property, when it only holds for the non-\`--slurp\` lazy path (\`--slurp\`'s \`get_inputs\` must fully validate+parse everything before writing any output at all). Both corrected to claim only what was actually verified.

Live-verifying the review's cross-file tracing also surfaced two real findings outside this PR's tests-only scope, filed as follow-ups rather than scope-creeping this PR:
- **#1563** — jq's own \`--validate\` early-return path skips an explicit \`out.flush()\` every sibling return in the same function calls, currently masked by \`BufWriter::drop\`'s incidental (and error-swallowing) flush.
- **#1564** — \`succinctly yq --validate\`'s DOM path silently discards an earlier valid file's output when a later file fails validation, live-verified, and inconsistent with the same flag's own M2-streaming path which preserves it for the identical two files. This is a real, reproducible bug (not just a coverage gap) that #1558's own jq-only scope note would never have surfaced — one reviewer angle specifically probed the yq side as a "twin mechanism" check even though this issue explicitly deferred it.

Re-verified: all 7 tests still pass, full suite/both clippy legs/fmt still clean.

Issue #1457, opened 2026-08-22
try_pattern_alternatives's break-propagation doc comment appears wrong: break retries like error, only halt propagates immediately


newhoggy:
## Context

While implementing #1365 (`?//` alternatives in `reduce`/`foreach`), I needed to determine how `break`/`halt` interact with `?//`'s retry-on-failure semantics, so I live-verified against the pinned oracle (`/usr/bin/jq` 1.7.1) rather than trusting `try_pattern_alternatives`'s existing doc comment (`src/jq/eval.rs`), which currently states:

> A `Break` always propagates via `partial` here, uncaught by `optional` ... this codebase now matches [jq] too (#693)

and treats `Break`/`Halt` together as "always propagate immediately, never retried."

## What I found

`break` inside a `?//` alternative's body is actually **retried** by real jq, exactly like `error(...)` — it does *not* propagate immediately. Only `halt` propagates immediately.

```bash
$ echo '{"a":1}' | jq -c 'label $out | (. as {a:$x} ?// {a:$y} | if $x==1 then break $out else [$x,$y] end)'
[null,1]
```

If `break` genuinely propagated immediately (uncaught by `?//`, escaping to the outer `label $out`), this whole expression would produce nothing (the label swallows the break with no value). Instead it prints `[null,1]` — alt2's output — meaning the `break $out` raised by alt1's body was caught by `?//`'s own retry and alt2 was tried instead.

Sanity checks performed (all against `/usr/bin/jq` 1.7.1, matching this codebase's pinned oracle):
- `break` in a *non-last* alternative retries (shown above).
- `break` in the *last* alternative (forced via a genuine type-error on alt1's pattern) produces no output — consistent with propagating once there's nothing left to retry, same as `error`'s own documented "errors in the final alternative pass through" rule.
- `halt` inside an alternative's body does **not** retry — it terminates the process immediately with no output, confirming `Halt` alone (not `Break`) is the one that propagates unconditionally.
- Baseline sanity: `break` with no `?//` in play propagates normally (unwinds to `label`), and `halt` behaves normally outside `?//` too — so this is specifically about how `?//`'s retry interacts with `break`, not a `break`/`halt` implementation issue in general.

## Where this needs fixing

[...]

PR #1566, opened 2026-08-23
fix(jq): retry the next ?// alternative when foreach's EXTRACT errors

### LINKED ISSUE #1458: foreach ?//: real jq retries EXTRACT errors by re-running UPDATE with the failed value as new state (unimplemented quirk)
## Context

PR #1455 (#1365) added `?//`-separated pattern alternatives to `reduce`/`foreach`'s `as` clause. For `foreach`, only UPDATE participates in the `?//` retry: when the winning alternative's EXTRACT call itself errors/breaks/halts, succinctly propagates that directly rather than retrying a different alternative.

Real jq's actual behavior here is stranger and was deliberately not replicated (see `try_foreach_step_alternatives`'s doc comment in `src/jq/eval.rs`).

## What real jq (1.7.1) actually does

```bash
$ echo '[{"a":1}]' | jq -c 'foreach .[] as {a:$x} ?// {a:$y} (0; .+1, .+2; if . == 2 then error("boom") else [$x,$y,.] end)'
[[1,null,1],[null,1,3],[null,1,4]]
```

Trace: alt1 (`{a:$x}`, x=1,y=null) wins UPDATE with no error, producing two outputs `1, 2`. EXTRACT runs per output:
- output=1: no error, emits `[1,null,1]`.
- output=2: `. == 2` is true → `error("boom")`.

At this point succinctly (and any "retry the whole alternative from the pre-step state" model) would either propagate the error or, if retrying, recompute UPDATE fresh from state `0` with alt2's bindings. Real jq does neither — the second and third outputs (`[null,1,3]`, `[null,1,4]`) only make sense as `.+1, .+2` **re-run with alt2's bindings but with `.` = `2`** — i.e. the *failed EXTRACT call's own input* becomes UPDATE's new state for the retry, not the original pre-step accumulator and not the value EXTRACT was actually trying to process.

Additional confirmed properties (all live-verified against `/usr/bin/jq` 1.7.1):
- The already-successfully-extracted output from the failed alternative (`[1,null,1]`) is **not** discarded when the alternative later fails and falls through — it survives in the final output.
- This reproduces with `try`/`catch` wrapping too (rules out a stray uncaught-error explanation) — see the original investigation's Test 4c/4d in the PR discussion.

## Why this wasn't implemented in #1455/PR #1455

This reads as an accidental byproduct of how jq's bytecode compiles `foreach`'s UPDATE/EXTRACT pipeline together with `?//`'s retry primitive, not a designed behavior — it doesn't fit any clean "retry from X" model that generalizes past this one example, and none of the issue's own acceptance-criteria examples exercise EXTRACT erroring specifically (only UPDATE erroring, which succinctly's current implementation gets right, verified against multiple oracle repros). Fully reverse-engineering it was out of scope for that PR.

## Scope for a fix

Reproducing this exactly would need `try_foreach_step_alternatives` (`src/jq/eval.rs`) to, on an EXTRACT error/break, retry the *whole remaining* UPDATE-then-EXTRACT sequence for the next alternative using the *failed EXTRACT call's input value* as UPDATE's new state (re-binding the pattern against the original `input_val`, not the failed value). This is a meaningfully different control-flow shape than the current "retry from the pre-attempt state" model UPDATE-only retries use, so it likely needs its own dedicated loop rather than a small tweak.

Given how obscure the trigger is (`?//` + a multi-output UPDATE + an EXTRACT that errors on a later-but-not-first output), this is low priority unless a real user hits it.

newhoggy:
## Summary

PR #1455 (#1365) added `?//`-separated pattern alternatives to `reduce`/`foreach`'s
`as` clause. For `foreach`, only UPDATE participated in the retry: when the winning
alternative's EXTRACT call itself errored/broke, succinctly propagated that directly
instead of retrying a different alternative — a deliberate scope-narrowing documented
in `try_foreach_step_alternatives`'s own doc comment, left as this follow-up.

Real jq's actual behavior is stranger than a simple "retry from pre-attempt state"
model: when EXTRACT errors on one of UPDATE's outputs, it retries by re-running UPDATE
**again** with the next alternative, feeding it **the failed EXTRACT call's own input**
as UPDATE's new state — not the pre-attempt state, and not this attempt's own final
UPDATE output.

```bash
$ echo '[{"a":1}]' | jq -c 'foreach .[] as {a:$x} ?// {a:$y} (0; .+1, .+2; if . == 2 then error("boom") else [$x,$y,.] end)'
[[1,null,1],[null,1,3],[null,1,4]]
```

The `3`/`4` only make sense as `.+1, .+2` re-run against state `2` — the value EXTRACT
choked on — with alt2's bindings.

## Change

`try_foreach_step_alternatives` (`src/jq/eval.rs`) now: on an EXTRACT
error/break on a non-last alternative, abandons the rest of that attempt's own
UPDATE outputs (and its own trailing `update_control`, if any), sets `state` to the
specific value EXTRACT failed on, and continues the outer alternatives loop — so the
next alternative's UPDATE-then-EXTRACT sequence runs fresh from there. This naturally
cascades through further alternatives if they also fail, since it's the same loop.
`Halt`, and a failure on the *last* alternative, still propagate immediately, matching
UPDATE's own existing rule.

Confirmed `?//` isn't parseable in yq mode at all (`as {a:$x} ?// ...` errors
`expected '|', found '?'`), so this is jq-mode-only in practice — no yq-side testing
needed.

## Verification against pinned jq 1.7.1

- The oracle example above, byte-for-byte.
- `try`/`catch` wrapping (rules out a stray-uncaught-error explanation).
- EXTRACT failure on the *last* alternative still propagates (single-alternative
  regression check).
- `halt` inside EXTRACT on a non-last alternative still stops the whole process
  immediately, no retry.
- `break $label` inside EXTRACT retries the same as `error`.
- 3-alternative cascade: first two alternatives' EXTRACT both fail, falls through to
  the third.
- The already-successfully-extracted output from a failing alternative
  (`[1,null,1]`) survives in the final output, not discarded.
- 2-arg `foreach` (no EXTRACT) and UPDATE-level retries unaffected (regression checks).

All six cases now have dedicated regression tests in `tests/jq_cli_tests.rs`.

Fixes #1458

## Test plan

- [x] `cargo test --features cli,simd,regex,serde --no-fail-fast` — 0 failures, all 54 groups green
- [x] `cargo fmt --check`
- [x] Live-verified every scenario above against `/usr/bin/jq` 1.7.1

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -21212,16 +21212,36 @@ fn eval_owned_input<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     }
 }
 
+/// Whether a `Control` escaping a `?//` alternative's attempt should retry
+/// the next alternative rather than propagate: `Halt` never retries (must
+/// bypass `try`/`catch` and `label`/`break` -- see [`Control::Halt`]'s own
+/// doc comment), and neither does `Error`/`Break` on the *last* alternative,
+/// since there's nothing left to fall through to.
+///
+/// Shared by both of [`try_foreach_step_alternatives`]'s retry decisions
+/// (UPDATE's own, and EXTRACT's -- #1458) so the two can't independently
+/// drift on what counts as retryable, the way `try_pattern_alternatives`'s
+/// own copy of this same predicate already once did (#1457).
+/// [`try_reduce_step_alternatives`] hand-rolls an equivalent match rather
+/// than calling this, since its single UPDATE-only retry predates this
+/// helper; consolidating it too is tracked separately (#1571) rather than
+/// attempted here, to keep this fix scoped to the bug it's actually fixing.
+fn is_retryable_control(control: &Control, is_last: bool) -> bool {
+    !is_last && matches!(control, Control::Error(_) | Control::Break(_))
+}
+
 /// Try each `?//`-separated pattern alternative (#1365) for one `foreach`
 /// step, in order, against a single `input_val`.
 ///
-/// Only UPDATE participates in the `?//` retry -- the state-threading rule
-/// is identical to [`try_reduce_step_alternatives`]'s (a pattern-match
-/// failure leaves `state` untouched; a body `Error`/`Break` still updates
-/// it first, from whatever partial output that attempt produced, *then*
-/// falls through; `Halt` alone always propagates immediately) -- see that
-/// function's own doc comment for the oracle-verified rationale and repro,
-/// which applies here unchanged.
+/// Only UPDATE participated in the `?//` retry until #1458 -- the state-
+/// threading rule for UPDATE's own retry is identical to
+/// [`try_reduce_step_alternatives`]'s (a pattern-match failure leaves
+/// `state` untouched; a body `Error`/`Break` still updates it first, from
+/// whatever partial output that attempt produced, *then* falls through;
+/// `Halt` alone always propagates immediately) -- see that function's own
+/// doc comment for the oracle-verified rationale and repro, which applies
+/// here unchanged. EXTRACT now participates too, with its own, stranger
+/// state-threading rule -- see below.
 ///
 /// EXTRACT runs only once UPDATE has picked a winning alternative, fanning
 /// out over every one of UPDATE's (possibly multiple) outputs using that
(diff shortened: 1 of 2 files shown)


newhoggy:
Ran a full multi-angle review before merging. Note: one finder agent ("Reuse angle") went outside its read-only research brief and directly edited the worktree mid-review, implementing its own suggested fix. The orchestrator caught this via an unexpected `git status` diff and correctly stashed it for isolation rather than discarding it — but the stash then needed manual recovery on my end since it turned out to contain exactly the fix I'd already independently applied by that point. Restored and verified byte-identical to my own work before proceeding.

**Strong convergence across 4 independent review angles** on one real finding: the new EXTRACT-retry match arm duplicated the exact "is this control retryable" predicate already hand-written 12 lines below at the UPDATE-retry site — with the concrete precedent of this exact predicate drifting once before (#1457, in `try_pattern_alternatives`). Extracted `is_retryable_control(control, is_last) -> bool` and routed both of `try_foreach_step_alternatives`'s own retry decisions through it. Left `try_reduce_step_alternatives`'s and `try_pattern_alternatives`'s own copies alone — consolidating those too is a genuinely separate, bigger change than this issue's own scope, filed as #1570.

**Also fixed:**
- A stale doc-comment sentence ("Only UPDATE participates in the `?//` retry") that directly contradicted the very next paragraph describing EXTRACT's new retry — exactly the misreading that produced #1458 in the first place (2 angles independently caught this).
- The missing regression test for EXTRACT failing on a *middle* (non-first/non-last) output of a 3-element UPDATE fan-out — the doc comment's own claimed subtle case, live-verified against jq 1.7.1 before locking it in (2 angles independently flagged this gap).
- A doc-comment phrasing ambiguity where the oracle repro's output looked like one nested array value when the actual (unwrapped) filter prints three separate top-level outputs.
- A test comment claiming `try`/`catch` verification that wasn't actually in the test body.

**Considered and left as-is:** an avoidable clone (`state = update_val.clone()` on a value that's provably never read again on that path) — the value is small (an EXTRACT input, not a full document) and this is a cold path (only reached when EXTRACT itself errors), so restructuring the loop's ownership to save one clone wasn't judged worth the added complexity.

Full test suite green (0 failures, 54/54 groups — two single spurious failures across intermediate runs, in unrelated tests, both confirmed passing in isolation; this repo currently has ~40 concurrent sessions building/testing, matching this project's own documented contention precedent). Clippy x2, no_std, fmt, doc all clean.

PR #1568, opened 2026-08-23
fix(jq): detect a truncated --seq -s record split across a file boundary

### LINKED ISSUE #1550: jq: --seq -s can't detect a truncated record split across a file boundary
**Severity: Low** — a narrow, multi-file edge case within an already-narrow scenario
(#1542's own trailing-record-drop rule).

## Summary

Found reviewing #1542's fix. `slurp_eof_line`'s drop check (`src/bin/succinctly/
jq_runner.rs`, `get_inputs`) inspects only `raw_inputs.last()` — the *last file's* own raw
text. Real jq's `-s` reader treats every file on the command line as one continuous byte
stream for `--seq` record-splitting purposes, so a truncated trailing record physically
located in a non-last file (with the actual last file empty, or itself RS-less) still
loses jq's EOF position — but succinctly's per-last-file check can't see across the file
boundary to find it.

Two files: `f1` = `\x1e1\n\x1e{"a":1` (truncated trailing record), `f2` = `` (empty):

```
$ jq --seq -s 'error("x")' f1 f2
jq: ignoring parse error: Unfinished JSON term at EOF at line 2, column 6
jq: error (at <unknown>): x

$ succinctly jq --seq -s 'error("x")' f1 f2
jq: error (at f2:0): x
```

succinctly only ever calls its trailing-record check on `f2`'s (empty) text, so it can't
detect `f1`'s truncation and falls through to #1520's ordinary "empty last file, EOF at
line 0" rule instead.

## Why this is separate from #1542

#1542's own fix and tests only exercise a single source (or a last source that's the one
actually truncated). This is the multi-file case where the *last* file isn't the one
carrying the problem — a genuinely different mechanism (needing knowledge of the whole
concatenated stream's record-parsing history, not just the last file's isolated content)
from anything #1542 implemented.

## Suggested fix direction

Not investigated in depth. Likely needs `seq_trailing_record_is_dropped`-equivalent logic
to walk backward across `raw_inputs` from the last file until it finds the actual last
non-empty source, rather than assuming that's always `raw_inputs.last()` — mirroring how
real jq's own `--seq` reader tracks record boundaries across the whole concatenated
stream, not per-file.

newhoggy:
Fixes #1550.

`slurp_eof_line`'s trailing-record drop check (`get_inputs`, `jq_runner.rs`) only ever inspected `raw_inputs.last()` — the physically last file on the command line. Real jq's `-s` reader treats every file as one continuous byte stream, so a truncated trailing record left behind in an earlier file, followed by one or more empty trailing files, still loses jq's EOF position (`(at <unknown>)`), but a check that only looks at the last file falls through to #1520's "empty last file, EOF at line 0" rule instead.

Added `last_non_empty_seq_source`, which walks `raw_inputs` backward past empty files to the source that actually carries the stream's last record, and runs the existing `seq_trailing_record_is_dropped` check against that instead of always trusting the physically-last entry.

Every case below verified live against pinned jq 1.7.1:

```
$ jq --seq -s 'error("x")' f1 f2         # f1 truncated, f2 empty
jq: error (at <unknown>): x

$ succinctly jq --seq -s 'error("x")' f1 f2   # before: (at f2:0)  |  after: (at <unknown>)
jq: error (at <unknown>): x
```

Added `test_jq_seq_slurp_truncated_record_across_file_boundary_1550`, covering: the exact reported two-file shape, a three-file variant (two empty trailing files, confirming the walk-back isn't limited to skipping just one), and a control (genuinely complete trailing record + two empty files) confirming the walk-back doesn't fire and #1520's own rule still applies unchanged.

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` (full suite, 895/895 jq CLI tests pass)
- [x] `cargo fmt --check`
- [x] New fix verified live against pinned jq 1.7.1 for the reported case, a 3-file variant, and the pre-existing #1520 control case (no regression)

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -1535,14 +1535,12 @@ fn get_inputs(
     // not `<unknown>`.
     let slurp_eof_line: Option<usize> = if args.slurp {
         raw_inputs.last().and_then(|(_, raw)| {
-            // #1550: the drop check itself has to walk back past any empty
-            // trailing files to the source that actually carries the
-            // stream's last record -- `raw_inputs.last()` alone can't see a
-            // truncated record left behind in an earlier file.
-            let dropped = args.seq
-                && !args.raw_input
-                && last_non_empty_seq_source(&raw_inputs)
-                    .is_some_and(seq_trailing_record_is_dropped);
+            // #1550: the drop check has to read across every file as one
+            // stream, not just `raw_inputs.last()` alone -- a truncated
+            // record's own opening RS byte and its closing/disambiguating
+            // bytes can live in different files.
+            let dropped =
+                args.seq && !args.raw_input && seq_stream_trailing_record_is_dropped(&raw_inputs);
             if dropped {
                 None
             } else {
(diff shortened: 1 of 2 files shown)


newhoggy:
## Review findings addressed

`/code-review`'s fan-out (5 independent angles) converged on the same root defect in the first version: the initial fix inspected only the physically-last *non-empty* file in isolation, discarding bytes that mattered:

1. **Spanning-and-closing regression** (found independently by 2 agents): a file with no RS byte of its own that's actually the *closing continuation* of a truncated record in an earlier file was misclassified as "abandoned text," turning a real, resolvable jq location into a spurious `<unknown>`. Verified live: `g1="\x1e{\"a\":1"`, `g2="}\n"` — oracle reports `(at g2:1)`, the first fix reported `(at <unknown>)`.
2. **Whitespace-disambiguation regression**: a bare number in one file, disambiguated by a lone trailing space in the *next* file, was also misclassified as `<unknown>` because the whitespace-only trailing file was skipped as "empty" via `.trim().is_empty()` — discarding exactly the byte that resolves the ambiguity.

Both were real, oracle-verified regressions introduced by the first version, not present on `main`.

**Fix**: replaced the isolate-one-file approach with `seq_stream_trailing_record_is_dropped`, which walks backward across files folding each RS-less file's *exact* bytes (never trimmed) onto whatever followed it, until it reaches a file containing an RS byte — then runs the existing single-source check against the fully reassembled tail. This matches the "every file is one continuous byte stream" model the code already claimed, rather than approximating it. Re-verified live against all 6 cases (original repro, the #1520 empty-trailing-file control, both regressions above, the 2-file spanning case, and the pre-existing single-file #1542 case) — all match pinned jq 1.7.1 exactly.

Also added an `ErrorAt::Live` (`input_line_number`) test for the new cross-file path, since #1542 previously found the Live and Fixed accessors could desync independently.

**Filed as a follow-up, out of scope here**: #1571 — a deeper, related bug where `--seq -s` silently drops a record's *value* (not just its error location) when the record's bytes genuinely span a file boundary. This fix only touches the location computation; fixing the value-level `parse_json_seq` path is a larger change (per-file parsing would need to become stream-aware too) better scoped separately.

Issue #1279, opened 2026-08-19
jq: generator-argument builtins don't fan out over a multi-output argument


newhoggy:
## Summary

Split out of #1045's own "Suggested approach" (point 2), which explicitly flagged this
as "a bigger, separate architectural question" not to be bundled with the `None`-arm
propagation fix (#1278/#1045).

## Confirmed repro (verified against pinned jq 1.7.1)

```
$ echo 1 | jq -c '[contains((1,2))]'
[true,false]
$ echo 1 | succinctly jq -c '[contains((1,2))]'
[true]
```

Real jq's `contains(x)` (like most generator-argument builtins, per jq's own `x as $b |
body` desugaring) fans out over *every* output of `x`, producing one result per output.
succinctly's `builtin_contains` — and every other builtin built the same way (`eval_single`
on the argument, then a single-value collapse via `result_to_owned`/`result_to_owned_ctrl`/
`result_to_owned_full`) — only ever uses the *first* output and silently discards the rest.

## Scope

Per #1045's own framing, this needs a builtin-by-builtin audit against real jq's actual
`.jq` prelude definitions (`src/builtin.jq`, fetchable) or live probing, since not every
generator-argument builtin necessarily fans out the same way real jq's `contains` does —
this is a bigger architectural question than the `None`-arm fix, likely requiring each
affected call site to be restructured from "resolve to one value" to "loop over every
value, emit one result per iteration," which changes the calling convention more
substantially than #1278's fix did.

Not scoped further here — left for whoever picks this up to investigate and scope before
implementing, per #1045's own explicit request that this get its own dedicated
investigation.

newhoggy:
## Tier review — **Tier 3**

Split out of #1045's own "Suggested approach", which flagged it as "a bigger, separate architectural question" — that framing is right, and it is the Tier 3 criterion directly.

Making generator-argument builtins fan out over a multi-output argument changes evaluation *shape* rather than fixing a local mistake: it affects how many outputs a builtin produces, which is observable in every downstream consumer. That is architectural, wants a design doc first, and cannot be pinned by patching one call site.

### Read with its siblings

- **#1277** — the ~24 remaining `result_to_owned`/`eval_owned_expr_ctrl` callers dropping a trailing escape.
- **#1280** — `eval_owned_expr_ctrl_full` collapsing an empty builtin result to `null`.

[...]

PR #1583, opened 2026-08-23
fix(jq): let a RejectMany slot's own escape outrank its count message

### LINKED ISSUE #1533: yq: ArgFanout::RejectMany masks an embedded error() with a generic "expected a single result" message
**Severity: Medium** (yq mode only — a real error gets replaced with a misleading message)

## Summary

Found during #1279's own code review (PR pending). `ArgFanout::RejectMany` (`src/jq/eval.rs:1447`, used by `setpath`/`delpaths` in yq mode) is meant to refuse a multi-output path/value argument outright. Its check, duplicated at three sites —

- `fanout_arg`, `src/jq/eval.rs:1516`: `ArgFanout::RejectMany if args.len() > 1 => { ... }`
- `fanout_two_args` (outer), `src/jq/eval.rs:1583`: `ArgFanout::RejectMany if outers.len() > 1 => { ... }`
- `fanout_two_args` (inner), `src/jq/eval.rs:1596`: `ArgFanout::RejectMany if inners.len() > 1 => { ... }`

— tests `len() > 1` and returns `Err("expected a single result but found N")` *before* the caller ever looks at `trailing`, the generator's own deferred `Control::Error` captured by `stream_outputs`. So when the multi-output generator's excess outputs come from a real `error(...)` call rather than genuinely distinct values, that error is discarded in favor of the generic "expected a single result" message.

The yq-mode gate that routes `setpath`/`delpaths` through `RejectMany` is also an inline `if S::TAG == EvalTag::Yq { ArgFanout::RejectMany } else { ArgFanout::All }`, duplicated verbatim at `src/jq/eval.rs:25109-25113` (`builtin_setpath`) and `:29365-29369` (`builtin_delpaths`), rather than a named predicate like the existing `ArgFanout::yq_native::<S>()`. The new `docs/compliance/yq/limitations.md` "Generator-argument fan-out" section claims "the predicate is named rather than inlined precisely so that stays a grep (CLAUDE.md's #106 lesson)" — true only for the `FirstOnly` gate (`yq_native`), not for this `RejectMany` gate, which the same section's own table cites for `setpath`/`delpaths`.

## Confirmed repro (verified against pinned yq v4.53.3)

```
$ printf 'null' | yq 'setpath((1,2,error("boom")); 1)'
Error: boom
$ printf 'null' | succinctly yq -o json 'setpath((1,2,error("boom")); 1)'
Error: expected a single result but found 2
```

Real yq propagates the embedded error. succinctly reports the generic "expected a single result" message instead, masking the actual failure. The same divergence reproduces for the value-side generator and for `delpaths`.

## Root cause

The `RejectMany if len > 1` guard fires unconditionally on output count, without first checking whether `trailing` already carries a `Control::Error`/`Break`/`Halt` that should take precedence.

## Suggested fix direction

Check `trailing` before (or instead of) the length guard — an error/break/halt already captured from the generator should propagate rather than be shadowed by the "expected a single result" message, matching how every other `fanout_arg`/`fanout_two_args` caller lets a captured control win. Separately, extract the `if S::TAG == EvalTag::Yq { RejectMany } else { All }` gate into a named constructor (e.g. `ArgFanout::reject_many_in_yq::<S>()`) alongside `yq_native`, both to remove the duplication and to make `docs/compliance/yq/limitations.md`'s "grep the predicate" claim actually true for this gate.

## Related

- #1279 — the fan-out work this bug was found reviewing.

newhoggy:
## Summary

Found reviewing #1279. `ArgFanout::RejectMany`'s own count check
(`args.len() > 1`) fires before `fanout_two_args` ever looks at that slot's
trailing control, so `setpath`'s own repro from #1533 masked a real
`error(...)` behind a generic count message:

```bash
$ printf 'null' | yq            'setpath((1,2,error("boom")); 1)'
Error: boom
$ printf 'null' | succinctly yq 'setpath((1,2,error("boom")); 1)'
Error: expected a single result but found 2
```

#1533's single-argument half (`delpaths`, via `fanout_arg`) was already
fixed earlier in this series via `clear_values_when_yq_argument_escaped` —
but that fix was deliberately *not* extended to `fanout_two_args`, because
doing so regressed two other shapes (`test`/`setpath`'s own per-builtin
escape-ordering, where emptying a slot's values pre-emptively skips `body`,
and `body` is where the *other* slot gets validated). That general rule
stays reverted; a prior commit in this branch's history left the
two-argument gap pinned as a known divergence rather than silently passing,
specifically so the eventual fix could flip the pin.

## Fix

The remaining gap is narrower than "any escape in a two-argument slot,"
though: `RejectMany`'s own count violation is itself only ever a *symptom*
of an escape — a generator produces 2+ values instead of collapsing to one
specifically because it didn't get to before raising. Deferring to that
slot's trailing control only when the count check is what's about to fire
doesn't touch the single-value-then-escape shapes the reverted fix broke.
Applied at both of `fanout_two_args`'s own `apply_arg_fanout` call sites
(outer and inner).

## Verification

- Live-verified against pinned yq v4.53.3 for the exact repro above, plus
  the "clean multi-output, no escape" guard (`setpath((["a"],["b"]); 1)`
  must still refuse), and both previously-fixed per-builtin-ordering cases
  (`test(("a", error("p")); "z")` still reports the flags-validation error,
  `setpath((["a"],error("p")); (1,error("v")))` still reports the path
  slot's escape) — all four match the oracle exactly.
- Flips the pin `test_yq_setpath_two_argument_escape_order_is_unfixed_1533`
  (renamed to
  `test_yq_setpath_two_argument_reject_many_propagates_an_embedded_error_1533`),
  while the regression guards for the previously-reverted general rule
  (`test_yq_two_argument_body_validation_outranks_a_slot_escape_1533`,
  `test_yq_fanout_two_args_argument_escape_reports_bare_not_prefix_then_raise`)
  continue to pass unchanged.
- Updated `docs/plan/jq-generator-argument-fanout.md` and
  `docs/compliance/yq/limitations.md`, both of which documented this as a
  still-open half of #1533.

Fixes #1533

## Test plan

- [x] `cargo test --features cli,simd,regex,serde --no-fail-fast` — 0 failures, all 54 groups green
- [x] `cargo fmt --check`
- [x] Live-verified against pinned yq v4.53.3

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -398,6 +398,13 @@ which was true only before #1534.
   escape, not competing evidence against it, so `fanout_two_args` now defers to that slot's
   trailing control whenever its own count check is what's about to fire. Confirmed by
   `test_yq_setpath_two_argument_reject_many_propagates_an_embedded_error_1533`.
+
+  Review found this alone still misattributed the error when *both* slots violate `RejectMany`
+  at once — reporting whichever slot (outer) it happened to check first, rather than real yq's
+  own consistent "inner always wins" rule (for `setpath`, path over value — live-verified across
+  every escaping/clean combination). `fanout_two_args` now always evaluates inner before
+  reporting outer's own violation, matching real yq's own evaluation order rather than just its
+  final answer. See `test_yq_setpath_reject_many_prefers_inner_violation_over_outer_1533`.
 - **`contains` on an escaping argument**
   ([#1553](https://github.com/rust-works/succinctly/issues/1553)). `contains` is ungated because
   real yq fans it out, but real yq still emits nothing when its argument escapes, where
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -388,11 +388,16 @@ which was true only before #1534.
   *outcome* — an error rather than a fan-out or a silent truncation — is what #1279 preserved;
   matching the per-slot wording is unstarted.
 - **Two-argument escape ordering** ([#1533](https://github.com/rust-works/succinctly/issues/1533),
-  half-closed). The escape-clearing above is deliberately *not* applied to `fanout_two_args`:
-  emptying one slot's values skips the body, which is where the other slot gets validated, and
-  which slot real yq reports is per-builtin and does not follow succinctly's outer/inner order
-  (`test` wants the flags, `setpath` wants the path). That needs a per-builtin ordering probe.
-  Pinned as a known divergence by `test_yq_setpath_two_argument_escape_order_is_unfixed_1533`.
+  now fully closed). The escape-clearing above is deliberately *not* applied to
+  `fanout_two_args`: emptying one slot's values skips the body, which is where the other slot
+  gets validated, and which slot real yq reports there is per-builtin and does not follow
+  succinctly's outer/inner order (`test` wants the flags, `setpath` wants the path) — that part
+  stays a known limitation with no shared rule. But the *specific* case #1533 was filed for —
+  `RejectMany`'s own `args.len() > 1` count check masking a real `error(...)` in the same
+  slot — needed no per-builtin probe: a count violation there is itself a symptom of the
+  escape, not competing evidence against it, so `fanout_two_args` now defers to that slot's
+  trailing control whenever its own count check is what's about to fire. Confirmed by
+  `test_yq_setpath_two_argument_reject_many_propagates_an_embedded_error_1533`.
 - **`contains` on an escaping argument**
   ([#1553](https://github.com/rust-works/succinctly/issues/1553)). `contains` is ungated because
   real yq fans it out, but real yq still emits nothing when its argument escapes, where
--- a/docs/plan/jq-generator-argument-fanout.md
+++ b/docs/plan/jq-generator-argument-fanout.md
@@ -394,9 +394,16 @@ duplicated five times), and the single-argument half of
 #1531's lazy pull is the one that changed a documented conclusion: it closed the last golden
 known-failure, the #820 eager-argument residue this design had recorded as out of scope.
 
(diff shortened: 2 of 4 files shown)


newhoggy:
Ran a full multi-angle review before merging. Found a genuine, significant gap requiring rework, not just polish.

**The fix's own compound case was wrong.** When BOTH slots of `setpath` independently violate `RejectMany` at once, my first version reported whichever slot (outer/value) it happened to check first — but real yq consistently prefers the inner/path slot's own violation, live-verified against pinned v4.53.3 across every combination:

```
setpath((1,2,error("PATH_ERR")); (1,2,error("VAL_ERR")))  => Error: PATH_ERR   (both escape)
setpath((1,2,error("PATH_ERR")); (1,2))                    => Error: PATH_ERR   (inner escapes, outer clean)
setpath((["a"],["b"]); (1,2,error("VAL_ERR")))              => inner's own count message, not VAL_ERR (inner clean, outer escapes)
```

`fanout_two_args` now always evaluates inner once before reporting outer's own violation of either kind — this is inner's *first* evaluation on this path (never an extra one), since the loop below is what would otherwise have done that on its first iteration anyway, so there's no side-effect duplication risk.

Added `test_yq_setpath_reject_many_prefers_inner_violation_over_outer_1533` covering all four escaping/clean combinations. Verified it actually catches the gap by reverting to the narrower (outer-first) version and confirming the test fails there before restoring the fix — the same discipline used for the earlier field-order test on #1458.

Also added the specific mirror-shape test the review named as missing (`setpath(["a"]; (1,2,error("boom")))` — outer escapes with multiple values, inner is a plain literal).

Updated both `docs/` files this PR touches, which had said the original fix was "fully closed" before this finding — narrowed to accurately describe what's actually now handled.

Full test suite green (0 failures, 54/54 groups). Clippy x2, no_std, fmt, doc all clean.

PR #1584, opened 2026-08-23
fix(jq): input_line_number reports 0, not the raw sentinel, for a dropped --seq -s record

### LINKED ISSUE #1549: jq: input_line_number reports a bogus huge number for a --seq -s dropped trailing record
**Severity: Low** — a wrong number from a rarely-called builtin, in an already-narrow
combination (`--seq --slurp` on a stream whose trailing record real jq itself never
resolves, #1542, *and* the filter also references `input_line_number`).

## Summary

Found reviewing #1542's fix. That issue made `InputLocations` represent "no real
position" via a sentinel `per_value` line (`UNKNOWN_LINE = u32::MAX`), checked once
inside `InputLocations::resolve()` so both the direct-lookup (`ErrorAt::Fixed`/`get()`)
and queue-based (`ErrorAt::Live`, reached whenever a filter references `input`/`inputs`/
`input_line_number`) paths agree on printing `<unknown>` for the error marker.

`input_line_number` itself, though, doesn't go through `resolve()` at all —
`remaining_inputs::pop()` (`src/jq/eval.rs`) sets `LAST_LINE` directly from the raw
`line` value popped off the shared queue, and `builtin_input_line_number` reads
`LAST_LINE` verbatim. For the `UNKNOWN_LINE` sentinel specifically, that means
`input_line_number` returns `4294967295` (the raw sentinel, cast to `i64`) instead of
real jq's own answer for this exact scenario.

```
$ printf '\x1e1\n\x1e{"a":1' | jq --seq -s 'input_line_number, error("x")'
jq: ignoring parse error: Unfinished JSON term at EOF at line 2, column 7
jq: error (at <unknown>): x
0

$ printf '\x1e1\n\x1e{"a":1' | succinctly jq --seq -s 'input_line_number, error("x")'
jq: error (at <unknown>): x
4294967295
```

The error marker is already correct (#1542's fix); only `input_line_number`'s own
numeric output diverges.

## Why this is separate from #1542

Fixing it properly means threading the "no real position" concept across the
`jq_runner.rs` (CLI binary) / `src/jq/eval.rs` (library) crate boundary that #1542's own
fix deliberately stayed within — `jq_runner.rs`'s `UNKNOWN_LINE` constant isn't visible to
`eval.rs` (the binary depends on the library, not the reverse), so `remaining_inputs::pop`/
`builtin_input_line_number` would need their own awareness of the sentinel, or the
`(OwnedValue, u32, u32)` queue tuple would need a real `Option`-shaped "no position"
representation instead of a magic value — a genuinely bigger, riskier change than
anything #1542 touched, for a bug with no crash/data-loss risk (the huge number doesn't
overflow `i64`, it's just visibly wrong).

## Suggested fix direction

Not investigated in depth. Likely needs either: (a) a shared sentinel constant defined in
the *library* crate (`src/jq/`) that both `jq_runner.rs` and `eval.rs` can reference, with
`builtin_input_line_number` mapping it back to `0` (real jq's own answer for "no line
known yet") at the point of read; or (b) widening the queue's per-document tuple to
`(OwnedValue, u32, Option<u32>)` so "no position" is a real variant instead of a
magic value, at every consumer.

newhoggy:
Fixes #1549.

`remaining_inputs::pop` (`src/jq/eval.rs`) sets `LAST_LINE` directly from the raw per-value `line` the CLI driver seeds the queue with, with no concept of "no real position." For #1542's `UNKNOWN_LINE` sentinel (`u32::MAX`, used when a `--seq -s` trailing record real jq's own incremental parser never resolves), that meant `input_line_number` returned the raw sentinel cast to `i64` (`4294967295`) instead of jq's own answer of `0`.

The `(at ...)` error marker already handled this correctly — `jq_runner.rs`'s `InputLocations::resolve` centrally checks for its own private `UNKNOWN_LINE` constant — but `input_line_number` reads `LAST_LINE` through a completely different path (`remaining_inputs::last_line()`) that never went through that check, and the CLI's sentinel was private to the binary crate, invisible to the library-crate code that actually needed to recognize it.

## Fix

Added `pub const UNKNOWN_INPUT_LINE` to the library crate (`src/jq/eval.rs`'s `remaining_inputs` module, re-exported via `src/jq/mod.rs`), so both sides agree on one sentinel instead of independently picking `u32::MAX`:
- `builtin_input_line_number` now maps the sentinel back to `0` at read time.
- `jq_runner.rs`'s own `UNKNOWN_LINE` now derives its value from the shared library constant instead of defining `u32::MAX` a second time.

```
$ jq --seq -s 'input_line_number, error("x")' <<< $'\x1e1\n\x1e{"a":1'
jq: ignoring parse error: Unfinished JSON term at EOF at line 2, column 7
jq: error (at <unknown>): x
0

$ succinctly jq --seq -s 'input_line_number, error("x")' <<< $'\x1e1\n\x1e{"a":1'
jq: error (at <unknown>): x
0   # was 4294967295 before this fix
```

No new divergence — this fix removes one, bringing `input_line_number` in line with jq's own answer.

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` (full suite, no regressions; one unrelated flake — `test_short_circuit_side_effect_shapes_already_match_jq_820` killed by SIGKILL under this machine's heavy concurrent load — reran alone and clean, then reran the full suite clean)
- [x] `cargo fmt --check`
- [x] New fix verified live against pinned jq 1.7.1 for the reported repro; verified normal (non-sentinel) `input_line_number` cases and #1520's own empty-last-file case are unaffected

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -1742,7 +1742,15 @@ fn json_seq_ends(raw: &[u8]) -> Vec<usize> {
 /// straight from [`per_value`](InputLocations::per_value) -- answers
 /// `<unknown>` the same way, rather than only the direct-lookup path
 /// checking a side flag `resolve` itself didn't know about.
-const UNKNOWN_LINE: u32 = u32::MAX;
+///
+/// Value comes from the library crate's own `jq::UNKNOWN_INPUT_LINE` (#1549)
+/// rather than an independently-picked `u32::MAX` here: this same raw value
+/// also crosses into the shared `input`/`inputs`/`input_line_number` queue
+/// (`seed_remaining_inputs`, below), so `builtin_input_line_number` needs to
+/// recognize the exact sentinel this side emits -- previously it didn't,
+/// and reported the raw `u32::MAX` (`4294967295`) instead of real jq's own
+/// `0` for a dropped trailing `--seq -s` record.
+const UNKNOWN_LINE: u32 = jq::UNKNOWN_INPUT_LINE;
 
 /// Source locations for the values returned by [`get_inputs`].
 ///
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -26767,8 +26767,19 @@ mod remaining_inputs {
         CURRENT.with(Cell::get)
     }
 
-    pub fn last_line() -> u32 {
-        LAST_LINE.with(Cell::get)
+    /// `None` when the last-popped document carries [`UNKNOWN_LINE`] (no
+    /// real EOF position, e.g. a dropped `--seq -s` trailing record, #1542)
+    /// instead of a real line -- checked once, here, rather than leaving it
+    /// to whichever caller remembers to compare against the sentinel (#1549:
+    /// `input_line_number` didn't, and reported the raw sentinel verbatim).
+    /// This is `last_line`'s only caller crate-wide, so folding the check in
+    /// here costs nothing and makes "unknown" a type-level fact instead of a
+    /// convention.
+    pub fn last_line() -> Option<u32> {
+        match LAST_LINE.with(Cell::get) {
+            UNKNOWN_LINE => None,
+            line => Some(line),
+        }
     }
 }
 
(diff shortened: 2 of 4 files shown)


newhoggy:
## Review findings addressed

`/code-review`'s 7-agent fan-out came back clean on correctness (no bugs — the sentinel chain was traced end-to-end: single producer, exactly two consumers, both using the same comparison sense, `cfg(std)` gating consistent, no dual-evaluator gap). Two quality findings worth acting on:

1. **Type-level enforcement over a convention** (altitude angle): `remaining_inputs::last_line()` has exactly one caller crate-wide, so leaving the `UNKNOWN_LINE` sentinel check to that caller (as a plain `u32` comparison) left the door open for a *second* call site to forget it and reproduce #1549 silently, with no compile error. Since this accessor is private with a single caller, this was a genuinely zero-blast-radius fix: `last_line()` now returns `Option<u32>` directly (matching `UNKNOWN_LINE` against `None` internally), and `builtin_input_line_number` simplifies to `.unwrap_or(0)`. This also resolves a smaller finding from the reuse/simplification angle (the builtin was reaching into `remaining_inputs::UNKNOWN_LINE` directly instead of the public `UNKNOWN_INPUT_LINE` alias it introduced) — the comparison moved inside the module that owns the constant, so there's no longer a second name to reach for.

Two findings noted but **not acted on**, with reasoning:

- **A shared `resolve_input_line`-style helper** to unify this check with `jq_runner.rs`'s pre-existing `InputLocations::resolve` — both consumers return genuinely different types (`Option<u32>` mapping to `0` vs. `InputLocation` construction), and this is a two-call-site, single-line comparison; extracting a cross-crate helper for it reads as premature abstraction given this repo's own stated bias against introducing abstractions beyond what the task requires.
- **A theoretical `u32::MAX`-line collision** (an input with exactly 4,294,967,295 lines would misreport as unknown): pre-existing in #1542's sentinel design (already affected the `(at ...)` marker), not introduced by this PR. Properly closing it means widening the queue's tuple type to `Option<u32>` everywhere — exactly the alternative #1549's own issue text considered and passed over as "a genuinely bigger, riskier change" in favor of the shared-sentinel approach this PR implements.

Re-verified against the oracle and the full local suite after the refactor — all green.

PR #1585, opened 2026-08-23
refactor(jq): unify Expr::Iterate's array/object loop bodies

newhoggy:
Partial follow-up to the remaining scope tracked on issue 1477 (not closing it — three items remain: Map's third copy of this shape, eval_fanout/accumulate_path_context_step consolidation, and the noted-but-deferred If/Select cond_result dedup).

`eval_pipe_with_path_context_internal`'s `Iterate` arm hand-rolled the identical per-element step twice — once for `OwnedValue::Array`, once for `OwnedValue::Object` — differing only in what key gets pushed onto `new_path` and how each container is walked. Factored the shared step (build `new_path`, call `continue_rest_with_borrowed_value`, accumulate via `accumulate_path_context_step`, check for stop) into a new `iterate_element_step` helper, called from both loops.

Deliberately kept the two containers' native iteration (`arr.iter().enumerate()` vs. the map's own `iter()`) rather than unifying them into one iterator — that would need a `Box<dyn Iterator>` or a hand-rolled either-wrapper, not worth the allocation/dynamic-dispatch cost to remove five duplicated lines, especially in a function with a documented history of perf regressions from careless consolidation in this exact area (this issue's own #1548-era review caught an unconditional clone from a superficially similar refactor).

## Verification

- Output identity against the pre-change binary for path-context reads, writes, and a 1M-element correctness check (`.items[] += 1 | .items | add`).
- Interleaved release-mode A/B timing (same filter, 1M elements, 6 reps): before ~0.90-0.91s, after ~0.92s — no measurable difference.

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` (full suite, no regressions)
- [x] `cargo fmt --check`
- [x] Output identity + interleaved A/B timing on a 1M-element `.[]` write, described above

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -22758,6 +22758,41 @@ fn continue_rest_with_borrowed_value<'a, W: Clone + AsRef<[u64]>, S: EvalSemanti
     eval_pipe_with_path_context_internal::<W, S>(rest, v, root, file_origin, current_path, optional)
 }
 
+/// One `Expr::Iterate` element's full step: append `key` to `current_path`,
+/// continue `rest` from `v` at that new path, and fold the result into
+/// `results` -- the one piece the array and object loop bodies in the
+/// `Iterate` arm share verbatim, differing only in what `key` is and how
+/// they walk their own container (#1477 follow-up to #1548's `Iterate`
+/// conversion, which left this duplication in place). Kept as a shared
+/// per-element step rather than a shared iterator over both container
+/// kinds -- unifying the *iteration* itself would need either a
+/// heap-allocated `Box<dyn Iterator>` or a hand-rolled either-style
+/// wrapper, neither of which is warranted just to remove five duplicated
+/// lines, and this exact function has a documented history of perf
+/// regressions from exactly this kind of well-intentioned consolidation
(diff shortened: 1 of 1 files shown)


newhoggy:
## Review findings addressed

`/code-review`'s fresh-eyes pass came back clean on correctness — traced `results` threading, the new `iterate_element_step`'s lifetime signature (matches the pre-existing `continue_rest_with_borrowed_value` pattern exactly), and confirmed `Field`/`Index`/`IndexNumber` correctly stay untouched (genuinely single-result, not loops). Ran a fresh live A/B against `origin/main` for Object-input break/error mid-iteration and confirmed byte-identical output/exit codes on both binaries — no bug.

One real gap found: the `Object` branch's `Some(stop)` early-return path (break/error mid-iteration) had no dedicated test — the two existing `label`/`break` path-context tests only exercised the `Array` branch. Since both branches now share `iterate_element_step`, this is exactly the code path this PR touches, so added object-input mirrors of both existing tests rather than leaving it as a follow-up. Confirmed via A/B (both before and after the refactor) that behavior was already correct either way — this closes a pre-existing coverage gap, not a bug the refactor introduced.

PR #1586, opened 2026-08-23
perf(jq): stop cloning every element before map(f)'s path-context evaluation

newhoggy:
Another partial follow-up to issue 1477's remaining scope (not closing it — two items remain: `eval_fanout`/`accumulate_path_context_step` consolidation, and the deferred `If`/`Select` `cond_result` dedup).

The path-context `Builtin::Map` arm (reached only when `f` itself needs `key`/`path`/`file_index`/`parent`, per `needs_path_context`'s gate — a plain `map(.foo)` never touches this code) eagerly homogenized the array/object being mapped into one `Vec<(OwnedValue, OwnedValue)>`, cloning every element up front even though the loop only ever needs a borrow — doubling peak memory for the container being mapped, for the entire duration of the call.

Restructured into separate array/object arms mirroring `Expr::Iterate`'s own two arms (from PR #1585, same issue), each borrowing `v: &OwnedValue` straight from the source container via a new `map_element_step` helper — `iterate_element_step`'s sibling, not the same helper, since `map(f)` always evaluates `f` and has no "rest is empty" fast path to skip the way `Iterate`'s continuation does.

## Measurement

First attempt at benchmarking this used `map(.n)`, which doesn't exercise this code path at all (`needs_path_context` gates it) — caught before trusting the (flat, no-difference) result, and redone with `map(.n + (key - key))`, which does. Interleaved before/after peak RSS, output identity gated first:

| Size | before (median) | after (median) | Δ |
|------|-----------------|-----------------|---|
| 200k elements, ~40MB input | 695.5 MB | 537.9 MB | **-22.7%** |
| 500k elements, ~1GB input | 8.43 GB | 7.13 GB | **-15.5%** |

Absolute savings scale with input size (~1.3GB saved on the 1GB case), consistent with removing exactly one full copy of the mapped container's data.

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` (full suite, no regressions)
- [x] `cargo fmt --check`
- [x] Output identity against the pre-change binary at both benchmark sizes
- [x] Added object-branch error/break test coverage proactively (mirroring the exact gap `/code-review` caught on PR #1585's `Iterate` unification) — both existing tests for the mid-element error/break paths only used array input

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -22808,24 +22808,35 @@ fn continue_rest_with_borrowed_value<'a, W: Clone + AsRef<[u64]>, S: EvalSemanti
     eval_pipe_with_path_context_internal::<W, S>(rest, v, root, file_origin, current_path, optional)
 }
 
-/// One `Expr::Iterate` element's full step: append `key` to `current_path`,
+/// One container element's full step: append `key` to `current_path`,
 /// continue `rest` from `v` at that new path, and fold the result into
-/// `results` -- the one piece the array and object loop bodies in the
-/// `Iterate` arm share verbatim, differing only in what `key` is and how
-/// they walk their own container (#1477 follow-up to #1548's `Iterate`
-/// conversion, which left this duplication in place). Kept as a shared
-/// per-element step rather than a shared iterator over both container
-/// kinds -- unifying the *iteration* itself would need either a
-/// heap-allocated `Box<dyn Iterator>` or a hand-rolled either-style
-/// wrapper, neither of which is warranted just to remove five duplicated
-/// lines, and this exact function has a documented history of perf
-/// regressions from exactly this kind of well-intentioned consolidation
-/// (#1548's own review caught an unconditional clone this shape can
-/// introduce if done carelessly).
+/// `results` -- the one piece the array and object loop bodies share
+/// verbatim in both `Expr::Iterate` and `Builtin::Map(f)`, differing only
+/// in what `key` is and how they walk their own container (#1477 follow-up
+/// to #1548's `Iterate` conversion, which left this duplication in place).
+/// `Builtin::Map(f)`'s two arms call this with `rest =
+/// core::slice::from_ref(f)` -- a slice that's always length 1, so
+/// `continue_rest_with_borrowed_value`'s own `rest.is_empty()` fast path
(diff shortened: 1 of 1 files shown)


newhoggy:
## Review findings addressed

`/code-review`'s 10-agent fan-out came back clean on correctness (7 angles traced the diff end-to-end, all confirmed no bugs, output identity, and a real live A/B against the pre-change binary). Two findings acted on:

1. **`map_element_step` was a provably redundant duplicate of `iterate_element_step`** (Simplification angle): `Builtin::Map`'s two call sites always pass `rest = core::slice::from_ref(f)` — a slice that's never empty — so `iterate_element_step`'s own `continue_rest_with_borrowed_value` call degenerates to the exact same `eval_pipe_with_path_context_internal` invocation `map_element_step` computed by hand. Verified this myself before accepting it (traced both call chains to the identical call). Deleted the second helper entirely; both `Map` arms now call `iterate_element_step` directly. This also resolves a related finding (a bare `#[allow(clippy::too_many_arguments)]` missing the STYLE-0004 citation three sibling allows in this same file already carry) — fixed on the one surviving helper.
2. **`results` had no capacity hint** (Efficiency angle): pre-sized via `Vec::with_capacity`, using the container's own length available right at the match site — avoids ~log2(n) reallocations while accumulating one output per input element.

Re-verified output identity and the same peak-RSS improvement (695.6MB → 537.8MB median, 200k elements) hold unchanged after the simplification.

**One substantive finding noted but deliberately not acted on in this PR** (Altitude angle): a suggestion that the array/object dispatch *shell* itself (not just the per-element step, already unified) could be collapsed further via a generic `impl FnMut` closure-based driver, potentially also serving 3-4 other independently-duplicated Array/Object sites elsewhere in this file (`builtin_map_values`, `update_path`'s two `Iterate` arms, `walk_path`'s `Iterate` arm). This is a legitimate architectural observation, but it's a materially larger, more invasive refactor than this PR's scope (a targeted clone fix) — and generic closure-based abstraction in exactly this perf-sensitive, regression-prone area (#1548's own history) needs its own careful implementation and benchmark verification, not an in-flight addition to an already-reviewed PR. Noting it on issue 1477's own tracking as a candidate for a future, separately-scoped pass.

Issue #1588, opened 2026-08-23
perf(jq): streaming duplicate-key detector doubles peak RSS on wide objects


newhoggy:
**Severity: Low** (no wrong output; a disclosed trade, on one path, on a shape that is rare)

## Summary

#1514 made the jq-mode duplicate-key detector 2-4x cheaper in time. On one path it paid for that
in memory: bare `keys_unsorted` over a **wide** object roughly **doubles peak RSS** against
`main`, because the streaming detector keeps an open-addressed table (16 bytes per key at its
half-load point) where the batch paths keep a sorted `Vec<u64>` (8 bytes per key).

Every other path improved or is unchanged. Filing so the trade is tracked rather than buried in a
design doc.

## Numbers

Peak RSS, Apple M4 Pro, MiB. `40c5ff93` is pre-#1385; `main` is `f9f2b5b4a`; tip is #1514's branch.

| workload | `40c5ff93` | `main` | tip |
|---|---|---|---|
| `wide/10mb` `.` | 20.8 | 63.0 | **57.1** |
| `wide/10mb` `keys_unsorted` | 20.8 | 29.7 | **52.9** |
| `wide/10mb` `keys_unsorted \| length` | 20.8 | 29.8 | 29.8 |
| `wide/100mb` `.` | 132.5 | 514.0 | **459.7** |
| `wide/100mb` `keys_unsorted` | 132.5 | 188.9 | **388.7** |
| `wide/100mb` `keys_unsorted \| length` | 132.5 | 189.0 | 189.0 |
| `users/10mb`, every filter | 16.4 | 16.5 | 16.5 |

`wide/100mb` is one object of 7.1M keys. `.` is *better* than `main` and `length` is identical;
only `keys_unsorted` doubles.

## Why

`DistinctKeyCursors` (`src/jq/document.rs`) answers "have I seen this key?" per key as it
streams, so it cannot sort — it has no list yet. It carries a `KeyHashes` table, which at load
factor 0.5 is 2 slots x 8 bytes per key, and doubles as it grows (so a transient old table is
live during a rehash).

The batch paths — `census`, `keys_repeat`, `effective_keys`, `spans_repeat` — all hash into a
`Vec<u64>` and sort, which is 8 bytes per key and also *faster* above L3. See #1514 for that
measurement.

## Directions

Roughly in order of how much they promise per unit of risk:

1. **Give the table a capacity hint.** The growth doubling means a transient old table on top of
[...]

Issue #1587, opened 2026-08-23
perf(bench): codegen-units=16 puts an unquantified ±10-27% into every CLI A/B


newhoggy:
**Severity: High** (no wrong output — every CLI A/B this project records carries an unquantified
±10-27% component, so past conclusions may be wrong and future ones are unreliable)

## Summary

`Cargo.toml` has no `[profile.release]`, so the crate builds at Rust's defaults —
**`codegen-units = 16`, no LTO**. Adding code to one module then changes how *unrelated* modules
are optimized, and the effect is large enough to swamp the changes we measure.

Concretely: `succinctly jq type` — a query that reports the root's type and touches no keys —
measured **+12% on an Apple M4 Pro and +27% on an AMD Ryzen 9 7950X** between two binaries whose
diff cannot reach that path. At `codegen-units=1` the same comparison is **+0.2%**.

Found while fixing #1514, where it first looked like an index-build regression.

## Evidence

All runs `scripts/ab-cli.py`, interleaved per repetition, 9 reps, both boxes verified idle.
`good` = `40c5ff93`, `main` = `f9f2b5b4a`, `s2`..`s7` = the #1514 series in order.

### 1. It does not bisect to the same commit on the two machines

`jq type` on `users/100mb.json`, `main` -> each stage, median delta:

| stage | M4 Pro | 7950X |
|---|---|---|
| s2 | −1.8% | +17.0% |
| s3 | −0.1% | +13.6% |
| s4 | −0.3% | +11.5% |
| s5 | +0.2% | +20.2% |
| s7 | **+12.2%** | **+27.2%** |

A real regression bisects to one commit on both machines. This does not: ARM sees nothing until
the last commit, x86 sees it from the first, with erratic magnitudes throughout.

### 2. It is per-byte work, not startup

`jq type`, `main` -> `s7`, M4 Pro:

| input | median delta | absolute |
|---|---|---|
| `users/1mb.json` | +3.2% | +0.1 ms |
| `users/10mb.json` | +9.4% | +1.3 ms |
| `users/100mb.json` | +12.7% | +13.3 ms |

~0.19 ms/MB — it scales with the input, so it is in reading plus `JsonIndex::build`. Which is
also why it hides on output-dominated workloads: `arrays/100mb` `keys_unsorted` spends most of
its time printing and reads −1.7%, while `users/100mb`, which is build-dominated, reads +10.8%.

### 3. `codegen-units=1` removes it

[...]

PR #1589, opened 2026-08-23
perf(jq): make the duplicate-key detector cheap on documents with no duplicates

### LINKED ISSUE #1514: perf(jq): #1385's duplicate-key detector costs 2-3x on object-heavy input with no duplicate keys

newhoggy:
Fixes #1514.

#1385's duplicate-key collapse is correct and stays; its **cost** was the problem. On a document
holding one wide object — none of whose objects carries a repeated key — the detector cost up to
**+141%** against the pre-#1385 baseline `40c5ff93`.

## Result

`codegen-units=1` on both sides (see #1587 for why that matters), median of 7, interleaved,
against `40c5ff93`. Control bands: M4 Pro −0.7%..+0.4%, 7950X −2.8%..+0.8%. Identity: **25
configurations, 0 differences on each box.**

| `wide`, 1 / 10 / 100 MB | M4 Pro `main` | M4 Pro this PR | 7950X `main` | 7950X this PR |
|---|---|---|---|---|
| `.` | +24 / +30 / +34% | **+16 / +19 / +24%** | +39 / +46 / +49% | **+24 / +28 / +30%** |
| `keys_unsorted` | +88 / +123 / +136% | **+50 / +57 / +96%** | +129 / +137 / +141% | **+78 / +97 / +142%** |
| `keys_unsorted \| select(true)` | +27 / +32 / +37% | **+4 / +4 / +5%** | +32 / +39 / +52% | **+2 / +4 / +3%** |
| `keys_unsorted \| map(.)` | +20 / +25 / +30% | **+4 / +6 / +12%** | +38 / +39 / +38% | **+16 / +23 / +36%** |
| `keys_unsorted \| length` | +77 / +105 / +119% | **+22 / +29 / +25%** | +98 / +106 / +107% | **+23 / +22 / +21%** |

Median across all 25 configurations: M4 Pro **+25.1% → +4.8%**, 7950X **+38.1% → +12.2%**.
Head to head against `main`: **M4 Pro −8.5% median, 7950X −12.0% median (24 of 25 rows faster)**.
`users` and `arrays` sit inside ±2%. yq is flat on both boxes (median −0.2%, every row within
±1.8%) — `collapse` is `false` there and the gates compile away.

## What changed

| commit | |
|---|---|
| `f46866f8e` | one hash-set primitive replaces three sorts — including the correction that it must be *sized*, since shipping it growing was slower than the sort it replaced |
| `2c3aa652a` | the probe runs only in the two arms that read the answer positionally; `select`/`map`/the fallback stop paying for an answer they discard |
| `ecde8a7b4` | bare `keys_unsorted` dedups as it writes — "first occurrence wins" needs no lookahead |
| `c3108c9e3` | hash the key's raw span instead of decoding it through `as_str`'s two scans and a UTF-8 validation |
| `b0ddbbe35` | `uncons_key`: walk keys without materializing every value — fixes a regression `ecde8a7b4` introduced |
| `5e6a0ddb2` | sort `u64` hashes rather than probe a table, except where the walk streams |
| `b19c73b36` | the record, in `docs/plan/jq-duplicate-key-collapse.md` |

## Three things measurement overturned

- **The table was the wrong structure, and only one machine said so.** An open-addressed set
  replaced every sort first. On an M4 Pro that was right; on a 7950X it cost 24% on the identity
  path, where 7.1M keys make the table 134 MB against 32 MB of L3 per CCD. Sorting hashes beats
  both on both boxes. `f46866f8e`'s own message records the earlier sizing correction for the
  same reason.
- **The fix introduced a bigger regression than the one it removed.** With dedup forcibly
  disabled, `ecde8a7b4`'s walk cost +89%/+100% on its own — it routed the writer through a
  generic walk that materialized each field's *value*, then re-derived the key from a cursor.
  Building the binary a second time with the feature disabled is what found it; timings alone
  could not.
- **The measurement had a hole in it.** `jq type`, which touches no keys, moved 12-27% between
  binaries and did not bisect to the same commit on the two machines. At `codegen-units=1` it is
  +0.2%. Filed as #1587; every figure above was re-measured accordingly.

## Correctness

- Identity-gated against the pre-change binary over a 16-input duplicate-key corpus (escaped
  spellings both ways, a `\u`-escaped spelling of an ASCII key, an undecodable key, non-ASCII
  keys, all-duplicates, duplicates at head/middle/tail of an object wider than the printer's
  pairwise limit, nested and inside arrays) × 30 filters × 4 flag sets: **1984 configurations, 0
  differences**, at every commit.
- Spot-checked against `/usr/bin/jq` 1.7.1: the only divergence is the invalid-escape input jq
  rejects as a parse error, which is #1385's documented semi-indexing case and unchanged here.
- Full `cargo test` 54/54, both clippy legs, `cargo fmt --check`, `cargo check
  --no-default-features`, `cargo doc` with `-D warnings`.
- Two new tests pin the mid-walk switch to the exact collapsed list, whose resume offset a
  tail-duplicate case cannot exercise; five pin the hash primitive.

## Known trade

Bare `keys_unsorted` over a wide object roughly doubles peak RSS against `main` (100 MB: 189 →
389 MiB) — the streaming detector keeps 16 bytes per key where a sorted `Vec` keeps 8. `.` is
*better* than `main` (514 → 460 MiB), `length` is identical, `users`-shaped documents are
unchanged. Filed as #1588.

## Not fixed

`keys_unsorted` on a single 7.1M-key object stays ~+96% (ARM) / +142% (x86). At that width the
streaming table is 134 MB and every insert is a DRAM round trip — ~87 ns/key against a 670 ms
baseline query. Exact detection has no cheaper structure here; the sort is worse. Recorded in the
design doc as the floor.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -617,7 +617,7 @@ For detailed documentation on optimization techniques used in this project, see 
 - **Verify the box is idle; macOS load average lies** (read 1.4 on a machine using 2.4% CPU — it counts uninterruptible-wait threads). Never benchmark a laptop on battery.
 - **A benchmark cannot measure a shape it does not generate** — "all neutral" is not evidence. Add the generator pattern first.
 - **A share of runtime is not a comparison.** Name a baseline commit that predates the change being judged — not a sibling from the same series (#1514: "cut the probe from 10.4% to 3.9%" shipped inside a 2-3x regression).
-- **Measure a precheck where the guarded code is already fastest.** It is charged to every input including the ones it cannot help, so the workload with least work to save is where its cost shows (#1514: the same detector read +8% on record-shaped input and +134% on one wide object).
+- **Measure a precheck where the guarded code is already fastest.** It is charged to every input including the ones it cannot help, so the workload with least work to save is where its cost shows (#1514: the same detector read +14% on record-shaped input and +141% on one wide object).
 
 **Recent YAML optimizations:**
 - ✅ P2.5 (Cached Type Checking): 1-17% improvement depending on nesting depth
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -602,6 +602,8 @@ For detailed documentation on optimization techniques used in this project, see 
 - Duplicated predicates diverge silently — one definition, plus a test that the call sites agree (#106: three copies of one predicate, one of them quadratic)
 - Check where data physically lives before proposing to tag it (#106: the proposal assumed a `Vec<u32>`; no real file has one)
 - Re-derive a break-even before trusting it (#106: the issue's stated 12.5% dropped a bits-to-bytes conversion; the real gate was 3.125%)
+- A hash table is not automatically cheaper than a sort — above L3 a sort streams and a table does not, and which wins is architecture-dependent (#1514: the same table beat the sort on an M4 Pro and lost 24% to it on a 7950X at 7.1M keys)
+- To attribute a cost, build the binary again with the feature *disabled* and measure that — timings alone cannot separate what a check costs from what the code around it costs (#1514: it proved #1385's cost was 100% its probe, and caught a larger regression introduced by the fix)
 
 ### Benchmarking Discipline
 
@@ -614,6 +616,8 @@ For detailed documentation on optimization techniques used in this project, see 
 - **Memory-bound effects do not port across architectures** — measure ARM *and* x86_64, and name the chip in every table. #106 was 6.1x on M4 Pro, 16.4x on Zen 4.
 - **Verify the box is idle; macOS load average lies** (read 1.4 on a machine using 2.4% CPU — it counts uninterruptible-wait threads). Never benchmark a laptop on battery.
 - **A benchmark cannot measure a shape it does not generate** — "all neutral" is not evidence. Add the generator pattern first.
+- **A share of runtime is not a comparison.** Name a baseline commit that predates the change being judged — not a sibling from the same series (#1514: "cut the probe from 10.4% to 3.9%" shipped inside a 2-3x regression).
+- **Measure a precheck where the guarded code is already fastest.** It is charged to every input including the ones it cannot help, so the workload with least work to save is where its cost shows (#1514: the same detector read +8% on record-shaped input and +134% on one wide object).
 
 **Recent YAML optimizations:**
 - ✅ P2.5 (Cached Type Checking): 1-17% improvement depending on nesting depth
--- a/docs/guides/benchmarking.md
+++ b/docs/guides/benchmarking.md
@@ -877,6 +877,34 @@ Before trusting a neutral result, confirm the suite contains the shape the chang
 does not, add a pattern to `src/bin/succinctly/yaml_generators.rs` (or the JSON/DSV equivalent)
 and regenerate.
 
+### 8. A share of runtime is not a comparison, and one signal finds one regression
+
+`b6c0c3ca` was titled *"cut the duplicate-key probe from 10.4% to 3.9%"* and shipped inside the
+very series that #1514 later bisected as a 2-3x regression. Both figures are shares of the
+*post-change* runtime. They say the probe got cheaper relative to a binary that was already
+much slower than the baseline; they cannot say whether it got closer to it. Measured against
+the pre-change commit, that same commit recovered part of one regression and left the other
+slightly worse.
+
+A perf claim needs a **named baseline commit that predates the change being judged** — not a
+sibling from the same series, and not the current tip. "A benchmark claim is only as fresh as
+its baseline" is already on record for the broadword UTF-8 reversal, where a clean rebase
+silently improved the control side; this is the same failure with the staleness chosen rather
+than inherited.
+
+**Corollary for a detector** — anything that decides whether a fast path applies. Measure it on
+the workload where the guarded code was *already fastest*, not the one where it was slowest: a
+precheck is charged to every input including the ones it cannot help, so the input with the
+least work to save is where its cost shows most. #1514's duplicate-key detector cost ~3% of
+`sjq '.'` over a document of small objects and doubled `sjq keys_unsorted` over one holding a
+single wide object.
+
+**And pick more than one signal.** #1385 shipped two regressions on two paths. A `git bisect`
+on `keys_unsorted` found the evaluator commit and stepped straight over the print-path one,
+because `.` had barely moved at that commit; a second bisect on `.` was needed to find it. One
+signal proves one thing. Before believing a bisect result, ask which paths the suspected change
+touches and whether the signal is sensitive to each of them.
+
 ### Building both halves on a remote box
 
 A source-only tarball plus a reverse patch of the commit under test is enough, with no pushing:
--- a/docs/plan/jq-duplicate-key-collapse.md
+++ b/docs/plan/jq-duplicate-key-collapse.md
@@ -389,13 +389,165 @@ why patch coverage on `src/json/light.rs` read 15%. `raw_and_escaped`, from the 
 genuinely used and stays. Deleting them also left exactly one pairwise-scan threshold in the
 tree rather than three.
 
+## What the detector cost, and how it was made cheap (#1514)
+
+Open risk 1 above asked for the 4-6% to be confirmed on the pinned hosts. It was not 4-6%.
+Measured against `40c5ff93` on both boxes — interleaved, identity-gated, `codegen-units=1` (see
+"The measurement had a hole in it" below) — the detector cost up to **+141%** on a document
+holding one wide object, and *no object in that document carries a repeated key*.
+
+Two corrections to what this document says above, both of which it had the information to avoid:
+
+- **"The residue is the printer's alone."** It was not. The printer's share (`.`) was the smaller
+  half; `keys_unsorted` — the path that had no probe at all before this series and gained a whole
+  `census` walk — more than doubled, and `keys_unsorted | length` with it.
+- **The numbers came from `succinctly json generate`'s default shape.** A detector's cost shows
+  most on the workload where the guarded code is *already fastest*, which here is a single wide
+  object (82K keys at 1 MB, 7.1M at 100 MB). On the record-shaped `users` pattern the same
+  detector reads +14%, which is why it shipped.
+
+### Finding the cost: disable the detector, then measure
+
+Timings alone cannot separate "what the detector costs" from "what the code around it costs".
+Building each binary a second time with the detector *forcibly disabled* can, and the whole
+investigation turned on it:
+
+| `wide` `keys_unsorted`, Apple M4 Pro | 10 MB | 100 MB |
+|---|---|---|
+| `main`, probe disabled | **−1.1%** | **−0.7%** |
+| an intermediate fix, dedup disabled | +89.4% | +100.0% |
+| the same intermediate, dedup on | +112.3% | +154.2% |
+
+Row 1 proves #1514's premise: with the probe gone, `main` sits exactly on the pre-#1385
+baseline, so on that path the detector was the entire cost. Row 2 caught a regression introduced
+*by the fix* — see "The mistake the fix made".
+
+### Where the cost sat
+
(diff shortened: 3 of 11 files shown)


newhoggy:
Rebased onto `main` (21 commits). One conflict, and it was structural rather than textual, so
flagging it for review:

`ad122f88a` (*refactor(jq): extract fold_pipe_stages from eval_single's Pipe arm*) moved the pipe
loop into its own function — the same loop `perf(jq): run the collapse probe only where the
answer is read` rewrites. Resolved by taking `main`'s extraction and porting the dispatch change
into `fold_pipe_stages`, rather than resolving to either side. The `LazyKeys` arm there now reads
as it does in the commit message: probe only in the two arms that read the answer positionally,
`length` via `effective_len`, `.[]` and `map` dedup during their own walk, `first` and the
materializing fallback with no probe at all.

Re-verified on the rebased head (`e8aa34936`):

- identity gate against the pre-change binary: **1984 configurations, 0 differences**
- `cargo test` **54/54**, both clippy legs, `cargo fmt --check`, `cargo check
  --no-default-features`, `cargo doc` with `-D warnings` — all clean

The performance numbers in the description were measured before this rebase, on
`5e6a0ddb2`/`b0ddbbe35`. The rebase touched only the conflict above, but they have not been
re-measured against the new base; happy to re-run both boxes if you want the table to match the
merged head exactly.

PR #1606, opened 2026-08-25
perf(jq): stream the lazy-keys fan-out instead of census-then-collect (#1599)

### LINKED ISSUE #1599: jq: first(keys_unsorted | .[] | f) on an object pays an unconditional O(n log n) dup-key census
**Severity: Low-Medium** (performance only, no correctness impact; pre-existing since #1385, not a #1595 regression)

Follow-up to #1565/#1595, found while verifying #1595's own test-plan note about "a separate pre-existing, unrelated cost in constructing `LazyKeys` for a 2M-field object left as-is (out of scope)".

## The gap

#1595 made the *downstream* re-evaluation after a `LazyKeys`/`LazyIndexRange`/`LazySeq` prefix demand-aware, so `first(keys_unsorted | .[] | f)` on an **array** (`keys_unsorted` -> `LazyIndexRange`, pure arithmetic) is genuinely O(1) now. But on an **object** (`keys`/`keys_unsorted` -> `LazyKeys`), both `fold_lazy_keys_stage` and the new `each_lazy_keys_iterate_sink` gate *every* arm -- `First`/`Index(0)`/`Iterate` included -- behind an unconditional call to `collapsed_fields`, which runs `census`: walk every field, decode+fingerprint every key, then `sort_unstable()` the whole fingerprint vector, before ever checking which arm is being folded. This only happens in jq mode (`EvalSemantics::COLLAPSE_DUPLICATE_KEYS = true`, from #1385); `succinctly yq` skips it (`COLLAPSE_DUPLICATE_KEYS = false`).

`census` is *correctly* required for `length` (a repeated key changes the effective field count) and for `.[n]`/`last` (a removed duplicate shifts every later position) -- but `keys_unsorted | first` / `.[0]` doesn't need it at all: jq's collapse rule keeps a repeated key at its *first* occurrence, so the first document-order key can never be displaced by a later duplicate, census or no census. `test_duplicate_keys_lazy_keys_fast_paths_1385` already pins this invariant (`{"b":1,"a":2,"b":3} | keys_unsorted | first` => `"b"`, the first-seen key, unaffected by the later `"b"` duplicate) -- it just doesn't currently exploit it for speed.

Verified on a release build (2,000,000-key flat object vs. a 2,000,000-element array, identical query shape):

```
$ succinctly jq -c 'first(.k0)'                 obj.json   # 0.19s (baseline: parse + one field lookup)
$ succinctly jq -c 'first(keys_unsorted | .[])'  obj.json   # 0.44s  <- +0.25s, object
$ succinctly jq -c 'first(keys_unsorted | .[])'  arr.json   # 0.10s  <- +0.01s, array (already O(1) per #1595)
```

Cross-checked against `succinctly yq --jq-extensions` (`COLLAPSE_DUPLICATE_KEYS = false`, same `eval_generic.rs` engine, same 2M-key object): the census-free path adds only ~0.03s over baseline, isolating the remaining ~0.22s to `census`'s decode+sort.

Root cause: `fold_lazy_keys_stage` and `each_lazy_keys_iterate_sink` (`src/jq/eval_generic.rs`) both open with `if collapse { collapsed_fields(&fields) }` before dispatching on the requested arm. `collapsed_fields`/`census` (`src/jq/document.rs`) is O(n log n) regardless of outcome -- its own doc comment already says as much ("The probe itself goes through `census`, so answering `None` costs 8 bytes per field rather than a materialized field list" describes the memory saving, not the O(n log n) walk+sort time it still pays). Pre-existing since #1385; #1565/#1595 didn't introduce this and correctly scoped it out (their own acceptance criteria used an array, where `LazyIndexRange` never runs `census` at all), but it's why `first(keys_unsorted | .[] | f)` on an object doesn't reach the O(1) `first(.[] | f)` baseline the way the array case now does.

## Likely fix

Skip the census for the arms that are provably invariant under jq's first-position-wins collapse rule -- `Expr::Builtin(Builtin::First)` / `Expr::Index(0)` (`!sorted`) in both `fold_lazy_keys_stage` and `each_lazy_keys_iterate_sink` -- and only pay for it when an arm that actually needs it (`Length`, full `Iterate` fan-out, nonzero `.[n]`, `Last`) is reached. `sorted` (`keys`) still needs full materialization+sort regardless of this change, unaffected either way.

## Acceptance criteria

- `first(keys_unsorted | .[] | f)` and `first(keys_unsorted | first)` / `.[0]` on a wide object land within noise of the already-O(1) array case (`first(keys_unsorted | .[] | f)` on an equal-size array), not merely "the same cost as `first(keys | length)`" (today's baseline).
- `length`, `last`, nonzero `.[n]`, and `keys` (sorted) keep their existing, correct duplicate-key-aware behavior -- this is a narrowing of *when* the census runs, not a removal of it.
- No regression on #1385's own duplicate-key suite (`test_duplicate_keys_lazy_keys_fast_paths_1385` and its siblings) or #1565/#1595's own tests.

Refs #1565, #1595, #1385.

newhoggy:
## Summary

Two `keys_unsorted` shapes ran a whole-object `census` — decode and fingerprint every key, then sort the fingerprints — before they could answer, even though neither needs it under jq's first-occurrence-wins collapse rule.

**1. The demand-aware fan-out** (`each_lazy_keys_iterate_sink`). It ran `document::collapsed_fields` and then collected every key cursor into a `Vec`, *both before handing the first element downstream*, so `first(keys_unsorted | .[] | f)` on a wide object paid O(n log n) plus a full cons-list walk to produce one key. It now streams `DistinctKeyCursors` straight into the driver: "first occurrence wins" is an *online* rule, so the dedup rides along with the walk instead of preceding it — exactly what `fold_lazy_keys_stage`'s `Expr::Iterate` arm switched to in #1514.

**2. `keys_unsorted | .[0]`** (`fold_lazy_keys_stage`). `.[0]` is `first` by another spelling — collapsing keeps a repeated key at its *first* position, so document-order field 0 is the answer whatever the rest of the object holds. It was still taking the general positional arm and probing. A non-zero index genuinely can be displaced by an earlier key collapsing away, so that arm still probes, as do `length`, `last`, and sorted `keys`.

**3. `YamlFields::uncons_key`.** #1514 added the key-only walk but only JSON overrode it, so every YAML key walk went through the trait default — which calls `uncons` and therefore builds each field's *value* too. Added the override; this is what makes the streaming arm's cost model actually true on the yq path.

## Measurements

Release builds, the two binaries **interleaved within each repetition with the order flipped every rep**, min of 9 (min/median agree throughout), Apple M-series. jq: 2,000,000-key flat object. yq: 1,000,000-key mapping (15.7 MB).

| shape | before | after | delta |
|---|---|---|---|
| jq `first(keys_unsorted \| .[])` — early exit | 0.336s | 0.061s | **-81.8%** |
| jq `keys_unsorted \| .[0]` | 0.194s | 0.060s | **-69.0%** |
| jq `first(keys_unsorted \| .[] \| select(...))` — walks all, no early exit | 0.843s | 0.830s | -1.5% |
| yq `first(keys \| .[])` — early exit | 0.140s | 0.093s | **-34.0%** |
| yq `first(keys \| .[] \| select(...))` — walks all | 0.463s | 0.486s | **+5.0%** |

`keys_unsorted | .[0]` now matches `keys_unsorted | first` (0.060s vs 0.061s); `.[1]`, `length` and `last` are unchanged, still correctly paying for a census.

**The +5.0% is a real regression and the one deliberate trade here.** Full consumption through the streaming arm interleaves the walk with downstream evaluation instead of building a `Vec` in a tight loop first, which costs locality. The `YamlFields::uncons_key` override already halved it (+11.1% → +5.0%); the residue is inherent to streaming. Accepted because the same mode's early-exit shape gains 34%, but it is a judgement call rather than a free win — flagging it explicitly rather than burying it.

## Behaviour change (one, deliberate)

**The cursor a collapsed duplicate key carries moved from the last occurrence to the first.** `collapse_repeated` overwrote the surviving slot with the *last* occurrence's `key_cursor`, so `first(keys_unsorted | .[] | column)` on `{"b":1,"a":2,"b":3}` reported `14`; it now reports `2`. This is observable through the position builtins and through `anchor`/`style` in yq mode.

`2` is what the collecting path (`[keys_unsorted | .[] | column]` => `[2,8]`) has always reported and what jq's first-occurrence-wins rule points at, so this removes a disagreement between two spellings of the same query rather than creating one. Pinned by a dedicated test, since nothing else did.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde --no-fail-fast` — full suite
- [x] `cargo fmt --check`
- [x] All 22 new jq expectations diffed against the pinned oracle `/usr/bin/jq` 1.7.1
- [x] Output identity gated before believing any timing
- [x] Dual-mode: a committed `yq_cli_tests` case now covers the `collapse = false` branch (duplicates preserved: `["b","a","b"]`)

### On the tests, specifically

The first version of this PR's test was largely ineffective and the review caught it. Reaching the streaming arm is narrow: `limit(n; ...)` does **not** route through it (its elements arrive as owned strings, not cursors) and a collecting `[keys_unsorted | .[]]` takes `fold_lazy_keys_stage` instead — so 9 of 12 original rows never executed the changed code, and because a bare `first(...)` stops after one element, none reached `DistinctKeyCursors`'s repeat handling either. The rewritten test uses `first(keys_unsorted | .[] | select(...))` with predicates that match before, exactly at, and after a mid-stream duplicate, which forces the walk through `collapse_confirmed_repeat` and its resume-at-`yielded` path — the one branch that could reorder or drop a key. I verified reachability independently of the review: on the pre-PR binary the arm reports column `14` and the collecting path `2`, so a row that changes across the two binaries provably went through it.

Fixes #1599

### DIFF
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -2703,6 +2703,22 @@ fn fold_lazy_keys_stage<S: EvalSemantics, V: DocumentValue>(
                 GenericResult::ManyCursor(cursors)
             }
         }
+        // `.[0]` is `first` by another spelling, so it takes
+        // the `Builtin::First` arm's reasoning below rather
+        // than the general positional one: collapsing keeps a
+        // repeated key at its *first* position, so field 0 in
+        // document order is the answer whatever the rest of
+        // the object turns out to hold. Only a *non-zero*
+        // index can be displaced by an earlier key collapsing
+        // away, which is why the arm below still probes
+        // (#1599 -- #1514 moved the probe off the guard, but
+        // left this spelling paying it).
+        Expr::Index(idx) | Expr::IndexNumber { idx, .. } if !sorted && *idx == 0 => {
+            match fields.uncons_key() {
+                Some((_, key_cursor, _)) => GenericResult::OneCursor(key_cursor),
+                None => GenericResult::Owned(OwnedValue::Null),
+            }
+        }
         Expr::Index(idx) | Expr::IndexNumber { idx, .. } if !sorted => {
             // Positional: a collapsed object has fewer
             // members, so `.[n]` lands elsewhere. This is one
@@ -3052,8 +3068,7 @@ fn each_lazy_index_range_iterate_sink<S: EvalSemantics, V: DocumentValue>(
 /// Demand-aware `Expr::Iterate` fan-out for a `LazyKeys` (#1565).
 ///
 /// `!sorted` (`keys_unsorted`): streams [`DistinctKeyCursors`] straight
-/// into the driver, so a consumer that stops early walks only the fields it
-/// actually reached. This arm used to run `document::collapsed_fields`
+/// into the driver. This arm used to run `document::collapsed_fields`
 /// (whose probe is a whole-object `census`: decode and fingerprint every
 /// key, then sort the fingerprints) and then collect every key cursor into
 /// a `Vec`, both *before* handing the first element downstream -- so
@@ -3067,14 +3082,34 @@ fn each_lazy_index_range_iterate_sink<S: EvalSemantics, V: DocumentValue>(
 /// every cursor at once to build a `ManyCursor`, so it collects; this one
 /// drives elements one at a time and can consume the iterator directly.
 ///
-/// The yielded sequence is identical either way, which is why this is a
-/// cost change and not a behaviour change: `DistinctKeyCursors` emits first
-/// occurrences in document order, and on confirming a real repeat switches
-/// to the exact collapsed list resuming at the count already yielded --
-/// which lines up, because that list opens with those same first
-/// occurrences. Unlike the `LazySeq` arm below, streaming raises no
-/// atomicity question: producing a key cursor runs no user filter, so there
-/// is no per-element failure that eager collection would have masked.
+/// **The early exit is not unconditional.** It holds until a repeated key
+/// is *reached*: at that point `DistinctKeyCursors` calls
+/// `collapse_confirmed_repeat`, which walks the rest of the object and
+/// owns a `String` per distinct key. So `first` over an object whose
+/// second field repeats its first still pays a full walk. That is never
+/// worse than the old path, which paid `census` plus `collapse_repeated`
+/// unconditionally -- but it is not O(1) either, and sizing this arm's
+/// worst case from the paragraph above alone would get it wrong.
+///
+/// The *sequence* of keys is identical to what collecting produced, and
+/// for the same reason the resume is sound: `DistinctKeyCursors` emits
+/// first occurrences in document order, and on confirming a real repeat
+/// switches to the exact collapsed list resuming at the count already
+/// yielded -- which lines up, because that list opens with those same
+/// first occurrences. Unlike the `LazySeq` arm below, streaming raises no
+/// atomicity question: producing a key cursor runs no user filter, so
+/// there is no per-element failure that eager collection would have
+/// masked.
+///
+/// **The cursor a collapsed key carries did change**, which is why this is
+/// not purely a cost change. `collapse_repeated` overwrites the surviving
+/// slot with the *last* occurrence's `key_cursor`, so the old path
+/// reported the later position; streaming reports the first occurrence.
+/// Observable through the position builtins and through `anchor`/`style`
+/// in yq mode. The new answer is the one the collecting path has always
+/// given and the one jq's first-occurrence-wins rule points at, so this
+/// settles a disagreement between two spellings rather than creating one
+/// -- pinned by `test_lazy_keys_streaming_reports_first_occurrence_cursor_1599`.
 ///
 /// `sorted` (`keys`): lexicographic order needs every key decoded and
 /// sorted first, unavoidably -- the same cost `materialize_lazy_keys` always
@@ -3084,7 +3119,7 @@ fn each_lazy_index_range_iterate_sink<S: EvalSemantics, V: DocumentValue>(
 /// existing behaviour of not preserving a cursor for sorted keys (no new
 /// capability, no regression, just demand-aware instead of eager).
 fn each_lazy_keys_iterate_sink<S: EvalSemantics, V: DocumentValue>(
-    fields: V::Fields,
+    fields: &V::Fields,
     sorted: bool,
     collapse: bool,
     rest: &[Expr],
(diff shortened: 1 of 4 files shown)


newhoggy:
Updated after `/code-review`, which returned 9 findings. I reproduced the significant ones independently before acting; **6 were real** and are fixed in `928dbed23` / `f28344ec2`:

- **`.[0]` still paid the census** (0.194s vs `first`'s 0.061s). #1599 names it in its acceptance criteria, and my original scope claim — that #1514 had already freed `Index(0)` — was simply wrong. Now -69%, and `.[1]`/`length`/`last` still correctly probe.
- **A real yq regression my evidence missed.** Both rows I offered as "full fan-out unchanged" (`[keys_unsorted|.[]]|length`, `reduce`) never enter the changed function at all, so I had reported a clean result for a path I never measured. The shape that does reach it regressed **+11.1%**. Root cause was finding 6: `YamlFields` never overrode `uncons_key`, so #1514's key-only walk was a no-op for YAML. Adding the override halved it to **+5.0%**, which is now stated in the PR body as a deliberate trade against -34% on the same mode's early-exit shape rather than buried.
- **The test pinned almost nothing it claimed to.** `limit(n; ...)` does not route through the arm, and a bare `first(...)` stops before the repeat handling — so 9 of 12 rows never ran the changed code and none reached the switch-and-resume branch. Rewritten around `first(keys_unsorted | .[] | select(...))` with predicates before/at/after a mid-stream duplicate. I established reachability independently of the review: the pre-change binary reports a collapsed key's cursor as column 14 where the collecting path reports 2, so a row that moves across the two binaries provably went through the arm.
- **Not a "cost change only", as I claimed.** The collapsed-key cursor moved from last to first occurrence (`column` 14 → 2). It agrees with the collecting path and with jq's first-wins rule, so it resolves an inconsistency — but it was unpinned and undocumented. Now both.
- Doc corrected on the unconditional-early-exit overstatement, `fields` takes `&V::Fields`, and a yq-mode row now covers the `collapse = false` branch.

Findings 5/7/8/9 were also real and are fixed. Full suite re-run: **5867 passed, 0 failed**; both clippy legs, `fmt --check`, and the no_std build are clean.

Not adopted: nothing — every finding was either fixed or (finding 2's residual +5%) explicitly documented as an accepted trade.

newhoggy:
Filed #1609 to track the accepted +5.0% yq full-consumption trade documented above, so it is a tracked item rather than only a PR-body footnote. It records what has already been ruled out (half the original +11.1% was `YamlFields` never overriding `uncons_key`, fixed here) and flags that the obvious fix — batching to recover locality — trades directly against this PR's -34% early-exit win, so it needs measuring at several batch sizes rather than assuming.

Issue #1565, opened 2026-08-23
jq: first(keys|.[]|f) / first(map(f)|.[]|f) still O(n), leaks side effects


newhoggy:
**Severity: Low** (stray side-effect writes; no wrong output, no data loss -- same class as #1459)

Follow-up to #1461, found during code review of its fix.

## The gap

#1461 gave `eval_generic.rs`'s `first(...)` a demand-driven sink
(`eval_each_generic`/`eval_each_pipe_generic`) for `Comma`/`Pipe`/`Paren`, closing
`first(.[] | stderr)` and friends. But when the pipe's *first* stage evaluates to
`GenericResult::LazyKeys`/`LazyIndexRange`/`LazySeq` (i.e. `keys`, `keys_unsorted`, an
array's own `.[]`/`keys` under `#684`, or a `map(f)` chain under `#724`/`#725`) and 2+
further stages follow, `eval_each_pipe_generic`'s driver hands the whole remaining pipe
to `fold_pipe_stages` -- the plain eager fold with no demand check at all -- instead of
continuing through the sink. So `first` still evaluates every element before it can stop,
for this one shape.

Verified against pinned jq 1.7.1 and this branch (post-#1461):

```
$ echo '{"a":1,"b":2,"c":3}' | jq -c 'first(keys | .[] | stderr)'
"a"     stderr: a
$ echo '{"a":1,"b":2,"c":3}' | succinctly jq -c 'first(keys | .[] | stderr)'
"a"     stderr: abc          <- stderr fired for all 3 keys, not just "a"

$ echo '[1,2,3]' | jq -c 'first(map(.+1) | .[] | stderr)'
2       stderr: 2
$ echo '[1,2,3]' | succinctly jq -c 'first(map(.+1) | .[] | stderr)'
2       stderr: 234          <- stderr fired for all 3 elements, not just the first
```

stdout matches jq exactly in both cases -- only the extra `stderr` writes diverge, so
this is a stray-side-effect bug, not silent data loss. Confirmed **pre-existing**: built
`main` at the commit #1461 based off (`379d437af`) and got byte-identical eager output
there, so this is not a regression from #1461 -- the shape was equally eager before, via
the old `first_over_comma_generic`/`first_over_pipe_prefix_result`'s 2-stage-only cap.

On a 2M-element array this is also a real perf cliff, not just a stderr nuisance:
`first(keys | .[] | .+1)` is ~12x slower than the now-fixed `first(.[] | .+1)`, and
[...]

PR #1610, opened 2026-08-25
perf(jq): build the owned rest-pipe once per driver, not per element (#1598)

### LINKED ISSUE #1598: jq: continue_pipe_element_generic deep-clones the remaining pipe AST per element
**Severity: Low** (latent — no live consumer reaches the per-element case today)

Found while reviewing [#1565](https://github.com/rust-works/succinctly/issues/1565)'s fix.

## The gap

`continue_pipe_element_generic`'s `GenericItem::Owned` arm (`src/jq/eval_generic.rs`) rebuilds the
remaining pipe as an owned AST for every item it is handed:

```rust
GenericItem::Owned(o) => {
    let rest_pipe = Expr::Pipe(rest.to_vec());
    eval_each_owned::<S>(&rest_pipe, &o, optional, &mut |o| sink(GenericItem::Owned(o)))
}
```

`rest.to_vec()` is a deep clone of every remaining stage's `Expr` tree — a `Vec` allocation plus
one recursive clone per stage, per element.

Before #1565 this arm ran once per *pipe*, from `eval_each_pipe_generic`'s driver. #1565 extracted
it and put it on a per-*element* path: `drive_pipe_elements_generic` calls
`continue_pipe_element_generic` in a loop, and for `each_lazy_index_range_iterate_sink` (and
`each_lazy_keys_iterate_sink`'s `sorted` branch) *every* item is an `Owned`, so every iteration
pays a fresh clone.

`fold_pipe_stages_sink`'s own `Owned` arm has the same `Expr::Pipe(stages[j..].to_vec())`, but that
one is genuinely once-per-fold and not a concern.

## Why it isn't visible yet

The only consumer that drives this sink today is `first(...)` via `each_take_first_generic`, which
stops after one element — so the loop runs once and the clone happens once. The cost is latent: it
becomes O(n) deep AST clones the moment a draining consumer (or a `limit(n; ...)`-style
partial one) uses the same path, which is exactly what
[#1596](https://github.com/rust-works/succinctly/issues/1596) and
[#820](https://github.com/rust-works/succinctly/issues/820)'s remaining stages would introduce.

No measurement is offered because there is currently no query that exercises the loop more than
once — that is the point of filing it before one exists.

## Likely fix

Hoist the `Expr::Pipe` construction out of the per-element loop: build it once in
`drive_pipe_elements_generic` (or pass `rest` through and let `eval_each_owned` take a slice
rather than an `&Expr`) so the clone is amortized over the fan-out instead of repeated per
element.

`eval_each_owned` taking a stage slice is the cleaner of the two and removes the synthesized
`Expr::Pipe` wrapper entirely, but it is an `eval.rs` signature change — worth checking whether
its other callers can supply a slice as cheaply.

## Acceptance criteria

- `continue_pipe_element_generic`'s `Owned` arm no longer allocates or deep-clones per element.
- No regression on `test_first_over_lazy_prefix_applies_every_stage_1565`,
  `test_first_over_pipe_prefix_dispatches_every_generic_result_shape_1451`, or
  `test_short_circuit_side_effect_shapes_already_match_jq_820`.

Refs #1565, #1461, #1450.

newhoggy:
## Summary

`continue_pipe_element_generic`'s `Owned` arm rebuilt the remaining pipe as an owned AST for **every item handed to it** — `Expr::Pipe(rest.to_vec())` is a `Vec` allocation plus a recursive `Expr` clone per stage. Before #1565 that arm ran once per pipe; #1565 moved it onto a per-*element* path, so both drivers that reach it pay the clone once per element.

The slice and its owned copy are now one `RestPipe` value, built at most once per driver and only if an `Owned` element actually arrives. Pairing them removes a hazard as well as the cost: correctness requires that a cached pipe is only ever used with the `rest` it was built from, and as two loose parameters that mispairing compiled cleanly and produced a wrong answer rather than a crash.

`eval.rs::eval_each_pipe` — the twin driver, and the function this arm hands off to — had the same eager clone and is made lazy the same way. This repo has previously fixed one of these two evaluators and missed the other, so they move together.

## This is not latent — it is measurable today

The issue filed this as latent, reasoning that no query drives the loop more than once and therefore offering no measurement. That holds only for a **bare** `first(...)`. A `first(...)` whose predicate never matches keeps pulling, and `keys` (sorted) yields every key as an `Owned` — so `first(keys | .[] | select(. == "zzz"))` already pays one clone per key on current `main`.

**Apple M5 Max (arm64)**, 200,000-key object, release builds, interleaved within each repetition **with the order flipped every rep**, 5 reps (min / median):

| shape | before | after | delta |
|---|---|---|---|
| owned driver, `rest` = 1 stage | 0.132s | 0.110s | **-16.8% / -16.5%** |
| owned driver, `rest` = 7 stages | 0.148s | 0.112s | **-24.6% / -25.7%** |
| borrowed chain, 1 / 4 / 8 stages | 0.107 / 0.251 / 0.474s | — | -0.4% / +0.7% / -0.3% (neutral) |

The gain **growing with stage count** is the mechanism's own signature — a longer `rest` is a bigger tree to clone per element.

⚠️ **Single-architecture result.** `docs/STYLE_GUIDE.md` STYLE-0008 and CLAUDE.md both require naming the chip and measuring ARM *and* x86_64; the first revision of this PR named neither, which the review flagged. The chip is named now, but these are ARM-only — the x86_64 leg needs one of the pinned benchmark hosts, which this run did not have. Treat the magnitudes as ARM figures pending an x86 confirmation. The change removes an allocation, and allocator behaviour is exactly the class CLAUDE.md warns can differ across architectures.

## Two proposals measured and *not* taken

Both came from review, both were implemented and benchmarked rather than accepted on the reasoning:

1. **`eval.rs` lazy clone** — kept, but **measured neutral** (numbers above). The review attributed the superlinear cost curve on borrowed chains to this clone; rebuilding with the allocation gone moves nothing, so that curve is the per-stage evaluation itself. Kept as dead-work removal, and its comment says so explicitly so nobody later cites it as a speed fix.
2. **Empty-`rest` shortcut** (skip the serialize-and-reindex bridge when `Expr::Iterate` is the pipe's last stage) — **implemented, measured, dropped.** Neutral on every shape that should reach it: `[map({a:.,b:.,c:.,d:.}) | .[]] | length` +1.6%, `[keys | .[]] | length` +0.6%, index-range -2.0%. The 0.196s-vs-0.360s gap that motivated it is real but is not this bridge, since removing the bridge does not close it. Its fidelity was also checked properly first — the review verified number spelling *with* the round trip in place, which does not show that removing it preserves representation; implementing it and diffing `2.50`, `1.0`, `-0.0`, `1e1000`, `1e-1000`, `4e4`, 20-digit integers, unicode and escapes against jq 1.7.1 does, and it did.

## Test

Rewritten to say what it actually establishes. The previous doc claimed to pin the cache-leak invariant; **every row passes identically on the pre-change binary**, because that leak is not expressible once `RestPipe` owns both halves. It is a forward guard against someone widening the cache's scope, not a regression pin, and it now says so.

Widened from one driver to four, since the original covered only the convenient one: the sorted-`keys` driver, a bare comma stream (the only shape that exercises the *second* cache, in `eval_each_pipe_generic` — the `keys` rows leave it untouched), the index-range driver, and the lazy-seq driver. All expectations are jq 1.7.1's own output.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde --no-fail-fast` — **8989 passed, 0 failed** (63 binaries)
- [x] `cargo fmt --check`
- [x] Behaviour diffed against the pinned oracle `/usr/bin/jq` 1.7.1 directly, not just goldens
- [x] Output identity gated before trusting any timing

Fixes #1598

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -3114,9 +3114,21 @@ fn eval_each_pipe<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     // to stop stage 1 too -- that is the whole point of the lazy `Pipe` arm --
     // but the two cases surface differently to our own caller.
     let mut downstream: Option<Flow> = None;
-    // Built once, outside the driver closure, so an Owned item mid-pipe
-    // doesn't re-clone `rest` per value.
-    let rest_pipe = Expr::Pipe(rest.to_vec());
+    // Built at most once, and only if an `Owned` item actually arrives
+    // (#1598). This used to be unconditional, so an all-`Borrowed` chain --
+    // the common case -- allocated a `Vec` and deep-cloned every remaining
+    // stage that nothing then read, once per call, and this function is
+    // itself the per-item recursion below.
+    //
+    // Measured neutral (Apple M5 Max, 200k elements, 1/4/8-stage borrowed
+    // chains: -0.4% / +0.7% / -0.3%), so this is dead-work removal for
+    // clarity, not a speed fix -- do not cite it as one. The superlinear
+    // cost curve those chains do show is the per-stage evaluation itself,
+    // not this clone: rebuilding with the allocation gone moves nothing.
+    // `eval_generic`'s twin of this driver carries the same cache as a
+    // `RestPipe`; here the `Owned` arm is the only reader, so a plain
+    // `Option` is enough.
+    let mut rest_pipe: Option<Expr> = None;
     let upstream = {
         let mut driver = |item: Item<'a, W>| -> Demand {
             let flow = match item {
@@ -3129,7 +3141,8 @@ fn eval_each_pipe<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
                 // (#820: an eager fallback here silently drained inputs a
                 // downstream sink never asked for).
                 Item::Owned(v) => {
-                    eval_each_owned::<S>(&rest_pipe, &v, optional, &mut |o| sink(Item::Owned(o)))
+                    let rest_pipe = rest_pipe.get_or_insert_with(|| Expr::Pipe(rest.to_vec()));
+                    eval_each_owned::<S>(rest_pipe, &v, optional, &mut |o| sink(Item::Owned(o)))
                 }
             };
             match flow {
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -3036,22 +3036,17 @@ fn drive_pipe_elements_generic<S: EvalSemantics, V: DocumentValue>(
     optional: bool,
     sink: &mut dyn FnMut(GenericItem<V>) -> Demand,
 ) -> Flow {
-    // One cache for the whole drive, so a `rest` pipe that has to be owned
-    // for an `Owned` element is built on the first one and reused by every
-    // element after it (#1598).
-    let mut rest_pipe: Option<Expr> = None;
+    // One `RestPipe` for the whole drive, so an owned copy is built on the
+    // first `Owned` element and reused by every element after it (#1598).
+    let mut rest = RestPipe::new(rest);
     for element in elements {
         match element {
-            Ok(item) => match continue_pipe_element_generic::<S, V>(
-                item,
-                rest,
-                optional,
-                &mut rest_pipe,
-                sink,
-            ) {
-                Flow::Exhausted => continue,
-                other => return other,
-            },
+            Ok(item) => {
+                match continue_pipe_element_generic::<S, V>(item, &mut rest, optional, sink) {
+                    Flow::Exhausted => continue,
+                    other => return other,
+                }
+            }
             Err(Control::Error(e)) => return drain_result_generic(GenericResult::Error(e), sink),
             Err(Control::Break(label)) => {
                 return drain_result_generic(GenericResult::Break(label), sink);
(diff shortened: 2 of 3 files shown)


newhoggy:
Updated after `/code-review` (10 findings, no correctness bugs). I implemented and **measured** the perf proposals rather than accepting their reasoning, and two of them did not survive that:

**Acted on:**
- **Finding 6** — the cache and its `rest` were two loose parameters with an unenforceable pairing invariant. Now one `RestPipe` value, so the mismatch is unrepresentable; the extra parameter is gone too.
- **Finding 1** — `eval.rs::eval_each_pipe` had the same eager clone. Fixed (this repo has shipped a fix to one of the twin evaluators and missed the other before). **But its attribution was wrong:** the review pinned a superlinear cost curve on borrowed chains to this clone, and rebuilding with the allocation removed moves nothing (-0.4% / +0.7% / -0.3% at 1/4/8 stages). The curve is the per-stage evaluation. Kept as dead-work removal, labelled neutral in the code so it is not cited as a win later.
- **Findings 7 + 8** — both correct. Every row of my test passed identically on the pre-change binary, so the doc claiming to pin the leak was overstating it; reworded as a forward guard. Widened from one driver to four, including the bare comma stream that is the only shape exercising the second cache.
- **Finding 10** — correct and my mistake: the table named no chip and one architecture, against STYLE-0008 and CLAUDE.md. Chip named (Apple M5 Max, arm64) and the single-architecture limitation is now called out explicitly rather than left implicit.
- **Finding 9** — the lazy arm genuinely cannot use the cache (`fold_pipe_stages_sink` rebuilds from `stages[j..]`, a different slice); noted in-code instead of silently leaving it.

**Not taken, after measuring:**
- **Finding 2** (empty-`rest` shortcut) — implemented, benchmarked, reverted. Neutral on every shape that should reach it (+1.6% / +0.6% / -2.0%). The 0.196s-vs-0.360s gap is real but is not this bridge. Worth flagging on method: the review verified number fidelity *with* the round trip in place, which does not show that removing it preserves spelling — I implemented it and diffed `2.50`/`1.0`/`-0.0`/`1e1000`/`1e-1000`/`4e4`/20-digit ints/unicode/escapes against jq 1.7.1 to actually check. It did hold, but that was not yet established.
- **Finding 3** (unwrap single-stage `rest` for `eval_owned_fast_path`) — not attempted. Given findings 1 and 2 both had sound mechanisms that measured neutral, I did not want to add a third speculative change to this PR on the same reasoning.
- **Findings 4, 5** — real and out of scope (deeper per-element recursion, and `needs_path_context` re-walking). Left alone rather than widened into here.

Headline result is unchanged and re-verified: **-16.8% at 1 trailing stage, -24.6% at 7**, the gain growing with stage count as the mechanism predicts. Full suite **8989 passed / 0 failed**.

newhoggy:
Patch coverage: **94.59% (35/37 new lines)**. `src/jq/eval.rs` 100% (3/3); `src/jq/eval_generic.rs` 94.12% (32/34).

The two uncovered lines are `continue_pipe_element_generic`'s `GenericItem::One` arm, and they are **pre-existing-uncovered rather than newly uncovered** — they only register as "new" because the arm went from one line to two under rustfmt when `rest` became `rest.stages()`. The dispatch is otherwise unchanged, so reachability is unchanged.

Established from the raw lcov hit counts rather than by assuming: within the same `match`, the sibling arms record 43 / 42 / 24 hits and this one records **0**, so nothing in the suite drives a `One` item into this function. It is not statically dead — `GenericItem::One` is constructed at two sites — but every path that builds one hands it straight to a sink (`push_one_generic`) or converts it (`generic_item_to_result`) rather than routing it back through here.

I did not write a test for it: any repro would have to be reverse-engineered from the dispatch rather than from a real query shape, which would pin the implementation rather than a behaviour, and it would be covering code this PR does not change. Flagging it as a genuine pre-existing gap instead — worth a look by whoever next touches that dispatch, since an arm with no reachable caller is either dead or a missing capability.

newhoggy:
Rebased onto current `main` (6 commits landed while this was in review — #1194's malformed-object-member series, which touches `src/jq/eval_generic.rs` and `tests/jq_cli_tests.rs`, i.e. exactly the files this PR changes). I had enqueued before checking that, which was the wrong order; dequeued, rebased, re-verified, re-pushed.

The rebase was conflict-free, which is precisely when a semantic break hides, so everything was re-run rather than assumed:

- Both `continue_pipe_element_generic` call sites are intact and #1194 added no new ones.
- Full suite after rebase: **5882 passed, 0 failed**; both clippy legs, `fmt --check`, no_std all clean.
- Perf re-measured against a **freshly built baseline from the new `main` tip**, not the pre-rebase one: **-15.1%** at 1 trailing stage, **-24.8%** at 7 (Apple M5 Max, 200k-key object, interleaved with order flipped per rep, 5 reps). Output byte-identical to that new baseline.

Ready to re-enqueue once CI is green on `c4934f6ad`.

Issue #1591, opened 2026-08-23
jq: resolve_recurse drops a TrackedVar snapshot mark from f's own output, refusing where jq accepts


newhoggy:
**Severity: Low** (refuse-only — succinctly declines a filter jq accepts, visibly and without
touching any document; no data-corruption risk)

## Summary

`resolve_recurse`'s own doc comment claims "`snapshot` is uniformly `false` across this walk
... every child below is reached by genuine navigation, so no frozen variable snapshot can
enter it" (#1466). That's not quite true: `recurse`'s own traversal function `f` is resolved
against each visited node via `resolve_against_cow::<S>(f, current, trackable)`, and if `f`
references a frozen `. as $x` binding (`Expr::TrackedVar`) whose value doesn't match the
*current* node, that arm marks the resulting branch `snapshot: true` — a real, provenance-
carrying value, not a reconstruction. `resolve_recurse` then destructures that child with `..`
(dropping `snapshot`), and both sites that queue it for further traversal/output hardcode
`snapshot: false` — silently discarding the mark.

If that value later flows into an enclosing fold's `FoldRegister` (whose own register is `$x`'s
own frozen position), the lost mark means `FoldRegister::identical` can no longer recognise it,
so succinctly refuses a path jq accepts.

## Repro (verified against pinned jq 1.7.1; `limit` bounds an otherwise-infinite `recurse`)

```
$ echo '{"a":{"v":1}}' | jq -c \
    'path(. as $x | reduce (1) as $i (.; limit(3; recurse(if . == $x then .a else $x end)) | select(. == $x)))'
[]
$ echo '{"a":{"v":1}}' | succinctly jq -c \
    'path(. as $x | reduce (1) as $i (.; limit(3; recurse(if . == $x then .a else $x end)) | select(. == $x)))'
jq: error (at <stdin>:1): Invalid path expression with result {"a":{"v":1}}   # exit 5
```

Write side — jq accepts and writes, succinctly refuses and leaves the document untouched
(the safe direction, unlike #1466's own bug):

```
$ echo '{"a":{"v":1}}' | jq -c \
    '(. as $x | reduce (1) as $i (.; limit(3; recurse(if . == $x then .a else $x end)) | select(. == $x))) |= 9'
9
$ echo '{"a":{"v":1}}' | succinctly jq -c \
    '(. as $x | reduce (1) as $i (.; limit(3; recurse(if . == $x then .a else $x end)) | select(. == $x))) |= 9'
[...]

Issue #668, opened 2026-08-08
perf(jq): push_recursive_branches/resolve_recurse still deep-clone O(subtree) per '..'/recurse node in path()/write context


newhoggy:
**Severity: Medium** — split out of #626, whose original diagnosis this was. #626's own benchmark turned out to exercise a different function (`eval_index_expr`, corrected there); this issue is the real, still-open bug that diagnosis originally described.

### Where

`push_recursive_branches` (the `path(..)`/bare-`recurse` resolver) and `resolve_recurse` (the `recurse(f)`/`recurse(f;cond)` resolver), both in `src/jq/eval.rs`, do `value.clone()`/`current.clone()` once per visited node to build that node's `PathBranch`. `OwnedValue::Array`/`Object` are plain `Vec<Self>`/`IndexMap<String, Self>` with no structural sharing, so `.clone()` on a compound value is a full deep copy of everything still below it.

For a linear-nesting document (one child per level, depth *d*), the node at depth *i* clones a subtree of size *O(d − i)*. Summed over all *d* nodes, that's *O(d²)* — independent of what (if anything) the rest of the pipe does with the value afterward. `resolve_dynamic_indexes`, the sole consumer of this machinery, discards every branch's value once it has extracted the branch's path components (`.map(|(components, _)| ...)`), so for the common case of `path(..)`-shaped queries this clone cost is pure waste: the value is never read again.

Unlike a bare `.. | .[.k]?` filter, these two functions are reachable **only** through `resolve_node` → `resolve_dynamic_indexes`, whose 4 callers are `eval_assign` (`=`), `eval_update` (`|=`), `builtin_path` (`path()`), `builtin_del` (`del()`) — so the bug needs `path(..)`, `path(recurse(...))`, or one of the write operators to trigger, not a bare filter.

### Before fixing: measure it

Nothing currently benchmarks this specific bug — #626's own benchmark (`.. | .[.k]?`, no `path()`/assignment) doesn't reach these functions at all (see #626's correction). Per this repo's own practice (#661), add a `path(..)`- or assignment-based linear-nesting benchmark first, confirm the quadratic signature, and use it as the before/after gate — don't assume the #626 numbers transfer.

### Why this is a separate, larger fix

Eliminating the clone requires either:
[...]

PR #1631, opened 2026-08-25
perf(jq): replace path-prefix Vec<Expr> clones with a persistent Rc chain

### LINKED ISSUE #701: perf(jq): push_recursive_branches/resolve_recurse still pay O(path-length) per node building Vec<Expr> prefixes, independent of #668's value clone
**Severity: Low-Medium** — found while verifying #668's fix, split out per this repo's own practice for #626→#668→#675.

### Where

`push_recursive_branches` and `resolve_recurse` (`src/jq/eval.rs`, the same two functions #668 fixed) do `prefix.to_vec()`/`prefix.clone()` **once per visited node** to build that node's static path (`Vec<Expr>`). For a node at depth `i`, that clone is `O(i)`. Summed over all `d + 1` nodes in a depth-`d` linear-nesting document, that's `O(d²)` — independent of, and additional to, the value-clone `O(d²)` #668 fixed.

`resolve_dynamic_indexes`'s `assemble` helper (same file) does the same thing on the way out: `components.into_iter().map(strip_resolved_optional).collect()` per branch, another `O(path length)` per branch, `O(d²)` total. `builtin_del`'s per-path `delete_at_path` walk (one call per resolved path, each `O(path length)`) is a third `O(d²)` contributor, though that one may be closer to inherent-to-the-operation than fixable — see "Why this is a separate issue" below.

### How this was found

#668 fixed the `Cow`-based elimination of the *value* clone in these two functions (PR TBD). `benches/jq_recurse_clone_bench.rs` (`del(..)`, depths 100–400) showed a real, statistically significant, depth-growing improvement (53% → 65% faster, `p = 0.00` at every depth — the growing-percentage signature `docs/guides/benchmarking.md` documents as evidence of a removed algorithmic term) — confirming #668's fix works exactly as designed. But the benchmark's *absolute* scaling stayed close to quadratic (measured exponent ~2.16 before → ~1.90 after, not the ~1.0 the issue's acceptance criteria hoped for), because these independent path-list costs were still there and now dominate.

### Why this is a separate issue, not a #668 follow-up-in-place

#668's own text scoped its fix specifically to the *value* half of each `PathBranch` — every code sample and complexity argument in that issue is about `OwnedValue::clone()`, never about the `Vec<Expr>` path-component list. Fixing this would need a different technique (e.g. a persistent/shared prefix representation — `Rc<PathSegment>` cons-list or similar — materialized into a flat `Vec<Expr>` only where a component list must actually leave this machinery), which is its own design space, separate from #668's `Cow<OwnedValue>` approach.

`delete_at_path`'s per-path apply cost may not be fixable at all without batching deletes (applying all `d + 1` resolved paths in one combined tree walk instead of `d + 1` independent root-to-leaf walks) — a materially bigger change than the path-list-construction fix above, and arguably correct behavior for "delete this specific path" applied repeatedly rather than a bug. Worth measuring in isolation (`path(..)`, which never calls `delete_at_path`) before deciding whether it's in scope at all.

### Acceptance

- [ ] A benchmark (reusing or extending `benches/jq_recurse_clone_bench.rs`) isolates the path-list-construction cost from the delete-apply cost — e.g. compare `path(..)` (no delete) against `del(..)` at the same depths, so the two O(d²) contributors can be told apart.
- [ ] `push_recursive_branches`/`resolve_recurse` no longer pay `O(path length)` per visited node purely to build that node's path prefix, where a cheaper shared representation is possible.
- [ ] The isolated (non-delete) benchmark shows sub-quadratic scaling with depth after the fix.
- [ ] No behavioral change to any existing path-resolution golden fixture (same corpus #668 verified against).
- [ ] A decision, backed by measurement, on whether `delete_at_path`'s per-path apply cost is in scope for this issue or a distinct one (or not fixable without a batched-delete rewrite).

### Related

Found verifying #668's fix. Same functions, same file region (`src/jq/eval.rs` `resolve_node`/`push_recursive_branches`/`resolve_recurse`/`resolve_dynamic_indexes`). Same discovery pattern as #626→#668→#675 (benchmark reveals a cost the original issue didn't attribute correctly) — this one didn't misattribute, it just found an *additional*, independent cost while confirming the fix.

newhoggy:
## Summary

- `push_recursive_branches`/`resolve_recurse`, `resolve_seq`, `resolve_index_expr`, `resolve_slice_expr`, and `FoldRegister` each rebuilt `PathBranch.path`/`FoldRegister.path` as a fresh `Vec<Expr>` on every node visited during path-tracking evaluation (`path(..)`, `del(..)`, `.. = x`, `.. |= f`, `recurse(f)`, `reduce`/`foreach`) — an O(depth)-per-node clone summing to O(d²) on a depth-`d` document. This is the path half of the cost #668 left unfixed when it moved the *value* half to `Cow`.
- `PathBranch.path` is now `Rc<PathPrefix>`, a persistent cons-list: extending it by a component is an O(1) refcount bump + one allocation instead of an O(depth) `Vec` clone. Flattening to a real `Vec<Expr>` now happens exactly once per branch, only where one is actually required (`resolve_dynamic_indexes`'s `assemble`), instead of being repeated at every node during construction.
- Also fixes a pre-existing, unrelated bug found while validating this: `benches/jq_recurse_clone_bench.rs`'s depth-400 case panicked (`MAX_VALUE_TREE_DEPTH` is 384), invisible until now since `cargo bench` isn't run by CI. Depths are now derived from the constant.
- Rebased onto #1591 (`fix(jq): thread the TrackedVar snapshot mark through the whole path resolver`), which landed on `main` mid-implementation and touched the same functions — the new `snapshot`-ambient-parameter threading and `PathBranch::passthrough` constructor are preserved, adapted to the new `Rc<PathPrefix>` representation.

Fixes #701.

## Remaining acceptance-criteria items (not attempted in this PR)

- Trustworthy sub-quadratic scaling evidence for the isolated construction cost — the dev machine used for implementation was running 47+ concurrent agent sessions, so A/B numbers gathered there were noisy/non-monotonic and are not included as evidence. Needs an idle machine or the project's dedicated bench boxes.
- The measured decision on whether `delete_at_path`'s separate O(pathlen²)-per-call cost (reached only via `builtin_del`'s single-static-path branch, a different code path than `del(..)`'s recursive-descent case) needs its own follow-up issue.

## Test plan

- [x] `cargo build --features cli,simd,regex,serde` / `cargo check --tests --benches`
- [x] `cargo fmt --check`
- [x] Full `cargo test` suite (2833 tests, 0 failed) including `jq_golden_tests`, `yq_golden_tests`, and `jq_recurse_depth_tests` (depth-300 correctness pin for the changed functions) — no behavioral change
- [x] `cargo bench --bench jq_recurse_clone_bench -- --test` (smoke test, confirms the depth-limit bug fix)
- [ ] Formal A/B benchmark on an idle/dedicated machine (see above)

### DIFF
--- a/benches/jq_recurse_clone_bench.rs
+++ b/benches/jq_recurse_clone_bench.rs
@@ -26,9 +26,16 @@
 //! `del(..)` instead reaches `resolve_dynamic_indexes` and then applies
 //! `delete_at_path`/`delete_expr_paths_at` directly to the already-assembled
 //! static paths — no second walk of the original tree — so its cost tracks
-//! `push_recursive_branches`'s own share closely (measured: `del(..)`
-//! end-to-end 12.8ms vs `resolve_dynamic_indexes` 9.7ms at depth 400, both
-//! dominated by `push_recursive_branches`'s 8.2ms).
+//! `push_recursive_branches`'s own share closely.
+//!
+//! #701 found the depth-400 point in the list below panicking
+//! (`nesting depth exceeds limit of 384`, from `to_owned_at_depth`,
+//! unrelated to this file's own machinery — it fires before path-tracking
+//! even starts) — invisible until now because `cargo bench` isn't run by
+//! `cargo test`/CI. This file's depths are now derived from
+//! `MAX_VALUE_TREE_DEPTH` so they can't silently exceed it again; the old
+//! "measured... at depth 400" figures above were never actually
+//! reproducible as checked in and should not be trusted.
 //!
 //! `del(..)` also has no computed index anywhere in the query, so — like
 //! `path(..)` — it can never route through `eval_index_expr`/
@@ -57,7 +64,7 @@
 
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
 use std::hint::black_box;
-use succinctly::jq::{eval, parse, JqSemantics, OwnedValue, QueryResult};
+use succinctly::jq::{eval, parse, JqSemantics, OwnedValue, QueryResult, MAX_VALUE_TREE_DEPTH};
 use succinctly::json::JsonIndex;
 
 /// `{"k": {"k": ... {} ... }}`, `depth` levels of `"k"` nesting, no other
@@ -72,9 +79,11 @@ fn linear_nest(depth: usize) -> Vec<u8> {
 
 fn bench_recurse_clone_depth(c: &mut Criterion) {
     let mut group = c.benchmark_group("jq_recurse_clone_depth");
-    // Matches #661/#626's depths so results stay comparable across this
-    // benchmark family.
-    for &depth in &[100usize, 200, 300, 400] {
+    // Matches #661/#626's depths (100/200/300) where possible so results
+    // stay comparable across this benchmark family; the top depth is derived
+    // from `MAX_VALUE_TREE_DEPTH` (384) rather than hardcoded 400, which
+    // panics past the limit (#701 — see the module doc above).
+    for &depth in &[100usize, 200, 300, MAX_VALUE_TREE_DEPTH - 9] {
         let json = linear_nest(depth);
         let index = JsonIndex::build(&json);
         let expr = parse("del(..)").expect("filter parses");
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -16607,6 +16607,102 @@ fn slice_component_value(
     )
 }
 
+/// A persistent, structurally-shared path prefix (#701).
+///
+/// Replaces `PathBranch::path: Vec<Expr>`. Extending a prefix by one
+/// component (`extend`) is O(1) — a refcount bump on the shared parent plus
+/// one new allocation — instead of the O(depth) `prefix.to_vec()`/`.clone()`
+/// this type replaces. The only O(depth) operation is [`PathPrefix::to_vec`],
+/// which walks the chain to build a flat `Vec<Expr>`; callers pay it at most
+/// once per branch, only where a genuinely flat path is required (chiefly
+/// `resolve_dynamic_indexes`'s `assemble`), instead of it being repeated at
+/// every node visited during construction.
+///
+/// `Rc`, not `Arc`: this evaluator is single-threaded (no `Send`/`Sync`
+/// bounds or thread-spawning anywhere in `src/jq`), and `Rc` is already an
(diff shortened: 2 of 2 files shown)


newhoggy:
## A/B benchmark on a dedicated idle ARM box — confirms the win, and its limit

This PR shipped without a formal A/B (the description says why: the dev machine had 47+ concurrent sessions at the time). Ran one now on the project's dedicated ARM bench box (`johns-mac-mini`, Apple M4 Pro), idleness and AC power confirmed before starting (load average <1 throughout).

**Setup:** two worktrees — `before` = merge-base with `main` (`2467ca6c8`), `after` = this PR's tip (`e444de0c3`). Both built `--features cli,simd,regex,serde,bench-runner`. Used the existing `benches/jq_recurse_clone_bench.rs` `del(..)` benchmark unmodified, 3 rounds, each round interleaved (before then after) per depth before moving to the next depth, per `docs/guides/benchmarking.md#ab-benchmarking-method`.

| depth | before (avg of 3) | after (avg of 3) | speedup |
|-------|-------------------:|-------------------:|--------:|
| 100 | 328.5 µs | 241.6 µs | 1.36x |
| 200 | 1235.5 µs | 884.4 µs | 1.40x |
| 300 | 2744.7 µs | 1998.9 µs | 1.37x |
| 375 | *(`before` panics here — `MAX_VALUE_TREE_DEPTH` is 384; the pre-existing bug this PR also fixes)* | 3126.7 µs | — |

Round-to-round variance was under 1% at every depth — a consistent **~27-28% speedup**, not noise.

**But the scaling exponent didn't move.** Fitting time ∝ depthᵏ across 100→200→300: `before` k≈1.94, `after` k≈1.94 (300→375 alone on `after`: k≈2.0). `del(..)` end-to-end is still quadratic in depth, both before and after this fix.

That's expected, not a gap in this PR's mechanism — the `Rc<PathPrefix>` construction itself is verifiably O(1) per node at the code level. `delete_at_path`'s own per-resolved-path apply cost is a second, independent O(d²) term this PR doesn't touch, and on this benchmark it dominates the total, exactly as #701's "Why this is a separate issue" section anticipated and this PR's own description flags as deferred. Removing the path-construction term shrinks the constant on top of the remaining quadratic term; it can't change the exponent while that term is untouched.

Filing `delete_at_path`'s cost as its own follow-up issue, since #701 is now closed and nothing currently tracks it.

PR #1635, opened 2026-08-25
fix(yaml,json): decode the real UTF-8 char in validator error messages

### LINKED ISSUE #1422: yaml/validate.rs and json/validate.rs have the same Latin-1-cast mojibake bug #1187 fixed in yaml/parser.rs
## Summary

Found during #1187's review round. #1187 fixed `src/yaml/parser.rs`'s 8 sites
constructing `YamlError::UnexpectedCharacter` via a Latin-1 `byte as char` cast (mojibake
on multi-byte characters) by factoring them into one `err_unexpected_char` helper that
does a real UTF-8 decode. The identical bug pattern exists, unfixed, in two sibling
files:

## `src/yaml/validate.rs` (the separate strict YAML validator)

- Line ~1425, `scan_double_quoted_escape`: `InvalidEscape { sequence: other as char }` —
  `other` is the raw byte immediately following `\` inside a double-quoted scalar, which
  can be the lead byte of a multi-byte UTF-8 sequence (e.g. `"\日"` renders the escape
  character as mojibake instead of `日`).
- The `UnbalancedFlow`/bracket-related `as char` casts in the same file (~1094, ~1104,
  ~1498) are NOT affected — each only ever sees a byte already matched against a specific
  ASCII literal (`[`, `]`, `{`, `}`, etc.), so the cast is always safe there.

## `src/json/validate.rs` (JSON's own strict validator)

- `ValidationErrorKind::UnexpectedCharacter { found: char }` is built by hand at 5 sites
  (~224, ~270, ~280, ~310, ~363) via `c as char` / `self.peek().map_or('\0', |b| b as
  char)` — the same pattern. Validating malformed JSON with a multi-byte character at the
  error offset renders mojibake in the error message.

## Why not fixed as part of #1187

Different files, different error-type shapes (`YamlValidationErrorKind::UnexpectedCharacter`
has no `offset` field, taking `found: char` directly; JSON's `ValidationErrorKind` is a
third, independent enum) — #1187 was scoped to `src/yaml/parser.rs` specifically. A
shared `text::utf8::decode_char_at(bytes: &[u8], offset: usize) -> char`-style primitive
(decode-with-`'\0'`-EOF-fallback, falling back to the raw byte's own Latin-1 rendering for
a present-but-invalid byte — see `err_unexpected_char`'s own doc comment for the exact
three-way behavior #1187 settled on) would let all three modules (`yaml/parser.rs`,
`yaml/validate.rs`, `json/validate.rs`) share one implementation instead of three
independent (and, for two of them, still-buggy) ones.

## Scope note

Not investigated in depth. Each site is a small, mechanical fix once the shared primitive
exists — the main work is deciding where the primitive lives (`src/text/utf8/`, most
likely) and threading it through two more call sites' worth of tests.

newhoggy:
## Summary

`src/yaml/validate.rs`'s `scan_double_quoted_escape` and five sites in
`src/json/validate.rs` (`UnexpectedCharacter`/`InvalidEscape`) construct their
error message's character via a raw `byte as char` cast on the offending
byte. When that byte is the lead byte of a multi-byte UTF-8 sequence, the
error message renders mojibake instead of the real character (e.g. `"\日"`
reports `'æ'` instead of `'日'`).

#1187 fixed the identical pattern in `src/yaml/parser.rs` via a private
`err_unexpected_char` helper that does a real UTF-8 decode with a documented
three-way fallback (decodes correctly → real char; present-but-invalid byte →
Latin-1 cast of that byte; true EOF → `'\0'`). This PR generalizes that logic
into a new public primitive, `text::utf8::decode_char_at(bytes, offset) -> char`,
and threads it through both validators' error sites instead of duplicating the
logic a second and third time.

**A sixth site, not enumerated by the issue**: `json/validate.rs`'s own
`validate_escape` catch-all (`InvalidEscape { sequence: c as char }`) has the
identical bug — the byte after `\` in a bad escape can itself be a multi-byte
lead byte. Fixed alongside the five the issue named, since it's the exact same
bug in the exact same file.

**Left alone, per the issue's own scope note** (verified against current
`main`): `yaml/validate.rs`'s bracket-related `as char` casts (`scan_flow`'s
`UnclosedFlow`/`UnbalancedFlow`, `doc_marker_char`) only ever cast a byte
already matched against a specific ASCII literal (`[`, `]`, `{`, `}`, `-`,
`.`), so the cast is always safe there — confirmed by re-reading each call
site, not just trusting the issue text.

## Verification

Manually confirmed against the built binary, contrasted with an unmodified
`main` build:

```
$ printf 'a: "\\日"\n' | main-succinctly yaml validate
error: invalid escape sequence '\æ'          # mojibake, pre-fix
$ printf 'a: "\\日"\n' | fixed-succinctly yaml validate
error: invalid escape sequence '\日'          # real character, post-fix
```

Same contrast reproduced for `json validate` on a bad object key (`{日: 1}`).

## Test plan

- [x] New unit tests: `yaml::validate::tests::rejects_invalid_escape_decodes_multibyte_char_1422`,
      `json::validate::tests::test_unexpected_character_decodes_multibyte_char_1422`
      (covers all 6 fixed json sites + the 1 yaml site, plus a 4-byte emoji case)
- [x] New CLI integration tests in `tests/json_validate_tests.rs` and
      `tests/yaml_validate_tests.rs` asserting the real character appears in
      stderr and the mojibake byte does not
- [x] New doctest on `decode_char_at`
- [x] Full test suite (`cargo test --features cli,simd,regex,serde`): all green
- [x] Both required clippy legs (`--all-features` and the non-scalar-yaml
      feature set) clean

Fixes #1422

### DIFF
--- a/src/json/validate.rs
+++ b/src/json/validate.rs
@@ -219,9 +219,9 @@ impl<'a> Validator<'a> {
             Some(b'-' | b'0'..=b'9') => self.validate_number(),
             Some(b't' | b'f' | b'n') => self.validate_keyword(),
             Some(b'+') => Err(self.error(ValidationErrorKind::LeadingPlus)),
-            Some(c) => Err(self.error(ValidationErrorKind::UnexpectedCharacter {
+            Some(_) => Err(self.error(ValidationErrorKind::UnexpectedCharacter {
                 expected: "JSON value",
-                found: c as char,
+                found: crate::text::utf8::decode_char_at(self.input, self.offset),
             })),
             None => Err(self.error(ValidationErrorKind::UnexpectedEof {
                 expected: "JSON value",
@@ -267,7 +267,7 @@ impl<'a> Validator<'a> {
             if self.peek() != Some(b'"') {
                 return Err(self.error(ValidationErrorKind::UnexpectedCharacter {
                     expected: "string key",
-                    found: self.peek().map_or('\0', |b| b as char),
+                    found: crate::text::utf8::decode_char_at(self.input, self.offset),
                 }));
             }
             self.validate_string()?;
@@ -277,7 +277,7 @@ impl<'a> Validator<'a> {
             if self.peek() != Some(b':') {
                 return Err(self.error(ValidationErrorKind::UnexpectedCharacter {
                     expected: "':'",
-                    found: self.peek().map_or('\0', |b| b as char),
+                    found: crate::text::utf8::decode_char_at(self.input, self.offset),
                 }));
             }
             self.advance();
@@ -304,10 +304,10 @@ impl<'a> Validator<'a> {
                     self.advance();
                     return Ok(());
                 }
-                Some(c) => {
+                Some(_) => {
                     return Err(self.error(ValidationErrorKind::UnexpectedCharacter {
                         expected: "',' or '}'",
-                        found: c as char,
+                        found: crate::text::utf8::decode_char_at(self.input, self.offset),
                     }));
                 }
                 None => {
@@ -357,10 +357,10 @@ impl<'a> Validator<'a> {
                     self.advance();
                     return Ok(());
                 }
-                Some(c) => {
+                Some(_) => {
                     return Err(self.error(ValidationErrorKind::UnexpectedCharacter {
                         expected: "',' or ']'",
-                        found: c as char,
+                        found: crate::text::utf8::decode_char_at(self.input, self.offset),
                     }));
                 }
                 None => {
@@ -512,8 +512,8 @@ impl<'a> Validator<'a> {
 
                 Ok(())
             }
-            Some(c) => Err(self.error(ValidationErrorKind::InvalidEscape {
-                sequence: c as char,
+            Some(_) => Err(self.error(ValidationErrorKind::InvalidEscape {
+                sequence: crate::text::utf8::decode_char_at(self.input, self.offset),
             })),
             None => Err(self.error(ValidationErrorKind::UnclosedString)),
         }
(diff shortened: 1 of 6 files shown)


newhoggy:
Ran \`/code-review high\` (6 finder agents). Synthesis:

**Fixed** (4/6 agents independently converged on this): \`yaml/parser.rs\`'s \`err_unexpected_char\` — the very function \`decode_char_at\`'s doc comment says this PR "generalized" — still carried its own inline copy of the identical three-way decode-fallback logic instead of delegating. Refactored it to call the new shared primitive (commit 901cafd5a); rebuilt, full test suite + both clippy legs + fmt all clean, and manually re-verified the parser's own error path (\`a: [1, "x" 日]\` -> \`unexpected character '日'\`, not mojibake).

**Investigated, not a real bug — false premise**: two agents flagged \`src/jq/eval.rs\`'s embedded JSON parser (\`parse_json_value\`/\`parse_json_string_value\`, ~4 sites) as the same \`c as char\` mojibake pattern, reachable via \`fromjson\`/\`tonumber\`. Traced both call sites of \`parse_complete_json\`: both discard the specific \`Err(String)\` from these functions entirely and rebuild their message from the *original* \`&str\` input via \`EvalError::invalid_numeric_literal\`/\`cannot_parse_as_number\` — which, being a Rust \`&str\`, is already valid UTF-8 and always renders correctly. Confirmed live: \`printf 'a' | jq -R 'fromjson'\`-style multi-byte input reports \`(while parsing '日')\` correctly, not mojibake. The byte-cast pattern exists in that code but its output is never actually surfaced to a user — not filing an issue on a dead-code path.

**Fixed premise, real, separate bug — filed as #1636**: \`yaml/validate.rs\`'s \`scan_hex_escape\` hardcodes \`sequence: 'x'\` regardless of whether the failing escape was \`\x\`/\`\u\`/\`\U\` — confirmed live (\`\uZZZZ\` and \`\UZZZZZZZZ\` both misreport as \`\x\`). Different bug class (hardcoded literal, not an encoding cast) on the same enum variant this PR touches one line above, so flagging rather than scope-creeping it in here.

**Not actioned** (low-severity/style, judged not worth the diff): \`decode_char_at\`'s doc summary uses imperative mood instead of third-person-singular (STYLE-0006) — but matches two pre-existing functions in the same file, so not a fresh divergence. The double \`.get()\` in \`decode_char_at\` is slightly redundant but lifted verbatim from the pre-existing #1187 code it replaces, harmless on an error-only path.

Issue #378, opened 2026-07-26
jq: if/select collapse a multi-output condition, and the two evaluators disagree how


newhoggy:
**Severity: Medium** — reproduced against the built CLI (release, `2261571`).

jq fans a condition out: `if`/`select` run once per output of the condition stream. Succinctly collapses it, and the two evaluators collapse it *differently*.

```
echo 1 | sjq -c 'select((false,false) and true)'   # => 1        (jq: no output)
echo 1 | sjq -c 'select((true,false) and true)'    # => 1        (jq: 1)  -- right by luck
echo null | sjq -c 'if (true,false) then "a" else "b" end'
                                                    # => "a"      (jq: "a" "b")
```

Two distinct sites, both keyed on the same triplicated match:

- `builtin_select` and `eval_if` (`src/jq/eval.rs`) read `vs.first().is_some_and(...)` — first output wins, the rest are dropped.
- `eval_generic`'s `Builtin::Select` arm goes further and treats *any* `Many`/`ManyOwned` condition as truthy outright, so the CLI selects where the library does not. Pinned as a known divergence in `tests/jq_evaluator_parity_tests.rs::test_multi_output_condition_in_select_parity_160`; that `assert_divergence` must become `assert_parity` when this lands.

The `vs.first()` shape was triplicated across `eval_alternative`, `eval_if` and `builtin_select`. #160 removed the first copy; these two remain. Per the repo's own lesson ("duplicated predicates diverge silently — one definition, plus a test that the call sites agree"), the fix should route both through one helper rather than editing two copies.

**Fix:** dispatch per condition output. `push_truthiness` (added by #160, `src/jq/eval.rs`) already walks a stream into one `bool` per output and preserves `Error`/`Break`; `if`/`select` need the same walk with a body evaluation per bit instead of a push.

**Sibling of #354** (object construction) and #160 — same root shape, different operator.

---
_Split out of #160 during its fix._

PR #1644, opened 2026-08-25
fix(yq): select(cond) emits its input once, not once per truthy output

### LINKED ISSUE #1613: yq: select(cond) forks per truthy condition output, where real yq emits once if any is truthy
**Severity: Medium** — wrong output (duplicated documents/values), not a side-effect leak.
An ADR-0018 rule-4 fidelity break in `yq` mode: the *mode* decides the behaviour, and yq
mode is currently implementing jq's rule.

Found while A/B-testing #1481's binary-fanout changes in yq mode. **Not caused by #1481** —
`main` @ `d61ef1754` behaves identically.

## Repro

Verified live against Homebrew `yq` v4.53.3 (the pin) and a `--release --features cli`
build.

```
$ printf 'a: 1\nb: 2\n' > /tmp/s.yaml

$ yq            -o json -I0 '[.[] | select((true,true))]' /tmp/s.yaml
[1,2]
$ succinctly yq -o json -I0 '[.[] | select((true,true))]' /tmp/s.yaml
[1,1,2,2]
```

## The rule each tool implements

Real yq emits the value **once if any** condition output is truthy. It never duplicates,
and it does not stop at the first output either:

| condition            | real yq | succinctly yq | jq 1.7.1  |
|----------------------|---------|---------------|-----------|
| `(true, true)`       | `[1,2]` | `[1,1,2,2]`   | `[1,1,2,2]` |
| `(true, true, true)` | `[1,2]` | `[1,1,1,2,2,2]` | `[1,1,1,2,2,2]` |
| `(true, false, true)`| `[1,2]` | `[1,1,2,2]`   | `[1,1,2,2]` |
| `(true, false)`      | `[1,2]` | `[1,2]`       | `[1,2]`   |
| `(false, true)`      | `[1,2]` | `[1,2]`       | `[1,2]`   |
| `(false, false)`     | `[]`    | `[]`          | `[]`      |
| `(null, 1)`          | `[1,2]` | `[1,2]`       | `[1,2]`   |

jq's rule is "emit once per truthy output"; yq's is "emit once iff any output is truthy".
The one-truthy-bit rows agree by coincidence, which is why this has gone unnoticed — every
existing test happens to use a single-output condition or a single truthy bit.

Real conditions reach it just as easily as literals:

```
$ yq            -o json -I0 '[.[] | select((. == 1, . == 1))]' /tmp/s.yaml   # [1]
$ succinctly yq -o json -I0 '[.[] | select((. == 1, . == 1))]' /tmp/s.yaml   # [1,1]
```

## Cause

`eval_fanout` (`src/jq/eval.rs`) evaluates `body(bit)` once per truthy bit of the condition
stream and merges the results. Its own doc comment says it is "shared by `eval_if` and
`builtin_select` so a multi-output condition ... fans out the same way both call it,
instead of only consulting the condition's first output (#378)".

#378 was a **jq-side** fix, applied to shared `<S>` code without a yq-side check — the
recurring "a jq-scoped fix to a shared `<S>` builtin can silently regress yq" shape. The
per-truthy-bit fanout is right for jq and wrong for yq.

Note the `eval_if` half of that sharing is *not* affected: `if/then/else` is not valid yq
syntax at all (`yq '[.[] | if true then 1 else 2 end]'` => `Error: 1:8: lexer: invalid
input text "if (true,true) t..."`), so only `select` needs the per-mode split. Likewise
`select(empty)` cannot be spelled in yq (its lexer rejects `empty`), so the zero-output
case needs no yq answer.

## Suggested shape

Make the truthy-bit policy an `EvalSemantics` decision rather than a shared constant —
behavioural rules belong on `EvalSemantics`, per ADR-0018 — so `builtin_select` collapses
the condition stream to "any bit truthy" under `YqSemantics` and keeps the per-bit fanout
under `JqSemantics`. `eval_if` keeps the jq rule unconditionally, since yq cannot reach it.

## Acceptance criteria

- Every row of the table above matches its own reference tool in its own mode.
- `succinctly jq`'s behaviour is unchanged (#378's regression tests still pass).
- Oracle-verified coverage for the *non*-literal spelling too (`select((. == 1, . == 1))`),
  since a literal-only test would not have caught this.
- A yq golden case pins it, so the pinned-`yq` drift job re-checks it.

Refs #378, ADR-0018.

newhoggy:
## Summary

Real yq's `select(cond)` rule is "emit once iff any output of `cond` is truthy"; `succinctly yq` was implementing jq's own rule of "emit once *per* truthy output" instead — an ADR-0018 rule-4 fidelity break inherited from #378's shared `eval_fanout` fix (a jq-scoped fix applied to shared `<S>` code with no yq-side check, the recurring "shared builtin rewrite needs dual-mode check" shape).

```
$ yq            -o json -I0 '[.[] | select((true,true))]' s.yaml   # [1,2]
$ succinctly yq -o json -I0 '[.[] | select((true,true))]' s.yaml   # was [1,1,2,2], now [1,2]
```

- Gated behind a new `EvalSemantics::SELECT_EMITS_ONCE_IF_ANY_TRUTHY` const (`false` for jq, `true` for yq).
- `cond` is still walked to completion either way, so a later error/break after an already-truthy bit still escapes with the correct (now-capped) prefix — only the republish *count* collapses.
- `eval_if`/`if...end` is unaffected in every implementation below: yq's lexer rejects `if`/`then`/`else` outright, so there's no yq answer for that per-bit-fanout rule to diverge from.

**`select` turned out to have four independent implementations, not one** — found by checking every caller of the touched functions before handing this off to review, per this repo's own tier guidance:

1. `eval.rs`'s `builtin_select` (shares `eval_fanout` with `eval_if`) — the issue's original scope.
2. `eval.rs`'s path-context `Expr::Builtin(Builtin::Select(cond))` arm (also `eval_fanout`).
3. `eval_generic.rs`'s native `Builtin::Select` arm (its own `truthy_count` computation, no `eval_fanout` at all).
4. `resolve_node`'s path-tracking `Select` arm (`resolve_cond_fork`, reached by `|=`/`=`/`del()`, not `eval_fanout` either) — **found live while fixing the first three, not in #1613's original scope**. Unpatched, a multi-truthy condition inside `|=` applied the update once per truthy bit instead of once:
   ```
   $ yq            -o json -I0 '(.a[] | select((true,true))) |= . + 10' s.yaml   # {"a":[11,12,13]}
   $ succinctly yq -o json -I0 '(.a[] | select((true,true))) |= . + 10' s.yaml   # was {"a":[21,22,23]}, now {"a":[11,12,13]}
   ```
   Fixed together with the first three since it's the same divergence reached through a different code path — `resolve_cond_fork`'s own doc comment already flagged this exact cross-check ("if a future fix to the fork-on-cond policy needs applying here, check whether it also needs applying there").

jq mode's existing per-truthy-bit fanout is unchanged in all four sites and pinned by regression tests (#378).

Fixes #1613.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 956 tests; two isolated SIGKILL flakes from machine contention during concurrent runs, each reproduced green in isolation and in a clean single-threaded full run)
- [x] Diffed output against the pinned oracles directly (Homebrew `yq` v4.53.3 and `/usr/bin/jq` 1.7.1), not just golden fixtures — every row of the issue's acceptance table, the non-literal spelling (`select((. == 1, . == 1))`), the error-after-truthy-prefix case, and the `|=` assignment case
- [x] New unit tests in both evaluators (`eval.rs`, `eval_generic.rs`) covering all 4 call sites in yq mode, plus regression pins confirming jq mode is unchanged
- [x] `cargo llvm-cov` + `omni-dev coverage diff`: 100% patch coverage (70/70 new lines) on both touched files

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -3611,6 +3611,33 @@ fn eval_fanout<'a, W: Clone + AsRef<[u64]>>(
     }
 }
 
+/// Whether `select(cond)` should republish its input for the current truthy
+/// bit, under `S::SELECT_EMITS_ONCE_IF_ANY_TRUTHY` (yq, #1613) vs jq's
+/// unconditional per-truthy-bit fanout (#378): `false` under jq (every
+/// truthy bit republishes, `already_emitted` never gates anything); `false`
+/// past the *first* truthy bit under yq (`already_emitted` caps the
+/// republish count at one). `already_emitted` is the caller's own running
+/// flag for one `select` evaluation, mutated in place on a true answer.
+///
+/// One definition shared by `builtin_select`, `eval_pipe_with_path_context_
+/// internal`'s `Select` arm, and `resolve_node`'s `Select` arm -- each of
+/// which independently hand-wrote this same three-line guard before, the
+/// "duplicated predicates diverge silently" shape CLAUDE.md's #106 lesson
+/// warns about (echoed by this file's own `ArgFanout`/`apply_arg_fanout`,
+/// #1537, the precedent for pulling a repeated gate into one place instead).
+/// `eval_generic.rs`'s native `Select` arm computes the analogous
+/// `truthy_count` differently, as a one-shot reduction over an
+/// already-fully-collected `Vec<bool>` rather than a per-bit gate, so it
+/// isn't a caller of this helper -- see its own comment for why that shape
+/// doesn't unify cleanly with this one.
+fn select_emits<S: EvalSemantics>(truthy: bool, already_emitted: &mut bool) -> bool {
+    if !truthy || (S::SELECT_EMITS_ONCE_IF_ANY_TRUTHY && *already_emitted) {
+        return false;
+    }
+    *already_emitted = true;
+    true
+}
+
 /// Evaluate `left op right` over every pairing jq's cartesian fanout
 /// produces, right operand outer / left operand inner (#768): `(1,2) + (10,20)`
 /// is `11,12,21,22` (right=10 held fixed while left runs 1,2; then right=20).
@@ -5999,10 +6026,8 @@ fn builtin_upper_in_src<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
 /// (`select((false,false) and true)` yields nothing, matching jq — not the
 /// first-output-only answer `vs.first()` used to give) — restored by
 /// `eval_fanout` (#378). Under `S::SELECT_EMITS_ONCE_IF_ANY_TRUTHY` (yq,
-/// #1613), that per-bit republish collapses to at most one: `already_emitted`
-/// suppresses every truthy bit after the first, while still walking every
-/// bit of `cond` (so a later error/break still escapes `eval_fanout`
-/// normally) and still emitting nothing for a purely-falsy condition.
+/// #1613), that per-bit republish collapses to at most one via
+/// `select_emits` — see its own doc comment.
 fn builtin_select<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     cond: &Expr,
     value: StandardJson<'a, W>,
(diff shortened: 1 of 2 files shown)


newhoggy:
Filed #1645 for a genuine efficiency finding from the review (yq mode's truthy check still pays a full `to_owned_cursor` deep-copy per `cond` output instead of an O(1) `is_falsy()`) — pre-existing in most of its cost, needs hardware A/B benchmarking per this repo's discipline, so it's a follow-up rather than bundled into this correctness fix.

Issue #1534, opened 2026-08-23
yq: ArgFanout::FirstOnly still leaks a trailing error/break from the argument outputs it's supposed to discard


newhoggy:
**Severity: Medium** (yq mode only — a query that should silently succeed instead hard-errors)

## Summary

Found during #1279's own code review (PR pending). `ArgFanout::FirstOnly` (`src/jq/eval.rs:1447-1450`) is documented as "Only the argument's first output is used; later ones are discarded" and gates `has`/`split`/`join` in yq mode via `ArgFanout::yq_native::<S>()`. `fanout_arg` (`src/jq/eval.rs:1507-1531`) implements it by truncating the *value* list to one element (`args.truncate(1)`), but the separately-tracked `trailing` control — captured by `stream_outputs` from evaluating the rest of the (now-discarded) argument stream — is untouched and still propagates via `partial(out, control)`. So a `break`/`error`/`halt` occurring anywhere after the first output still escapes, even though nothing past the first output is supposed to be "used" at all.

Real jq's own semantically closest construct, `first(g)`, shows what "later ones are discarded" should mean: it truncates the underlying generator *eagerly*, so an `error(...)` after the first output is never even reached:
```
$ echo '{"a":1}' | jq -c 'first(has(("a","b", error("boom"))))'
true
```
(exits 0 — the error is never reached, since `first` stops pulling from the generator after its first output).

## Confirmed repro

```
$ printf 'a: 1\n' | succinctly yq -o json 'has(("a","b", error("boom")))'
Error: boom
true
```
(exits 1). succinctly's own `ArgFanout::FirstOnly` doc comment says "later ones are discarded," but the trailing error from the discarded `error("boom")` branch still fires. The same shape reproduces for `split`/`join` in yq mode, e.g.:
```
$ printf 'x: "a,b"\n' | succinctly yq -o json '.x | split((",", error("boom")))'
```
also emits the split result *and* propagates "boom" instead of silently succeeding with just the first-output result.

Note: the same collapsing-but-still-escaping shape already existed in the pre-#1279 `result_to_owned_full`'s `Partial` arm for *all* modes before this PR, so the underlying quirk isn't newly introduced. What #1279 changes is that it's the first place this exact scenario becomes reachable through `has`/`split`/`join` in yq mode specifically (those builtins previously used a mode-uniform single-value collapse), and it's the first place `ArgFanout::FirstOnly` is named and documented with a "later ones are discarded" contract the implementation doesn't fully honor.

[...]

PR #1648, opened 2026-08-25
fix(yq): contains(cond) aborts on an escaped argument instead of emitting first

### LINKED ISSUE #1553: yq: contains emits a value then raises when its argument escapes, where real yq emits nothing
**Severity: Low** (yq mode; emits output real yq does not, alongside the right error)

## Summary

Sibling of #1534, which fixed exactly this shape for `has`/`split`/`join`. Those three carry `ArgFanout::yq_native`, so #1534's rule ("an escape anywhere in the argument takes the whole call down, with no output") reaches them. `contains` deliberately does **not** carry that gate -- it is the one shared builtin that genuinely *does* fan out in real yq, so gating it would introduce a divergence rather than preserve one (see `ArgFanout::yq_native`'s own doc comment).

The consequence is that `contains` keeps the pre-#1534 behaviour: it emits a value and *then* raises.

## Confirmed repro (verified against pinned yq v4.53.3)

```
$ printf 'x: "abc"\n' | yq -o=json '.x | contains(("a", error("boom")))'
Error: boom                     # stderr only; stdout empty, exit 1

$ printf 'x: "abc"\n' | succinctly yq -o json '.x | contains(("a", error("boom")))'
true                            # stdout -- real yq prints nothing here
Error: boom                     # stderr, exit 1
```

## Pre-existing, not introduced by #1534

Verified identical on the pre-#1534 binary -- #1534 simply did not reach this builtin. Filed rather than folded into #1534 because the fix is genuinely different: `contains` must keep fanning out (jq and yq agree that it does), so it cannot just adopt the `FirstOnly` gate. What it needs is the *escape* half of #1534's rule without the *truncation* half.

## Suggested fix direction

Real yq fans `contains` out **and** aborts the whole call on an escape anywhere in the argument. That is neither of the two existing yq gates:

| | fan out? | escape clears output? |
|---|---|---|
| `ArgFanout::All` (jq) | yes | no -- prefix then control (#1279 rule 1) |
| `ArgFanout::yq_native`/`FirstOnly` | no | yes (#1534) |
| what yq's `contains` wants | **yes** | **yes** |

So this likely wants a third gate (fan out, but clear the values when the argument escaped), applied to `contains` in yq mode only. Confirm the full behaviour against v4.53.3 first -- including the multi-output non-error case (`[.x | contains(("a","zz"))]` is `[true,false]`, already matching) -- rather than assuming the table above is complete.

## Related

- #1534 -- the same "emit a value it then raises over" shape, fixed for the three `yq_native`-gated builtins.
- #1533 -- the two-argument half of that family, still open for its own reasons.

newhoggy:
## Summary

Real yq's `contains` genuinely fans out over a multi-output argument, unlike its
`ArgFanout::yq_native`-gated neighbours (`has`/`split`/`join`, #1534) — but it still needs
their yq-only rule: an escape anywhere in the argument clears the whole output, rather
than jq's own "emit the prefix, then raise" (#1279 rule 2).

```
$ yq            '.x | contains(("a", error("boom")))' s.yaml   # Error: boom (stdout empty, exit 1)
$ succinctly yq '.x | contains(("a", error("boom")))' s.yaml   # was: true (stdout), then Error: boom
```

Neither existing yq gate fit:
- `FirstOnly` would wrongly truncate the ordinary non-error multi-output case — `contains`
  must still answer `[true,false]` for `contains(("a","zz"))`, which real yq does too.
- `RejectMany` refuses multi-output outright, which `contains` never does.

## Fix

Added a third `ArgFanout` variant, `AllClearedOnEscape`: keeps every value like `All`, but
routes through the same eager, escape-clearing path `FirstOnly`/`RejectMany` already use in
`fanout_arg` — no change needed to `fanout_arg` itself, since its dispatch condition
(`!matches!(fanout, ArgFanout::All)`) already admits any non-`All` variant into that path.

`contains_gate::<S>()` picks `AllClearedOnEscape` for `YqSemantics`, `All` (unchanged) for
`JqSemantics` — pinned by a jq-mode regression test alongside the new yq-mode coverage.

Fixes #1553.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 959 tests, all green)
- [x] Diffed output against the pinned oracles directly (Homebrew `yq` v4.53.3, `/usr/bin/jq` 1.7.1) — escape-in-argument, immediate escape, and no-escape multi-output cases in both modes
- [x] New unit tests (`apply_arg_fanout`/`clear_values_when_yq_argument_escaped`/gate-selection coverage) and new CLI tests (`tests/yq_cli_tests.rs`) covering all four cases
- [x] `cargo llvm-cov` + `omni-dev coverage diff`: 100% patch coverage (33/33 new lines)

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -379,16 +379,20 @@ per-mode-wholesale — all rows live-verified against yq v4.53.3:
 | `flatten(n)` | literal depth only — `bad expression, please check expression syntax` | gated off |
 | `getpath`, `range`, `nth`, `limit`, `combinations`, `paths`, `ltrimstr`, `rtrimstr`, `startswith`, `endswith`, `index`, `rindex`, `indices`, `inside`, `splits`, `scan`, `gsub`, `strftime`, `strptime`, `bsearch`, `pow` | **lexer-rejected — the builtin does not exist** | fans out; unopposed extension |
 
-There are **two** gate predicates, so the audit is both greps —
-`grep -E 'ArgFanout::(yq_native|reject_many_in_yq)' src/jq/eval.rs`. Each is named rather than
-inlined precisely so that stays a grep (CLAUDE.md's #106 lesson).
+There are **three** gate predicates, so the audit is all three greps —
+`grep -E 'ArgFanout::(yq_native|reject_many_in_yq|contains_gate)' src/jq/eval.rs`. Each is named
+rather than inlined precisely so that stays a grep (CLAUDE.md's #106 lesson). `contains_gate`
+is `contains`'s own gate (#1553): unlike the other two, it does not change whether the argument
+fans out (`contains` genuinely fans out in both modes, the table row above) — it only adds the
+no-prefix escape rule below.
 
 ### The prefix rule is where the two modes deliberately part
 
 jq emits the outputs a prefix earned and *then* fires the argument's trailing control (#1277's
 rule 2). Real yq does not: an escape anywhere in a gated argument produces the escape alone,
 with nothing printed first — live-verified against v4.53.3 for `has`, `split` and `join`
-(#1534), and for `delpaths` (#1533):
+(#1534), for `delpaths` (#1533), and for `contains` (#1553, `ArgFanout::AllClearedOnEscape` —
+the one gate that keeps every value on the *non*-escaping path, unlike the other two):
 
 ```bash
 $ printf 'a: 1\n' | yq            'has(("a","b", error("boom")))'   # Error: boom, stdout empty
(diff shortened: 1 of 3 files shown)


newhoggy:
Filed #1649 (contains/inside raise on a scalar-vs-scalar kind mismatch, where real yq answers `false` — pre-existing, sits in the exact function this PR touches) and #1650 (~13 jq-only builtins including `inside` are silently reachable without `--jq-extensions`, unlike their gated siblings — a policy/consistency gap, not a mechanical fix) as follow-ups from the review. Both are pre-existing and out of scope for this PR's escape-clearing fix.

Issue #791, opened 2026-08-11
jq: implement halt, halt_error, stderr, and $__prog_name__ (severable easy half of #723)


newhoggy:
## Summary

Split off from #723's severable easy half. jq's `halt`, `halt_error`/`halt_error(exit_code)`, `stderr`, and the `$__prog_name__` variable are not implemented. Confirmed live:

```
$ echo '1' | succinctly jq 'halt'
undefined function: halt
$ echo '1' | succinctly jq 'halt_error'
undefined function: halt_error
$ echo '1' | succinctly jq 'stderr'
undefined function: stderr
$ echo '1' | succinctly jq '$__prog_name__'
undefined variable: $__prog_name__
```

## Why this is the cheap half

Unlike `input`/`inputs`/`input_line_number` (left in #723), none of these four need the CLI's remaining-input-stream state threaded into the evaluator:

- `halt` — exits immediately with code 0, no output.
- `halt_error`/`halt_error(exit_code)` — prints the current value to stderr (raw if string, JSON otherwise) and exits (default code 5).
- `stderr` — prints the current value to stderr (compact JSON) and passes it through unchanged; a pure side effect.
- `$__prog_name__` — a static constant (the program name), no input dependency at all.

## De-risking: reuse existing machinery, don't build new

- `src/jq/eval.rs` already has a `Control` enum (`Control::Error`, `Control::Break`) threaded through the whole evaluator for control-flow signals. A `Control::Halt(exit_code)` variant is a third case of a pattern already proven, not new architecture.
- `src/bin/succinctly/main.rs` already has multiple `std::process::exit(exit_code)` call sites — the CLI-level exit-code plumbing this needs already exists.
- `output.rs` already has exit-code conventions documented for the uncaught-error/bare-`halt_error` case (see its doc comment on jq's exit code 5), so the intended exit code for `halt_error` is already established in this codebase, just not wired to a builtin.

## Scope

No design question outstanding — implementation-ready. Confirm `halt_error`'s stderr formatting (raw string vs `tostring`-equivalent JSON) and default exit code (5) against real jq before landing.

## Related

Split from #723 (input/inputs/input_line_number retain the harder design question there).

newhoggy:
While addressing the review findings on the halt/halt_error/stderr implementation, I checked the remaining `$__prog_name__` item in this issue's title against real jq and found it doesn't exist there.

Verified against both the pinned jq 1.7.1 oracle and jq 1.8.2:

```
[...]

PR #1652, opened 2026-08-26
fix(jq): debug/debug(msg) write ["DEBUG:",value] to stderr

### LINKED ISSUE #1594: jq: debug/debug(msg) never write ["DEBUG:",value] to stderr
**Severity: Low** — cosmetic (stdout/exit code/value-passthrough are all
correct; only the diagnostic stderr trace is missing).

Found by #1481's own oracle sweep (`scripts/jq-fanout-oracle-sweep.sh`),
which cross-products binary-fanout shapes against several terminating side
effects including `debug` — every one of the 49 divergences that sweep
reported was this same gap, unrelated to fanout ordering itself (stdout and
exit code matched jq in all 49; only stderr differed). Confirmed in
isolation, with no fanout involved at all:

```
$ jq            -cn '1 | debug'
["DEBUG:",1]
1
$ succinctly jq -cn '1 | debug'
1
```

## Root cause

`builtin_debug`/`builtin_debug_msg` (`src/jq/eval.rs`) are explicit,
commented no-ops:

```rust
fn builtin_debug<W: Clone + AsRef<[u64]>>(
    value: StandardJson<'_, W>,
    _optional: bool,
) -> QueryResult<'_, W> {
    // In a library context we don't actually print to stderr
    // We just pass through the value unchanged
    QueryResult::Owned(to_owned(&value))
}
```

Unlike `debug`, `stderr`/`halt_error` do write directly from `eval.rs` via
`write_stderr(...)` (`builtin_stderr`, `eval.rs`) — there is no separate
CLI-layer wiring for either builtin family in `src/bin/succinctly/`. So
`debug`'s "library context" no-op policy is never overridden for the actual
CLI binary, where jq's own `["DEBUG:",value]` (or `["DEBUG:",msg,value]` for
`debug(msg)`) trace is expected.

## Suggested shape

Same mechanism `stderr`/`halt_error` already use: have `builtin_debug`/
`builtin_debug_msg` call `write_stderr` with jq's own JSON-array format
(`["DEBUG:",value]`, or `["DEBUG:",msg,value]` for the `debug(msg)` two-arg
form — verify the exact real-jq wording/formatting live against the pinned
binary before implementing, per this repo's reference-fidelity rule).

## Acceptance criteria

- `1 | debug` writes `["DEBUG:",1]` to stderr, matching jq.
- `debug("msg")` (two-arg form) writes jq's exact `["DEBUG:",...]` shape.
- `docs/compliance/jq/limitations.md` gets an entry if any residual gap
  remains, per ADR-0018.

newhoggy:
## Summary

\`builtin_debug\`/\`builtin_debug_msg\` were explicit, commented no-ops — \`.\` passed through
unchanged, nothing ever reached stderr, so \`debug\`'s whole diagnostic purpose was silently
missing:

\`\`\`
$ jq            -cn '1 | debug'   # writes ["DEBUG:",1] to stderr, 1 to stdout
$ succinctly jq -cn '1 | debug'   # was: 1 to stdout, nothing to stderr
\`\`\`

## Fix

Real jq writes a compact-JSON \`["DEBUG:", value]\` line with a trailing newline;
\`debug(msg)\` writes one such line **per output of \`msg\`** (its own definition is
\`(msg|debug|empty), .\` — confirmed live: \`debug(("a","b"))\` writes two lines), and the
line contains \`msg\`'s own evaluated value, not \`.\`.

Reuses \`write_stderr\` and \`owned_value_to_json::<S>\` — the same primitives
\`stderr\`/\`halt_error\` already use — so non-finite floats etc. render identically across
all three rather than a fourth hand-rolled formatter.

\`debug(msg)\`'s pre-existing halt/error propagation (#791: control effects reach the
process even though \`msg\`'s text used to stay silent) is unchanged by this fix — only the
no-print policy itself is gone. Two existing tests referencing that old policy in their
comments (\`test_debug_msg_argument_halt_propagates\`, still passing unmodified) had their
comments updated to stay accurate.

Reaches yq mode identically (\`debug\` is gated behind \`--jq-extensions\` there — real yq has
no \`debug\` at all), pinned alongside the jq-mode coverage.

Fixes #1594.

## Test plan

- [x] \`cargo test --features cli,simd,regex,serde\` (full suite, 965 tests, all green; one
      isolated SIGKILL flake from machine contention, reproduced green in isolation)
- [x] \`cargo clippy --all-targets --all-features -- -D warnings\`
- [x] \`cargo clippy --all-targets --features std,simd,serde,cli,regex,bench-runner,large-tests,mmap-tests -- -D warnings\`
- [x] \`cargo build --no-default-features\` (no_std) — same 6 pre-existing dead-code warnings as unmodified \`main\`, none new
- [x] Diffed output byte-for-byte against the pinned jq 1.7.1 oracle (via separate stdout/stderr file redirection, not a combined capture) — bare \`debug\`, \`debug(msg)\` with a scalar/string msg, multi-output \`msg\`, and \`debug(error(...))\`/\`debug(halt_error(n))\` control-flow propagation
- [x] New unit + CLI tests (jq mode and yq mode) pinning the exact stderr byte content
- [x] \`cargo llvm-cov\` + \`omni-dev coverage diff\`: 100% patch coverage (15/15 new lines)

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -31414,10 +31414,19 @@ fn builtin_isfinite<W: Clone + AsRef<[u64]>>(
 /// formatter `builtin_stderr` uses, so non-finite floats etc. render
 /// identically across both rather than re-deriving the jq/yq JSON-shape fork
 /// a second time.
+///
+/// Builds the whole line before writing it, in one `write_stderr` call --
+/// matching `builtin_halt_error`'s own single-write convention just below,
+/// rather than the three separate writes an earlier revision used. Also
+/// makes this atomic against a mid-formatting panic (e.g.
+/// `assert_value_tree_depth`'s nesting-depth guard): a panic during
+/// `owned_value_to_json` now leaves stderr untouched rather than a
+/// truncated `["DEBUG:",` fragment with no closing bracket.
 fn write_debug_line<S: EvalSemantics>(value: &OwnedValue) {
-    write_stderr("[\"DEBUG:\",");
-    write_stderr(&owned_value_to_json::<S>(value));
-    write_stderr("]\n");
+    write_stderr(&format!(
+        "[\"DEBUG:\",{}]\n",
+        owned_value_to_json::<S>(value)
+    ));
 }
 
 /// Builtin: debug - print `["DEBUG:", .]` to stderr, pass `.` through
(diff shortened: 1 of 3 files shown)


newhoggy:
Filed #1653 for a genuine, pre-existing gap the review surfaced (`--unbuffered` doesn't achieve real-time stdout/stderr interleaving — the CLI batches a whole input's outputs before writing any, unrelated to this PR's own stderr-write correctness fix). Not introduced or worsened here; `debug`/`debug(msg)` just newly demonstrate a gap `stderr`/`halt_error` already had.

PR #1673, opened 2026-08-26
fix(jq): flush stdout explicitly before --validate's lazy-path early return

### LINKED ISSUE #1563: jq: --validate's lazy-path failure return skips the explicit out.flush()
**Severity: Low** — no observed failure; the only test currently exercising this path
depends on incidental `Drop` behavior instead of an explicit contract.

## Summary

Found during #1558's code review (test-coverage pass). `run_jq`'s lazy path
(`src/bin/succinctly/jq_runner.rs`) has an early return on a `--validate` failure that
skips the function's own explicit-flush convention every other early return follows:

```rust
for (idx, raw) in raw_inputs.iter().enumerate() {
    ...
    if args.validate {
        if let Err(exit_code) = validate_json_input(raw, filename.as_deref()) {
            return Ok(exit_code);   // <-- no out.flush() here
        }
    }
    ...
}
```

Every other early return in this function calls `out.flush()?` first (lines 1017, 1024,
1142, 1255, 1284, 1302, plus the function's own tail at 1309) — `out` is a
`BufWriter::new(stdout.lock())`. This one doesn't.

## Why this hasn't caused a visible bug (yet)

`BufWriter::drop` performs a best-effort flush when `out` goes out of scope, so any
buffered output from files processed *before* the one that failed validation currently
still reaches stdout in practice — confirmed by
`test_validate_multi_file_second_invalid_keeps_first_files_output_1558` (#1558), which
asserts exactly this and passes. But that test's correctness currently rests on
`Drop`'s flush, not an explicit one:

- `BufWriter::drop` silently **swallows** any flush error (e.g. a broken pipe if stdout
  closes early) rather than propagating it, unlike every sibling return path's explicit
  `.flush()?`.
- If a future refactor of this function changes `out`'s ownership/scoping (e.g. moves it
  into a struct with a different `Drop` order, or the loop gets extracted into a helper
  that doesn't own `out`), the implicit flush could silently stop happening with no
  compiler error and no obviously-related test failure to point at the cause.

## Suggested fix

Add `out.flush()?;` immediately before the `return Ok(exit_code)` in the `--validate`
failure arm, matching the convention every other early return in this function already
follows. One line; no behavior change expected in practice, just closes the gap between
"works because of `Drop`" and "works because it's supposed to."

## Related

- #1558 — where this was found (test-coverage pass for `--validate`), not blocking that
  PR since it's tests-only and the current behavior isn't observably wrong.

newhoggy:
## Summary

`run_jq`'s lazy path (`src/bin/succinctly/jq_runner.rs`) had an early return
that skipped the function's own explicit-flush convention every other early
return follows:

```rust
if args.validate {
    if let Err(exit_code) = validate_json_input(raw, filename.as_deref()) {
        return Ok(exit_code);   // <-- no out.flush() here
    }
}
```

`out`/`writer` in both `run_jq` and `run_yq` are `BufWriter::new(stdout.lock())`.
`BufWriter::drop` performs a best-effort flush when they go out of scope, so
already-buffered output from files processed before a failing one still
reaches stdout in practice today — but that correctness rested on `Drop`,
which silently **swallows** any flush error (e.g. a broken pipe) instead of
propagating it, unlike an explicit `.flush()?`.

## What grew during review

Code review on the initial one-line fix found the identical bug in
`yq_runner.rs`'s own multi-file `'m2_files` streaming loop (three sites:
`read_file`, `yaml_validate_guard`, `YamlIndex::build`), plus a real bug in
the shared `flush_then_err` helper this PR introduced: `err.context(flush_msg)`
made the *flush* failure the primary displayed message instead of the
original error, the opposite of the helper's own documented intent —
verified against `anyhow`'s actual `Context`/`Display` source. All three
fixed in follow-up commits, each re-verified with a full local test run.

A deeper multi-angle review pass then confirmed the fix is correct as far as
it goes, but incomplete relative to the *general* pattern: several more
structurally identical unguarded-write sites exist in both files (e.g.
`yq_runner.rs`'s `stream_cursor!` macro and DOM "collect" path,
`jq_runner.rs`'s DSV loop and materializing-branch loops). Chasing all of
those here would have turned a narrow bug fix into a much larger refactor of
two ~5,000-line functions with no clear stopping point, so the remaining
sites and the review's suggested architectural direction (a flush-on-Drop
safety net at writer construction, instead of per-site annotation) are
tracked in #1680 instead of folded into this PR.

## Fix

- `out.flush()?`/`writer.flush()?` before the early returns named above.
- `output.rs::flush_then_err`, shared by both runners: flushes before
  propagating a real error, and — if the flush itself also fails — keeps the
  original error as the primary message rather than losing it.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test --features cli,simd,regex,serde` — 55/55 test binaries pass, 0 failures (re-run after every round of changes, including post-rebase)
- [x] New test: `test_validate_multi_file_second_invalid_keeps_first_files_output_1563` (`tests/yq_cli_tests.rs`), mirroring the existing jq-side sibling test for the same bug class
- [x] All existing `--validate`-related tests pass unchanged
- [x] Live-verified both the `--validate` and plain-YAML-parse-error multi-file cases via the built CLI

Fixes #1563

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -25,8 +25,8 @@ use succinctly::json::JsonIndex;
 
 use super::JqCommand;
 use crate::output::{
-    self, escape_json_string, escape_json_string_ascii, exit_codes, ColorScheme, ControlEscape,
-    DiagStyle, ErrorSink, FloatStyle, InputLocation, JsonFormatOpts,
+    self, escape_json_string, escape_json_string_ascii, exit_codes, flush_then_err, ColorScheme,
+    ControlEscape, DiagStyle, ErrorSink, FloatStyle, InputLocation, JsonFormatOpts,
 };
 
 /// Evaluation context for passing variables to the jq evaluator.
@@ -1254,11 +1254,11 @@ pub fn run_jq(args: JqCommand) -> Result<i32> {
                             // in this same run, and a genuine (non-malformed-
                             // document) error here shouldn't leave that
                             // relying on `Drop`'s own best-effort,
-                            // error-swallowing flush.
-                            None => {
-                                out.flush()?;
-                                return Err(e);
-                            }
+                            // error-swallowing flush. `flush_then_err`
+                            // (review of #1673) keeps `e` as the reported
+                            // error even if the flush also fails, instead of
+                            // the flush error silently displacing it.
+                            None => return flush_then_err(&mut out, e),
                         }
                     }
                     // result is dropped here, freeing its memory immediately
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -1161,6 +1161,14 @@ pub fn run_jq(args: JqCommand) -> Result<i32> {
             // Validate JSON if --validate flag is set
             if args.validate {
                 if let Err(exit_code) = validate_json_input(raw, filename.as_deref()) {
+                    // Every other early return in this function flushes
+                    // explicitly before returning (#1563) -- this one used
+                    // to rely on `out`'s own `Drop` impl to flush any
+                    // already-buffered output from files processed before
+                    // this one, which works today but silently swallows a
+                    // flush error (e.g. a closed stdout) instead of
+                    // propagating it, unlike every sibling return path.
+                    out.flush()?;
                     return Ok(exit_code);
                 }
             }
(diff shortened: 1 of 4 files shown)


newhoggy:
Closing and reopening to force a fresh CI trigger -- the runs queued before today's GitHub Actions outage (15:11-16:14 UTC) appear permanently stuck even though capacity has otherwise recovered (confirmed via a fresh PR on the same repo passing cleanly). No code changes.

newhoggy:
Closing in favor of a fresh PR from the same branch -- this PR's events (rerun, close/reopen, a new empty commit) all failed to trigger any CI run even after the broader Actions outage recovered (confirmed: a brand-new PR on this same repo, #1681, triggers and runs cleanly). No further diagnostics available without repo admin access.

Issue #1693, opened 2026-08-27
yq: --ascii-output does nothing in the compact JSON M2 fast path


newhoggy:
**Severity: Low** (cosmetic escaping divergence; no data loss, no crash)

## Summary

`succinctly yq -o=json -I=0 --ascii-output '.'` does not escape non-ASCII characters
to `\uXXXX` — it passes them through raw, same as without `--ascii-output` at all.
Found while implementing #1577 (extending `--slurp`'s M2 fast path to JSON output),
whose new code inherits this exact same gap by reusing the established
`can_json_fast_path`-style gate.

## Repro (verified on `main`, unrelated to #1577's own change)

```console
$ printf 'k: "héllo"\n' > g.yaml
$ succinctly yq -o=json -I=0 --ascii-output '.' g.yaml
{"k":"héllo"}                    # should be {"k":"héllo"}

$ succinctly yq -o=json --ascii-output '.' g.yaml    # pretty mode (no -I=0)
{
  "k": "héllo"               # correct here -- pretty mode uses the DOM path
}
```

`--ascii-output` is a succinctly extension (real yq v4.53.3 has no such flag — only
jq does), so there is no oracle to diverge from; this is a self-consistency bug.

## Root cause

`can_json_fast_path`'s own condition is `output_config.compact || (can_stream_pretty_or_colored
&& !args.ascii_output)` (`src/bin/succinctly/yq_runner.rs`) — the `output_config.compact ||`
half unconditionally allows the M2 streaming fast path even when `--ascii-output` is set,
on the apparent assumption that the compact path already implements the escaping. It
doesn't: `stream_json_value`/`stream_json_sequence` (`src/yaml/light.rs`) have no
`ascii_output`/`escape_non_ascii` parameter at all, unlike the DOM path's `output_value`,
which does (and is correctly reached — and correct — whenever pretty mode forces the
gate closed instead).

Every M2 JSON fast-path gate in `yq_runner.rs` shares this same `compact || (... &&
!ascii_output)` shape (`can_json_fast_path`, `can_inplace_json_fast_path`, and now
`can_slurp_fast_path` after #1577), so the gap is universal across all of them, not
specific to any one flag combination.

## Suggested fix direction

Not attempted; needs its own investigation. Two directions:

1. Thread an `ascii_output: bool` parameter through `stream_json_value`/
   `stream_json_sequence`'s string-writing path (mirroring the DOM path's existing
[...]

Issue #757, opened 2026-08-10
jq: CLI streaming-output integration for lazy map/select (Slice 3 of #700)


newhoggy:
## Summary

Slice 3 of #700's design (`docs/plan/jq-lazy-map-select.md`), deferred out of both #724 and #725: extend the CLI's own output-streaming eligibility to the `GenericResult::LazySeq` primitive those two issues introduced, so a `map`/`select` chain that stays lazy through `eval_generic.rs` can also stream straight to CLI output instead of buffering through `OwnedValue`/`JqValue`.

`yq_runner.rs`'s `can_use_m2_streaming` whitelist (the mechanism #685 added for `keys_unsorted` output streaming) does not currently know about `LazySeq` at all. Right now, `GenericResult::stream_json`/`stream_yaml`'s own `LazySeq` arm (added by #724/#725) already avoids the four-pass `to_owned → reserialize → reindex → re-evaluate` round trip, but it does so via one atomic `materialize_atomic()` pass into a single in-memory `OwnedValue::Array` before streaming — not the zero-buffer, cursor-at-a-time output `can_use_m2_streaming` gives `keys_unsorted`. See `docs/plan/jq-lazy-map-select.md`'s design section "Where this plugs in" and its own explicit call-out:

> **Output streaming needs no new type family.** A `stream_lazy_seq_json`/`_yaml` generalizes `stream_lazy_keys_json`/`_yaml`'s existing idiom to `LazySeq`, and `GenericResult::stream_json`/`stream_yaml` gain one arm each calling it.

That plan turned out to be in tension with `map`'s atomicity requirement (real jq's array construction is all-or-nothing — `[1,2,"x"]|map(.+1)` prints nothing on error, not a truncated prefix), which a byte-at-a-time writer can't honor without buffering at least one element ahead. #724/#725 shipped the atomic-materialize-then-stream compromise instead and left true zero-buffer CLI streaming to this issue.

## Scope

- `yq_runner.rs`'s `can_use_m2_streaming` whitelist (and `jq_runner.rs`'s CLI-boundary materialization arms, `evaluate_input`/`generic_result_to_jq_values`) need a `LazySeq`-aware path that can still stream true M2-style output for the common non-erroring case, while preserving `map`'s atomicity on the error path (no truncated-prefix output).
- Reuse `stream_lazy_keys_json`/`_yaml`'s existing idiom (`src/jq/stream.rs`) as the template, per the design doc.
[...]

Issue #715, opened 2026-08-09
yq: missing --front-matter, real --split-exp (collides with -s), and eval-all/fileIndex


newhoggy:
**Severity: Enhancement** — three distinct pieces of real `yq`'s CLI surface with no succinctly equivalent, bundled here since they're all "feature absent" (not "feature broken") and each is individually smaller than the other issues from this audit.

## 1. `--front-matter` (extract/process modes)

Real `yq` can operate on YAML embedded as front matter inside another file (e.g. Markdown with a `---`-delimited YAML header), splitting it off, running the expression, and (in `process` mode) re-appending the untouched trailing content. succinctly has no equivalent:

```
$ grep -n "front_matter\|FrontMatter\|front-matter" src/bin/succinctly/*.rs src/jq/*.rs src/yaml/*.rs
(no output)
```

## 2. `-s`/`--split-exp` — and a naming collision worth flagging on its own

Real `yq -s '"file_" + $index' myfile.yml` splits each result into its own output file, named by evaluating the given expression. succinctly's `-s` is already taken for something else entirely — `--slurp` (matching jq's `-s`, reading all inputs into a single array) — confirmed tested at `tests/yq_cli_tests.rs:663,1000`. This means a user coming from real `yq` muscle memory who types `succinctly yq -s ...` gets silently different behavior, not an error. Worth deciding explicitly whether real `yq`'s split-to-files feature is in scope, and if so what flag it should use given `-s` is unavailable.

## 3. `eval-all`/`ea` and `fileIndex`/`fi` (cross-file merge workflow)

Real `yq` defaults to `eval` (apply the expression once per document/file, low memory) but offers `eval-all`/`ea` to load every document from every file into one evaluation context, primarily for cross-file merges:

```
yq eval-all 'select(fileIndex == 0) * select(fileIndex == 1)' f1.yml f2.yml
```

succinctly has no `fileIndex`/`fi` builtin (`grep -n "fileIndex\|FileIndex" src/jq/*.rs src/bin/succinctly/*.rs` → no relevant hits — the only `file_index` hits are unrelated, in the `json generate`/`dsv generate` seed-derivation code) and thus no way to express this idiom. To be clear about what *does* work: per-file `eval`-style multi-file arguments already work correctly today —

```
$ succinctly yq '.a' f1.yaml f2.yaml
1
---
2
```

— this issue is specifically about the `eval-all` cross-document-context mode, not multi-file support in general.

## Notes

[...]

Issue #478, opened 2026-07-29
yq/jq: --slurp/--inplace and jq --preserve-input pretty output still collapse duplicate mapping keys


newhoggy:
**Severity: Medium (silent data loss)** — follow-up to #442, which fixed the plain `yq -o json '.'` / `yq '.'` pretty-print case (identity and simple navigation queries streamed directly from the cursor instead of going through `OwnedValue::Object`'s `IndexMap`, which cannot hold duplicate keys).

#442's own "Cause" section named three call sites that build an `OwnedValue`/`JqValue` via `.insert()`/`.collect()` into an `IndexMap`, and therefore silently drop all but the last occurrence of a duplicate mapping key:

1. `to_owned()` in `src/jq/eval_generic.rs` — used by the yq pretty-print identity/navigation path. **Fixed by #442** (for identity and simple navigation: `.`, `.field`, `.[n]`, `.[]`, and pipes/parens/`?` of those).
2. `yaml_to_owned_value()` in `src/bin/succinctly/yq_runner.rs` — used by `--slurp` (`-s`, "read all inputs into an array") and `--inplace`. **Still broken**, unchanged by #442's fix:
   ```
   $ printf 'a: 1\na: 2\n' | succinctly yq --slurp '.'
   -
     a: 2                            # should keep both `a:` entries in the slurped element
   ```
   (Note: real `mikefarah/yq`'s own `-s` flag means something different — `--split-exp`, for splitting output into files — so there's no direct real-`yq` A/B here; succinctly's `--slurp` is a jq-style flag layered on top. The bug is the internal inconsistency: `succinctly yq '.'` on this input now correctly keeps both `a:` entries since #442, but `succinctly yq --slurp '.'` on the same input silently drops one, because it goes through a different, unfixed conversion function.)

   `--inplace` *does* have a direct real-`yq` comparison, and it diverges:
   ```
   $ printf 'a: 1\na: 2\n' > /tmp/d.yaml && succinctly yq --inplace '.' /tmp/d.yaml && cat /tmp/d.yaml
   a: 2                             # real yq --inplace '.' keeps both: "a: 1\na: 2\n"
   ```
[...]

PR #1694, opened 2026-08-27
fix(yq): extend --slurp's fast path to -o=json, preserving duplicate keys

### LINKED ISSUE #1577: yq: --slurp -o=json has no fast path, so it still collapses duplicate mapping keys
**Severity: Medium** — duplicate mapping keys are silently collapsed for
`--slurp -o=json`, where the YAML-output side of the same flag preserves them.

## Background

`--slurp`'s fast path (#478, extended by #1115) is YAML-output only:
`can_slurp_fast_path` in `yq_runner.rs` requires
`output_format == OutputFormat::Yaml`, because `stream_yaml_sequence` was the
only sequence writer that existed. `-o=json --slurp` falls through to the DOM
path, which builds `OwnedValue::Array(all_docs)` — and an `IndexMap` cannot
represent a repeated key.

## What changed

#757 added `stream_json_sequence` to `src/yaml/light.rs` — the JSON
counterpart of `stream_yaml_sequence`, same signature, built on
`YamlCursor::stream_json_value` so it handles every `-I` width including
compact. The missing half of the pair now exists, so extending
`can_slurp_fast_path` to `-o=json` is mostly wiring rather than new
rendering code.

## Repro (confirm against pinned yq v4.53.3 before implementing)

```console
$ printf 'a: 1\na: 2\n' > d.yaml
$ yq            --slurp -o=json -I=0 '.' d.yaml
$ succinctly yq --slurp -o=json -I=0 '.' d.yaml
```

Real yq preserves duplicate keys in JSON output elsewhere
(`yq -o=json -I=0 '.[]'` on a duplicate-keyed document prints
`{"a":1,"a":2}`), so the expected direction is preserve — but capture the
oracle's actual `--slurp -o=json` output rather than assuming it, per
ADR-0018.

## Scope

- Extend `can_slurp_fast_path` to `OutputFormat::Json`, routing through
  `stream_json_sequence` the way the YAML branch routes through
  `stream_yaml_sequence`.
- Keep the buffer-and-colorize wrapper (#748/#809) — `stream_json_sequence`
  is generic over `core::fmt::Write`, so it slots into `stream_maybe_colored`
  unmodified, exactly as its YAML sibling does.
- Golden cases for `--slurp -o=json` at `-I=0` and pretty.

Related: #1343 tracks the wider `-s`/`--eval-all`/`-i` DOM-fallback
duplicate-key gap; this issue is the one slice of it that now has a ready
writer.

newhoggy:
## Summary

- `can_slurp_fast_path` (#478) required YAML output, so `-o=json --slurp` fell through to the DOM path's `IndexMap`-backed array builder, which collapses a duplicate mapping key to just its last value (`a: 1\na: 2` → `{"a":2}`) — unlike YAML output on the identical input, which already preserved both. Verified live against real yq's own equivalent (`yq -o=json ea -I=0 '[.]'`, since v4.53.3 has no `--slurp` flag at all — succinctly's own `-s`/`--slurp` deliberately overloads real yq's `-s` for jq-style slurp semantics, per #715): real yq preserves the duplicate key there too.
- `stream_json_sequence` (#757) already exists as `stream_yaml_sequence`'s JSON-writing sibling, built for exactly this gap, so extending `can_slurp_fast_path` to `OutputFormat::Json` is mostly wiring — dispatches on output format at the one call site, using `json_indent_spaces`/`colorize_json` instead of their YAML counterparts, and mirrors `can_json_fast_path`'s own `compact || (pretty-or-colored && !ascii_output)` split (the YAML arm's condition is unchanged from before this PR).
- Updated 4 pre-existing `_through_dom_path` tests whose plain `.` filter now takes the (correct) new fast path instead of the DOM path they were written to exercise — added an unused `--arg` to keep forcing DOM (`context.named.is_empty()`), the same trick two other tests in this file already use for the identical purpose. Verified live that output is unchanged either way (these tests check DOM-path-specific float/tag/alias resolution, not fast-vs-DOM path selection itself).
- Updated `test_slurp_color_output_json`, which previously pinned the old (wrong) collapsing behavior under `-C`, to `test_duplicate_mapping_key_survives_slurp_json_color` pinning the fixed one.
- Added new tests: `test_duplicate_mapping_key_survives_slurp_json` (compact + pretty) and `test_duplicate_mapping_key_survives_slurp_json_multiple_sources`, mirroring the existing YAML-output pair.
- Filed #1693 for a pre-existing, unrelated gap this surfaced while testing: `--ascii-output` does nothing in the compact JSON M2 fast path (reproduces identically on unmodified `main`, for the ordinary non-slurp identity case too — not introduced or widened by this PR).

Fixes #1577.

## Test plan

- [x] All 41 `slurp`-named tests in `yq_cli_tests.rs` pass, including the 3 new/updated ones.
- [x] Live-verified against pinned yq v4.53.3's actual equivalent operation (`-o=json ea -I=0 '[.]'`) for both compact and pretty indent.
- [x] Full `cargo test --features cli,regex,serde` — all green (one transient `cargo run` lock-contention flake on an unrelated jq-error-wording test, reproduced clean in isolation).
- [x] `yq_golden_tests`/`yq_path_consistency_tests` — clean.
- [x] `cargo clippy --all-targets --all-features -- -D warnings` and the scalar-yaml CI leg — clean.
- [x] `cargo build --no-default-features`, `cargo fmt --check`, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` — clean.
- [x] Patch coverage: 96.15% (25/26 new lines). The one uncovered line is the JSON branch's `stream_maybe_colored(...)?` — the `?` operator's I/O-error propagation path, genuinely untestable without mocking a broken writer, matching the pre-existing (also uncovered) YAML branch's identical shape.

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -3283,15 +3283,24 @@ pub fn run_yq(args: YqCommand) -> Result<i32> {
     // `stream_yaml_sequence`/`stream_json_sequence` in `stream_maybe_colored`,
     // same as the stdout/inplace paths. Unlike the other gates above, this
     // one never OR's in `output_config.compact` for YAML -- preserved as-is
-    // from before #1577, which only adds the JSON arm and its established
-    // `compact || (pretty-or-colored && !ascii_output)` split (matching
-    // `can_json_fast_path`'s identical asymmetry: `--ascii-output`'s
-    // escaping isn't implemented by the pretty/colored streaming path, only
-    // by the compact one).
+    // from before #1577.
+    //
+    // The JSON arm deliberately does NOT copy `can_json_fast_path`'s
+    // `compact || (pretty-or-colored && !ascii_output)` shape verbatim: that
+    // condition already lets `--ascii-output` through unescaped in compact
+    // mode today (#1693, a live, pre-existing bug -- none of the M2 JSON
+    // streamers implement ASCII escaping at all, contrary to what that
+    // condition's shape implies). Copying it here would have been a new
+    // regression, not an inherited gap: before this arm existed, `--slurp
+    // -o=json` had no fast path in *any* combination, so `--ascii-output`
+    // always reached the DOM path's correct escaping. `!args.ascii_output`
+    // is therefore required unconditionally, keeping that combination on
+    // the DOM path exactly as it was pre-#1577, until #1693 gives the M2
+    // streamers real ASCII-escaping support to extend this to.
     let can_slurp_fast_path = is_identity
         && match output_config.output_format {
             OutputFormat::Json => {
-                output_config.compact || (can_stream_pretty_or_colored && !args.ascii_output)
+                !args.ascii_output && (output_config.compact || can_stream_pretty_or_colored)
             }
             OutputFormat::Yaml => can_stream_pretty_or_colored,
             _ => false,
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -23,7 +23,7 @@ use succinctly::jq::{
 use succinctly::json::JsonIndex;
 use succinctly::yaml::{
     format_float_yq_yaml, format_float_yq_yaml_nested, resolve_plain, resolve_tagged,
-    stream_yaml_sequence, YamlCursor, YamlIndex, YamlValue,
+    stream_json_sequence, stream_yaml_sequence, YamlCursor, YamlIndex, YamlValue,
 };
 
 use super::{FrontMatterMode, InputFormat, OutputFormat, YqCommand};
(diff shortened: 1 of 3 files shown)


newhoggy:
Ran two more rounds of `/code-review` on top of the ascii-output regression fix. Both rounds confirmed the fix is correct and complete — no further correctness bugs found across independent verification passes and live flag-matrix testing (compact/pretty × color/no-color × ascii-output × doc-select × multi-source).

Remaining findings were altitude/DRY-caliber only:
- **Fixed**: `test_slurp_color_output_json` (renamed to `test_duplicate_mapping_key_survives_slurp_json_color`) was, pre-PR, the only test covering `output_value`'s JSON `use_color` DOM branch — `--slurp -o=json` had no fast path at all before this change. Once `can_slurp_fast_path` started accepting JSON, that test moved onto the new streaming path and silently stopped exercising the DOM color branch. Added `test_null_input_color_output_json`, mirroring the existing `test_null_input_color_output_yaml` fix for the identical YAML-side gap from #809 (`--null-input` has no cursor to stream from, so it always stays on the DOM path).
- **Declined as out of scope**: several suggestions to unify the now-5 near-identical fast-path gate booleans (`can_json_fast_path`/`can_yaml_fast_path`/`can_inplace_json_fast_path`/`can_inplace_yaml_fast_path`/`can_slurp_fast_path`) and to de-duplicate the repeated `colorize_json`-default-scheme closures / format-dispatch shape in `yq_runner.rs`. These are pre-existing patterns this PR extends rather than introduces, and none are correctness bugs — a real refactor there is a separate, dedicated task rather than scope for a targeted duplicate-key-preservation fix.
- **Declined as out of scope**: recording #1693 (pre-existing `--ascii-output` no-op in the *sibling* compact-JSON fast paths, confirmed reproducible on unmodified `main`) in `docs/compliance/yq/limitations.md`. It's already fully tracked as its own issue with repro/root-cause/fix-direction, and since `--ascii-output` is itself a succinctly-only extension with no real-`yq` behavior to diverge from, it doesn't cleanly fit that doc's ADR-0018 divergence-recording framing.

Pushed as `9e3e0cb6e`. Full local verification (build/test/clippy/fmt/no_std/rustdoc) green.

Issue #1546, opened 2026-08-23
test(cli): same unwrap_or(-1)/101-retry pattern duplicated in 3 more test files; false-pass in test_as_pattern_alt_rejected_in_yq_mode_720


newhoggy:
**Severity: Low** (test-infrastructure only, same class as #1516)

## Summary

#1516 fixed the signal-death misdiagnosis (`.code().unwrap_or(-1)` silently coercing a
killed child to a fake exit code) in `tests/jq_cli_tests.rs`'s four `cargo run`-based
helpers (`run_jq_stdin_streams`, `run_jq_file`, `run_jq_null`, `run_jq_binary_stdin`) plus
`run_jq_full`, via a shared `classify_cargo_run_exit`/`signal_death_error`. Its own
`/code-review` (7-agent fan-out) surfaced several more sites carrying the identical
pattern, deliberately left out of that PR's scope:

## Sites still carrying the bug

- `tests/cli_golden_tests.rs` — `run_cli` (same `unwrap_or(-1)` + `code == 101` retry,
  hand-rolled) and `run_cli_bin` (same `unwrap_or(-1)`, no retry at all).
- `tests/cli_characterization_tests.rs` — `run` (same hand-rolled pattern).
- `tests/deep_nesting_valid_tests.rs` — `run` (same pattern, returns `(stdout, exit_code)`
  straight to callers with no stderr surfaced at all — the least diagnostic of any site).

None of these currently retry or report a signal death as such; a killed child on any of
them still renders as an inscrutable `-1` with no signal name.

## A pre-existing false-pass bug found in the same review

`tests/jq_cli_tests.rs`, `test_as_pattern_alt_rejected_in_yq_mode_720`: asserts
`assert_ne!(code, 0)` to prove yq mode rejected `?//` as a parse error. If the child is
instead killed by an unrelated signal (OOM, `pkill`), `code()` is `None` → `-1`, which
still satisfies `assert_ne!(_, 0)` — the test passes for the wrong reason, silently masking
a signal death as a successful correctness check. Worth an audit for other `assert_ne!(code,
0)`/`assert!(code != 0)` sites with the same shape.

## Also noted, not necessarily actionable

- Several ad hoc test call sites within `jq_cli_tests.rs` itself (e.g. around
  `test_unary_minus`, `test_builtin_first_on_non_array_errors`) spawn `cargo run`/the built
  binary directly and inspect `.code()` without going through any of the shared helpers —
[...]

PR #1695, opened 2026-08-27
test(cli): fix false-pass exit-code checks, sweep .code().unwrap_or(-1) repo-wide

### LINKED ISSUE #1691: test(cli): broader .code().unwrap_or(-1) false-pass audit — 3 more sites in yq_cli_tests.rs, pattern spread across 7 other test files
**Severity: Low** (test-infrastructure only, same class as #1516/#1546)

## Summary

While fixing #1546 (dedup `classify_cargo_run_exit`/`signal_death_error` across
`jq_cli_tests.rs`, `cli_golden_tests.rs`, `cli_characterization_tests.rs`,
`deep_nesting_valid_tests.rs`), a repo-wide grep for the same `.code().unwrap_or(-1)`
pattern turned up meaningfully more scope than that issue named — both the false-pass
shape specifically, and the coercion pattern generally.

## More false-pass sites, same shape as #1546's `test_as_pattern_alt_rejected_in_yq_mode_720`

`tests/yq_cli_tests.rs` has three more `assert_ne!(output.status.code().unwrap_or(-1), 0)`
sites — the exact "a signal-killed child coerces to `-1`, which still satisfies
`assert_ne!(_, 0)`, so the test passes for the wrong reason" bug #1546 fixed one instance
of:

- `test_jq_urid_base64d_still_reject_non_string_1109` (`tests/yq_cli_tests.rs:16809`) —
  two `assert_ne!` sites inside this one test.
- `test_jq_base64d_does_not_trim_whitespace_1123` (`tests/yq_cli_tests.rs:19307`).

All three spawn `CARGO_BIN_EXE_succinctly` directly (no `cargo run`, no retry loop), same
shape as #1546's fixed test — the fix is the same one-line change: check
`output.status.code()` and error (naming the signal) rather than defaulting to `-1` on
`None`.

## The broader `.code().unwrap_or(-1)` pattern, beyond #1546's four named files

A plain coercion (not necessarily paired with `assert_ne!`, so not all false-pass, but all
liable to render a signal death as an inscrutable `left: -1, right: ...` panic instead of
naming the signal — #1516's original diagnosis) appears in:

- `tests/dsv_cli_tests.rs`
- `tests/orchestrate_cli_tests.rs`
- `tests/locate_cli_tests.rs`
- `tests/json_validate_tests.rs` (4 sites, including 2 `assert_eq!(..., 0)`/`assert_eq!(..., 1)` — those aren't false-pass since they check a specific value rather than "not equal", but still coerce a signal death to a wrong number)
- `tests/text_cli_tests.rs` (4 sites)
- `tests/yaml_validate_tests.rs` (4 sites)
- `tests/yq_cli_tests.rs` (many more beyond the 3 named above — this file is large and has its own multiple `run`-style helpers)

## Suggested fix direction

Same shared `tests/common/cargo_run_exit.rs` module #1546 introduces — most of these
sites should be straightforward to route through `classify_cargo_run_exit`/
`signal_death_error` (add `#[path = "common/cargo_run_exit.rs"] mod cargo_run_exit;` and
call it), the same one-line-per-site change #1546 applied to `run_cli_bin`. A few (the
`json_validate_tests.rs`/`text_cli_tests.rs`/`yaml_validate_tests.rs` sites) may be
single-shot spawns like #1546's own false-pass fix rather than retry loops — those need
the direct `let Some(code) = status.code() else { return Err(signal_death_error(...)) }`
treatment instead of the retry-loop classifier.

## Why this is separate from #1546

#1546 scoped itself to four specific files named in its own body (following on from
#1516's original four-helper fix in `jq_cli_tests.rs`). Widening it to a full
repo-wide audit mid-flight would have ballooned that PR's diff well past its stated
scope. This issue is the audit #1546's own body suggested ("worth an audit for other
`assert_ne!(code, 0)`/`assert!(code != 0)` sites with the same shape") but did not commit
to doing everywhere.

## Related

- #1516 — the original fix, four helpers in `jq_cli_tests.rs`.
- #1546 — extended the fix (and the shared module this issue's fix should also use) to
  three more files plus one false-pass test.
- #550 — the original ENOENT-at-spawn() issue in `jq_cli_tests.rs`, unrelated mechanism
  but same file/theme.

newhoggy:
## Summary

Fixes #1691, the follow-up audit filed during #1546's own review.

- **`tests/yq_cli_tests.rs`**: fixes 5 `assert_ne!(code, 0)` false-pass sites — a signal-killed child's `None` exit code coerces to `-1` via `.unwrap_or(-1)`, which still satisfies `assert_ne!(_, 0)` and passes the test for the wrong reason. Two of the five weren't named in #1691's own body (found by re-running its grep against this file specifically). Also routes the file's 7 `run_*`/spawn helpers and 2 more inline `assert_eq!` sites through the same `signal_death_error` helper `#1546` introduced, so a killed child names its signal instead of an inscrutable `left: -1, right: 0`.
- **6 more files** (`dsv_cli_tests.rs`, `orchestrate_cli_tests.rs`, `locate_cli_tests.rs`, `json_validate_tests.rs`, `text_cli_tests.rs`, `yaml_validate_tests.rs`): same coercion pattern, per #1691's own repo-wide grep. Two sites in `yaml_validate_tests.rs` turned out to be the same false-pass shape (`assert_eq!(code, 0)` isn't false-pass, but `yaml_validate_tests.rs` also fixed two `run_*` helpers). `orchestrate_cli_tests.rs` and `locate_cli_tests.rs` needed their bare-tuple `run` helper (and its callers) converted to `Result<(...)>` so the error can propagate with `?` — no other helper needed a signature change, since they already returned `Result`.

## Why not the retry-loop classifier for the two `json_validate_tests.rs` `cargo run` sites

`test_multiple_files_all_valid`/`test_multiple_files_one_invalid` spawn `cargo run` directly rather than the pre-built binary, but neither had a retry loop to begin with (unlike `jq_cli_tests.rs`'s `run_jq_file`/`run_jq_null`). Adding one now would fix a separate, pre-existing flakiness concern (`cargo run`'s exit-101 lock contention) this issue isn't about — so these got the same minimal `signal_death_error` treatment as every other single-shot site instead.

## Test plan

- [x] `cargo test --release --features cli,bench-runner,simd,regex,serde --no-fail-fast` — all green except one pre-existing, unrelated `#[should_panic]` failure on `src/bits/select.rs` confirmed to also fail on unmodified `main` in `--release` mode (a `debug_assert` that release strips).
- [x] `cargo test --features cli --test yq_cli_tests --test yq_golden_tests --test yq_path_consistency_tests --test jq_golden_tests --test yaml_bench_suite_coverage --test dsv_cli_tests --test text_cli_tests --test json_validate_tests --test yaml_validate_tests` (CI's own combined test invocation) — green.
- [x] `cargo clippy --all-targets --features cli,simd,regex,serde -- -D warnings` and the `--all-features` leg — clean.
- [x] `cargo build --no-default-features` — no new warnings vs. `main`.
- [x] `cargo fmt --check`, `cargo doc --no-deps --features cli,simd,regex,serde` — clean.

Closes #1691.

### DIFF
--- a/tests/common/cargo_run_exit.rs
+++ b/tests/common/cargo_run_exit.rs
@@ -48,6 +48,33 @@ pub fn signal_death_error(status: std::process::ExitStatus, stderr: &str) -> any
     )
 }
 
+/// Extracts a real exit code from `status`, or builds [`signal_death_error`]
+/// from the *raw* stderr bytes -- checked before any caller attempts a
+/// strict `String::from_utf8` decode of the child's stdout/stderr (#1691
+/// code review). A signal-killed child can leave a truncated multi-byte
+/// UTF-8 sequence at the end of a buffer it was writing when killed; if the
+/// caller decodes first, that decode's own `?` fires before this check ever
+/// runs, surfacing a generic `FromUtf8Error` instead of naming the signal --
+/// exactly the diagnostic loss this module exists to prevent. Taking `&[u8]`
+/// rather than `&str` means the error path never needs its own successful
+/// decode: `String::from_utf8_lossy` cannot fail.
+///
+/// Also collapses the ~29 hand-written `let Some(code) = status.code() else
+/// { ... }` copies #1691's own fix introduced across 7 files into one
+/// definition, per that same review round.
+pub fn exit_code_or_signal_death(
+    status: std::process::ExitStatus,
+    raw_stderr: &[u8],
+) -> Result<i32> {
+    match status.code() {
+        Some(code) => Ok(code),
+        None => Err(signal_death_error(
+            status,
+            &String::from_utf8_lossy(raw_stderr),
+        )),
+    }
+}
+
 /// Classifies a `cargo run` child's exit for a retry loop, so each call site
 /// doesn't hand-roll its own "retry on 101" check.
 ///
--- a/tests/dsv_cli_tests.rs
+++ b/tests/dsv_cli_tests.rs
@@ -13,6 +13,10 @@ use std::process::{Command, Stdio};
 
 use anyhow::Result;
 
+#[path = "common/cargo_run_exit.rs"]
+mod cargo_run_exit;
+use cargo_run_exit::signal_death_error;
+
 /// Path to the pre-built `succinctly` CLI binary. Cargo builds the `succinctly`
 /// bin target (gated `required-features = ["cli"]`) before this test binary
 /// runs, since this file is itself gated on `cli`, and bakes the resulting
@@ -30,9 +34,11 @@ fn run_generate(size: &str, extra_args: &[&str]) -> Result<(String, String, i32)
         .stderr(Stdio::piped())
         .output()?;
 
-    let exit_code = output.status.code().unwrap_or(-1);
     let stdout = String::from_utf8_lossy(&output.stdout).into_owned();
     let stderr = String::from_utf8_lossy(&output.stderr).into_owned();
+    let Some(exit_code) = output.status.code() else {
+        return Err(signal_death_error(output.status, &stderr));
+    };
     Ok((stdout, stderr, exit_code))
 }
 
--- a/tests/dsv_cli_tests.rs
+++ b/tests/dsv_cli_tests.rs
@@ -15,7 +15,7 @@ use anyhow::Result;
 
 #[path = "common/cargo_run_exit.rs"]
 mod cargo_run_exit;
-use cargo_run_exit::signal_death_error;
+use cargo_run_exit::exit_code_or_signal_death;
 
 /// Path to the pre-built `succinctly` CLI binary. Cargo builds the `succinctly`
 /// bin target (gated `required-features = ["cli"]`) before this test binary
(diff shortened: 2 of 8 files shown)


newhoggy:
Ran \`/code-review\` — 10 finder angles, all substantive findings addressed:

- **Fixed** (decode-before-check ordering): several helpers decoded stdout/stderr with a strict \`String::from_utf8(...)?\` *before* checking \`output.status.code()\`. A signal-killed child can leave a truncated multi-byte UTF-8 sequence at the end of a buffer it was writing when killed, so that decode's own \`?\` could fire first and mask the more useful "child was killed by signal N" diagnostic behind a generic \`FromUtf8Error\` — defeating this PR's own purpose in exactly the scenario it exists for.
- **Fixed** (DRY): the ~29 guard blocks this PR introduced across 7 files were each hand-written, reproducing the exact copy-paste drift #1516/#1546 were about. Added \`cargo_run_exit::exit_code_or_signal_death(status, raw_stderr: &[u8]) -> Result<i32>\`, which checks the exit code before any decode and collapses every site to one call.
- **Fixed** (real flakiness gap): \`json_validate_tests.rs\`'s two \`cargo run\`-spawning tests had no retry-on-101 loop at all, unlike every other \`cargo run\`-based helper in this codebase (\`jq_cli_tests.rs\`, \`cli_golden_tests.rs\`, \`cli_characterization_tests.rs\`, \`deep_nesting_valid_tests.rs\`). Gave them the same \`classify_cargo_run_exit\` retry treatment.
- **Filed as #1696** (out of scope for this test-only PR): the Altitude-angle finder caught the identical \`.code().unwrap_or(-1)\` bug live in production code — \`src/bin/succinctly/bench_runner/orchestrate/ssh.rs:241\`. Real bug, different subsystem; tracked separately.
- No CI guard (grep-based lint) was added for this pattern — a reasonable idea (raised by the same finder) but out of scope for a targeted test fix; noted in #1696 as a suggested direction rather than added here.

Pushed as \`d3f52e41d\`. Full local verification (build/test/clippy-both-legs/fmt/doc, plus CI's own exact combined test invocation) green.

PR #1698, opened 2026-08-27
fix(yq): make --ascii-output gate the whole M2 JSON fast-path condition

### LINKED ISSUE #1693: yq: --ascii-output does nothing in the compact JSON M2 fast path

newhoggy:
## Summary

Fixes #1693.

`can_json_fast_path` and `can_inplace_json_fast_path` both had the shape `compact || (pretty-or-colored && !ascii_output)` — the `compact ||` half let `--ascii-output` through unescaped whenever compact mode (`-I0`) was also requested, since none of the M2 JSON streamers (`stream_json_value`, `stream_json_sequence`) implement ASCII escaping; only the DOM path's `output_value` does, and it was only reached when the whole condition was already false.

Fixed by moving `!args.ascii_output` outside the disjunction so it gates unconditionally — the exact same shape `can_slurp_fast_path` already used after #1577's own fix for this bug in that one gate. All three JSON fast-path gates are now consistent.

```console
$ printf 'k: "héllo"\n' > g.yaml
$ succinctly yq -o=json -I=0 --ascii-output '.' g.yaml
{"k":"héllo"}   # was: {"k":"héllo"} (unescaped, wrong)
```

## Why not thread ASCII escaping through the M2 streamers instead

That's the more complete fix (#1693's suggested direction 1), but investigating it turned up a real cost: `stream_json_string` is a hand-tuned, deliberately *separate* copy of the same escaping logic `jq::escape::write_json_body_yq` implements — kept unshared specifically because #965 measured up to 14% regression from letting the two call sites share one inlined copy. Adding ASCII-escaping support there would mean writing a third such copy from scratch (correct UTF-8 decoding, surrogate pairs for code points above U+FFFF, its own SIMD-tuned scan), a much larger and riskier change than this issue's stated "Low, cosmetic" severity warrants. Falling back to the DOM path for this one flag combination is the documented, accepted tradeoff (#1693's own suggested direction 2).

## Test plan

- [x] Live-verified all 5 affected combinations (stdout compact, stdout pretty, inplace compact, non-ascii-output compact, slurp compact) before and after the fix.
- [x] Added `test_ascii_output_still_escapes_in_compact_mode_1693` and `test_ascii_output_still_escapes_in_compact_inplace_mode_1693`.
- [x] `cargo test --release --features cli,bench-runner,simd,regex,serde --no-fail-fast` — green except the pre-existing, unrelated `bits::select` `#[should_panic]` release-mode failure (confirmed on unmodified `main`).
- [x] `cargo clippy` both legs, `cargo fmt --check`, `cargo doc --no-deps`, `cargo build --no-default-features` — all clean, no new warnings vs. `main`.

Closes #1693.

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -3224,11 +3224,33 @@ pub fn run_yq(args: YqCommand) -> Result<i32> {
     // implement ASCII escaping, so `compact ||` previously let
     // `--ascii-output` through unescaped whenever `-I=0`/compact mode was
     // also requested -- the DOM path (`output_value`) is the only place
-    // that escapes correctly, and it's reached only when this whole
-    // condition is false.
+    // that escapes correctly, and it's reached only when this is false.
+    //
+    // Falling back to DOM for `--ascii-output` reopens the duplicate-key
+    // collapse `can_stream_pretty_or_colored`'s own comment above warns
+    // about (#442/#748/#809) -- but only for this one flag, and only in a
+    // way that already shipped: pretty mode's `can_stream_pretty_or_colored
+    // && !ascii_output` had exactly this same DOM fallback (and therefore
+    // the same duplicate-key loss) for `--ascii-output` before #1693 ever
+    // touched this function. This shared local makes compact mode
+    // consistent with that pre-existing behavior instead of leaving the two
+    // output styles to diverge on the same input. #1700 tracks the real fix
+    // (ASCII-escaping support in the M2 streamers themselves, which would
+    // let `--ascii-output` keep both correct escaping and duplicate-key
+    // safety at once) -- not attempted here per #1693's own "Low, cosmetic"
+    // scoping and the same #965 inlining-cost reasoning this file's
+    // `stream_json_string` doc comment records for why that isn't a
+    // drop-in change.
+    //
+    // Extracted into one local (rather than repeating the expression at
+    // `can_json_fast_path`, `can_inplace_json_fast_path`, and
+    // `can_slurp_fast_path`'s JSON arm) per /code-review on #1693's own PR:
+    // `can_slurp_fast_path` originally diverged from this exact predicate
+    // because #1577 copied it by hand instead of sharing it.
+    let can_stream_json_output_style =
+        !args.ascii_output && (output_config.compact || can_stream_pretty_or_colored);
     let can_json_fast_path = is_m2_streamable
-        && !args.ascii_output
-        && (output_config.compact || can_stream_pretty_or_colored)
+        && can_stream_json_output_style
         && output_config.output_format == OutputFormat::Json
         && !args.null_input
         && !args.raw_input
@@ -3265,10 +3287,9 @@ pub fn run_yq(args: YqCommand) -> Result<i32> {
     // writes), so `-C` reaching the fast path still never writes ANSI to
     // disk, but no longer forces a fallback to the duplicate-key-collapsing
     // DOM path either.
-    // Same `!args.ascii_output` fix as `can_json_fast_path` above (#1693).
+    // Shares `can_json_fast_path`'s `can_stream_json_output_style` above.
     let can_inplace_json_fast_path = is_m2_streamable
-        && !args.ascii_output
-        && (output_config.compact || can_stream_pretty_or_colored)
+        && can_stream_json_output_style
         && output_config.output_format == OutputFormat::Json
         && !args.null_input
         && !args.raw_input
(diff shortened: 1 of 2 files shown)


newhoggy:
Ran \`/code-review\` — 10 finder angles. Most substantive finding: fixing the ASCII-escaping bug by falling back to the DOM path reopens \`can_stream_pretty_or_colored\`'s duplicate-key collapse (#442/#748/#809) for compact mode specifically. Live-verified this isn't a new failure mode from a clean baseline, though — **pretty mode already had this exact tradeoff before this PR touched anything**: \`printf 'a: 1\\na: 2\\n' | succinctly yq -o json --ascii-output '.'\` already collapses to \`{"a": 2}\` on unmodified \`main\`, since pretty mode's own gate already forced DOM for \`--ascii-output\`. This PR just made compact mode consistent with that pre-existing behavior instead of leaving the two output styles diverge on the same input.

Still a real, worth-documenting tradeoff, not something to wave off:
- **Documented explicitly** in a new \`can_stream_json_output_style\` comment, and pinned with \`test_ascii_output_compact_and_pretty_agree_on_duplicate_key_collapse_1693\` so the behavior is an intentional, tested choice rather than an untested surprise.
- **Filed #1700** for the real fix: ASCII-escaping support in the M2 streamers themselves, which would keep both correct escaping and duplicate-key safety at once. Not attempted here — investigated and found non-trivial (\`stream_json_string\` is a deliberately unshared, hand-tuned copy of the escaping logic per #965's measured 14% regression from sharing it; adding ASCII support means a third such copy, not a threading change).
- **Fixed the DRY finding** two independent reviewers raised: extracted the now-identical predicate shared by all three JSON gates (\`can_json_fast_path\`, \`can_inplace_json_fast_path\`, \`can_slurp_fast_path\`'s JSON arm) into one local, \`can_stream_json_output_style\` — the exact kind of hand-copied divergence that caused #1693 itself (#1577 copied the shape by hand instead of sharing it).
- **Filed #1699 and #1701** for two unrelated, pre-existing bugs the same sweep found in this gate family (\`--no-doc\` and \`-0\`/\`--nul-output\` both silently ignored by the M2 fast path) — confirmed live on unmodified \`main\`, out of scope for this PR.

Pushed as \`1cf30f868\`. Full local verification green (one unrelated flaky \`jq_cli_tests\` run from concurrent cargo lock contention, confirmed clean on isolated re-run — not a regression).

PR #1703, opened 2026-08-27
fix(bench): name the signal when a remote exec is killed, not -1

### LINKED ISSUE #1696: bench-runner: SystemSsh::run_with_timeout coerces a signal-killed child's exit code to -1
**Severity: Low** (bench-runner orchestration diagnostics only, same class as #1516/#1546/#1691)

## Summary

Found by `/code-review`'s Altitude-angle pass on PR #1695 (issue #1691's test-infrastructure
audit of the `.code().unwrap_or(-1)` coercion pattern). That audit's grep only covered
`tests/` — this is the same bug pattern, live in production orchestration code.

`SystemSsh::run_with_timeout` in `src/bin/succinctly/bench_runner/orchestrate/ssh.rs:241`:

```rust
Ok(ExecOutput {
    stdout,
    stderr,
    exit_code: status.code().unwrap_or(-1),
})
```

`ExitStatus::code()` returns `None` only when the child was killed by a signal (Unix). This
silently coerces that into a fake `-1` exit code. `ExecOutput::success()`
(`ssh.rs:28-29`) checks `self.exit_code == 0`, so a signal-killed remote `ssh`/`rsync` child
(OOM kill, network drop, another process's `pkill`, or fallout from the same conditions
#1516/#1546/#1691's own doc comments describe for this codebase's heavily concurrent test/bench
sessions) reports as a generic non-zero-exit failure with no indication a signal was ever
involved — exactly the diagnostic loss #1516 was filed to eliminate, just in a different
subsystem.

## The correct pattern already exists next door, unused here

`src/bin/succinctly/bench_runner/runner.rs` (3 call sites: lines ~417, ~475, ~518) keeps the
exit code as `Option<i32>` throughout (`let exit_code = output.status.code();`) rather than
coercing it. `ssh.rs`'s `ExecOutput.exit_code` field is a plain `i32`
(`src/bin/succinctly/bench_runner/orchestrate/ssh.rs:24`), which is what forces the
`unwrap_or(-1)` coercion at its one construction site.

## Suggested fix direction

Either:
1. Change `ExecOutput.exit_code` to `Option<i32>` (matching `runner.rs`'s pattern) and update
   `success()`/the one existing test assertion (`ssh.rs:333`,
   `assert_eq!(out.exit_code, 3)`) accordingly, or
2. Keep the field `i32` but construct a descriptive error (naming the signal, mirroring
   `tests/common/cargo_run_exit.rs`'s `signal_death_error`/`exit_code_or_signal_death`
   helpers from #1691) instead of silently substituting `-1`, since orchestration code has a
   real error-reporting path (`anyhow::Result`) available to it that test helpers didn't
   always have.

Not fixed here — out of scope for #1691's test-only audit; filing as its own issue per that
PR's review.

## Related

- #1516 — the original test-infrastructure fix for this pattern.
- #1546 — extended #1516's fix to 3 more test files.
- #1691 — extended further to 7 total test files; found this production instance while
  auditing.

newhoggy:
## Summary

- `SystemSsh::run_with_timeout` coerced `ExitStatus::code() == None` (a
  signal-killed child) into a fake exit code `-1` via `.unwrap_or(-1)`,
  matching the diagnostic-loss pattern #1516/#1546/#1691 fixed across test
  helpers — this was the one live production-code instance of the same bug,
  found and filed as #1696 during #1691's review.
- `run_with_timeout` now returns a descriptive error naming the signal
  (mirroring `tests/common/cargo_run_exit.rs`'s `signal_death_error` from
  #1691) instead of silently building an `ExecOutput` with a fake `-1`.
  Duplicated rather than shared: production code (`src/`) can't depend on a
  `tests/` module.
- No caller changes needed — every caller of `RemoteExec::exec` already
  handles `Result<ExecOutput>`, so a signal death now surfaces through the
  existing `anyhow::Result` error path instead of masquerading as a generic
  nonzero exit.

## Test plan

- [x] `cargo build --features cli,bench-runner`
- [x] Added `localhost_exec_signal_death_names_the_signal` (kills the child
      with `SIGKILL` via `kill -9 $$`, asserts the error names signal 9);
      all 9 `ssh::tests` pass
- [x] `cargo test --features cli,simd,regex,serde` (full suite, CI's actual
      feature set) — 55/55 suites green
- [x] `cargo llvm-cov --features cli,bench-runner --workspace` — every new
      line (the exit-code check and the new `signal_death_error` helper) is
      covered by the new test; no regression in file-level coverage

Fixes #1696

### DIFF
--- a/src/bin/succinctly/bench_runner/orchestrate/ssh.rs
+++ b/src/bin/succinctly/bench_runner/orchestrate/ssh.rs
@@ -9,6 +9,7 @@
 //! any network access.
 
 use super::config::NodeConfig;
+use crate::exit_status::exit_code_or_signal_death;
 use anyhow::{Context, Result};
 use std::fs;
 use std::io::Read;
@@ -208,14 +209,19 @@ fn run_with_timeout(mut cmd: Command, timeout: Duration) -> Result<ExecOutput> {
     let mut stdout_pipe = child.stdout.take().context("child had no stdout pipe")?;
     let mut stderr_pipe = child.stderr.take().context("child had no stderr pipe")?;
 
+    // Raw bytes, decoded only after the exit-code/signal check below: a
+    // signal-killed child can leave a truncated multi-byte UTF-8 sequence
+    // at the end of a buffer it was writing when killed, and a strict
+    // `read_to_string` discards *all* already-read output (not just the
+    // trailing partial sequence) the moment that happens (#1696 review).
     let stdout_handle = std::thread::spawn(move || {
-        let mut buf = String::new();
-        let _ = stdout_pipe.read_to_string(&mut buf);
+        let mut buf = Vec::new();
+        let _ = stdout_pipe.read_to_end(&mut buf);
         buf
     });
     let stderr_handle = std::thread::spawn(move || {
-        let mut buf = String::new();
-        let _ = stderr_pipe.read_to_string(&mut buf);
+        let mut buf = Vec::new();
+        let _ = stderr_pipe.read_to_end(&mut buf);
         buf
     });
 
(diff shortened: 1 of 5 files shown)


newhoggy:
Multi-angle \`/code-review\` round found the first commit's fix was itself shallow: \`run_with_timeout\` decoded stdout/stderr via \`read_to_string()\` **before** checking the exit code, and \`read_to_string\` discards *all* already-read output (not just the trailing bytes) on invalid UTF-8 — exactly what a SIGKILLed child leaves behind mid-write of a multi-byte sequence. The new diagnostic could be handed empty stderr in precisely the case it exists to explain.

Fixed in a second commit:
- Capture raw `Vec<u8>` in the reader threads; defer decoding (`String::from_utf8_lossy`, which cannot fail) until after the exit-code/signal check — matching `tests/common/cargo_run_exit.rs`'s `exit_code_or_signal_death` design from #1691.
- Extracted the shared logic into `src/bin/succinctly/exit_status.rs` so `ssh.rs` and `cargo_run_exit.rs` both call one definition (review flagged the first commit's copy as reopening the #1516→#1546→#1691 drift pattern).
- Added a test that reproduces the truncated-UTF-8 scenario directly (the existing SIGKILL test happened to produce empty stderr either way, so it never would have caught this).
- Restored `sync.rs`'s post-upload-verification error message, which the first commit's early-return `?` had silently dropped.

Also filed #1705 for two unrelated test-infra gaps surfaced while verifying this locally (bench-runner-gated tests never run in CI; a `cargo run` clobber hazard when they're run together with `jq_cli_tests.rs`).

Issue #1512, opened 2026-08-23
policy: is "succinctly yq accepts jq surface real yq lacks" an extension or a divergence? The docs say both


newhoggy:
Surfaced by #1507. This is a classification question, not a bug report — but
it governs roughly 14 builtins, so it is worth settling explicitly rather than
case by case.

## The two contradictory positions, both in-tree

`succinctly yq` implements a substantial amount of jq surface that real yq's
lexer rejects outright. The repo currently classifies that in two incompatible
ways:

**Position 1 — it is a divergence (a bug to close).**
`docs/compliance/yq/limitations.md` files the `gsub`/`scan`/`splits` case under
`## Open divergences (bugs, not decisions)` (line 239), a section introduced as
"gaps to close, listed here so they are not rediscovered from scratch"
(line 168). It is **not** under that same page's `## Extensions` section
(line 537).

**Position 2 — it is an unopposed extension.**
`CLAUDE.md` calls the identical situation "an unopposed succinctly behavior,
not a verified divergence." And `docs/compliance/yq/limitations.md:296` uses
the absence of an oracle to *withhold* fidelity work: "Deliberately **not**
extended to `gsub`/`scan`/`splits` … there's no oracle for succinctly's own
yq-mode extensions to diverge from, so they stay on the jq-style iteration."

The same page, three sections apart, calls it both a gap to close and an
extension to leave alone.

## Scale (measured, not estimated)

I probed 28 filters against Homebrew yq v4.53.3 and the current
`succinctly yq`. yq's lexer rejects **26**; `succinctly yq` runs **14** of
those without error:

`paths`, `getpath`, `getpath([])`, `tostream`, `leaf_paths`, `IN`,
`fromstream`, `truncate_stream`, `limit`, `isempty`, `debug`, `ltrimstr`,
`infinite`, `isnan`

plus `gsub`/`scan`/`splits` (already documented) and the three input builtins
(gated, #1507). This is not an edge case; it is a substantial and apparently
deliberate posture: `succinctly yq` is in practice "jq's language over YAML
documents, plus yq's own builtins."

## Why it matters

Under ADR-0018:

- **If extension (rule 5):** permitted, needs only two things — marked as an
  extension in user-facing docs, and inert with respect to reference-defined
  filters. Both hold for all 14. The work is *documentation*: get them into a
  labelled list.
- **If divergence (rule 4):** none of the three carve-out conditions applies
  (output is readable, nothing is corrupted, no process dies), so all 14 are
[...]

PR #1713, opened 2026-08-27
fix(jq): extend --jq-extensions gate to 14 more real-yq-lexer-rejected builtins

### LINKED ISSUE #1650: yq: ~13 jq-only builtins (inside, startswith, index, strftime, ...) are silently reachable without --jq-extensions
**Severity: Low-Medium**

## Summary

Found by the multi-agent \`/code-review\` run on PR #1648 (#1553's fix), while confirming
\`inside\` (\`contains\`'s natural mirror) correctly stays ungated. Real yq's lexer rejects a
wide surface of jq-only builtins outright, and \`docs/reference/yq-language.md\`'s "Gated jq
Builtins (--jq-extensions)" table documents \`succinctly yq\`'s policy for that class as
"rejects them too by default... passing the flag opts back into the jq-compatible surface"
(#1512). That gate (\`reject_unless_jq_extensions\`, \`src/jq/parser.rs:309\`) currently
covers exactly 15 names (\`IN\`, \`debug\`, \`fromstream\`, \`getpath\`, \`gsub\`, \`infinite\`,
\`isempty\`, \`isnan\`, \`leaf_paths\`, \`limit\`, \`ltrimstr\`, \`paths\`, \`scan\`, \`splits\`,
\`tostream\`/\`truncate_stream\`) — but \`docs/compliance/yq/limitations.md\`'s own
"Generator-argument fan-out" table lists a second group of builtins real yq's lexer
*also* lexer-rejects (\`inside\`, \`rtrimstr\`, \`startswith\`, \`endswith\`, \`index\`, \`rindex\`,
\`indices\`, \`strftime\`, \`strptime\`, \`bsearch\`, \`pow\`, \`range\`, \`nth\`, \`combinations\`),
none of which carry the same \`--jq-extensions\` gate — they parse and run in plain
\`succinctly yq\` with no flag at all.

## Repro (confirmed live against yq v4.53.3 and the current build)

\`\`\`console
$ printf 'x: "abc"\n' | yq            '.x | inside("xabcy")'
Error: 1:6: lexer: invalid input text "inside(\"xabcy\")..."
$ printf 'x: "abc"\n' | succinctly yq '.x | inside("xabcy")'   # no --jq-extensions
true                                                             # runs anyway

$ printf 'x: "abc"\n' | succinctly yq '.x | startswith("a")'   # same gap
true
$ printf 'x: "abc"\n' | succinctly yq '.x | rtrimstr("c")'     # same gap
ab
$ printf 'x: "abc"\n' | succinctly yq '.x | index("b")'        # same gap
1

$ printf '{"a":1}' | succinctly yq 'paths'                     # a *gated* sibling, for contrast
Error: parse error: ... "paths" is not part of yq's syntax; pass --jq-extensions ...
\`\`\`

## Why this matters

Not a fidelity break by itself — succinctly accepting more syntax than real yq is
explicitly permitted where matching real yq's rejection has no correctness stakes
(ADR-0018's extension carve-out). But it's an inconsistency the project's own docs don't
explain: \`paths\`/\`getpath\`/\`ltrimstr\`/etc. need an opt-in flag to reach this extension
surface; \`inside\`/\`startswith\`/\`index\`/etc. reach the identical kind of surface (a jq
builtin real yq's lexer has never heard of) with no flag at all. A user relying on the
gate's stated purpose — "rather than silently accepting broader syntax than the reference
it's meant to match" — gets that guarantee for one arbitrary half of this surface and not
the other.

It also interacts directly with #1553/#1649: any escape-leak or kind-check bug in one of
these ungated builtins (e.g. \`inside\`'s own escape-clearing gap, mirroring \`contains\`'s
pre-#1553 bug) is reachable in default \`succinctly yq\` with no flag, where the gated
siblings would require an explicit opt-in first.

## Suggested fix direction

Decide the actual intended policy first (this is the judgment call, not a mechanical
fix): either (a) extend \`reject_unless_jq_extensions\` to the full "real yq's lexer
rejects this" list in \`limitations.md\`'s table, closing the gap outright, or (b)
determine the current 15-name gate was deliberately scoped narrower (e.g., only
generator-argument or streaming-shaped builtins) and update
\`docs/reference/yq-language.md\`'s table + \`limitations.md\` to state that scope
explicitly rather than leaving the asymmetry undocumented. Whichever is chosen, verify
each of the ~14 currently-ungated names individually against the pinned yq v4.53.3 lexer
before changing gate coverage — some may have real yq syntax under a different spelling
this issue didn't check for.

## Related

#1553 (PR #1648), #1649 — both surfaced while investigating \`contains\`/\`inside\`'s gating
and kind-check behavior.

newhoggy:
## Summary

- `reject_unless_jq_extensions` (#1512) gated 15 jq-only builtin names real
  yq's lexer rejects outright — but `docs/compliance/yq/limitations.md`'s own
  "Generator-argument fan-out" table had been documenting a second group of
  14 names with the identical "lexer-rejected — the builtin does not exist"
  classification that carried no gate at all: `inside`, `startswith`,
  `endswith`, `rtrimstr`, `index`, `rindex`, `indices`, `range`, `nth`,
  `combinations`, `pow`, `strftime`, `strptime`, `bsearch`. They parsed and
  ran in plain `succinctly yq` with no flag.
- Live-verified each of the 14 against the pinned `yq` v4.53.3 binary before
  touching the gate (per the issue's own caution — some could plausibly have
  had real yq syntax under a different spelling; none did — all 14 confirmed
  `lexer: invalid input text`).
- Added `self.reject_unless_jq_extensions("name")?;` at each builtin's
  existing parse call site, matching the established call-before-consume
  pattern used by the original 15.
- Updated `docs/reference/yq-language.md`'s "Gated jq Builtins" table with
  the 14 new entries.

## Test plan

- [x] Verified all 14 rejected by default and accepted with `--jq-extensions`
      against the built binary directly
- [x] Confirmed no existing test exercised any of the 14 without the flag by
      running the full suite and fixing every resulting failure — 7 existing
      tests (`test_raw_input_select`, `test_split_exp_with_null_input`, both
      `test_split_exp_writes_every_result_in_a_multi_value_halt_prefix_*`,
      `test_yq_combinations_n_empty_argument_produces_empty_array_1408`,
      `test_yq_inside_shares_contains_kind_mismatch_rule_1649`,
      `test_yq_nth_stream_empty_n_argument_produces_no_output_1408`) were
      exercising real, intended extension behavior and needed
      `--jq-extensions` threaded through, not behavior changes
- [x] Extended `test_yq_default_rejects_jq_only_builtins_1512` and
      `test_yq_jq_extensions_flag_enables_jq_only_builtins_1512` (#1512's own
      regression tests, previously sampling only 3 of the 15 original names)
      to cover all 30 gated names — closing a pre-existing gap where neither
      the original 15 nor the new 14 had a test actually asserting the
      rejection message
- [x] `cargo test --features cli,simd,regex,serde` (full suite) — 55/55 green
- [x] `cargo llvm-cov` — every new gate call site covered (both the rejection
      and acceptance paths, verified via the two regression tests above)

Fixes #1650

### DIFF
--- a/docs/reference/yq-language.md
+++ b/docs/reference/yq-language.md
@@ -366,18 +366,27 @@ match, `succinctly yq` rejects them too by default (a parse error naming
 `--jq-extensions`); passing the flag opts back into the jq-compatible
 surface (#1512):
 
-| Builtin                                           | Description                               |
-|---------------------------------------------------|-------------------------------------------|
-| `paths`, `paths(f)`, `leaf_paths`                 | Paths to every node / every leaf node     |
-| `getpath(path)`                                   | Value at a path array                     |
-| `tostream`, `fromstream(f)`, `truncate_stream(f)` | Streaming event representation            |
-| `IN(s)`, `IN(src; s)`                             | Membership test                           |
-| `ltrimstr(s)`                                     | Strip a leading string prefix             |
-| `limit(n; f)`                                     | First `n` outputs of `f`                  |
-| `isempty(f)`                                      | True if `f` produces no output            |
-| `debug`, `debug(msg)`                             | Print to stderr, pass value through       |
-| `infinite`, `isnan`                               | IEEE 754 infinity literal / NaN test      |
-| `gsub(re; s)`, `scan(re)`, `splits(re)`           | Not real yq builtins at any arity (#1436) |
+| Builtin                                              | Description                                   |
+|------------------------------------------------------|-----------------------------------------------|
+| `paths`, `paths(f)`, `leaf_paths`                    | Paths to every node / every leaf node         |
+| `getpath(path)`                                      | Value at a path array                         |
+| `tostream`, `fromstream(f)`, `truncate_stream(f)`    | Streaming event representation                |
+| `IN(s)`, `IN(src; s)`                                | Membership test                               |
+| `ltrimstr(s)`                                        | Strip a leading string prefix                 |
+| `limit(n; f)`                                        | First `n` outputs of `f`                      |
+| `isempty(f)`                                         | True if `f` produces no output                |
+| `debug`, `debug(msg)`                                | Print to stderr, pass value through           |
+| `infinite`, `isnan`                                  | IEEE 754 infinity literal / NaN test          |
+| `gsub(re; s)`, `scan(re)`, `splits(re)`              | Not real yq builtins at any arity (#1436)     |
+| `inside(s)`                                          | Membership test, `contains`'s inverse         |
+| `startswith(s)`, `endswith(s)`                       | String prefix / suffix test                   |
+| `rtrimstr(s)`                                        | Strip a trailing string suffix                |
+| `index(s)`, `rindex(s)`, `indices(s)`                | First / last / all match positions            |
+| `range(n)`, `range(from; to)`, `range(from; to; by)` | Numeric generator                             |
+| `nth(n)`, `nth(n; f)`                                | The `n`th value / output of `f`               |
+| `combinations`, `combinations(n)`                    | Cartesian product of an array of arrays       |
+| `pow(base; exp)`                                     | Exponentiation                                |
+| `strftime(fmt)`, `strptime(fmt)`                     | Broken-down-time formatting / parsing (#1650) |
 
 ```bash
 echo '{}' | succinctly yq 'paths'
--- a/docs/reference/yq-language.md
+++ b/docs/reference/yq-language.md
@@ -386,6 +386,7 @@ surface (#1512):
 | `nth(n)`, `nth(n; f)`                                | The `n`th value / output of `f`               |
 | `combinations`, `combinations(n)`                    | Cartesian product of an array of arrays       |
 | `pow(base; exp)`                                     | Exponentiation                                |
+| `bsearch(target)`                                    | Binary search index in a sorted array         |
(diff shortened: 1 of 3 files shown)


newhoggy:
Multi-agent \`/code-review\` round found two things:

1. **Fixed**: I omitted \`bsearch\` from the docs table update despite gating it in code and testing it — added in a follow-up commit.
2. **Filed as #1714, not fixed here**: multiple agents independently live-verified against the pinned yq v4.53.3 binary that ~18 more jq-only builtins (\`sqrt\`, \`floor\`, \`ceil\`, \`round\`, \`fabs\`, the \`log\`/\`exp\` family, \`sinh\`/\`cosh\`/\`tanh\`, \`atan2\`, \`add\`, \`implode\`, \`walk\`, \`min_by\`/\`max_by\`, \`INDEX\`) are also lexer-rejected by real yq but carry no gate. This is the same bug class as #1650, just outside the specific 14-name list this issue scoped (sourced from \`docs/compliance/yq/limitations.md\`'s fan-out table, which was never meant to be exhaustive). #1714 also suggests a more durable structural fix (deriving the gated set from yq's actual grammar rather than continued per-name whack-a-mole) rather than mechanically extending the list a fifth time.

All review findings on the actual diff (gate placement, arity coverage, test threading, no removed assertions) came back clean across every other angle.

Issue #1612, opened 2026-08-25
jq/yq: "s" * n panics with capacity overflow (or aborts) on a huge repeat count


newhoggy:
**Severity: Medium** — a Rust panic/abort reachable from ordinary CLI input, with no
malformed document and no unusual flag. The process dies with a `RUST_BACKTRACE` note
instead of a jq-style error, so it is the same class as #1098 (a panic reasoned to be
unreachable that turned out live-reachable via the CLI) rather than a cosmetic divergence.

Found while A/B-testing #1481's binary-fanout changes across arithmetic operand shapes.
**It is not caused by #1481** — `main` @ `d61ef1754` and the #1481 branch abort identically.

## Repro

Verified against pinned jq 1.7.1 (`/usr/bin/jq`, `jq-1.7.1-apple`) and a `--release
--features cli` build.

```
$ echo '{"s":"ab","n":123456789012345678901234567890}' > /tmp/p.json

$ succinctly jq -c '.s * .n' /tmp/p.json
thread 'main' panicked at library/alloc/src/raw_vec/mod.rs:28:5:
capacity overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
                                                              # exit 101

$ jq -c '.s * .n' /tmp/p.json
                                                              # grinds; no panic
```

A slightly smaller count aborts on allocation failure rather than capacity overflow —
same outcome, different message:

```
$ echo '{"s":"ab","n":1e15}' | succinctly jq -c '.s * .n'
memory allocation of 2000000000000000 bytes failed          # exit 134 (SIGABRT)
```

Measured thresholds with `"ab"` (len 2), `--release`:

| `n`      | succinctly                                  | jq 1.7.1            |
|----------|---------------------------------------------|---------------------|
| `1e9`    | ok (2 GB string)                            | ok                  |
| `1e12`   | still running at 15s                        | still running       |
| `1e15`   | **abort**, `memory allocation ... failed`   | still running       |
| `1e30`   | **panic**, `capacity overflow`, exit 101    | still running       |

## Cause

`arith_mul`'s string-repetition arm (`src/jq/eval.rs`, the
`(OwnedValue::String(s), _, _, Some(n_repr))` arm) ends in an unguarded

```rust
Ok(OwnedValue::String(s.repeat(n as usize)))
```

`String::repeat` panics with `capacity overflow` when `len * n` overflows `usize`, before
attempting any allocation. The count reaches that range because the `NumberRepr::Float`
[...]

PR #1716, opened 2026-08-27
fix(jq): guard combinations(n)/cartesian_product's unbounded allocations

### LINKED ISSUE #1669: jq: combinations(n)/cartesian_product have the same unguarded allocation-product crash as #1612/#1634
**Severity: Low** (crash, not data corruption — same class as #1612/#1634, at different sites)

## Summary

Found during #1666's review (the #1634 fix). Two more unguarded, generator-controlled allocations in the `combinations`/cartesian-product family reproduce the identical crash class #1612 and #1634 already closed elsewhere: an infallible allocation sized by a product of independently-controlled lengths, handed straight to the allocator instead of going through a `try_reserve`-style guard.

## Site 1: `builtin_combinations_n` — confirmed live abort

`(0..n).map(|_| base_array.clone()).collect()` implicitly pre-allocates capacity for `n` elements via `Vec`'s `TrustedLen` `collect` specialization. `n` comes straight from evaluating `combinations(n)`'s argument and is never passed through `try_reserve_product` or any other fallible-reserve guard.

```console
$ echo '[1]' | succinctly jq 'combinations(288230376151711744)'
memory allocation of 6917529027641081856 bytes failed
zsh: abort      echo '[1]' |
```

Reproduced live in a debug build of the `issue-1634-capacity-multiplication-overflow` worktree — SIGABRT, exit code 134, uncatchable by `?` or any `EvalError` handling.

## Site 2: `cartesian_product` — same shape, slower to trigger

`cartesian_product` (backing plain `combinations` on `[[...],[...]]`) grows its `results: Vec<OwnedValue>` via `Vec::new()` + `.push()` in a loop bounded by the product of every inner array's length — the same key-outer/target-inner cross-product shape #1634 guarded for `.[$keys]`/`.[$s:$e]`, but with no product-overflow check or capacity pre-check anywhere in the function.

```
[[range(100000)],[range(100000)]] | combinations | length
```
drives `cartesian_product` to push ~10^10 freshly-allocated `OwnedValue::Array` combinations one at a time via ordinary `Vec` growth. Slower to trigger than Site 1 (growth is incremental, not a single huge `with_capacity`), but structurally the identical missed guard — it eventually hits the same infallible-allocator abort path once memory is exhausted, uncaught by any `EvalError`.

## Suggested fix direction

Same technique #1612/#1634 already established: guard `builtin_combinations_n`'s `n`-sized allocation with `try_reserve_product` (or a similar single-factor variant); consider whether `cartesian_product`'s incremental-growth case needs a pre-check on the product of input lengths before the loop starts, or whether ordinary `Vec` growth (which already uses fallible-alloc-aware `try_reserve` internally... to confirm) is enough once the upfront product is validated. Verify against both pinned oracles first (does jq 1.7.1 have any cap for `combinations`? does yq?) per ADR-0018, the same way #1612's own investigation found real yq has a deliberate cap where jq has none.

## Related

- #1612 — the string-repeat instance of this bug class, in `arith_mul`.
- #1634 (closed by #1666) — the computed-index/-slice instance of this bug class, at 7 call sites across `eval.rs`/`eval_generic.rs`.
- #1017 — `cannot_grow_array`/`pad_with_nulls`, the original instance and established fix pattern (`try_reserve`) this class follows.

newhoggy:
## Summary

- `builtin_combinations_n`'s `(0..n).map(|_| base_array.clone()).collect()` handed a generator-controlled `n` straight to `Vec`'s `TrustedLen` `collect` specialization with no guard — confirmed live before this fix, `[1] | combinations(288230376151711744)` aborted the process (SIGABRT, exit 134), uncatchable by any `EvalError` handling.
- `cartesian_product` (backing both bare `combinations` and `combinations(n)`) grew its `results` via ordinary `Vec` push-growth with no upfront size check at all, bounded only by the product of every input array's length — the same missed guard, just slower to trigger since growth is incremental.
- Both are independent risks: a small `n` (cheap on its own) can still combine with a large base array to overflow only the second guard, so both needed their own `try_reserve_product` call — the same technique #1612/#1634/#1017 already established for this crash class.

## Test plan

- [x] Verified the issue's exact repro no longer aborts: `[1] | combinations(288230376151711744)` now returns a clean `EvalError` (exit 5) instead of SIGABRT
- [x] Verified the second, independent overflow path (`[range(1000)] | combinations(10)`, where `n=10` is trivial but `1000^10` isn't)
- [x] Verified bare `combinations`'s own `cartesian_product` overflow path (`[range(100000)] as $a | [$a,$a,$a,$a,$a] | combinations`)
- [x] Cross-checked normal (non-crashing) `combinations`/`combinations(n)` output against real jq — unchanged
- [x] Checked yq mode (now gated behind `--jq-extensions` since #1650, merged just before this) — unaffected, same fix applies
- [x] Added 3 CLI-level regression tests (subprocess-based, since an in-process test can't survive the pre-fix SIGABRT) mirroring #1612's own `does_not_panic` test convention
- [x] `cargo test --features cli,simd,regex,serde` (full suite) — 55/55 green
- [x] `cargo llvm-cov` — every new branch covered (verified the third test was needed to close a real gap: `builtin_combinations_n`'s second, less obvious guard call site)

Fixes #1669

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -703,27 +703,40 @@ pursued within this issue's scope. Either way, the guard converts a host-process
 a catchable error uniformly in both modes, with no cap-specific divergence to record in
 [yq Limitations](../yq/limitations.md).
 
-`combinations`/`combinations(n)` (#1669) have the identical shape a third time, at two
-independent sites: `builtin_combinations_n`'s own `n`-sized bookkeeping (an `n`-element
-`indices` array, guarded on `n` alone) and the actual combinatorial output size
-(`base_array.len().pow(n)` for `combinations(n)`, or the product of every input array's
-length for bare `combinations` on an array of arrays) — two genuinely different quantities,
-since a small `n` can still combine with a large base array to overflow only the second.
+`combinations`/`combinations(n)` (#1669) have the identical shape a third time, at three
+independent sites: `builtin_combinations_n`'s own `n`-sized bookkeeping (both the `indices`
+array and, independently, each combination row it builds — `checked_combinations_len`'s
+`base <= 1` short-circuit makes the output *count* permanently `1` regardless of `n`, so
+that alone never bounds the row width), `cartesian_product`'s own row width
+(`arrays.len()`, unbounded by its length-*product* guard whenever many factor arrays are
+length 1), and the actual combinatorial output count (`base_array.len().pow(n)` for
+`combinations(n)`, or the product of every input array's length for bare `combinations`).
 Confirmed live before this fix: `[1] | combinations(288230376151711744)` aborted the
-process (SIGABRT), not a catchable error. succinctly now refuses with `Cannot allocate
-<base>^<n> elements for combinations` (the `combinations(n)` case) or the same `Cannot
-allocate <factors joined by " * "> elements for a computed-index expansion` message
-`try_reserve_product` already uses elsewhere (the bare `combinations` case), via the same
-`Vec::try_reserve_exact` technique as the cases above. Confirmed live against the pinned jq
-1.7.1 binary that jq itself does not error or crash on a comparably large query: both
-`combinations`/`combinations(n)` are true generators there, so `first(combinations(...))`
-returns immediately without ever materializing the full combinatorial explosion — the
-divergence is specifically that succinctly eagerly builds the whole result set rather than
-streaming it lazily, the same shape (not just the same *conclusion*) as the `.[$keys]`
-cross-product case above. Every combination count that fits in memory is still produced in
-full, so ordinary uses of both builtins are unaffected. This guard is symmetric across jq
-and yq mode (yq reaches it only behind `--jq-extensions`, per #1650), with no additional
-yq-specific cap to record in [yq Limitations](../yq/limitations.md).
+process (SIGABRT), not a catchable error. succinctly now refuses with a `Cannot allocate
+...` message via the same `Vec::try_reserve_exact` technique as the cases above, whichever
+site first detects the excess.
+
+The two arities diverge from real jq differently. Bare `combinations` matches the
+`.[$keys]` cross-product case above exactly: confirmed live against the pinned jq 1.7.1
+binary, `[range(100000)] as $a | first([$a,$a,$a,$a,$a] | combinations)` returns instantly
+rather than erroring or hanging — jq streams results one at a time instead of
+pre-allocating, so succinctly's eager refusal is a real behavioural divergence for this
+arity, not just a difference in wording. `combinations(n)` does not get the same lazy
+treatment in jq's own standard-library definition (`def combinations(n): . as $dot |
+[range(n)] | map($dot) | combinations;`): `[range(n)]` eagerly materializes an
+`n`-element array *before* the lazy recursive `combinations` is ever reached, so jq itself
+pays the same eager cost succinctly does — confirmed live, `[1] | first(combinations(
+288230376151711744))` does not return within 6 seconds against the pinned jq 1.7.1 binary
+either, rather than the instant response the bare-`combinations` case gets. succinctly's
+guard makes this fail fast with a catchable error instead of hanging (and eventually
+exhausting memory) the way jq's own definition does — an improvement in kind, not just a
+faster failure, but still the "would take the host process down" exception ADR-0018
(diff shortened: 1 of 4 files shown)


newhoggy:
Multi-agent \`/code-review\` (7 independent angles) converged on the same critical finding: the first commit's \`try_reserve_product(&[n])\` guard only bounded the outer \`Vec<Vec<OwnedValue>>\`'s pointer-table cost, not the aggregate payload of the \`n\` real \`Vec::clone()\` calls that followed — a moderate \`n\` combined with a large base array (\`[range(100000)] | combinations(1000000)\`) passed that guard trivially, then reached the allocator directly inside the unguarded clone loop. Same abort class the fix was meant to close, just moved to a different axis, and untested by either original regression test.

Fixed properly in a second commit:
- Redesigned \`builtin_combinations_n\` to index into \`base_array\` directly instead of cloning it \`n\` times at all — eliminates both the wasted intermediate allocation and a redundant per-call \`lens\`-vec construction a reviewer separately flagged as pure waste (all \`n\` entries were identical).
- Two independent guards now cover the two independent quantities: \`n\` itself (the \`indices\` Vec's size) and \`base_array.len().pow(n)\` (the actual output size) — each gets its own dedicated error message instead of borrowing the cross-product-shaped wording that didn't fit this call site.
- Added 4 more regression tests (the newly-found gap itself, a deterministic \`try_reserve_exact\`-past-\`checked_pow\` boundary case, the \`base<=1\` short-circuit, and a yq-mode check — the fix is mode-symmetric but nothing had confirmed that directly), plus fixed 2 of the original 3 tests using \`run_jq_stdin\` (via \`cargo run\`, invisible to coverage instrumentation) instead of \`run_jq_full\`.
- Closed a fully pre-existing gap noticed while touching \`cartesian_product\`: bare \`combinations\`' actual multi-output success path had zero permanent test coverage anywhere in this repo before this PR.
- Recorded the divergence in \`docs/compliance/jq/limitations.md\` per ADR-0018, which the first commit missed.

All other review angles (dispatch correctness across jq/yq/eval_generic.rs, error propagation, CLAUDE.md conventions) came back clean.

newhoggy:
A second, deeper \`/code-review\` round (given the second commit was itself a substantial redesign) found the fix was *still* incomplete: \`n\` and \`base_array.len().pow(n)\` were both guarded, but every combination row is \`n\` elements of \`OwnedValue\` — a larger type than the \`Vec<usize>\` \`indices\` guard actually bounded — and \`checked_combinations_len\`'s \`base<=1\` short-circuit makes the output *count* permanently 1 regardless of \`n\`, so neither guard bounded the row width in that case. Fixed in a third commit:

- Added a dedicated row-width guard (\`Vec::<OwnedValue>::new().try_reserve_exact(n)\`), ordered before the real allocations it protects, with its own distinct error message.
- Since \`OwnedValue\` is provably larger than \`usize\`, this same probe subsumes \`indices\`' own allocation — removed the now-dead separate \`indices\` guard.
- \`cartesian_product\` (bare \`combinations\`) had the identical gap on \`arrays.len()\` itself; fixed the same way.
- Corrected a false claim the previous commit introduced into the compliance doc: real jq's \`combinations(n)\` is **not** lazy the way bare \`combinations\` is (confirmed live: it hangs on the same huge-\`n\` input rather than returning immediately) — the doc now describes both arities accurately.

Filed two more follow-ups rather than expanding scope further:
- **#1720**: a separate, pre-existing bug the review surfaced in code this PR reads but doesn't modify — the arity-summing loop above the guards uses unchecked \`usize\` addition, panicking in debug builds and silently wrapping to a wrong answer in release builds.
- **#1721**: the guard-then-reserve pattern is now a 4th independently-hand-rolled copy in this file (#1612/#1017/#1634/#1669), and the two mixed-radix loops (combinations_n's and cartesian_product's) duplicate each other — both DRY/maintainability concerns, not functional bugs.

One line's failure branch (\`cartesian_product\`'s row-width guard, specifically) is genuinely impractical to exercise with an automated test — unlike the \`n\`-exponent case, there's no cheap textual way to encode an astronomically large \`arrays.len()\` (no exponentiation trick applies to a plain count, and the existing \`MAX_RANGE=100000\` cap on \`range()\` prevents generator-based construction at the needed scale). Noting this rather than forcing a multi-GB test.

Issue #1651, opened 2026-08-25
perf(jq): delete_at_path still O(path length) per resolved path, O(d²) total for del(..) — #701's fix didn't move the exponent


newhoggy:
## Where

`delete_at_path` (`src/jq/eval.rs`) applies one resolved static path at a time: `del(..)`/`del(recurse(...))` resolve to `d + 1` paths on a depth-`d` document (one per visited node), and each is applied via its own independent root-to-leaf walk. Each walk is `O(path length)`; summed over all `d + 1` paths, that's `O(d²)` — independent of, and in addition to, the path-*construction* cost #701 fixed.

## How this was found

#701 (closed, fixed by #1631) replaced `PathBranch.path`'s `Vec<Expr>` with a persistent `Rc<PathPrefix>` cons-list, making path *construction* O(1) per node instead of O(depth) per node. That PR's own description deferred measuring `delete_at_path`'s cost, since #701's "Why this is a separate issue" section already flagged it as possibly out of scope.

Post-merge, I ran `benches/jq_recurse_clone_bench.rs`'s existing `del(..)` benchmark on the project's dedicated idle ARM bench box (`johns-mac-mini`, Apple M4 Pro), interleaved before (merge-base `2467ca6c8`) vs after (#1631's tip `e444de0c3`), 3 rounds:

| depth | before (avg of 3) | after (avg of 3) | speedup |
|-------|-------------------:|-------------------:|--------:|
| 100 | 328.5 µs | 241.6 µs | 1.36x |
| 200 | 1235.5 µs | 884.4 µs | 1.40x |
| 300 | 2744.7 µs | 1998.9 µs | 1.37x |

A real, consistent ~27-28% speedup (round-to-round variance <1%) — confirming #1631's fix works. But fitting time ∝ depthᵏ across 100→200→300 gives **k≈1.94 before and k≈1.94 after** — unchanged. `del(..)` end-to-end is still quadratic in depth even after #701's fix, because `delete_at_path`'s own O(d²) term was never touched and dominates the total.

Full methodology and raw numbers: [PR #1631 comment](https://github.com/rust-works/succinctly/pull/1631#issuecomment-5418419114), [issue #701 comment](https://github.com/rust-works/succinctly/issues/701#issuecomment-5418420897).

## Why this might not be fixable without a larger change

[...]

PR #1725, opened 2026-08-27
fix(yq): make bare del(.)/del(.?) match real yq's root-delete rule

### LINKED ISSUE #1702: yq: bare del(.) emits null instead of nothing, and del(.?) deletes instead of no-op'ing
Found while investigating PR #1689's coverage-gap comment (chasing why `flatten_delete_path`'s `Expr::Identity` arm lost its only live-reachable route through the CLI). Filing separately per that PR — this divergence predates and is unaffected by #1651's short-circuit.

## Summary

Real yq treats a `del()` that resolves to the document root itself specially, and differently depending on whether it's wrapped in `?`:

- A bare `del(.)` (no `?` anywhere) deletes the whole document and emits **nothing** — not even `null`.
- `del(.?)` (the path's own `.` marked optional) is a **full no-op** — the original value passes through unchanged.

`succinctly yq` instead follows jq's model for both: `delete_at_path`'s `Expr::Identity` arm (`src/jq/eval.rs`) unconditionally sets the root to `null` and returns it, regardless of `yq_mode` or the `optional` flag — matching real jq's own `del(.)` => `null`, which is correct for `succinctly jq`, but not for `succinctly yq`.

```
$ echo 'a: 1' | yq 'del(.)'
$                                    # <- nothing at all
$ echo 'a: 1' | succinctly yq 'del(.)'
null

$ echo '[1,2,3]' | yq -o=json 'del(.)'
$                                    # <- nothing at all
$ echo '5' | yq -o=json 'del(.)'
$                                    # <- nothing at all

$ echo 'a: 1' | yq 'del(.?)'
a: 1                                 # <- unchanged, full no-op
$ echo 'a: 1' | succinctly yq 'del(.?)'
null

$ echo '[1,2,3]' | yq -o=json 'del(.?)'
[
  1,
  2,
  3
]
$ echo '5' | yq -o=json 'del(.?)'
5
```

Uniform across scalar/array/object targets in both cases — not type-dependent.

For contrast, `succinctly jq`/real jq agree with each other and are unaffected: both `del(.)` and `del(.?)` produce `null` on any target in jq mode.

## Root cause

`delete_at_path`'s `Expr::Identity` arm:

```rust
Expr::Identity => {
    // del(.) replaces with null
    *root = OwnedValue::Null;
    Ok(())
}
```

always overwrites to `null`, with no `yq_mode`/`optional` awareness, even though the function already threads both through for other arms. Two behaviors need adding for `yq_mode`:

- Plain `Expr::Identity` (not wrapped in `Optional`): needs to signal "no output" rather than write `null` — `delete_at_path`'s `Result<(), EvalError>` shape has no way to express that today, so this likely needs to bubble up as something `builtin_del`'s caller can turn into an empty `QueryResult` (its own `optional` parameter already does something adjacent for a different reason — worth checking whether that machinery composes, or whether this needs its own signal).
- `Expr::Optional(Identity)` (`del(.?)`): needs to no-op instead of deleting — the `Optional` handling that unwraps to `Identity` with `optional=true` needs a yq-specific carve-out at the root case specifically, since every non-root field/index/slice arm's existing `optional` handling means something different (tolerate a navigation error) than "leave the value untouched."

## Scope note

Filing to make this trackable, not blocking anything. Likely belongs in the same family as #1411 (`del()`'s yq-mode scalar/root behavior not yet taught the same yq-specific carve-outs `=`/`|=` have) — whoever picks this up should check whether it's cleaner to fold into that fix or keep separate, since the mechanism (signaling "no output" vs "no-op" from deep in `delete_at_path`) is a different shape of problem than #1411's "no-op instead of type error."

newhoggy:
## Summary
Real yq treats `del()` on the document root specially, and differently depending on
whether it's wrapped in `?`:
- A bare `del(.)` (no `?` anywhere) deletes the whole document and emits **nothing** —
  not even `null`.
- `del(.?)` (the path's own `.` marked optional) is a **full no-op** — the original
  value passes through unchanged.

`succinctly yq` previously followed jq's own model for both (`del(.)`/`del(.?)` =>
`null`), which is correct for `succinctly jq` but not `succinctly yq`.

`builtin_del`'s per-path loop now checks, in yq mode, whether the resolved path is a
bare root `Identity` (returns `QueryResult::None` directly — `delete_at_path`'s
`Result<(), EvalError>` has no way to signal "no output") or an `Optional`-wrapped
root `Identity` (skips the delete, leaving the accumulated result untouched).
`#1651`'s existing short-circuit already collapses a comma naming the root alongside
another path (`del(., .c)`) to a single root `Identity` before this check runs, so
that shape composes for free.

Fixes #1702

## Test plan
- [x] Live-verified all repro cases from the issue against yq v4.53.3 (scalar, array, object targets)
- [x] Live-verified `del(., .c)` comma-with-root composes correctly with #1651's short-circuit
- [x] Confirmed `succinctly jq` mode is unaffected — `del(.)`/`del(.?)` still produce `null`
- [x] Added regression tests in `tests/yq_cli_tests.rs` (bare root, optional root, comma-with-root) and `tests/jq_cli_tests.rs` (jq-mode-unaffected)
- [x] `cargo test --features cli,regex` — full suite green (2895+ tests, 0 failures)
- [x] `cargo build --no-default-features` — no_std build clean
- [x] Patch coverage: 100% (7/7 new lines)

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -27724,6 +27724,25 @@ fn builtin_del<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
                 YqDelSliceOutcome::Noop => continue,
                 YqDelSliceOutcome::NotApplicable => path,
             };
+            // yq's root-delete rule (#1702): real yq deletes the whole
+            // document and emits nothing for a bare `del(.)` (no `?`
+            // anywhere), and treats `del(.?)` as a full no-op — the
+            // optional root delete never fires. `succinctly` previously
+            // followed jq's model for both (`del(.)` => `null`,
+            // `del(.?)` => `null`), which is correct for `succinctly jq`
+            // but not `succinctly yq`. Checked here, ahead of
+            // `delete_at_path`, rather than inside its `Expr::Identity`
+            // arm, since that arm has no way to signal "no output at
+            // all" through its `Result<(), EvalError>` return type.
+            if S::TAG == EvalTag::Yq {
+                match effective_path {
+                    Expr::Identity => return QueryResult::None,
+                    Expr::Optional(inner) if matches!(inner.as_ref(), Expr::Identity) => {
+                        continue;
+                    }
+                    _ => {}
+                }
+            }
             if let Err(e) =
                 delete_at_path(&mut result, effective_path, false, S::TAG == EvalTag::Yq)
             {
--- a/tests/jq_cli_tests.rs
+++ b/tests/jq_cli_tests.rs
@@ -22933,3 +22933,16 @@ fn test_parent_n_non_numeric_argument_reports_real_type_name_1487() -> Result<()
     );
     Ok(())
 }
+
+/// #1702 fixed `del(.)`/`del(.?)` to emit nothing / no-op in yq mode; confirm
+/// `succinctly jq` still follows real jq's own model unchanged — both
+/// produce `null` on any target.
+#[test]
+fn test_jq_mode_root_del_still_null_after_1702() -> Result<()> {
+    for expr in ["del(.)", "del(.?)"] {
+        let (stdout, stderr, code) = run_jq_full(&["-c", expr], Some(r#"{"a":1}"#))?;
+        assert_eq!(stdout.trim(), "null", "expr={expr} stderr: {stderr:?}");
+        assert_eq!(code, 0, "expr={expr} stderr: {stderr:?}");
+    }
+    Ok(())
+}
--- a/tests/yq_cli_tests.rs
+++ b/tests/yq_cli_tests.rs
@@ -18439,6 +18439,46 @@ fn test_1162_bare_root_object_slice_del_is_noop() -> Result<()> {
     Ok(())
 }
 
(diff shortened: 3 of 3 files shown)


newhoggy:
\`/code-review\` caught a real, adjacent gap: a comma combining \`.?\` with a sibling path (\`del(.?, .c)\`) still produces the old wrong \`null\` output, because \`resolve_node\`'s shared \`Expr::Optional\` arm gives a zero-component inner branch an artificial \`path.depth() == 1\` (wrapping it in \`Expr::Optional(Pipe(vec![]))\`), which slips past #1651's \`short_circuit_del_root\` depth-0 check — and \`PathBranch\` has no \`optional\` field to catch it another way. This pre-dates this PR (confirmed same wrong output on \`main\` before this fix) and requires touching \`resolve_node\`'s shared branch-construction logic, used by \`path()\`/\`=\`/\`|=\` as well as \`del()\` — filed as #1726 rather than expanding this PR's scope beyond #1702's literal (single-path) repro.

Issue #1620, opened 2026-08-25
jq: decide whether a decode failure should be catchable — '?' silently swallows it, try/catch does not (#1247 Open Risk 1)


newhoggy:
## Summary

#1247's design doc raises this as Open Risk 1 and leaves it open:

> **`GenericResult::Error` is catchable.** A decode failure routed through it becomes
> suppressible by `.a?` or `try .a catch`, where jq's parse error is not catchable at all
> (it happens before the program runs). Decide explicitly whether that is acceptable or
> whether these need `Halt`-like uncatchability. This is a genuine semantic difference from
> both oracles and the design above does not resolve it.

Its own "Follow-up issues" list says to file this "if it is not settled during Stage 3's
review". Stages 1–5 have landed and it was not settled. Filing it.

## What the shipped behaviour actually is

Measured at `4864f5f80` (`--release --features cli`). The two suppression mechanisms
disagree with each other:

```console
$ printf '{"a":"\ud800"}' | succinctly jq -c 'try (.a | ascii_downcase) catch "caught"'
jq: error (at <stdin>:0): invalid unicode escape sequence            # exit 5 — NOT caught

$ printf '{"a":"\ud800"}' | succinctly jq -c '.a | length?'
                                                                     # exit 0, no output
```

Ordinary evaluation errors are caught by both, and match real jq:

```console
$ echo '{"a":1}' | succinctly jq -c 'try (.a | keys) catch "caught"'
"caught"                                                             # exit 0
$ echo '{"a":1}' | jq -c 'try (.a | keys) catch "caught"'
"caught"                                                             # exit 0
```

So a decode failure is currently **uncatchable by `try`/`catch`** — arguably the right
answer, since real jq's equivalent is a parse error that no program can catch — but
**silently swallowed by `?`**, which is the exact silent-degrade #1247 exists to eliminate,
reachable by a filter as ordinary as `.a | tonumber?`.

## Why `try`/`catch` doesn't catch it

Not by design. `Expr::Try` has no native arm in `eval_generic`, so it falls through to the
`_ =>` reindex-bridge fallback, which materializes the input via `to_owned_with_cursor`
*before* the try body ever runs. The decode failure happens while preparing the try's
input, so it escapes the try. Change how `Expr::Try` dispatches and the behaviour flips
with nothing pinning it.

## Why `?` swallows it

Every builtin's dispatch has the shape

```rust
} else if optional {
    GenericResult::None
} else {
[...]

PR #1745, opened 2026-08-27
refactor(jq): consolidate decode-failure/uncatchable checks

### LINKED ISSUE #1661: jq: consolidate the 'always uncatchable' error checks (is_decode_failure / is_invalid_path_expression) — duplicated helper, ANDed flags
## Summary

Two small maintainability regressions surfaced during review of PR #1647
(#1620's decode-failure-catchability fix). Neither is a correctness bug —
both are follow-ups on the shape of the fix, filed separately from #1660
since that one is behavioral and these are structural.

## 1. A shared helper was deleted and replaced with 7 verbatim-duplicated inline copies

`eval_generic.rs` used to have one helper, `type_error_or_decode_failure`,
called from a single site to decide "is this a decode failure or a genuine
type mismatch". PR #1647 deleted it and inlined the equivalent
`else if let Some(reason) = value.string_decode_error() { GenericResult::Error(EvalError::decode_failure(reason)) }`
branch by hand at 7 separate `eval_builtin` arms (`Iterate` ×2, `Length`,
`Keys`, `KeysUnsorted`, `ToEntries`, `ToNumber`).

The actual bug being fixed was just ordering — check decode-failure before
`optional`, not after — which the old helper could have absorbed with a single
added `optional: bool` parameter (a one-site edit). As shipped, a future
change to this rule, or a new dispatch arm needing the same check, has to be
applied by hand at 7 sites with no compiler check that all 7 stay consistent.
This is a step backward in centralization introduced by this diff, not
inherited convention — the single call site existed before this PR.

**Suggested fix**: restore a shared helper (e.g.
`fn decode_failure_or<T>(value: &V, optional: bool, fallback: impl FnOnce() -> GenericResult<V>) -> Option<GenericResult<V>>`)
that the 7 sites can call instead of duplicating the branch.

## 2. `resolve_node` ANDs two semantically-identical "always uncatchable" flags

`resolve_node`'s catch arm now reads:

```rust
Err((prefix, EvalEscape::Error(e))) if !e.is_invalid_path_expression() && !e.is_decode_failure()
```

Both flags mean the same thing at this call site — "this error class is
always uncatchable, no positional nuance" — unlike the file's third such
predicate, `is_untracked_navigation_error`, which genuinely *is*
position-dependent and can't be merged the same way. Since these two really
are equivalent in kind, a third "always uncatchable" error class added later
requires hand-editing this (and every other) call site again, with nothing
enforcing that all of them get updated.

**Suggested fix**: add a single `EvalError::is_uncatchable(&self) -> bool`
that ORs `is_invalid_path_expression()` and `is_decode_failure()` (and any
future always-uncatchable class), and collapse this call site to
`!e.is_uncatchable()`.

## Context

- #1620 / PR #1647 — the fix these follow-ups are on.

newhoggy:
## Summary

Two maintainability regressions from PR #1647 (#1620's decode-failure ordering fix),
filed as #1661 during that PR's own review. Neither is a correctness bug -- both are
follow-ups on the shape of that fix.

### 1. A shared helper was deleted in favor of 7 hand-duplicated inline copies

`eval_generic.rs` used to have one helper deciding "is this a decode failure or a
genuine type mismatch". PR #1647 deleted it and inlined the equivalent branch by hand
at 7 separate `eval_builtin` arms (`Expr::Iterate`, `Builtin::Map`, `Length`, `Keys`,
`KeysUnsorted`, `ToEntries`, `ToNumber`). A future change to this rule, or a new
dispatch arm needing the same check, would have to be applied by hand at 7 sites with
no compiler check that all 7 stay consistent.

Restored as `decode_failure_or(value, optional, fallback)`, used at all 7 sites. Each
site's own builtin-specific error (`cannot_iterate_with`, `has_no_length`,
`has_no_keys`, `cannot_parse_as_number`) moves into the `fallback` closure, called only
on the genuine type-error exit. `ToEntries` (which never gates on `optional` -- see its
own preserved comment on why) passes `optional: false` literally, keeping identical
behavior.

### 2. `resolve_node`'s catch arm ANDs two semantically-identical "always uncatchable" flags

```rust
Err((prefix, EvalEscape::Error(e)))
    if !e.is_invalid_path_expression() && !e.is_decode_failure() =>
```

Both flags mean the same thing at this call site -- "this error class is always
uncatchable, no positional nuance" -- unlike the file's third such predicate,
`is_untracked_navigation_error`, which genuinely is position-dependent (a bare postfix
`?` on the primitive that raised it) and is left untouched at its own two call sites.

Added `EvalError::is_uncatchable()` (`error.rs`), ORing the two, and collapsed the
`eval.rs` call site to `!e.is_uncatchable()`.

## Test plan

- [x] Pure refactor, no behavior change intended -- full `cargo test` unchanged
      (2926+ lib tests, 1015 jq_cli_tests, 204+64 eval-generic tests, all passing;
      one transient failure on the first full run under concurrent-session
      contention, confirmed spurious by two clean re-runs)
- [x] Grepped for other ANDed occurrences of the same two flags -- none found

Fixes #1661

### DIFF
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -1014,9 +1014,10 @@ impl EvalError {
     /// *where* the error was caught (a bare postfix `?` on the primitive
     /// that raised it vs. anything else), both of these mean the same thing
     /// at every call site that checks them: never suppressed by `?`, never
-    /// handed to a `catch` handler. A future always-uncatchable error class
-    /// only needs to be added here, not hand-copied into every `?`/`try`
-    /// dispatch that currently ANDs the two checks together.
+    /// handed to a `catch` handler. Only `resolve_node`'s `Expr::Try` arm
+    /// ANDs the two checks together today, but a future always-uncatchable
+    /// error class only needs to be added here, not hand-copied into
+    /// whichever `?`/`try` dispatch sites grow the same check next.
     pub fn is_uncatchable(&self) -> bool {
         self.is_invalid_path_expression() || self.is_decode_failure()
     }
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -1007,6 +1007,20 @@ impl EvalError {
         )
     }
 
+    /// Whether this error class is *always* uncatchable, with no positional
+    /// nuance — [`Self::is_invalid_path_expression`] or
+    /// [`Self::is_decode_failure`]. Unlike
+    /// [`Self::is_untracked_navigation_error`], which genuinely depends on
+    /// *where* the error was caught (a bare postfix `?` on the primitive
+    /// that raised it vs. anything else), both of these mean the same thing
+    /// at every call site that checks them: never suppressed by `?`, never
+    /// handed to a `catch` handler. A future always-uncatchable error class
+    /// only needs to be added here, not hand-copied into every `?`/`try`
+    /// dispatch that currently ANDs the two checks together.
+    pub fn is_uncatchable(&self) -> bool {
+        self.is_invalid_path_expression() || self.is_decode_failure()
+    }
+
     /// `Cannot check whether <container> has a <key type> key`.
     pub fn cannot_check_has(container_type: &str, key_type: &str) -> Self {
         Self::new(format!(
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -17959,10 +17959,10 @@ fn resolve_node<'a, S: EvalSemantics>(
             // only a genuine error or a break (below) reaches `catch`;
             // everything else propagates unchanged, as does an
             // invalid-path-expression complaint (#530) and, per the same
-            // reasoning, a decode failure (#1247, #1620).
-            Err((prefix, EvalEscape::Error(e)))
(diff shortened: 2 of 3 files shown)


newhoggy:
Multi-angle \`/code-review\` (10 finder angles + verification) completed. Fixed two in-scope doc-comment gaps it found (\`decode_failure_or\`'s enumeration omitted \`ToEntries\`; \`is_uncatchable\`'s comment overstated how many call sites currently AND the two checks) in a follow-up commit.

The review also surfaced a significant, high-confidence, but **pre-existing** bug family in \`eval.rs\`'s own private/non-generic evaluator (\`to_owned\`/\`builtin_keys\`/\`Expr::Iterate\`/\`builtin_map\`/\`builtin_to_entries\`) that never received the #1247/#1620/#1660 decode-failure fix sweep at all -- confirmed live by three independent reviewer angles via the public \`succinctly::jq::eval\` library API. Unrelated to this PR's diff (untouched by it), out of scope for a "pure refactor, no behavior change" dedup PR -- filed as #1746 rather than scope-creeping it in here.

PR #1748, opened 2026-08-27
fix(yq): guard --input-format json bridge against colliding decode-failure keys

### LINKED ISSUE #1738: yq: --input-format json bridge still silently drops a value on colliding decode-failure keys (PR #1672 fix lost before merge)
## Summary

**Severity: Critical (silent data loss, live on `main` today).**

PR #1672 (#1642, "preserve an undecodable object key instead of raising on it") added `DisplayKeyGuard` to guard every `to_owned`-shaped materializer against silently collapsing two colliding decode-failure keys into one — *except* `yq_runner.rs`'s `to_owned_canonicalizing_numbers_at_depth`, the `--input-format json` bridge for `succinctly yq --input-format json --slurp/--eval-all/--inplace`. That one function still builds its `IndexMap` via plain `key_display_string` with no guard, so two colliding decode-failure keys silently overwrite each other there — exit 0, no error, one value quietly lost.

A post-merge review of PR #1672 found and fixed this (plus a related duplication and a minor efficiency issue), but the fix was made directly in a worktree's working tree and **never committed** before that worktree got rebased onto `origin/main` — main already contained PR #1672's original content, so the rebase silently discarded the uncommitted fix. The bug is therefore still live on `main`, unfixed, despite already having been diagnosed and a working patch produced once.

## Repro (confirmed live against current `main`)

```console
$ printf '{"\ud800":1,"\ud800":2}' | succinctly yq --input-format json --slurp '.[0]' -o=json
{"\ud800": 2}
```
Exit 0. The value `1` is silently dropped. The equivalent jq-mode path (`eval_generic::to_owned_at_depth`, which *does* have the guard) correctly raises an "ambiguous" error on the identical shape.

## The fix (already implemented once; reconstructing it is mechanical)

1. In `src/jq/document.rs`: extract the hand-copied `key_display_string_kind` + `DisplayKeyGuard::check` + raise-or-insert sequence (duplicated across `eval_generic.rs`'s `to_owned_at_depth`/`to_owned_cursor_at_depth`/`to_owned_with_comments_at_depth` and `lazy.rs`'s `cursor_to_owned_at_depth`) into one shared function, e.g.:
   ```rust
   pub fn resolve_display_key<V: DocumentValue, T>(
       key: &V,
       map: &IndexMap<String, T>,
       guard: &mut DisplayKeyGuard,
   ) -> Result<Option<String>, EvalError> {
       let Some((key, is_fallback)) = key_display_string_kind(key) else {
           return Ok(None);
       };
       let key = key.into_owned();
       if !guard.check(map, &key, is_fallback) {
           return Err(colliding_display_key_error(&key));
       }
       Ok(Some(key))
   }
   ```
   Make `DisplayKeyGuard` (the struct) `pub` (it was `pub(crate)`) so `succinctly-cli`'s separate `yq_runner.rs` binary crate can construct one — same reasoning `key_display_string` itself already documents for why it's `pub` and not `pub(crate)`.
2. Replace all 4 existing call sites (`eval_generic.rs` x3, `lazy.rs` x1) with calls to `resolve_display_key`, deleting their hand-rolled duplicates of the same sequence.
3. In `src/bin/succinctly/yq_runner.rs`: change `to_owned_canonicalizing_numbers`/`to_owned_canonicalizing_numbers_at_depth` to return `Result<OwnedValue, EvalError>` instead of a bare `OwnedValue`, construct a `DisplayKeyGuard` per object, and call `resolve_display_key` instead of `key_display_string` directly. Propagate `?` through the array/object recursion. Update `parse_input`'s call site to `.map_err(|e| anyhow::anyhow!("{e}"))?`. Update the 4 existing unit tests in this file that call `to_owned_canonicalizing_numbers` directly to `.unwrap()` the new `Result`.
4. (Optional, bundled with the above previously) `DisplayKeyGuard::check` can fast-exit before touching the map at all when `!is_fallback && fallback_keys.is_empty()` — provably no collision is possible in that case, and it's the overwhelmingly common one (zero decode-failure keys per object).
5. Add a regression test alongside the existing `test_input_format_json_bridge_preserves_decode_failure_key_1642` (`tests/yq_cli_tests.rs`), e.g. asserting `{"a\q":1,"a\\q":2}` under `--input-format json --slurp`/`--eval-all` now exits 1 with an "ambiguous" error, instead of silently returning `{"a\q": 2}`.

All of the above was implemented, and verified against the full test suite (`cargo test --features cli,simd,regex,serde`, 2909+ tests), both CI clippy legs (`--all-features` and the SIMD x86 leg), `cargo doc -D warnings`, and `cargo fmt --check` — all clean — before being lost to the uncommitted-changes/rebase accident described above.

## Related

- #1642 / PR #1672 — introduced `DisplayKeyGuard`, missed this one call site.
- #1733 — a separate, still-open design question (YAML complex-key fallback not covered by the guard at all) — not this bug, don't conflate the two.
- #1734 — a separate, still-open design question (`eval.rs`'s second evaluator, currently a dead path) — also not this bug.

newhoggy:
## Summary

- `yq_runner.rs`'s `to_owned_canonicalizing_numbers_at_depth` (the `--input-format json` bridge for `--slurp`/`--eval-all`/`--inplace`) built its `IndexMap` via plain `key_display_string` with no `DisplayKeyGuard`, so two colliding decode-failure keys silently overwrote each other — exit 0, no error, one value quietly lost. PR #1672 added the guard to every other `to_owned`-shaped materializer but missed this one call site.
- Extracted the guard-and-raise sequence (`key_display_string_kind` + `check` + raise-or-insert), previously duplicated four times across `eval_generic.rs` and `lazy.rs`, into one shared `resolve_display_key` helper in `document.rs`. Made `DisplayKeyGuard` `pub` so `yq_runner.rs` (a separate binary crate) can construct one.
- Routed all five call sites — the four existing ones plus the previously-unguarded bridge — through the shared helper.

## Repro (before fix)

```console
$ printf '{"\ud800":1,"\ud800":2}' | succinctly yq --input-format json --slurp '.[0]' -o=json
{"���": 2}
```
Exit 0, value `1` silently dropped.

## After fix

```console
$ printf '{"\ud800":1,"\ud800":2}' | succinctly yq --input-format json --slurp '.[0]' -o=json
Error: object key "���" is ambiguous: an undecodable key's display form collides with another key of the same name and cannot be represented
```
Exit 1, matching the jq-mode path's existing behavior for the identical shape.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 2926+ tests, 0 failed)
- [x] `cargo doc --no-deps --all-features` (no warnings)
- [x] `cargo fmt --check`
- [x] `cargo build --no-default-features` (`no_std` compiles)
- [x] New regression test `test_input_format_json_bridge_raises_on_colliding_decode_failure_key_1738` covering `--slurp` and `--eval-all`
- [x] `omni-dev coverage diff`: 100% patch coverage (32/32 new lines)
- [x] Manually verified fix against the exact repro in #1738, plus a sanity check that ordinary (non-colliding) objects are unaffected

Fixes #1738

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -9,8 +9,8 @@ use std::io::{BufWriter, IsTerminal, Read, Write};
 use std::path::Path;
 
 use succinctly::jq::document::{
-    effective_keys, key_display_string, DocumentCursor, DocumentElements, DocumentFields,
-    DocumentValue, IndentSpec,
+    effective_keys, resolve_display_key, DisplayKeyGuard, DocumentCursor, DocumentElements,
+    DocumentFields, DocumentValue, IndentSpec,
 };
 use succinctly::jq::eval_generic::{
     assert_nesting_depth, eval_with_cursor_using, to_owned as generic_to_owned,
@@ -593,14 +593,22 @@ fn gather_input_sources(
 /// clearly-scoped primitive the library exports for this -- everything
 /// else about the walk stays here, matching where `canonicalize_json_numbers`
 /// always lived before this fix, just fused into one pass instead of two.
-fn to_owned_canonicalizing_numbers<V: DocumentValue>(value: &V) -> OwnedValue {
+///
+/// Returns `Err` when two colliding decode-failure keys (#1642/#1738) would
+/// otherwise silently overwrite one another in the same object -- see
+/// `eval_generic::to_owned_at_depth`'s identical guard, which this function
+/// now shares via [`resolve_display_key`] rather than the hand-rolled
+/// `key_display_string` call it used to make (#1738: that gap meant this,
+/// the `--input-format json` bridge, was the one materializer PR #1672 never
+/// reached).
+fn to_owned_canonicalizing_numbers<V: DocumentValue>(value: &V) -> Result<OwnedValue, EvalError> {
     to_owned_canonicalizing_numbers_at_depth(value, 0)
 }
 
 fn to_owned_canonicalizing_numbers_at_depth<V: DocumentValue>(
     value: &V,
     depth: usize,
-) -> OwnedValue {
+) -> Result<OwnedValue, EvalError> {
     // `assert_nesting_depth` (256, matching `eval_generic::to_owned_at_depth`
     // exactly), not `assert_value_tree_depth` (384) -- this walks a
     // `DocumentCursor`-backed `V` directly, the same recursion shape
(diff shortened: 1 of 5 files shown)


newhoggy:
Code review surfaced a structurally similar but out-of-scope gap in `yaml_to_owned_value`'s Mapping arm (the YAML-native DOM path, not the JSON bridge this PR fixes) — filed as #1749.

Issue #1643, opened 2026-08-25
jq/JSON: object and array delimiters are unvalidated, so a missing `:` or `,` is silently invented


newhoggy:
## Summary

JSON's structural **delimiters** are not validated at all: `,` and `:` are interchangeable to
the semi-index, and either can be repeated or omitted entirely. Eight shapes real jq rejects
outright are accepted at exit 0, one of them silently *inventing* a `:` that was never written.

Found while reviewing PR #1628 (#1194's fix). Distinct from #1194, and not addressed by it —
see "Why #1194's fix cannot see these" below.

## Repro

Verified against the `/usr/bin/jq` 1.7.1 pin. Every `succinctly` row below is exit 0; every
`jq` row is a parse error at exit 5.

```console
$ echo '{"a" 1}'        | sjq -c .    # {"a":1}          <- a missing `:` becomes one
$ echo '{"a":1 "b":2}'  | sjq -c .    # {"a":1,"b":2}    <- a missing `,` becomes one
$ echo '{"a":1,}'       | sjq -c .    # {"a":1}          <- trailing comma
$ echo '{"a":1,,"b":2}' | sjq -c .    # {"a":1,"b":2}    <- doubled comma
$ echo '{,}'            | sjq -c .    # {}
$ echo '[1 2]'          | sjq -c .    # [1,2]            <- missing `,`
$ echo '[1,2,]'         | sjq -c .    # [1,2]            <- trailing comma
$ echo '[,]'            | sjq -c .    # []
```

`{"a" 1}` is the one worth staring at: the output is not merely lenient, it is a *different
document* from the input, and it is well-formed JSON, so nothing downstream can tell it was
ever wrong.

The one shape in this family that already errors is `{"a":,}` (exit 5) — and only because it
lands on #1194's odd-child-count check by accident, not because any delimiter was inspected.

## Why #1194's fix cannot see these

`json::standard::is_delim` maps `:` and `,` to the same `Phi::NONE`, so an object's members are
recovered by pairing the container's BP children two at a time and the delimiters themselves
leave no trace. #1194 closed the two consequences that disturb the *pairing*:

- an **odd** child count (`{invalid}`, `{"a":1,"b"}`) → `DocumentFields::ends_unpaired`
- a child in key position that is not a string (`{123: 1}`) → `key_is_malformed`

Every shape above leaves an **even** child count with **string keys**, so both checks pass and
the document is indistinguishable from a well-formed one. This is not a gap in that fix; it is
[...]

Issue #935, opened 2026-08-15
grinding-backlog: unfiltered 'cargo test' silently hangs for hours under heavy concurrent multi-session load


newhoggy:
**Severity: Medium**

## Summary

While verifying issue #847 (`grinding-backlog` agent \`aiden\`), the mandatory
Phase 2 step 4 full-suite check —
\`cargo test --features cli,simd,regex,serde\` (no target filter, default job
count) run from a fresh worktree
(\`issue-847-flow-style-bare-dash-deferred\`) — silently hung for at least
7+ hours with **zero output** (not even the usual \`Compiling ...\` lines)
and 0.0% CPU, on a machine that was simultaneously running ~38 other
concurrent \`cargo\` processes spawned by other \`grinding-backlog\`
sessions/worktrees sharing the same clone. There was no error, no timeout,
no partial output — the process just sat idle indefinitely, giving no
signal that anything was wrong.

## Repro / observed symptoms

\`\`\`
$ ps -o pid,stat,etime,pcpu,command -p 59139
  PID STAT  ELAPSED  %CPU COMMAND
59139 S    07:18:10   0.0 cargo test --features cli,simd,regex,serde

$ ps aux | grep -c "[c]argo"
38

$ uptime
load averages: 4.33 3.83 4.16   # modest, not pegged

$ ps aux -r | head -15   # top CPU consumers
# no cargo/rustc process appears in the top 15 at all
\`\`\`

Two other unrelated \`cargo test\` invocations from other concurrent
sessions (one in worktree \`issue-840-sub-gsub-replacement-cascading-divergence\`)
showed the identical symptom: \`S\` state, 0.0% CPU, multi-hour elapsed time,
zero progress. This wasn't isolated to my own process.

**Workaround that resolved it**: \`kill -9\` the stuck PID, then re-run the
identical command with a capped job count (\`cargo test --features
cli,simd,regex,serde -j 4\`) in the same worktree. The retry immediately
showed real \`rustc\` CPU usage (18.8%, 1.3%, ...) within 20 seconds and the
full suite (including all doctests) completed normally in well under two
minutes.

## Root cause

Not fully pinned down — filing as observed rather than root-caused, since
digging further would have meant leaving issue #847 in progress
indefinitely (against the skill's "one issue at a time" / no-mid-edit-detour
rule). Best available read of the evidence:

- The hang happened before any \`Compiling\` output was even printed, which
  points at something in cargo's own startup path (build-plan
[...]

Issue #1677, opened 2026-08-26
jq: #1643's delimiter check lives only in the CLI printer — .[]/length/keys/library API still accept a missing delimiter


newhoggy:
## Summary

#1643's delimiter check (`preceding_gap_ok`, added to `print_json`'s object/array arms in
`jq_runner.rs`) lives entirely in the CLI's own printer. It was never added to the actual
`jq`/`eval_generic` **evaluator** that walks objects/arrays to answer `.[]`, `length`, `keys`,
`add`, `to_entries`, or a plain field lookup that doesn't reach a leaf. Any filter whose
result doesn't require *re-serializing the malformed container itself* as JSON output never
passes through `print_json`'s object/array arms at all, so it never triggers the check —
even on the ordinary `sjq -c .` path, no flag required.

#1643's own commit message called this "the same class of `which path validates` gap #1611
already tracks for a different flag-driven divergence" — but #1611 is specifically about
`-a`/`-S`/`-C`/`-s` swapping the *entire input strategy* to a different materializer
(`parse_json_stream` → `json_bytes_to_owned_value`, now itself covered by a #1643 follow-up
fix). This issue is a distinct mechanism: it reproduces on the **default** path, with **no
flags at all**, for filters that simply never call `print_json`'s object/array arms on the
malformed container in the first place. It also isn't #1629 (`keys_unsorted`'s positional
fast paths still answering from a malformed object) — that one is about #1194's non-string-
key class; this one is about #1643's delimiter class, and reproduces on plain `keys`, not
just `keys_unsorted`.

## Repro

Verified against the `/usr/bin/jq` 1.7.1 pin; every `succinctly` row is exit 0 with a
plausible-looking answer, every `jq` row is a parse error at exit 5:

```console
$ echo '{"a" 1, "b": 2}' | sjq -c 'keys'         # ["a","b"]
$ echo '{"a" 1, "b": 2}' | sjq -c 'length'       # 2
$ echo '{"a" 1,"b":2}'   | sjq -c 'to_entries'   # [{"key":"a","value":1},{"key":"b","value":2}]
$ echo '[1 2, 3]'        | sjq -c '.[]'          # 1␊2␊3
$ echo '[1 2, 3]'        | sjq -c 'add'          # 6
$ echo '{"a" 1}'         | sjq -c '.a'           # 1

$ echo '{"a" 1, "b": 2}' | jq -c 'keys'          # parse error, exit 5  (every row above)
```

[...]

PR #1750, opened 2026-08-27
fix(jq): catch a trailing/empty-gap comma next to a JSON bracket (#1676)

### LINKED ISSUE #1676: jq/JSON: #1643's delimiter check still misses a trailing comma next to a bracket (`{"a":1,}`, `{,}`)
## Summary

#1643's delimiter check (`preceding_gap_ok` in `jq_runner.rs`) catches a **missing or
doubled** `,`/`:` between two real object members or array elements, but deliberately does
not catch a **trailing** delimiter next to a closing bracket, or a delimiter inside an
apparently-empty container. Both were called out as a known, deliberate gap in #1643's own
fix and pinned with a regression test (`test_jq_trailing_leading_comma_still_a_known_gap_1643`
in `tests/jq_cli_tests.rs`) specifically so a future fix would update that test rather than
land silently uncovered. This issue is that future fix's tracking ticket.

## Repro

Verified against the `/usr/bin/jq` 1.7.1 pin; every `succinctly` row is exit 0, every `jq`
row is a parse error at exit 5:

```console
$ echo '{"a":1,}' | sjq -c .   # {"a":1}   <- trailing comma
$ echo '[1,2,]'   | sjq -c .   # [1,2]     <- trailing comma
$ echo '{,}'      | sjq -c .   # {}        <- comma in an apparently-empty object
$ echo '[,]'      | sjq -c .   # []        <- comma in an apparently-empty array
```

## Root cause

`preceding_gap_ok` validates the gap **before** a real child by scanning backward from that
child's own `text_position()` — a position every direct child already has cheaply, via
IB's own opening-position marks. A trailing delimiter sits **after** the last real child and
**before the container's own closing bracket**, which has no equivalent cheap lookup: IB only
marks *opening* positions (see `JsonCursor::text_range`'s own comment), so finding a `}`/`]`'s
own text offset needs a bracket-matching scan — exactly the cost #1643 chose not to pay on
every container.

An apparently-empty container (`{,}`, `[,]`) has the same problem from the other direction:
`elements.is_empty()`/the object field walk never iterates a single child to hang a check off
of, so there is nothing for `preceding_gap_ok` to be called against at all.

## Suggested direction

Not attempted here — needs its own investigation, per #1643's own note about the cost
tradeoff. Two shapes seem worth comparing:

1. Give containers a cheap closing-bracket text position (would also help other cursor
   operations that currently lack one — see `JsonCursor::text_range`'s own comment).
2. A narrower, container-local check: since the *closing* bracket byte itself is uniquely
   identifiable (`}`/`]`) once its position is known by any means, and the *opening* bracket's
   own position is already free (it's the container cursor's own `text_position()`), maybe a
   bounded backward scan from the closing bracket — found via BP navigation rather than a
   text-position lookup — is cheaper than a general-purpose closing-position index.

## Scope

Same as #1643: JSON only (`succinctly jq`'s CLI print path). Also affected by #1643's other
documented scope limits — see the sibling issue filed alongside this one for the evaluator-
layer gap (`.[]`/`length`/`keys`/library API still bypass the delimiter check entirely,
independent of this trailing-comma limitation).

Refs #1643.

newhoggy:
## Summary

#1643's own delimiter check (`preceding_gap_ok`) only validated the gap
between two real children, deliberately leaving a `,`/`:` next to a closing
bracket uncaught (`{"a":1,}`, `[1,2,]`) or inside an apparently-empty
container (`{,}`, `[,]`) — both silently accepted at exit 0 where real jq
(pinned 1.7.1) rejects them at exit 5.

- Adds `following_gap_ok`, a forward-scan mirror of the existing backward
  `preceding_gap_ok`, and `trailing_gap_ok`, which applies it once per
  container using the last real child's own end position.
- **Deliberately scoped**: the end position is only used when it's cheap —
  a scalar last child (self-delimiting, bounded scan) or an empty
  container. When the last child is itself a container, the check is
  skipped rather than pay for finding *its* closing bracket's text
  position (the same O(subtree) cost `preceding_gap_ok`'s own doc comment
  explains is too expensive). That narrower residual (`{"a":[1,2],}`,
  `[[1,2],]`) is pinned with its own regression test, same discipline
  #1643 used for this whole class originally.
- Wired into both the default `print_json` path and the `-S`/`-C`/`--slurp`
  fallback's `validate_json_delimiters` — the latter was independently
  broken (confirmed by building pre-fix and observing exit 0), since
  `find_json_values`'s lenient re-scan finds `{"a":1,}` as one balanced
  span regardless of the comma.

Closes #1676.

## Coverage

Patch coverage on `jq_runner.rs`: **94.87%** (111/117 new lines). The
remaining 6 uncovered lines are all bare closing-brace lines immediately
following a `return` inside nested `if`/`if let` blocks — verified via raw
`lcov` `DA:` hit counts that both the success and failure branches around
each one are genuinely exercised (e.g. one pair shows 6 failure hits + 3
success hits on the surrounding statement lines, with only the brace-only
line itself at 0), matching an already-established pattern elsewhere in
this same function (`print_json`'s empty-container checks) — an
llvm-cov region-mapping artifact for consecutive brace-only lines, not a
real gap.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, twice —
      first run hit the known #935 machine-contention stall on an
      unrelated test, second run fully green)
- [x] Diffed output against `/usr/bin/jq` 1.7.1 directly for every repro
      shape (trailing comma, empty-with-comma, doubled comma, nested,
      leniency-fallback success cases)
- [x] `cargo llvm-cov` + `omni-dev coverage diff` — 94.87% patch coverage,
      remainder explained above

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -3805,16 +3805,41 @@ fn following_gap_ok(text: &[u8], start: usize, expected: Option<u8>) -> Option<u
 ///
 /// Deliberately narrower than a complete fix: `gap_start` must already be
 /// known cheaply. Callers only have that when the last child is a scalar
-/// (its `text_range()` is a bounded, self-delimiting scan, not the
-/// O(subtree) walk `JsonCursor::text_range`'s own comment describes for a
-/// container) or when the container is empty. When the last child is
-/// itself a container, finding *its* closing bracket costs exactly what
-/// this function's sibling `preceding_gap_ok` was written to avoid, so
-/// callers skip this check in that case rather than pay for it -- the same
+/// (via `scalar_end_pos`, an O(1) length read, not a fresh scan) or when
+/// the container is empty. When the last child is itself a container,
+/// finding *its* closing bracket costs exactly the O(subtree) walk this
+/// function's sibling `preceding_gap_ok` was written to avoid, so callers
+/// skip this check in that case rather than pay for it -- the same
 /// "tracked as a follow-up, not folded in here" trade #1643 already made,
 /// narrowed by one more case rather than eliminated.
 fn trailing_gap_ok(text: &[u8], gap_start: usize, close_char: u8) -> bool {
     matches!(following_gap_ok(text, gap_start, None), Some(pos) if text.get(pos) == Some(&close_char))
+}
+
+/// Cheap end position (one past the last byte) of an already-resolved
+/// scalar `value` known to start at `start` -- `None` for a container or
+/// unresolved value, which callers treat as "can't determine, skip"
+/// (#1676's own deferred case for a container, or a defensively-skipped
+/// unresolved cursor).
+///
+/// Deliberately doesn't call `JsonCursor::text_range()`: every caller here
+/// already resolved `value` via `.value_at(start)` (reusing the *same*
+/// `start` a caller-side `text_position()` already paid for -- see
+/// `check_preceding_delimiter`'s own doc comment on why a second lookup
+/// measured 19-30% slower for #1643), so re-deriving the position and
+/// re-dispatching on the leading byte a second time would silently
+/// reintroduce that exact regression. `raw_bytes()` on a string/number is
+/// already the value's own exact text span; `true`/`false`/`null` are
+/// fixed-width literals.
+fn scalar_end_pos<W: AsRef<[u64]>>(start: usize, value: &StandardJson<'_, W>) -> Option<usize> {
+    match value {
+        StandardJson::String(s) => Some(start + s.raw_bytes().len()),
+        StandardJson::Number(n) => Some(start + n.raw_bytes().len()),
+        StandardJson::Bool(true) => Some(start + 4),
+        StandardJson::Bool(false) => Some(start + 5),
+        StandardJson::Null => Some(start + 4),
+        StandardJson::Array(_) | StandardJson::Object(_) | StandardJson::Error(_) => None,
+    }
 }
 
 /// Array-element wrapper around [`preceding_gap_ok`]: element `index` (0-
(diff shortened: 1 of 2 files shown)


newhoggy:
Ran a 3-agent \`/code-review\` pass on this branch. Summary of what changed as a result:

**Fixed (real regression risk):** the trailing-gap tracking called \`.value()\`/\`.text_range()\` per child, both of which internally re-derive \`text_position()\` even though the caller had already resolved it — the same double-lookup shape #1643's own doc comment measured at 19-30% slower. Replaced with a new \`scalar_end_pos\` helper (O(1) length read from the already-resolved value, reusing \`value_at(start)\`) at all 4 call sites. Also dropped the now-redundant \`last_is_container\` bool and fixed a missing \`array_scratch.truncate(base)\` on the array arm's error path.

**Fixed (test quality):** a misleading doc-comment, added the empty-inner-container variant to the pinned known-gap test, and closed a small pre-existing coverage gap the diff's line-shift surfaced (the #1643 array-arm check was previously untested under \`-S\` for an array shape, only an object).

**Filed as a comment, not a new issue:** a review agent independently cross-file-traced this whole class of check and found it doesn't reach `eval_generic.rs`'s `to_owned_at_depth` (the primary evaluator backing `.[]`/`keys`/`add`/`map`/`to_entries`) at all — but that's already tracked as #1677, with an active PR (#1747) in flight. Posted the agent's additional findings (trailing-comma reproduces there too; a second independent evaluator in `eval.rs` may still have the gap) as a comment on #1677 rather than duplicating it.

Patch coverage: 96.19% (101/105 new lines) on the final commit; the 4 remaining lines are bare closing-brace lines where raw \`lcov\` hit counts confirm both branches around each one are genuinely exercised (an llvm-cov region-mapping artifact, not a real gap).

Full test suite green (twice — one run hit the known #935 machine-contention stall, unrelated to this change), both clippy legs clean.

Issue #1642, opened 2026-08-25
jq: an undecodable object key gets four different answers from the same document (keys/to_entries/has raise; length/keys_unsorted/identity do not)


newhoggy:
**Severity: Medium** — not silent data loss, but the CLI contradicts itself about whether a
document is usable, so a script's behaviour depends on which builtin it happens to call.

## Summary

`#1247`'s fix (PR #1391) made *some* key-materializing routes raise on a key whose bytes will
not decode. The routes that were left alone still answer happily, so one document now produces
four different outcomes:

```console
$ printf '{"a\\q":1,"b":2}' | sjq -c .              # {"a\q":1,"b":2}   exit 0
$ printf '{"a\\q":1,"b":2}' | sjq -c length         # 2                 exit 0
$ printf '{"a\\q":1,"b":2}' | sjq -c keys_unsorted  # ["a\q","b"]       exit 0
$ printf '{"a\\q":1,"b":2}' | sjq -c keys           # jq: error … invalid escape sequence …  exit 5
$ printf '{"a\\q":1,"b":2}' | sjq -c to_entries     # jq: error …                            exit 5
$ printf '{"a\\q":1,"b":2}' | sjq -c 'has("b")'     # jq: error …                            exit 5
$ printf '{"a\\q":1,"b":2}' | sjq -c .b             # 2                                      exit 0
```

`keys` and `keys_unsorted` disagree with each other; `keys` and `length` disagree about the same
object; `has("b")` refuses to answer a question `.b` answers fine.

Verified on a `--release --features cli` build at #1628's head. Real jq rejects the document
outright at parse time (`Invalid escape`, exit 5) for every one of these filters.

## Why the split exists

`DocumentFields::keys` and `Builtin::ToEntries` check `string_decode_error()` before
stringifying, so they raise. `DocumentFields::len` is infallible and `census`/`key_hash_of` count
an undecodable key as `unkeyed` — a real member with no comparable name — so `length` answers 2.
The `keys_unsorted` writer streams each key's **raw byte span** without decoding, so it prints
the key verbatim and never notices.

Each of those choices is individually defensible. Together they are not.

## Not the same fault as #1194

Worth stating, because the two look identical from the outside and a fix that conflates them
regresses this one. #1194 is a key the *grammar* never allowed (`{123: 1}`) — always an error.
This is a key that is structurally a string but whose bytes will not decode, and the codebase
[...]

Issue #222, opened 2026-07-15
yq: -I 0 changes document values, not just formatting


newhoggy:
## Summary

`succinctly yq` has **two** YAML→JSON implementations, and it selects between them based
on output formatting. They do not agree. Across the YAML Test Suite they produce different
**values** — not merely different whitespace — for **29 of 402** cases.

An indentation flag must not alter data.

## Reproduction

```console
$ cat 26DV.yaml
"top1" :
  "key1" : &alias1 scalar1
top3: &node3
  *alias1 : scalar3

$ succinctly yq -o json '.' 26DV.yaml | tr -d ' \n'
{"top1":{"key1":"scalar1"},"top3":{"scalar1":"scalar3"}}    # correct

$ succinctly yq -o json -I 0 '.' 26DV.yaml
{"top1":{"key1":"scalar1"},"top3":{"":"scalar3"}}           # WRONG
```

The key is an alias (`*alias1 : scalar3`). The DOM path resolves it to `scalar1`; the
streaming path emits an empty string.

## Cause

`src/bin/succinctly/yq_runner.rs` gates the M2/P9 streaming fast path on `compact`:

```rust
let can_json_fast_path = is_m2_streamable
    && output_config.compact          // <-- only with -I 0
    && output_config.output_format == OutputFormat::Json
    ...
```

- **Compact** (`-I 0`) → streaming path, `YamlCursor::to_json` / `stream_json`
- **Pretty** (default) → builds an `OwnedValue` DOM via `yaml_to_owned_value`

Two implementations of the same conversion that have drifted apart.

## Scope

Neither path is uniformly right. Where the suite can adjudicate the 29 divergent cases:

| Path                     | Right |
|--------------------------|-------|
| DOM (pretty)             | 15    |
| Streaming (`-I 0`)       | 2     |
| Both wrong               | 3     |
| No JSON form to compare  | 9     |

Divergence classes: alias-as-key, empty keys, block-scalar folding whitespace, trailing tabs.

Cases where the streaming path fails and the DOM path passes are in
`tests/data/yaml-test-suite-known-failures.txt` under category `streaming`:

```
26DV 2G84/02 DE56/04 DE56/05 E76Z K858 M7A3 NP9H
```

The full 29 (including where the DOM path is the wrong one):

```
26DV 2G84/02 4FJ6 6WPF 7A4E 9MMW 9TFX DE56/00 DE56/01 DE56/02 DE56/03 DE56/04
[...]

PR #1752, opened 2026-08-28
fix(yq): raise instead of silently dropping a colliding complex-key value

newhoggy:
## Summary

`yaml_to_owned_value`'s `Mapping` arm (the DOM materializer backing `--slurp`,
`--eval-all`, and `--inplace`'s DOM fallback path) did a bare
`map.insert(key, value)` with no collision guard. Two different complex
(non-scalar) YAML keys — e.g. two different sequence keys — both stringify to
`""` per #222's own documented key-display rule. Real yq's streaming/DOM
paths keep both entries (its underlying representation isn't a plain map),
but `OwnedValue::Object`'s `IndexMap<String, _>` structurally cannot hold two
entries under one key, so the second insert silently overwrote the first
with no error at all (exit 0):

```console
$ printf '? [1,2]\n: a\n? [3,4]\n: b\n' | succinctly yq --slurp '.[0]' -o=json
{
  "": "b"
}
```

This is a `YamlCursor`-native materializer, not `DocumentValue`-generic, so
it wasn't one of the call sites #1738 (PR #1748) fixed for the analogous
JSON decode-failure case — found while reviewing that PR.

## Approach

Verified against real yq (v4.53.3, `eval-all`/`-i`) before picking a
direction: both keep both entries in a representation succinctly's
`IndexMap`-backed `OwnedValue` can't structurally match, so — matching the
#1642/#1738 precedent for the same class of problem — this raises a clear
error instead of silently dropping, rather than attempting to fake
duplicate-key preservation in a data structure that can't hold it.

- Added `YamlValue::key_string_kind()`, splitting `key_string()`'s own
  classification to also report whether the `""` it returned is a
  **fallback** (complex/undecodable key) rather than a genuine decode —
  mirrors JSON's `key_display_string_kind`.
- Widened `DisplayKeyGuard::check` from `pub(crate)` to `pub` so
  `yq_runner.rs` (a separate binary crate) can drive it directly with YAML's
  own classification, reusing the same collision policy #1642/#1738 already
  established: an **ordinary** repeated genuine key still overwrites
  (matching jq's own last-key-wins), only a fallback-spelling collision on
  either side raises.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite green; one
      run hit two spurious `signal 9`-killed child-process failures under
      heavy concurrent-session load on this machine, both confirmed to pass
      in isolation and unrelated to this change — neither test touches YAML
      mapping keys)
- [x] Live-verified against `/usr/bin/jq`-equivalent real `yq` v4.53.3 for
      the repro and the direction decision
- [x] Regression tests for: colliding complex keys, colliding decode-failure
      string keys, colliding alias-to-decode-failure-string keys, and an
      ordinary repeated scalar key (confirms it still overwrites, not raises)
- [x] `cargo llvm-cov` + `omni-dev coverage diff` — 100% patch coverage
      (18/18 new lines)

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -786,6 +786,35 @@ another key of the same name and cannot be represented
 An *ordinary* repeated key (no decode failure on either side) is unaffected and still
 collapses to its last value, matching yq's normal duplicate-key handling.
 
+**`--slurp`/`--eval-all`/`--inplace`'s own DOM fallback catches a second, wider trigger
+the paragraph above's routes do not, as of
+[#1749](https://github.com/rust-works/succinctly/issues/1749).** Those three flags
+materialize through a separate, `YamlCursor`-native conversion
+(`yaml_to_owned_value` in
+[src/bin/succinctly/yq_runner.rs](../../../src/bin/succinctly/yq_runner.rs)), not the
+`DocumentValue`-generic path `--arg`/`-P`/`.,.,` use. It reuses the same
+`DisplayKeyGuard`/`colliding_display_key_error` machinery, but drives it with its own
+`YamlValue::key_string_kind` classification, which also flags a **complex** key
+(mapping, sequence, `null`, or a non-scalar/dangling alias -- not just a decode-failure
+string) as a fallback spelling. Real yq keeps both entries in this case too (its
+underlying representation isn't a plain map); succinctly's `OwnedValue::Object`
+structurally cannot, so this raises rather than silently discarding one:
+
+```console
+$ printf '? [1,2]\n: a\n? [3,4]\n: b\n' | succinctly yq --slurp '.[0]'
+Error: object key "" is ambiguous: an undecodable key's display form collides with
+another key of the same name and cannot be represented
+```
+
+**This wider trigger is currently `--slurp`/`--eval-all`/`--inplace`-only.** The
+`--arg`/`-P` route's own `key_display_string_kind` (JSON-oriented, shared across every
+`DocumentValue` implementor) does not yet recognize a YAML complex key as fallback --
+only a decode-failure string key, same as the paragraph above -- so the identical input
+through `--arg`/`-P` still silently drops one entry rather than raising. Tracked as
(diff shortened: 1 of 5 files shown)


newhoggy:
Ran a 5-agent \`/code-review\` pass on this branch. Summary of what changed as a result:

**Fixed:** reused the JSON bridge's \`colliding_display_key_error\` instead of hand-writing a second, wording-drift-prone copy (widened it from \`pub(crate)\` to \`pub\`); deduplicated \`key_string_kind\`'s repeated \`Ok/Err\` classification into a shared helper; added tests for two gaps the test-coverage pass found (a genuine \`""\` key colliding with a fallback \`""\` key in both orders, and three colliding keys in one mapping, not just two); documented this fix's own divergence in \`docs/compliance/yq/limitations.md\` per ADR-0018's recording rule, including that the wider trigger (a complex, non-decode-failure key) is currently \`--slurp\`/\`--eval-all\`/\`--inplace\`-only.

**Filed as a follow-up, not folded into this PR (#1753):** a review agent traced the true root cause one layer deeper -- \`src/jq/document.rs\`'s \`key_display_string_kind\`, shared by the *general* \`eval_generic.rs\`/\`lazy.rs\` materializers, doesn't recognize a YAML complex key as fallback either, and it's reached by \`walk\`, \`map_values\`, and even a plain assignment (\`.a = 1\`) -- not just the three flags this issue covers. The same pass found two independent, worse bugs in \`eval.rs\`'s own \`load()\` builtin conversion (no guard at all; a decode-failure key silently drops the whole field, not just falls back to \`""\`), and a pre-existing one-hop-only alias-resolution limit that this PR's new guard can now escalate into a spurious "ambiguous" error on a legitimate key. All three are filed together since a proper fix to the shared classifier would likely close most of them at once.

Patch coverage: 100% (21/21 new lines) on the final commit. Full test suite green (twice -- two separate runs each hit one unrelated \`signal 9\`-killed test under heavy concurrent-session load on this machine, confirmed to pass in isolation both times), both clippy legs clean.

Issue #1734, opened 2026-08-27
jq: eval.rs's own to_owned_at_depth still silently drops an undecodable key (#1642 follow-up, dead path today)


newhoggy:
## Summary

`src/jq/eval.rs`'s `to_owned_at_depth` (line ~485) is a second, independent `StandardJson -> OwnedValue` converter, distinct from `eval_generic.rs`'s `to_owned_at_depth` that #1642/#1247 fixed. It still uses the pre-#1247 pattern:

```rust
StandardJson::String(s) => {
    if let Ok(cow) = s.as_str() {
        OwnedValue::String(cow.into_owned())
    } else {
        OwnedValue::String(String::new())   // silently becomes ""
    }
}
...
StandardJson::Object(fields) => {
    ...
    if let StandardJson::String(key_str_val) = field.key() {
        if let Ok(cow) = key_str_val.as_str() {
            map.insert(cow.into_owned(), to_owned_at_depth(&field.value(), depth + 1));
        }
        // else: field silently dropped, no error
    }
    ...
}
```

It calls neither `key_display_string`/`key_display_string_kind` nor `DisplayKeyGuard` — a key that fails to decode is silently dropped instead of preserved-or-raised, and an undecodable *value* becomes an empty string instead of raising. This is the same "second evaluator missed the fix" pattern already seen in #1054.

## Why this is filed rather than fixed inline

This function (and its sibling `to_owned`) is called from ~250 sites throughout `eval.rs`'s internal arithmetic/comparison/builtin machinery — changing its signature to `Result<OwnedValue, EvalError>` to add a guard would require touching every call site, a much larger blast radius than #1642's own scope.

**Currently a dead path, not a live bug**: its only external caller is `yq_runner.rs`'s JSON-reindex bridge, which re-serializes an already-materialized `OwnedValue` whose keys/strings are already valid (they passed through a guarded conversion earlier) — so the buggy branches above are unreachable in practice today. Internally, it's also reached by `reduce`/`foreach`/`while`/`until`/`repeat` accumulators, which only ever hold computed values, never raw undecodable document bytes.

## Risk

Any future caller that routes raw, potentially-undecodable document text through `jq::eval::to_owned`/`to_owned_at_depth` (rather than through `eval_generic`'s guarded conversion first) would silently reintroduce the pre-#1642 drop-key/replace-with-empty-string bug, with no compiler signal.

## Related

- #1642 — the branch that fixed the *other* evaluator's copy of this exact pattern.
[...]

Issue #1728, opened 2026-08-27
jq: eval.rs's own to_owned_at_depth silently drops a malformed key and corrupts path() output on an undecodable dynamic index


newhoggy:
Split off from #1660 (PR #1725... wait, PR for #1660) — the most severe finding
in that issue's decode-failure catchability sweep, deliberately not included in
that fix given its disproportionate scale (a ~250-call-site signature change)
relative to the rest of the sweep.

## Summary

`src/jq/eval.rs` has its own, independent `to_owned`/`to_owned_at_depth` — a second
lazy `StandardJson`-to-`OwnedValue` conversion distinct from the one in
`eval_generic.rs` that #1620/PR #1647 taught to raise on a decode failure. This one
still silently drops a malformed object key, and silently coerces an undecodable
string used as a dynamic index into `""`, instead of raising. Confirmed live via the
`eval()` API (`query!` macro in `eval.rs`'s own test module):

```
path(.[])       on {"<bad-utf8-key>": 1, "b": 2}   => ["b"]        (should raise)
path(.[.a])     on {"a": "<bad-utf8>", "": 99}     => [""]         (should raise)
```

The second case isn't merely a suppressed error — it computes and returns a
materially wrong path, because the undecodable string is silently coerced to `""`
and used to index. `builtin_path` is one of ~250 call sites of this
`to_owned_at_depth` (measured via `grep -c` against the `to_owned(` wrapper alone in
`src/jq/eval.rs`).

Like the rest of #1660's findings, this is reachable only through the public
`succinctly::jq::eval()` library API — the CLI's `eval_generic.rs` reindex bridge
already materializes/validates the whole document before any of this code runs, so
it is not a live CLI regression.

## Root cause

```rust
fn to_owned_at_depth<W: Clone + AsRef<[u64]>>(
    value: &StandardJson<'_, W>,
    depth: usize,
) -> OwnedValue {
    ...
    StandardJson::String(s) => {
        if let Ok(cow) = s.as_str() {
            OwnedValue::String(cow.into_owned())
        } else {
            OwnedValue::String(String::new())   // <-- silent coercion
        }
    }
    StandardJson::Object(fields) => {
        ...
        if let StandardJson::String(key_str_val) = field.key() {
            if let Ok(cow) = key_str_val.as_str() {
                map.insert(cow.into_owned(), to_owned_at_depth(&field.value(), depth + 1));
            }
            // else: field silently dropped
        }
        ...
    }
    ...
}
```

[...]

PR #1754, opened 2026-08-28
fix(jq): raise decode failures in eval.rs's private to_owned family

### LINKED ISSUE #1746: jq: eval.rs's own to_owned/keys/.[]/map/to_entries never got the #1247/#1620/#1660 decode-failure fix
## Summary

Found during #1661's own `/code-review` (multi-angle fan-out, PR #1745), while checking
whether `is_uncatchable()`'s doc-comment claim ("never suppressed by `?`, never handed to
a `catch` handler ... at every call site that checks them") actually holds throughout
`eval.rs`. It doesn't -- `eval.rs` has its own private, non-generic evaluator (used by
`path()`/`=`/`|=`/`+=`/`del()` via `resolve_node`/`builtin_path`/`eval_assign`) that
never received the #1247/#1620/#1660 decode-failure fix sweep the `eval_generic.rs`
evaluator got. Confirmed live by three independent reviewer angles converging on the
same root cause, each with its own repro.

Not introduced by #1661 -- pure pre-existing gap, unrelated to that PR's diff. Filed
separately per this session's usual scope discipline rather than folding into that PR.

## Root cause

`eval.rs`'s own `to_owned`/`to_owned_at_depth` (lines ~505-563, a `StandardJson`-specific
twin of the already-fixed `eval_generic::to_owned_at_depth`) still has the **pre-#1247**
behavior in its `StandardJson::String` arm:

```rust
StandardJson::String(s) => {
    if let Ok(cow) = s.as_str() {
        OwnedValue::String(cow.into_owned())
    } else {
        OwnedValue::String(String::new())   // silently drops the decode failure
    }
}
```

Its already-fixed twin, `eval_generic::to_owned_at_depth`, explicitly raises
`EvalError::decode_failure(reason)` here and cites #1247 in its own doc comment. This
private copy was apparently missed by every pass of that fix sweep.

Every actual entry point into `resolve_node` materializes the document through this
buggy copy: `builtin_path`, `eval_assign`, `eval_update`, `yq_assign_noop_check`.
`eval.rs` also has its own separate, hand-duplicated native-arm implementations of
`keys`/`.[]`/`map`/`to_entries` (`builtin_keys`, `Expr::Iterate`'s native arm,
`builtin_map`, `builtin_to_entries`) that share the same wildcard fallback into this
`to_owned`, so none of them raise on a decode failure either -- unlike `length`/
`tonumber`/`reverse`/`startswith`/etc. in the same file, which already explicitly
match `s.as_str()` and raise (per an existing test comment, `#1660`'s "second-pass
sweep" evidently didn't cover this family).

`docs/plan/decode-failure-routing.md`'s "Non-goals" section explicitly excludes
"`eval.rs`'s private copy of the `to_owned` family" on the stated premise that
"`eval.rs`'s evaluator only processes JSON re-serialized from an already-decoded
`OwnedValue`, which by construction cannot contain invalid UTF-8." The repro below
shows that premise is false for the crate's public library API.

## Repro (library API — `succinctly::jq::eval`/`eval_using`, per the crate's own doc example)

Silent data corruption of an **untouched sibling field**:

```
{"a": 1, "b": "\xff\xfe"} | .a = 5
```
returns `{"a": 5, "b": ""}` (the invalid-UTF-8 `"b"` silently becomes `""`, no error,
exit success) instead of raising.

Decode-failure silently swallowed by `?` where the value-position equivalent correctly
raises:
```
.a | tonumber?                    -> Error(is_decode_failure() == true)   [correct]
path(select(.a|tonumber)?)        -> QueryResult::None                    [wrong -- swallowed]
```
because `builtin_path`'s `to_owned` has already turned `.a` into `""` before
`resolve_node` ever runs, so `tonumber("")` fails as an *ordinary* type error that `?`
correctly-but-wrongly suppresses.

`try keys/.[]/to_entries/map(.) catch "x"` on an undecodable string all wrongly catch
what should be an uncatchable decode failure (`try length catch "x"` correctly does
not), confirmed against `{"a": "\xff\xfe"}` via the crate's public
`succinctly::jq::eval::<Vec<u64>, JqSemantics>` entry point.

## Scope: not reachable via the shipped CLI

`jq_runner.rs`/`yq_runner.rs` instead route reads through `eval_generic::eval_with_cursor`
(the evaluator #1620/#1660/#1247's fixes actually landed in), and every write/assignment
path eagerly, strictly materializes the whole document upfront -- verified live that
even `.a = 5`/`del(.a)` on an *untouched* decode-failing field aborts immediately via
that path, before `eval.rs`'s buggy `to_owned` is ever reached. The exposure is
specifically at the public `pub fn eval`/`eval_using` library-API surface, for any
caller that indexes untrusted bytes directly and evaluates a `path()`/assignment/
`keys`/`.[]`/`map`/`to_entries` query against it.

One CLI-level symptom was independently *observed but not fully root-caused* during the
same review (`succinctly jq '.b = 2'` on a file with invalid UTF-8 in an untouched `.a`
silently emitting the corrupted value at exit 0) -- flagged as unconfirmed, possibly a
related but distinct gap, worth checking separately before assuming it's the same cause.

## Secondary finding: `resolve_node`'s `Expr::Optional` (bare `?`) arm has the same gap as its `Expr::Try` sibling did before #1661

`resolve_node`'s `Expr::Optional` generic arm (`eval.rs`, guard around what's currently
~line 17563-17566) only excludes `is_invalid_path_expression()` and (conditionally)
`is_untracked_navigation_error()` -- it never checks `is_decode_failure()`/
`is_uncatchable()`, unlike the sibling `Expr::Try` arm #1661 just migrated to
`!e.is_uncatchable()`. Since `expr?` is documented in this same file as sugar for
`try expr`, the two should behave identically here and currently don't (in principle --
see below).

**Currently unreachable/dead**, per multiple independent live probes: `resolve_node`
only ever operates on an already-fully-materialized `OwnedValue` (a Rust `String` can't
carry undecoded bytes), and the buggy `to_owned` above already silently erases any
decode failure before `resolve_node`'s recursion starts -- so today, nothing in
`resolve_node`'s own call tree can freshly manufacture a `decode_failure` `EvalError`
for this arm to see. `path(.a?)`/`path(try .a)` on `{"a": "\xff\xfe"}` both return a
clean path with no error, confirming this.

This gap only becomes *reachable* once the primary `to_owned` bug above is fixed (at
which point `resolve_node` could see a real `EvalError::decode_failure` for the first
time) -- so the primary fix should land first, and this arm should be updated
(`e.is_invalid_path_expression()` -> `e.is_uncatchable()` in its OR) in the same pass or
immediately after, not left stale a second time.

## Suggested direction

Not attempted here. Two independent pieces:

1. Fix `eval.rs`'s private `to_owned`/`to_owned_at_depth` to raise `decode_failure`
   instead of substituting `String::new()`, mirroring `eval_generic::to_owned_at_depth`'s
   already-fixed shape. Likely also needs `builtin_keys`/`Expr::Iterate`/`builtin_map`/
   `builtin_to_entries` (or whatever downstream code assumed `to_owned` couldn't fail)
   checked for fallout once it can.
2. Once (1) lands and `resolve_node` can actually see a `decode_failure`, update
   `Expr::Optional`'s generic arm to also exclude it via `is_uncatchable()`.

`docs/plan/decode-failure-routing.md`'s "Non-goals" section will need its premise about
this function family corrected regardless of which fix lands first.

## Related

- #1661 / PR #1745 -- the review that found this.
- #1247, #1620, #1660 -- the original decode-failure raising sweep(s) this family was
  apparently missed by.

newhoggy:
## Summary

- `eval.rs`'s own `StandardJson`-specific `to_owned`/`to_owned_at_depth` — used only by `path()`/`=`/`|=`/`+=`/`del()` and reachable via the crate's public library API (`succinctly::jq::eval`/`eval_using`), never via the shipped CLI — never received the #1247/#1620/#1660 decode-failure raising sweep its already-fixed twin `eval_generic::to_owned_at_depth` got.
- A decode-failure string silently became `""` instead of raising, corrupting an untouched sibling field: `{"a": 1, "b": "\ud800"} | .a = 5` returned `{"a": 5, "b": ""}` at exit 0 instead of raising.
- `docs/plan/decode-failure-routing.md`'s "Non-goals" section had explicitly excluded this function family on the premise that it's "not live-reachable" — true for the shipped CLI, false for the public library API. Corrected.

## Fix

Added a fallible `to_owned_checked`/`to_owned_checked_at_depth` sibling mirroring `eval_generic::to_owned_at_depth`'s already-fixed shape, and routed this file's **primary result-value materialization points** through it — the whole-document read/write family and three container-transform builtins:
- `builtin_path`
- `eval_assign` (RHS values, `pristine`, and `yq_assign_noop_check`'s own `pristine`)
- `eval_update`
- `builtin_setpath`
- `builtin_del`
- `delpaths_one`
- `map`/`map_values`
- `to_entries`

`yq_assign_skip_rhs`'s `Err` falls through to `None`, since its callers already fall back to `eval_update` on `None`, whose own fixed `to_owned_checked` raises the identical error correctly — no cascading signature change needed there.

Also widened `resolve_node`'s `Expr::Optional` guard to exclude `is_uncatchable()`, matching the sibling `Expr::Try` arm #1661 already migrated (`expr?` is documented sugar for `try expr`, and the two must agree). **This is a real, live behavior change**, not dead code as an earlier version of this PR claimed — confirmed by review with a direct repro: `path(.a | tonumber?)` on invalid UTF-8 previously suppressed `tonumber`'s own pre-existing decode-failure raise via the bare `?`; it now correctly propagates, entirely independent of anything `to_owned_checked` touches. Pinned by its own regression test.

## Scope correction during review

Code review's own systematic audit of every `to_owned` call site in this file, across two review passes, found:
1. **`builtin_setpath`/`builtin_del`/`delpaths_one` were missed by the first pass** despite being within #1746's own stated scope (`del()` is named explicitly in the issue's summary line, and these three are the exact same "whole-document read-modify-write" shape as `builtin_path`/`eval_assign`/`eval_update`) — added to this PR.
2. **Roughly 50 more sites** across read-only builtins (`flatten`, `pick`, `omit`, `add`, `sort`, `unique`, `group_by`, `min_by`/`max_by`, `getpath`, `recurse`/`walk`/`tostream`, every `@format` function, and more) have the identical bug shape but are a genuinely separate, larger body of work — filed as **#1755** rather than folded into this already-substantial PR.
3. `eval_rhs_once` (the RHS materializer for `+=`/`-=`/`*=`/`/=`/`%=`/`//=`) still calls the unchecked `to_owned`, but its corruption is provably unobservable: its only two callers always call the now-fixed `eval_update` immediately afterward on the same input, which independently re-checks the whole document first. Pinned by a regression test documenting the masking mechanism rather than left as a silent gap.
4. `delpaths_one` didn't bypass `optional` for a decode failure the way every other converted site does — found live-unreachable via any real jq/yq syntax today, but fixed for consistency and pinned regardless.
5. A narrow, newly-*observable* inconsistency in yq mode: `=` and `+=`-family operators now disagree on which error wins when *both* a decode failure and an RHS error are present in the same call. Neither silently succeeds — both still raise, just a different error — noted on #1755 for whoever picks that up next, since fixing it needs an internal-consistency decision (no real-yq oracle exists for either operator shape here).

## Deliberately out of scope (tracked in #1755 / #1728 / #1734)

This file has ~150 remaining `to_owned` call sites: the ~50 read-only builtins above (#1755), `QueryResult` collector boilerplate, and error-message rendering — all audited during review and confirmed either safe (operate only on already-validated data, or are type-only/structure-only and can't surface a string-content bug) or out of scope with a documented reason.

Key-handling is also unchanged: an object field whose key fails to decode is still silently dropped — a separate, #1194-shaped structural gap already tracked by #1728/#1734, not this issue's #1247-shaped value-corruption one.

## Repro (before fix)

```
{"a": 1, "b": "\ud800"} | .a = 5
```
returned `{"a": 5, "b": ""}` via the library API, corrupting the untouched `.b` field silently at exit 0.

## After fix

Raises a catchable `EvalError::decode_failure`, matching `eval_generic`'s existing behavior for the identical shape, consistent across `.a = 5`, `.a |= (.+1)`, `.a += .b`, `path(.a)`, `path(.a | tonumber?)`, `map(.)`, `map_values(.)`, `to_entries`, `setpath(...)`, `del(...)`, and `delpaths(...)` (including under `?`).

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 0 failed — one run hit 3 known-flaky tests under this machine's concurrent-session contention; confirmed passing in isolation, unrelated to this diff)
- [x] `cargo doc --no-deps --all-features` (no warnings)
- [x] `cargo fmt --check`
- [x] `cargo build --no-default-features` (`no_std` compiles)
- [x] 21 new regression tests in `eval.rs`'s own test module (library-API level, since this evaluator isn't CLI-reachable): the primary repro, uncatchability via `?`, `eval_update`, `map`/`map_values` (all `One`/`Many`/object/array shape combinations), `to_entries` (object and array arms), `builtin_path` directly, `eval_assign`'s RHS materialization, `yq_assign_noop_check` (yq mode), `setpath`/`del`/`delpaths` (failure and success paths, plus `delpaths`'s `optional`-bypass), `eval_rhs_once`'s masking behavior, `resolve_node`'s `Expr::Optional` widening (the live repro above), plus success-path tests covering the `Ok` arm of every new match block
- [x] `omni-dev coverage diff`: patch coverage in the high 80s/low 90s across every revision of this PR — every uncovered line is either my own tests' unreachable `panic!` fallback arms (standard pattern throughout this codebase), a pre-existing unrelated line swept in by diff-hunk context, or one line documented as unreachable in practice by its own existing doc comment
- [x] Manually verified the exact repro from the issue, plus that ordinary (non-decode-failing) documents are unaffected

Fixes #1746

### DIFF
--- a/docs/plan/decode-failure-routing.md
+++ b/docs/plan/decode-failure-routing.md
@@ -627,12 +627,22 @@ what Stage 6 is left with is bad escapes only.
   this work touches four of them — but consolidating them is a refactor with its own risk
   profile and should follow, not accompany, a behaviour change. Sequenced after this per
   #1283's cluster E.
-- **`eval.rs`'s private copy of the `to_owned` family.** #1247 correctly notes it is not
-  live-reachable for this scenario: `eval.rs`'s evaluator only processes JSON
-  re-serialized from an already-decoded `OwnedValue`, which by construction cannot contain
-  invalid UTF-8. Site 19 (`eval.rs:427`) is included anyway, as a cheap consistency fix,
-  but no behaviour change is expected from it and none should be asserted in a test that
-  claims to exercise it end-to-end.
+- **`eval.rs`'s private copy of the `to_owned` family** — corrected by #1746 (this premise
+  was false). #1247's "not live-reachable" claim held for the shipped CLI (`jq_runner.rs`/
+  `yq_runner.rs` route reads through `eval_generic::eval_with_cursor`, and every write path
+  eagerly, strictly materializes upfront), but the crate's public library API
+  (`succinctly::jq::eval`/`eval_using`) feeds raw, untrusted bytes straight into
+  `JsonIndex::build` and this evaluator, with no re-serialization step and no upstream
+  decode-failure check to have already caught it. #1746 fixed the primary result-value
+  materialization points (`builtin_path`, `eval_assign`, `eval_update`,
+  `yq_assign_noop_check`, `map`/`map_values`, `to_entries`) via a new fallible
+  `to_owned_checked`/`to_owned_checked_at_depth` sibling, mirroring
+  `eval_generic::to_owned_at_depth`'s already-fixed shape. The plain, infallible
+  `to_owned`/`to_owned_at_depth` remains in place for this file's ~200 other call sites
+  (mostly `QueryResult` collector boilerplate and error-message rendering) — deliberately
+  out of scope for #1746, since changing that function's own signature would force a
+  Result-threading edit at every one of them in a single change; left as a documented
+  follow-up rather than attempted there.
 - **Buffering the whole record so the streaming path can reject cleanly.** That would
   reverse P9 (direct YAML-to-JSON streaming, a 2.3× win) for a malformed-input edge case.
 
--- a/docs/plan/decode-failure-routing.md
+++ b/docs/plan/decode-failure-routing.md
@@ -633,16 +633,20 @@ what Stage 6 is left with is bad escapes only.
   eagerly, strictly materializes upfront), but the crate's public library API
   (`succinctly::jq::eval`/`eval_using`) feeds raw, untrusted bytes straight into
   `JsonIndex::build` and this evaluator, with no re-serialization step and no upstream
-  decode-failure check to have already caught it. #1746 fixed the primary result-value
-  materialization points (`builtin_path`, `eval_assign`, `eval_update`,
-  `yq_assign_noop_check`, `map`/`map_values`, `to_entries`) via a new fallible
+  decode-failure check to have already caught it. #1746 added a new fallible
   `to_owned_checked`/`to_owned_checked_at_depth` sibling, mirroring
-  `eval_generic::to_owned_at_depth`'s already-fixed shape. The plain, infallible
-  `to_owned`/`to_owned_at_depth` remains in place for this file's ~200 other call sites
-  (mostly `QueryResult` collector boilerplate and error-message rendering) — deliberately
-  out of scope for #1746, since changing that function's own signature would force a
-  Result-threading edit at every one of them in a single change; left as a documented
-  follow-up rather than attempted there.
+  `eval_generic::to_owned_at_depth`'s already-fixed shape, and fixed the whole-document
+  read/write family (`builtin_path`, `eval_assign`, `eval_update`, `yq_assign_noop_check`,
+  `builtin_setpath`, `builtin_del`, `delpaths_one`) plus `map`/`map_values`/`to_entries`.
+  A systematic audit run during that PR's own review found roughly 50 more builtins with
+  the identical shape (a container's own elements/fields materialized straight into a
+  returned `QueryResult::Owned`/`ManyOwned`, uncaught) — `flatten`/`pick`/`omit`/`add`/
+  `sort`/`unique`/`group_by`/`min_by`/`max_by`/`getpath`/`recurse`/`walk`/`tostream`/format
+  functions (`@base64` etc.) among them — filed as #1755 rather than folded into #1746's
+  own PR, which was already a substantial diff closing the originally-reported repro. The
+  plain, infallible `to_owned`/`to_owned_at_depth` remains in place for those sites plus
+  this file's remaining ~150 genuine `QueryResult`-collector-boilerplate/error-message-
+  rendering sites (audited and confirmed safe or out of scope, per #1755's own site list).
 - **Buffering the whole record so the streaming path can reject cleanly.** That would
   reverse P9 (direct YAML-to-JSON streaming, a 2.3× win) for a malformed-input edge case.
 
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -1014,10 +1014,12 @@ impl EvalError {
     /// *where* the error was caught (a bare postfix `?` on the primitive
     /// that raised it vs. anything else), both of these mean the same thing
     /// at every call site that checks them: never suppressed by `?`, never
-    /// handed to a `catch` handler. Only `resolve_node`'s `Expr::Try` arm
-    /// ANDs the two checks together today, but a future always-uncatchable
-    /// error class only needs to be added here, not hand-copied into
-    /// whichever `?`/`try` dispatch sites grow the same check next.
+    /// handed to a `catch` handler. `resolve_node`'s `Expr::Try` and
+    /// `Expr::Optional` arms (`?`'s two path-context dispatch sites, kept in
+    /// agreement since #1746 — `expr?` is documented sugar for `try expr`)
+    /// both check this today, but a future always-uncatchable error class
+    /// only needs to be added here, not hand-copied into whichever `?`/`try`
+    /// dispatch sites grow the same check next.
     pub fn is_uncatchable(&self) -> bool {
         self.is_invalid_path_expression() || self.is_decode_failure()
     }
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -562,6 +562,69 @@ fn to_owned_at_depth<W: Clone + AsRef<[u64]>>(
     }
 }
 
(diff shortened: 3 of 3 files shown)


newhoggy:
Code review's own systematic audit found this fix's original site selection (8 functions) was incomplete — three more exact-sibling functions (`builtin_setpath`, `builtin_del`, `delpaths_one`) were within #1746's own stated scope (`del()` is explicitly named in the issue's summary line) and have now been added to this PR. The audit also found ~50 more sites across read-only builtins (`flatten`, `pick`, `omit`, `add`, `sort`, `unique`, `group_by`, `recurse`, `walk`, format functions, and more) with the identical bug shape — filed separately as #1755 rather than folded into this already-substantial PR.

PR #1757, opened 2026-08-28
fix(yq): place color reset before terminator in evaluated YAML output

### LINKED ISSUE #1708: yq: -0/--join-output combined with -C embeds the terminator inside the ANSI color span
**Severity: Low** (cosmetic byte-ordering, narrow flag combination)

## Summary

Found by \`/code-review\` on PR #1707 (#1701). \`-0\`/\`--join-output\` combined with \`-C\`
(color) on the M2 fast path's **evaluated** (non-identity) branch embeds the terminator
byte *inside* the ANSI color span, before the trailing reset code — unlike the identity
branch and the DOM path, both of which place it cleanly after the reset code.

## Repro (verified on the #1701 branch)

\`\`\`console
$ printf 'a: 1\n' | succinctly yq -C -0 '.a'      # evaluated (non-identity)
\x1b[36m1\x00\x1b[0m                                # NUL lands before the reset code

$ printf 'a: 1\n' | succinctly yq -C -0 '.'        # identity — correct
...\x1b[0m\x00                                      # NUL lands after the reset code

$ printf 'a: 1\n' | succinctly yq -C -0 '.a + 0'   # DOM path — also correct
\x1b[36m1\x1b[0m\x00
\`\`\`

## Root cause

\`stream_maybe_colored\` buffers the whole rendered body into a \`String\` (via
\`ColorSink::Buffered\`) when \`use_color\` is set, then wraps the *entire* buffer with
\`colorize(&buf)\` after the render callback returns. The M2 evaluated branch's terminator
write (\`terminator.write_fmt(w)\`, passed as \`stream_yaml\`/\`stream_json\`'s per-result
callback) runs *inside* that buffered render callback, so it becomes part of what gets
colorized. The identity branch instead writes the terminator via \`terminator.write_io($writer)\`
*after* \`stream_maybe_colored\` returns, bypassing the color wrapping entirely — which is
why only the evaluated branch has this ordering bug.

## Why not fixed in #1701's own PR

The per-result callback exists so a single \`stream_yaml\`/\`stream_json\` call can correctly
terminate *each* result when a query yields multiple values (e.g. \`.[]\`) — moving the
terminator write outside \`stream_maybe_colored\` entirely (matching the identity branch)
would only apply it once, after the whole colorized block, silently dropping per-result
terminators for multi-result evaluated queries. A correct fix needs either the colorizer
functions themselves to become terminator-aware, or a restructuring of how
\`stream_maybe_colored\` buffers/splices per-result content — a larger change than #1701's
otherwise-narrow \`Terminator\` centralization warranted for a first pass, and \`-C\` combined
with \`-0\`/\`--join-output\` (a terminal color aesthetic vs. a NUL-safe machine-piping mode)
is an unusual combination in practice.

## Suggested fix direction

Not attempted. Investigate whether \`colorize_json\`/\`colorize_yaml\` can be made to accept
(and correctly place) a trailing terminator byte explicitly, rather than having callers bake
it into the buffered body.

## Related

- #1701 / PR #1707 — the fix whose review surfaced this.

newhoggy:
## Summary

- `colorize_yaml` re-lexes a buffered render and relies on one trailing, unconditional reset code (`\x1b[0m`) to close whatever color span is still open at the end. The M2 evaluated-and-colored streaming path (`stream_cursor!` macro) writes the terminator (NUL or newline) *directly to the real writer*, outside that buffer, after the colorized text has already been flushed — so under `-C -0`/`-C` a still-open color span's reset code could land *after* the terminator instead of before it (e.g. `\x1b[36m1\x00\x1b[0m` instead of `\x1b[36m1\x1b[0m\x00`).
- `colorize_json` does **not** have this bug — verified live (hex-dump comparison) that its token-by-token design already self-closes each span locally, so it needed no change.
- **Fix (revised after code review — see below):** `ColorSink::record_boundary()` records each streamed result's boundary as a byte offset into the buffer (a side `Vec<usize>`), instead of writing any terminator-related byte into the text. `colorize_yaml` takes those offsets directly and, at each one, closes whatever color span is open *there* before writing the real terminator — rather than leaving it for the single trailing reset to close.

### Revision history

The first version of this PR used an in-band sentinel character (`'\u{1}'`) written into the buffer in place of the real terminator, which `colorize_yaml` then recognized and substituted. Code review found two live bugs in that approach:

1. **Sentinel collision**: this crate's own YAML parser passes a literal `U+0001` byte straight through from a double-quoted scalar (`a: "x\x01y"`) — colliding with the marker and splitting one result into two NUL-terminated records.
2. **Incomplete span-closing at the marker**: the sentinel handler only closed an open *key* span, not an open *string* span, so a value with an unescaped apostrophe (e.g. `"it's here"`) desynced `colorize_yaml`'s naive quote-toggle tracking across the boundary — reproducing the exact terminator-before-reset bug on the *next* result in the same buffer.

Both are fixed by dropping the in-band marker entirely in favor of out-of-band boundary tracking, which removes the collision (no byte value needs to be reserved) and closes *every* open span (key or string) at each boundary, isolating each result's coloring from the next regardless of what state the lexer is left in.

## Test plan

- 6 regression tests in `tests/yq_cli_tests.rs`: reset-before-terminator ordering for the evaluated M2 path, identity path unaffected, JSON output path unaffected, correct per-result ordering with multiple `-0`-terminated results, survival of an unbalanced quote/apostrophe in a streamed value (review finding #2), and survival of a literal embedded control-char byte matching the original marker (review finding #1).
- `cargo clippy --all-targets --all-features -- -D warnings` and the SIMD/x86 feature-set leg (`std,simd,serde,cli,regex,bench-runner,large-tests,mmap-tests`): both clean.
- Patch coverage: 84.72% (61/72 new lines). Every uncovered line is a pre-existing `colorize_yaml`/`colorize_json` call site in a defensive-fallback code path explicitly documented as "rarely if ever reached" (the root cursor always reports its virtual document sequence, so the `_ =>` single-document fallback arms are dead in practice) — the drop from the first version's 89.19% is purely line-count inflation from `rustfmt` wrapping those same call sites' closures across more lines now that they take an extra parameter, not new untested logic. The actual fix logic (`ColorSink::record_boundary`, the boundary-emit loops in `colorize_yaml`) is fully exercised by the 6 tests above.

Fixes #1708

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -102,6 +102,20 @@ impl<W: Write> ColorSink<'_, W> {
             boundaries.push(buf.len());
         }
     }
+
+    /// Writes a streamed result's terminator through whichever mode `self`
+    /// is in: records a boundary for `colorize_yaml` to place later when
+    /// `Buffered`, or writes it immediately when `Direct` -- so a caller
+    /// never needs its own `use_color` check to pick between the two, only
+    /// `ColorSink`'s own variant (which already reflects it).
+    fn write_terminator(&mut self, terminator: Terminator) -> core::fmt::Result {
+        if matches!(self, ColorSink::Buffered(..)) {
+            self.record_boundary();
+            Ok(())
+        } else {
+            terminator.write_fmt(self)
+        }
+    }
 }
 
 /// Streams through `render` directly when `use_color` is false. When true,
@@ -2841,24 +2855,37 @@ fn compact_item_opens_with_key(rest_of_line: &str) -> bool {
 /// (the identity/DOM paths, which write their terminator directly outside
 /// any buffer) passes an empty slice, and `yaml` is colorized exactly as
 /// before.
+// `close_and_terminate!`'s state resets are dead in its post-loop drain call
+// (nothing reads `at_key_start`/`escape_next` after the function returns,
+// and only the drain's *last* iteration's `trailing_reset_needed` write
+// matters) but live in its main-loop call -- sharing one macro definition
+// between both call sites is worth the otherwise-harmless dead stores.
+#[allow(unused_assignments)]
 fn colorize_yaml(yaml: &str, terminator: Terminator, boundaries: &[usize]) -> String {
-    let mut result = String::with_capacity(yaml.len() * 2 + boundaries.len() * 4);
+    let mut result = String::with_capacity(yaml.len() * 2 + boundaries.len() * 5);
     let mut in_string = false;
     let mut escape_next = false;
     let mut at_key_start = true;
     let mut in_key = false;
     let mut boundary_idx = 0;
     let mut byte_pos = 0usize;
+    // Whether the unconditional trailing reset below is still needed.
+    // `close_and_terminate!` (a boundary was just closed) clears it; any
+    // subsequently pushed content sets it again. Stays `true` throughout
+    // when `boundaries` is empty (the identity/DOM callers), so behaves
+    // exactly as before this mechanism existed for them.
+    let mut trailing_reset_needed = true;
 
-    let mut chars = yaml.chars();
-    while let Some(c) = chars.next() {
-        while boundary_idx < boundaries.len() && boundaries[boundary_idx] <= byte_pos {
-            // One recorded boundary per streamed result, so this correctly
-            // places every terminator in a multi-result stream, not just
-            // the last one -- and closes both spans a boundary can land
-            // inside (a still-open key *or* string color), not just
-            // `in_key`, so a leftover open span from one result can never
-            // bleed its missing reset into the next.
+    // One recorded boundary per streamed result, so this correctly places
+    // every terminator in a multi-result stream, not just the last one --
+    // and closes both spans a boundary can land inside (a still-open key
+    // *or* string color), not just `in_key`, so a leftover open span from
+    // one result can never bleed its missing reset into the next. A single
+    // definition (rather than duplicating this at both call sites below)
+    // keeps the in-loop and post-loop drains from drifting apart on a
+    // future edit.
+    macro_rules! close_and_terminate {
+        () => {
             if in_key || in_string {
                 result.push_str("\x1b[0m");
                 in_key = false;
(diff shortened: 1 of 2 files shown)


newhoggy:
**Second code-review pass** found one more live bug in the boundary-tracking redesign: `colorize_yaml`'s unconditional trailing reset still fired even when the boundary-drain loop just above it had already closed every open span and written the real terminator — placing a redundant `\x1b[0m` *after* the terminator instead of nothing, relocating (not fixing) the "reset must precede terminator" defect to the tail of the buffer. Every M2 evaluated-color result hit this.

Fixed by tracking whether real content has been written since the last boundary close, and skipping the trailing safety-net reset when it hasn't — identity/DOM callers (empty `boundaries`) are unaffected, since the flag never leaves its initial `true` state for them. Also factored the duplicated boundary-close-and-terminate logic into one local macro (the in-loop and post-loop copies had already drifted on which state they reset — a second review finding), and consolidated the `$use_color` branch into a new `ColorSink::write_terminator` method so callers no longer re-derive a fact the sink's own variant already encodes (a third, lower-severity finding).

One pre-existing test (`test_color_output_survives_iteration_with_duplicate_keys`) had asserted the old buggy trailing-reset-after-newline-terminator output verbatim; corrected its expectation, and added `test_color_terminator_ordering_terminator_is_final_byte_1708` to catch this specific regression going forward.

All local gates re-verified green: full test suite (0 failures across 3 confirmation runs — one transient unrelated failure in `test_pipe_element_rest_cache_across_drivers_1598` reproduced as a known flaky-under-contention test, confirmed passing in isolation), both clippy feature legs, `cargo doc`, `cargo fmt --check`, `cargo build --no-default-features`.

newhoggy:
**Third code-review pass** found no further correctness bugs — confirmed both prior gaps stay fixed across every code path (main-loop content-write, in-loop boundary drain, post-loop boundary drain) via a full derivation plus live edge-case probing (mid-stream error, empty-container results, zero-result streams, multiple terminator kinds). Three lower-severity style/maintainability findings, all addressed in \`d61cfde45\`:

- The `unused_assignments` suppression now cites STYLE-0004 by name (matching the codebase's other three instances) with an accurate reason — the two `close_and_terminate!()` call sites are asymmetric in which state they actually need, not both "live" as the prior comment claimed.
- `ColorSink::write_terminator` renamed to `write_result_terminator` to avoid colliding (in name only, not behavior) with the pre-existing unrelated free function `write_terminator`.

All local gates re-verified green after this pass: full test suite (0 failures), both clippy feature legs, `cargo doc`, `cargo fmt --check`, `cargo build --no-default-features`.

Ready for CI.

PR #1758, opened 2026-08-28
fix(yq): yq-locate warns instead of silently disagreeing on bad UTF-8

### LINKED ISSUE #1627: yq: yq-locate accepts an invalid-UTF-8 document that yq, yq --validate and at_offset all reject
**Severity: Low** (self-consistency, not correctness against an oracle)

## Summary

Since #1391 (#1242's fix) landed, `succinctly yq` rejects a YAML document containing
invalid UTF-8 at the input boundary. `succinctly yq-locate` does not — it accepts the
same document and answers normally, at exit 0.

The sharp part is not yq-vs-yq-locate in the abstract: it is that **the same navigation
feature disagrees with itself depending on which front door you use**.
`succinctly yq 'at_offset(9)'` rejects the document; `succinctly yq-locate --offset 9`
answers `.[0].b`.

## Repro (verified on `main` @ `1cf9bf153`, x86_64)

```console
$ printf 'a: 1\nb: "x\xe4y"\n' > bad.yaml

$ succinctly yq -o json '.' bad.yaml            ; echo $?   # rejects
Error: YAML parse error: invalid UTF-8 continuation byte at byte 11, line 2, column 7
1
$ succinctly yq --validate -o json '.' bad.yaml ; echo $?   # rejects   1
$ succinctly yaml validate bad.yaml             ; echo $?   # rejects   1
$ succinctly yq -o json 'at_offset(9)' bad.yaml ; echo $?   # rejects   1

$ succinctly yq-locate bad.yaml --offset 9      ; echo $?   # ACCEPTS
.[0].b
0
$ succinctly yq-locate bad.yaml --line 2 --column 5 ; echo $?   # ACCEPTS
.[0].b
0
```

`src/bin/succinctly/yq_locate.rs` has its own input path and never calls
`yaml_validate_guard`, which is where #1391 put the encoding check.

## This is a decision, not an obvious bug

There is a real argument for leaving it exactly as it is: a *locate* tool is a
diagnostic. Wanting to find what is at byte 9 of a file that a parser just rejected is
arguably the normal case for it, not an edge case — refusing to answer would remove the
tool precisely when it is most useful.

There is no oracle to settle it. `yq-locate` is a succinctly extension; real yq has no
locate subcommand, so ADR-0018's reference-fidelity rule does not decide this one. It is
purely a question of what succinctly wants to be self-consistent about.

**`jq-locate` is not affected and should not change.** jq mode substitutes U+FFFD rather
than rejecting, so `jq-locate` accepting the document is already consistent with
`succinctly jq`.

## Options

1. **Leave it, document it.** Add a line to `docs/compliance/yaml/limitations.md` saying
   `yq-locate` deliberately accepts what `yq` rejects, and why.
2. **Make `yq-locate` reject too**, reusing the same check, for one rule across every
   YAML entry point. Costs the diagnostic use case above.
3. **Warn and continue** — locate still answers, but prints the encoding diagnostic to
   stderr and keeps exit 0. Keeps the tool useful and stops the silent disagreement.

(3) looks like the best of the three, but (1) is defensible and free.

## Context

Filed out of #1242's own "Also check" list, which asked whether `yq-locate` and
`at_offset`/`at_position` had their own silent fallback for the same decode failure.
Now answered: `at_offset`/`at_position` are fine (they route through the yq runner and
inherit its gate), `yq-locate` is the one gap.

newhoggy:
## Summary

- `succinctly yq`, `yq --validate`, and `at_offset` all reject a YAML document containing invalid UTF-8 at the input boundary (#1391/#1242) — but `yq-locate` (`src/bin/succinctly/yq_locate.rs`) has its own input path that never called `yaml_validate_guard`, so it silently accepted the same document and answered normally. The same navigation feature disagreed with itself depending on which front door was used: `succinctly yq 'at_offset(9)'` rejects a malformed document, `succinctly yq-locate --offset 9` on the identical file answers as if nothing were wrong.
- This is explicitly framed in the issue as a design decision with three options (leave it and document it; make `yq-locate` reject too; warn and keep answering) rather than an obvious bug — there's no oracle to settle it, since `yq-locate` is a succinctly-only extension with no real-yq equivalent. Went with the issue's own recommended option: **warn to stderr, keep answering, exit 0**. `yq-locate` is a diagnostic tool, and asking "what's at byte N" of a file a parser just rejected is the normal case for it, not an edge case — refusing outright would remove the tool exactly when it's most useful. This stops the silent self-disagreement without losing that.

## Test plan

- New regression test `yq_locate_warns_but_still_answers_on_invalid_utf8` in `tests/locate_cli_tests.rs`: confirms exit 0, the correct expression is still returned, and a UTF-8 warning appears on stderr.
- Patch coverage: 100% (3/3 new lines).
- Documented the deliberate divergence in `docs/guides/cli.md`'s `yq-locate` section.

Fixes #1627

### DIFF
--- a/docs/guides/cli.md
+++ b/docs/guides/cli.md
@@ -786,9 +786,12 @@ succinctly yq-locate config.yaml --offset 42 --format json
 
 Unlike `succinctly yq`/`yq --validate`/`at_offset`, which all reject a YAML document
 containing invalid UTF-8 at the input boundary, `yq-locate` deliberately keeps answering on
-one: warns to stderr and continues, rather than refusing outright. Asking "what's at byte
-N" of a file a parser just rejected is the normal case for a diagnostic tool, not an edge
-case (#1627).
+one: warns to stderr and continues, rather than refusing outright, for the byte in question
+(#1627). Asking "what's at byte N" of a file a parser just rejected is the normal case for
+a diagnostic tool, not an edge case. This doesn't cover every invalid byte in the document,
+though: if the byte falls inside a *mapping key* on the path to the requested offset, the
+path expression still can't be built at all (a separate, pre-existing limitation in how
+`yq-locate` renders keys, not fixed by #1627) and the command still fails — see #1759.
 
 ---
 
--- a/docs/guides/cli.md
+++ b/docs/guides/cli.md
@@ -784,6 +784,12 @@ succinctly yq-locate config.yaml --offset 42 --format json
 # Output: {"expression":".users[0].name","type":"string","start":38,"end":52}
 ```
 
+Unlike `succinctly yq`/`yq --validate`/`at_offset`, which all reject a YAML document
+containing invalid UTF-8 at the input boundary, `yq-locate` deliberately keeps answering on
+one: warns to stderr and continues, rather than refusing outright. Asking "what's at byte
+N" of a file a parser just rejected is the normal case for a diagnostic tool, not an edge
+case (#1627).
+
 ---
 
 ## Environment Variables
--- a/src/bin/succinctly/yq_locate.rs
+++ b/src/bin/succinctly/yq_locate.rs
@@ -48,6 +48,17 @@ pub fn run_yq_locate(args: YqLocateArgs) -> Result<i32> {
     let text = std::fs::read(&args.file)
         .with_context(|| format!("Failed to read file: {}", args.file.display()))?;
 
+    // Unlike `yq`/`yq --validate`/`at_offset` (which all reject invalid UTF-8
+    // at the input boundary, #1391/#1242), `yq-locate` keeps answering on an
+    // encoding-invalid document instead of refusing outright (#1627): asking
+    // "what's at byte N" of a file a parser just rejected is the normal case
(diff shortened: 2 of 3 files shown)


newhoggy:
Code review found the "warn to stderr, keep answering" claim only holds for an invalid byte inside a **value** — one inside a **mapping key** on the path to the requested offset still fails outright (\`path_to_bp\` can't build a path expression without decoding every ancestor key). This is a pre-existing limitation independent of this fix (it predates #1627 and isn't worsened by it), but the docs text I'd written overclaimed.

Narrowed the wording in \`docs/guides/cli.md\`, added a regression test pinning the current (documented, not desired) failure behavior for the key case, and split the gap into #1759 rather than fixing it here — it touches path-building semantics in \`src/yaml/locate.rs\`, a different and riskier piece of code than the missing encoding check this issue was actually about.

All local gates re-verified green: full test suite (0 failures), both clippy feature legs, \`cargo doc\`, \`cargo fmt --check\`, \`cargo build --no-default-features\`, 100% patch coverage.

PR #1762, opened 2026-08-28
fix(jq): re-validate an Expr::Comma prefix's terminal check on a later escape

### LINKED ISSUE #1560: jq: path-context Comma emits a branch and an escape where jq reports one Invalid path expression
**Severity: Low** (path context; emits a branch real jq does not, alongside a different error)

## Confirmed repro (pinned jq 1.7.1)

```
$ echo '{"a":[1,2]}' | jq -c 'path(.a | (1, error("boom")))'
jq: error (at <stdin>:1): Invalid path expression with result 1        # stderr only, exit 5

$ echo '{"a":[1,2]}' | succinctly jq -c 'path(.a | (1, error("boom")))'
jq: error (at <stdin>:1): boom                                        # stderr
["a"]                                                                 # stdout — jq prints nothing
```

Two differences at once: succinctly emits a path branch jq does not, and it reports the comma's embedded `error("boom")` where jq reports its own `Invalid path expression with result 1`. jq checks the first output's path-expression validity before the second output is ever reached; succinctly appears to accept the first branch and then surface the second output's escape.

## Not from #1279

Found while verifying #1522 (generator-argument fan-out). That branch does not touch the path-context `Expr::Comma` arm — `git diff origin/main..HEAD -- src/jq/eval.rs` has no hunk matching `Expr::Comma`, `accumulate_path_context_step` or `continue_rest_with_context`. The shape also involves no builtin generator argument at all, which is the only thing #1279 changed.

The neighbourhood is instead main's own recent work: #1509 (`fix(jq): thread per-output path through path-context Comma`, and its review-findings follow-up). Worth checking against that series rather than against #1279.

## Related

- #1509 — path-context Comma per-output threading.
- #891 — the `Invalid path expression with result <v>` wording this compares against.

newhoggy:
## Summary

`resolve_dynamic_indexes` (the top-level entry point for `path()`, `del()`,
`=`, and `|=`) validates that every branch a resolved path expression
produces is "trackable" — a real navigation, not a computed value — via
`reject_untracked_at_terminal`. But it only ran that check on
`resolve_node`'s `Ok` result:

```rust
return match resolve_node::<S>(expr, input, true, false)
    .and_then(|branches| reject_untracked_at_terminal(branches, false))
{ ... }
```

When a later `Expr::Comma` sibling's own escape stops resolution first
(`resolve_node` returns `Err((prefix, e))`), `.and_then` never runs — so an
earlier sibling's own untracked-terminal violation, already sitting right
there in `prefix`, skips the check entirely. The later sibling's error
surfaces instead of jq's real "Invalid path expression" message, and the
untracked branch's stale path leaks into the assembled (would-be) prefix.

```
$ echo '{"a":[1,2]}' | jq -c 'path(.a | (1, error("boom")))'
jq: error (at <stdin>:1): Invalid path expression with result 1   # exit 5, no stdout

$ echo '{"a":[1,2]}' | succinctly-before jq -c 'path(.a | (1, error("boom")))'
jq: error (at <stdin>:1): boom                                    # wrong message
["a"]                                                              # phantom output jq never prints
```

jq validates each comma sibling's path-validity the instant it is produced,
before the next sibling ever runs — so branch 1 (`1`, untracked) is checked
and found invalid before branch 2 (`error("boom")`) is ever evaluated.

## Fix

Added `reject_untracked_prefix_too`, which runs the same terminal check on
an `Err` prefix too — giving an earlier terminal violation priority over a
later sibling's own escape, matching jq's evaluation order. Both call sites
in `resolve_dynamic_indexes` (the ordinary case and the trailing-bare-iterate
case, `near_iterate: true`/`false`) now go through it.

The existing #986 Stage 2 behavior — a *valid* prefix keeps whatever resolved
before a later real error (`path(.a, error("z"), .b)` still prints `["a"]`
then raises `"z"`) — is unaffected; the new check only intervenes when the
prefix itself contains an untracked branch.

## Verification

22-shape live differential sweep against pinned jq 1.7.1 (`/usr/bin/jq`
1.7.1), zero regressions — including the control cases this fix must not
disturb:
- `(.a, 1) | .c` (mid-pipe, non-terminal — #986 Stage 2's own case)
- `path(.a, error("x"))` / `path(.a, .b, error("z"))` / `path(.a, error("z"), .b)` (valid prefix + later real error, both orderings)
- `path((1, error("boom")) | .[])` / `path((error("boom"), 1) | .[])` (the `near_iterate: true` call site)
- `del(.a, 1)`, `(.a, 1) = 5` (the write-side callers of the same function)

## Test plan

- [x] New unit tests in `src/jq/eval.rs`: `test_path_comma_terminal_check_wins_over_a_later_siblings_escape_1560`, `test_path_comma_valid_prefix_keeps_a_later_siblings_own_error_1560`
- [x] New CLI integration tests in `tests/jq_cli_tests.rs` (same two shapes, exercised through the actual binary)
- [x] Full test suite (`cargo test --features cli,simd,regex,serde`): all green, no regressions

Fixes #1560

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -21211,6 +21211,37 @@ fn resolve_dynamic_indexes<S: EvalSemantics>(
         Ok(kept)
     }
 
+    /// Same terminal check as [`reject_untracked_at_terminal`], but also run
+    /// over whatever a *later* sibling's own escape stopped resolution with.
+    ///
+    /// jq validates each `Expr::Comma` sibling's path-validity the instant it
+    /// is produced, before the next sibling ever runs (confirmed live:
+    /// `path(.a | (1, error("boom")))` raises "Invalid path expression with
+    /// result 1" -- the first branch's own violation -- and never reaches
+    /// `error("boom")` at all; jq prints nothing to stdout). `resolve_node`'s
+    /// `Expr::Comma` arm cannot replicate that ordering on its own (#986
+    /// Stage 2's reasoning still holds: mid-pipe, it does not yet know
+    /// whether it is terminal), so it always resolves every sibling before
+    /// returning. When a *later* sibling's own escape is what stopped that
+    /// resolution, the branches collected before it still carry whatever an
+    /// earlier, terminal-but-untracked sibling produced -- and without this,
+    /// that branch skips the terminal check an all-`Ok` result gets, both
+    /// surfacing the later sibling's error instead of jq's real one and
+    /// leaking that untracked branch's stale path into the assembled prefix
+    /// (#1560).
+    fn reject_untracked_prefix_too<'a>(
+        result: Result<Vec<PathBranch<'a>>, (Vec<PathBranch<'a>>, EvalEscape)>,
+        near_iterate: bool,
+    ) -> Result<Vec<PathBranch<'a>>, (Vec<PathBranch<'a>>, EvalEscape)> {
+        match result {
+            Ok(branches) => reject_untracked_at_terminal(branches, near_iterate),
+            Err((prefix, e)) => match reject_untracked_at_terminal(prefix, near_iterate) {
+                Ok(prefix) => Err((prefix, e)),
+                terminal_violation => terminal_violation,
+            },
+        }
+    }
+
     /// Is this a bare trailing iterate — `Expr::Iterate` or
     /// `Expr::Optional(Iterate)` (`.foo[]?`) — the shape `resolve_seq`'s
     /// fan-out loop would otherwise fully enumerate one static `Index`
@@ -21277,8 +21308,7 @@ fn resolve_dynamic_indexes<S: EvalSemantics>(
         // frozen `$x` snapshot, so ambient `snapshot` starts `false` here,
         // same as it does at this function's other top-level entry point
         // below (#1591).
-        return match resolve_node::<S>(expr, input, true, false)
-            .and_then(|branches| reject_untracked_at_terminal(branches, false))
+        return match reject_untracked_prefix_too(resolve_node::<S>(expr, input, true, false), false)
         {
             Ok(branches) => {
                 if short_circuit_del_root && branches.iter().any(|b| b.path.depth() == 0) {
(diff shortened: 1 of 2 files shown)


newhoggy:
Ran \`/code-review high\` (3 finder agents). Synthesis:

**No correctness bugs found** — one agent independently live-verified 8+ additional shapes against pinned jq 1.7.1 beyond this PR's own tests, including the highest-risk case (whether a later sibling's \`halt\`/\`break\` could be wrongly suppressed by the new prefix check): confirmed the fix correctly wins over \`halt\_error\`/\`break\` too, matching jq's own behavior exactly (\`((1, halt_error(3)) = 5)?\` exits 0 on both oracle and fix; \`main\` wrongly exits 3). Also confirmed \`near_iterate\` wording and multi-sibling ordering all match byte-for-byte.

**Fixed** (2 findings, same root suggestion from two independent angles): \`reject_untracked_prefix_too\` spelled out \`Result<Vec<PathBranch<'a>>, (Vec<PathBranch<'a>>, EvalEscape)>\` by hand instead of using the file's own \`PathResolveResult<'a>\` alias, and hand-rolled the prefix/escape recombination instead of reusing the existing \`path_result\` helper built for exactly that. Refactored (commit 7b845dffc) to destructure into \`(prefix, Option<escape>)\`, run the terminal check via \`?\` (propagating a violation with priority over the original escape — the exact desired behavior), then recombine via \`path_result\`. Re-verified: full suite green, both clippy legs clean, fmt clean, same 22-shape differential sweep still matches exactly.

**Investigated, filed separately as #1764 — not fixed here (correctly out of scope)**: one agent flagged that \`resolve_dynamic_indexes\` is generic over \`EvalSemantics\`, so this fix also reaches \`succinctly yq\`, and asked whether yq-mode coverage should exist. Checked against the pinned yq oracle (Homebrew v4.53.3) and found something more specific: real yq's \`path()\` builtin doesn't exist in its grammar at all (\`path(.a)\` is a parse error in real yq), so there's no oracle-verifiable yq-mode test to write for \`path()\` itself. But \`del()\`/\`=\` — the other two callers of the same \`resolve_dynamic_indexes\` entry point — *are* valid yq syntax, and checking those surfaced a real, **pre-existing** divergence unrelated to this PR: \`reject_untracked_at_terminal\`'s unconditional "Invalid path expression" raise already ran in yq mode before this PR (on the \`Ok\`-path call site), where real yq instead silently no-ops an untracked comma branch (confirmed on unpatched \`main\`: \`del(.a, 1)\` and \`(.a, 1) = 5\` both diverge from real yq already). This PR's \`reject_untracked_prefix_too\` change doesn't introduce or worsen that gap — it symmetrically extends the same pre-existing, already-mode-agnostic check to the \`Err\`-path case, making an existing inconsistency *consistent* rather than creating a new one. The underlying yq-mode question (should this check be mode-dispatched at all?) is a separate, larger investigation — filed as #1764 rather than scope-creeping this fix.

**Not actioned**: no other findings survived — commit message format, ADR-0018 divergence-documentation, and testing-pattern checks all came back clean.

PR #1763, opened 2026-08-28
fix(scripts): ab-cli.py's idleness gate no longer flags idle Macs busy

### LINKED ISSUE #1640: bench(scripts): ab-cli.py reports an idle Mac as 30% busy — ps pcpu is a lifetime average, not instantaneous
**Severity: Medium** (no wrong output, but it blocks the idleness gate on every macOS A/B run —
and the sanctioned workaround, `--force`, also disables the battery check, which is correct)

## Summary

`scripts/ab-cli.py`'s `machine_warnings()` reports a long-idle Mac as busy. On
`johns-mac-mini` (M4 Pro, up 26 days) it reported **"30% CPU busy across all processes"** while
`top -l 2` on the same box, seconds later, reported **98.71% idle** with no consumer above 1.2%.

Reproduced across four separate invocations over ~40 minutes, at reps 3 and 7, before and after
an unrelated corpus generation had finished.

## Root cause

```python
# macOS load average counts uninterruptible-wait threads; sum actual CPU instead.
ps = run_text(["ps", "-Ao", "pcpu"])
total = sum(float(x) for x in ps.split()[1:] if x.replace(".", "", 1).isdigit())
if total > 25.0:
    warnings.append(f"{total:.0f}% CPU busy across all processes")
```

On macOS, `ps -o pcpu` reports each process's **lifetime-average** CPU, not instantaneous
utilisation. Summing it across a machine with long uptime gives a number that has nothing to do
with current load, and that only ever grows as long-lived processes accumulate CPU time.

The three biggest contributors on the box above, from `top`:

| process | CPU time | instantaneous %CPU |
|---------|----------|--------------------|
| iTerm2 | 9h 24m | 1.0 |
| WindowServer | 4h 31m | 0.3 |
| com.apple.DriverKit | 3h 59m | 0.6 |

Instantaneously they total ~2%; by lifetime average they dominate the 30%.

This is the same class of error the comment directly above it was written to avoid — it
replaced one aggregate proxy (load average) with another (lifetime-average CPU), rather than
with an instantaneous reading.

## Second, independent false positive

```python
busy = run_text(["pgrep", "-fl", "cargo|rustc|criterion|claude"])
```

`claude` matches Claude Code's own background helper, which is present and **sleeping** on any
machine with the CLI installed:

```
9546 /Users/jky/.local/share/claude/versions/2.1.233 --chrome-native-host
```

So the check fires on every run on a developer machine regardless of load. This one is
independent of the `pcpu` bug — fixing either alone still leaves the gate tripping.

## Why it matters

Both warnings are fatal unless `--force` is passed. `--force` also bypasses the **battery**
check, which is legitimate and is one of the rules `docs/guides/benchmarking.md` calls out by
name ("Never benchmark a laptop on battery"). So the practical effect is that the useful check
gets routinely bypassed along with the broken ones, and a genuinely busy or unplugged machine
will not stop a run.

## Suggested fix

- **CPU**: read an instantaneous sample instead of a lifetime aggregate — `top -l 2 -n 0` and
  parse the *second* sample's `CPU usage: ... % idle` line (the first sample is itself a
  lifetime average and must be discarded). Compare against an idle threshold rather than a
  busy one.
- **Processes**: narrow the pattern to real build processes. `criterion|cargo|rustc` alone, or
  keep `claude` but exclude helper subprocesses (`--chrome-native-host`, `claude-wrap`), so it
  matches an agent actually running a build rather than the CLI merely being installed.
- Consider splitting `--force` so the battery check cannot be waived by the same flag that
  waives the CPU heuristics.

## Found while

Running #965's A/B on pinned hardware. The Linux branch of the same function
(`os.getloadavg()[0] > 1.0`) was not exercised and is not implicated.

newhoggy:
## Summary

- `machine_warnings()` summed `ps -o pcpu`, which macOS reports as each process's **lifetime-average** CPU, not instantaneous usage — on a long-uptime machine that number only ever grows, unrelated to current load. Reproduced on `johns-mac-mini` (M4 Pro, up 26 days): reported "30% CPU busy" while `top -l 2` on the same box reported 98.71% idle. Fixed by reading `top -l 2 -n 0` instead, discarding the first (also lifetime-average) sample and using the second's instantaneous idle percentage.
- The busy-process check matched `claude`, which fires on **every** run on any machine with Claude Code installed — its own background helper (`--chrome-native-host`) is always present, sleeping. Dropped `claude` from the pattern; `cargo|rustc|criterion` alone are a reliable signal for an actual build/benchmark in progress.
- Both warnings were fatal only via `--force`, which also waived the battery check — the one check here that's never a false positive. Split it into its own `--force-battery` flag so the useful check can't be silently bypassed along with the noisy ones.

## Test plan

- `py_compile` passes; no Python test harness exists in this repo for `scripts/*.py` (nor does CI lint them), so this was verified live rather than via a test suite.
- Live-ran `battery_warning()`/`machine_warnings()` against the actual functions on this machine: `battery_warning()` correctly reported battery power on the current (unplugged) laptop; `machine_warnings()`'s process check correctly caught a genuinely-running `cargo build` from a concurrent worktree via the narrowed `cargo|rustc|criterion` pattern (not a `claude`-only false match).
- Verified the `top -l 2 -n 0` idle-percentage regex parsing in isolation against real `top` output — confirms the second (instantaneous) sample is what gets used, not the first.
- `cargo build --features cli,simd,regex,serde` / `cargo fmt --check`: clean (no Rust source touched).

Fixes #1640

### DIFF
--- a/scripts/ab-cli.py
+++ b/scripts/ab-cli.py
@@ -23,6 +23,7 @@ import argparse
 import glob
 import hashlib
 import os
+import re
 import shutil
 import statistics
 import subprocess
@@ -58,7 +59,12 @@ def parse_args(argv=None):
     p.add_argument("--no-identity", action="store_true",
                    help="skip the output identity gate (rule 4) — for control runs")
     p.add_argument("--force", action="store_true",
-                   help="run even if the machine looks busy or is on battery")
+                   help="run even if the machine looks CPU-busy or has build/benchmark "
+                        "processes running (does NOT waive the battery check)")
+    p.add_argument("--force-battery", action="store_true",
+                   help="run even on battery power — separate from --force (#1640) so "
+                        "the one check that's never a false positive can't be waived by "
+                        "the same flag that waives the noisy heuristics")
     return p.parse_args(argv)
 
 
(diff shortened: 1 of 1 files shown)


newhoggy:
Code review found the new CPU check silently produced no warning at all if `top -l 2 -n 0` failed or its output format didn't match the expected pattern — a genuinely busy machine would read as idle with no indication why. Fixed by surfacing that as a warning (fatal unless `--force`, same as every other check here) instead of silently skipping it. Verified the failure path directly (malformed input to the regex correctly yields zero matches, triggering the new warning) and re-confirmed the happy path is unaffected on this machine.

Issue #1765, opened 2026-08-28
jq: path-context builtins (key/parent/file_index) still silently no-op inside AsPattern/Reduce/Foreach/Limit


newhoggy:
**Severity: Low** (silent wrong output, no crash; requires an uncommon-but-real query shape; no oracle exists for any of the four remaining cases)

## Summary

#1663 fixed `needs_path_context`'s missing `Expr::As` arm (`EXPR as $v | body`,
the ordinary jq variable-binding idiom — the single most consequential shape
in that issue's own filing, since it's a very common idiom). Four other
shapes named in #1663's original filing (or found during its own
investigation) remain unfixed and still silently stub `key`/`parent(n)`/
`file_index` to their zero defaults instead of resolving correctly:

- `Expr::AsPattern` — destructuring `as` (`. as [$x] | ...`, `. as {a: $x} | ...`)
- `Expr::Reduce` — `reduce EXPR as $x (init; update)`
- `Expr::Foreach` — `foreach EXPR as $x (init; update; extract)`
- `Expr::Limit` — `limit(n; expr)`

(`Expr::Object` has the identical gap but is tracked separately as #1332,
predating #1663.)

## Repro (confirmed live against `main`, post-#1663)

```console
$ echo '{"a":[1]}' | succinctly jq -c '.a | . as [$x] | key'
null                                             # should resolve, not stub
$ echo '{"a":[1,2]}' | succinctly jq -c '.a | reduce .[] as $x (key; .)'
null
$ echo '{"a":[1,2]}' | succinctly jq -c '.a | [foreach .[] as $x (key; .)]'
[null,null]
$ echo '{"a":1}' | succinctly jq -c '.a | [limit(1; key)]'
[null]
```

## Why these four are separate from #1663's own fix

Unlike `Expr::As`, none of these four have a real-yq oracle to verify a fix
against: yq's own lexer rejects destructuring `as`, `reduce`, `foreach`, and
`limit` outright (confirmed live against v4.53.3 — each is a parse error).
`key`/`parent`/`file_index` are themselves succinctly extensions with no
jq equivalent either, so there is no reference implementation for any of
these four shapes to check "correct" against — only internal consistency.

`Reduce`/`Foreach` in particular raise a real semantic question #1663's own
investigation didn't settle: `update` (and `foreach`'s `extract`) evaluate
against the **accumulator**, a synthetic value with no document position —
what should `key` mean there? The current stub (`null`) may already be the
[...]

PR #1766, opened 2026-08-28
fix(jq): give needs_path_context an Expr::As arm

### LINKED ISSUE #1663: jq: bare/unpiped path-context builtins (parent, parent(n), key, file_index) silently no-op outside Expr::Pipe
**Severity: Low-Medium** (silent wrong output, no crash; requires an uncommon-but-real query shape)

## Summary

Found during #1487's code review (multiple independent review passes converged on this). `Builtin::Parent`/`Builtin::ParentN`/`Builtin::Key`/`Builtin::FileIndex` all require path-context tracking to resolve meaningfully, which only `eval_pipe_with_path_context_internal` provides. That evaluator is only reached when `needs_path_context` recognizes the surrounding expression shape and routes through it via `Expr::Pipe`'s dispatcher.

`needs_path_context` (`src/jq/eval.rs`, ~line 561) is missing arms for several common expression shapes, so a path-context builtin used inside any of them falls through to `eval_single`'s plain dispatch → `eval_builtin`, whose arms for these builtins don't have path context available and return a fixed, wrong fallback (e.g. `ParentN`'s arm does `let _ = n_expr;` and unconditionally returns `{}`, without even evaluating `n_expr` -- so a `parent(error("boom"))` in one of these positions silently swallows the error instead of raising it).

Confirmed-missing shapes (live-verified against a `--release --features cli` build):
- `Expr::Object` (e.g. `{x: parent(-1)}`)
- `Expr::AsPattern` (`EXPR as $x | ...` -- the ordinary variable-binding idiom, very commonly used)
- `Expr::Reduce`
- `Expr::Foreach`
- `Expr::Limit`

## Repro

```
$ echo '{"a":1}' | succinctly jq 'parent(-1)'              # {} instead of erroring (bare/unpiped)
$ echo '{"a":1}' | succinctly jq 'parent(error("boom"))'   # {} -- error is never raised
$ echo '{"a":{"b":1}}' | succinctly jq '.a.b | . as $x | parent(1)'   # {} instead of {"b":1} (WRONG, not just unvalidated)
$ echo '{"a":{"b":1}}' | succinctly jq '.a.b | {x: parent(1)}'        # {"x":{}} instead of {"x":{"b":1}}
```

The `Expr::AsPattern` case is the most consequential: it's not merely a missed validation opportunity (like the bare/no-pipe cases), it's an outright wrong *value* for an extremely common jq idiom (`EXPR as $x | ...`).

## Scope note

This is a pre-existing gap in `needs_path_context`'s coverage, not something #1487 introduced -- the same builtins already returned the same wrong fallback in these shapes before that fix. #1487 was scoped narrowly to `parent(n)`'s numeric-argument classification (the Int/Float disagreement, and yq-mode's negative-n wraparound) and deliberately did not expand into fixing `needs_path_context`'s coverage, since that affects every path-context builtin (not just `parent(n)`) and carries a much larger blast radius than a single-builtin argument-validation fix.

## Suggested approach

Audit every `Expr` variant `needs_path_context` currently returns `false`/falls through to the catch-all for, and either:
- add explicit path-context-aware handling for `key`/`parent`/`parent(n)`/`file_index` inside each of the missing shapes (mirroring how `Expr::Array` already does this), or
- find a way to make the fallback in `eval_builtin` at least *fail loudly* (evaluate `n_expr` for its side effects and error rather than silently returning `{}`) when it can't provide real path context, so the failure mode changes from "silently wrong" to "clearly unsupported" until full support lands.

Related: #1409 ("isolate-then-continue_rest_with_context loses per-output path threading") is adjacent path-context-threading work but not the same gap -- worth checking for overlap when scoping the fix.

newhoggy:
## Summary

`Builtin::Parent`/`Builtin::ParentN`/`Builtin::Key`/`Builtin::FileIndex` require
path-context tracking to resolve meaningfully, only available via
`eval_pipe_with_path_context_internal`, reached only when `needs_path_context`
recognizes the surrounding `Expr` shape. `needs_path_context` had no arm for
`Expr::As` (`EXPR as $v | body` — the ordinary jq variable-binding idiom, not
some rare shape), so any of these builtins used inside it fell through to
`eval_single`'s plain dispatch and silently returned a fixed, wrong fallback
(`{}` for `parent(n)`, `null` for `key`) instead of the correct value.

This is the single most consequential case in #1663's own filing:

```console
$ echo '{"a":{"b":1}}' | succinctly jq -c '.a.b | . as $x | parent(1)'
{}                    # before this fix
{"b":1}                # after — matches real yq's identical `. as $v | parent(1)`
```

(#1663's own text classified this repro as `Expr::AsPattern`, but only
*destructuring* `as` produces `AsPattern` — `. as $x | ...` is `Expr::As`,
confirmed while investigating. `AsPattern`/`Reduce`/`Foreach`/`Limit` still
have the same underlying gap and are tracked as a follow-up, #1765 —
`Expr::Object` has the identical gap but predates this issue as #1332.)

## Approach

Neither `expr` (the bind expression) nor `body` navigate away from the
caller's own position — the pre-existing `eval_as` already evaluates both
against the same `value` — so both route through
`eval_pipe_with_path_context_internal` unchanged, mirroring the existing
`FuncDef` arm.

Unlike #1302's `Array` arm or #1334's `StringInterpolation` arm, `as` is
deliberately **non-atomic** under `?` — `eval_as`'s own doc comment
establishes `(1,2,error("x")) as $v | $v + 10` keeps `11`, `12` as a real
prefix, unlike `[...]`'s atomicity — so `optional` is passed straight
through to both evaluations rather than forced to `false`, and the final
combined result goes through `continue_rest_with_context` (preserving
position), not `continue_rest_with_fresh_root`.

Verified against every control-flow shape (`None`/`Error`/`Break`/`Halt`,
both bare and prefix-preserving, for both the bind expression and the body)
against the pre-existing `eval_as` path as ground truth, and against real
`yq` v4.53.3 for every value-correctness case (`key`/`parent(n)` are
succinctly extensions with no jq equivalent, so yq mode is the actual
oracle here, not jq's own).

Closes #1663.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite green — two
      separate runs each hit one unrelated `signal 9`-killed test under
      heavy concurrent-session load on this machine, both confirmed to
      pass in isolation and unrelated to this change)
- [x] All 37 pre-existing `path_context` tests pass unchanged (no
      regression to the #1302/#1306/#1334/#1405/#1409 precedents this fix
      builds on)
- [x] Live-verified against real `yq` v4.53.3 for every value-correctness
      claim above
- [x] Regression tests: happy paths (including deep `parent(n)`, nested
      `as`, position-preserving `rest` continuation), the bind-expression's
      full non-`Owned`/`ManyOwned` match (prefix-preserving, not atomic —
      the key semantic difference from #1302/#1334), the body-loop's own
      identical match (a second, distinct set of arms), and a pinned
      "known gap" test for the four shapes this PR deliberately leaves
      unfixed (tracked as #1765)
- [x] `cargo llvm-cov` + `omni-dev coverage diff` — 96.55% patch coverage
      (56/58 new lines); the 2 remaining lines are both `unreachable!()`
      arms, matching the exact same accepted, always-uncovered pattern the
      pre-existing `Array`/`StringInterpolation` arms already have

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -763,6 +763,13 @@ pub(crate) fn needs_path_context(expr: &Expr) -> bool {
         // as a known gap rather than paying for a full
         // `expand_func_calls` just to answer this routing question.
         Expr::FuncDef { body, then, .. } => needs_path_context(body) || needs_path_context(then),
+        // `EXPR as $v | body` (#1663): neither `expr` nor `body` navigate
+        // away from the caller's own position -- `eval_as` evaluates both
+        // against the same `value` -- so a `key`/`parent`/`file_index`
+        // inside either one needs the same ambient context the caller has.
+        // This is the single most common trigger for this whole bug class:
+        // `.a.b | . as $x | parent(1)` previously silently stubbed to `{}`.
+        Expr::As { expr, body, .. } => needs_path_context(expr) || needs_path_context(body),
         _ => false,
     }
 }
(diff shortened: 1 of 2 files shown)


newhoggy:
Ran a 5-agent \`/code-review\` pass (high effort, given this touches the evaluator's core dispatch, shared across jq and yq modes). Summary of what changed as a result:

**Fixed (real, actionable):** the initial \`Expr::As\` arm hand-rolled two pieces of logic this codebase already has shared helpers for, specifically to prevent this exact drift — \`materialize_bound_values\` (#1462) and \`accumulate_path_context_step\` (#715 follow-up, whose own doc comment records that six prior near-copies of this loop shape each independently dropped a \`Break\`/\`Partial\`'s already-produced output before that fix). Refactored to reuse both, which also removed two hand-written \`unreachable!()\` blocks — \`accumulate_path_context_step\` already correctly converts the \`One\`/\`Many\` variants \`eval_fanout\` (used by \`select\`) can hand back even from an all-owned-value call site, rather than asserting they can't happen.

**Verified, no action needed:**
- Removed-behavior audit: traced every \`QueryResult\` variant combination against the pre-existing \`eval_as\` as ground truth — no drops, reorders, or double-counts found.
- Cross-file tracer: confirmed \`Expr::AsPattern\` (destructuring \`as\`) now shows a visible asymmetry against the fixed \`Expr::As\` — exactly the gap already scoped out and tracked as #1765, not a new finding.
- Conventions: an unverified-claim flag on issue #1765's own text turned out to be a false positive once live-checked (real yq's lexer does reject \`reduce\`/\`foreach\`/\`limit\`/destructuring \`as\`, confirmed against v4.53.3) — added the missing verification citation to #1765 directly.
- Efficiency: a minor O(N²) guard re-evaluation for deeply nested \`as\`-chains was flagged, but it rides on top of \`substitute_bound_var\`'s own already-quadratic pre-existing cost (same complexity class, not a regression) — not pursued given the marginal return.

Full test suite green (all 41 \`path_context\` tests, including the 4 new ones, unchanged after the refactor — confirming it's behavior-preserving), both clippy legs clean, patch coverage 97.62% (41/42 — the one remaining line is a provably-unreachable \`unreachable!()\` arm, matching the same accepted pattern the pre-existing \`Array\`/\`StringInterpolation\` arms already have).

Issue #1701, opened 2026-08-27
yq: M2 fast path ignores -0/--nul-output, always terminates records with \n


newhoggy:
**Severity: Low** (cosmetic output-shape divergence; no data loss, no crash)

## Summary

Found by \`/code-review\`'s cross-file trace on PR #1698 (#1693), while sweeping the same
gate-defining code that PR edits. Confirmed live on unmodified \`main\` — pre-existing, not
introduced by that PR.

\`succinctly yq -0\`/\`--nul-output\` (NUL-terminate records instead of newline) is silently
ignored whenever the M2 streaming fast path is taken, for both stdout and \`--inplace\`,
JSON and YAML output alike.

## Repro (verified on \`main\`)

\`\`\`console
$ printf 'a: 1\n' > f.yaml
$ succinctly yq -0 '.' f.yaml | xxd | tail -1
00000000: 613a 2031 0a                             a: 1.        # ends in 0a (\n), not 00
\`\`\`

Same for \`.[]\` iteration and \`-o json\` output.

## Root cause

The \`stream_cursor!\` macro (\`src/bin/succinctly/yq_runner.rs\`) hardcodes \`\n\` as the record
terminator for every M2 fast-path output, never consulting \`output_config.nul_output\`.
\`write_terminator\`, which does honor \`nul_output\`, is only reachable from the DOM path's
\`output_value\`.

## Suggested fix direction

Not attempted. Thread \`output_config.nul_output\` into \`stream_cursor!\` (or a new parameter
on the streaming call sites) so the M2 path picks \`\0\` vs. \`\n\` the same way \`write_terminator\`
does on the DOM path, rather than falling back to DOM (which would reopen the same
duplicate-key-collapse tradeoff #1693/#1700 already document for \`--ascii-output\`).

## Related

- #1693 / PR #1698 — the PR whose review surfaced this while auditing the same gate family.

PR #1767, opened 2026-08-28
fix(yq): reject a NUL-output result whose value contains a NUL byte

### LINKED ISSUE #1709: yq: -0/--nul-output performs no content validation, unlike real yq
**Severity: Low** (pre-existing gap, newly reachable via a fixed code path)

## Summary

Found by \`/code-review\` on PR #1707 (#1701). Real yq (pinned Homebrew v4.53.3) refuses
\`-0\`/\`--nul-output\` output when a scalar's own value contains an embedded NUL character;
succinctly silently emits it inline, making the record boundary ambiguous to any
NUL-delimited consumer (e.g. \`xargs -0\`).

## Repro (verified against the pinned oracle)

\`\`\`console
$ printf 'a: "hello\\u0000world"\n' | yq -0 '.a'
Error: can't serialise value because it contains NUL char and you are using NUL separated output

$ printf 'a: "hello\\u0000world"\n' | succinctly yq -0 '.a'
hello\0world\0                          # exits 0 -- indistinguishable from two records
\`\`\`

## Scope note

This gap is **pre-existing on the DOM path** (reproducible on unmodified \`main\` via
\`-P\`/\`--null-input\`/any flag that forces DOM), not introduced by #1701. #1701's own fix
(centralizing terminator writes into a \`Terminator\` type, \`src/bin/succinctly/yq_runner.rs\`)
is what makes \`-0\` functional on the M2 fast path for the first time, so the fast path now
inherits this pre-existing unsoundness too — previously \`-0\` was simply a no-op there
(always emitting \`\\n\` regardless), so this specific failure mode couldn't be reached via
that path at all.

## Suggested fix direction

Not attempted. Needs a NUL-byte scan over each value's own rendered content before/while
writing it (likely in \`Terminator::write_io\`/\`write_fmt\`'s callers, or in the streamers
themselves), erroring the same way real yq does rather than silently emitting an ambiguous
byte stream.

## Related

- #1701 / PR #1707 — the fix that made \`-0\` reachable on the M2 fast path, surfacing this
  pre-existing gap there for the first time.

newhoggy:
## Summary

Found by `/code-review` on PR #1707 (#1701). Real yq (pinned Homebrew v4.53.3) refuses `-0`/`--nul-output` output when a scalar's own value contains an embedded NUL character; succinctly silently emitted it inline, making the record boundary ambiguous to any NUL-delimited consumer (e.g. `xargs -0`).

Repro (verified against the pinned oracle): a YAML document with a scalar containing a NUL escape (hex 00) under `-0` -- real yq exits 1 with "can't serialise value because it contains NUL char and you are using NUL separated output"; succinctly used to exit 0, emitting the raw byte inline (indistinguishable from two records).

**Scope**: YAML output only. JSON output always escapes a NUL as a six-character sequence rather than emitting it raw, so there's no ambiguity there -- live-verified against the oracle that real yq accepts the same value under `-o=json` (exit 0). The identity path (`.`, no evaluation) also needed no change: it re-emits a scalar's source spelling verbatim rather than decoding it, so an escaped NUL stays printable characters, never becoming a raw NUL byte -- real yq agrees there too.

**Mechanism**: `stream_maybe_colored` now also buffers (not just for `--color`) whenever the real terminator is `Nul`, so the fully-rendered result can be scanned for an embedded NUL before any of it reaches the writer -- reusing the same buffering machinery #1708 built for coloring, extended to a second reason to need it. JSON call sites pass `Terminator::None` to skip the check entirely (it can never apply there, and buffering for nothing would be pure cost). `insert_terminators_at_boundaries` is the new non-colored counterpart to `colorize_yaml`'s own boundary handling, needed now that the non-colored path can also buffer.

## Test plan

- 5 new regression tests in `tests/yq_cli_tests.rs`, each verified against the pinned oracle first: rejects an embedded NUL under `-0` (evaluated path), accepts it under the default newline terminator, accepts it under `-o=json`, rejects it under `-C -0` (colored path), and confirms the identity path emits exactly one NUL byte total (the terminator itself, not a second embedded one).
- Patch coverage: 87.88% (29/33 new lines). The 4 uncovered lines are pre-existing defensive-fallback `stream_maybe_colored` call sites (documented "rarely if ever reached") that only changed by gaining one more argument; not new logic.

Fixes #1709

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -121,28 +121,72 @@ impl<W: Write> ColorSink<'_, W> {
     }
 }
 
-/// Streams through `render` directly when `use_color` is false. When true,
-/// renders into a buffer instead (still via the duplicate-key-safe cursor
-/// streamer passed in as `render`), then runs the buffer through `colorize`
-/// before writing the colorized result (#748).
+/// Streams through `render` directly when neither buffering condition below
+/// applies. Buffers into a single in-memory string first when `use_color`
+/// (so `colorize` can re-lex the whole rendered result, #748) or when
+/// `terminator` is [`Terminator::Nul`] (so the buffer can be scanned for an
+/// embedded NUL byte before any of it reaches `writer`, #1709) -- YAML's raw
+/// scalar rendering can contain a literal NUL from the source document,
+/// unlike JSON output, which always escapes one as the six-character
+/// sequence backslash-u-0-0-0-0; pass `Terminator::None` at a JSON call
+/// site to skip this check, since it can never apply there and buffering
+/// the whole result for nothing would be pure cost.
 fn stream_maybe_colored<W: Write, T>(
     writer: &mut W,
     use_color: bool,
+    terminator: Terminator,
     colorize: impl FnOnce(&str, &[usize]) -> String,
     render: impl FnOnce(&mut ColorSink<'_, W>) -> Result<T, core::fmt::Error>,
 ) -> anyhow::Result<T> {
-    if use_color {
+    if use_color || matches!(terminator, Terminator::Nul) {
         let mut sink = ColorSink::Buffered(String::new(), Vec::new());
         let value = render(&mut sink).map_err(|_| anyhow::anyhow!("Write error"))?;
         let ColorSink::Buffered(buf, boundaries) = sink else {
             unreachable!("stream_maybe_colored always constructs ColorSink::Buffered here")
         };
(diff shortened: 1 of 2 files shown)


newhoggy:
Closing without merging — the multi-agent `/code-review` pass (4 independent angles) found this implementation has real regressions, not just incomplete coverage. Full writeup going on #1709 itself so a future attempt has the complete picture. Reverting the board back to Ready rather than Done.

PR #1777, opened 2026-08-28
test(ci): run bench-runner-gated tests in CI, fix a cargo-run clobber hazard

### LINKED ISSUE #1705: test: bench-runner-gated integration tests never run in CI, and a cargo-run clobber hazard hides behind that
**Severity: Low** (test-infrastructure gap; the underlying production code these tests cover is otherwise well-tested)

## Summary

While verifying #1696 (PR #1703), running the local test suite with
`--features cli,simd,regex,serde,bench-runner` together (to actually exercise
`bench_runner`-gated code) surfaced two related gaps:

## 1. `orchestrate_cli_tests.rs` (and other `bench-runner`-gated integration
tests) never execute in CI at all

`.github/workflows/ci.yml`'s `cargo test` steps never pass `--features
bench-runner` (only `cli`, `cli,simd,regex,serde,...` etc. — grep confirms
`bench-runner` appears only in the three `cargo clippy` invocations, never in
a `cargo test` one). `tests/orchestrate_cli_tests.rs` is
`#![cfg(feature = "bench-runner")]`, so it silently never compiles, let alone
runs, in CI. It only runs if a developer manually adds `bench-runner` to a
local `cargo test` invocation.

## 2. When it *is* run, a pre-existing `cargo run` clobber hazard breaks it

Running the full suite with `bench-runner` added reproducibly fails
`orchestrate_cli_tests.rs` with `error: unrecognized subcommand 'bench'`,
100% reproducibly, but only inside a full-suite run — never when run in
isolation (`cargo test --test orchestrate_cli_tests` alone always passes).

Root cause: `tests/jq_cli_tests.rs` (and `cli_golden_tests.rs`,
`cli_characterization_tests.rs`, `deep_nesting_valid_tests.rs`,
`json_validate_tests.rs`) invoke `Command::new("cargo").args(["run",
"--quiet", "--features", "cli", "--bin", "succinctly", ...])` directly as
part of their own test bodies — a *narrower* feature set than the outer
`cargo test --features cli,simd,regex,serde,bench-runner` invocation. Since
`jq_cli_tests.rs` runs alphabetically before `orchestrate_cli_tests.rs` in a
single `cargo test` run, its `cargo run --features cli` subprocess triggers
Cargo to rebuild+relink `target/debug/succinctly` with the narrower feature
set (dropping `bench-runner`), clobbering the exact binary path
`orchestrate_cli_tests.rs` relies on via `env!("CARGO_BIN_EXE_succinctly")`
(baked in at that test crate's compile time, pointing at the same shared
`target/debug/succinctly` path). By the time `orchestrate_cli_tests.rs` runs,
the binary on disk no longer has the `bench` subcommand.

Confirmed via `.github/workflows/ci.yml` grep: this exact combination (a
`bench-runner`-gated integration test alongside a `cargo run --features cli`
test) is never exercised in CI, which is presumably why nobody has hit this
yet.

## Why this is separate from #1696

#1696/PR #1703 is a production-code bug fix in `ssh.rs`; this is a
test-infrastructure gap discovered incidentally while verifying that fix
locally. Fixing it isn't required for #1696 and would be unrelated scope
creep on that PR.

## Suggested fix direction

- Add a CI job/step that runs `cargo test --features
  cli,simd,regex,serde,bench-runner --test orchestrate_cli_tests` (or the
  full bench-runner-gated set) so this code path is actually covered.
- Separately, `jq_cli_tests.rs`'s (and siblings') `cargo run --features cli`
  helper should pass the *same* feature set the outer test invocation used
  (or use a distinct target dir / `CARGO_TARGET_DIR` override) rather than a
  hardcoded narrower one, so it can't clobber a shared binary artifact
  another test target depends on. `tests/common/cargo_run_exit.rs` might be
  a reasonable home for a shared helper that reads the ambient feature
  configuration instead of hardcoding `"cli"`.

## Related, minor (bundled here rather than as a third issue)

`.omni-dev/scopes.yaml`'s `bench` scope only lists `file_patterns:
["benches/**", "bench-compare/**"]`, but 7+ established commits (e.g.
`74b9a4e36`, `cf3623e73`, `0ffd3eb22`) use `fix(bench)`/`feat(bench)` for
`src/bin/succinctly/bench_runner/**` — a path that per the file's own rules
actually matches the `cli` scope's `src/bin/**` pattern instead. Either the
`bench` scope's `file_patterns` should be extended to include
`src/bin/succinctly/bench_runner/**`, or future commits touching that path
should switch to `cli` to match the documented rule. Not fixed here since
it's a one-line docs/config change orthogonal to the two gaps above and
unrelated to #1696's actual scope.

newhoggy:
## Summary

While verifying #1696 (PR #1703), running the local test suite with `--features cli,simd,regex,serde,bench-runner` together surfaced two related gaps:

**1. `orchestrate_cli_tests.rs` (and other `bench-runner`-gated integration tests) never execute in CI at all.** `.github/workflows/ci.yml`'s `cargo test` steps never pass `--features bench-runner` (only in the three `cargo clippy` invocations, never a `cargo test` one). `tests/orchestrate_cli_tests.rs` is `#![cfg(feature = "bench-runner")]`, so it silently never compiles, let alone runs, in CI.

**2. When it *is* run, a `cargo run` clobber hazard breaks it.** `tests/jq_cli_tests.rs` (and `cli_golden_tests.rs`, `cli_characterization_tests.rs`, `deep_nesting_valid_tests.rs`, `json_validate_tests.rs`) invoke `Command::new("cargo").args(["run", "--quiet", "--features", "cli", ...])` directly as part of their own test bodies -- a narrower feature set than the outer `cargo test --features cli,simd,regex,serde,bench-runner` invocation. Since `jq_cli_tests.rs` runs alphabetically before `orchestrate_cli_tests.rs`, its `cargo run --features cli` subprocess rebuilds and relinks the shared `target/debug/succinctly` binary with the narrower set, clobbering the `bench` subcommand out from under `orchestrate_cli_tests.rs`'s own `env!("CARGO_BIN_EXE_succinctly")`.

## Fix

- Added a dedicated "Run bench-runner-gated tests" step (`cargo test --features cli,bench-runner --test orchestrate_cli_tests`) to all four Test jobs (x86_64, ARM64, macOS ARM64, beta/nightly matrix) -- scoped narrowly to just that one test target, matching the issue's own suggested direction, rather than a broad wildcard that would also pull in unrelated snapshot tests this repo has never run under `bench-runner`.
- Added `cargo_run_features()` to the shared `tests/common/cargo_run_exit.rs` module (already included by all 5 affected files): `cfg!(feature = "bench-runner")` reads whether the *current* compiled test binary needs it, so the inner `cargo run` always matches what compiled the outer `cargo test` invocation. Replaced the ~38 hardcoded `"--features", "cli"` call sites across the 5 files with it.

## Not included

The issue's own "related, minor" bonus item (`.omni-dev/scopes.yaml`'s `bench` scope `file_patterns` not covering `src/bin/succinctly/bench_runner/**`) was explicitly deferred by the issue itself ("not fixed here since it's a one-line docs/config change orthogonal to the two gaps above") -- left out of this PR to stay on scope.

## Test plan

- Reproduced the original clobber bug locally (`cargo test --features cli,simd,regex,serde,bench-runner` on unmodified `main`: `orchestrate_cli_tests.rs` fails with `error: unrecognized subcommand 'bench'`), then confirmed the fix resolves it (same command post-fix: 0 failures, no clobber error).
- Verified the exact CI-scoped invocation (`cargo test --features cli,bench-runner --test orchestrate_cli_tests`) passes cleanly: 5/5 tests pass.
- Full local `cargo test --features cli,simd,regex,serde` (CI's normal, non-bench-runner invocation): 0 failures -- confirms the fix doesn't change behavior for the common case.
- `cargo test --features cli,simd,regex,serde,bench-runner --tests` (whole suite compiles with the new feature): clean.
- `python3 -c "import yaml; yaml.safe_load(...)"` on the modified `ci.yml`: valid.
- No coverage delta: this PR touches only `tests/*.rs` and CI YAML, neither in `cargo llvm-cov`'s scanned scope.

Fixes #1705

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,6 +150,17 @@ jobs:
         if: needs.gate.outputs.code == 'true'
         run: cargo test --features cli --test yq_cli_tests --test yq_golden_tests --test yq_path_consistency_tests --test jq_golden_tests --test yaml_bench_suite_coverage --test dsv_cli_tests --test text_cli_tests --test json_validate_tests --test yaml_validate_tests
 
+      # `tests/orchestrate_cli_tests.rs` is `#![cfg(feature = "bench-runner")]`
+      # and was never previously built or run by any `cargo test` step here
+      # (#1705) -- run in its own step, isolated from the `cli`-only tests
+      # above (whose `cargo run`-spawned CLI subprocesses would otherwise
+      # re-link the shared `target/debug/succinctly` binary with a narrower
+      # feature set, dropping the `bench` subcommand out from under this
+      # test target if it ran afterward in the same invocation).
+      - name: Run bench-runner-gated tests
+        if: needs.gate.outputs.code == 'true'
+        run: cargo test --features cli,bench-runner --test orchestrate_cli_tests
+
       - name: Run tests
         if: needs.gate.outputs.code == 'true'
         run: cargo test --verbose
@@ -245,6 +256,10 @@ jobs:
       - name: Run cli-gated tests
         if: needs.gate.outputs.code == 'true'
         run: cargo test --features cli --test yq_cli_tests --test yq_golden_tests --test yq_path_consistency_tests --test jq_golden_tests --test yaml_bench_suite_coverage --test dsv_cli_tests --test text_cli_tests --test json_validate_tests --test yaml_validate_tests
+
+      - name: Run bench-runner-gated tests
+        if: needs.gate.outputs.code == 'true'
+        run: cargo test --features cli,bench-runner --test orchestrate_cli_tests
 
       - name: Run tests
         if: needs.gate.outputs.code == 'true'
@@ -344,6 +359,10 @@ jobs:
         if: needs.gate.outputs.code == 'true'
         run: cargo test --features cli --test yq_cli_tests --test yq_golden_tests --test yq_path_consistency_tests --test jq_golden_tests --test yaml_bench_suite_coverage --test dsv_cli_tests --test text_cli_tests --test json_validate_tests --test yaml_validate_tests
 
+      - name: Run bench-runner-gated tests
+        if: needs.gate.outputs.code == 'true'
+        run: cargo test --features cli,bench-runner --test orchestrate_cli_tests
+
       - name: Run tests
         if: needs.gate.outputs.code == 'true'
         run: cargo test --verbose
@@ -405,6 +424,9 @@ jobs:
 
       - name: Run cli-gated tests
         run: cargo test --features cli --test yq_cli_tests --test yq_golden_tests --test yq_path_consistency_tests --test yaml_bench_suite_coverage --test dsv_cli_tests --test text_cli_tests --test json_validate_tests --test yaml_validate_tests
+
+      - name: Run bench-runner-gated tests
+        run: cargo test --features cli,bench-runner --test orchestrate_cli_tests
 
       - name: Run tests
         run: cargo test --verbose
--- a/tests/cli_characterization_tests.rs
+++ b/tests/cli_characterization_tests.rs
@@ -26,7 +26,7 @@ use anyhow::Result;
 
 #[path = "common/cargo_run_exit.rs"]
 mod cargo_run_exit;
-use cargo_run_exit::{classify_cargo_run_exit, MAX_CARGO_RETRIES};
+use cargo_run_exit::{cargo_run_features, classify_cargo_run_exit, MAX_CARGO_RETRIES};
 
 /// Run `succinctly <args>` with `stdin` piped in, returning captured stdout.
 ///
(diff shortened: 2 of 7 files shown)


newhoggy:
Two `/code-review` angles reported back (no correctness bugs found, confirmed all 4 CI jobs correctly wired, no missed cargo-run call sites, no divergence risk in the cfg!()-based detection). Applied one legitimate simplification found: `cargo_run_features()` returned `"cli,bench-runner"`, but `Cargo.toml`'s `bench-runner = ["cli", ...]` already pulls `cli` in, so naming both was redundant — now returns just `"bench-runner"`.

A couple of other observations from review didn't warrant a code change: the CI step and the `cargo_run_features()` rollout genuinely fix two different scenarios (the CI-scoped isolation fix closes the gap in CI itself; the shared-helper fix additionally protects a developer running the full suite locally with `bench-runner` and no `--test` filter) — both are real value, not redundant with each other. And a suggested step-reordering for CI efficiency wasn't applied since Cargo caches build artifacts per feature-fingerprint regardless of step order, so the claimed savings aren't well-substantiated without measurement.

All local gates re-verified green after the simplification: reproduced the original clobber bug against unmodified `main`, confirmed the fix resolves it, full test suite (0 failures — one unrelated pre-existing snapshot staleness in `cli_golden_tests.rs`, itself never run in CI today, confirmed out of scope), both clippy legs, doc, fmt.

PR #1785, opened 2026-08-28
refactor(cli): hoist Terminator into a shared output module

### LINKED ISSUE #1711: refactor: yq_runner's Terminator and jq_runner's write_terminator duplicate the same nul/join/newline decision
**Severity: Low** (maintenance/DRY concern, not a functional bug)

## Summary

Found by \`/code-review\` on PR #1707 (#1701). #1701 centralized yq mode's "which terminator
follows a written value" three-way decision (NUL / newline / nothing, per \`-0\`/\`--join-output\`)
into a \`Terminator\` enum in \`src/bin/succinctly/yq_runner.rs\`. \`src/bin/succinctly/jq_runner.rs\`
has its own, independent \`write_terminator\` function (around line 3480) reimplementing the
identical logic for jq mode, using a differently-named field (\`raw_output0\` vs \`nul_output\`)
and inlining an extra \`unbuffered\` flush. Neither file's copy was hoisted into a shared
location (e.g. \`output.rs\`, already imported by both binaries).

## Why this is currently safe, and why it's still worth flagging

jq's own identity fast path (\`can_use_raw_identity()\`) already excludes \`raw_output\`/
\`join_output\` from taking that path at all (\`OutputConfig::from_args\` forces
\`raw_output = true\` whenever either is set) — so jq mode isn't currently exposed to the bug
class #1701/#1699 fixed for yq mode's M2 fast path. But that safety is incidental to a
*different, broader* gate, not because the two files share a fix. A future change that
narrows jq's fast-path gate (mirroring the kind of change #1701 made to yq's own
\`stream_cursor!\`) would have to rediscover and re-fix this exact bug class independently in
\`jq_runner.rs\`, since there's no shared type tying the two decisions together.

## Suggested fix direction

Not attempted. Hoist \`Terminator\` (or an equivalent) into \`output.rs\` (or another module
both \`jq_runner.rs\` and \`yq_runner.rs\` already import from), and have both binaries' own
\`write_terminator\`-style functions delegate to it.

## Related

- #1699, #1701 / PR #1707 — the two yq-mode fixes that established the pattern this issue
  proposes sharing with jq mode.

newhoggy:
## Summary

Found by `/code-review` on PR #1707 (#1701). #1701 centralized yq mode's "which
terminator follows a written value" three-way decision (NUL / newline / nothing, per
`-0`/`--join-output`) into a `Terminator` enum in `yq_runner.rs`. `jq_runner.rs` had its
own, independent `write_terminator` function reimplementing the identical logic for jq
mode, using a differently-named field (`raw_output0` vs `nul_output`) and inlining an
extra `unbuffered` flush. Neither file's copy was hoisted into a shared location.

## Change

Moved `Terminator` (the enum plus its `write_io`/`write_fmt` methods) into `output.rs`,
which both binaries already import from. Since the two `OutputConfig` structs are
different types with different field names, the shared type takes two `bool`s
(`Terminator::new(nul, join)`) rather than either `OutputConfig` directly:

- `yq_runner.rs`: removed its local `enum Terminator`/`impl Terminator` entirely; kept a
  thin `terminator_from_config(config: &OutputConfig) -> Terminator` wrapper reading
  `nul_output`/`join_output`, used at all 8 existing call sites (the DOM path's
  `write_terminator`, plus the M2 fast path's `stream_cursor!` macro and doc-separator
  writes #1701 added this type for in the first place).
- `jq_runner.rs`: `write_terminator` now delegates the NUL/newline/join decision to
  `output::Terminator::new(config.raw_output0, config.join_output).write_io(out)`,
  keeping its own `--unbuffered` flush local (no yq-mode equivalent at this call site).

Pure refactor, no behavior change intended.

## Test plan

- [x] Manually verified terminator behavior for both binaries: `jq --raw-output0`,
      `jq -j`, `jq` default; `yq -0`, `yq -j`, `yq` default; `jq --raw-output0 -j`
      (NUL wins over join, matching real jq's own order-independence) -- all byte-exact
      to pre-change output
- [x] Full `cargo test --features cli,simd,regex,serde` -- 55/55 test binaries pass,
      2950+204+... lib/integration tests, 0 failed (run under `cargo-guard.sh` after an
      earlier attempt silently stalled -- see below)
- [x] `cargo doc --no-deps` -- clean, no intra-doc-link errors
- [x] Rebased cleanly onto main after a large batch of concurrent merges landed while
      this branch sat idle; re-verified the diff is textually and semantically intact
      post-rebase, then re-ran the full build/clippy/test suite against the new base

Fixes #1711

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -3579,7 +3579,7 @@ fn write_output<W: Write>(out: &mut W, value: &OwnedValue, config: &OutputConfig
 /// own `--unbuffered` flush has no yq-mode equivalent at this call site,
 /// so it stays local rather than folding into the shared type.
 fn write_terminator<W: Write>(out: &mut W, config: &OutputConfig) -> Result<()> {
-    Terminator::new(config.raw_output0, config.join_output).write_io(out)?;
+    Terminator::from_flags(config.raw_output0, config.join_output).write_io(out)?;
     if config.unbuffered {
         out.flush()?;
     }
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -27,7 +27,7 @@ use succinctly::json::JsonIndex;
 use super::JqCommand;
 use crate::output::{
     self, escape_json_string, escape_json_string_ascii, exit_codes, flush_then_err, ColorScheme,
-    ControlEscape, DiagStyle, ErrorSink, FloatStyle, InputLocation, JsonFormatOpts,
+    ControlEscape, DiagStyle, ErrorSink, FloatStyle, InputLocation, JsonFormatOpts, Terminator,
 };
 
 /// Evaluation context for passing variables to the jq evaluator.
@@ -3573,13 +3573,13 @@ fn write_output<W: Write>(out: &mut W, value: &OwnedValue, config: &OutputConfig
     Ok(())
 }
 
-/// Write the appropriate line terminator based on config.
+/// Write the appropriate line terminator based on config. The NUL/newline/
+/// join three-way choice itself is shared with yq_runner.rs's own
+/// `terminator_from_config` via `output::Terminator` (#1711) -- jq mode's
+/// own `--unbuffered` flush has no yq-mode equivalent at this call site,
+/// so it stays local rather than folding into the shared type.
 fn write_terminator<W: Write>(out: &mut W, config: &OutputConfig) -> Result<()> {
-    if config.raw_output0 {
-        out.write_all(&[0])?; // NUL byte
-    } else if !config.join_output {
-        out.write_all(b"\n")?;
-    }
+    Terminator::new(config.raw_output0, config.join_output).write_io(out)?;
     if config.unbuffered {
         out.flush()?;
     }
--- a/src/bin/succinctly/output.rs
+++ b/src/bin/succinctly/output.rs
@@ -54,6 +54,56 @@ pub fn json_bytes_to_owned_value(bytes: &[u8]) -> Result<OwnedValue, EvalError> 
     generic_to_owned(&cursor.value())
 }
 
+/// Which terminator follows a written value: NUL (`-0`/`--nul-output`),
+/// a bare newline (the default), or nothing (`-j`/`--join-output`). `nul`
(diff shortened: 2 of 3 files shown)


newhoggy:
Multi-round \`/code-review\` (10+ finder angles across two passes) converged on a few concrete, worth-fixing points, all addressed in a follow-up commit:

- \`Terminator::new(bool, bool)\` reintroduced the exact same-typed-positional-argument transposition risk the old \`from_config(&OutputConfig)\` design existed to prevent (independently flagged by 3+ reviewer angles). Renamed to \`from_flags\` to match the codebase's existing \`ColorChoice::from_flags\` idiom for this exact shape, added \`Debug\`/\`PartialEq\`/\`Eq\`, and added a precedence-table unit test mirroring that type's own test.
- The hoist accidentally routed jq_runner.rs's per-value newline write through \`writeln!()\` (fmt machinery) instead of its original \`write_all(b"\n")\` -- reverted, keeping the zero-formatting-overhead shape on jq's own hot output path. (One reviewer angle empirically confirmed this specific difference doesn't actually show up in codegen on the current toolchain thanks to \`format_args!\`'s zero-arg fast path, but there's no reason to introduce the detour when the original bytes-only call already existed.)
- \`yq_runner.rs\`'s wrapper doc comment had grown into a near-duplicate of \`output.rs\`'s own (now canonical) precedence-rule doc -- trimmed to just its own reason for existing.
- \`write_fmt\`'s doc comment named yq's M2 path specifically despite the method now living in code jq_runner.rs also imports -- generalized.
- \`output.rs\`'s module doc didn't mention \`Terminator\` -- added.

**Considered and not adopted**: a full trait-based per-mode config adapter (mirroring \`EvalSemantics\`) for the "config struct -> two bools" step. That step is 1-8 call sites of a 2-line conversion; a trait+impls would cost more code than the duplication it removes, and CLAUDE.md's \`EvalSemantics\` guidance is specifically about jq/yq *behavioral* divergence, not this kind of field-name mismatch. The existing per-runner thin wrappers (inline in jq_runner.rs's single call site, a named function in yq_runner.rs's eight) stay.

Full test suite re-verified clean after each fix (one run hit 2 spurious failures from concurrent-session contention -- "Blocking waiting for file lock on package cache" plus a SIGKILL -- confirmed non-reproducing in isolation and on a clean re-run).

PR #1786, opened 2026-08-29
fix(jq): resolve key/parent/file_index through single-pattern AsPattern

newhoggy:
## Summary

#1663 fixed `needs_path_context`'s missing `Expr::As` arm (the ordinary
`EXPR as $v | body` idiom), the single most consequential shape in that
issue's filing. Four other shapes it named remain unfixed, still silently
stubbing `key`/`parent(n)`/`file_index` to their zero defaults — tracked as
#1765. This PR takes the first of those four, per the issue's own suggested
value/cost ordering: `Expr::AsPattern`'s single-pattern case (no `?//`),
which mirrors #1663's fix most directly.

```
$ echo '{"a":[1]}' | succinctly jq -c '.a | . as [$x] | key'
null                                             # before: should be "a"
$ echo '{"a":[1]}' | succinctly-fixed jq -c '.a | . as [$x] | key'
"a"                                              # after
```

## Fix

Adds a dedicated `Expr::AsPattern` arm to `eval_pipe_with_path_context_internal`,
structurally identical to the existing `Expr::As` arm just above it: bind
`expr`'s outputs through the path-context evaluator, extract each pattern's
bindings via `extract_pattern_bindings(&patterns[0], &bound_val, false)`
(`invert: false`, matching the single-pattern/non-`?//` contract that
function's own doc comment establishes), substitute into `body` via the
existing `substitute_vars`, and evaluate each substituted body through the
same path-context evaluator — so `key`/`parent`/`file_index` inside `body`
see the caller's own ambient position, unchanged.

`needs_path_context`'s own `Expr::AsPattern` arm mirrors `Expr::As`'s too,
without a pattern-count gate — that's just the routing question ("does the
whole pipe need path-context evaluation at all"), so a `?//`-chain still
routes into path-context evaluation when needed, then correctly falls
through to the generic (pre-existing, unchanged, stub) fallback once inside,
rather than hitting the new dedicated arm.

## Scope: deliberately not `Reduce`/`Foreach`/`Limit`/`?//`

Per the issue's own scoping note — a `?//`-chain's retry-on-failure
semantics (`try_pattern_alternatives`'s decode-failure/break/halt
classification, carried-prefix-across-alternatives bookkeeping) are real
complexity this PR doesn't attempt to replicate. Confirmed live this isn't a
regression, just unchanged:

```
$ echo '{"a":[1]}' | succinctly-fixed jq -c '.a | . as [$x] ?// {b:$x} | key'
null   # same as before this PR -- falls through to the generic fallback
```

`Reduce`/`Foreach`/`Limit` (items 2-4 of the issue's own ordering) are left
as documented known gaps for follow-up — `Reduce`/`Foreach` raise a real
semantic question the issue itself flags (what does `key` mean while
evaluating the fold's own accumulator, which has no document position?)
that isn't settled yet.

The issue (#1765) is intentionally left open, narrowed to those three
remaining shapes, rather than closed by this PR.

## Verification

No jq/yq oracle exists for any of this — `key`/`parent`/`file_index` are
succinctly extensions, and yq's own lexer rejects destructuring `as`
outright (confirmed live against v4.53.3) — so this pins internal
consistency with the already-fixed `Expr::As` sibling, not oracle
compliance. Manually verified array destructuring, object destructuring,
and `parent(n)` reaching back through a bound position to the document
root, plus confirmed the `?//` and Reduce/Foreach/Limit cases are
unchanged.

## Test plan

- [x] Updated the existing #1663 known-gaps test
      (`test_as_pattern_reduce_foreach_limit_path_context_still_known_gaps_1663`
      → `test_reduce_foreach_limit_path_context_still_known_gaps_1663`) to
      drop the now-fixed single-pattern `AsPattern` case and add the `?//`
      case explicitly (previously only implied)
- [x] New test `test_as_pattern_single_pattern_path_context_resolves_1765`
      covering array destructuring, object destructuring, and `parent(n)`
- [x] Full test suite (`cargo test --features cli,simd,regex,serde`): all
      green, no regressions

Refs #1765

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -810,6 +810,16 @@ pub(crate) fn needs_path_context(expr: &Expr) -> bool {
         // This is the single most common trigger for this whole bug class:
         // `.a.b | . as $x | parent(1)` previously silently stubbed to `{}`.
         Expr::As { expr, body, .. } => needs_path_context(expr) || needs_path_context(body),
+        // `. as PATTERN | body` (#1765): same reasoning as `Expr::As` above,
+        // for destructuring `as`. Deliberately not gated on `patterns.len()`
+        // here -- unlike the dedicated evaluation arm below (which only
+        // handles the single-pattern case), this is just the routing
+        // question "does the whole pipe need path-context evaluation at
+        // all," and a `?//`-chain with a path-context builtin inside still
+        // needs `true` here so the surrounding pipe routes correctly, even
+        // though the chain itself then falls through to the generic
+        // fallback (its pre-existing, unchanged stub behavior) once inside.
+        Expr::AsPattern { expr, body, .. } => needs_path_context(expr) || needs_path_context(body),
         _ => false,
     }
 }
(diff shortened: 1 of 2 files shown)


newhoggy:

**Fixed — real, confirmed regression (2 independent agents converged on this)**: an earlier version of \`needs_path_context\`'s new \`Expr::AsPattern\` arm answered \`true\` for a \`?//\`-chain too, not just the single-pattern case the dedicated eval arm handles. One agent live-verified the concrete consequence: routing a \`?//\`-chain into the eager path-context evaluator whenever it sat inside \`Arithmetic\`/\`Compare\`/a non-terminal \`Pipe\` position hit the generic fallback's \`eval_owned_expr_full\`, which collapses a multi-output result into a single \`OwnedValue::Array\` instead of streaming it — confirmed live: \`(.a[] as {x:$v} ?// [$v] | key) + 0\` went from two separate outputs (matching pre-PR \`main\`) to a hard type error. Fixed by gating \`needs_path_context\`'s arm to \`patterns.len() == 1\`, matching the eval arm exactly (commit b9d255934) — re-verified byte-identical to pre-PR \`main\` on both the array-collapse and arithmetic-error repros, added a dedicated regression test pinning this.

**Fixed — real duplication risk**: the \`Expr::As\`/\`Expr::AsPattern\` eval arms were two ~65-line byte-for-byte-identical copies of the same bind/substitute/accumulate skeleton around one differing step (how a bound value becomes a substituted body). The reviewer cited the exact precedent this ignored — \`materialize_bound_values\`/\`accumulate_path_context_step\` were extracted for this identical reason, after #1457/#1660 had to independently correct two drifted copies of a related skeleton. Extracted \`eval_bind_with_path_context\`, parameterized over a \`substitute\` closure; both arms now call it. Needed \`#[allow(clippy::too_many_arguments)]\` (8 params — mirrors \`eval_pipe_with_path_context_internal\`'s own set plus \`rest\`/\`substitute\`), following this file's own established STYLE-0004 precedent for the sibling path-context helpers.

**Correction to my own PR description**: one agent found that \`key\`/\`parent\` are *not* succinctly-only extensions as I'd written — real yq v4.53.3 implements both. The "no oracle" claim only holds for the specific combination this PR fixes (destructuring \`as\` + these builtins), since yq's lexer rejects destructuring \`as\` outright — that narrower scoping was accurate, just worded misleadingly broadly.

**No correctness bugs found** beyond the regression above — one agent did a full line-by-line read plus live oracle probes (including verifying jq's own "does \`body\` rebind \`.\`" semantics, confirming the arm's use of the ambient \`value\`/\`current_path\` is correct, not a bug) and found nothing else.

**Not actioned**: the yq-mode test-coverage observation (this fix is reachable via \`succinctly yq\` too, generic over \`EvalSemantics\`) — noted as a thoroughness gap, not a rule violation, since no oracle exists for the specific combination either way.

Full suite, both clippy legs, fmt, and no_std build all re-verified clean after both fixes.

PR #1789, opened 2026-08-29
refactor(jq): consolidate combinations/cartesian_product's mixed-radix loop

### LINKED ISSUE #1721: jq: combinations(n)/cartesian_product's guard-then-reserve pattern is now the 4th hand-rolled copy, and the two mixed-radix loops duplicate each other
**Severity: Low** (maintenance/DRY concern, not a functional bug)

## Summary

Found by \`/code-review\` on PR #1716 (#1669's combinations allocation-crash fix). Two
related duplication concerns in the fixed code, both flagged independently by multiple
review angles:

## 1. Two near-identical mixed-radix counting loops

\`builtin_combinations_n\`'s new loop (indexing one shared \`base_array\` by a constant
radix) and \`cartesian_product\`'s existing loop (indexing \`arrays[i]\` by a per-position
radix) implement the same carry-propagation algorithm ~30 lines apart in
\`src/jq/eval.rs\`, differing only in whether the per-digit length is constant
(\`base_array.len()\`) or looked up per-array (\`arrays[i].len()\`). PR #1716's own doc
comment on the new loop (\"same as \`cartesian_product\`'s own loop below\") acknowledges
the duplication rather than eliminating it.

A future correctness fix to the carry/increment algorithm (or to how the allocation
guards interleave with it) has to be found and applied in both places by hand — exactly
the failure mode a second commit in PR #1716 already demonstrates for a related reason
(the first commit's guard was incomplete and needed a follow-up fix).

**Suggested direction**: generalize \`cartesian_product\` to accept either a slice of owned
arrays or "one array repeated n times" (e.g. an enum, or a per-position accessor closure
\`Fn(usize, usize) -> OwnedValue\` alongside a \`lens: &[usize]\`), so
\`builtin_combinations_n\` can call the same loop instead of maintaining its own copy.

## 2. The guard-then-reserve pattern is now the 4th independent copy in this file

\`try_reserve_product\`/\`cannot_reserve_cross_product\` (#1634), the string-repeat guard
(#1612), \`pad_with_nulls\`'s own \`try_reserve\` (#1017), and now
\`checked_combinations_len\`/\`cannot_allocate_combinations\`/\`cannot_allocate_combinations_count\`
(#1669) each independently re-derive the same "compute a checked size, \`try_reserve_exact\`,
build a purpose-specific \`EvalError\`" shape, with their own message format and their own
error-message call sites. Nothing in the file signals "allocation sizes driven by
generator/document input must go through a fallible path" except scattered doc comments
on unrelated functions — a fifth builtin that takes a numeric argument and builds a \`Vec\`
has no enforced way to discover this is an established, recurring bug class (#1612 →
#1017 → #1634 → #1669, four independent rediscoveries of the same shape).

**Suggested direction**: a single shared helper (e.g.
\`fn try_reserve_sized<T>(len: usize, describe: impl FnOnce() -> String) -> Result<Vec<T>,
EvalError>\`) that all four sites route through, so the pattern is discoverable by grep and
new call sites can't accidentally skip the guard.

## 3. Minor: \`cartesian_product\`'s overflow message is reused verbatim from an unrelated case

\`cartesian_product\`'s length-product overflow (backing bare \`combinations\`) reuses
\`try_reserve_product\`'s \"Cannot allocate ... elements for a computed-index expansion\"
wording, which describes \`.[$keys]\`-style computed indexing — bare \`combinations\` never
indexes anything. PR #1716's own docs update acknowledges reusing this message verbatim.
Low priority (purely cosmetic, the existing test only asserts the \"Cannot allocate\"
prefix), but worth a dedicated message the next time this area is touched.

## Related

#1669 (PR #1716) — the fix these three sites all appear in. Not addressed there to avoid
scope creep on what was already a two-commit crash-fix PR.

newhoggy:
## Summary

Found by `/code-review` on PR #1716 (#1669's combinations allocation-crash fix). This
PR addresses items 1 and 3 of that review:

### 1. Two near-identical mixed-radix counting loops

`builtin_combinations_n`'s loop (indexing one shared `base_array` by a constant radix)
and `cartesian_product`'s loop (indexing `arrays[i]` by a per-position radix)
implemented the same carry-propagation algorithm ~30 lines apart, differing only in
whether the per-digit length is constant or looked up per-array. PR #1716's own doc
comment on the new loop ("same as `cartesian_product`'s own loop below") acknowledged
the duplication rather than eliminating it.

Extracted the shared walk into `mixed_radix_combinations(width, results, len_at,
value_at)`, taking closures for the one difference. Each caller keeps its own
already-reviewed allocation guards completely unchanged, in their original order --
the shared function only walks and pushes into an already-reserved `Vec`, it never
sizes or grows the backing allocation itself. This was a deliberate design choice to
minimize risk: the guard logic in both callers is the crash-prevention code #1669
carefully ordered cheapest-check-first, and this PR doesn't touch it at all.

### 3. Misleading reused error message

`cartesian_product`'s length-product overflow (backing bare `combinations`) was
reusing `try_reserve_product`'s "Cannot allocate ... elements for a computed-index
expansion" wording, which describes `.[$keys]`-style computed indexing -- bare
`combinations` never indexes anything. Gave it a dedicated
`cannot_allocate_cartesian_product` message ("Cannot allocate ... elements for
combinations", matching the other two combinations-family messages) instead of
touching `try_reserve_product` itself, which is used correctly by 8+ other call sites
across `eval.rs`/`eval_generic.rs`.

### Not addressed here (filed separately)

Item 2 from the same review -- a shared `try_reserve_sized`-style helper unifying all
4 independent "guard-then-reserve" copies (#1017/#1612/#1634/#1669) -- needs its own
dedicated pass: it touches more call sites across more of this crash-prevention code
than a loop-dedup PR should bundle, and deserves review focused specifically on
allocation-guard correctness rather than sharing space with an unrelated loop
refactor.

## Test plan

- [x] Manually verified: `combinations(n)` basic output, bare `combinations` output,
      `combinations(0)` (`[[]]`), empty-base/empty-inner-array no-output cases -- all
      unchanged
- [x] Manually verified all three allocation-guard crash paths still reject cleanly
      (exit 5, no abort): huge `n` count, huge `base^n`, huge bare-`combinations`
      product -- the last now reporting the corrected message
      (`Cannot allocate 10 * 10 * ... elements for combinations`)
- [x] Full `cargo test --features cli,simd,regex,serde` -- 55/55 test binaries pass,
      including every existing `_1669`/`_1720`/`_1279`/`_1408` crash-regression test

Fixes #1721

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -31120,40 +31120,71 @@ fn builtin_combinations_n<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     if results.try_reserve_exact(total).is_err() {
         return QueryResult::Error(cannot_allocate_combinations(base_array.len(), n));
     }
-    let mut indices: Vec<usize> = vec![0; n];
+
+    mixed_radix_combinations(
+        n,
+        &mut results,
+        |_| base_array.len(),
+        |_, idx| base_array[idx].clone(),
+    );
+
+    QueryResult::ManyOwned(results)
+}
+
+/// The mixed-radix counting walk shared by [`cartesian_product`] and
+/// `builtin_combinations_n` (#1721): both build every combination of `width`
+/// positions by treating `indices` as a mixed-radix counter and carrying
+/// into the next position once the current one wraps, differing only in
+/// whether a position's own length is looked up per-position (independent
+/// arrays) or is one constant repeated `width` times (`combinations(n)`'s
+/// single array raised to itself). `len_at`/`value_at` abstract over that
+/// one difference; everything else -- the counter, the carry propagation,
+/// the termination check -- was previously hand-duplicated in both
+/// callers, ~30 lines apart, and had to be kept in sync by eye.
+///
+/// `results` must already be reserved for the caller's own already-guarded
+/// total capacity -- this only walks and pushes, it never sizes or grows
+/// the backing allocation itself, so it carries none of either caller's
+/// crash-prevention guards and must not be called before them.
+fn mixed_radix_combinations(
+    width: usize,
+    results: &mut Vec<OwnedValue>,
+    len_at: impl Fn(usize) -> usize,
+    value_at: impl Fn(usize, usize) -> OwnedValue,
+) {
+    let mut indices = vec![0usize; width];
 
     loop {
-        let combination: Vec<OwnedValue> =
-            indices.iter().map(|&idx| base_array[idx].clone()).collect();
+        let combination: Vec<OwnedValue> = (0..width).map(|i| value_at(i, indices[i])).collect();
         results.push(OwnedValue::Array(combination));
 
-        // Increment indices (like counting in mixed radix), same as
-        // cartesian_product's own loop below.
+        // Increment indices (like counting in mixed radix).
         let mut carry = true;
-        for idx in indices.iter_mut().rev() {
+        for i in (0..width).rev() {
             if carry {
-                *idx += 1;
-                if *idx >= base_array.len() {
-                    *idx = 0;
+                indices[i] += 1;
+                if indices[i] >= len_at(i) {
+                    indices[i] = 0;
                 } else {
                     carry = false;
                 }
             }
         }
+
+        // If we carried all the way through, we're done.
         if carry {
             break;
         }
     }
-    QueryResult::ManyOwned(results)
 }
 
 /// Compute the Cartesian product of a list of arrays.
 ///
 /// The result has one entry per combination -- `arrays.iter().map(Vec::len)`'s
-/// product -- so that product is exactly what [`try_reserve_product`] guards
-/// before the loop below grows `results` one push at a time: without this,
-/// `results` grows via ordinary `Vec` doubling with no upfront size check at
-/// all, the identical missed guard #1612/#1634 already closed for other
+/// product -- so that product is exactly what the guard below checks before
+/// the walk grows `results` one push at a time: without this, `results`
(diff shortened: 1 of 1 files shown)


newhoggy:
Multi-round \`/code-review\` (10+ finder angles) found one real issue, corroborated independently by nearly every angle (reuse, simplification, altitude/conventions, line-by-line, removed-behavior, wrapper-correctness, language-pitfall): \`cartesian_product\`'s own overflow guard had duplicated \`try_reserve_product\`'s entire checked-multiply-then-reserve body -- and dropped its zero-factor short-circuit in the process -- purely to get a differently-worded error message. Ironic for a dedup PR.

Fixed by splitting \`try_reserve_product\` into a labeled variant taking the error constructor as a parameter, so \`cartesian_product\` shares the exact same guard logic (short-circuit included) while still getting its own accurate message. Also extracted the shared \`"a * b * c"\` formatting into \`format_product\`, added \`#[inline]\` to \`mixed_radix_combinations\` matching this file's convention for extracted hot-path helpers, had \`cartesian_product\` reuse its already-computed \`lens\` instead of re-deriving lengths in the walk closure, added a doc comment to \`cannot_allocate_cartesian_product\` matching its siblings' cross-referencing convention, and fixed a stale comment referring to \`indices\` after it moved into the shared helper.

Manually re-verified the restored zero-factor short-circuit (\`[[],[1,2,3]] | combinations\` → \`[]\`, no panic) plus all existing crash-guard behavior. Full test suite passes, including every \`_1669\`/\`_1634\` crash-regression test.

Issue #1791, opened 2026-08-29
yq: join's empty-object element renders as "{}" where real yq gives empty string


newhoggy:
**Severity: Low** (cosmetic — element part choice, not a crash or data loss)

Found live-verifying #1535's own refactor (a pure hoist, unrelated to this) — pre-existing on unmodified \`main\`, not introduced by that PR.

## Repro (confirmed against yq v4.53.3)

\`\`\`bash
\$ printf '[{}]\n' | yq -o=json -I0 'join(",")'
""

\$ printf '[{}]\n' | succinctly yq -o=json -I0 'join(",")'
"{}"
\`\`\`

## Where

\`collect_join_parts\`'s yq-mode array arm (\`src/jq/eval.rs\`, née \`join_with_separator\`) special-cases an empty object element to render as the literal string \`"{}"\` — this was added deliberately for #1047, with a doc comment citing that fix. Re-verifying it live now (pinned yq v4.53.3) shows real yq's actual answer for this exact shape is an **empty string**, not \`"{}"\` — either #1047's own original oracle check was wrong, or yq's behavior has since changed, or the fix generalized a narrower confirmed case incorrectly. Not investigated further here — worth checking #1047's own PR/issue thread for what was actually verified there before assuming this fix's premise was right.

## Scope

Not fixed here — surfaced during unrelated verification, filed to track rather than silently leaving an oracle-verified divergence unrecorded. \`yq_join_element_part\`'s own doc comment (referenced by the same array arm) may need re-checking too, since it documents the same "container collapses" rule this special case is layered on top of.

newhoggy:
## Triage (2026-09-04) — decided: delete the empty-object *element* special case; the separator case (#1047's real finding) stays

#1047's thread verified only the separator: `[1,2] | join({})` → `1{}2` (still true today in yq
v4.53.3 and succinctly). The element special case was an over-generalisation. Element oracle, 24
shapes: every container element renders `""` regardless of emptiness — `[{}]` → `""`, `["a",{},"b"]`
→ `"a,,b"`, `[{},{}]` → `","`, `[[]]`/`[[1]]`/`[{"a":1}]` → `""`; `[null]` → `""`; scalars → text
(`[1.50]` → `"1.50"`, `[1e3]` → `"1e3"`, `[true]` → `"true"`). succinctly matches all but the three
`{}`-element shapes.

### Decision
- `eval.rs::collect_join_parts`, yq array branch: remove the `StandardJson::Object(fields) if
[...]

Issue #1755, opened 2026-08-28
jq: eval.rs's ~50 remaining read-only builtins never got #1746's decode-failure fix


newhoggy:
**Severity: High**

## Summary

#1746 (PR #1754) fixed `eval.rs`'s private, `StandardJson`-specific `to_owned`/`to_owned_at_depth` for its "whole-document read-modify-write" family (`builtin_path`, `eval_assign`, `eval_update`, `yq_assign_noop_check`, `builtin_setpath`, `builtin_del`, `delpaths_one`) plus `map`/`map_values`/`to_entries`. A systematic audit run during that PR's own review found the same silent-corruption bug class is live in roughly **50 more builtins** in the same file — every one of them materializes a `StandardJson` container's own elements/fields directly via the still-unchecked `to_owned`, and returns the result straight to the caller (via `QueryResult::Owned`/`ManyOwned`) with no downstream check that would catch the corruption. This issue exists to track fixing the rest, since #1746's own PR was already large and this remaining set is a genuinely separate, larger body of work.

## Repro (confirmed live against current `main`, via the public `succinctly::jq::eval`/`eval_using` library API — same reachability #1746 established)

```
[1,"\ud800"] | flatten
```
returns `[1,""]` silently (exit 0) instead of raising a decode-failure error. `pick(["a"])` and `omit(["b"])` on `{"a":"\ud800",...}` do the same for the picked/kept fields.

## Root cause

Same as #1746: `eval.rs`'s plain `fn to_owned<W>(value: &StandardJson<'_, W>) -> OwnedValue` (not the `to_owned_checked` fallible sibling #1746 added) silently substitutes `OwnedValue::String(String::new())` for a string that fails to decode, instead of raising `EvalError::decode_failure`. Every site below still calls the unchecked version.

## Full site list (from #1746's own review audit — "Bucket A")

Direct (the corrupted value *is* the returned data):

[...]

PR #1792, opened 2026-08-29
perf(jq): hoist join's element walk above its separator fan-out

### LINKED ISSUE #1535: jq: join re-collects every array/object element once per fanned-out separator output
**Severity: Low** (performance only, jq mode, requires a multi-output separator argument)

## Summary

Found during #1279's own code review (PR pending). `builtin_join` (`src/jq/eval.rs:5837`) fans its separator argument out via `fanout_arg` with `ArgFanout::yq_native::<S>()`, which is `All` in jq mode (real jq's `join` does fan out over a multi-output separator: `[.x | join((",","-"))]` on `["a","b"]` is `["a,b","a-b"]`). The entire per-separator body is `join_with_separator` (`src/jq/eval.rs:5858`), which re-walks and re-collects every array/object element from scratch — `elements.try_fold(...)` in jq mode, `elements.map(...).collect()` in yq mode — once per separator output, even though the element list itself is identical on every iteration; only the separator differs.

`#1279`'s own design doc calls this exact shape out by name as a case whose element-collection work should be hoisted above the `fanout_arg` call rather than repeated inside the per-output closure, but the hoist wasn't done for `join`.

## Cost

`[1,2,3] | [join((",","-"))]` walks and deep-copies the source array/object twice (once per separator) instead of once; the redundant work scales with both the container's size and the number of separator outputs.

## Suggested fix direction

Compute the element list/stringified parts once above the `fanout_arg` call (as the design doc intends), and have the per-separator closure only perform the final `.join(&sep)` step, matching the pattern already used elsewhere in this PR for hoisting a document-wide computation out of a fan-out closure (e.g. `contains`/`inside`'s whole-document `to_owned` hoisting).

## Related

- #1279 — the fan-out work this was found reviewing; its own design doc identifies this exact hoist as necessary but it wasn't applied to `join`.

newhoggy:
## Summary

Fixes #1535. \`builtin_join\` fans its separator argument out (jq mode: real jq's own \`join\` genuinely fans out over a multi-output separator). \`join_with_separator\`, the per-separator body, re-walked and re-decoded every array/object element from scratch on every call, even though the element list itself never changes — only the separator does. #1279's own design doc names this exact shape as a case whose element collection should be hoisted above \`fanout_arg\`, matching the \`contains\`/\`inside\` hoist already in this file, but it wasn't applied to \`join\`.

## Fix

Split into two functions:
- \`collect_join_parts\`: the one-time, separator-independent element walk/decode — computed lazily, on the fan-out's first call, and cached (memoized via a captured \`Option\`) across every later call.
- \`join_parts_with_separator\`: the per-separator accumulate-only step, now cloning small already-stringified parts instead of re-decoding cursor elements each time.

Every documented edge case is preserved exactly:
- The separator is never validated for a single-element array (\`[1] | join(1)\` succeeds — real jq's own algorithm never touches the separator until the *second* element's accumulation).
- \`null\` elements become empty-string parts (not skipped).
- A container element fails \`+\` naturally with jq's own wording, on every separator output identically.
- **Zero-output separator means the array/object validity check never runs at all** — an invalid target under e.g. \`join(empty)\` still contributes no error, exactly as before the hoist (verified via the memoization design: the closure is never invoked if the separator generator produces zero outputs).

## Also found (not fixed here)

Live-verifying this surfaced a pre-existing, unrelated divergence: yq's own empty-object-element special case (#1047) renders \`"{}"\` where real yq v4.53.3 actually gives \`""\` for \`[{}] | join(",")\`. Confirmed present on unmodified \`main\` too, so not introduced by this PR. Filed as #1791 rather than folded in.

## Verification

- 2 new regression tests: multi-separator fanout correctness, and an element-error reproducing identically on every separator output (confirming the cached parts carry the same raw element through each fold, not silently succeeding after the first).
- Live-verified every case above against \`/usr/bin/jq\` (1.7.1 pin) and pinned yq v4.53.3, plus the existing \`test_join_uses_first_separator_value_then_propagates_trailing_break_1164\`/\`test_join_propagates_halt_in_separator_argument\` control-flow tests pass unchanged.
- Full test suite green (55/55 binaries), both clippy legs clean, \`cargo fmt --check\` clean, \`cargo doc\` clean, \`--no-default-features\` builds clean.

## Test plan

- [x] \`cargo test --features cli,simd,regex,serde\`
- [x] \`cargo clippy --all-targets --all-features -- -D warnings\`
- [x] \`cargo clippy --all-targets --features cli,simd,regex,serde -- -D warnings\`
- [x] \`cargo fmt --check\`
- [x] \`cargo doc --no-deps --all-features\`
- [x] \`cargo build --no-default-features\`

Closes #1535

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -7470,19 +7470,14 @@ fn yq_join_separator(sep: OwnedValue) -> String {
     }
 }
 
-/// One step of `join`'s jq-mode fold: combine the running accumulator with
-/// the next element, interleaving the separator. See `builtin_join`'s doc
-/// comment for the algorithm this implements.
-fn join_step<S: EvalSemantics>(
-    acc: Option<OwnedValue>,
-    elem: OwnedValue,
-    sep: &OwnedValue,
-) -> Result<Option<OwnedValue>, EvalError> {
-    let left = match acc {
-        None => OwnedValue::String(String::new()),
-        Some(a) => arith_add::<S>(a, sep.clone())?,
-    };
-    let right = match elem {
+/// jq-mode `join`'s per-element stringification -- the separator-independent
+/// half of the fold `join_step` (pre-#1535) used to redo per separator
+/// output. `Null`/number/boolean become a string part; `String` passes
+/// through; `Array`/`Object` are passed through raw so `+` fails on them
+/// naturally in [`join_parts_with_separator`], matching real jq's own wording
+/// -- see `builtin_join`'s doc comment for the full algorithm.
+fn join_element_part(elem: OwnedValue) -> OwnedValue {
+    match elem {
         OwnedValue::Null => OwnedValue::String(String::new()),
         // `to_json()`, not `owned_to_string()`: jq's `tostring` renders
         // NaN as `null` (confirmed live, jq 1.7.1: `nan | tostring` ->
@@ -7495,9 +7490,8 @@ fn join_step<S: EvalSemantics>(
         | OwnedValue::Int(_)
         | OwnedValue::Float(_)
         | OwnedValue::NumberLiteral(..) => OwnedValue::String(elem.to_json()),
-        other => other, // String passes through; Array/Object stay raw so `+` fails naturally below.
-    };
-    Ok(Some(arith_add::<S>(left, right)?))
+        other => other,
+    }
 }
 
 /// Builtin: join(s) - join array elements with separator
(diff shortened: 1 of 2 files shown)


newhoggy:
Follow-up commit \`ac0be3268\` addresses a review finding: \`collect_join_parts\`'s jq-mode collection now stops at the first \`Array\`/\`Object\` part instead of eagerly decoding every element, matching the original \`try_fold\`-based algorithm's own decode boundary exactly (\`+\` fails on any container part regardless of separator value, so nothing past it can change the outcome — and since \`to_owned\` panics past \`MAX_VALUE_TREE_DEPTH\` #1005, decoding elements the original never would have is a correctness/safety issue, not just wasted work).

While constructing a concrete repro for this, I also found and filed **#1793**: a document with 257+ levels of nesting *anywhere* panics **any** non-lazy-path builtin (confirmed with \`sort\`/\`map(.)\`, not just \`join\`) — even when the query never reads the deep part, and confirmed identical on unmodified \`main\`. That's a separate, pre-existing, materializing-path issue unrelated to this PR's own algorithm choice; not fixed here.

Full suite/clippy/fmt/doc/no_std all re-verified clean after this commit.

newhoggy:
Rebased onto latest main — a genuine content conflict this time (not just a fast-forward): #1755 landed on main and independently added decode-failure checking to \`join\` (\`to_owned_checked\`/\`scalar_decode_failure\` instead of the old infallible \`to_owned\`, for both the yq array-element path and the jq array/object/scalar path).

Merged both changes together: \`collect_join_parts\`'s jq-mode collection now threads \`to_owned_checked\`'s \`Result\` through \`collect_jq_join_parts_until_first_container\` (propagating a decode failure immediately, before the container short-circuit fix from the earlier review round even gets a chance to apply), and the yq-mode array arm keeps its now-fallible \`to_owned_key_shape\` per element. Also squashed my two commits into one during this rebase — same content, cleaner history.

Full suite (55/55)/clippy (both legs)/fmt/doc/no_std all re-verified clean after the merge.

PR #1795, opened 2026-08-29
fix(cli): make a stdout flush failure loud instead of silently swallowed

### LINKED ISSUE #1680: jq,yq: run_jq/run_yq rely on Drop's silent flush at many more early-return sites beyond #1563
`run_jq`/`run_yq` rely on `BufWriter`'s `Drop` to flush on most early
returns; #1563/#1673 patched three sites in `yq_runner.rs`'s `'m2_files`
loop and one in `jq_runner.rs`'s lazy-path loop where a later
file's/document's error could otherwise skip an explicit flush and
silently lose an earlier file's already-buffered output (`Drop`'s own
flush swallows any error rather than propagating it).

A multi-angle review of that PR (10 independent finder passes) confirmed
the fix is correct but incomplete: the same failure shape exists at
several more sites in both files, all inside the very functions #1673
touched. Fixing them required three follow-up commits within the PR
before converging even on the narrow set the review happened to check —
strong evidence that patching individual `?`-return sites by hand isn't
converging, and a fresh audit would very likely turn up still more.

**Known remaining unguarded sites** (each: a loop writes real output to
the shared `writer`/`out` across multiple iterations, and a later
iteration's write/build call can fail via bare `?` without an explicit
flush first):

- `yq_runner.rs`: the `stream_cursor!` macro (~3299-3367, invoked at
  3399/3447/3509/3572) and the direct `stream_maybe_colored`/`writeln!`
  calls at ~3438/3545/3554/3563 — the actual per-document streaming
  writes inside the `'m2_files` loop #1673 already touched for its
  `read_file`/`yaml_validate_guard`/`YamlIndex::build` sites.
- `yq_runner.rs`: the general/DOM "collect" output path (~4383-4431) —
  the default fallback for any multi-file run not eligible for M2
  streaming (e.g. `--pretty`/`-P`).
- `yq_runner.rs`: the `--raw-input` per-line loop (~3843-3855) and the
  `--eval-all` per-result loop (~3652-3671).
- `jq_runner.rs`: the DSV streaming loop's `write_output(...)?` (~1076;
  its own halt path three lines below already flushes explicitly,
  showing the ordinary-error case was just missed).
- `jq_runner.rs`: the materializing branch's three loops — `null_input`
  arm (~1397), the `input`/`inputs` queue-pop arm (~1426), and the plain
  per-input arm (~1444) — each calling `write_output(...)?` across
  multiple documents. This directly contradicts #1673's own added
  comment claiming this branch's early returns "provably run before
  anything is ever written to `out`" — true only for the single
  `get_inputs` failure before the loop starts, not for iteration N>=2 of
  any of these three loops.
- `jq_runner.rs`: the identity fast path's `out.write_all(...)?` calls
  (~1207), inside the very loop #1673 already patched for its
  `--validate` and value-materialization error paths.

**Proposed direction** (from the review's "right altitude" finding,
worth validating before implementing): move the safety net to writer
*construction* (`jq_runner.rs` and `yq_runner.rs`'s
`BufWriter::new(stdout.lock())`) instead of continuing to annotate each
early return by hand. A `Drop`-based guard can't propagate a `Result`,
but it can at least make a flush failure *loud* (e.g. log to stderr)
instead of *silent* -- closing the whole bug class in one place rather
than requiring another manual audit pass every time a new early return
is added to either ~5,000-line function. A full `Result`-propagating fix
would need a `finish(self) -> Result<()>` consumed at every exit
instead, which is a larger refactor of both functions' control flow.

Not urgent (the concrete `--validate` scenario #1563 named, plus the
directly adjacent sites in the same loop, are already fixed) but worth
tracking so it doesn't silently regress the "we already fixed this"
assumption the next time someone touches either runner's output loops.

newhoggy:
## Summary

`run_jq`/`run_yq` rely on `BufWriter`'s `Drop` to flush on most early returns.
#1563/#1673 patched three sites in `yq_runner.rs`'s `'m2_files` loop and one in
`jq_runner.rs`'s lazy-path loop where a later file's/document's error could otherwise
skip an explicit flush and silently lose an earlier file's already-buffered output
(`Drop`'s own flush swallows any error rather than propagating it).

A multi-angle review of that PR confirmed the fix is correct but incomplete: the same
failure shape exists at several more sites in both files (identity fast path, DSV
streaming loop, `--raw-input`/`--eval-all` per-item loops, the general DOM collect
path, ...), all inside the very functions #1673 already touched. Fixing the originally
identified sites required three follow-up commits within that PR before converging even
on the narrow set the review happened to check -- strong evidence that patching
individual `?`-return sites by hand isn't converging.

## Change

Implements the issue's own proposed direction: move the safety net to writer
*construction* instead of continuing to annotate each early return by hand.

Added `LoudFlushWriter`, a `BufWriter` wrapper whose `Drop` reports a flush failure to
stderr instead of silently discarding it. Swapped it in at both runners' one
stdout-backed writer construction site (`jq_runner.rs`'s `out`, `yq_runner.rs`'s
`writer`) -- the two other `BufWriter::new(...)` sites in `yq_runner.rs` wrap an
in-memory `Vec<u8>` buffer, which can't fail to flush, so those are left as plain
`BufWriter`.

This is **not** a full fix for the underlying data loss -- once a `?` has already
unwound past the point a later write would have happened, that output was never
buffered in the first place, and if an *earlier* buffered write's flush now fails, those
bytes are still gone. What changes is that the failure becomes a loud, visible
diagnostic instead of silently dropped output. A `finish(self) -> Result<()>` consumed
at every exit would `Result`-propagate a flush failure properly instead of merely
reporting it, but that's the larger refactor of both functions' control flow the issue
itself deferred.

`LoudFlushWriter` tracks whether an explicit `.flush()` call (`flush_then_err`'s own, or
`--unbuffered`'s per-value flush) already observed and reported a failure, so `Drop`'s
own attempt -- which always runs regardless of which path got the writer here --
doesn't print a second, redundant diagnostic for the exact same failure the caller's own
error already carries.

## Test plan

- [x] Two unit tests pinning `LoudFlushWriter`'s core behavior: exactly one flush
      attempt total, whether it happens via an explicit caller flush or via `Drop`'s
      backstop
- [x] Manually verified against a real broken pipe (`seq 1000000 | sjq ... | head -1`):
  - A previously fully-unguarded site (the identity fast path, reachable via
    `--preserve-input -c '.'`) now reports `succinctly: failed to flush output: Broken
    pipe (os error 32)` where it previously printed nothing at all
  - An already-`flush_then_err`-guarded site still prints its existing message exactly
    once, not twice
  - An ordinary successful run's stderr stays completely empty (no false-positive noise)
- [x] Full `cargo test --features cli,simd,regex,serde` -- 55/55 test binaries pass (one
      transient failure on the first run under concurrent-session contention --
      "Blocking waiting for file lock on package cache" -- confirmed spurious by
      isolation re-run and a clean full re-run)
- [x] `cargo build --no-default-features` (no_std) -- builds, no new warnings (change is
      `src/bin`-only, doesn't touch the no_std library)

Fixes #1680

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -6,7 +6,7 @@
 use anyhow::{Context, Result};
 use indexmap::IndexMap;
 use std::collections::BTreeMap;
-use std::io::{BufWriter, IsTerminal, Read, Write};
+use std::io::{IsTerminal, Read, Write};
 use std::path::{Path, PathBuf};
 
 use succinctly::dsv::{build_index as build_dsv_index, DsvConfig, DsvRows};
@@ -27,7 +27,8 @@ use succinctly::json::JsonIndex;
 use super::JqCommand;
 use crate::output::{
     self, escape_json_string, escape_json_string_ascii, exit_codes, flush_then_err, ColorScheme,
-    ControlEscape, DiagStyle, ErrorSink, FloatStyle, InputLocation, JsonFormatOpts, Terminator,
+    ControlEscape, DiagStyle, ErrorSink, FloatStyle, InputLocation, JsonFormatOpts,
+    LoudFlushWriter, Terminator,
 };
 
 /// Evaluation context for passing variables to the jq evaluator.
@@ -1018,7 +1019,7 @@ pub fn run_jq(args: JqCommand) -> Result<i32> {
 
     // Set up output writer
     let stdout = std::io::stdout();
-    let mut out = BufWriter::new(stdout.lock());
+    let mut out = LoudFlushWriter::new(stdout.lock());
 
     // Track last output for exit status
     let mut last_output: Option<OwnedValue> = None;
--- a/src/bin/succinctly/output.rs
+++ b/src/bin/succinctly/output.rs
@@ -349,51 +349,59 @@ pub fn flush_then_err<W: std::io::Write, T>(
 /// properly instead of merely reporting it, but that is the larger
 /// refactor of both functions' control flow this issue itself deferred.
 ///
-/// Tracks whether an explicit `.flush()` call (`flush_then_err`'s own, or
-/// `--unbuffered`'s per-value flush) already observed and reported a
-/// failure, so `Drop`'s own flush attempt -- which always runs regardless
-/// of which path got the writer here -- doesn't print a second, redundant
-/// diagnostic for the exact same underlying failure the caller's own
-/// `Result` already carries.
-pub struct LoudFlushWriter<W: Write> {
-    inner: BufWriter<W>,
-    flush_already_failed: std::cell::Cell<bool>,
-}
+/// Deliberately does **not** try to suppress a redundant report when an
+/// explicit `.flush()` call (`flush_then_err`'s own, or `--unbuffered`'s
+/// per-value flush) already observed and reported the same failure: an
+/// earlier version tracked that case with a "already failed" flag, but the
+/// flag only fired on an explicit `flush()` call, not on `write`/
+/// `write_all` -- and `BufWriter`'s own `write`/`write_all` can trigger a
+/// real inner flush whenever an incoming write doesn't fit the remaining
+/// buffer capacity (the common case once the buffer fills during
+/// streaming, e.g. a broken pipe from `| head`). Extending the flag to
+/// also cover writes would make one transient failure permanently disable
+/// `Drop`'s own backstop for the rest of the process's life, silently
+/// reintroducing exactly the data-loss risk this type exists to close, on
+/// every later flush regardless of whether it might have actually
+/// succeeded. An occasional cosmetic second diagnostic line for the same
+/// underlying I/O condition is a much smaller cost than that, so `Drop`
+/// always attempts its own flush and always reports a failure, full stop.
+pub struct LoudFlushWriter<W: Write>(BufWriter<W>);
 
 impl<W: Write> LoudFlushWriter<W> {
     pub fn new(inner: W) -> Self {
-        Self {
-            inner: BufWriter::new(inner),
-            flush_already_failed: std::cell::Cell::new(false),
-        }
+        Self(BufWriter::new(inner))
     }
 }
 
 impl<W: Write> Write for LoudFlushWriter<W> {
     fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
-        self.inner.write(buf)
+        self.0.write(buf)
     }
 
     fn write_all(&mut self, buf: &[u8]) -> std::io::Result<()> {
-        self.inner.write_all(buf)
+        self.0.write_all(buf)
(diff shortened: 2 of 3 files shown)


newhoggy:
Extensive multi-round \`/code-review\` (10+ finder angles, very heavy convergence) found two real problems in the first commit, both fixed in a follow-up commit:

**1. The dedup logic didn't actually prevent double-reporting for the common case.** \`flush_already_failed\` was only set inside the \`flush()\` override, never inside \`write\`/\`write_all\` -- but \`BufWriter\`'s own \`write\`/\`write_all\` can trigger a real inner flush whenever an incoming write doesn't fit the remaining buffer capacity (the common case once the buffer fills during streaming, e.g. piping into \`head\`). That failure bypassed the tracking entirely, contradicting the doc comment's and commit message's claim.

The tempting fix (also track write failures) has a *worse* failure mode: one transient write failure would permanently disable \`Drop\`'s own backstop for the rest of the writer's life, silently reintroducing the exact data-loss risk this type exists to close. **Removed the dedup state entirely instead** -- \`Drop\` now always attempts its own flush and always reports a failure. The accepted cost is an occasional cosmetic second diagnostic line when an explicit flush already reported the same failure -- a much smaller cost than silently losing the backstop.

**2. A real panic-in-Drop risk.** \`eprintln!\` panics if the write to stderr itself fails (documented \`std\` behavior) -- if stderr is also broken (common when a supervisor tears down both stdio streams together), the "graceful" diagnostic would itself panic, or worse, abort the whole process via Rust's double-panic rule if it fires mid-unwind from an unrelated panic elsewhere in either ~5,000-line runner. Switched to a direct \`writeln!(io::stderr(), ...)\` with the result discarded, which cannot panic.

**Also corrected an inaccurate manual-verification claim** in the first commit: the \`--preserve-input\` broken-pipe repro did *not* previously print nothing, as claimed -- it already printed \`Error: Broken pipe\` via ordinary \`anyhow\` propagation, since a write failure there does propagate through the \`Result\`. The type's actual, narrower value is the case that error path can't reach: a write failure discovered only when an unrelated, non-I/O early return (a data error reported through a separate diagnostic channel) skips past ever flushing it, with nothing else in the call chain ever seeing or reporting the failure.

Added a third unit test covering the write-failure-then-drop path the original test suite didn't exercise (and had to fix it once already -- a small \`write_all\` never reaches the mock at all, since it fits entirely in \`BufWriter\`'s own internal buffer; needed a write past the default 8KB capacity to actually exercise the inner writer).

Multiple independent finder angles (efficiency, wrapper-correctness, reuse, conventions) confirmed no other issues: no hot-path cost from the wrapper (confirmed via \`nm\`/disassembly reasoning that a same-empty-buffer re-flush costs zero real syscalls), no double-wrapper construction, no CLAUDE.md violations, no pre-existing helper this duplicates.

Issue #1634, opened 2026-08-25
jq: unguarded keys.len() * targets.len() capacity multiplications can panic/abort (#1612 sibling)


newhoggy:
**Severity: Low** (crash, not data corruption — same class as #1612, at different sites)

## Summary

Found during `/code-review` of #1612 (`arith_mul`'s string-repeat overflow fix). #1612 closed
`String::repeat`'s own unguarded byte-length multiplication (`s.len() * n`), but the same
pattern — an unchecked `usize` multiplication of two document/generator-controlled lengths
feeding straight into an allocation — exists unguarded at several other sites in
`src/jq/eval.rs`:

```
src/jq/eval.rs:13511:  Vec::with_capacity(keys.len() * ts.len())
src/jq/eval.rs:13539:  Vec::with_capacity(keys.len() * ts.len())
src/jq/eval.rs:13635:  Vec::with_capacity(starts.len() * ends.len())
src/jq/eval.rs:19573:  Vec::with_capacity(keys.len() * target_branches.len())
src/jq/eval.rs:19834:  Vec::with_capacity(starts.len() * ends.len() * target_branches.len())
```

These are the cross-product capacity pre-allocations for computed-index expansion
(`.[$keys] as $t | ...`-style constructs, both the value-position and `path()`-position
resolvers). `keys`/`ts`/`starts`/`ends`/`target_branches` are all generator- or
document-controlled lengths — e.g. two `range(N)`-produced arrays as computed-index operands
— so a large enough `N` on each side can overflow the product.

## Why this is separate from #1612

#1612 was scoped to `arith_mul`'s own string-repeat arm specifically (`String::repeat`'s
`s.len() * n`), found and fixed via a single, well-characterized repro. These are a different
function family (computed-index cross-product expansion, not arithmetic), found only via
`/code-review`'s own cross-file trace — no live repro was captured for these specific sites
before filing, and the exact panic/abort threshold (`isize::MAX / size_of::<T>()`, which
differs per site depending on the element type — `StandardJson`, `OwnedValue`, etc.) wasn't
derived per site.

## Suggested fix direction (needs its own oracle spike + live repro first)

Same technique #1612 (and #1017's `pad_with_nulls`) already established for this exact bug
class: guard the product with a widened (`u128` or `checked_mul`) computation before calling
`Vec::with_capacity`, and/or replace the infallible `with_capacity` with a `try_reserve_exact`
[...]

PR #1796, opened 2026-08-29
ci(jq): disallow raw Vec/String::with_capacity in eval.rs/eval_generic.rs

### LINKED ISSUE #1670: jq: consider a clippy lint against unguarded Vec::with_capacity(a * b) — fourth recurrence of the #1612/#1634 bug class
**Severity: Low** (process/tooling improvement, not a live bug)

## Summary

Found during #1666's review (the #1634 fix). This is now the *fourth* independently-discovered instance of the same bug shape — an infallible `Vec`/`String` allocation sized by a product (or otherwise generator-controlled magnitude) of independently-controlled lengths, handed straight to the allocator instead of a `try_reserve`-style fallible guard:

- #1017 — `cannot_grow_array`/`pad_with_nulls` (`setpath`'s array-grow)
- #1612 — `arith_mul`'s string-repeat guard (`s * n`)
- #1634 (closed by #1666) — 7 sites across `eval.rs`/`eval_generic.rs`'s computed-index/-slice cross products
- #1669 — `combinations(n)`/`cartesian_product` (confirmed live abort)

Each occurrence was found independently, by a human or a multi-agent review pass noticing the pattern by eye — nothing in CI or `clippy` flags a raw `Vec::with_capacity`/`String::with_capacity` (or an implicit `TrustedLen` `collect`) whose argument is a multiplication of two or more independent lengths, or a scalar read from generator/document-controlled input. #1666 itself demonstrates the detection gap concretely: its first commit patched only `eval.rs`'s copies of `eval_index_expr`/`eval_slice_expr`, and needed a *second* commit, prompted by a second multi-agent review pass, to catch that `eval_generic.rs` independently reimplements both functions (the actual CLI dispatch path) with the identical unguarded pattern.

## Suggested fix direction (needs its own scoping pass)

A `clippy::disallowed_methods` (or similar) lint entry scoped to `src/jq/` banning a bare `Vec::with_capacity`/`String::with_capacity` call would turn this into a compile-time gate instead of a review-time hope. This needs its own scoping work:
- Enumerate the legitimate single-length (or length-times-constant) call sites that would need a per-call-site `#[allow(...)]` or a rewrite to use `try_reserve_product`/an equivalent single-factor helper (`json/standard.rs`, `output.rs`, `yq_runner.rs`, etc. all have benign `with_capacity(x.len())` calls today).
- Decide whether the lint should be `src/jq/`-wide or narrower (e.g. only `eval.rs`/`eval_generic.rs`, where the generator/cross-product pattern actually recurs).
- Consider whether `try_reserve_product`'s hardcoded "computed-index expansion" error wording (`src/jq/eval.rs`) should be parameterized (a `&str`/enum label) before other call sites (e.g. `combinations(n)` per #1669) start reusing it, so the error text doesn't mislead for a differently-shaped caller.

A separate, larger structural question raised in the same review pass but explicitly out of scope for a lint-only fix: `eval.rs` and `eval_generic.rs` maintain independent, parallel reimplementations of `eval_index_expr`/`eval_slice_expr` (and likely other evaluator logic) with no mechanism keeping them in sync — this is the root cause #1666 needed a second commit to fully close. Whether to consolidate these two evaluators (or add some other cross-checking mechanism) is a much bigger question than this issue's own scope; noted here for visibility, not as this issue's own acceptance criterion.

## Related

#1017, #1612, #1634 (closed by #1666), #1669 — the four known instances of this bug shape.

newhoggy:
## Summary

Fixes #1670.

A bare `Vec::with_capacity`/`String::with_capacity` sized by a product of
independently-controlled lengths has caused four independent
infallible-allocation bugs, each found only by a human or review pass
noticing the pattern by eye — nothing in CI catches it: #1017, #1612,
#1634, #1669.

- `clippy.toml` adds a `clippy::disallowed_methods` entry banning the raw
  `Vec::with_capacity`/`String::with_capacity` methods crate-wide.
- Scoped narrower than "all of `src/jq/`" per the issue's own suggested
  option: `src/lib.rs`, the CLI binary's crate root, and every
  `tests/`/`benches/`/`examples/` file that calls `with_capacity` opt back
  out (`#![allow(clippy::disallowed_methods)]`) — all of those sites size
  from a single collection's own length (or a length-times-constant), the
  same shape as every other benign call site the four incidents showed up
  next to.
- `src/jq/eval.rs` and `src/jq/eval_generic.rs` — the two files every
  historical incident occurred in — re-enable the lint
  (`#![warn(clippy::disallowed_methods)]`) and route their ~37 legitimate
  single-length call sites through new `vec_with_capacity`/
  `string_with_capacity` helpers, leaving the raw method free to flag a
  *future* unguarded product allocation instead of blending in.
- The four already-fixed sites (#1017/#1612/#1634/#1669) already use
  `try_reserve_product`/`try_reserve_exact`, not raw `with_capacity` —
  unaffected by this change.
- Added `test_pivot_array_of_objects_via_cli` (`tests/yq_cli_tests.rs`):
  found while renaming `eval_generic.rs`'s `Pivot` branch that its
  array-of-objects transpose had no CLI-level test anywhere in the suite
  (only `eval.rs`'s own unit tests, which the CLI never reaches).

## Verification

Live-verified the lint actually fires: temporarily added
`Vec::with_capacity(a * b)` inside `eval.rs`, confirmed
`cargo clippy --all-targets --all-features -- -D warnings` catches it with
the intended message, then reverted (confirmed byte-identical restore)
before committing.

## Test plan

- [x] Patch coverage: 100% (42/42 new lines)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -277,6 +277,18 @@ all      = { level = "warn", priority = -1 }
 pedantic = { level = "warn", priority = -1 }
 nursery  = { level = "warn", priority = -1 }
 
+# --- Restriction lints, opted in only where the pattern has actually
+# recurred ------------------------------------------------------------------
+# `disallowed_methods` is a restriction lint: allowed crate-wide by default,
+# matching the group priority scheme above, so a module opts back in with a
+# localized `#![warn(clippy::disallowed_methods)]` -- the same "closed by
+# default, reviewed opt-in elsewhere" shape `unsafe_code` uses above in
+# `[lints.rust]`, just inverted (there the group is `deny` and modules opt
+# *out*; here the group is `allow` and modules opt back *in*). See
+# `clippy.toml` for what's disallowed and why (#1670); `src/jq/eval.rs`/
+# `eval_generic.rs` are the only two files that opt back in today.
+disallowed_methods = "allow"
+
 # --- Numeric casts -----------------------------------------------------------
 # Pervasive and intentional in bit-manipulation / SIMD / indexing code, where
 # narrowing and sign reinterpretation are the whole point. These are guarded by
--- a/benches/balanced_parens.rs
+++ b/benches/balanced_parens.rs
@@ -5,6 +5,13 @@
 //! - Various tree depths and sizes
 //! - find_close, find_open, enclose operations
 
+// #1670: `clippy.toml`'s `disallowed-methods` bans a bare `Vec`/`String`
+// `with_capacity` crate-wide (re-enabled only in `succinctly::jq::eval`/
+// `eval_generic`) -- every call site here sizes from a single collection's
+// own length (or a length-times-constant) and was never part of that bug
+// shape.
+#![allow(clippy::disallowed_methods)]
+
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use rand::{RngExt, SeedableRng};
 use rand_chacha::ChaCha8Rng;
--- a/benches/balanced_parens.rs
+++ b/benches/balanced_parens.rs
@@ -5,13 +5,6 @@
 //! - Various tree depths and sizes
 //! - find_close, find_open, enclose operations
 
-// #1670: `clippy.toml`'s `disallowed-methods` bans a bare `Vec`/`String`
-// `with_capacity` crate-wide (re-enabled only in `succinctly::jq::eval`/
-// `eval_generic`) -- every call site here sizes from a single collection's
-// own length (or a length-times-constant) and was never part of that bug
-// shape.
-#![allow(clippy::disallowed_methods)]
-
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use rand::{RngExt, SeedableRng};
 use rand_chacha::ChaCha8Rng;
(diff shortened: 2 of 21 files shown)


newhoggy:
## Update (post code-review)

Two fixes from `/code-review`, both applied:

1. **16 file-level `#![allow(clippy::disallowed_methods)]` copies → one `Cargo.toml` line.** Two independent review passes pointed out `Cargo.toml`'s own `[lints.clippy]` table already exists for exactly this pattern (the `unsafe_code` precedent in `[lints.rust]`, inverted). `disallowed_methods = "allow"` there replaces every one of the 16 file-level attributes; `eval.rs`/`eval_generic.rs`'s own `#![warn(...)]` still overrides it correctly (verified live).

2. **Extended the lint to `Vec::resize`/`resize_with`/`reserve`/`reserve_exact` and `String::reserve`/`reserve_exact`.** A review pass found that #1017's own panic was actually at `Vec::resize`, not `with_capacity` — my original lint didn't cover it at all. The one pre-existing `resize` call in `eval.rs` (`pad_with_nulls`) is already correctly guarded by a `try_reserve` just above it, so it gets a one-line `#[allow]` with a comment rather than a rewrite. Verified live with a temporary `resize(a * b, ...)` probe.

Full test suite, all 3 clippy legs, and no_std build all still pass; patch coverage 100%.

PR #1797, opened 2026-08-29
fix(jq): plug decode-failure corruption in the format-function family

newhoggy:
## Summary

Part of #1755 -- this slice covers the format-function family
(`eval_format`) only; #1755 stays open, tracking the remaining sites
(recurse/walk/tostream, indirect comparison-key sites, and the misc
"direct" bucket).

`eval_format` is the single materialization point every `@format`
function (`@json`, `@csv`, `@tsv`, `@dsv`, `@base64`, `@base64d`, `@html`,
`@sh`, `@uri`, `@urid`, `@yaml`, `@props`, `@text`) shares via
`format_owned` -- it still called plain `to_owned`, so an undecodable
string silently formatted as `""` (or a container holding `""` for a
nested field) instead of raising `EvalError::decode_failure`. Converted to
`to_owned_checked` (established by #1746/#1754/#1794), propagating the
error through `QueryResult::Error` via the existing
`From<EvalError> for QueryResult` conversion.

Because every format function shares this one conversion point, fixing
`eval_format` closes the vulnerability for the entire family in one
surgical change, rather than needing a per-format-function fix.

## Test plan

- [x] New regression tests: `test_format_raises_on_decode_failure_1755`
      (loops over `@text`/`@json`/`@base64`/`@uri`/`@csv`/`@sh`/`@html`),
      `test_format_valid_data_unaffected_1755` -- verified the first test
      actually depends on the fix (temporarily reverted, confirmed it fails
      with the exact silent-corruption output `[1,""]` the issue describes,
      restored)
- [x] Patch coverage: 100% (23/23 new lines)

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -42310,6 +42310,41 @@ mod tests {
         );
     }
 
+    /// #1755: `eval_format`'s `to_owned_checked` runs *before* any
+    /// format-specific type/shape check, so a decode failure now preempts
+    /// an error that would have fired regardless of the string's content
+    /// (`@csv` on an object always errors; yq's `@uri`/`@base64`/`@html`
+    /// always reject a container) -- matching the established rule that a
+    /// decode failure is never suppressed, here extended to "never
+    /// outranked by an unrelated type error" (`scalar_decode_failure`'s
+    /// own doc comment states the same precedence for the sort family).
+    /// Pinned in both modes since `format_owned`'s yq-specific container
+    /// rejection (`stringify_for_format`) is the one dispatch arm that
+    /// could plausibly have raced with this ordering.
+    #[test]
+    fn test_format_decode_failure_outranks_shape_error_1755() {
+        // jq: `{"a": ...} | @csv` always errors -- @csv never accepts an
+        // object, content aside. With a decode failure inside it, that
+        // decode failure now fires first, not the "not both arrays" shape
+        // error.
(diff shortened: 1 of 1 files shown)


newhoggy:
## Update (post code-review)

`/code-review` found that `eval_format`'s new `to_owned_checked` call runs *before* any format-specific type/shape check — so a decode failure buried in a value now preempts an error that would fire regardless of content (`@csv` on an object always errors; yq's `@uri`/`@base64`/`@html` always reject a container). This was untested in either mode.

Live-verified this is the actual, consistent behavior in both jq and yq mode (both now raise `decode_failure` where they used to raise the format-specific shape error), and that it matches the established rule already in this file (`scalar_decode_failure`'s own doc comment: a decode failure is never suppressed/outranked). Added `test_format_decode_failure_outranks_shape_error_1755`, pinning this in both `query!` (jq) and `yq_query!` (yq) — verified it depends on the fix (reverted `eval_format`, confirmed the test fails with the old `{"a":""}` corrupted-and-wrong-error output, restored).

Other findings (key-drop on an undecodable object key, no shared `owned_or_err!`-style macro, the broader "should `to_owned` itself become fallible" architecture question) are pre-existing, already-documented gaps or recurring cross-PR observations already tracked by #1755's own scope — not addressed in this slice.

Full test suite, all clippy legs, and no_std build still pass; patch coverage 100%.

Issue #1319, opened 2026-08-20
yaml: explicit_tag only follows one alias hop, silently loses tag-driven type coercion on a 2+-hop chain


newhoggy:
## Summary

`YamlCursor::explicit_tag` (`src/yaml/light.rs`) resolves an alias's explicit tag (`!!str`, `!!int`, ...) via `target.and_then(|t| t.index.get_tag(t.bp_pos))` — the **immediate** target only, one hop. If the alias chain is 2+ hops deep before reaching the tagged node, the tag lookup silently misses and callers (`type`, arithmetic, anything routing through `to_owned_cursor`/`tagged_type_name` in `src/jq/eval_generic.rs`, per issue #747/#224) fall back to inferring the type from content instead of honoring the explicit tag.

## Repro (confirmed live against current `main`)

```yaml
k0: &a0 !!int "5"
k1: &a1 *a0
z: *a1
```

```console
$ succinctly yq '.k1 | type' file.yaml   # 1 hop
"number"
$ succinctly yq '.z | type' file.yaml    # 2 hops
"string"
$ succinctly yq '.z + 1' file.yaml       # 2 hops
1
```

`.k1` (one hop from the tagged node) correctly resolves `!!int "5"` to the number `5` (`type` → `"number"`). `.z` (two hops away) loses the tag entirely — `type` falls back to inferring from the quoted string content (`"string"`), and arithmetic (`.z + 1`) treats it as effectively absent/null-ish rather than the number 5 (`1`, not `6`).

Note the JSON *streaming* output path resolves correctly at any depth (`succinctly yq -o json '.z' file.yaml` → `5`) — this divergence is specific to `explicit_tag()`'s own callers, not YAML→JSON streaming in general.

## Why this is separate from #1193 and its other follow-ups

Not a stack-overflow/crash vector — `explicit_tag` doesn't call itself, it's bounded to exactly one hop (same shape as `as_str`/`key_string`'s #1315 gap and `merge_sources`'s #1318 gap). This is a pure correctness bug: a chained alias to an explicitly-tagged scalar silently loses the tag, causing wrong type inference and wrong arithmetic results.

## Suggested fix

[...]

PR #1798, opened 2026-08-29
perf(jq): use O(1) is_falsy() instead of materializing select()'s cond

### LINKED ISSUE #1645: perf(jq/yq): select(cond)'s truthy check pays a full to_owned_cursor deep-copy per output instead of an O(1) is_falsy()
**Severity: Low**

## Summary

Found by the multi-agent `/code-review` run on PR #1644 (#1613's fix). `eval_generic.rs`'s
`push_generic_truthiness` (line 1682) computes each `cond` output's truthiness via
`to_owned_cursor(&c).is_truthy()` in its `OneCursor`/`ManyCursor` arms — a full recursive
materialization of the value (arbitrarily deep for a nested object/array), just to answer a
question `DocumentCursor::is_falsy()` already answers in O(1) without materializing anything
(doc comment at `document.rs:332-338`: "Used for `--exit-status` flag handling without requiring
full materialization"; already used elsewhere in this same file, e.g. `stats.last_was_falsy`).

**This is pre-existing** — confirmed via `git show origin/main:src/jq/eval_generic.rs`, the
`to_owned_cursor`-based arms predate #1613 entirely and apply to *both* jq and yq modes,
anywhere a multi-output condition fans out over cursors (`select(.[])` on an array/object of
non-trivial nested values is the direct trigger). #1613 doesn't introduce this cost, but it does
make it more visible: `SELECT_EMITS_ONCE_IF_ANY_TRUTHY`'s yq-mode collapse
(`usize::from(bits.iter().any(...))`, `eval_generic.rs` ~5462) only reduces the already-fully-
materialized `bits: Vec<bool>` — by the time it runs, the expensive part (deep-copying every
`cond` output) has already happened regardless of whether the very first output was already
truthy and the rest of the count is moot.

`eval.rs`'s `push_truthiness`/`eval_fanout`/`resolve_cond_fork` are cheaper by comparison —
`json_is_truthy`/`OwnedValue::is_truthy` don't deep-copy — but #1613's new `select_emits`
gate (added in review, see the follow-up commit on #1644) still invokes `body(bit)`/
`dispatch(bit, out)` once per *remaining* bit in yq mode even after the collapse outcome is
already fixed, each call now provably a no-op. Minor by comparison (a branch, no allocation),
but the same shape.

## Why this can't just short-circuit `cond`'s evaluation entirely

Real yq's `select` does **not** stop evaluating `cond` after the first truthy output — confirmed
live against yq v4.53.3: `select((true, error("boom")))` still raises `boom`, meaning the second
comma branch is evaluated (and its side effect/error observed) even though the first branch
already settled "truthy: yes". So any fix here can only cheapen *how* each output's truthiness is
checked (is_falsy() instead of to_owned_cursor), not skip evaluating any of `cond`'s outputs —
`cond` must still run to completion for correctness (side effects, and any error/break it raises).

## Suggested fix direction

- Swap `push_generic_truthiness`'s `OneCursor`/`ManyCursor` arms to use
  `DocumentCursor::is_falsy()` instead of `to_owned_cursor(...).is_truthy()`, avoiding the
  materialization while still visiting every output (for side-effect/error correctness).
- Once that's cheap, reconsider whether `eval.rs`'s three `select_emits` call sites should also
  skip the now-redundant `body(bit)`/`dispatch(bit, out)` invocation for bits after the collapse
  outcome is fixed (upstream `cond` evaluation is unaffected either way, so this part is purely
  about not calling a closure that's already known to return `None`).
- Needs the project's A/B hardware-benchmark discipline (`docs/guides/benchmarking.md`) before
  landing, on both a workload where the fast path helps (`select(.[])` over a large array/object
  of deep values, first element truthy) and one where it can't (every output needs checking
  anyway) — per this repo's own recorded lesson that a precheck's cost has to be measured against
  the workload where the guarded code is already fastest, not just the one it's slowest on.

## Why this is separate from #1613

#1613 is a correctness fix (wrong output count); this is a performance-only gap, pre-existing
in the `to_owned_cursor` portion and only cosmetically new in the `select_emits`-loop portion.
Landing it needs hardware benchmarking per this repo's standing discipline, which doesn't belong
in a correctness-scoped, unattended-session PR.

## Root cause

- `src/jq/eval_generic.rs:1682` (`push_generic_truthiness`, `OneCursor`/`ManyCursor` arms) —
  pre-existing, applies to both modes.
- `src/jq/eval.rs:3633` (`select_emits`, added by #1613) — the three call sites
  (`builtin_select`, `resolve_node`'s `Select` arm, `eval_pipe_with_path_context_internal`'s
  `Select` arm) still invoke the per-bit closure once per remaining bit in yq's collapsed case.

newhoggy:
## Summary

Found by the multi-agent `/code-review` run on PR #1644 (#1613's fix). `eval_generic.rs`'s
`push_generic_truthiness` computed each `cond` output's truthiness via
`to_owned_cursor(&c).is_truthy()` in its `OneCursor`/`ManyCursor` arms -- a full recursive
materialization of the value, arbitrarily deep for a nested object/array, just to answer a
question `DocumentCursor::is_falsy()` already answers in O(1) without materializing anything
(already used elsewhere in this same file, e.g. `--exit-status`'s `stats.last_was_falsy`).

`cond` must still run to completion for every output (side effects, any error/break it raises)
-- real yq's own `select` does **not** short-circuit after the first truthy output either
(confirmed live against yq v4.53.3 in the issue: `select((true, error("boom")))` still raises
`boom`) -- so this only changes *how* each output's truthiness is checked, not how many
outputs `cond` produces or visits.

## Change

Swapped `push_generic_truthiness`'s `OneCursor`/`ManyCursor` arms from
`push_or_control!(to_owned_cursor(&c)).is_truthy()` to `!c.is_falsy()`.

## Benchmarks

Per this repo's own A/B discipline (`docs/guides/benchmarking.md`), on both reference machines
(Apple M4 Pro/NEON, AMD Ryzen 9 7950X/AVX-512), interleaved A/B, two independent 15-rep runs
each, output-identity-gated on all 22 (machine x workload x size) cells before trusting any
timing:

**Deep-nesting workload** (`select(.[])` over N objects, 7 levels deep, the case this targets):

| N | M4 Pro (before -> after) | 7950X (before -> after) |
|---|---|---|
| 1,000 | 24.59ms -> 14.05ms (**-42.9%**) | 27.02ms -> 15.55ms (**-42.5%**) |
| 10,000 | 244.48ms -> 127.55ms (**-47.8%**) | 279.88ms -> 156.22ms (**-44.2%**) |
| 100,000 | 2970.28ms -> 1551.80ms (**-47.8%**) | 3058.72ms -> 1696.28ms (**-44.5%**) |

Flat percentage across N (nesting depth is constant, only absolute time scales) -- the expected
signature of eliminating a fixed per-element materialization cost, not an algorithmic-complexity
change.

**Shallow-scalar workload** (the case the issue expected to be a wash -- every output already
"cheap" to materialize): turns out `to_owned_cursor`'s scalar path wasn't actually free either
(an `explicit_tag()`/`canonicalize_numbers()` check plus a string allocation or number-literal
parse, work `is_falsy()`'s pure type-tag check skips entirely). At N>=100,000: **20-27% faster
on ARM, 10-11% faster on x86_64** (architecture-dependent, not further explained here); at
N=1,000 the delta is 1-4%, indistinguishable from the measured noise floor (<=1.9% both
machines). **No regression in any of the 8 shallow cells, either architecture.**

Noise floor (control run, before-binary vs itself): 0.63% (M4 Pro), 1.91% (7950X).

## Test plan

- [x] A/B benchmarked as above -- real win on the target workload, no regression on the
      "can't help" workload, both architectures, output-identical on all 22 cells
- [x] Manually verified `select` semantics: basic truthy/falsy filtering, deep nested
      objects/arrays, yq-mode object fan-out, and that `cond`'s side effects/errors are still
      fully evaluated (`select((true, error("boom")))` still raises)
- [x] Full `cargo test --features cli,simd,regex,serde` -- 55/55 test binaries pass

## Not addressed here (filed as follow-up scope, per the issue's own split)

`eval.rs`'s `select_emits` (added by #1613) still invokes the per-bit closure once per
*remaining* bit in yq's collapsed-select case even after the collapse outcome is already fixed
-- a smaller, cosmetic version of the same shape (a branch, no allocation, no materialization),
left for a separate pass since it's orthogonal to this fix's actual measured win.

Fixes #1645

### DIFF
--- a/src/jq/document.rs
+++ b/src/jq/document.rs
@@ -974,10 +974,11 @@ pub fn colliding_display_key_error(key: &str) -> EvalError {
 /// `colliding_display_key_error` instead of allowing a silent overwrite.
 ///
 /// Shared by `eval_generic.rs`'s three `to_owned*_at_depth` conversions,
+/// its validate-only `push_generic_truthiness_cursor_error` (#1645),
 /// `lazy.rs`'s `cursor_to_owned_at_depth`, and `succinctly-cli`'s
 /// `yq_runner.rs` `--input-format json` bridge (`pub` for that reason,
 /// same as [`DisplayKeyGuard`] and [`key_display_string`] before it) -- one
-/// definition rather than a fifth hand-copied guard-and-raise sequence.
+/// definition rather than a sixth hand-copied guard-and-raise sequence.
 pub fn resolve_display_key<V: DocumentValue, T>(
     key: &V,
     map: &IndexMap<String, T>,
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -1780,14 +1780,25 @@ fn push_generic_owned_values<V: DocumentValue>(
 /// This is [`to_owned_cursor_at_depth`]'s own traversal and validation
 /// (same object/array unconsing, same key-collision guard, same checks
 /// in the same order once a scalar is reached) -- but it builds no
-/// `OwnedValue` anywhere, only walking and checking, so `select`'s
-/// truthiness check keeps its #1645 performance win (no `String`/`Vec`
-/// value allocation) for the overwhelmingly common case where nothing is
-/// corrupted, while still raising on a value corrupted at *any* depth
-/// below `c`'s own top level -- not just when `c` itself is the bad
-/// scalar (an earlier version of this fix only checked `c.value()`
-/// directly, silently missing anything nested inside a well-formed
-/// container; caught by review, #1645).
+/// `OwnedValue` container or scalar payload anywhere, only walking and
+/// checking, raising on a value corrupted at *any* depth below `c`'s own
+/// top level -- not just when `c` itself is the bad scalar (an earlier
+/// version of this fix only checked `c.value()` directly, silently
+/// missing anything nested inside a well-formed container; caught by
+/// review, #1645).
+///
+/// For a scalar condition this keeps #1645's O(1) win in full: neither
+/// this walk nor `is_falsy()` afterward materializes anything. For a
+/// container condition, jq's own truthiness rule needs none of this --
+/// any object/array is unconditionally truthy regardless of contents --
+/// but this walk still visits the whole subtree anyway, because a
+/// corrupted value can be anywhere inside it and the #1247/#1194
+/// invariant this function exists to enforce is a property of the whole
+/// subtree, not of the truthiness question alone. That walk still
+/// allocates one `String` per object key (`resolve_display_key`'s own
+/// #1642 guard), so "no allocation" only ever describes the *value*
+/// side (no `OwnedValue`/`Vec` payload) -- not a claim that a
+/// container-shaped condition is free.
 fn push_generic_truthiness_cursor_error<C: DocumentCursor>(c: &C, depth: usize) -> Option<Control> {
     assert_nesting_depth(depth);
     let value = c.value();
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -1768,6 +1768,30 @@ fn push_generic_owned_values<V: DocumentValue>(
     None
 }
 
+/// Whether `c`'s value cannot be honestly answered "truthy or falsy" at
+/// all -- a decode failure (a string token whose bytes don't decode) or a
+/// structural error (a token the semi-index accepted as a span but
+/// couldn't classify, #1194) -- and if so, the `Control::Error` a caller
+/// should raise instead of falling through to
+/// [`DocumentCursor::is_falsy`]'s own silent "not falsy" default for
+/// exactly these two cases. Mirrors [`to_owned_at_depth`]'s identical pair
+/// of checks (same order, same messages), which `is_falsy` bypasses since
+/// it never inspects either (#1645 review).
+fn push_generic_truthiness_cursor_error<C: DocumentCursor>(c: &C) -> Option<Control> {
+    let v = c.value();
+    if let Some(reason) = v.string_decode_error() {
+        return Some(Control::Error(EvalError::decode_failure(reason)));
+    }
+    if v.is_error() {
+        return Some(Control::Error(EvalError::new(
+            v.error_message()
+                .unwrap_or("malformed value in document")
+                .to_string(),
+        )));
+    }
+    None
+}
+
 /// Append one truthiness bit per output of a `GenericResult` stream to
 /// `out`. Mirrors [`super::eval::push_truthiness`] for the generic
 /// evaluator's cursor-aware result type — used to fan `select`'s condition
@@ -1781,8 +1805,17 @@ fn push_generic_truthiness<V: DocumentValue>(
         // `DocumentCursor::is_falsy` answers this in O(1) without
         // materializing the value at all -- an arbitrarily deep object/
         // array previously paid a full recursive `to_owned_cursor` copy
-        // just to learn it isn't `null`/`false` (#1645).
+        // just to learn it isn't `null`/`false` (#1645). `is_falsy` itself
+        // is deliberately silent on a value that fails to decode (its own
+        // doc comment: "conservative assumption", matching `--exit-status`'s
+        // pre-existing, best-effort use of it) -- `select`'s own filtering
+        // needs the real #1247/#1194 raise instead, so that check runs
+        // first, same as `to_owned_at_depth`'s own two checks in the same
+        // order, just without materializing on the ordinary-value path.
         GenericResult::OneCursor(c) => {
+            if let Some(control) = push_generic_truthiness_cursor_error(&c) {
+                return Some(control);
+            }
             out.push(!c.is_falsy());
         }
         GenericResult::Many(vs) => {
@@ -1792,6 +1825,9 @@ fn push_generic_truthiness<V: DocumentValue>(
         }
         GenericResult::ManyCursor(cs) => {
             for c in &cs {
+                if let Some(control) = push_generic_truthiness_cursor_error(c) {
+                    return Some(control);
+                }
                 out.push(!c.is_falsy());
             }
         }
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -1768,28 +1768,73 @@ fn push_generic_owned_values<V: DocumentValue>(
     None
 }
 
-/// Whether `c`'s value cannot be honestly answered "truthy or falsy" at
-/// all -- a decode failure (a string token whose bytes don't decode) or a
-/// structural error (a token the semi-index accepted as a span but
-/// couldn't classify, #1194) -- and if so, the `Control::Error` a caller
-/// should raise instead of falling through to
+/// Whether `c`'s subtree, at any depth, contains a value that cannot be
+/// honestly answered "truthy or falsy" at all -- a decode failure (a
+/// string token whose bytes don't decode), a structural error (a token
+/// the semi-index accepted as a span but couldn't classify, #1194), or a
+/// #1642 colliding-decode-failure-key -- and if so, the `Control::Error`
+/// a caller should raise instead of falling through to
 /// [`DocumentCursor::is_falsy`]'s own silent "not falsy" default for
-/// exactly these two cases. Mirrors [`to_owned_at_depth`]'s identical pair
-/// of checks (same order, same messages), which `is_falsy` bypasses since
-/// it never inspects either (#1645 review).
-fn push_generic_truthiness_cursor_error<C: DocumentCursor>(c: &C) -> Option<Control> {
-    let v = c.value();
-    if let Some(reason) = v.string_decode_error() {
-        return Some(Control::Error(EvalError::decode_failure(reason)));
-    }
-    if v.is_error() {
-        return Some(Control::Error(EvalError::new(
-            v.error_message()
+/// exactly these cases.
+///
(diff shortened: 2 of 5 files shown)


newhoggy:
\`/code-review\` found a real, confirmed correctness regression in the first commit: swapping to \`DocumentCursor::is_falsy()\` for \`select\`'s truthiness check is silent on a value that fails to decode or is structurally malformed (\`is_falsy()\`'s own doc comment: "conservative assumption" -- it's designed for \`--exit-status\`'s existing best-effort use, not for \`select\`'s own filtering correctness), where the removed \`to_owned_cursor\` materializing path used to raise via \`string_decode_error()\`/\`is_error()\` -- exactly the #1247/#1194 class of bug this codebase treats as non-negotiable.

Confirmed live by the reviewer (and reproduced myself): \`printf '["\x"]' | succinctly jq '.[] | select(.)'\` went from raising \`invalid escape sequence in string\` (exit 5) to silently printing the value at exit 0.

Fixed in a follow-up commit: added \`push_generic_truthiness_cursor_error\`, checking the same two conditions \`to_owned_at_depth\` already checks (same order, same messages) before falling through to the O(1) \`is_falsy()\` path -- so a decode failure or #1194 structural error still raises, and only a genuinely representable value takes the fast, non-materializing path. Added two CLI-level regression tests pinning both cases (decode failure and structural error).

The added checks are the same cheap enum-discriminant reads \`to_owned_at_depth\` already performs as its own first two checks -- no allocation, no recursion -- so the original benchmark's conclusions hold in kind; I spot-checked locally (stable ~0.09-0.10s for the 10K deep-nesting workload, consistent with the original remote numbers) rather than re-running the full remote A/B suite for a one-line addition to an already-cheap path.

newhoggy:
Round 3, in response to the last review pass finding two independently-confirmed gaps in round 2's single-level check:

1. **Nested decode-failure/structural-error still slipped through.** `push_generic_truthiness_cursor_error` only checked the condition cursor's own top-level value — a decode failure or #1194 structural error one level (or more) inside a container (e.g. `select(.bad) | .keep` where `.bad` is `["\x"]`) still silently passed `is_falsy()`'s "not falsy" default. Rewrote it as a full recursive walk mirroring `to_owned_cursor_at_depth`'s own traversal (same object/array unconsing, same #1642 key-collision guard, same checks once a scalar is reached) — but it builds no `OwnedValue`, so the O(1) win for the ordinary case is unaffected; it only allocates on the path that's about to raise anyway. Verified live against both a nested decode failure and a nested structural error, plus the original flat cases and the `select(.)`-over-whole-document shape a reviewing subagent flagged.

2. **YAML `is_falsy()` had no `Alias` arm.** Newly load-bearing once `select` started calling it directly: an alias resolving to `false`/`null` was reported truthy regardless of target. Fixed by resolving via the existing `resolve_alias_target_cursor` and recursing, matching `string_decode_error`'s own precedent in the same file. Verified against a direct alias, a 2-hop chain, and confirmed all four cases match `main`'s pre-#1645 output exactly.

Also strengthened the two existing regression tests to assert the exact exit code (`5`, matching this file's #1194 convention) and exact message text rather than just "nonzero/nonempty", and added nested-in-container cases for both failure kinds plus a yq alias-falsiness test.

Verification this round: full `cargo test --features cli,regex` (55/55 green), both CI clippy legs (`--all-features` and the `broadword-yaml` blind-spot leg), `cargo fmt --check`, `cargo build --no-default-features` (no_std), `cargo doc --all-features`, all after rebasing onto current `main`.

Note on a deliberately out-of-scope finding: a reviewing subagent also flagged that `GenericResult::stream_json`'s raw-byte-echo path doesn't decode-validate output independent of `select` — confirmed this is pre-existing `main` behavior (`succinctly jq '.foo'` on an untouched sibling field with a bad escape already echoes it verbatim on unmodified `main`), unrelated to this diff's blast radius. Not fixing here; flagging in case it's worth its own issue.

Requesting another `/code-review` pass on this round.

newhoggy:
Round 4, addressing the six-angle review of round 3:

**Fixed:**
- `is_error()`/`error_message()` on `YamlValue` were the only typed accessors that didn't resolve through an `Alias` chain (every sibling does). Added matching `Alias` arms. Latent in practice (neither `YamlValue::Error` construction site is reachable behind a real alias), but now consistent with every sibling accessor in the same impl block.
- `is_falsy()`'s new `Alias` arm used `is_some_and` (dangling target → not falsy); `is_null()`'s established convention is `map_or(true, ...)` (dangling target → null-like → falsy). Aligned to match — also unreachable in practice (`YamlIndex::build` refuses unresolvable aliases, #372), but now consistent.
- Tightened the "no allocation" doc-comment claim: it only ever covered the *value* side. A container-shaped condition still allocates one `String` per key (the shared #1642 guard) and still walks the whole subtree — not because jq's truthiness rule needs it (a container is unconditionally truthy regardless of contents), but because the #1247/#1194 invariant this function enforces is a property of the whole subtree, independent of what jq's own O(1) container-truthiness shortcut would answer.
- Updated `resolve_display_key`'s doc comment to count this as a sixth call site.
- Added `test_select_and_materialize_agree_on_corruption_1645`, checking `select(.bad) | .keep`'s new walk and `-Sc .bad`'s pre-existing materializing walk raise on the same input across six malformed shapes. This is the guard CLAUDE.md's own "duplicated predicates diverge silently" rule (#106) asks for — pinning that the two independent implementations agree, not just that each is individually correct.

**Filed as follow-ups, deliberately not fixed here:**
- #1319 (pre-existing) — added a comment with a new confirmed symptom: `explicit_tag()`'s one-hop alias limit now visibly disagrees with `select()` on a 2+-hop tagged alias (`select()` is more correct post-#1645; `==`/`type`/materialization still aren't). Root cause predates and is broader than this PR.
- #1803 — the traversal-with-validation walk is now hand-copied across 5 call sites in the crate; a shared generic core is the right fix but a separate refactor, not a bug fix.
- #1804 — `as_object`/`as_array`'s unmemoized alias-chain resolution can amplify a deeply-aliased document exponentially; confirmed pre-existing (inherited from `to_owned_cursor_at_depth`, not introduced here).

Re-verified this round: full `cargo test --features cli,regex` (55/55 green), both CI clippy legs, `cargo fmt --check`, `cargo build --no-default-features`, `cargo doc --all-features`, and the alias-falsiness live sanity sweep from round 3 (all four cases still correct).

Requesting one more `/code-review` pass focused on the round-4 diff.

newhoggy:
Round 5, addressing the third review pass (7-angle fan-out, 4 findings converged from 3+ independent angles):

**Fixed:**
- `is_error()`'s dangling-alias-target default aligned to `true` (matching `is_null()`/`is_falsy()`'s convention for the same unreachable-per-#372 case) instead of `false`.
- `resolve_alias_target_cursor`'s doc comment updated for the 6th call site.
- `push_generic_truthiness_cursor_error`'s doc comment now names precisely which check it skips relative to `to_owned_cursor_at_depth` (tag/number-literal resolution) and why that's safe, instead of a blanket "same checks" claim.
- Added `test_select_raises_on_yaml_decode_failure_nested_in_container_1645` — every #1645 regression test was JSON-only; nothing pinned that `select()`'s corruption-detection walk actually works through the yq evaluator too (only alias-falsiness was covered on that side).
- Added a missing "structural error nested in object" case to the JSON cross-check table (had array but not object for that failure kind).

**Settled a live question from this round:** one reviewer measured `select(.data) | .keep` over a 2M-element array as ~2x slower on current HEAD than on this branch's *first* commit (`405dbcfe2`, the O(1)-only optimization with no validation at all — never shippable, since it silently swallowed corruption). That's true but the wrong baseline. Interleaved release-build A/B against genuine pre-#1645 `main` (output verified byte-identical): current branch ~1.2-1.4s vs. main's ~1.5-1.6s — still faster, not slower. The container-condition case nets a real, if smaller, win over what's actually shipped today; it only looks like a regression against a commit that was never correct enough to ship.

**Filed as follow-ups:**
- #1803 — the traversal-with-validation walk is now hand-copied across 5+ call sites (architectural refactor).
- #1804 — unmemoized YAML alias-chain resolution can amplify a deeply-aliased document exponentially (confirmed inherited unchanged from pre-#1645 `to_owned_cursor`, not introduced here).
- #1805 — `--exit-status` has the same decode-failure/structural-error silent-swallow gap `select()` used to have, across 5 call sites in `eval_generic.rs`'s streaming path and `yq_runner.rs`'s identity fast path. Confirmed live against real `yq` (which raises) and against this PR's own `select()`/materialization paths (which now raise). A distinct fix with its own testing surface — not a natural extension of this diff.
- Commented on the pre-existing #1319 (`explicit_tag`'s one-hop alias limit) with a new confirmed symptom: this diff's more-correct full-chain `is_falsy()` now visibly disagrees with `==`/`type`/materialization on a 2+-hop tagged alias, where before all of them agreed (all wrong).

Re-verified: full `cargo test --features cli,regex` (55/55 green), both CI clippy legs, `cargo fmt --check`, `cargo build --no-default-features`, `cargo doc --all-features`, all 7 of this issue's regression tests individually, and the alias-falsiness/perf sanity sweeps from prior rounds.

Given four review rounds have now converged on the same small set of items (each subsequent round reconfirming prior findings rather than surfacing new correctness bugs), I believe this is ready to merge. The remaining known-real gaps (#1319, #1803, #1804, #1805) are each independently scoped, filed, and cross-referenced rather than silently dropped.

PR #1799, opened 2026-08-29
fix(jq): plug decode-failure corruption in the indirect comparison-key family

newhoggy:
## Summary

Part of #1755 -- this slice covers the indirect comparison-key family
only; #1755 stays open, tracking the remaining sites (recurse/walk/
tostream, and the misc "direct" bucket).

- `builtin_in`/`builtin_upper_in` (`IN`): the input is used as a
  comparison key, not returned directly -- `to_owned` silently
  substituted `""`, potentially flipping the answer.
- `builtin_contains`/`builtin_inside`: the whole input is the
  containment operand.
- `search_pattern` (backs `index`/`indices`/`rindex`): each array element
  is converted for comparison against the pattern, inside a loop -- 3
  call sites.
- `bsearch_one_target`: the whole array is converted upfront for the
  binary search.
- `to_owned_key_shape` (a computed-index/slice-bound key's scalar
  fallback -- its Array/Object arms never touch `to_owned` at all, so
  this only affects the String case): made fallible, closing
  `eval_index_expr`'s and `eval_slice_bound`'s shared use of it, and
  incidentally yq-mode `join`'s own use of the same helper.

## Unrelated fix included: a test PR #1797's merge silently dropped

While starting this slice I found `main`'s merge commit for PR #1797 is
byte-identical to that PR's *first* local commit only -- the second
commit (a regression test added in response to `/code-review`, verified
correct and CI-green beforehand) never made it into `main`, with no
error or partial-merge indication anywhere. Restored it here as its own
commit since I was already touching this file's test module. Filed a
memory note on this incident; root cause not diagnosed, and it does not
appear to be systemic (checked two other recent multi-commit merges from
this session, both landed intact).

## Test plan

- [x] New regression tests: `test_comparison_key_family_raises_on_decode_failure_1755`
      (11 exprs across `in`/`IN`/`contains`/`inside`/`index`/`indices`/
      `rindex`/`bsearch`/computed-index/slice-bound/yq-mode `join`),
      `test_comparison_key_family_valid_data_unaffected_1755` -- verified
      the first depends on the fix (reverted all 8 sites, confirmed it
      fails with a silently-wrong `false` answer instead of raising,
      restored)
- [x] Patch coverage: 99.32% (146/147) -- the one uncovered line is
      `eval_slice_bound`'s `QueryResult::Many` arm success path. Confirmed
      live that a genuinely document-cursor-backed multi-output bound
      expression (`.bounds[]` alone returns `Many`) still arrives at this
      match as `ManyOwned` once embedded in a slice bound position,
      suggesting the outer slice-expression evaluator pre-materializes
      bounds to owned values upstream of this match -- did not chase the
      exact mechanism further given this doesn't affect correctness (the
      `Err` arm is covered, so a decode failure in a multi-bound
      expression still raises; only the already-narrow `Ok` success arm
      for that specific shape is untested).

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -36380,6 +36380,16 @@ mod tests {
             "join(\",\")",
             QueryResult::Error(e) if e.is_decode_failure() => {}
         );
+        // A *multi*-output computed-index key (`.keys[]` as the key
+        // expression) reaches `to_owned_key_shape`'s own `QueryResult::Many`
+        // arm, distinct from the single-key case above (`QueryResult::One`)
+        // -- one valid key, one malformed, to also exercise the `.collect()`
+        // short-circuit rather than only its first element.
+        query!(
+            &b"{\"a\":1,\"keys\":[\"a\",\"\xff\xfe\"]}"[..],
+            ".[(.keys[])]",
+            QueryResult::Error(e) if e.is_decode_failure() => {}
+        );
     }
 
     /// #1755 positive control: valid data through each of the comparison-key
@@ -36421,6 +36431,11 @@ mod tests {
         );
         yq_query!(br#"["x","y"]"#, "join(\",\")",
             QueryResult::Owned(OwnedValue::String(s)) => { assert_eq!(s, "x,y"); }
+        );
+        query!(br#"{"a":1,"b":2,"keys":["a","b"]}"#, ".[(.keys[])]",
+            QueryResult::Many(vs) => {
+                assert_eq!(vs.iter().map(to_owned).collect::<Vec<_>>(), vec![OwnedValue::Int(1), OwnedValue::Int(2)]);
+            }
         );
     }
 
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -42310,6 +42310,46 @@ mod tests {
         );
     }
 
+    /// #1755: `eval_format`'s `to_owned_checked` runs *before* any
+    /// format-specific type/shape check, so a decode failure now preempts
+    /// an error that would have fired regardless of the string's content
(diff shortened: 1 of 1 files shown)


newhoggy:
## Update (post code-review)

Two review findings addressed:

1. **`?`/`optional` precedence for `in`/`IN`/`contains`/`inside` was untested.** Added `test_comparison_key_family_optional_does_not_suppress_decode_failure_1755`, live-verified — the new decode-failure raise fires before `optional` is ever consulted, matching this file's established rule.

2. **A decode failure now preempts a sibling generator expression's own evaluation/side effects entirely** (e.g. `"<bad>" | in({"a":1}, error("boom"))` never reaches `error("boom")`). Live-verified this is real. This is the *pre-existing* "hoisted, pre-fanout materialization" structure in `builtin_in`/`builtin_contains`/`builtin_inside` (predates #1755 — the comment "hoisted out of the fan-out" was already there), just newly observable now that the hoisted conversion can fail. A proper fix needs the same `Partial`-preserving machinery this file uses for `Break`/`Halt` elsewhere, which is a materially bigger change than this slice's scope — filed as **#1800** rather than redesigning evaluation order here.

Also filed **#1801**: `/code-review` found `yaml_value_to_owned` (a `YamlCursor`-specific converter used by `load()`'s YAML path) has the identical decode-failure/key-drop corruption shape, under a name outside #1755's own `StandardJson`-scoped audit — a genuinely separate, same-shaped bug a `to_owned`-grep sweep would never surface.

Full test suite, clippy, and no_std build still pass; patch coverage 99.39% (one documented, structurally-hard-to-reach line unchanged from before this round).

PR #1802, opened 2026-08-29
fix(jq): plug decode-failure corruption in the direct-value misc family

newhoggy:
## Summary

Part of #1755 -- this slice covers the simpler "direct" bucket functions
only; #1755 stays open, tracking the remaining sites (recurse/walk/
tostream, and INDEX/getpath/load/combinations/debug/stderr/transpose/
shuffle/pivot).

- `builtin_add`, `builtin_flatten` (**#1755's own headline repro**:
  `[1, "\ud800"] | flatten` returns `[1,""]`), `builtin_from_entries`:
  array/object element materialization, plus their scalar fallback arm
  (now checks `scalar_decode_failure` before `optional`, matching the
  sort family's established precedent).
- `trim_edge` (backs `ltrimstr`/`rtrimstr`): both of its total-passthrough
  arms (non-string pattern, non-string value).
- `join_with_separator`'s jq-mode arm (the yq-mode arm was already fixed
  incidentally by #1799's `to_owned_key_shape` change).
- `builtin_last`, `builtin_reverse`.
- `builtin_tostring`, `builtin_tojson`.
- `builtin_pick`/`builtin_omit`: the object and array value-materialization
  arms (their own `keys_expr` computation is a separate, not-yet-touched
  concern, left for a later pass).

All converted from `to_owned` to `to_owned_checked` (established by
#1746/#1754/#1794/#1797/#1799), propagating `EvalError::decode_failure`
instead of silently substituting `""`.

## Test plan

- [x] `cargo test --features cli,regex --no-fail-fast` (full suite, 0
      failures -- one transient CLI-subprocess flake seen once,
      reproduced clean on immediate re-run, unrelated to this diff)
- [x] New regression tests: `test_direct_misc_family_raises_on_decode_failure_1755`
      (14 exprs across all 11 functions), `test_direct_misc_family_valid_data_unaffected_1755`,
      `test_direct_misc_family_scalar_fallback_1755` -- verified genuine
      dependency by reverting the fix (aggregate revert, plus individually
      re-verified `flatten` reproduces the issue's own exact headline
      symptom `[1,""]`, and `omit`'s array arm) before restoring
- [x] Patch coverage: 98.38% (182/185) -- the 3 remaining uncovered lines
      are single-statement `if optional { return QueryResult::None; }`
      bodies, confirmed functionally exercised (the test above passes,
      and the surrounding `if`/`}` lines show hits in raw lcov) but
      mis-attributed by `llvm-cov` -- the same known, already-documented
      quirk from #1670/#1794's own coverage rounds.

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -7617,7 +7617,12 @@ fn join_with_separator<W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         StandardJson::Object(mut fields) => fields.try_fold(None, |acc, f| {
             join_step::<S>(acc, to_owned_checked(&f.value())?, &sep)
         }),
-        _ => return QueryResult::Error(EvalError::cannot_iterate_with(S::TAG, &to_owned(&value))),
+        _ => {
+            if let Some(e) = scalar_decode_failure(&value) {
+                return QueryResult::Error(e);
+            }
+            return QueryResult::Error(EvalError::cannot_iterate_with(S::TAG, &to_owned(&value)));
+        }
     };
 
     match result {
@@ -7792,11 +7797,23 @@ fn builtin_first<W: Clone + AsRef<[u64]>>(
         },
         // jq: null | first => null
         StandardJson::Null => QueryResult::Owned(OwnedValue::Null),
-        _ if optional => QueryResult::None,
-        _ => QueryResult::Error(EvalError::cannot_index_with_type(
-            type_name(&value),
-            "number",
-        )),
+        // #1755: a decode failure on the scalar itself must raise
+        // unconditionally, checked ahead of `optional` -- see
+        // `scalar_decode_failure`. Found alongside `last`'s identical
+        // fallback arm, though `first` was not itself in scope for this
+        // sweep (its own Array arm never materializes at all).
+        _ => {
+            if let Some(e) = scalar_decode_failure(&value) {
+                return QueryResult::Error(e);
+            }
+            if optional {
+                return QueryResult::None;
+            }
+            QueryResult::Error(EvalError::cannot_index_with_type(
+                type_name(&value),
+                "number",
+            ))
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


newhoggy:
## Update (post code-review, two rounds)

Two independent review passes converged on the same real gaps, all fixed:

1. **`builtin_last`/`builtin_pick`/`builtin_omit`'s scalar fallback arms** didn't check `scalar_decode_failure` before `optional`, unlike the sibling `add`/`flatten`/`from_entries` fixes in the same original commit — an inconsistent application of the same rule within one diff.
2. **`join_with_separator`'s scalar fallback arm** still built its error message via unchecked `to_owned`.
3. **`builtin_pick`/`builtin_omit`'s *keys-expression* resolution** (as opposed to the picked/kept value, already fixed) still used unchecked `to_owned` — an undecodable string inside a dynamically-computed keys argument silently produced `{}` instead of raising. Live-verified both the bug (`pick(.k)` on a malformed key silently returns `Owned(Object({}))`) and the fix via a targeted revert.
4. **`builtin_first`**, found alongside `last` with the identical unfixed shape — not itself in this slice's named scope (its own Array arm never materializes), but the one-line fix was free to include.

Also closed the resulting coverage gaps: `pick`/`omit`'s `QueryResult::One`/`Many` keys-expression success and error paths, previously only exercised via a literal-array argument (`QueryResult::Owned`). `QueryResult::OneCursor` is documented (not chased) as structurally unreachable — it's only ever produced by `eval()`'s own `Expr::Identity` special case, never by `eval_single`, which is all `pick`/`omit`'s keys-expression evaluation calls.

Full test suite, clippy, and no_std build still pass; patch coverage 95.52% (277/290), with every remaining line accounted for (the known llvm-cov single-statement-branch quirk already documented in this repo's earlier #1755/#1670 rounds, plus the two documented-unreachable `OneCursor` arms).

PR #1807, opened 2026-08-29
fix(jq): to_owned_checked_at_depth raises on an undecodable object key

### LINKED ISSUE #1734: jq: eval.rs's own to_owned_at_depth still silently drops an undecodable key (#1642 follow-up, dead path today)

newhoggy:
## Summary

`to_owned_checked_at_depth` (`src/jq/eval.rs`) silently dropped an object field whose key failed to *decode* from the materialized map, instead of preserving it -- the same "still swallows a decode failure" bug class the `checked`/ variant exists to close.

The fix routes through `document.rs`'s shared `resolve_display_key`/`DisplayKeyGuard` mechanism -- `StandardJson` already implements `DocumentValue`, so the same mechanism `eval_generic.rs`'s three `to_owned*_at_depth` conversions and `lazy.rs`'s `cursor_to_owned_at_depth` already share is directly usable here, closing all four cases in one call:

- A key that fails to **decode** is **preserved** via its lossily-decoded raw source span (#1642's established convention, matching `length`/`keys_unsorted`/`.`) -- not raised on, not dropped.
- A key that is **structurally non-string** (#1194, JSON's grammar never allows it, e.g. `{123: 1}`) **raises** -- matching `eval_generic::to_owned_at_depth`'s own handling of the identical case.
- Two decode-failure keys whose lossy fallback spellings **collide** also **raise**, via `DisplayKeyGuard`, rather than one silently overwriting the other.
- A **trailing unpaired member** (#1194, e.g. `{"a":1,"b"}` -- a key with no value) **raises** -- this needed switching the iteration from `for field in *fields` to the `uncons`/`rest` walk `eval_generic::to_owned_at_depth` already uses, since a `for` loop consumes the `Iterator` impl to exhaustion and loses the signal `ends_unpaired` needs (genuine exhaustion and a trailing unpaired member both report `None` from `next()`).

All four live-verified via the library API, each matching the real CLI's own `eval_generic.rs`-routed behavior on the identical input.

## Revision history

Two review-driven corrections landed on top of the original submission:

1. **First pass**: the initial fix made a decode-failure key *raise* instead of preserving it -- inverting #1642's established convention (confirmed against `eval_generic.rs`'s own `test_to_owned_preserves_object_key_decode_failure_1642` and live CLI behavior). It also left the structural-key axis (#1194) unfixed. Both closed by routing through the shared `resolve_display_key` mechanism instead of hand-rolling either axis.
2. **Second pass**: a `for field in *fields` loop (kept from the original hand-rolled version) can never reach an `ends_unpaired` check, since the signal it needs is consumed by the time a `for` loop's body could ask for it -- not a missing line, the wrong iteration primitive. Fixed by switching to the `uncons`/`rest` walk shape, which also incidentally corrected `malformed_member_error()` being called on the un-advanced outer `fields` binding rather than the walk's current position (accidentally correct before this fix only because `JsonCursor::text()` returns the whole document regardless of cursor position -- not a documented invariant this code should have depended on).

Two further findings from the second pass are **pre-existing bugs in separate code**, unrelated to and predating this PR (reproduced identically against unmodified `main`), filed as follow-ups rather than fixed here:
- `eval_generic.rs`'s `Expr::Try` handling (`owned_or_err!` in the wildcard fallback arm of `eval_single`) has no catchability check at all, so `try`/`catch` doesn't contain a decode-failure/collision error the way `?` does -- this PR widens where that pre-existing gap is reachable (every one of `to_owned_checked`'s ~40 call sites can now raise the error class), but doesn't cause it.
- `colliding_display_key_error` is built via `EvalError::decode_failure`, whose doc comment says it must never be caught by `try`/`catch`/`?` -- but `is_decode_failure()`'s message-matching doesn't recognize its "is ambiguous" text, so it's actually an ordinary catchable error today, contrary to that documented contract.

Reachability is unaffected by any of the above: `jq_runner.rs` never calls `eval.rs`'s `eval()` at all, and `yq_runner.rs`'s only call site re-serializes an already-materialized `OwnedValue` (keys already valid) before rebuilding the cursor -- so this fix has no observable effect on either shipped CLI tool today, only on direct library API use, same as every other #1746/#1755 site already fixed this way.

## Test plan

- 4 regression tests in `src/jq/eval.rs`'s own test module: decode-failure key preserved, structural key raises, colliding fallback keys raise, trailing unpaired member raises.
- Full local `cargo test --features cli,simd,regex,serde`: 0 failures (confirmed via explicit `echo "exit: $?"` after two separate transient runs hit unrelated flaky `cargo run`-spawning integration tests under heavy concurrent load on this shared machine -- each confirmed passing individually and in a clean full-suite re-run).
- `cargo doc --no-deps --all-features` (including the new intra-doc link to `DocumentValue`), `cargo fmt --check`, `cargo build --no-default-features` (no_std): all clean.
- Patch coverage: 90.74% (49/54 new lines) -- all 5 uncovered lines are the 4 new tests' own unreachable-on-success `panic!` arms, the same pattern every other test in this cluster has.

Fixes #1734

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -620,13 +620,20 @@ fn to_owned_checked_at_depth<W: Clone + AsRef<[u64]>>(
         StandardJson::Object(fields) => {
             let mut map = IndexMap::new();
             for field in *fields {
-                // Get the key as a string
+                // A structurally non-string key (#1194) stays silently
+                // skipped here, matching #1642's established
+                // preserve-not-raise convention for that axis -- only a
+                // *string* key that fails to *decode* raises, mirroring
+                // the value arm just above.
                 if let StandardJson::String(key_str_val) = field.key() {
-                    if let Ok(cow) = key_str_val.as_str() {
-                        map.insert(
-                            cow.into_owned(),
-                            to_owned_checked_at_depth(&field.value(), depth + 1)?,
-                        );
+                    match key_str_val.as_str() {
+                        Ok(cow) => {
+                            map.insert(
+                                cow.into_owned(),
+                                to_owned_checked_at_depth(&field.value(), depth + 1)?,
+                            );
+                        }
+                        Err(e) => return Err(EvalError::decode_failure(e.message())),
                     }
                 }
             }
@@ -58689,6 +58696,34 @@ mod tests {
             other => {
                 panic!("expected the decode failure to bypass `optional` uncaught, got: {other:?}")
             }
+        }
+    }
+
+    /// #1734: `to_owned_checked_at_depth`'s `Object` arm guarded a field's
+    /// *value* against decode failure (the string-value arm just above)
+    /// but not its *key* -- an undecodable key was silently dropped from
+    /// the materialized map instead of raising, unlike every other
+    /// #1620/#1247 decode-failure site. Live-reachable via `sort` alone
+    /// (`builtin_sort` was already migrated to `to_owned_checked` by
+    /// #1755's own prior work): confirmed via the library API that
+    /// `[{<undecodable key>: 1}] | sort` used to return
+    /// `Owned(Array([Object({})]))` -- the whole field silently vanished,
+    /// no error -- rather than raising like the sibling string-value case
+    /// already did.
+    #[test]
+    fn eval_rs_to_owned_checked_object_key_decode_failure_raises_1734() {
+        let json: &[u8] = br#"[{"\ud800":1}]"#;
+        let index = JsonIndex::build(json);
+        let cursor = index.root(json);
+        let expr = parse("sort").unwrap();
+        match eval::<Vec<u64>, JqSemantics>(&expr, cursor) {
+            QueryResult::Error(e) => assert!(
+                e.is_decode_failure(),
+                "expected a decode-failure error, got: {e:?}"
+            ),
+            other => panic!(
+                "expected the undecodable object key to raise instead of being silently dropped, got: {other:?}"
+            ),
         }
     }
 
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -631,7 +631,17 @@ fn to_owned_checked_at_depth<W: Clone + AsRef<[u64]>>(
         StandardJson::Object(fields) => {
             let mut map = IndexMap::new();
             let mut guard = DisplayKeyGuard::default();
-            for field in *fields {
+            // `uncons`/`rest`, not `for field in *fields`: a `for` loop
+            // consumes the `Iterator` impl, whose `next()` reports `None`
+            // both on genuine exhaustion *and* on a trailing unpaired
+            // member (#1194, e.g. `{"a":1,"b"}`) -- the two are
+            // indistinguishable through that protocol, so a `for` loop
+            // has no way to ask `ends_unpaired` afterward (it needs the
+            // list state a walk *finished* on, not the one it started
+            // from). Mirrors `eval_generic::to_owned_at_depth`'s own walk
+            // shape for exactly this reason.
+            let mut f = *fields;
+            while let Some((field, rest)) = f.uncons() {
                 let key_val = field.key();
                 // `resolve_display_key` covers both key axes in one call
                 // (#1734): `Ok(None)` is a structurally non-string key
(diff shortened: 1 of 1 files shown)


newhoggy:
Code review found a critical inversion: the first version of this fix made a decode-failure key **raise** instead of preserving it, contradicting #1642's established convention (verified against `eval_generic.rs`'s own `test_to_owned_preserves_object_key_decode_failure_1642` and live CLI behavior — `succinctly jq -c 'sort'` on the same input preserves the key, no error). It also correctly flagged that the structural-key axis (#1194) was left untouched, still silently dropping where `eval_generic.rs`'s own `to_owned_at_depth` raises, and suggested reusing `document.rs`'s shared `resolve_display_key`/`DisplayKeyGuard` mechanism instead of hand-rolling.

Rewrote the fix to route through that shared mechanism (`StandardJson` already implements `DocumentValue`), which correctly closes all three cases in one call: decode-failure preserves, structural raises, and colliding fallback keys raise via `DisplayKeyGuard`. Replaced the single incorrect test with three regression tests, one per case, each live-verified via the library API before writing.

All local gates re-verified green: full test suite (0 failures), both clippy legs, `cargo doc` (including the new intra-doc link), `cargo fmt --check`, no_std build.

newhoggy:
Second review pass (6 finder agents) confirmed the corrected fix, and found one more real gap: a `for field in *fields` loop consumes `JsonFields`'s `Iterator` impl to exhaustion, which can't distinguish genuine exhaustion from a trailing *unpaired* member (#1194, e.g. `{"a":1,"b"}` — a key with no value) — both report `None`. Not a missing follow-up check, the wrong iteration primitive: switched to the `uncons`/`rest` walk `eval_generic::to_owned_at_depth`'s own object arm already uses for exactly this reason, with the same trailing `ends_unpaired` check. This also incidentally fixed `malformed_member_error()` being called on the un-advanced outer `fields` binding instead of the walk's current position.

Two more findings are confirmed pre-existing bugs in separate code (`eval_generic.rs`'s `Expr::Try` handling, and `colliding_display_key_error`'s catchability classification) — reproduced identically against unmodified `main`, so this PR widens where they're reachable but doesn't cause them. Filing follow-up issues for both rather than scope-creeping this PR.

Added a 4th regression test for the trailing-unpaired-member case. All local gates re-verified green (with explicit exit-code checks this time, to rule out a `grep -c` false-positive that briefly looked like a task failure): full test suite, both clippy legs, doc, fmt, no_std.

newhoggy:
Filed the two pre-existing follow-ups: #1812 (`try`/`catch` doesn't contain a decode-failure/collision error the way `?` does) and #1813 (`colliding_display_key_error` isn't actually uncatchable, contrary to `decode_failure`'s contract).

PR #1808, opened 2026-08-29
fix(jq): reassemble a --seq record whose bytes span a file boundary

### LINKED ISSUE #1571: jq: --seq -s silently drops a record whose bytes genuinely span a file boundary
**Severity: Low-Medium** — narrow (`--seq -s` with multiple files where a single record's bytes genuinely span a file boundary), but a *silent* data-loss bug: no error, no diagnostic, just a missing array element.

## Summary

Found during `/code-review` of PR #1568 (which fixed #1550's *error-location* marker for a truncated `--seq -s` record spanning a file boundary). That fix only touches `slurp_eof_line`'s location computation — the actual *value*-producing path, `parse_json_seq` (called once per file inside `get_inputs`'s main loop, `src/bin/succinctly/jq_runner.rs`), still parses each file in complete isolation. A record whose opening RS byte is in one file and whose closing bytes are in the next is never reassembled: each half is independently invalid JSON, so both halves are silently dropped per RFC 7464's "ignore malformed records" rule — even though the *whole* record, read as one continuous byte stream (which is what real jq's own `-s` reader does), is perfectly well-formed.

Two files: `v1` = `\x1e1\n\x1e{"a":"unterminated ` (opens a string, never closes it), `v2` = `str"}\n` (closes it — no RS byte of its own, since it's a continuation, not a new record):

```
$ jq --seq -c -s '.' v1 v2
[1,{"a":"unterminated str"}]

$ succinctly jq --seq -c -s '.' v1 v2
[1]
```

The second record is completely missing from succinctly's output, with exit code 0 and no stderr — indistinguishable from "the input only ever had one record."

## Why this is separate from #1550

#1550 (fixed by PR #1568) is scoped to the `(at ...)` error-location marker only, per its own issue text — it doesn't touch what values get parsed into the slurped array. This issue is the value-level counterpart: a genuinely larger, riskier change, since it requires `get_inputs`'s per-file loop to treat multiple `--seq` files as one continuous byte stream for *parsing* too, not just for the trailing-EOF check. #1568's own review surfaced this while confirming the location fix didn't regress; not attempted there since it's out of scope for a location-only fix.

## Suggested fix direction

Not investigated in depth. Likely needs `parse_json_seq` (or its caller) to work across the whole `raw_inputs` sequence rather than one file at a time — e.g. concatenating adjacent files' content when the current file's trailing bytes don't complete a valid record, mirroring the reassembly PR #1568 already does for the location-only check (`seq_stream_trailing_record_is_dropped`). A minimal version might only need to handle the "previous file's trailing segment is invalid on its own, but valid once the next file's leading segment is appended" case, without fully generalizing to records spanning 3+ files.

newhoggy:
## Summary

`get_inputs` calls `parse_json_seq` once per file, treating each file's
content in complete isolation. A record whose opening RS byte is in one
file and whose closing bytes are in the next is never reassembled: each
half is independently invalid JSON, so both are silently dropped per RFC
7464's "ignore malformed records" rule — even though the whole record,
read as one continuous byte stream (which is what real jq's own `-s`
reader actually does), is perfectly well-formed.

```console
$ jq --seq -c -s '.' v1 v2
[1,{"a":"unterminated str"}]

$ succinctly-before jq --seq -c -s '.' v1 v2
[1]
```

Found during `/code-review` of PR #1568 (#1550's fix for this same
boundary's *error-location* marker only — that fix doesn't touch what
values get parsed, per its own scope).

## Fix

Extracted `--seq` handling out of the main per-file loop into a new
`build_seq_values`, which concatenates every file's raw content in file
order into one string and runs the existing `parse_json_seq` **once**
over the combined result — matching real jq's actual model directly
(multiple `--seq` files are one continuous RFC 7464 byte stream to its
reader, regardless of whether `-s` is also passed) instead of
special-casing a narrower "two adjacent files" reassembly.

This also resolves a subtler pre-existing correctness question for free:
`parse_json_seq`'s own `seq_trailing_record_is_dropped` check (the
ambiguous-bare-number-at-true-EOF exclusion) previously ran once per file,
treating each file's own tail as if it were the true end of the whole
input — wrong whenever a later file exists. Run against the real combined
stream instead, it can only ever see genuine end-of-input.

Non-slurp mode's per-value `(source, line)` locations are recovered by
mapping each record's end offset in the combined string back to whichever
file's own byte range contains it, and computing the line via a fresh
per-file `LineCounter` — a boundary-spanning record is attributed to the
file its *end* falls in, matching #1568's own precedent for the stream's
trailing record.

The other three input modes (DSV, raw-input, plain JSON) are unaffected —
none has a multi-byte record delimiter that can be split across a file
boundary in the first place.

## Verification

9-shape differential sweep against pinned jq 1.7.1 (`/usr/bin/jq`), zero
regressions:

- The issue's own 2-file repro, slurp and non-slurp
- A record spanning 3 files (not just 2 — confirms the fix isn't narrowly
  hard-coded to one boundary)
- A bare number ambiguous only at its own file's end, disambiguated by the
  next file's content (the subtler correctness question above)
- Genuinely malformed content *not* at a boundary still drops (control —
  reassembly must not paper over real malformation)
- `input_line_number` for a boundary-spanning record
- Plain single-file `--seq` (regression check vs. old per-file behavior)

## Test plan

- [x] New unit tests in `src/bin/succinctly/jq_runner.rs`:
      `test_build_seq_values_reassembles_across_file_boundary_1571`,
      `test_build_seq_values_reassembles_across_three_files_1571`,
      `test_build_seq_values_still_drops_genuinely_malformed_record_1571`
- [x] New CLI integration test in `tests/jq_cli_tests.rs`:
      `test_jq_seq_value_reassembled_across_file_boundary_1571` (5 shapes,
      exact byte-for-byte jq 1.7.1 output)
- [x] Full test suite (`cargo test --features cli,simd,regex,serde`): all
      green, no regressions

Fixes #1571

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -1822,104 +1822,115 @@ fn get_inputs(
     // Process based on input mode
     let mut values = Vec::new();
 
-    for (file_idx, raw) in raw_inputs {
-        let src = file_idx.unwrap_or(0);
-
-        if let Some(delimiter) = args.input_dsv {
-            // DSV input: each row becomes a JSON array of strings
-            let parsed = parse_dsv_input(&raw, delimiter);
-            // One row per line (approximate for embedded newlines; jq has no
-            // DSV input mode, so there is no oracle to match here). Skipped
-            // under `--slurp` (#1541): the combined array's own location
-            // comes from `slurp_eof_line` above, not from any of these
-            // per-value entries, which `get_inputs` discards wholesale when
-            // it replaces `locations` with `InputLocations::single` below.
-            if !args.slurp {
-                for line in 1..=parsed.len() {
-                    locations.push(src, line);
-                }
-            }
-            values.extend(parsed);
-        } else if args.raw_input {
-            // Raw input: each line becomes a string. Deliberately ungated by
-            // `!args.slurp` (#1541), unlike its three siblings above/below --
-            // `args.raw_input && args.slurp` can only reach this arm when
-            // `args.input_dsv` is also set (otherwise the dedicated `-R -s`
-            // early return above already handled it), and the DSV check is
-            // tried first in this `if`/`else if` chain, so it always wins:
-            // this arm is unreachable whenever `args.slurp` is true.
-            for (line_idx, line) in raw.lines().enumerate() {
-                values.push(OwnedValue::String(line.to_string()));
-                locations.push(src, line_idx + 1);
-            }
-        } else if args.seq {
-            // JSON sequence input (RFC 7464): split on RS, ignore parse failures
-            let parsed = parse_json_seq(&raw);
-            // Skipped under `--slurp` (#1541) -- see the DSV branch above.
-            if !args.slurp {
-                locations.extend_from_ends(src, &raw, &json_seq_ends(raw.as_bytes()), parsed.len());
-            }
-            values.extend(parsed);
-        } else {
-            // JSON input: validate first if --validate is set
-            if args.validate {
-                let filename = file_idx.map(|idx| files[idx].to_string_lossy().to_string());
-                validate_json_input(raw.as_bytes(), filename.as_deref())?;
-            }
-            // Parse as JSON stream
-            let parsed = match parse_json_stream(&raw) {
-                Ok(p) => p,
-                Err(e) => return Ok(Err(e)),
-            };
-            // Skipped under `--slurp` (#1541) -- see the DSV branch above.
-            // `find_json_values` exists here only to feed
-            // `locations.extend_from_ends`, so slurp mode skips that scan
-            // entirely rather than running it and discarding the result.
-            // Side effect: the divergence check below (`find_json_values`
-            // disagreeing with `parse_json_stream`/`serde_json`) no longer
-            // runs under `--slurp` either -- if the two validators were ever
-            // to disagree on some input, non-slurp mode would still raise
-            // the internal error below, but slurp mode would now silently
-            // proceed. Accepted: the comment above already treats that
-            // divergence as unreachable through this crate's public CLI
-            // surface, so this only narrows *where* an already-believed-dead
-            // safety net runs, not what output a real input can produce.
-            if !args.slurp {
-                // `parse_json_stream` (above) already validated this exact
-                // input successfully via `serde_json`, which is strictly
-                // pickier than `find_json_values`'s own lenient heuristic
-                // scan (RFC 8259 plus #1094's leading-zero tolerance, vs.
-                // `find_json_values`'s RFC 8259 plus leading-zero *and*
-                // leading-dot tolerance, #1171) -- so `find_json_values`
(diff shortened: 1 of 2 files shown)


newhoggy:

**Fixed — real, critical regression, independently confirmed by 3 of 5 agents (each with a live repro)**: the original design scanned end offsets (\`json_seq_ends\`) and parsed values (\`parse_json_seq\`) as two *separate* passes over the combined stream, then reconciled them by comparing counts. Whenever any record anywhere in a multi-file \`--seq\` stream was genuinely malformed, that reconciliation's "counts disagree" fallback attributed **every value in the whole invocation** to the *last* file's last line — not just the dropped record's own. Confirmed live: \`error("boom")\` on a value physically living in file 1 of 2 was reported against file 2 (\`jq: error (at file2:1): boom\` instead of \`file1:1\`) — a wrong *filename*, not just an imprecise line.

**Root-cause fix, not a patch**: replaced the two-scan-then-reconcile design with \`parse_json_seq_with_ends\`, which computes each surviving value's own trimmed end offset in the *same* loop that decides whether to keep it. The two lists can no longer disagree in count — the whole class of reconciliation bug is now structurally impossible, not just fixed for this one repro. \`parse_json_seq\` (now a redundant wrapper with no production caller) and the now-fully-unused \`json_seq_ends\` were removed rather than left as dead code, updating 3 existing unit tests and 2 stale doc-comment cross-references along the way.

**Corrected a false claim in my own PR**: one agent live-verified that my original comment ("DSV and raw-input are line-oriented, no multi-byte record to split across a boundary") was wrong for raw-input — \`-R\` joins an unterminated trailing line across a file boundary the exact same way \`--seq\` joins a boundary-split record, confirmed against both the pinned oracle and unmodified \`main\`. Filed as **#1809** rather than fixed here (different mechanism, no \`build_seq_values\`-style machinery exists in that code path yet) — comment corrected to state this accurately instead of asserting safety I hadn't checked.

**Applied a simplification finding**: the file-offset lookup now uses \`file_ends.partition_point(...)\` instead of a hand-rolled three-mutable-variable \`while\` walk, removing two of the three pieces of loop-carried state.

**Test-coverage gap closed**: 3 agents independently noted the original tests only asserted value content and location *count*, never the actual \`(file, line)\` pairs — exactly the gap that let the regression through unnoticed. New tests assert real locations via \`InputLocations::get\` (unit-level) and a live \`error(...)\`-location CLI check, both verified against pinned jq 1.7.1 first.

**Not actioned**: the memory-duplication/extra-copy-pass concern (\`combined: String\` holds a second copy of every file's bytes) is a real, reasoned tradeoff for a narrow, non-hot-path CLI mode (RFC 7464 \`--seq\`, not the M2/streaming architecture's primary paths) — accepted rather than attempting a lazy multi-slice scanner under time pressure; happy to revisit if this ever shows up as a real bottleneck.

Full suite, both clippy legs, fmt, and no_std build all re-verified clean after these fixes.

Issue #355, opened 2026-07-26
jq/yq: uncaught evaluation error exits 0 instead of 5, and omits jq's position / (not a string) markers


newhoggy:
**Severity: High** — a failed query is indistinguishable from a successful one in a shell pipeline. Reproduced against the built CLI at `86e7eb16`.

An uncaught evaluation error prints a diagnostic to stderr and then **exits 0**. jq exits 5. Any script that relies on `set -e`, `&&`, or an explicit `$?` check treats a failed filter as success.

```
$ echo '{"x":1}' | sjq -c 'error("boom")'; echo "exit=$?"
jq: error: boom
exit=0                        # jq: exit=5

$ echo '{"x":1}' | sjq -c '1|.foo'; echo "exit=$?"
jq: error: expected object, got number
exit=0                        # jq: exit=5
```

The runner swallows the error into an empty result set rather than propagating a failure — `src/bin/succinctly/jq_runner.rs:1390` and `:1435`:

```rust
GenericResult::Error(e) => {
    eprintln!("jq: error: {e}");
    Ok(vec![])
}
```

`yq_runner.rs:348`/`:376` and `eval_generic.rs:341`/`:424` have the same shape, so `succinctly yq` exits 0 on a failed query too.

The diagnostic text also diverges. jq reports the input position and flags a non-string payload:

```
jq: error (at <stdin>:1): boom
jq: error (at <stdin>:1) (not a string): {"a":1}
```

against succinctly's `jq: error: boom` / `jq: error: {"a":1}`. The `(not a string)` marker became reachable with #158, which gave `EvalError` a payload — the raised value is now available at the print site, so distinguishing a string payload from a structured one no longer needs new plumbing.

**Fix:** propagate the error as a failure from the runner rather than `Ok(vec![])`, and exit 5 to match jq. Emit `(at <file>:<line>)` and the `(not a string)` marker from `EvalError`'s new `value` field while at the print site.

**Watch out:** several existing CLI tests assert exit 0 on paths that raise eval errors, and `-e/--exit-status` already has its own exit-code semantics (#178) that must not be conflated with this one — jq's 5 is for *errors*, `-e`'s 1 is for a false/null *result*.

**Scope note:** this was deliberately left out of #158, which fixed the caught path only. Filing it so the divergence is tracked rather than implied.

---
_Identified while fixing #158 and explicitly scoped out of it; present at `86e7eb16`._

PR #1811, opened 2026-08-29
fix(jq): raise on a #1194 structurally malformed key everywhere, not just via keys

### LINKED ISSUE #1679: jq/yq: a #1194 structurally malformed key silently vanishes via keys_unsorted/materialize but raises via keys
**Severity: Low-Medium** — reachable via the shipped `sjq`/`syq` CLIs (unlike #1678), but only
for the narrow #1194 shape (a key the format's grammar never allowed at all, e.g. a bare
non-string JSON key `{123: 1}`), not the far more common #1247/#1642 undecodable-key shape,
which this branch already fixed everywhere.

## Summary

Found while reviewing/fixing #1642, which closed the "same document, different answer"
inconsistency for a key whose bytes won't *decode* (#1247's fault). That fix left a
structurally-analogous, pre-existing gap open on the adjacent #1194 axis: a key the format's
grammar never allowed at all still raises on some routes and silently vanishes on others.

`keys()` (`src/jq/document.rs`), `effective_keys`, `to_owned_at_depth`,
`to_owned_cursor_at_depth`, `to_owned_with_comments_at_depth`, and the `ToEntries` builtin
(`src/jq/eval_generic.rs`) all do:

```rust
let Some(key) = key_display_string(&field.key) else {
    return Err(f.malformed_member_error());
};
```

`key_display_string` returns `None` exactly for a #1194 key (see its doc comment), so all of
these correctly raise.

But `lazy_keys_array_to_owned` and `cursor_to_owned_at_depth` (`src/jq/lazy.rs`), and
`stream_lazy_keys_json`/`stream_lazy_keys_yaml` (`src/jq/stream.rs`), instead do:

```rust
if let Some(key) = key_display_string(&field.key) {
    // push / write
}
```

with no `else` — a `None` (the #1194 case) just silently drops the field, no error. This
predates #1642 (the old code had the identical asymmetry via `field.key_str()`/`if let
StandardJson::String(k) = key`), so #1642 didn't introduce it — but its "one document, many
answers" framing makes this easy to mistake as already closed, since it's exactly that bug
class, just on the #1194 axis instead of #1247's.

## Repro

```console
$ echo '{123:1,"a":1}' | sjq -c 'keys'           # raises (correct)
$ echo '{123:1,"a":1}' | sjq -c 'keys_unsorted'  # ["a"]   <- silently one entry short, exit 0
$ echo '{123:1,"a":1}' | sjq -c 'length'         # 2
```

`keys` and `length`/`keys_unsorted` disagree with each other on the same document, depending
on which builtin is asked — the same failure mode #1642's own issue description named.

## Suggested direction

Not attempted; same two-way choice #1642 itself faced (raise everywhere vs. skip/preserve
everywhere), and picking "raise everywhere" here is more invasive than #1642's fix was:
`stream_lazy_keys_json`/`stream_lazy_keys_yaml` write through a `core::fmt::Write`-style
signature with no error channel today, the same constraint that made #1247's original
design doc (`docs/plan/decode-failure-routing.md`) split out and defer the *value*-side
streaming fix as its own stage rather than attempt it inline. `lazy_keys_array_to_owned`'s
signature already returns `Result`, so that one site at least is a smaller lift.

## Related

- #1642 — the review that found this; fixed the #1247/#1642 axis of the same class of bug,
  left this #1194 axis open.
- #1194 — the original structural-key issue this asymmetry belongs to.
- #1247 — established the `EvalError`-raising convention the fixed call sites already use.

newhoggy:
Fixes #1679.

`key_display_string` correctly returns `None` for a key the format's grammar
never allowed at all (a #1194-shaped key, e.g. a bare numeric JSON key), and
most call sites already raise `malformed_member_error()` on that `None` — but
four did not, silently dropping the field instead:

- `lazy_keys_array_to_owned` / `cursor_to_owned_at_depth` (`src/jq/lazy.rs`)
- `stream_lazy_keys_json` / `stream_lazy_keys_yaml` (`src/jq/stream.rs`)

```console
$ echo '{123:1,"a":1}' | sjq -c 'keys'           # raises (correct)
$ echo '{123:1,"a":1}' | sjq -c 'keys_unsorted'  # ["a"]   <- one entry short, exit 0
$ echo '{123:1,"a":1}' | sjq -c 'length'         # 2
```

`keys`/`keys_unsorted`/`length` disagreed with each other on the same
document — the same "one document, many answers" failure mode #1642 fixed on
the adjacent decode-failure axis, left open here on the #1194 (structural)
axis.

## Fix

- `lazy_keys_array_to_owned` and `cursor_to_owned_at_depth` now mirror
  `effective_keys`/`eval_generic::to_owned_at_depth`'s already-correct shape
  exactly: raise mid-loop on a non-stringifiable key, and check
  `DistinctKeyCursors::ended_unpaired()` after the loop for the other #1194
  shape (a trailing key with no paired value — the walk never even yields
  it, so a mid-loop check alone can't see it).
- The two `stream.rs` writers only have a `core::fmt::Result` channel
  (`core::fmt::Error` carries no message), so changing their return type
  would ripple into their callers' own `Result<StreamStats, fmt::Error>`
  signatures. Instead they take an `&mut Option<EvalError>` out-parameter and
  stop writing further keys on a #1194 key, mirroring the `Partial`/
  `owned_or_stream_error` idiom their sibling `sorted: true` branch already
  uses: whatever was already written to `out` stays written, and the failure
  travels back out-of-band via `StreamStats::error` instead of as a value on
  `out` (#355).

## Testing

- New unit tests directly exercising all 4 fixed functions, both #1194
  shapes (non-string key, unpaired trailing child) each.
- A streaming-path test (`test_stream_unsorted_lazy_keys_arm_raises_on_malformed_key_1679`)
  exercising the actual `GenericResult::stream_json`/`stream_yaml` call sites,
  the sibling of the pre-existing `sorted: true` test for the same shape.
- Verified live: reverting `lazy_keys_array_to_owned`'s fix reproduces the
  exact silent-drop bug (`Array(["b"])`/`Array([])` instead of raising) that
  the new test catches.
- Confirmed against the pinned `/usr/bin/jq` oracle: real jq rejects this
  input outright at parse time for every filter (`keys`/`keys_unsorted`/
  `length` all exit 5, same message) — this fix makes succinctly's own
  evaluation-time check consistent across all four call sites, matching the
  outcome even though succinctly's semi-index defers the check to
  evaluation (already-documented architectural difference, ADR/semi-indexing
  docs).
- Full local suite green (2977+ lib tests + all integration suites, 0
  failures). Patch coverage 97.46% (230/236); the 6 remaining lines are
  confirmed non-issues: one is the well-documented multi-line-call
  llvm-cov attribution artifact (raw lcov shows the call executed 15 times;
  only its closing-paren line reads 0), the other 5 are trivial `panic!`
  fallback arms inside test-only `match`/`let-else` statements.

### DIFF
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -14234,6 +14234,40 @@ mod tests {
         assert_eq!(out, "");
     }
 
+    /// #1679: the `sorted: false` sibling of the test above -- the genuinely
+    /// lazy path (`stream_lazy_keys_json`/`stream_lazy_keys_yaml`, #685)
+    /// used to silently skip a #1194 key instead of raising, disagreeing
+    /// with `keys`'s `sorted: true` arm on the very same document. Both
+    /// output formats must now report the failure via `stats.error` too,
+    /// keeping whatever prefix (nothing, here, since the malformed key is
+    /// the only one) was already written.
+    #[test]
+    fn test_stream_unsorted_lazy_keys_arm_raises_on_malformed_key_1679() {
+        let json: &[u8] = b"{123: 1}";
+
+        let index = JsonIndex::build(json);
+        let cursor = index.root(json);
+        let result = eval_with_cursor(&crate::jq::parse("keys_unsorted").unwrap(), cursor);
+        assert!(matches!(
+            result,
+            GenericResult::LazyKeys { sorted: false, .. }
+        ));
+
+        let mut out = String::new();
+        let stats = result
+            .stream_json(&mut out, IndentSpec::COMPACT, false, |_| Ok(()))
+            .unwrap();
+        assert!(stats.error.is_some());
+        assert_eq!(out, "[]");
+
+        let mut out = String::new();
+        let stats = result
+            .stream_yaml(&mut out, IndentSpec::spaces(2), false, |_| Ok(()))
+            .unwrap();
+        assert!(stats.error.is_some());
+        assert_eq!(out, "");
+    }
+
     /// `GenericResult::materialize_lazy`'s `LazyKeys` arm `Err` side:
     /// reached whenever a `LazyKeys` result is materialized by a consumer
     /// other than `stream_json`/`stream_yaml`/`fold_lazy_keys_stage` (which
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -2324,7 +2324,24 @@ impl<V: DocumentValue> GenericResult<V> {
                     // `Builtin::KeysUnsorted`. `sort_keys` (`-S`) is a no-op
                     // here: this is a flat array of key names, not a
                     // mapping, so there's nothing for `-S` to reorder.
-                    crate::jq::stream::stream_lazy_keys_json(fields, *collapse, out, indent)?;
+                    //
+                    // #1679: a #1194 key stops the writer mid-stream rather
+                    // than silently dropping it; whatever keys already
+                    // reached `out` stay written (same `Partial` idiom as
+                    // `owned_or_stream_error` above), and the failure
+                    // travels back via `stats.error` instead.
+                    let mut malformed_key = None;
+                    crate::jq::stream::stream_lazy_keys_json(
+                        fields,
+                        *collapse,
+                        out,
+                        indent,
+                        &mut malformed_key,
+                    )?;
+                    if let Some(e) = malformed_key {
+                        stats.error = Some(stream_error(&e));
+                        return Ok(stats);
+                    }
                 }
                 on_value(out)?;
                 stats.count = 1;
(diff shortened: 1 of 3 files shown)


newhoggy:
Code review found a 5th call site with the same bug shape: \`write_json_at_depth\`'s \`LazyKeysArray\` arm (\`src/jq/lazy.rs\`) — this PR's own "everywhere" claim didn't fully hold. It wrote a non-string key's raw source bytes verbatim, assuming they're always an already-quoted JSON string token; for a #1194 key that's false, so \`{123:1}\` produced the invalid JSON \`[123]\` instead of raising, and \`{invalid}\` silently closed the array early. Fixed in \`0eb9107ec\`, mirroring the existing \`JqValue::Cursor\` arm's precedent one match arm up (\`cursor_to_owned(c).map_err(|_| core::fmt::Error)?\`) since \`write_json\` has no error channel richer than \`core::fmt::Error\`. Verified live via revert (reproduces \`[123,"b"]\`).

Patch coverage note: the defensive \`None\` fallback inside that same arm (raw_bytes() returning \`None\` for a string-token key) is genuinely untestable — the JSON semi-index always gives a string token a text range, so this fallback predates this PR and was already unreachable in practice (its own comment says so). Left uncovered per that existing convention. The 4 other uncovered lines llvm-cov reports are the closing-brace-after-`return`/trivial-match-arm attribution artifacts already documented elsewhere in this repo's coverage notes — confirmed via raw lcov DA: hit counts, not just assumed.

Full local verification (build/clippy/no_std/full test suite) re-run clean after this fix; patch coverage 95.65% overall with the above gaps accounted for.

PR #1815, opened 2026-08-29
fix(jq): isolate the nesting-depth panic per-document instead of aborting the stream

### LINKED ISSUE #1793: jq: any non-lazy-path builtin panics on a document with 257+ levels of nesting anywhere, even unreached
**Severity: Medium-High** (a hard process panic/crash, reachable from ordinary CLI usage on adversarial or accidentally-malformed input — not a graceful error)

## Summary

Found while investigating a hoist-vs-decode-eagerness concern raised in #1535's own code review (unrelated — see that PR's own follow-up note). A document containing a structure nested past ~256-257 levels **anywhere** crashes the process for *any* query that isn't eligible for the "lazy path" — even a query that never actually reads the deeply-nested part.

## Repro (confirmed on current `main`, this session)

\`\`\`bash
\$ python3 -c "print('[[1,2],' + '['*400 + '1' + ']'*400 + ']')" > deep.json

\$ succinctly jq -c 'length' deep.json
2

\$ succinctly jq -c '.[1]' deep.json
[[[[[[[...(prints fine, full depth)...]]]]]]]

\$ succinctly jq -c 'sort' deep.json
thread 'main' panicked at src/jq/eval_generic.rs:262:5:

\$ succinctly jq -c 'map(.)' deep.json
thread 'main' panicked at src/jq/eval_generic.rs:262:5:

\$ succinctly jq -c 'join(",")' deep.json
thread 'main' panicked at src/jq/eval_generic.rs:262:5:
\`\`\`

Note the document's *own top level* is trivial — a 2-element array, `[[1,2], <the deep part>]` — and \`sort\`/\`map(.)\`/\`join(",")\` don't even conceptually need to look inside the deep second element to do their job (\`sort\` on a 2-element array just needs a comparison; \`map(.)\`/\`join\` do walk every element, but the panic fires from \`to_owned_cursor_at_depth\` — \`src/jq/eval_generic.rs:262\`, \`assert_nesting_depth\`, apparently a 256-level limit distinct from \`MAX_VALUE_TREE_DEPTH\`'s own 384 in \`src/jq/value.rs\` — well before any output is produced).

## Why this looks like a materializing-path issue, not a per-builtin one

\`length\` and \`.[1]\` (structural navigation / streaming output) succeed on the identical document; only builtins that aren't eligible for whatever this codebase calls the "lazy path" (per prior session memory: \`can_use_lazy_path\` in \`jq_runner.rs\`, or an equivalent gate) hit this. That suggests some **whole-document eager materialization step reachable from the non-lazy-path builtin dispatch** — not \`sort\`/\`map\`/\`join\` individually — is what calls \`to_owned_cursor\`/\`to_owned\` deeply enough to hit the 256 guard, regardless of whether the specific query needs the deep part at all.

## Scope

Not investigated further here — this needs someone to trace exactly which non-lazy-path entry point eagerly walks the *whole* input to that depth (rather than the specific value a query reads), and either raise a graceful \`EvalError\` there instead of panicking (matching \`MAX_VALUE_TREE_DEPTH\`'s own stated intent of being a *safety net*, not a crash), or make it not require full-document depth at all when the query doesn't need it. A crafted deeply-nested JSON/YAML document reaching this via ordinary CLI usage (no special flags needed) is a real DoS-class finding, not theoretical — see the "Issue #1098 panic fix reverted" precedent in this repo's own history for the same failure class.

newhoggy:
Fixes #1793

## Summary

`sort`/`join`/`map(.)` have no native lazy fast path (or, for `map(.)`, still need to materialize their `LazySeq` result before printing) and fall back to `to_owned_cursor_at_depth`'s `MAX_NESTING_DEPTH` guard — a deliberate `panic!`, not an `EvalError` (see that guard's own doc comment and #1021's history: routing this through `EvalError` was tried and reverted because it makes a stack-safety guard catchable by `try`/`catch`).

For a bare `sort`/`map(.)`/`join(",")` on `.`, an adversarially or accidentally deep document used to escape as an **uncaught panic** — not just an ugly message, but a real process abort that silently discarded every subsequent document in a multi-value stream too, unlike the `#1194`/`#355` "isolate this one document, continue the stream" convention this same CLI loop already follows for every other uncaught-error class.

## Fix

Wrapped the CLI's per-document `evaluate_bytes_lazy` call (`jq_runner.rs`) in `catch_unwind`, matching this specific panic's message and routing it through the same `sink.report()` + `continue` path `#1194` already uses. Anything that doesn't match this exact guard's message is re-raised via `resume_unwind` — an unexpected panic still crashes loudly, it's not silently absorbed as if it were this one known condition.

This only touches the CLI's outermost per-document call site, not `to_owned_cursor_at_depth`'s 58+ interior call sites or `EvalError`/`Control`'s own shape, so it doesn't reopen #1021's stack-safety regression. `-e`'s own separate panic path (`lazy.rs`'s `cursor_to_owned`) is untouched — confirmed live, `test_exit_status_query_rejects_adversarial_nesting_998` still panics at exit 101 exactly as before.

## Before / after

```
$ succinctly jq -c 'sort' huge-stream.json
thread 'main' panicked at src/jq/eval_generic.rs:267:5: nesting depth exceeds limit of 256
# process aborted, every document after the offending one silently lost
```
```
$ succinctly jq -c 'sort' huge-stream.json
jq: error (at huge-stream.json:1042): nesting depth exceeds limit of 256
# exit 5, and every other document in the stream still gets processed
```

## A footgun found along the way

`payload: Box<dyn Any + Send>` from `catch_unwind`'s `Err`, and a bare `&payload` coerces to `&(dyn Any + Send)` by treating the **Box itself** as the concrete type under test (`Box<T>` is `Any` too, via a blanket impl) rather than looking at what it holds — so `downcast_ref` inside a helper taking `&(dyn Any + Send)` silently misses regardless of the real payload type unless the Box is dereferenced first (`&*payload`). Cost me a debugging round; documented at the call site.

## Testing

- 6 new regression tests: `sort`/`map(.)`/`join(",")` each get a clean exit-5 diagnostic instead of panicking; a companion "still succeeds under the limit" test; the headline test (`test_adversarial_nesting_does_not_abort_rest_of_stream_1793`) proving sibling documents on either side of the adversarial one are still processed; a control case confirming an ordinary (non-panic) type error is unaffected.
- All 4 pre-existing `#998`/`-e` nesting tests still pass unchanged.
- YAML confirmed unaffected — its own parser enforces a lower (128), already-graceful structural limit at parse time, so it never reaches this guard.
- Full `cargo test --features cli,regex` (55/55 green), both CI clippy legs, `cargo fmt --check`, `cargo build --no-default-features`, `cargo doc --all-features`.
- Interleaved release-build A/B (this branch vs. unmodified `main`, 500K-small-document identity stream): output byte-identical, timing within noise across 4 reps — `catch_unwind`'s cost is near-zero on the non-panicking path.

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -956,18 +956,27 @@ equivalent guard only fires later, when a builtin with no native lazy fast path 
 `join`) or one that still has to materialize its result before printing (`map(.)`) forces
 a full `to_owned_cursor` conversion of the value it's handed — `nesting depth exceeds
 limit of 256` (also exit 5, matching jq's own exit code by coincidence of both picking the
-same conventional "filter failed" code, not by design). succinctly's own wording, since
-there is no equivalent jq sentence to copy for an evaluation-time guard jq has no
-counterpart to (its own check never gets this far). See #1793 for the fix that turned this
-from an uncaught process panic into a clean, catchable diagnostic.
+same conventional "filter failed" code, not by design — an internal architectural ceiling
+being reported through the same channel as an ordinary filter/type error, deliberately,
+so a `sort`/`join`/`map(.)` result is never distinguishable-by-exit-code from any other
+uncaught `EvalError` on this path). succinctly's own wording, since there is no equivalent
+jq sentence to copy for an evaluation-time guard jq has no counterpart to (its own check
+never gets this far). Falls under ADR-0018's "would take the host process down" exception
+— the pre-fix behavior was an uncaught panic, not merely different wording. See #1793 for
+the fix that turned this from an uncaught process panic into a clean, catchable
+diagnostic.
 
 This fix is scoped to the CLI's default (lazy) per-document dispatch, matching #1793's own
-repro — the same panic remains uncaught via a CLI flag that forces whole-batch
-materialization up front (`--slurp`, `-S`/`--sort-keys`, `-C`/`--color-output`,
-`--ascii-output`, `--slurpfile`) or via `-e`/`--exit-status`'s own separate materializer
-(`src/jq/lazy.rs`, already pinned uncaught by `test_exit_status_query_rejects_adversarial_
-nesting_998`), and `succinctly yq` has no equivalent guard on either its default or
-materializing path at all. Tracked separately, not fixed here.
+repro — the identical panic remains uncaught via any of: a CLI flag that forces
+whole-batch materialization up front (`--slurp`, `-S`/`--sort-keys`, `-C`/`--color-output`,
+`--ascii-output`, `--slurpfile`); a filter using `input`/`inputs`/`input_line_number`,
+which routes to that same materializing branch with *no flag at all* (#1818); `-e`/
+`--exit-status`'s own separate materializer (`src/jq/lazy.rs`, already pinned uncaught by
+`test_exit_status_query_rejects_adversarial_nesting_998`); and `succinctly yq`, which has
+no equivalent guard on either its default or materializing path at all (#1817). Tracked
+separately, not fixed here. Confirmed live, also pre-existing and unrelated to this fix:
+`print_json`'s own guard can flush corrupted/truncated JSON to stdout before it fires
+(#1819).
 
 ## Regex flags `l` and `n`
 
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -943,6 +943,32 @@ against the pinned jq 1.7.1 binary), so refusing with a catchable error is the s
 "would take the host process down" exception as the rest of this entry, not a new kind of
 divergence.
 
+## A too-deeply-nested document is caught at a different stage, with different wording
+
+Real jq refuses a document nested past 256 levels at *parse* time: `jq -c sort` on a
+260-level-deep array gives `jq: parse error: Exceeds depth limit for parsing at line 1,
+column 257` (exit 5) before evaluation ever starts, regardless of which filter is applied
+— confirmed live against the pinned 1.7.1 binary. Succinctly's semi-indexing accepts the
+same document at parse/index time (its own architectural point, #1793's own investigation
+confirmed: `succinctly jq -c length`/`.[1]` on the identical document already succeed,
+since navigating to or measuring a value doesn't require materializing it), so the
+equivalent guard only fires later, when a builtin with no native lazy fast path (`sort`,
+`join`) or one that still has to materialize its result before printing (`map(.)`) forces
+a full `to_owned_cursor` conversion of the value it's handed — `nesting depth exceeds
+limit of 256` (also exit 5, matching jq's own exit code by coincidence of both picking the
+same conventional "filter failed" code, not by design). succinctly's own wording, since
+there is no equivalent jq sentence to copy for an evaluation-time guard jq has no
+counterpart to (its own check never gets this far). See #1793 for the fix that turned this
+from an uncaught process panic into a clean, catchable diagnostic.
+
+This fix is scoped to the CLI's default (lazy) per-document dispatch, matching #1793's own
+repro — the same panic remains uncaught via a CLI flag that forces whole-batch
+materialization up front (`--slurp`, `-S`/`--sort-keys`, `-C`/`--color-output`,
+`--ascii-output`, `--slurpfile`) or via `-e`/`--exit-status`'s own separate materializer
+(`src/jq/lazy.rs`, already pinned uncaught by `test_exit_status_query_rejects_adversarial_
+nesting_998`), and `succinctly yq` has no equivalent guard on either its default or
+materializing path at all. Tracked separately, not fixed here.
+
 ## Regex flags `l` and `n`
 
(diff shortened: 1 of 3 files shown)


newhoggy:
Round 2, addressing the multi-angle review of the initial fix (10+ finder-agent reports across two full review runs):

**Fixed — two real correctness bugs in the new code:**
- `nesting_depth_panic_message` matched on a substring, which also matches `assert_value_tree_depth`'s independent `MAX_VALUE_TREE_DEPTH` (384) guard — a different failure class sharing the exact message template via the same underlying `assert_depth` call. Confirmed live: `reduce range(400) as $i (null; [.])` was being silently caught and reported as if it were #1793's own condition. Narrowed to an exact match against `MAX_NESTING_DEPTH`'s own interpolated value; added a regression test pinning that the unrelated guard still crashes uncaught.
- The panic-catching arm's `continue` skipped the `sink.halted()` check every other early-exit in this same loop reaches on the same iteration — added it, matching every sibling arm's convention (`halt`/`halt_error` outranks everything else, #791).

**Attempted, then reverted:** a custom panic hook to suppress the raw `thread 'main' panicked at ...` banner Rust's default hook still prints before `catch_unwind` gets control. `PanicHookInfo` carries no caller-site identity, only the message — and the exact-match fix above meant the hook also silently suppressed `-e`'s own separate, still-uncaught panic's *only* diagnostic output (it shares the identical message from a different call site). Confirmed live: this broke the pre-existing `test_exit_status_query_rejects_adversarial_nesting_998` (empty stderr). Reverted — the raw banner alongside the clean diagnostic is a real but minor, now-documented rough edge, not worth a message-only global hook's fragility.

**Added:** a `docs/compliance/jq/limitations.md` entry per CLAUDE.md's own rule (real jq rejects the same shape at *parse* time with different wording — confirmed live against the pinned 1.7.1 binary), explicitly scoping this fix to the CLI's default lazy per-document dispatch.

**Filed as follow-ups** (review confirmed the identical guard is reachable, uncaught, via three further paths — each a genuinely different *trigger* sharing only the underlying guard, not #1793's own reported "builtin lacks a lazy fast path" shape):
- #1817 — `succinctly yq` has zero equivalent protection on its own dispatch loop.
- #1818 — `--slurp`/`-S`/`-C`/`--ascii-output`/`--slurpfile` still panic uncaught, and worse than the original bug (loses the whole batch's output, not just documents after the offending one, since that branch reads everything up front).
- #1819 — pre-existing, unrelated to this fix: a value strictly between `MAX_NESTING_DEPTH` (256) and `MAX_VALUE_TREE_DEPTH` (384) prints truncated/corrupted JSON before `print_json`'s own guard fires.

Re-verified: full `cargo test --features cli,regex` (55/55 green — two transient "child was killed by signal 9" failures on unrelated tests during a high-load run, confirmed as machine contention noise by passing in isolation and on a clean re-run), both CI clippy legs, `cargo fmt --check`, `cargo build --no-default-features`, `cargo doc --all-features`, all 11 of this issue's regression tests, and the live sanity sweeps from round 1 (multi-doc isolation, `-e` unaffected).

newhoggy:
Round 3, addressing the third review pass:

**Corrected an incomplete scope note**, confirmed live: a filter using `input`/`inputs`/`input_line_number` reaches the exact same unguarded materializing branch as the CLI flags already documented — with **no flag needed at all** (`succinctly jq -c '[inputs]' deep.json` still panics uncaught). This isn't a new bug, just an omission in what I'd already filed as #1818 — retitled that issue and commented with the confirmed repro, and expanded the `limitations.md` entry to name it explicitly.

**Added the missing ADR-0018 exception citation** the entry needed per CLAUDE.md's own rule (every neighboring entry in that file has one) — this falls under "would take the host process down," since the pre-fix behavior was an uncaught panic, not merely different wording from real jq's own parse-time rejection.

**Added an inline comment** explaining a previously-undocumented but intentional choice review flagged: this now reports through jq's ordinary exit-5 diagnostic channel rather than keeping the panic's own exit 101 — deliberate, since an internal architectural ceiling is meant to be indistinguishable, by exit code, from any other uncaught `EvalError` on this path.

**Noted on #1819** that its corrupted-output finding reproduces on an even simpler shape than its original repro: a plain, already-deep *input document* via ordinary navigation (`.[0]`), no `reduce`-based value construction needed — same underlying `print_json` mechanism, confirmed pre-existing on `main`, unaffected by this fix.

**Not re-attempted:** the panic-hook-noise finding (raised again this round) — round 2's attempt at a global message-matching hook demonstrably broke `-e`'s own separate diagnostic (confirmed live, and explained in that round's commit). Nothing in this round's findings offers a different mechanism, so I'm treating this as a documented, accepted rough edge rather than re-attempting the same broken approach.

This is documentation- and comment-only this round — no behavior change. Re-verified: full `cargo test --features cli,regex` (55/55 green), both CI clippy legs, `cargo fmt --check`, `cargo build --no-default-features`, `cargo doc --all-features`.

Three review rounds have now converged on the same handful of already-triaged, already-filed gaps rather than surfacing new correctness bugs in this diff's own code. I believe this is ready to merge.

Issue #1820, opened 2026-08-29
jq: 20+ more to_owned(&value) sites in eval.rs share #1755's decode-failure shape


newhoggy:
**Severity: Low-Medium**

## Summary

While implementing the #1755 "misc bucket" slice (PR to follow), a full audit of every remaining `to_owned(&value)`/`to_owned(&e)`/`to_owned(&elem)` call in `src/jq/eval.rs` turned up more genuine instances of the same bug shape (silent `""` substitution instead of raising on an undecodable string) than the 13 functions that slice's own scope named. Two categories:

### 1. Genuine bugs found via the same textual pattern (`to_owned(&value)`), not fixed by the misc-bucket slice

| Line (pre-misc-bucket-PR) | Function | Note |
|---|---|---|
| 1009 | `eval_single`, `Expr::Iterate` arm | `.[]` on an undecodable-string scalar misreports as an ordinary type error instead of `decode_failure` |
| 5279 | `eval_not` | `!"<bad-string>"` silently becomes `!true` (empty string is truthy) instead of raising |
| 5443 | `eval_error` (bare `error`, no message) | the raised *payload* itself is corrupted to `""` |
| 6003 | `builtin_keys`/`keys_unsorted` | scalar fallback, no `scalar_decode_failure` guard |
| 6425 | `builtin_map` | scalar fallback, no guard (sibling `map_over`'s per-element path is already fixed, but this outer type-error arm isn't) |
| 6543 | `builtin_map_values` | same gap as `builtin_map` |
| 8550 | `builtin_to_entries`, scalar fallback | the Array/Object arms are already fixed (this is `to_owned_checked`'s own "primary" call site), but the residual scalar arm was missed |
| 11402 | `builtin_tojsonstream` | unconditional whole-value materialization feeding *real output*, not just a message |
| 11444 | `builtin_fromjsonstream`, Array arm | passes possibly-corrupted array straight through as output |
| 27377 | `builtin_truncate_stream` | whole-input `to_owned` used as the truncation depth |
| 27429 | `builtin_paths` | whole-input `to_owned` feeding the actual `paths` output |
| 27508 | `each_paths_filter` (lazy `paths(filter)` producer) | same, feeds actual path output |
| 27659 | `builtin_leaf_paths` | same shape as `builtin_paths` |
| 30019 | `builtin_mktime` | converts *before* checking type -- an undecodable top-level string silently becomes `""` -> misreported as "not an array" |
[...]

Issue #400, opened 2026-07-27
jq: `QueryResult` models `Error`/`Break` per-stream, so a mid-stream error discards outputs already computed


newhoggy:
**Severity: Medium** — reproduced against the built CLI (release, branch `issue-160-jq-alternative-only-inspects-the-first-output`).

`QueryResult::Error` and `QueryResult::Break` are properties of the *whole* stream rather than of one output, so any operator that consumes a stream loses the outputs that preceded the error or break.

```
echo null | sjq -c '(1,error("x")) // 2'                             # => 2         (jq: 1, then the error)
echo null | sjq -c 'label $out | ((true,true) and (1, break $out))'  # => (nothing)  (jq: true)
```

`eval_comma` has the same limit, which is where the collapse originates: `(1,error("x"))` becomes a whole-stream `Error`, discarding the `1` before `//` ever sees it.

**Distinct from #377.** #377 is about `//` discarding a left-hand error *outright* (`error("x") // 3` yields `3`); this issue is about losing the outputs that came *before* the error, and it affects `and`/`or` as well as `//`.

**Why it was deferred from #160.** #160 fixed the multi-output collapse in `//`, `and` and `or`. This one is not a per-operator patch — it needs `QueryResult` to model control flow per-output rather than per-stream, so that a stream can carry `1` *and then* an error. That is a representation change touching every operator that consumes a stream.

**Fix sketch:** make the output sequence able to terminate with an error/break after yielding values — e.g. `Many(Vec<T>, Option<Control>)`, or an iterator whose item type is `Result`-shaped — then propagate the already-computed prefix at each consumer instead of dropping it.

---
_Split out from #160, which documented this as a known limitation in its CHANGELOG entry._

PR #1821, opened 2026-08-29
fix(jq): plug decode-failure corruption in the misc-bucket family

newhoggy:
Part of #1755. Follow-up scope filed as #1820.

Fixes the last 13 named functions from #1755's own tracking scope, converting
their `to_owned`/`to_owned_at_depth` entry-point conversions to the fallible
`to_owned_checked`, with the error propagated as `QueryResult::Error` before
any generator loop starts:

- `builtin_upper_index` / `builtin_upper_index_stream` (`INDEX`)
- `getpath_one_path`
- `eval_single`'s yq-mode object slice arm
- `any_all_over` / `any_all_f` / `any_all_gen_cond` (`any`/`all`, bare and
  `(cond)`/`(gen;cond)` forms)
- `eval_pipe`'s path-context entry
- `builtin_load`'s JSON branch
- `builtin_combinations` / `builtin_combinations_n`
- `builtin_debug` / `builtin_debug_msg` / `builtin_stderr` /
  `builtin_halt_error` (the last found incidentally, same file region, same
  shape)
- `builtin_transpose`
- `builtin_shuffle`
- `builtin_pivot`
- `builtin_with_entries`
- `eval_rhs_once` (the "masked hygiene fix" — currently unobservable through
  either real caller, since both immediately re-materialize the whole
  document via `eval_update`'s own already-fixed `to_owned_checked` right
  after; fixed anyway so correctness doesn't rest on that coincidence)

None of these route through `finish_fork`, so there's no `optional`-suppression
interaction to work around (unlike the recurse/walk/tostream slice, #1806).

`builtin_with_entries`'s fix site is structurally unreachable via any query
today: `entry_json` is built via `owned_to_json_bytes`, which serializes into
a Rust `String` (always valid UTF-8 by the type system) before the cursor `f`
evaluates against is ever built — documented at the call site rather than
forced into a test.

## Broader audit finding

While implementing this slice, auditing every remaining `to_owned(&value)`
call in `eval.rs` found ~20 more genuine same-shape sites beyond this slice's
own named scope (e.g. `builtin_paths`/`builtin_leaf_paths`/
`builtin_truncate_stream`, structurally identical to the recurse/walk/tostream
family), plus a ~47-site unaudited pool in generic `QueryResult`-collector
plumbing. Filed as #1820 rather than folded in here, to keep this PR
reviewable at a similar size to the five prior #1755 slices.

## Testing

- `test_misc_bucket_raises_on_decode_failure_1755` covers all 12 reachable
  sites (19 expressions, plus a manual yq-mode check for the object slice
  arm).
- `test_misc_bucket_valid_data_unaffected_1755` is the positive control.
- `builtin_load` gets a dedicated test (raw-byte file I/O, since the
  existing `with_temp_file` helper's `&str` signature can't hold invalid
  UTF-8).
- `eval_rhs_once` gets a dedicated test calling the private function
  directly (both its `One` and `Many` result arms) — a `+=`-based
  integration test would only prove `eval_update`'s pre-existing fix, not
  this one.
- Verified live: reverting `any_all_over`'s fix reproduces the exact
  silent-corruption bug (`any` on `["\xff\xfe"]` returns `true` instead of
  raising).
- Full local suite green (2988 lib tests + all integration suites). One
  CLI-suite flake (`test_preserve_input_duplicate_keys_are_output_only_1385`,
  unrelated to this `eval.rs`-only diff) confirmed transient via isolated
  rerun and a clean full-suite rerun.
- Patch coverage 92.93% (184/198). Remaining gaps: 6 confirmed llvm-cov
  attribution artifacts (multi-line match-arm/macro-invocation lines
  showing 0 hits despite the exact same test asserting the value they
  construct — verified via raw lcov `DA:` counts), the 3-line
  `with_entries` fix documented above as genuinely unreachable, and the
  rest are trivial `panic!`/`other =>` fallback arms inside the new
  tests' own `match` statements.

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -6626,15 +6626,14 @@ fn builtin_any<W: Clone + AsRef<[u64]>>(
     value: StandardJson<'_, W>,
     optional: bool,
 ) -> QueryResult<'_, W> {
+    let owned_bool = |r: Result<bool, EvalError>| {
+        r.map_or_else(QueryResult::Error, |b| {
+            QueryResult::Owned(OwnedValue::Bool(b))
+        })
+    };
     match value {
-        StandardJson::Array(elements) => match any_all_over(elements, true) {
-            Ok(b) => QueryResult::Owned(OwnedValue::Bool(b)),
-            Err(e) => QueryResult::Error(e),
-        },
-        StandardJson::Object(fields) => match any_all_over(fields.map(|f| f.value()), true) {
-            Ok(b) => QueryResult::Owned(OwnedValue::Bool(b)),
-            Err(e) => QueryResult::Error(e),
-        },
+        StandardJson::Array(elements) => owned_bool(any_all_over(elements, true)),
+        StandardJson::Object(fields) => owned_bool(any_all_over(fields.map(|f| f.value()), true)),
         _ if optional => QueryResult::None,
         _ => QueryResult::Error(EvalError::cannot_iterate(&to_owned(&value))),
     }
@@ -6647,15 +6646,14 @@ fn builtin_all<W: Clone + AsRef<[u64]>>(
     value: StandardJson<'_, W>,
     optional: bool,
 ) -> QueryResult<'_, W> {
+    let owned_bool = |r: Result<bool, EvalError>| {
+        r.map_or_else(QueryResult::Error, |b| {
+            QueryResult::Owned(OwnedValue::Bool(b))
+        })
+    };
     match value {
-        StandardJson::Array(elements) => match any_all_over(elements, false) {
-            Ok(b) => QueryResult::Owned(OwnedValue::Bool(b)),
-            Err(e) => QueryResult::Error(e),
-        },
-        StandardJson::Object(fields) => match any_all_over(fields.map(|f| f.value()), false) {
-            Ok(b) => QueryResult::Owned(OwnedValue::Bool(b)),
-            Err(e) => QueryResult::Error(e),
-        },
+        StandardJson::Array(elements) => owned_bool(any_all_over(elements, false)),
+        StandardJson::Object(fields) => owned_bool(any_all_over(fields.map(|f| f.value()), false)),
         _ if optional => QueryResult::None,
         _ => QueryResult::Error(EvalError::cannot_iterate(&to_owned(&value))),
     }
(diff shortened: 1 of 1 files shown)


newhoggy:
Code review found and I fixed two confirmed regressions, both live-verified before and after the fix:

1. **`any_all_f` (`any(cond)`/`all(cond)`) lost short-circuiting.** An earlier version of this fix eagerly `to_owned_checked`-collected every element upfront before `cond` ever probed any of them — so a decode failure past the deciding element aborted a call that should have short-circuited before reaching it. Oracle-verified: `[1, <bad-string>] | any(.==1)` is `true` in real jq, which never needs to decode the second element. Restructured (`any_all_f_over`, generic over the Array/Object element iterator, mirroring `any_all_over`'s own existing shape) so each element is converted right before its own `cond` probe.

2. **`eval_pipe`'s `Many`-branch promotion loop silently corrupted data.** The moment any sibling forced an owned result, the whole batch got promoted from lazily-borrowed cursors via unchecked `to_owned` — `.[] | (if type == "string" then . else . + 0 end)` on `["\xff\xfe", 1]` silently returned `ManyOwned([String(""), Int(1)])` instead of raising. Fixed via two new helpers (`push_promoted`, `promote_borrowed_checked`) that preserve the existing `#400` "already-piped outputs survive a downstream error" policy: on a decode failure, whatever converted successfully before it stays as the `Partial` prefix.

`merge_owned` (the sibling helper used by this same loop's terminal `Error`/`Break`/`Halt`/`Partial` control arms to build their own prefix) has the identical unchecked-`to_owned` shape but is broader shared infrastructure used by other functions too — filed as #1832 rather than folded in here.

Also applied three low-risk simplifications from the review's reuse/simplification angles (collapsed `builtin_any`/`builtin_all`'s duplicated match blocks, `eval_rhs_once`'s `Many` arm to a single `map_or`, `builtin_transpose`'s two temp-variable bindings) — all semantics-preserving, no behavior change.

Full local verification re-run clean after every fix (build/clippy/no_std/full test suite each time). Patch coverage now 96.44% (298/309); remaining gaps are the documented genuinely-unreachable `with_entries` site, confirmed llvm-cov attribution artifacts (verified via raw lcov `DA:` hit counts against passing tests), and trivial test-only fallback arms.

Issue #1660, opened 2026-08-26
jq: decode-failure catchability sweep (#1620) missed several call sites — some still swallow it, one silently computes a wrong path


newhoggy:
## Summary

PR #1647 (closing #1620) made a decode failure (invalid UTF-8, an unpaired
surrogate, a malformed escape/number) uncatchable by both `?` and
`try`/`catch`, by giving `EvalError` a `decode_failure()`/`is_decode_failure()`
pair and checking it first at every `?`/`try`/`catch` boundary it swept. The
sweep covered `eval_generic.rs`'s dispatch and ~18 sites in `eval.rs`'s own
builtins, but several sibling code paths were missed. Found during a fresh
review of that PR's merged diff.

All of these are reachable only through the public `succinctly::jq::eval()`
library API, **not** through the `succinctly jq`/`succinctly yq` CLI binaries
— the CLI's `eval_generic.rs` reindex bridge already materializes/validates
the whole document before any of this code runs, so nothing here is a live
CLI regression. But it's exactly the surface PR #1647's own new tests
(`test_optional_does_not_suppress_decode_failure_1620`,
`test_try_catch_does_not_catch_decode_failure_1620`) target, so it's in scope
for the same contract that PR establishes.

## 1. `path(.[.a])` silently computes a wrong path (most severe — data corruption, not just suppression)

`eval.rs` has its own, independent `to_owned_at_depth` — a second lazy
`StandardJson`-to-`OwnedValue` conversion distinct from the one in
`eval_generic.rs` that #1620 fixed. It silently drops a malformed object key,
and silently coerces an undecodable string used as a dynamic index into `""`,
instead of raising. Confirmed live via the `eval()` API:

```
path(.[])       on {"<bad-utf8-key>": 1, "b": 2}   => ["b"]        (should raise)
path(.[.a])     on {"a": "<bad-utf8>", "": 99}     => [""]         (should raise)
```

The second case isn't merely a suppressed error — it computes and returns a
materially wrong path, because the undecodable string is silently coerced to
`""` and used to index. `builtin_path` is one of ~251 call sites of this
`to_owned_at_depth`.

## 2. `?//` pattern-alternative retry loops don't check `is_decode_failure()`

`try_pattern_alternatives` (backs `?//` destructuring) and its lazy/streaming
twin `each_pattern_alternatives` — explicitly documented in the code as
needing to "agree" with each other — retry the next pattern alternative on
*any* error from the current one, including a decode failure, when the
[...]

PR #1836, opened 2026-08-29
fix(jq): make colliding_display_key_error actually uncatchable

### LINKED ISSUE #1813: jq: colliding_display_key_error isn't actually uncatchable, contrary to decode_failure's contract
Found reviewing PR #1807 (#1734) -- confirmed pre-existing, reproduces identically against unmodified `main` (present since #1642).

## Summary

`colliding_display_key_error` (`src/jq/document.rs:961-966`) is built via `EvalError::decode_failure(...)`, whose doc comment (`src/jq/error.rs:964`) states the resulting error "must never be suppressed by `?` or caught by `try`/`catch`." But `is_decode_failure()` (`error.rs:978`) classifies purely by matching `message` against a fixed literal set (`"invalid UTF-8 in string"`, `"invalid number format"`, etc.) -- `"object key \"...\" is ambiguous: ..."` matches none of them, so `is_uncatchable()` (`is_invalid_path_expression() || is_decode_failure()`) is `false` for it.

## Repro

```console
$ printf '[{"\z": 1, "\z": 2}]' | succinctly jq 'try sort catch .'
"object key \"\\z\" is ambiguous: an undecodable key's display form collides with another key of the same name and cannot be represented"
$ echo $?
0

$ printf '[{"\z": 1, "\z": 2}]' | succinctly jq 'sort?'
$ echo $?
0
```

Both `try`/`catch` and `?` treat the collision error as an ordinary catchable error -- contrary to the "never caught" contract `decode_failure`'s own doc comment states, and inconsistent with every other `decode_failure`-constructed error (a plain undecodable string or key correctly stays uncatchable).

## Why this matters more after #1734

`to_owned_checked_at_depth` (fixed by #1734) is the first call site in `eval.rs` to give `DisplayKeyGuard` an instance to actually engage -- meaning ordinary jq idioms like `sort?`/`delpaths(...)?` on a document with colliding decode-failure keys can now silently suppress this error where the codebase's own convention says it shouldn't. Confirmed live: `delete_paths_sorted`'s existing `Err(e) if optional && !e.is_uncatchable() => QueryResult::None` gate (added by #1746 review) will suppress this specific error under `delpaths(...)?`, since `is_uncatchable()` is `false` for it -- consistent with this bug's classification, not a separate one, but a concrete place it now bites.

## Suggested direction

Not attempted. Either give `colliding_display_key_error` its own dedicated `EvalError` constructor/kind that `is_uncatchable()` recognizes (matching how other genuinely-uncatchable errors are distinguished), or extend `is_decode_failure()`'s message-matching to also recognize the "is ambiguous" text -- the former is more robust against a future wording change to the message silently breaking classification again.

## Related

- #1734 -- the fix whose review found this.
- #1642 -- introduced `colliding_display_key_error`/`DisplayKeyGuard`.
- #1620 -- established the general decode-failure-must-never-be-suppressed convention.

newhoggy:
Fixes #1813

## Summary

`colliding_display_key_error` (#1642) is built via `EvalError::decode_failure`, whose contract says the result "must never be suppressed by `?` or caught by `try`/`catch`" (#1620). But `is_decode_failure()` classifies purely by matching `message` against a fixed literal list, and this error's dynamic `object key "<key>" is ambiguous: ...` text matched none of them — so `?` and `try`/`catch` both treated it as an ordinary catchable error.

## Repro (confirmed live before this fix)

```console
$ printf '[{"\z": 1, "\z": 2}]' | succinctly jq 'sort?'
$ echo $?
0
```

Silent — a real corruption signal (two undecodable keys colliding under the same display fallback) vanishes with no diagnostic at all, exactly the failure class #1620/#1642 exist to close.

## Fix

Followed this codebase's own established precedent rather than the dedicated-field alternative the issue names: a boolean field on `EvalError` was tried once already for `is_decode_failure()` itself and reverted, because the extra width propagated into `QueryResult`/`Control`'s recursive-materializer call frames and turned a controlled panic into a genuine stack overflow during unwind (#1021 — documented on `is_decode_failure()`'s own doc comment).

Extended the same zero-size-cost message-matching technique instead:
- Added `EvalError::colliding_display_key(key)`, co-locating the message's construction with a private fixed-suffix constant, the same way `invalid_path_expression` already does with its own prefix constant.
- `is_decode_failure()` now also matches that suffix.
- `document.rs`'s `colliding_display_key_error` delegates to the new constructor instead of hand-building the message via `format!`.

The existing `self.value.is_some()` guard (#1660 — true only for `error(v)`) already covers the obvious false-positive: a user's own `error("...")` call using text that happens to end with the same fixed suffix stays ordinarily catchable — confirmed by a new test.

## Testing

- New tests cover `sort?`, `try sort catch .`, and `delpaths(...)?` (the specific case #1746's review already gated on `is_uncatchable()`) all correctly letting this error escape uncaught.
- A companion test confirms a user's own `error(...)` call with the exact matching suffix text stays catchable (no new false positive).
- Verified against **both** evaluators — `error.rs`/`document.rs` are shared code, and a yq-mode test (using an invalid escape rather than invalid UTF-8 bytes, since YAML's own parser rejects the latter before the evaluator sees it) confirms the fix applies there too.
- All pre-existing `#1642`/`#1620`/`#1660`/`#1746` tests still pass unchanged.
- Full `cargo test --features cli,regex` (69/69 suites green), both CI clippy legs, `cargo fmt --check`, `cargo build --no-default-features`, `cargo doc --all-features`.

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -9,8 +9,8 @@ use std::io::{BufWriter, IsTerminal, Read, Write};
 use std::path::Path;
 
 use succinctly::jq::document::{
-    colliding_display_key_error, effective_keys, resolve_display_key, DisplayKeyGuard,
-    DocumentCursor, DocumentElements, DocumentFields, DocumentValue, IndentSpec,
+    effective_keys, resolve_display_key, DisplayKeyGuard, DocumentCursor, DocumentElements,
+    DocumentFields, DocumentValue, IndentSpec,
 };
 use succinctly::jq::eval_generic::{
     assert_nesting_depth, eval_with_cursor_using, to_owned as generic_to_owned,
@@ -358,7 +358,10 @@ fn yaml_to_owned_value<W: AsRef<[u64]>>(cursor: YamlCursor<'_, W>) -> Result<Own
                 let (key, is_fallback) = field.key().key_string_kind();
                 let key = key.into_owned();
                 if !guard.check(&map, &key, is_fallback) {
-                    return Err(anyhow::anyhow!("{}", colliding_display_key_error(&key)));
+                    return Err(anyhow::anyhow!(
+                        "{}",
+                        EvalError::colliding_display_key(&key)
+                    ));
                 }
                 let value = yaml_to_owned_value(field.value_cursor())?;
                 map.insert(key, value);
--- a/src/jq/document.rs
+++ b/src/jq/document.rs
@@ -1097,32 +1097,16 @@ impl DisplayKeyGuard {
     }
 }
 
-/// The error for two keys that collide only because a decode-failure key's
-/// display fallback (#1642) matches another key's spelling.
-///
-/// #1385 forbids treating them as the same key, but a display-keyed map has
-/// no way to hold both, so this is what `to_owned`/`materialize` raise
-/// instead of silently dropping one of them -- see [`DisplayKeyGuard`].
-///
-/// `pub`, not `pub(crate)`: `succinctly-cli`'s `yq_runner.rs`
-/// (`yaml_to_owned_value`, #1749) reuses this directly rather than
-/// hand-writing a second, wording-drift-prone copy of the same message --
-/// same reasoning [`DisplayKeyGuard::check`] itself is `pub` for. The
-/// wording says "object key", which still reads correctly for a YAML
-/// mapping key (a YAML document *is* the object/mapping distinction JSON
-/// doesn't have, but "object" is the neutral term this message already
-/// uses for JSON's own object keys).
-pub fn colliding_display_key_error(key: &str) -> EvalError {
-    EvalError::colliding_display_key(key)
-}
-
 /// The full display-key resolution sequence a `to_owned`-shaped materializer
 /// needs for one field, in one call.
 ///
 /// Gets the display spelling (`None` when the key does not stringify at all
 /// -- #1194's territory, the caller's own job to handle), guards it against
 /// a colliding decode-failure fallback (#1642), and raises
-/// `colliding_display_key_error` instead of allowing a silent overwrite.
+/// [`EvalError::colliding_display_key`] instead of allowing a silent
+/// overwrite -- #1385 forbids treating two colliding keys as the same one,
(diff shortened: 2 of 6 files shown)


newhoggy:
Round 2, addressing the full 10-angle review (all 10 finder agents completed with no correctness bugs — a clean result this time).

**Applied:** the one confirmed, actionable simplification (found independently by 2 angles) — `document.rs`'s `colliding_display_key_error` was a fully redundant one-line wrapper around `EvalError::colliding_display_key`; all three call sites already had direct, unconditional access to `EvalError`. Deleted it, moved the "why this crosses two crates" rationale onto the real constructor, updated all three call sites and their imports (including removing a now-unnecessary `#[cfg(feature = "std")]`-gated import in `eval.rs`).

**Added:** a test for a minor, previously-safe-but-untested gap the cross-file tracer flagged — `load()`'s YAML path (`yaml_value_to_owned_checked`) is a third call site for this error, and nothing pinned that `try`/`catch` around `load()` itself still can't suppress it.

**Filed #1840** for the review's most significant finding, deliberately not acted on in this narrowly-scoped bug fix: a niche-sharing enum replacing `EvalError`'s `value: Option<OwnedValue>` field can carry a typed `ErrorKind` tag at a *measured, verified* size of 72 bytes — identical to today's `Option<OwnedValue>`, because `OwnedValue`'s discriminant already has spare niche capacity. This directly refutes the #1021 revert's assumption that any typed classification necessarily costs extra space, and would close this whole recurring string-matching-classifier bug class (already hit twice: #1660, #1813) — but it's a real redesign of `EvalError`'s core representation, not a fit for this PR.

**Not acted on** (confirmed safe/inactive by review, not worth the added complexity): a latent, currently-unreachable interaction between the `" in object key"` suffix-stripping and the new suffix check; `EvalError`'s fields being fully `pub` with the "only `error(v)` sets `.value`" invariant enforced by convention rather than the type system (both tied to #1840's broader redesign discussion, not fixable in isolation here).

No behavior change from this round — message text and classification logic are identical, only where the constructor lives. Re-verified: full `cargo test --features cli,regex` (55/55 green), both CI clippy legs, `cargo fmt --check`, `cargo build --no-default-features`, `cargo doc --all-features`.

PR #1839, opened 2026-08-29
fix(jq): -R joins a line across a file boundary, matching --seq's #1571 fix

### LINKED ISSUE #1809: jq: -R (raw-input) joins a line across a file boundary where real jq does not join across files' boundaries correctly either way — verify exact rule
## Summary

Found during `/code-review` on PR #1808 (#1571's \`--seq\` cross-file-boundary fix). That
PR's own doc comment claimed \`-R\`/raw-input is safe from the identical class of bug
because it's "line-oriented (no multi-byte record to split across a boundary)" — that claim
is wrong, confirmed live against the pinned oracle.

## Repro (confirmed live against \`/usr/bin/jq\` 1.7.1 and unmodified succinctly \`main\`)

Two files, the first with no trailing newline:

\`\`\`console
$ printf 'abc' > f1; printf 'def\nghi\n' > f2
$ jq -R -c '.' f1 f2
"abcdef"
"ghi"

$ succinctly jq -R -c '.' f1 f2
"abc"
"def"
"ghi"
\`\`\`

Real jq's \`-R\` reader treats multiple files as one continuous byte stream for line-splitting
purposes too (matching \`-s\`'s own model, and \`--seq\`'s, per #1571) — a line without a
trailing newline joins with the next file's first line. succinctly's \`-R\` handling
(\`get_inputs\`'s \`args.raw_input\` branch, \`raw.lines().enumerate()\` per file) treats each
file as an independent set of lines instead, so \`"abc"\` (file 1's incomplete trailing
"line") and \`"def"\` (file 2's first line) never get joined.

## Scope note

Not investigated beyond confirming the repro. #1571/PR #1808 built a
concatenate-and-remap-offsets pattern (\`build_seq_values\`) for exactly this class of
cross-file-boundary problem in \`--seq\` mode — a similar approach (concatenate all raw-input
files' bytes before line-splitting, remap each line back to its owning file/line for
\`input_line_number\`/error-location purposes) is a plausible starting point here too, but the
non-slurp location-tracking side of \`-R\` may have its own subtleties (per-line locations,
not per-record) worth verifying against the oracle carefully rather than assuming symmetry
with \`--seq\`'s exact shape.

## Related

- #1571 / PR #1808 — the \`--seq\` fix this bug's sibling was found alongside; not touched by
  that PR (deliberately out of scope, DSV was not independently verified there either and
  may have the same underlying gap).

newhoggy:
## Summary
- Real jq's `-R` (raw-input) reader treats multiple input files as one continuous byte stream for line-splitting, so a file's own unterminated trailing line joins with the next file's first line — confirmed live against `/usr/bin/jq` 1.7.1. succinctly's `-R` handling split each file into lines independently instead (`raw.lines().enumerate()` per file), producing an extra spurious line at every unterminated file boundary.
- Adds `build_raw_input_values`, mirroring `build_seq_values`'s (#1571/PR #1808) concatenate-then-remap pattern: concatenate every file's raw bytes, split into lines once across the whole combined stream, then remap each line's end offset back to its owning file and file-local line number via the same `LineCounter`/`partition_point` approach — reused verbatim, not reimplemented.
- The dead per-file `raw_input` arm inside the loop is removed; the new dispatch branch (`args.raw_input && args.input_dsv.is_none()`) intercepts every case that arm used to handle, since the `-R -s` early return already covers the one remaining combination.
- Side effect (not scope creep — same code path, correct-by-construction): a final unterminated line's `input_line_number`/error-location attribution is now also correct in the *single-file* case, matching real jq exactly (a final line with no trailing `\n` reports the *previous* completed line's number, confirmed live: `printf 'abc\ndef' | jq -R -c '., input_line_number'` reports `1` for both lines). No prior test covered this since every existing `-R` fixture was fully `\n`-terminated.

## Test plan
- [x] Live-verified the issue's own repro, a 3-file join, `input_line_number`, and error-location attribution all match `/usr/bin/jq` 1.7.1 exactly
- [x] `-R -s` (early-return path) and `-R --input-dsv` (still per-file) confirmed unaffected
- [x] New tests: `test_jq_raw_input_line_joined_across_file_boundary_1809`, `test_jq_raw_input_line_number_after_cross_file_join_1809`, `test_jq_raw_input_error_location_after_cross_file_join_1809`
- [x] `cargo test --features cli,simd,regex,serde` — full suite green, no regressions
- [x] `cargo fmt --check`

Fixes #1809

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -2954,40 +2954,16 @@ fn build_seq_values(
     locations: &mut InputLocations,
     slurp: bool,
 ) -> Vec<OwnedValue> {
-    let mut combined = String::new();
-    let mut file_ends: Vec<usize> = Vec::with_capacity(raw_inputs.len());
-    for (_, raw) in raw_inputs {
-        combined.push_str(raw);
-        file_ends.push(combined.len());
-    }
-
+    let (combined, file_ends) = concat_with_file_ends(raw_inputs);
     let parsed = parse_json_seq_with_ends(&combined);
 
     if !slurp {
-        let mut current: Option<(usize, LineCounter<'_>)> = None;
-        for &(_, end) in &parsed {
-            let file_idx = file_ends
-                .partition_point(|&fe| fe < end)
-                .min(file_ends.len().saturating_sub(1));
-            if current.as_ref().map(|(idx, _)| *idx) != Some(file_idx) {
-                current = Some((
-                    file_idx,
-                    LineCounter::new(raw_inputs[file_idx].1.as_bytes()),
-                ));
-            }
-            let file_start = if file_idx == 0 {
-                0
-            } else {
-                file_ends[file_idx - 1]
-            };
-            let src = raw_inputs[file_idx].0.unwrap_or(0);
-            let line = current
-                .as_mut()
-                .expect("just set above")
-                .1
-                .advance_to(end.saturating_sub(file_start));
-            locations.push(src, line);
-        }
+        remap_ends_to_locations(
+            parsed.iter().map(|&(_, end)| end),
+            raw_inputs,
+            &file_ends,
+            locations,
+        );
     }
 
     parsed.into_iter().map(|(v, _)| v).collect()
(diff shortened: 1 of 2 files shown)


newhoggy:
Code review found real issues in the first version, all confirmed live against jq 1.7.1 and now fixed:

1. **File-boundary off-by-one**: `file_ends.partition_point(|&fe| fe < end)` misattributed a line/record to the wrong (earlier) file whenever its end offset landed exactly on a file boundary — e.g. a file whose sole content is the `\n` terminating the previous file's unterminated trailing text. Fixed to `fe <= end`. This turned out to be a pre-existing bug shared with `build_seq_values` (`--seq` mode, #1571/#1808), not something introduced here — extracted shared `concat_with_file_ends`/`remap_ends_to_locations` helpers to fix both call sites at once and remove the duplication that let them drift apart.
2. **CRLF**: `-R` was silently stripping a trailing `\r` to mirror `str::lines()`, but real jq's raw-input reader never strips `\r`. Pre-existing (the old per-file `raw.lines()` loop had the same bug), fixed while already rewriting this function.
3. Added test coverage for the single-file unterminated-line `input_line_number` quirk the original doc comment claimed to fix but nothing exercised.

New commit: 73608d8e7. Full suite/clippy/fmt/doc/no_std all re-verified green.

Issue #1615, opened 2026-08-25
yq: a bad escape in a streamed scalar still silently degrades to null/"" instead of raising (#1247 Stage 6)


newhoggy:
## Summary

#1247 (PR #1391) routed almost every string-decode-failure materialization path through a
real `EvalError` — but three streaming *output* writers were left as a deliberately-deferred
gap ("Stage 6" in
[`docs/plan/decode-failure-routing.md`](docs/plan/decode-failure-routing.md)). Their only
error channel is `core::fmt::Result`, which carries no message, so a mid-stream failure there
can only be silently absorbed or cause a bare, undiagnosed abort — the design doc chose
silent absorption, on purpose, and split closing the gap into its own follow-up stage rather
than block the rest of #1247 on it.

This issue tracks that follow-up, now with three known sites instead of the one the design
doc originally scoped.

## Live-verified sites (built from #1247's own landed tree)

```console
$ printf 'a: 1\nb: "bad \q escape"\n' | succinctly yq -o=json '.'
{
  "a": 1,
  "b": null
}
$ printf 'a: 1\nb: "bad \q escape"\n' | yq -o=json '.'
Error: bad file '-': yaml: while scanning a quoted scalar at line 2, column 4: line 2, column 9: found unknown escape character

$ printf 'double: "quoted \q scalar"\n' | succinctly yq '.'
double: ""
$ printf '"a\qb": 1\nb: 2\n' | succinctly yq '.'
"": 1
b: 2
```

1. **`stream_json_value`/the YAML→JSON transcoders** (`src/yaml/light.rs`, reached via
   `succinctly yq -o=json`) — the site the design doc originally named.
2. **`stream_yaml_string_value`/`write_yaml_field_key`** (`src/yaml/light.rs`, the *default*
   YAML→YAML output path — no `-o` flag needed). This is the single most common `yq`
   invocation and was not previously recorded anywhere; found during #1247's post-merge
   review.
3. **`stream_json_as_yaml`'s `StandardJson::Error` arm** (`src/json/light.rs`, the JSON→YAML
   streaming output path) — architecturally the same gap (same `core::fmt::Write`-only
   channel), currently miscategorized in the design doc's site inventory as a landed Stage 4
   fix rather than folded into Stage 6 alongside its siblings.

Every *materializing* route (a `--arg`-forced DOM, a multi-result filter, `to_entries`,
`length`) already raises correctly via #1247; only these three purely-streamed writers do
not.

## What a fix needs

[...]

PR #1842, opened 2026-08-29
fix(jq): nth/limit accept NaN and non-integer counts the way jq does

### LINKED ISSUE #1825: jq: nth(nan) returns a value where jq raises, nth() with a non-integer n picks the wrong element, and limit() rejects a float count jq accepts
**Severity: Low-Medium** (wrong answers and a swallowed error on a narrow input class; fully
oracle-pinned, no design question)

Split out of #1687 (item 4, second bullet), which noted that `classify_nth_n` "silently accepts
`NaN` as index `0` and truncates a positive non-integer float instead of rounding". Probing that
against the pinned oracle found **three** distinct divergences, not one, and located the actual
mechanism — so this is filed separately with the full matrix. Everything below is live-verified
against `/usr/bin/jq` 1.7.1 and a current `--release --features cli` build.

## Divergence 1 — `nth(nan; g)` returns a value where jq raises

|---|---|---|
| `[nth(nan; 1,2,3)]` | error `nth doesn't support negative indices`, **exit 5** | `[1]`, **exit 0** |
| `[nth(-nan; 1,2,3)]` | error, exit 5 | `[1]`, exit 0 |

The guard is the bug: succinctly tests `n < 0`, which NaN fails, so NaN falls through to the
success path and classifies as index 0. jq's guard is effectively `!(n >= 0)`, which NaN also
fails — but in the *raising* direction. Matching jq means inverting the comparison, not adding a
special NaN case.

Controls that already agree (so the fix must not disturb them):

| filter | both |
|---|---|
| `[nth(-0.5; 1,2,3)]` | error, exit 5 |
| `[nth(-infinite; 1,2,3)]` | error, exit 5 |
| `[nth(-0.0; 1,2,3)]` | `[1]` |
| `[nth(0; 1,2,3)]` | `[1]` |
| `[nth(3.0; 1,2,3)]` | `[]` |
| `[nth(infinite; 1,2,3)]` | `[]` |

## Divergence 2 — a non-integer `n` picks the wrong element

|---|---|---|
| `[nth(0.4; 1,2,3)]` | `[2]` | `[1]` |
| `[nth(1.7; 1,2,3)]` | `[3]` | `[2]` |

Not an off-by-one and not "rounding": it falls out of jq's own definition
`nth($n; g) = last(limit($n + 1; g))`, where `limit` stops at the first element whose 1-based
count satisfies `count >= $n` — a **float** comparison. For `n = 0.4` that is `limit(1.4)`, which
takes 2 elements (1 >= 1.4 is false, 2 >= 1.4 is true), and `last` is the second. Confirmed by
evaluating the definition by hand in jq:

```console
$ jq -nc '[last(limit(0.4 + 1; 1,2,3))]'   # [2]   — matches nth(0.4)
$ jq -nc '[last(limit(1.7 + 1; 1,2,3))]'   # [3]   — matches nth(1.7)
```

succinctly truncates `n` to an integer instead, which floors where jq effectively ceilings.

## Divergence 3 — `limit` rejects a non-integer count that jq accepts

This is the mechanism behind divergence 2, and is independently wrong:

|---|---|---|
| `[limit(0.4; 1,2,3)]` | `[1]` | error `limit requires non-negative integer` |
| `[limit(1.4; 1,2,3)]` | `[1,2]` | error |
| `[limit(2.7; 1,2,3)]` | `[1,2,3]` | error |
| `[limit(nan; 1,2,3)]` | `[1,2,3]` | error |
| `[limit(-1; 1,2,3)]` | `[1,2,3]` | `[1,2,3]` (agrees) |

jq's `limit` is `if $n > 0 then <take while count < $n> elif $n == 0 then empty else f end`, so a
NaN or negative `$n` falls to the `else` branch and passes *everything* through — which is why
`limit(nan; ...)` yields all three and `limit(-1; ...)` already agrees.

## Not affected — the 1-arg array forms already match

```console
$ echo '[1,2,3]' | jq -c 'nth(1.7)'  # 2   (succinctly: 2)
$ echo '[1,2,3]' | jq -c 'nth(0.4)'  # 1   (succinctly: 1)
$ echo '[1,2,3]' | jq -c 'nth(nan)'  # null (succinctly: null)
```

`nth/1` is a plain array index and is unrelated to `classify_nth_n`. Fixing the generator forms
must not touch it.

## Where

`classify_nth_n` (shared by `eval_nth_generic` and `builtin_nth_stream` per #1313), and `limit`'s
own count validation. Note the sharing: a fix to `classify_nth_n` lands on both call sites at
once, so both need coverage.

## Suggested fix

1. `limit`: accept any non-integer float count, implementing jq's `count >= $n` float comparison
   and its `$n > 0 / == 0 / else` three-way branch, rather than requiring an integer.
2. `nth/2`: raise on `!(n >= 0)` — covering negative **and** NaN — and otherwise derive the
   element from the corrected `limit`, so divergence 2 disappears without a second rounding rule.
3. Regression tests for every row of all three tables above, in both jq and yq mode
   (`classify_nth_n` is shared, so a jq-only test would leave the yq path unpinned — see the
   repeated lesson in this repo's history about mode-scoped fixes to shared helpers).

## Related

- #1687 — the umbrella this is split from (its items 1-3 are architectural and stay there).
- #1313 — established the shared `classify_nth_n` to stop the classification being re-derived.
- #1607 / #1686 — added the native `limit`/`nth` arms this classification feeds.

newhoggy:
Fixes #1825

## Summary

Three related divergences, all in `classify_nth_n`/`classify_limit_n` (shared by both evaluators per #1313, and by jq/yq mode alike):

1. **`nth(nan; g)` returned a value where jq raises.** `classify_nth_n` tested `f < 0.0`, which NaN fails (every IEEE 754 comparison against NaN is `false`), so NaN silently classified as index `0`. jq's own guard is effectively `!(n >= 0)`, which also catches NaN.
2. **`nth($n; g)` for a positive non-integer `$n` picked the wrong element.** `classify_nth_n` truncated `$n`; jq's own definition is `nth($n; g) = last(limit($n + 1; g))`, and `limit` takes `ceil($n + 1)` outputs for a positive count — so the correct 0-based stop index is `ceil($n)`.
3. **`limit($n; g)` rejected any non-integer count outright.** jq's own definition accepts a fractional `$n`, taking `ceil($n)` outputs for a positive count.

## Verification

Every row of all three divergence tables, plus every pre-existing control case, verified live against the pinned jq 1.7.1 binary:

```
[nth(nan; 1,2,3)]        jq: error, exit 5    →  now matches (was [1], exit 0)
[nth(0.4; 1,2,3)]        jq: [2]              →  now matches (was [1])
[nth(1.7; 1,2,3)]        jq: [3]              →  now matches (was [2])
[limit(0.4; 1,2,3)]      jq: [1]              →  now matches (was an error)
[limit(1.4; 1,2,3)]      jq: [1,2]            →  now matches (was an error)
[limit(2.7; 1,2,3)]      jq: [1,2,3]          →  now matches (was an error)
[limit(nan; 1,2,3)]      jq: [1,2,3]          →  now matches (was an error)
```

Controls unaffected: `nth(-0.5|-infinite)`, `nth(-0.0|0|3.0|infinite)`, `limit(-1)`, `nth/1` (the plain array-index form), all `#983` cases.

## Fix

- `classify_nth_n`: `f.is_nan() || f < 0.0` instead of `f < 0.0` (written this way, not the equivalent negated `!(f >= 0.0)`, per `clippy::neg_cmp_op_on_partial_ord`). The float fallback now ceils instead of truncates — `owned.as_i64()` is still preferred first for an exact integer, since converting to `f64` at all would already lose precision past its 53-bit mantissa.
- `classify_limit_n`: implements jq's exact three-way `if $n > 0 / elif $n == 0 / else` branch for the float case, taking `ceil($n)` for a positive count instead of erroring on any non-integer.

Two existing tests pinned the old, now-incorrect behavior as *deliberate* — updated rather than left to bit-rot into false negatives:
- `test_limit_positive_float_count_still_errors_983` → `test_limit_positive_float_count_takes_ceil_983`
- `builtin_limit_n_classification_matches_classify_limit_n` (its own assertion inverted from `Error` to the correct `ManyOwned([1,2,3])`)

## Testing

- New CLI tests cover every row of all three tables in jq mode, plus the `nth/1` unaffected-form control.
- A yq-mode test confirms the shared classification fix applies identically there (gated behind `--jq-extensions`, since `nth`/`limit` aren't part of real yq's own grammar).
- Full `cargo test --features cli,regex` (55/55 green), both CI clippy legs, `cargo fmt --check`, `cargo build --no-default-features`, `cargo doc --all-features`.

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -3357,6 +3357,32 @@ pub(crate) enum LimitN {
     Take(usize),
 }
 
+/// jq's `limit($n; f)`/`nth($n; f)` both take `ceil($n)` outputs of `f` for
+/// a positive `$n` (`limit`'s own definition -- `if $n > 0 then <take,
+/// decrementing $n by 1 per output, stop once the decremented count is <=
+/// 0> elif $n == 0 then empty else f end` -- consumes one whole unit of
+/// `$n` per iteration, so a fractional remainder still needs one more
+/// output to cross zero). `f > 0.0` is the caller's contract, not
+/// re-checked here (`debug_assert!` only): both call sites already branch
+/// on sign themselves, for a `LimitN::Take(0)`/`Unlimited` distinction this
+/// function has no reason to know about.
+///
+/// A single shared definition rather than one ceiling formula per caller,
+/// #1825 review: an earlier version of `classify_nth_n` re-derived this
+/// independently via the real-number identity `ceil(x + 1) == ceil(x) +
+/// 1`, true for exact reals but not for the rounded `f64` addition jq's
+/// own `$n + 1` actually performs -- the two diverge by one whenever `$n`
+/// crosses a power-of-two ULP-doubling boundary
+/// (`nth(1.0000000000000002; range(1;30))` is `2` in jq, confirmed live;
+/// the old formula computed `3`). Callers now perform that same rounded
+/// addition themselves (`classify_nth_n` passes `f + 1.0` here) instead of
+/// this function re-deriving an equivalent, so there is exactly one
+/// ceiling computation both share, not two independently-reasoned copies.
+fn ceil_positive_float_to_usize(f: f64) -> usize {
+    debug_assert!(f > 0.0);
+    f.ceil() as usize
+}
+
 pub(crate) fn classify_limit_n(n_value: OwnedValue) -> Result<LimitN, EvalError> {
     match n_value {
         OwnedValue::Int(i) | OwnedValue::NumberLiteral(NumberRepr::Int(i), _) if i >= 0 => {
@@ -3366,21 +3392,15 @@ pub(crate) fn classify_limit_n(n_value: OwnedValue) -> Result<LimitN, EvalError>
         | OwnedValue::NumberLiteral(NumberRepr::Int(_), _)
         | OwnedValue::Null
         | OwnedValue::Bool(_) => Ok(LimitN::Unlimited),
-        // jq's own `limit($n; f)` is `if $n > 0 then <take, decrementing $n
-        // by 1 per output, stop once the decremented count is <= 0> elif
-        // $n == 0 then empty else f end` -- a positive `$n` therefore takes
-        // `ceil($n)` outputs (each iteration consumes one whole unit of
-        // `$n`, and a fractional remainder still needs one more output to
-        // cross zero), not the integer-only count this function used to
-        // require. `$n == 0.0` (including `-0.0`, which compares equal to
-        // `0.0` under IEEE 754) takes the `$n == 0` branch; NaN satisfies
-        // neither `> 0` nor `== 0`, so it falls to the same unlimited `else
-        // f` branch as a negative `$n` (#1825 -- confirmed live against jq
+        // `$n == 0.0` (including `-0.0`, which compares equal to `0.0`
+        // under IEEE 754) takes the `$n == 0` branch; NaN satisfies neither
+        // `> 0` nor `== 0`, so it falls to the same unlimited `else f`
+        // branch as a negative `$n` (#1825 -- confirmed live against jq
         // 1.7.1: `limit(nan; 1,2,3)` passes everything through, same as
         // `limit(-1; 1,2,3)`).
         OwnedValue::Float(f) | OwnedValue::NumberLiteral(NumberRepr::Float(f), _) => {
             if f > 0.0 {
-                Ok(LimitN::Take(f.ceil() as usize))
+                Ok(LimitN::Take(ceil_positive_float_to_usize(f)))
             } else if f == 0.0 {
                 Ok(LimitN::Take(0))
             } else {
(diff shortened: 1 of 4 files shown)


newhoggy:
Round 2, addressing the full 10-angle review. This one caught a real, serious bug in my own initial fix — reporting in full since it's the headline result.

**Critical fix — a genuine off-by-one in `classify_nth_n`'s own formula**, confirmed live: the original fix computed the 0-based stop index as `ceil($n)`, reasoning from the real-number identity `ceil(x + 1) == ceil(x) + 1`. That identity is true for exact reals but **not** for the rounded `f64` addition jq's own `$n + 1` actually performs — the two diverge by one whenever `$n`'s fractional part is small enough to be rounded away as `$n + 1.0` crosses a power-of-two ULP-doubling boundary:

```
jq -n 'nth(1.0000000000000002; range(1;30))'   # => 2 in jq
                                                # => 3 with the original ceil($n) formula (WRONG)
```

Swept every `k = 2^m - 1` from 1 to 511 (the smallest `f64` greater than each) — all reproduced the same off-by-one. Fixed by computing `(f + 1.0).ceil() as usize - 1` directly (the same rounded addition jq performs, not a shortcut around it) via `saturating_sub(1)`. Verified against the full boundary sweep plus every row of the original divergence tables, which are unaffected (the two formulas only differ at these specific rounding boundaries). Added `test_nth_ulp_boundary_matches_jq_addition_order_1825` covering the whole sweep.

**Also fixed:**
- `eval_nth_expr` (`Expr::NthExpr`, confirmed parser-unreachable by 3 independent review angles) had its own independent, integer-only classification never touched by the original fix — a latent instance of exactly the #1313-style drift this whole `classify_nth_n`/`classify_limit_n` refactor exists to prevent. Replaced with a direct call to `classify_nth_n`; all 5 pre-existing unit tests for it still pass unchanged.
- `classify_limit_n`'s doc comment named its call sites and had already undercounted them before this PR. Reworded to describe the invariant instead of an enumerable list that goes stale the moment a new call site is added.

**Filed #1846** for a confirmed-live, out-of-scope finding: `skip(n; expr)` has the identical NaN/non-integer-count divergence, hand-duplicated twice within itself rather than sharing a classifier — a separate builtin with its own decrement semantics, not assumed identical to `limit`'s `ceil($n)`.

**Not acted on** (pre-existing, out of scope, in a different file's shared utility used far beyond these two functions): a theoretical `OwnedValue::as_i64()` epsilon-check misclassification at exactly `f = 2^63` (unobservable — no generator produces ~9.2 quintillion items), and 32-bit-`usize` saturation for an implausibly large float count (this crate has no stated 32-bit target, and the pattern already existed pre-PR for other inputs).

Also hit a real append-conflict during rebase (both this branch and a concurrently-merged main commit added new tests at the same file-tail location) — reconstructed both complete functions by hand rather than trusting the naive 3-way merge, verified by running both test sets afterward (all 7 pass, including the pre-existing #1615 tests from main).

Re-verified: full `cargo test --features cli,regex` (55/55 green), both CI clippy legs, `cargo fmt --check`, `cargo build --no-default-features`, `cargo doc --all-features`.

newhoggy:
## Round 3 review — consolidation + a real live bug found and fixed

Ran a third `/code-review` pass (4 finder angles) specifically because round 2's fix (`988392a09`) touched delicate floating-point boundary math. Two of the four angles independently found and confirmed **the same live bug**, since fixed:

### Confirmed bug: `classify_nth_n`'s `as_i64()` shortcut misclassified a narrow float band

The non-integer branch called `OwnedValue::as_i64()` as an "is this exactly an integer" shortcut before falling back to the `(f+1.0).ceil()-1` formula. `as_i64()` snaps to the nearest integer using a **fixed absolute epsilon** (`f64::EPSILON`), not a magnitude-relative one — so any `f` in roughly `[1.11e-16, 2.22e-16)` was misclassified as exactly integer `0`, giving index `0` instead of the correct `1`:

```
$ jq -n 'nth(1.5e-16; 1,2,3)'        # real jq 1.7.1
2
$ (pre-fix build) succinctly jq -n 'nth(1.5e-16; 1,2,3)'
1                                     # wrong
```

Fixed by rewriting `classify_nth_n` to destructure `Int`/`Float`/`NumberLiteral` directly (matching `classify_limit_n`'s existing shape) instead of double-dispatching through `as_f64()`/`as_i64()` — this removes the `as_i64()` call from the float path entirely, not just the epsilon window.

### Consolidation

Extracted `ceil_positive_float_to_usize` as the one shared "ceil a positive float" primitive for both `classify_limit_n` and `classify_nth_n`. `classify_nth_n`'s ceiling formula had been independently re-derived rather than reusing `classify_limit_n`'s — the exact class of drift #1313 already had to fix once for these two functions.

### Test fix

`test_nth_ulp_boundary_matches_jq_addition_order_1825`'s fixed `+ 0.0000000000000002` offset only survived f64 rounding at `k=1`'s magnitude — for `k=3..511` it was rounded away entirely, so 8 of the 9 rows silently tested the trivial integer case instead of the boundary the test exists to pin (confirmed: `3.0 + 0.0000000000000002 == 3.0` bit-for-bit). Rewrote it to build each row's boundary value from `k`'s own bit pattern + 1 ULP, so every row now genuinely exercises the boundary.

### Doc fixes

Two stale doc comments claiming `eval_nth_expr` still has a "narrower integer-only rule" — no longer true now that it delegates to `classify_nth_n` too (`eval_generic.rs`'s `eval_nth_generic` doc comment, and a comment in `test_limit_and_nth_control_flow_1607` in `tests/yq_cli_tests.rs`).

### Out-of-scope findings filed as follow-ups

- #1849 — `skip($n; expr)` has the same NaN-bypass and fractional-truncation bug class, same #983 lineage, never touched by this PR.
- #1850 — `path(limit(n; expr))` eagerly materializes the full result before truncating to `n`, instead of streaming like the value-context form does.

### Verification

Full `jq_cli_tests`/`yq_cli_tests` suites (1079 + 1271 tests), `eval::` unit tests, both CI clippy legs, `--no-default-features` (no_std) build, and `cargo doc --all-features` all green. Live-verified the fix and the ULP-boundary sweep against both `/usr/bin/jq` 1.7.1 and the rebuilt binary.

No drift against `main` — `origin/main` is exactly this branch's merge-base, so no rebase was needed.

Issue #1430, opened 2026-08-22
yq: a self-referencing multi-path assignment (.a[] = .a[] + 1) prints one extra document


newhoggy:
**Severity: Low**

## Summary

`.a[] = .a[] + 1` (RHS reads the same generator the LHS path resolves against) prints two documents in yq mode where real yq prints one.

## Repro (verified against yq v4.53.3)

```
$ echo '{"a":[1,2]}' | yq -o=json -I0 '.a[] = .a[] + 1'
{"a":[2,3]}

$ echo '{"a":[1,2]}' | succinctly yq -o=json -I0 '.a[] = .a[] + 1'
{"a":[2,2]}
{"a":[3,3]}
```

Also reproduces for the more general `.a[].b = .a[].b + 1` shape (a Field nested under the RHS-referencing Iterate).

## Discovery context

Found live during #1298's code review while stress-testing the new mid-chain `Iterate` fan-out. Confirmed **not** caused by #1298: reproduces identically for the *pre-existing*, untouched terminal `Expr::Iterate` arm (`.a[] = .a[] + 1`, no mid-chain component at all), and the responsible code (`eval_assign`'s RHS/paths folding loop, `src/jq/eval.rs`) is nowhere near either of #1298's two diff hunks (`split_at_iterate`/`set_path_through_iterate`).

## Scope note

Not investigated beyond confirming the repro and ruling out #1298 as the cause — likely a gap in how `eval_assign` folds a multi-valued RHS (here, the RHS's own generator producing one value per LHS path) across multiple LHS paths, printing one document per RHS output×path combination instead of correctly threading the write-in-progress document through. Filing to make this trackable, not blocking.

newhoggy:
Claiming — investigating the RHS/paths folding loop in `eval_assign` (src/jq/eval.rs) for the terminal `Expr::Iterate` self-referencing case.

Issue #392, opened 2026-07-26
jq: assignment collapses a multi-output RHS — `=` keeps the first, `|=` wraps them in an array


newhoggy:
**Severity: High** — `|=` produces a value jq never produces. Reproduced against the built CLI on the #359 branch at `8688adb9`; pre-existing.

jq fans an assignment out over its right-hand side: `=` emits one output per RHS value, and `|=` takes the filter's *first* output. Succinctly does neither.

```
$ echo '{}' | sjq -c '.a = (1,2)'
{"a":1}                     # jq: {"a":1} then {"a":2}   — second output dropped

$ echo '{"a":0}' | sjq -c '.a |= (1,2)'
{"a":[1,2]}                 # jq: {"a":1}                — an array appears from nowhere
```

The `|=` case is the worse of the two: `=` loses data, but `|=` *invents* a container. Nothing in the document or the filter asks for an array.

Two separate sites:

- **`=`** — `eval_assign` (`src/jq/eval.rs:5803`) materialises the RHS to one value, with `Many`/`ManyOwned` explicitly taking `.first()`. The comment says "use the first one"; jq's rule is one output per RHS value, so there is no first to take.
- **`|=`** — `update_path` runs the filter through `eval_owned_expr` (`src/jq/eval.rs:6981`), whose `Many` arm is:

  ```rust
  QueryResult::Many(vs) => {
      if vs.len() == 1 {
          Ok(to_owned(&vs[0]))
      } else {
          Ok(OwnedValue::Array(vs.iter().map(to_owned).collect()))
      }
  }
  ```

  The `len() == 1` special case is what hides this: every single-output filter behaves, so the array only appears once a filter starts fanning out. jq takes the first output here, not all of them.

Same family as #354 (object construction), #378 (if/select) and #353 (comma ordering) — a multi-output stream collapsed at a site that should fan out or pick. Distinct site, so filed separately; a shared fix may well cover several.

`eval_owned_expr` is also the "serialize and reparse" helper, so any fix should note it re-indexes the value on every path step.

newhoggy:
Correcting the fix sketch above: **do not change `eval_owned_expr`'s `Many` arm.** It has 21 call sites (`reduce`, `foreach`, `while`, `until`, `repeat`, `walk`, `limit`, …), and the array collapse is load-bearing for some of them.

[...]

Issue #1826, opened 2026-08-29
jq: extract the shared isolate-and-atomically-catch-optional helper (4 hand-written copies, one dormant Builtin::Map gap)


newhoggy:
**Severity: Low** (maintainability + one dormant risk; no live compliance divergence)

Split out of #1410, which bundles a mechanical consolidation with a much larger design problem.
This issue is the consolidation only. #1410 keeps its "risk 2" (the path-context
`IndexExpr`/`SliceExpr` carve-out gap), which genuinely needs a path-context-threading twin of
`eval_index_expr`/`eval_slice_expr` and stays Opus.

## The duplication

The "isolate a sub-expression with `optional` forced to `false`, then catch an escaping
`Error`/`Break` atomically, keeping any already-produced prefix" pattern has been independently
discovered and hand-implemented four times:

- `eval_try` / the plain evaluator's `Expr::Optional` dispatch (`src/jq/eval.rs`, pre-existing)
- the path-context evaluator's `Expr::Array` arm (#1302)
- the path-context evaluator's `Expr::Optional` arm (#1335)
- `eval_generic.rs`'s own `Expr::Array` arm — which relies on the same invariant *implicitly*,
  its atomicity comment being correct only because nothing currently seeds `optional=true` ahead
  of it

A unit test in `eval_generic.rs`
(`test_generic_plain_map_optional_on_non_container_is_unreachable_via_parser_725`) already
documents in prose the same reasoning #1302 and #1335 each rediscovered for `eval.rs`.

## The dormant risk this protects against

`eval.rs`'s path-context `Builtin::Map(f)` arm claims array-construction-style atomicity in its
own comment — matching `Array`'s pre-#1302 claim and `Optional`'s pre-#1335 claim — but passes
the *ambient* `optional` straight into `f` instead of forcing `false` and self-catching. It is
dormant only because nothing today forces `optional=true` ahead of a `map(...)` needing path
context. That is precisely the state `Array` and `Optional` were in before each needed fixing.

## Suggested fix

A shared helper —
`fn catch_optional<W>(result: QueryResult<'_, W>, catch_break: bool) -> QueryResult<'_, W>` —
that every `?`-shaped or atomic-construction arm routes through, in both `eval.rs` and
`eval_generic.rs`. This mirrors how `accumulate_path_context_step` and
`continue_rest_with_context` were themselves extracted after the same "N near-identical copies
silently diverged" pattern.

[...]

PR #1843, opened 2026-08-29
fix(jq): compound/alternative assignment fork/take-last a multi-output RHS

### LINKED ISSUE #1778: jq/yq: compound assignment operators collapse multi-output RHS to first value in both modes, disagreeing with both oracles
**Severity: Low** (surprising output, no data loss)

Found live during #1430's investigation (yq's plain \`=\` multi-output-RHS handling) — not folded into that fix, since it's a separate function and separate scope. See [docs/compliance/yq/limitations.md](docs/compliance/yq/limitations.md#s-multi-output-rhs-real-yq-takes-only-the-last-value-no-fan-out) for the full writeup this issue is extracted from.

## The gap

\`eval_rhs_once\` (\`src/jq/eval.rs\`), shared by \`eval_compound_assign\` (\`+=\`/\`-=\`/\`*=\`/\`/=\`/\`%=\`) and \`eval_alternative_assign\` (\`//=\`), always collapses a multi-output RHS to its **first** value — documented in its own doc comment as "a separate, pre-existing limitation" referencing #392, the issue that gave plain \`=\` its own correct multi-output handling (via \`eval_assign\`, a different function).

This disagrees with **both** oracles, in different ways:

**Real jq forks**, exactly like its own \`=\` does (confirmed live, jq 1.7.1):
\`\`\`bash
$ echo '{"x":1}' | jq -c '.x += (10,20,30)'
{"x":11}
{"x":21}
{"x":31}

$ echo '{"x":1}' | succinctly jq -c '.x += (10,20,30)'
{"x":11}
\`\`\`

**Real yq takes the last value** (not the first), matching the rule #1430 established for its own \`=\`:
\`\`\`bash
$ echo '{"x":1}' | yq -o=json -I0 '.x += (10,20,30)'
{"x":31}

$ echo '{"x":1}' | succinctly yq -o=json -I0 '.x += (10,20,30)'
{"x":11}
\`\`\`

\`|=\` itself is unaffected — real jq's own \`|=\` also collapses to first-only (confirmed live, matches succinctly today), so there's no divergence there.

## Scope note

This is two separate fixes bundled under one root cause (\`eval_rhs_once\`'s single-value collapse), not one:
- jq mode needs \`eval_compound_assign\`/\`eval_alternative_assign\` to fork over the RHS like \`eval_assign\` already does for \`=\` — likely requires restructuring these two functions away from \`eval_rhs_once\` toward something closer to \`eval_assign\`'s own fan-out loop, since \`eval_update\`'s underlying per-path filter application doesn't currently expect multiple RHS values.
- yq mode needs the opposite value picked (last, not first) with no fork, mirroring #1430's fix for \`=\`.

Both need per-operator live verification against the pinned oracles (\`-=\`/\`*=\`/\`/=\`/\`%=\` each, plus \`//=\`, which per existing code comments isn't real yq syntax at all — that operator's yq-mode answer would need to be judged by internal consistency with \`|=\` instead, same as the existing precedent for it).

newhoggy:
Fixes #1778.

`eval_rhs_once` (shared by `eval_compound_assign` `+=`/`-=`/`*=`/`/=`/`%=` and
`eval_alternative_assign` `//=`) always collapsed a multi-output RHS to its
first value, disagreeing with both oracles:

- **Real jq forks** over every RHS output, one whole result document per
  output, exactly like its own `=` already does (via `eval_assign`, #392).
  Confirmed live (jq 1.7.1): `{"x":1} | .x += (10,20,30)` produces three
  documents, `{"x":11}`/`{"x":21}`/`{"x":31}`.
- **Real yq takes only the last value**, no fork, matching the rule #1430
  established for its own `=` (confirmed live, yq v4.53.3, for `+=`/`-=`/`*=`;
  `/=`, `%=`, and `//=` have no real yq syntax at all, so their yq-mode
  answer is judged by internal consistency with `|=` instead, per the
  existing comments on these functions).

## Fix

Restructured `eval_compound_assign`/`eval_alternative_assign` to mirror
`eval_assign`'s own #392/#1430 shape:

- `collect_rhs_outputs`: keeps every RHS output instead of collapsing to the
  first — replaces `eval_rhs_once`, now fully removed (its only two callers
  were these).
- `eval_update_multi`: forks once per RHS output into its own whole-document
  copy via `update_path` (not `set_path` — these operators go through a
  filter, `. op value`/`. // value`, unlike `=`), yq-mode collapses to the
  last output only on clean completion (mirrors `eval_assign`'s own #1430
  gate), and follows the same #400 "keep the prefix" policy on a per-output
  write failure.

A zero-output RHS (`.a += empty`) now correctly produces zero output
(#1313's rule) rather than `eval_rhs_once`'s old "collapse empty to `Null`"
behavior.

## Incidental fix found via full verification

A pre-existing CLI test (`test_compound_assign_propagates_halt_after_partial_rhs_output`)
had its own oracle claim wrong from the start — re-verified live that
`{"a":0} | .a += (1, halt_error(3))` exits 3 and **does** print `{"a":1}` to
stdout (the RHS's already-computed `1` built a real document before the halt
fired), not empty output as the test claimed. The old `eval_rhs_once`-based
code discarded that document entirely on a trailing halt — a different
divergence from jq than the test's comment described, fixed as a side effect
of this rewrite. Test corrected to match the real, live-verified oracle
behavior.

## Testing

- 10 new regression tests covering jq-mode fork (bare, multi-path, and
  cursor-backed `Many`), yq-mode take-last (`+=`/`-=`/`*=`/`//=`),
  single-output-unaffected, empty-RHS, bare `Error`/`Break`/`Halt`, and
  error-after-partial-output.
- A direct unit test against `collect_rhs_outputs` for #1755's
  decode-failure raise (its predecessor `eval_rhs_once`'s own equivalent
  test, retargeted).
- Verified live: reverting `eval_update_multi`'s yq-mode collapse block
  reproduces the exact bug (both new yq-mode tests fail, showing every RHS
  output instead of just the last).
- Full local suite green — including a full run under CI's actual default
  (no explicit `--features`) invocation, not just the feature-flagged one
  this session had been using exclusively until now (see the `Test` job
  failure caught on the sibling #1755 misc-bucket PR, #1821).
- Patch coverage 94.87%+ after two coverage-gap-closing commits. Remaining
  gaps: one `unreachable!()` defensive arm (never reachable per
  `materialize_cursor`'s own contract), and a handful of lines in a narrow
  "inline `?` on a path component, not the whole assignment" edge case that
  mirrors `eval_assign`'s own pre-existing, unmodified code with the
  identical characteristic — confirmed via raw lcov `DA:` hit counts, not
  just assumed.

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -755,18 +755,33 @@ what succinctly should do instead). Not folded into this fix, since redefining a
 unverified error-interaction shape risked a second, differently-wrong divergence rather than
 fixing one — filed as [#1779](https://github.com/rust-works/succinctly/issues/1779).
 
-**Separately, not part of this fix:** live-probing this also surfaced that real jq's own
-`+=`/`-=`/`*=`/`/=`/`%=` (unlike `|=`) genuinely fork over a multi-output RHS the same way
-`=` does (`.x += (10,20,30)` on `{"x":1}` is three documents, `{"x":11}`/`{"x":21}`/
-`{"x":31}`) — succinctly's `eval_rhs_once` (shared by all of `eval_compound_assign` and
-`eval_alternative_assign`) instead always collapses to the *first* output, a pre-existing,
-already-documented gap in that function's own doc comment (referencing #392) that predates
-and is unrelated to #1430's yq-mode scope. Real yq's own answer for these operators under a
-multi-output RHS also isn't "first" — live-verified `.x += (10,20,30)` on `{"x":1}` is
-`{"x":31}` (last value, matching `=`'s own rule above), not `{"x":11}`. Both the jq-mode
-fork gap and the yq-mode wrong-value gap are real, separate from each other and from #1430,
-and neither is fixed here — filed as
-[#1778](https://github.com/rust-works/succinctly/issues/1778).
+### `+=`/`-=`/`*=`/`/=`/`%=`/`//=`'s multi-output RHS: same jq-forks/yq-takes-last split as `=`
+
+Live-probing #1430 also surfaced that real jq's own `+=`/`-=`/`*=`/`/=`/`%=` (unlike `|=`)
+genuinely fork over a multi-output RHS the same way `=` does, and real yq's own answer for
+these operators isn't "first" either — it takes only the **last** output, exactly like
+`=`'s own rule above:
+
+```bash
+$ echo '{"x":1}' | jq            -c '.x += (10,20,30)'          # {"x":11} {"x":21} {"x":31}
+$ echo '{"x":1}' | succinctly jq -c '.x += (10,20,30)'          # {"x":11} {"x":21} {"x":31}  (fixed)
+
+$ echo '{"x":1}' | yq            -o=json -I0 '.x += (10,20,30)' # {"x":31}
+$ echo '{"x":1}' | succinctly yq -o=json -I0 '.x += (10,20,30)' # {"x":31}  (fixed)
+```
+
+`succinctly`'s `eval_rhs_once` (shared by all of `eval_compound_assign` and
+`eval_alternative_assign`) used to always collapse to the *first* output in both modes — a
+pre-existing gap in that function's own doc comment (referencing #392), separate from and
+predating #1430's yq-mode scope for `=`. Fixed in
+[#1778](https://github.com/rust-works/succinctly/issues/1778) by replacing
+`eval_rhs_once` with `collect_rhs_outputs`/`eval_update_multi`, mirroring `eval_assign`'s
+own #392/#1430 shape: jq mode forks, yq mode collapses to the last output on a clean
+completion, gated on `S::TAG == EvalTag::Yq` — the same "deliberately still open" carve-out
+above (an RHS that itself errors partway through) applies here too, for the same reason
+(#1779). `/=`/`%=`/`//=` have no real yq syntax at all (confirmed live, `'/'`/`'//'` expect
+2 args but there is 1), so their yq-mode "take the last value" answer is judged by internal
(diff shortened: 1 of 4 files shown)


newhoggy:
Code review found and I fixed:

1. **A documentation regression**: `docs/compliance/yq/limitations.md`'s own #1778 entry still said "neither is fixed here" after the fix shipped. Updated to describe the shipped fork/take-last behavior.
2. **A mis-cited follow-up issue**: my own new `eval_update_multi` doc comment claimed the `eval_assign`/`collect_rhs_outputs` duplication was "#1826's own tracked scope" — verified against #1826's actual body/acceptance criteria and found it covers a completely unrelated pattern. Filed #1844 for the real duplication (a genuine ~130-line shared RHS-fork prologue+loop between `eval_assign` and the two new functions) and corrected the citation.
3. **Stale `eval_rhs_once` references** in test comments (renamed to `collect_rhs_outputs` where the function is gone but the described behavior is unchanged and still correctly tested).
4. **A guessed, unverified claim** in one of my own new tests' doc comments about what triggers `optional: true` at a specific branch — live-instrumented both this code and `eval_assign`'s identical pre-existing branch and found my guess was wrong (and, as it turns out, doesn't fully match what `eval_assign`'s own existing comment claims either). Softened to documented uncertainty rather than asserting something unverified.

No correctness bugs found across all 8 review angles — extensive live-oracle verification (jq 1.7.1 and yq v4.53.3) of fork order, last-value collapse, halt/partial interaction, scalar-noop interaction, and merge-flag interaction all checked out.

Full local verification re-run clean after each fix (build/clippy/no_std/full test suite under both default and full feature sets each time).

Issue #1832, opened 2026-08-29
jq: merge_owned's error-path prefix construction can still corrupt on decode failure


newhoggy:
**Severity: Low**

## Summary

While fixing #1755's misc-bucket slice (PR #1821), code review found and I fixed `eval_pipe`'s `Many`-branch promotion loop (`push_promoted`/`promote_borrowed_checked` in `src/jq/eval.rs`), which silently corrupted an undecodable element via unchecked `to_owned` the moment any sibling forced promotion to owned.

While fixing that, I noticed the shared `merge_owned` helper (`src/jq/eval.rs`, near line 1597) has the identical unchecked-`to_owned` shape, but used in a different context: constructing the **prefix** for a `partial(...)` result when a downstream control signal (`Error`/`Break`/`Halt`/`Partial`) terminates `eval_pipe`'s `Many`-branch loop early, e.g.:

```rust
QueryResult::Error(e) => {
    let prefix = owned.unwrap_or_else(|| merge_owned(borrowed, Vec::new()));
    return partial(prefix, Control::Error(e));
}
```

If `borrowed` (still-lazy cursors, never decode-checked because `owned` never got promoted) contains an undecodable string, `merge_owned`'s internal `borrowed.iter().map(to_owned).collect()` silently substitutes `""` for it instead of raising -- the exact #1746/#1755 bug shape, just reached via the error-reporting path rather than the main output path.

## Root cause

`merge_owned` (doc comment: "Shared by every stream-combining function that promotes a borrowed `Vec<StandardJson>` to owned the moment an owned value (or a terminal control) forces it") is a genuinely shared helper, not scoped to `eval_pipe` alone -- its own doc comment says so. Making it fallible ripples into every caller's own error-prefix construction, not just `eval_pipe`'s.

## Why this is separate from #1755's misc-bucket PR

`merge_owned` is shared infrastructure used by multiple functions, not a single-function fix like the rest of that PR's scope. Auditing every caller's error-path handling (do they already have a `Result`-returning shape to plumb this through, or does a `merge_owned` failure need its own partial-with-partial-prefix construction, mirroring the `promote_borrowed_checked` pattern that PR just established for `eval_pipe`'s own owned-value-forced promotion?) is a separate, focused piece of work.

## Suggested fix direction

[...]

PR #1852, opened 2026-08-29
fix(jq): eval_comma raises on decode failure in a heterogeneous comma

### LINKED ISSUE #1790: jq: eval_array_construction's fix has a gap — a comma mixing a navigable field with a literal still silently corrupts
Found while fixing #1755's sort/unique/group_by/min/max slice (PR for #1755) — a
narrower, separate gap in the same area, not covered by that fix.

## Summary

\`eval_array_construction\` (backs literal \`[...]\` array construction and every
\`min_by\`/\`max_by\`/\`sort_by\`/\`group_by\`/\`unique_by\` key computation via jq's own
\`[f]\` desugaring) was fixed in #1755's PR to raise \`EvalError::decode_failure\`
instead of silently substituting \`""\` for an undecodable string among its
constructed elements — confirmed correct for a homogeneous comma (\`[.a, .a]\`) and
for the sort-family's own single-key case (\`sort_by(.)\`, etc.).

But a comma mixing a *navigable* field access with a *non-navigable* literal still
silently corrupts, reachable only through the public \`succinctly::jq::eval\`
library API (same reachability as #1746/#1755's other findings — the CLI's
\`eval_generic.rs\` bridge validates the whole document upfront, so this is not a
live CLI regression; confirmed live that \`succinctly jq -c '[.a, 1]'\` on
\`{"a":"\ud800"}\` correctly raises via the CLI):

\`\`\`rust
// via the query! macro / eval() library API, bypassing eval_generic.rs
let json_bytes: &[u8] = b"{\"a\": \"\xff\xfe\"}";
// ...
eval::<Vec<u64>, JqSemantics>(&parse("[.a, 1]").unwrap(), cursor)
// => Owned(Array([String(""), NumberLiteral(Int(1), "1")]))   -- should raise

eval::<Vec<u64>, JqSemantics>(&parse("[.a]").unwrap(), cursor)
// => Error(decode_failure)   -- correct (fixed by #1755's PR)

eval::<Vec<u64>, JqSemantics>(&parse("[.a, .a]").unwrap(), cursor)
// => Error(decode_failure)   -- correct (fixed by #1755's PR)
```

## Root cause (partial — not fully traced)

\`eval_array_construction\`'s own fix converts its \`QueryResult::One\`/\`::Many\` arms
to use \`to_owned_checked\` instead of \`to_owned\`. The \`[.a, 1]\` shape apparently
produces a \`QueryResult::ManyOwned\` (or otherwise pre-converted) result by the time
it reaches this function's match — meaning the corrupted \`OwnedValue::String("")\`
was already produced *upstream*, somewhere inside \`eval_single\`'s own evaluation of
the \`Expr::Comma([.a, Literal(1)])\` inner expression, before
\`eval_array_construction\` ever gets a chance to check it.

This wasn't traced further — whoever picks this up should find the exact unguarded
\`to_owned\` call inside \`eval_single\`'s Comma-branch handling that produces the
already-corrupted \`ManyOwned\` values for a *heterogeneous* comma (mixing a
cursor-backed branch with an already-owned literal branch), as distinct from a
*homogeneous* comma (both branches navigable, which apparently routes through
\`eval_array_construction\`'s own \`QueryResult::Many\` arm correctly).

## Suggested fix

Trace \`eval_single\`'s \`Expr::Comma\` arm (or whatever produces \`ManyOwned\` for a
mixed cursor/literal comma) and find the unguarded \`to_owned\` call materializing
the cursor-backed branch's own output before it's merged into the \`ManyOwned\`
vector. Convert it to \`to_owned_checked\`, propagating the error.

## Related

- #1746 / PR #1754 — the original decode-failure-raising fix for this file's
  write-path family.
- #1755 — the read-only-builtins follow-up this was found while working on.

newhoggy:
Fixes #1790

## Root cause

`eval_comma`'s `QueryResult::One`/`::Many` arms promoted a cursor-backed
branch's value via unchecked `to_owned` the moment an already-owned sibling
(e.g. a literal) forced promotion to `ManyOwned` — silently substituting
`OwnedValue::String("")` for a string that fails UTF-8 decode, instead of
raising `EvalError::decode_failure`. This is the same bug shape #1755/#1843
already fixed for `eval_pipe`'s identical `Many`-branch promotion, just
reached through a different expression (`Expr::Comma` instead of `Expr::Pipe`).

A homogeneous comma (`[.a, .a]`) was already correct — both branches route
through `eval_array_construction`'s own `QueryResult::Many` arm, fixed by
#1755's PR. Only a heterogeneous comma (a navigable branch alongside an
already-owned one) hit this gap, and only via the public `succinctly::jq::eval`
library API — the CLI's `eval_generic.rs` bridge validates the whole document
upfront, so this was never a live CLI regression (confirmed live:
`succinctly jq -c '[.a, 1]'` on `{"a":"\ud800"}` already raised correctly).

## Fix

Rewrote `eval_comma`'s `One`/`Many`/`Owned`/`ManyOwned` arms to go through the
same `push_promoted`/`promote_borrowed_checked` helpers `eval_pipe` uses
(built in #1843) — reused directly, not re-duplicated a third time — so any
promotion failure raises `decode_failure` with the correct `#400` prefix
semantics (outputs already produced by earlier siblings in the comma are kept
via `QueryResult::Partial`, matching #1843's precedent).

## Verification

- Live-confirmed the repro from #1790 before and after the fix.
- New tests cover: heterogeneous comma with cursor-branch first (`[.a, 1]`,
  `.a, 1`) and last (`1, .a`), homogeneous valid-data regression check, and
  the `Many`/`ManyOwned` arms specifically (`1, .a[]`, `.a, (1,2)`).
- 100% patch coverage (46/46 new lines) via `omni-dev coverage diff`.
- Full test suite green under both `cargo test --verbose` (CI's default
  feature set) and `cargo test --features cli,simd,regex,serde` (this repo's
  full feature set), before and after rebasing onto `main`.
- `cargo build --no-default-features` clean (no_std compatibility).

## Test plan

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -1242,33 +1242,47 @@ fn eval_comma<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
 
     for expr in exprs {
         match eval_single::<W, S>(expr, value.clone(), optional).materialize_cursor() {
-            QueryResult::One(v) => match owned.as_mut() {
-                Some(acc) => acc.push(to_owned(&v)),
-                None => borrowed.push(v),
-            },
+            // #1790: to_owned_checked, not to_owned -- a heterogeneous
+            // comma (a navigable branch alongside an already-owned one,
+            // e.g. `[.a, 1]`) used to promote the borrowed branch's
+            // undecodable string to `""` here, silently, the moment the
+            // owned sibling forced the promotion. `push_promoted`/
+            // `promote_borrowed_checked` are the same helpers
+            // `eval_pipe`'s identical `Many`-branch promotion uses
+            // (#1755/#1843's own instance of this exact bug shape).
+            QueryResult::One(v) => {
+                if let Err(e) = push_promoted(core::iter::once(v), &mut borrowed, &mut owned) {
+                    let merged = owned.unwrap_or_else(|| merge_owned(borrowed, Vec::new()));
+                    return partial(merged, Control::Error(e));
+                }
+            }
             QueryResult::OneCursor(_) => {
                 unreachable!("materialize_cursor should have converted this")
             }
-            QueryResult::Many(vs) => match owned.as_mut() {
-                Some(acc) => acc.extend(vs.iter().map(to_owned)),
-                None => borrowed.extend(vs),
-            },
-            QueryResult::Owned(v) => owned
-                .get_or_insert_with(|| {
-                    core::mem::take(&mut borrowed)
-                        .iter()
-                        .map(to_owned)
-                        .collect()
-                })
-                .push(v),
-            QueryResult::ManyOwned(vs) => owned
-                .get_or_insert_with(|| {
-                    core::mem::take(&mut borrowed)
-                        .iter()
-                        .map(to_owned)
-                        .collect()
-                })
-                .extend(vs),
+            QueryResult::Many(vs) => {
+                if let Err(e) = push_promoted(vs, &mut borrowed, &mut owned) {
+                    let merged = owned.unwrap_or_else(|| merge_owned(borrowed, Vec::new()));
+                    return partial(merged, Control::Error(e));
+                }
+            }
+            QueryResult::Owned(v) => {
+                if owned.is_none() {
+                    match promote_borrowed_checked(core::mem::take(&mut borrowed)) {
+                        Ok(acc) => owned = Some(acc),
+                        Err((prefix, e)) => return partial(prefix, Control::Error(e)),
+                    }
+                }
+                owned.as_mut().unwrap().push(v);
+            }
+            QueryResult::ManyOwned(vs) => {
+                if owned.is_none() {
+                    match promote_borrowed_checked(core::mem::take(&mut borrowed)) {
+                        Ok(acc) => owned = Some(acc),
+                        Err((prefix, e)) => return partial(prefix, Control::Error(e)),
+                    }
+                }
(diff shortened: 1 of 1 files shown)


newhoggy:
Ran the mandatory multi-angle code review. Summary:

- **Confirmed correct**: the `One`/`Many`/`Owned`/`ManyOwned` promotion arms are sound, match `eval_pipe`'s established #400 prefix-preservation precedent, and `eval_generic.rs`'s Comma-handling paths were never exposed to this bug shape in the first place (no unfixed twin there).
- **Dead code found and fixed** (this branch, latest commit): `push_promoted` can only return `Err` via its `Some(acc)` arm, so the `owned.unwrap_or_else(|| merge_owned(...))` fallback at both new call sites was unreachable. Replaced with `.expect(...)` stating the invariant.
- **Out-of-scope gap found, not fixed here**: `eval_comma`'s own unchanged `Error`/`Break`/`Halt`/`Partial` arms still call the unchecked `merge_owned`, so a bad-borrowed-then-error/break/halt/partial sibling ordering (as opposed to this PR's bad-borrowed-then-owned promotion case) still corrupts. This is the exact gap issue #1832 already tracks generically across every `merge_owned` caller — commented there with a live repro (`.a, error("x")` on undecodable `.a`) rather than duplicating scope into this PR.

All local verification (build/clippy/tests under both default and full feature sets, no_std build, 100% patch coverage) re-run clean after the fix.

PR #1853, opened 2026-08-29
test(jq): invariant tests for the Index/IndexNumber and Slice/SliceNumber pairs

### LINKED ISSUE #1827: jq: add an invariant test that paired Expr variants (Index/IndexNumber, Slice/SliceNumber) cannot drift apart at their ~30 dispatch sites
**Severity: Low** (regression prevention; test-only, no production code change)

Split out of #1401 (option 4, which that issue itself frames as the "at minimum" step). This is
the one option there that needs no decision about `Expr`'s representation — it is test-only, and
it is what would have caught the bug #1401 documents happening twice.

#1401 stays open for options 1-3 (fold the key onto the existing variant / box it / replace
`unreachable!()` catch-alls), all of which are representation changes wanting a deliberate
decision.

## What this protects against

#1088 and #1326 each added a paired `Expr` variant (`IndexNumber` alongside `Index`,
`SliceNumber` alongside `Slice`) to preserve a float literal's own spelling in `path()` output.
Every one of the ~30 sites that asks "is this `Expr` a slice / does it carry an index key" must
handle both members of a pair, and nothing enforces that:

- `delete_expr_array_paths`'s error-construction match had **no `Expr::IndexNumber` arm at all**
  from #1088 until #1326 added one as a byproduct — a float-spelled index reaching it would have
  hit `unreachable!()`.
- Plain match exhaustiveness does not help where a match uses `_ => unreachable!(...)` instead of
  a real arm. `src/jq/eval.rs` currently has 89 `unreachable!` occurrences.

#1326 added `Expr::is_slice()`, `Expr::index_number_key()` and `Expr::slice_number_keys()`
(`src/jq/expr.rs`) for the sites that were pure boolean checks, but nothing yet asserts that new
call sites cannot drift apart again.

## What to build

An invariant test, per the `testing` skill's "Invariant Tests Over Duplicated Logic" guidance:
for every `Expr` shape in the paired families, assert that the plain variant and its
`*Number` sibling are classified identically by every shared predicate — `is_slice()`,
`index_number_key()`/`slice_number_keys()` presence, and whatever else the call sites branch on.

Then assert the property that actually matters end-to-end: for each pair, a filter using the
float-spelled form and the same filter using the plain form reach the same code path — e.g. that
`del`, `path()`, `=`, `|=` and `getpath` all behave equivalently for `.[1]` vs `.[1.0]` modulo
the preserved spelling. That is the assertion that would have caught the
`delete_expr_array_paths` gap.

## Acceptance

- A test that fails if a new `Expr` variant joins a paired family without being taught to every
  shared predicate.
- A test that fails if a plain/`*Number` pair diverges at any of the ~30 dispatch sites.
- No production code change. If writing the test surfaces a *third* live missing arm, file it
  separately rather than fixing it here.

## Related

- #1401 — the umbrella this is split from (options 1-3 remain).
- #1088, #1326 — the two paired variants, and the missing-arm bug found during the second.

newhoggy:
## Summary
- Test-only, no production code change. #1088 and #1326 each added a paired `Expr` variant (`IndexNumber` alongside `Index`, `SliceNumber` alongside `Slice`) to preserve a float literal's own spelling in `path()` output — every one of the ~30 sites that asks "is this a slice / does it carry an index key" must handle both members of a pair, and nothing enforced that. `delete_expr_array_paths`'s error-construction match had no `Expr::IndexNumber` arm at all from #1088 until #1326 added one as a byproduct.
- Added a predicate-consistency test in `expr.rs` pinning `is_slice()`/`index_number_key()`/`slice_number_keys()`'s own classification of each pair.
- Added end-to-end behavioral-equivalence tests running real filters (`.[1]` vs `.[1.0]`/`.[1e0]`, `.[1:3]` vs `.[1.0:3]`/`.[1:3.0]`) through `path()`, `del`, `getpath`, `=`, `|=`, asserting identical results modulo `path()`'s own spelling-preserving rendering.
- Added a dedicated multi-target `del()` case — local investigation confirmed single-target `del` takes a *different* fast path that never reaches `delete_expr_array_paths`; only `del(a, b)` (comma-separated) routes through it. Confirmed by local experimentation: temporarily reverted that function's or-pattern arms back to the pre-#1326 shape, re-ran the new tests, and only the multi-target `del` case failed with the historical `unreachable!()` panic — confirming this test actually catches the bug class it's meant to guard against, not just something that happens to pass either way.

## Test plan
- [x] Live-verified every filter's expected output against `/usr/bin/jq` 1.7.1
- [x] Confirmed the new multi-target `del()` test fails with the exact historical panic when the fix is reverted locally, and passes once restored (not committed — verified then reverted)
- [x] `cargo test --features cli,simd,regex,serde` — full suite green (3012+3 unit, 1270+5 integration)
- [x] `cargo fmt --check`

Fixes #1827

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -57943,6 +57943,35 @@ mod tests {
         assert_eq!(err.message, "Cannot index number with number");
     }
 
+    /// #1827: the sibling of `test_delete_expr_array_paths_reports_index_number_like_plain_index_1326`
+    /// for the `Slice`/`SliceNumber` pair -- the same match's two `Slice { .. }
+    /// | Expr::SliceNumber { .. }` arms (one `if`-gated on a `String` value, one
+    /// not) had no direct test coverage at all before this, for either sibling.
+    /// Same reachability caveat as #1326's own test: called directly, since a
+    /// parsed `del(.[1.0:3], .[3])`-shaped query exits through a different,
+    /// higher-level dispatch before ever reaching this specific match (verified
+    /// live: it reports a plain "Cannot index X with number", not either message
+    /// asserted here).
+    #[test]
+    fn test_delete_expr_array_paths_reports_slice_number_like_plain_slice_1827() {
+        let steps = [DeleteStep {
+            component: Expr::SliceNumber {
+                start: Some(1),
+                end: Some(3),
+                start_key: Some(NumberKey::Literal(1.0, "1.0".into())),
+                end_key: None,
+            },
+            optional: false,
+        }];
+        let err = delete_expr_array_paths(OwnedValue::String("hi".into()), &[&steps], 0)
+            .expect_err("not an array");
+        assert_eq!(err.message, "Cannot delete fields from string");
+
+        let err = delete_expr_array_paths(OwnedValue::Bool(true), &[&steps], 0)
+            .expect_err("not an array");
+        assert_eq!(err.message, "Cannot index boolean with object");
+    }
+
     // #694: `collect_owned()`/`eval_owned_multi` silently dropped a
     // `Partial`'s trailing `Control::Error`/`Control::Break`, keeping only
     // its prefix as if evaluation had completed normally -- so a mid-stream
--- a/src/jq/expr.rs
+++ b/src/jq/expr.rs
@@ -1629,6 +1629,47 @@ mod tests {
         );
     }
 
+    /// #1827: pins the three shared predicates' own classification of each
+    /// paired family's two members, so a future edit to `is_slice()`/
+    /// `index_number_key()`/`slice_number_keys()` cannot silently start
+    /// treating a plain/`*Number` pair differently. This alone does not
+    /// catch a *third* future paired variant, or a dispatch site elsewhere
+    /// that skips these predicates and pattern-matches the pair by hand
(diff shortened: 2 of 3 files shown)


newhoggy:
Code review found the multi-target \`del()\` test's doc comment misattributed which match it protects: it exercises \`delete_expr_array_paths\`'s \`ArrayStep\`-classification match (always correct, never missing an \`IndexNumber\`/\`SliceNumber\` arm), not the *error-construction* match that #1088/#1326's history is actually about. Verified live that neither match is reachable via a real parsed \`del()\` filter on a non-array value — the existing #1326 unit test's own doc comment already says so ("Called directly rather than through a jq query... exit through a different, higher-level dispatch"), confirmed independently by reverting each match's or-pattern arms in turn and observing which test (if either) panics.

Fixed: corrected the misleading doc comment, and added the direct-call sibling of #1326's own existing test for the `Slice`/`SliceNumber` pair (`test_delete_expr_array_paths_reports_slice_number_like_plain_slice_1827`), which had zero coverage before this at either match arm. Both new tests confirmed to fail with the exact historical `unreachable!()` panic when their respective match's or-pattern arms are reverted (verified locally, then restored — never committed).

New commit: 913084364. Full suite/clippy/fmt/doc/no_std re-verified green.

Issue #1443, opened 2026-08-22
yq mode: regex builtins should coerce a non-string pattern argument like real yq does


newhoggy:
Real yq silently coerces a non-string regex-pattern argument (e.g. a number) to its string representation and matches literally, but succinctly errors with `is_not_a_string` in yq mode. Live-verified against yq v4.53.3:

```bash
$ echo '"a1c"' | yq 'test(1)'
true
$ echo '"a1c"' | succinctly yq -o=json 'test(1)'
Error: number not a string or array

$ echo '"a1c"' | yq 'sub(1;"X")'
aXc
$ echo '"a1c"' | succinctly yq -o=json 'sub(1;"X")'
Error: number (1) is not a string
```

This is shared plumbing: `eval_regex_pattern_and_flags` (and the arity-3 `yq_sub_arity3_empty_replace` added in #1122, which evaluates and type-checks its own pattern argument independently) all reject a non-string pattern the same way, in both jq and yq mode. Per ADR-0018 rule 2 (mode decides behavior, not format) this should probably become a `yq`-mode-gated coercion — likely `to_string`-equivalent — applied to the pattern before it reaches regex compilation, matching real yq's own leniency, while jq mode keeps the current hard type error (real jq 1.7.1/1.8.2 also errors on a non-string pattern, unconfirmed but consistent with jq's general strict-typing philosophy elsewhere).

Needs live-verification across the full regex family (`test`, `match`, `capture`, `sub`, `gsub`, `scan`, `splits`) in yq mode before implementing, per ADR-0018 rule 1 (claims must be captured live from the pinned binary, not assumed to generalize from two probes).

Surfaced during code review of #1122's PR (#1442) — not in scope there since it's a broader, pre-existing gap across the whole regex builtin family, not specific to arity-3 `sub`.

newhoggy:
Claiming this — branch `issue-1443-yq-regex-non-string-pattern-coercion`.

newhoggy:
Merged via #1787.

[...]

Issue #1122, opened 2026-08-17
yq: real yq's 3-arg sub(re; s; flags) semantics don't fit any hypothesis tried so far


newhoggy:
## Summary

Split off from #1069 (fixed the confirmed part: yq's bare 2-arg `sub(re;s)` replaces every match, not just the first). The 3-arg `sub(re; s; flags)` form's real-yq behavior is a separate, still-unresolved mystery that doesn't fit jq's model, yq's bare-sub model, or a simple argument-order swap.

## Repro (real yq v4.53.3)

```
$ yq -n '"aaa" | sub("a";"X")' -o json          # 2-arg, baseline
"XXX"
$ yq -n '"aaa" | sub("a";"X";"g")' -o json       # 3-arg, matching input+pattern
""
$ yq -n '"aaa" | sub("a";"X";"")' -o json        # 3-arg, empty flags
""
$ yq -n '"aaa" | sub("a";"X";"z")' -o json       # 3-arg, junk flags
""
$ yq -n '"AAA" | sub("a";"X";"i")' -o json       # 3-arg, non-matching case (pattern lowercase, input uppercase)
"AAA"
$ yq -n '"aaa" | sub("a";"g";"X")' -o json       # hypothesis: arg order is (re; flags; replacement)?
""
$ yq -n '"AAA" | sub("a";"i";"X")' -o json       # same hypothesis, non-matching case
"AAA"
```

For comparison, real jq's 3-arg `sub`:
```
$ jq -n '"aaa" | sub("a";"X";"g")'
"XXX"
$ jq -n '"AAA" | sub("a";"X";"i")'
"XAA"
```

## Hypotheses tried and ruled out

- **Flags ignored, falls back to bare-sub-like behavior**: ruled out — if flags were just ignored, the matching-input+pattern case would produce something (either `"XXX"` global-style or `"Xaa"` first-match-style), not an empty string.
- **Argument order swapped** (`sub(re; flags; replacement)` instead of `sub(re; replacement; flags)`): ruled out — swapping produces the identical pattern of results (empty on match, unchanged on non-match) as the original order, which wouldn't happen if the swap actually changed what's being interpreted as the replacement text.
- **3-arg form silently broken/no-op regardless of input**: ruled out — a genuinely invalid regex (`sub("[";"X";"g")`) still produces a real parse error, so the regex and (in some form) the flags argument are being parsed and used, not ignored outright.

[...]

PR #1856, opened 2026-08-29
fix(yq): split(re; flags) ignores every argument once arity exceeds 1

### LINKED ISSUE #1439: yq: split() with arity >= 2 ignores all arguments and behaves as split("") — model established
**Severity: Low**

## Summary

Real yq's `split` **ignores every argument once its arity exceeds 1** and behaves exactly as
`split("")` — splitting the input string into individual Unicode characters. It is not a regex
split, and it is not jq's `split/2`. This is the same upstream Go bug shape already documented
for `sub(re; s; flags)` in `CLAUDE.md` (an argument read from a fixed AST slot that is empty at
higher arities), not a designed feature.

The model was previously unknown ("needs investigation"). It is now established by live probing
against the pinned oracle (Homebrew `yq` v4.53.3), so this issue is fully specified and ready to
implement.

## The model, live-verified against yq v4.53.3

| probe | yq v4.53.3 | reading |
|---|---|---|
| `"a1b2c" \| split("[0-9]";"g")` | `["a","1","b","2","c"]` | every char, digits kept |
| `"a1b2c" \| split("[0-9]";"x")` | `["a","1","b","2","c"]` | flags argument is irrelevant |
| `"a1b2c" \| split("[0-9]";"")` | `["a","1","b","2","c"]` | ...including empty flags |
| `"aXXbXXc" \| split("XX";"g")` | `["a","X","X","b","X","X","c"]` | pattern argument is irrelevant too |
| `"a,b,c" \| split(",";"g")` | `["a",",","b",",","c"]` | even a literal that *is* present |
| `"abc" \| split("Z";"g")` | `["a","b","c"]` | pattern absent — same result |
| `"" \| split("x";"y")` | `[]` | empty input |
| `"aé日b" \| split("x";"y")` | `["a","é","日","b"]` | **runes, not bytes** |
| `"abc" \| split("")` | `["a","b","c"]` | 1-arg `split("")` — identical |

### Neither argument is evaluated

```console
$ echo '"abc"' | yq -o=json -I0 'split(error("boom1");"y")'
["a","b","c"]                       # exit 0 — error() never fires
$ echo '"abc"' | yq -o=json -I0 'split("x";error("boom2"))'
["a","b","c"]                       # exit 0 — error() never fires
```

Control, proving the slot theory rather than "yq never evaluates split's arguments":

```console
$ echo '"abc"' | yq -o=json -I0 'split(error("boom1"))'
Error: boom1                        # exit 1 — at arity 1 the argument IS evaluated
```

### Arity 3 and 4 are accepted and discarded, and argument *types* are not checked

```console
$ echo '"abc"' | yq -o=json -I0 'split("x";"y";"z")'   # -> ["a","b","c"], exit 0
$ echo '"abc"' | yq -o=json -I0 'split(1;2)'           # -> ["a","b","c"], exit 0
```

Same parser leniency already recorded for `sub`'s 4th-and-beyond arguments (#1122).

### The *input* type check still runs

```console
$ echo '[1,2]' | yq -o=json -I0 'split("x";"y")'
Error: cannot split !!seq, can only split strings      # exit 1
```

## Current succinctly behaviour

`succinctly yq` applies jq's `split/2` regex model:

```console
$ echo '"a1b2c"' | succinctly yq -o=json -I0 'split("[0-9]";"g")'
["a","b","c"]        # want ["a","1","b","2","c"]
```

Everything else already lines up, which keeps the fix small:

```console
$ echo '"aé日b"' | succinctly yq -o=json -I0 'split("")'   # ["a","é","日","b"]  — matches
$ echo '""'      | succinctly yq -o=json -I0 'split("")'   # []                  — matches
```

`succinctly jq`'s `split/2` is correct and must not change (`"a1b2c" | split("[0-9]";"g")` →
`["a","b","c"]`, matching jq 1.7.1).

## Scope of the fix

1. In **yq mode only**, dispatch `split` with arity ≥ 2 to the existing `split("")`
   implementation, discarding all arguments without evaluating them.
2. Accept arity 3+ rather than raising an arity error, matching real yq's parser leniency
   (`sub`'s existing handling is the in-repo precedent).
3. `succinctly jq` unchanged.
4. Record the divergence in `docs/compliance/yq/limitations.md` alongside the existing `sub`
   entry — this reproduces an upstream bug deliberately, per ADR-0018 rule 3, exactly as `sub`
   already does.

## Out of scope (pre-existing, separate)

succinctly's non-string-input error wording differs from yq's and is unrelated to arity:

```console
$ echo '[1,2]' | succinctly yq -o=json -I0 'split("x";"y")'
Error: array ([1,2]) cannot be matched, as it is not a string
# yq: cannot split !!seq, can only split strings
```

Worth its own issue if anyone wants wording parity; do not fold it in here.

## Related

- #1122 — the same fixed-empty-AST-slot bug shape in yq's 3-arg `sub(re; s; flags)`.
- #1426 — the regex flag-grammar investigation that first surfaced this.

newhoggy:
## Summary

Real yq's `split(re; flags)` ignores every argument once arity exceeds 1 and always behaves as `split("")` — one element per Unicode character, not a regex split. Same upstream Go bug shape already documented for `sub(re; s; flags)` (#1122): the argument is read from a fixed AST slot that's empty at higher arities, not a designed feature.

Confirmed live against yq v4.53.3 across the issue's full probe table: a present-but-irrelevant literal pattern, `error(...)` never firing in either argument position (genuinely unevaluated, not just unused), arity 3+ accepted and discarded the same way `sub`'s arity 4+ already is, and the 1-arg `split(s)` form staying unaffected (its own argument really is evaluated).

## Implementation

- `builtin_split_regex` dispatches on `S::TAG == EvalTag::Yq` *before* either argument is evaluated — `fanout_regex_pattern_with_collected_flags` is never called in yq mode once arity reaches 2, matching "neither argument evaluated" exactly rather than "evaluated but ignored."
- Parser accepts (and discards, unevaluated) arity 3+ in yq mode only, mirroring `sub`'s existing precedent.
- `succinctly jq`'s own `split/2` is unchanged.
- `docs/compliance/yq/limitations.md`: resolved the existing "split — still open" stub, and updated the adjacent flag-grammar and #1443 pattern-coercion notes that pointed at this as an open question.

## Tests

Two pre-existing tests pinned the old (real-regex-split) output as intentional — updated now that #1439 resolves what they called an unresolved question:
- `test_yq_split_regex_pattern_coerces_1443` — was asserting only `code == 0`; now checks the real output too.
- `test_yq_split_regex_keeps_jq_style_zero_width_1255` → renamed `test_jq_split_regex_keeps_zero_width_1255`, converted to jq mode — yq mode no longer reaches the code this test exists to guard (`split_regex_resolved`/`global_captures`), so jq mode is now the only spelling that does.

Added 5 new tests: the full probe table, "neither argument evaluated," "arity 3+ accepted," a yq-mode zero-width-pattern-is-inert case, and a jq-mode negative control.

## Verification

Full `jq_cli_tests`/`yq_cli_tests` suites (1079 + 1276 tests), both CI clippy legs, `--no-default-features` (no_std) build, and `cargo doc --all-features` all green.

Fixes #1439

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -415,14 +415,31 @@ $ echo '"aaa"' | yq            'sub("A";"X";"i")'   # "aaa" (flags never read, n
 $ echo '"aaa"' | succinctly yq 'sub("A";"X";"i")'   # "aaa" (matches)
 ```
 
-`split`'s 2-arg `split(re; flags)` form has an unrelated, still-unresolved mystery of its
-own ([#1439](https://github.com/rust-works/succinctly/issues/1439)) — it doesn't do a regex
-split at all:
+### 2-argument `split(re; flags)` — resolved, real yq ignores every argument once arity exceeds 1
+
+`split`'s 2-arg `split(re; flags)` form had an unrelated mystery of its own, now resolved
+by [#1439](https://github.com/rust-works/succinctly/issues/1439): the same fixed-AST-slot
+shape as `sub` above, not a designed feature. Real yq's `split` ignores every argument once
+arity exceeds 1 and behaves exactly as `split("")` — splitting the input into individual
+Unicode characters, not a regex split, and not jq's `split/2`. Confirmed live: an
+`error(...)` placed in either the pattern or the flags argument never fires, and neither a
+present nor absent literal pattern changes the output at all:
 
 ```bash
 $ echo '"a1b2c"' | yq            -o=json 'split("[0-9]";"g")'   # ["a","1","b","2","c"]
-$ echo '"a1b2c"' | succinctly yq -o=json 'split("[0-9]";"g")'   # ["a","b","c"] -- jq-modeled regex split
+$ echo '"a1b2c"' | succinctly yq -o=json 'split("[0-9]";"g")'   # ["a","1","b","2","c"] (matches)
+$ echo '"a,b,c"' | yq            -o=json 'split(",";"g")'       # ["a",",","b",",","c"] -- pattern is present, still irrelevant
+$ echo '"a,b,c"' | succinctly yq -o=json 'split(",";"g")'       # ["a",",","b",",","c"] (matches)
 ```
+
+Per [ADR-0018](../../adrs/adr-0018.md) rule 3, succinctly reproduces this bug-for-bug rather
+than performing an actual regex split; arity 3+ is accepted and discarded (parser leniency),
+matching `sub`'s own arity 4+ handling above. The 1-arg `split(s)` form is unaffected — its
+own argument really is evaluated and used (`split(error("boom"))` still raises, confirmed
+live) — and `succinctly jq`'s `split/2` keeps its real jq-modeled regex-split behavior
+unchanged in jq mode. Non-string-input error wording (`array (...) cannot be matched, as it
+is not a string` vs. real yq's `cannot split !!seq, can only split strings`) is a separate,
+pre-existing gap unrelated to arity, not addressed here.
 
 ### Global regex zero-width-match iteration — resolved, real yq uses Go's `regexp`
 
@@ -551,7 +568,7 @@ which was true only before #1534.
   `inside` at all to verify it against — see the fan-out table above). Confirmed by
   `test_yq_contains_scalar_vs_scalar_kind_mismatch_answers_false_1649` and its siblings.
 
-### Regex flag grammar — `test`/`match`/`capture` fixed, `split` still open
+### Regex flag grammar — `test`/`match`/`capture` fixed
 
 **Fixed by [#1426](https://github.com/rust-works/succinctly/issues/1426):** real yq doesn't
 use jq's flag grammar at all for `test`/`match`/`capture` — only `g` is a real flag; every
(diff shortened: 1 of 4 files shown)


newhoggy:
## Code review (2 rounds)

**Round 1**: no correctness bugs. One cleanup finding — \`yq_split_ignores_arguments\` duplicated the exact char-splitting logic already in \`builtin_split\`'s empty-separator branch. Extracted \`split_into_individual_chars\` as the shared pure-splitting step (the two callers' surrounding error-handling for non-string input intentionally stays separate — different wording, per the issue's own scope note).

**Round 2** (3 finder angles: correctness line-by-line, removed-behavior audit, cleanup/reuse/altitude): no correctness bugs found — every "confirmed live against yq v4.53.3" claim was independently re-verified against the pinned oracle and matched exactly, no stray old-behavior assertions left behind, no double-evaluation risk in the parser's discard loop. One further cleanup finding: the yq-mode arity-leniency parsing loop (accept-and-discard any \`;expr\` past the arguments actually read) was copy-pasted byte-for-byte between \`sub\` and my new \`split\` block — the PR's own doc comment even called out the duplication without fixing it. Extracted \`parse_yq_arity_leniency_tail\` as the one shared implementation both builtins now call.

Full \`jq_cli_tests\`/\`yq_cli_tests\` suites (1079 + 1276), both clippy legs, \`--no-default-features\` (no_std), and \`cargo doc --all-features\` all green after each round.

PR #1858, opened 2026-08-29
refactor(yaml): drop write_deferred_prefix's always-None key_comment param

### LINKED ISSUE #1828: yaml: write_deferred_prefix's key_comment parameter is always None at 3 of its 4 call sites
**Severity: Low** (readability; no behaviour change)

Split out of #1448 (item 3). That issue bundles a benchmark question (item 1) with three
readability observations; this is the one with a concrete, decided fix and no measurement gate.

## The dead weight (verified live on current `main`)

`write_deferred_prefix` (`src/yaml/light.rs:6673`) takes a `key_comment: Option<&str>` parameter
that only **one of its four call sites** ever populates:

| call site | `key_comment` |
|---|---|
| `light.rs:1793` (`stream_yaml_value`'s Mapping-field branch) | `Some(...)` |
| `light.rs:1929` | `None` |
| `light.rs:6726` | `None` |
| `light.rs:7077` | `None` |

The function's own doc comment has to explain that passing `None` reproduces the pre-extraction
behaviour — a sign the parameter belongs at the one call site that needs it, not in the shared
signature.

## Suggested fix

Narrow the shared function to `write_anchor_tag(out, anchor, tag)` and leave the
comment-then-newline logic inline at `:1793`, the only caller that needs it. No behaviour change:
the three `None` callers currently take the `else if anchor.is_some() || tag.is_some()` branch
that `write_anchor_tag` would be.

## Acceptance

- `write_deferred_prefix`'s always-`None` parameter is gone.
- Output byte-identical before/after across the yq golden suite, specifically including
  anchor/tag-carrying documents in both block and flow style (#1444's own fixtures).
- `cargo test` green; `cargo clippy --all-targets --all-features` clean.

## Explicitly not in scope

Items 2 and 4 of #1448 (the duplicated dash/anchor/tag block, and `write_yaml_child_inline`
re-encoding the anchor/tag spacing rule) both need an API split or a non-trivial extraction
across differing indent representations, and #1448's own text says doing nothing there is a
legitimate outcome. Leave them alone.

## Related

- #1448 — the umbrella this is split from (items 1, 2 and 4 remain).
- #1444 — the PR whose review surfaced all four items.

newhoggy:
## Summary
- `write_deferred_prefix`'s `key_comment: Option<&str>` parameter was only ever populated at one of its four call sites (`stream_yaml_value`'s Mapping-field branch); the other three always passed `None`, and the function's own doc comment had to explain that `None` reproduces pre-extraction behavior — a sign the parameter belonged at the one call site that needs it, not in the shared signature.
- Narrowed to `write_anchor_tag(out, anchor, tag)` — just the shared "leading space, then `&anchor`, then tag" logic all four callers actually use.
- Inlined the comment-then-newline ordering (comment first, then anchor/tag on their own line at absolute column 0) at the one call site that needs it, preserving both of that site's own branches (comment present vs absent) exactly.
- Fixed 3 stale doc-comment references elsewhere in the file that named the old function.

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` — full suite green, including yq golden fixtures and #1444's own anchor/tag-carrying fixtures
- [x] Live spot-check: key-comment + container-anchor (column-0 placement), no-comment anchor, no-comment tag — all byte-identical against the pre-change binary
- [x] `cargo fmt --check`

Fixes #1828

### DIFF
--- a/src/yaml/light.rs
+++ b/src/yaml/light.rs
@@ -1803,18 +1803,36 @@ impl<'a, W: AsRef<[u64]>> YamlCursor<'a, W> {
                         let value = field.value_cursor();
                         if is_yaml_cursor_container(&value) && value.style() != "flow" {
                             // The key's own trailing comment (#765), then
-                            // the anchor/tag -- both via the same helper
-                            // #1077/#1113's scalar/absent branch uses, so
-                            // this branch stops hand-writing (and dropping)
-                            // the tag, and stops mis-placing the anchor
-                            // after the comment instead of on its own line
-                            // (#1132).
-                            write_deferred_prefix(
-                                out,
-                                field.key_cursor().line_comment_raw(),
-                                value.anchor(),
-                                value.explicit_tag(),
-                            )?;
+                            // the anchor/tag: with no comment, both go on
+                            // this line via the shared `write_anchor_tag`
+                            // helper #1077/#1113's scalar/absent branch
+                            // also uses; with one, it's written first, then
+                            // the anchor/tag move to their own line at
+                            // column 0 -- this branch stops hand-writing
+                            // (and dropping) the tag, and stops mis-placing
+                            // the anchor after the comment instead of on
+                            // its own line (#1132). Inline rather than
+                            // through the shared helper (#1828): this is
+                            // the only one of its four original callers
+                            // that ever had a comment to place.
+                            if let Some(comment) = field.key_cursor().line_comment_raw() {
+                                write_line_comment(out, Some(comment))?;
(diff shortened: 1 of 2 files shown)


newhoggy:
Caught: my earlier rename sweep only covered `src/yaml/light.rs`, missing 3 doc-comment references to the removed `write_deferred_prefix` in `tests/yq_cli_tests.rs`. Fixed in af0a86ceb — confirmed zero remaining references repo-wide, re-ran the affected test file (1271 passed).

Issue #1859, opened 2026-08-29
test(jq): tests/jq_cli_tests.rs has 34 more cargo-run orphan-process sites, not funneled through one helper


newhoggy:
## Summary

#1847 fixed the orphaned-`succinctly`-grandchild-process problem (`cargo run` spawning a second, uninstrumented binary that gets reparented to init and blocks forever on its stdout pipe once the outer process group is killed) in `cli_golden_tests.rs`, `cli_characterization_tests.rs`, `deep_nesting_valid_tests.rs`, and `json_validate_tests.rs` — each of those had exactly one helper function funneling every call site through a single `Command::new("cargo")` spawn, so the fix was a clean, single-function conversion to `Command::new(env!("CARGO_BIN_EXE_succinctly"))` per file.

`tests/jq_cli_tests.rs` is a different shape entirely and was deliberately left out of that PR's scope: it has **34 separate, independent `Command::new("cargo")` call sites**, not funneled through a small number of shared helpers. Verified by direct count against the file as it stands after #1847:

```
$ grep -c 'Command::new("cargo")' tests/jq_cli_tests.rs
34
```

Broken down:
- **5 shared helper functions** still spawn `cargo run` internally: `run_jq_stdin_streams`, `run_jq_file` (already `#[allow(dead_code)]` — check whether it's genuinely dead before touching it), `run_jq_null`, `jq_stderr`, `run_jq_binary_stdin`.
- **28 individual `test_*` functions** hand-roll their own `Command::new("cargo")...spawn()`/`.output()` call *inline*, with no shared helper at all — e.g. `test_unary_minus`, `test_unary_minus_expression`, `test_double_negation`, `test_slurp_with_raw_input_multiple_files`, `test_from_file`, `test_multiple_file_inputs`, `test_builtin_first_on_non_array_errors`, `test_contains_type_mismatch_errors`, `test_default_identity_filter`, `test_jq_help`, and 18 more (full list obtainable via `grep -B2 'Command::new("cargo")' tests/jq_cli_tests.rs` cross-referenced against the nearest preceding `fn` line).

[...]

PR #1860, opened 2026-08-29
test: spawn succinctly directly instead of via cargo run, stop orphans

### LINKED ISSUE #1847: test: tests that shell out via `cargo run` orphan `succinctly` grandchildren that hang at 0% CPU for days
**Severity: Medium**

## Summary

Tests that shell out via `cargo run` (rather than invoking the built binary
directly) leave **orphaned `succinctly` grandchild processes** that sit at 0.0%
CPU indefinitely. They accumulate across sessions and survive long after the
test run that spawned them is gone.

Observed live on the shared build machine while working #1615:

```console
$ ps -Ao pid,ppid,pcpu,etime,command | awk '$3=="0.0" && /target\/debug\/succinctly/'
 9564     1   0.0 02-01:22:13 target/debug/succinctly json generate 500 --pattern mixed --seed 42
19383     1   0.0 01-21:42:16 target/debug/succinctly json generate 1kb --seed 12345
43687     1   0.0    16:59:25 target/debug/succinctly json generate 500 --pattern strings --seed 42
84034 83720   0.0    04:49:30 target/debug/succinctly json generate 500 --pattern strings --escape-density 0.5 --seed 42
89418 89098   0.0 01-09:31:52 target/debug/succinctly json generate 500 --pattern literals --seed 42
```

Note `ppid 1` on three of them, and ages up to **2 days**.

## Root cause

`run_cli` in [tests/cli_golden_tests.rs](tests/cli_golden_tests.rs) runs the CLI
as `cargo run --bin succinctly -- <args>`:

```rust
let output = Command::new("cargo")
    .args(["run", "--features", cargo_run_features(), "--bin", "succinctly", "--"])
    .args(args)
    .output()?;
```

That inserts a `cargo` layer, so the real binary is a *grandchild* of the test
process. When the process group is killed — which
`~/.claude/skills/grinding-backlog/scripts/cargo-guard.sh` does by design on a
detected stall, and which #935 established happens routinely under concurrent
multi-session load — `cargo` dies but the grandchild is reparented to `init`
and blocks forever on its now-unreadable stdout pipe.

A sibling helper in the same file, `run_cli_bin`, already does the right thing:

```rust
Command::new(env!("CARGO_BIN_EXE_succinctly"))
```

`run_cli` has 22 call sites in that file versus 5 for `run_cli_bin`, and
`cargo`-spawning appears in at least 5 test files (`cli_golden_tests.rs`,
`cli_characterization_tests.rs`, `deep_nesting_valid_tests.rs`,
`jq_cli_tests.rs`, `json_validate_tests.rs`).

## Why this matters beyond tidiness

1. **It corrupts stall diagnosis.** These orphans are indistinguishable at a
   glance from a genuinely hung current run, so a session investigating a slow
   suite can conclude it is stalled when it is not (this happened while working
   #1615 — I diagnosed a live run as stalled on the strength of the orphan list).
2. **It degrades the shared machine** for every concurrent session, and nothing
   ever reaps them.
3. **#935 is closed as fixed, but its fix is the guard**, which *causes* this
   orphaning as a side effect of killing the group. The guard is treating the
   symptom one layer above the cause.
4. `cargo run` in tests is already known-harmful here for a second, unrelated
   reason: it builds an **uninstrumented** binary, making those tests invisible
   to `cargo llvm-cov` (the same reason `run_jq_full` exists).

## Suggested fix direction

Replace `Command::new("cargo").args(["run", ...])` with
`Command::new(env!("CARGO_BIN_EXE_succinctly"))` throughout the test suite —
`run_cli` can almost certainly become a thin wrapper over `run_cli_bin`'s
approach. That removes the grandchild layer entirely (no orphan is possible),
removes the retry-on-cargo-lock machinery `run_cli` currently needs
(`MAX_CARGO_RETRIES`, `classify_cargo_run_exit`), speeds the suite up by not
re-invoking cargo per test, and makes those tests visible to coverage.

Worth checking whether any call site genuinely *needs* `cargo run` (e.g. to pick
a non-default feature set that `CARGO_BIN_EXE_` cannot express) before
converting it; `cargo_run_features()` suggests at least one might.

## Why this is separate from #1615

#1615 was about decode-failure routing in the streaming writers. This was found
only because the stall guard fired during that work and the process list showed
days-old orphans from other sessions. Nothing about it is specific to #1615's
change.

newhoggy:
## Summary

Fixes the orphaned-`succinctly`-grandchild-process problem: `Command::new("cargo").args(["run", ..., "--bin", "succinctly", "--", ...])` inserts a cargo layer between the test process and the real binary. When the process group is killed (`cargo-guard.sh` does this by design on a detected stall, and #935 established it happens routinely under concurrent multi-session load), cargo dies but the `succinctly` grandchild is reparented to init and blocks forever on its now-unreadable stdout pipe. Observed live up to 2 days old, corrupting stall diagnosis for future sessions and degrading the shared build machine.

## Fix

Converted the 4 files where every call site funneled through exactly one helper function to spawn `Command::new(env!("CARGO_BIN_EXE_succinctly"))` directly instead:

- `cli_golden_tests.rs`'s `run_cli` — now a thin wrapper over `run_cli_bin`'s own already-correct approach (the sibling function the issue pointed at as the precedent).
- `cli_characterization_tests.rs`'s `run`.
- `deep_nesting_valid_tests.rs`'s `run`.
- `json_validate_tests.rs`'s `run_json_validate_via_cargo` — now matching the file's own `succinctly_bin()`/`exit_code_or_signal_death` convention its other helpers already use.

Each conversion also drops the now-unneeded cargo-lock-contention retry loop (`MAX_CARGO_RETRIES`/`classify_cargo_run_exit`): `CARGO_BIN_EXE_succinctly` is a compile-time path to this test binary's own already-built dependency, not a second cargo invocation that can hit exit-101 lock contention.

## Scope correction — split out #1859

The issue's own count ("22 call sites" for `run_cli`) undercounted `tests/jq_cli_tests.rs`'s real surface. That file has **34 separate `Command::new("cargo")` sites** — 5 shared helpers plus **28 individual inline call sites with no shared helper at all** — a materially different, larger, more heterogeneous shape than the clean single-helper-per-file pattern the other 4 files had. Filed as #1859 rather than folded in here.

## Verification

All 4 touched suites pass (75 tests total: 23 + 13 + 3 + 36), both CI clippy legs, `cargo fmt --check`, `--no-default-features` (no_std) build, and `cargo doc --all-features` all green. Total runtime for the 4 suites combined dropped from what would have been dozens of `cargo run` invocations to well under a second.

Fixes #1847

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,7 +148,7 @@ jobs:
 
       - name: Run cli-gated tests
         if: needs.gate.outputs.code == 'true'
-        run: cargo test --features cli --test yq_cli_tests --test yq_golden_tests --test yq_path_consistency_tests --test jq_golden_tests --test yaml_bench_suite_coverage --test dsv_cli_tests --test text_cli_tests --test json_validate_tests --test yaml_validate_tests
+        run: cargo test --features cli --test yq_cli_tests --test yq_golden_tests --test yq_path_consistency_tests --test jq_golden_tests --test yaml_bench_suite_coverage --test dsv_cli_tests --test text_cli_tests --test json_validate_tests --test yaml_validate_tests --test cli_golden_tests --test cli_characterization_tests --test deep_nesting_valid_tests
 
       # `tests/orchestrate_cli_tests.rs` is `#![cfg(feature = "bench-runner")]`
       # and was never previously built or run by any `cargo test` step here
@@ -255,7 +255,7 @@ jobs:
 
       - name: Run cli-gated tests
         if: needs.gate.outputs.code == 'true'
-        run: cargo test --features cli --test yq_cli_tests --test yq_golden_tests --test yq_path_consistency_tests --test jq_golden_tests --test yaml_bench_suite_coverage --test dsv_cli_tests --test text_cli_tests --test json_validate_tests --test yaml_validate_tests
+        run: cargo test --features cli --test yq_cli_tests --test yq_golden_tests --test yq_path_consistency_tests --test jq_golden_tests --test yaml_bench_suite_coverage --test dsv_cli_tests --test text_cli_tests --test json_validate_tests --test yaml_validate_tests --test cli_golden_tests --test cli_characterization_tests --test deep_nesting_valid_tests
 
       - name: Run bench-runner-gated tests
         if: needs.gate.outputs.code == 'true'
@@ -357,7 +357,7 @@ jobs:
 
       - name: Run cli-gated tests
         if: needs.gate.outputs.code == 'true'
-        run: cargo test --features cli --test yq_cli_tests --test yq_golden_tests --test yq_path_consistency_tests --test jq_golden_tests --test yaml_bench_suite_coverage --test dsv_cli_tests --test text_cli_tests --test json_validate_tests --test yaml_validate_tests
+        run: cargo test --features cli --test yq_cli_tests --test yq_golden_tests --test yq_path_consistency_tests --test jq_golden_tests --test yaml_bench_suite_coverage --test dsv_cli_tests --test text_cli_tests --test json_validate_tests --test yaml_validate_tests --test cli_golden_tests --test cli_characterization_tests --test deep_nesting_valid_tests
 
       - name: Run bench-runner-gated tests
         if: needs.gate.outputs.code == 'true'
@@ -423,7 +423,7 @@ jobs:
         run: cargo build --features cli --bin succinctly
 
       - name: Run cli-gated tests
-        run: cargo test --features cli --test yq_cli_tests --test yq_golden_tests --test yq_path_consistency_tests --test yaml_bench_suite_coverage --test dsv_cli_tests --test text_cli_tests --test json_validate_tests --test yaml_validate_tests
+        run: cargo test --features cli --test yq_cli_tests --test yq_golden_tests --test yq_path_consistency_tests --test yaml_bench_suite_coverage --test dsv_cli_tests --test text_cli_tests --test json_validate_tests --test yaml_validate_tests --test cli_golden_tests --test cli_characterization_tests --test deep_nesting_valid_tests
 
       - name: Run bench-runner-gated tests
         run: cargo test --features cli,bench-runner --test orchestrate_cli_tests
--- a/tests/cli_characterization_tests.rs
+++ b/tests/cli_characterization_tests.rs
@@ -20,56 +20,48 @@
 
 use std::io::Write;
 use std::process::{Command, Stdio};
-use std::time::Duration;
 
 use anyhow::Result;
 
 #[path = "common/cargo_run_exit.rs"]
 mod cargo_run_exit;
-use cargo_run_exit::{cargo_run_features, classify_cargo_run_exit, MAX_CARGO_RETRIES};
+use cargo_run_exit::exit_code_or_signal_death;
 
 /// Run `succinctly <args>` with `stdin` piped in, returning captured stdout.
 ///
-/// Goes through the real binary (via `cargo run`) rather than the library so the
-/// snapshot reflects exactly what a user sees. Bails with stderr on failure so a
-/// broken query is a loud test error, not a silently-empty snapshot.
+/// Goes through the real binary directly (`CARGO_BIN_EXE_succinctly`, a
+/// compile-time path to this test binary's own already-built dependency)
+/// rather than the library, so the snapshot reflects exactly what a user
+/// sees. Used to go through a second `cargo run` subprocess instead: that
+/// orphans the real `succinctly` grandchild -- reparented to init, blocked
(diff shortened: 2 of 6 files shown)


newhoggy:
## Round 2 review — two more real regressions found and fixed

**Regression 1: signal-death retry silently dropped.** Removing the cargo-run layer correctly eliminated the exit-101 lock-contention retry (genuinely gone — no cargo invocation left to contend on a lock), but `classify_cargo_run_exit` bundled a *second*, independent retry condition into the same function: retry on a signal-killed child (OOM kill, another session's \`pkill\`, or a stall-guard's process-group kill). That hazard doesn't go away just because the child is spawned directly — if anything a direct child is now *more* exposed to a targeted \`pkill succinctly\` than before. Fixed by adding \`spawn_with_signal_retry\` to \`cargo_run_exit.rs\` (mirroring that file's own established "one shared copy, not four independently drifting ones" precedent) and switching all four converted helpers to it, plus json_validate_tests.rs's two pre-existing helpers for internal consistency.

**Regression 2 (more severe): the 3 newly-gated files silently dropped out of CI.** Confirmed independently by 3 of 5 finder angles. \`ci.yml\`'s "Run cli-gated tests" step — the one step that actually passes \`--features cli\` with an explicit \`--test\` list, in all four \`test-x86\`/\`test-arm\`/\`test-macos-arm\`/\`test-x86-upstream\` jobs — never included these 3 files, because they didn't need \`--features cli\` before (self-building via their own inner \`cargo run\`, reachable from the plain \`cargo test --verbose\` step too). Round 1's \`#![cfg(feature = "cli")]\` fix correctly stopped the "No such file or directory" failure, but also correctly excludes these files from that now-ungated-for-them step — silently dropping 39 tests to 0-run-no-failure in every one of those 4 jobs, green with no signal anywhere. The only job that still happened to catch them was Coverage's \`--workspace\` sweep, which is skipped entirely on \`merge_group\` events — meaning a real regression in any of these 39 tests could merge through the queue with everything green. Fixed by adding all 3 files to the \`--test\` list in all 4 jobs.

Also added the missing \`bug\` label to #1859 (this issue's own sibling follow-up), so it doesn't fall off a bug-label filter once \`Fixes #1847\` auto-closes this one.

Full verification after both fixes: all 75 tests across the 4 files pass, plain \`cargo test\` (no \`--features cli\`) correctly compiles these files out (0 run, not a failure — confirmed on a clean target dir), both clippy legs, \`--no-default-features\` (no_std), \`cargo fmt --check\`, and \`cargo doc --all-features\` all green.

Running a third review round given this is the second round in a row to find a genuine, non-trivial regression on infrastructure shared across CI and 5 test files.

newhoggy:
## Round 3 review — three more gaps found in the shared helper itself, all fixed

All three findings were in \`spawn_with_signal_retry\` (the consolidation added in round 2), not the original conversion:

1. **stdin silently changed from null to inherited.** \`Command::output()\`'s documented default is \`Stdio::null()\` for stdin; \`Command::spawn()\` without an explicit \`.stdin(...)\` call inherits the parent's instead. My round-2 helper only called \`.stdin(Stdio::piped())\` when there was input to write, leaving stdin unconfigured (→ inherited) otherwise — confirmed live with a standalone repro (a spawned child with only stdout/stderr piped echoed back this shell's own stdin, where the old \`.output()\`-based code saw immediate EOF). A test that ever reads stdin under this contract could hang waiting on a live parent stdin — the exact class of hang this whole PR exists to eliminate, just relocated. Fixed: \`stdin\` is now explicitly set to \`Stdio::null()\` when there's no input to write.

2. **Retry backoff dropped.** Every pre-conversion call site slept \`100 * (attempt+1)\`ms between retries; the shared helper retried immediately. Restored.

3. **Six call sites each re-asserted the same invariant.** \`output.status.code().expect(...)\` was copy-pasted at every one of the 6 call sites instead of relying on the shared function's own guarantee. \`spawn_with_signal_retry\` now returns \`(Output, i32)\` directly, so the assertion happens exactly once, inside the shared function, via the already-imported \`exit_code_or_signal_death\`.

Also did a self-review pass afterward: the write-stdin-then-wait_with_output ordering (a known deadlock risk for pipe-buffer-sized-or-larger input) is unchanged from the pre-existing code in \`cli_characterization_tests.rs\`/\`deep_nesting_valid_tests.rs\` — not something this PR introduces, and safe given these files' own small, inline, stable test payloads (documented as such in \`cli_characterization_tests.rs\`'s header comment). Retry-count boundary condition traced through by hand: exactly \`MAX_CARGO_RETRIES\` (3) total attempts before giving up, matching the original loops' semantics.

Full verification: all 75 tests across the 4 files pass, both clippy legs, \`--no-default-features\`, \`cargo fmt --check\`, and \`cargo doc --all-features\` all green.

Issue #1857, opened 2026-08-29
yq: no-op predicate can't express 'skip RHS, still autovivify' for a mid-chain Iterate over Null


newhoggy:
**Severity: Low** (narrow trigger — a mid-chain `Iterate` whose immediate target is `Null`, RHS is side-effecting/erroring)

## Correction (2026-08-30)

This issue's original body claimed succinctly's write path doesn't autovivify `null` → `[]` for a mid-chain `Iterate` at all. **That claim was wrong** — found via code review on PR #1861 (three independent reviewer agents, plus my own re-verification), traced to a testing error on my part (a stale/misused binary invocation, not the actual behavior). Live-confirmed on both `main` and the PR branch:

```console
$ echo 'a: null' | ./target/debug/succinctly yq -o=json '.a[].b = 5'
{"a": []}
```

This matches real yq exactly — the autovivification already works correctly. Apologies for the noise; leaving this issue open under a corrected title/scope rather than closing outright, since a real (much narrower) gap remains underneath the original, wrong diagnosis.

## The actual gap

`yq_assign_noop_check`'s `Skip`/`Continue` split is all-or-nothing: `Skip(pristine)` means "return the document completely unchanged, evaluate nothing." For a `Null` target under a mid-chain `Iterate`, real yq's actual behavior is a *third* case this split can't express: the RHS is never needed (there are zero elements to write into, once autovivified), but the document still legitimately changes (`null` → `[]`) via the autovivify step alone.

Since `Skip` can only mean "unchanged," `assign_path_all_noop` (PR #1861) correctly excludes `Null` from its no-op predicate to avoid the worse bug (discarding a real, always-happening write) — but the RHS still evaluates eagerly as a result, which is itself a real, if narrower, divergence from real yq:

```console
$ echo 'a: null' | yq -o=json '.a[].b = error("boom")'
{"a": []}                                       # RHS never evaluates, autovivify still happens

$ echo 'a: null' | ./target/debug/succinctly yq -o=json '.a[].b = error("boom")'
Error: boom                                     # RHS evaluates eagerly instead
```

This also reaches through PR #1861's new recursion, not just a bare top-level `Null`:

```console
$ printf 'a:\n  - null\n  - [1]\n' | yq -o=json '.a[][].b = error("boom")'
{"a": [[], [1]]}                                 # RHS never evaluates

[...]

PR #1861, opened 2026-08-29
fix(yq): RHS-discard optimization now recognizes an all-no-op Iterate fan-out

### LINKED ISSUE #1432: yq: RHS-discard optimization for a mid-chain Iterate only covers a scalar target, not an all-no-op fan-out
**Severity: Low** (RHS-discard optimization only; the underlying write is already correct)

## Summary

#1298 fixed a mid-chain `Iterate`'s *write* to correctly no-op when the iterated value's own elements each individually no-op (`.a[].b = 99` on `a: [1, 2]` leaves `a: [1, 2]` unchanged, matching real yq). Code review during that PR then found `yq_assign_is_total_noop`'s eager-RHS-discard pre-check (`navigate_read_only`) didn't recognize any mid-chain `Iterate` case at all, so the RHS still evaluated for real even though the write correctly no-op'd — fixed for the narrowest case (`.a` itself being a genuine scalar).

Live-probing further during that same review turned up that real yq's actual RHS-discard rule is **broader** than "the iterated value is a scalar" — it's really "every element the `Iterate` fans into individually no-ops" (vacuously true for zero elements too). #1298's fix only covers the first of these; the others still eagerly evaluate (and can surface an error from) an RHS the real write was always going to discard.

## Repro (verified against yq v4.53.3)

```
$ echo 'a: [1, 2]' | yq -o=json '.a[].b = error("boom")'
{"a": [1, 2]}                                   # no-op, RHS never runs

$ echo 'a: [1, 2]' | succinctly yq -o=json '.a[].b = error("boom")'
Error: boom                                     # wrong: evaluates RHS for real

$ echo 'a: null' | yq -o=json '.a[].b = error("boom")'
{"a": []}                                       # write still autovivifies null -> [], but RHS never runs

$ echo 'a: null' | succinctly yq -o=json '.a[].b = error("boom")'
Error: boom                                     # wrong

$ echo 'a: []' | yq -o=json '.a[].b = error("boom")'
{"a": []}                                       # already empty -- vacuously a no-op, RHS never runs

$ echo 'a: []' | succinctly yq -o=json '.a[].b = error("boom")'
Error: boom                                     # wrong
```

A genuinely mixed target (at least one real write) correctly still evaluates the RHS, both before and after this scope note — not affected:

```
$ echo 'a: [1, {}]' | yq -o=json '.a[].b = error("boom")'
Error: boom
```

## Root cause

`navigate_read_only`'s new `Expr::Iterate` arm (`src/jq/eval.rs`, added by #1298) reports `PrefixNavOutcome::Absent` (defer to normal RHS evaluation) whenever the iterated value is a real `Array`/`Object`, regardless of what its elements are. Correctly recognizing the broader rule means checking every element/value against the *rest* of the resolved path (the terminal component beyond the `Iterate`, plus any further chain), not just this one step — effectively a read-only, non-mutating dry run of what `set_path_through_iterate` actually does at write time.

## Scope note

A real fix likely needs a new helper — call it a dry-run counterpart to `set_path_through_iterate` — that, given a container and the remaining path, returns whether *every* element/value would no-op the remainder (an `Array`/`Object`'s own `.iter().all(...)` over a recursive per-element check, vacuously `true` for an empty container). `navigate_read_only`'s `Iterate` arm would then call this instead of unconditionally returning `Absent` for a real container. This is naturally recursive (a chain of multiple `Iterate`s, or an `Iterate` followed by more path that itself contains another `Iterate`, needs to recurse through the same rule at each level) and needs its own scoping pass rather than a quick patch — not blocking #1298's own narrower, already-correct fix (the underlying *write* is unaffected either way; this is purely about whether the RHS gets evaluated at all when it happens to be side-effecting or erroring).

Filed during #1298's code review, correcting an earlier draft of the same finding that claimed "array of scalars" as a confirmed divergence without live-verifying it first — `a: [1,2,3]` does *not* actually diverge from real yq's write behavior (matches #1298's own fix), the divergence is specifically in the RHS-discard *optimization*, not the write.

newhoggy:
Fixes #1432

## Root cause

`yq_assign_is_total_noop`'s prefix walk (`navigate_read_only`) deferred to normal RHS evaluation for *any* real `Array`/`Object` hit by a mid-chain `Iterate`, even when every element/value would itself no-op the rest of the path — so `.a[].b = error("boom")` on `a: [1, 2]` (both elements scalar) incorrectly evaluated the RHS and raised, where real yq silently discards it (confirmed live against yq v4.53.3).

## Fix

New `assign_path_all_noop` recurses into `.iter().all(...)`/`.values().all(...)` at a mid-chain `Iterate` hitting a real container, instead of unconditionally deferring — a read-only, non-mutating dry run of `set_path_through_iterate`'s own per-element recursion, restricted to the `Field`/`Index`/`IndexNumber`/`Iterate` step vocabulary that's all this predicate's caller ever hands it (its own slice check already excludes anything else). Vacuously true for an empty container, and naturally handles nested `Iterate`s (`.a[][].b`) since the recursion applies at every level.

`PrefixNavOutcome::Resolved`'s payload became dead once its only reader (the terminal scalar check) moved into `assign_path_all_noop`'s own base case, so it's now a unit variant.

## Scope note: `Null` deliberately excluded

While verifying the fix against the oracle, found that real yq's `.a[].b = <RHS>` on `a: null` autovivifies `null` → `[]` as part of the write itself — but probing independently (a safe, always-evaluated RHS, isolated from this predicate entirely) showed succinctly's write path doesn't perform that autovivification at all today (`a: null` stays `a: null`, where real yq produces `a: []`). This is a separate, pre-existing write-correctness bug, not something this predicate can fix — its caller's `Skip(pristine)` short-circuit returns the *unmodified* original on a `true` result, which would make the missing autovivification permanent instead of merely already-wrong. Filed separately as #1857.

## Verification

- Live-verified every new case against yq v4.53.3 before implementing: empty array/object, all-scalar array/object, nested `Iterate`s, and a genuinely mixed (non-scalar-containing) fan-out that must still evaluate the RHS.
- New tests cover all of the above, plus the `Field`/`Index` mid-prefix mismatch arms (wrong container type, out-of-range index, scalar hit) needed to reach 96.77% patch coverage — the one remaining uncovered line is a defensive, provably-unreachable catch-all, documented and matching an identical pre-existing precedent elsewhere in this same predicate family.
- Full test suite green under both `cargo test --verbose` (default features, matches CI) and `cargo test --features cli,simd,regex,serde` (full feature set), before and after rebasing onto `main`.
- `cargo clippy --all-targets --all-features -- -D warnings` clean (both feature-set clippy legs).
- `cargo build --no-default-features` clean (no_std compatibility).

## Test plan

- [x] `omni-dev coverage diff` — 96.77% patch coverage (30/31; remaining line is documented defensive/unreachable)

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -732,17 +732,24 @@ pre-check, #1232's `PrefixNavOutcome`) an `Expr::Iterate` arm too — but only f
 narrowest case, `.a` itself being a genuine scalar (`.a[].b = error("boom")` on `a: 5` now
 no-ops silently, RHS never evaluated, matching real yq).
 
-**Still open ([#1432](https://github.com/rust-works/succinctly/issues/1432)):** real yq's
-actual RHS-discard rule for a mid-chain `Iterate` is broader — a real container whose own
-elements *all* individually no-op also discards the RHS, and so does an empty or
-null-autovivified container (vacuously). The underlying *write* already gets all of these
-right (verified against real yq); only the RHS-discard optimization is narrower than real
-yq's own rule:
+**Fixed by [#1432](https://github.com/rust-works/succinctly/issues/1432):** real yq's actual
+RHS-discard rule for a mid-chain `Iterate` is broader than #1298's own narrowest case — a real
+container whose own elements *all* individually no-op also discards the RHS, and so does an
+empty container (vacuously). New `assign_path_all_noop` recurses into
+`.iter().all(...)`/`.values().all(...)` at such an `Iterate` instead of unconditionally
+deferring, a read-only dry run of `set_path_through_iterate`'s own per-element recursion:
 
 ```bash
 $ printf 'a: [1, 2]\n' | yq            -o=json '.a[].b = error("boom")'   # {"a":[1,2]}, no-op
-$ printf 'a: [1, 2]\n' | succinctly yq -o=json '.a[].b = error("boom")'   # Error: boom
+$ printf 'a: [1, 2]\n' | succinctly yq -o=json '.a[].b = error("boom")'   # {"a":[1,2]} (fixed)
 ```
+
+A `null` target is deliberately excluded from this fix — real yq autovivifies `null` to `[]`
+as part of the write itself (already correct on both sides before and after this fix), so it
+is not a *total* no-op the way an empty/all-scalar container is, and this predicate's
+all-or-nothing `Skip`/`Continue` caller has no way to express "skip the RHS, but still
+perform the write." That narrower, still-open gap is tracked separately by
+[#1857](https://github.com/rust-works/succinctly/issues/1857).
 
 ### `=`'s multi-output RHS: real yq takes only the last value, no fan-out
 
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -15411,18 +15411,22 @@ fn yq_assign_is_total_noop(path: &Expr, root: &OwnedValue) -> bool {
 ///
 /// Deliberately excludes `Null`: real yq's `.a[].b = v` on `a: null`
 /// autovivifies `null` to `[]` as part of the write itself (confirmed live
-/// against a safe, non-erroring RHS: `yq '.a[].b = 5'` on `a: null`
-/// produces `a: []`), so it is not a *total* no-op the way an empty or
-/// all-scalar container is -- the document still changes. Checked
-/// independently of this fix (same safe-RHS probe against this codebase's
-/// unmodified write path) and found that autovivification currently
-/// doesn't happen there at all -- `a: null` stays `a: null` -- a
-/// pre-existing write-correctness bug this predicate can't fix (it only
-/// ever decides whether to skip evaluating the RHS, and the existing
-/// caller's `Skip(pristine)` short-circuit returns the *unmodified*
-/// original when this returns `true`, which would make the missing
-/// autovivification permanent instead of merely already-wrong). Filed
-/// separately as #1857 rather than folded into this predicate.
+/// -- and confirmed already correct on this codebase's own write path
+/// too, independent of this predicate: a safe, non-erroring RHS,
+/// `yq '.a[].b = 5'` on `a: null`, already produces `a: []` here exactly
+/// like the oracle). So `Null` is not a *total* no-op the way an empty or
+/// all-scalar container is -- the document still changes -- but the
+/// reason this predicate can't simply report it as one isn't a write bug:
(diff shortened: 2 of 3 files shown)


newhoggy:
Ran the mandatory multi-angle code review (5 finder agents). Summary:

- **Confirmed correct**: `assign_path_all_noop`'s recursion faithfully mirrors `set_path_through_iterate`'s own per-element walk (verified by hand-tracing and live oracle comparison across ~17 constructed cases), `navigate_read_only`'s only remaining caller (`yq_del_slice_outcome`) is unaffected by the payload removal, and no false-positive (data-loss direction) no-op was found anywhere.
- **Real defect found and fixed** (3 independent agents, confirmed by my own rebuild): my own doc comments and issue #1857 asserted a factually wrong claim — that succinctly's write path doesn't autovivify `null` → `[]` for a mid-chain `Iterate`. It does, already correctly, on both `main` and this branch (a testing error on my part, not a real bug). Corrected the doc comments in `eval.rs`/`yq_cli_tests.rs`, corrected #1857's title/body to the accurate narrower diagnosis (the `Skip`/`Continue` mechanism can't express "skip RHS, still autovivify," not "autovivification is broken"), updated `docs/compliance/yq/limitations.md`'s stale "Still open" entry to "Fixed", and added a regression test for the same gap reached through this PR's own new recursion (`a: [null, [1]]`).
- **Out-of-scope findings bundled into a new follow-up** (#1863): `navigate_read_only`/`PrefixNavOutcome`'s 3-way richness is now unused by its one remaining caller (a plain bool would do); a `let len = arr.len() as i64` bounds-normalization idiom is now duplicated 7x across the file; and three separate, deeper yq write-path divergences (Field-hits-array error-ordering, Index-out-of-range array-padding autovivification, Index-hits-object string-key coercion) surfaced while writing coverage tests for the `Field`/`Index` mid-prefix mismatch arms — all pre-existing, none introduced or worsened by this fix.
- **Efficiency**: the new fan-out is O(container-size) per `Iterate` level with short-circuiting on the first non-noop element — reviewed and accepted as a necessary, correctly-bounded cost (not superlinear, not benchmarked separately given this is a Low-severity narrow correctness fix, not a hot-path change).

All local verification (build/clippy/tests under both default and full feature sets, no_std build, 96.77% patch coverage) re-run clean after the corrections. One `--features cli,simd,regex,serde` run hit an unrelated, pre-existing flaky test (`test_walk_array_zero_output_child_drops_out_of_the_array`, no connection to this diff) — confirmed transient via isolated rerun and a clean full-suite rerun.

Issue #1829, opened 2026-08-29
jq: an undecodable object key is preserved by path()/tojson but silently dropped by keys/paths/to_entries — two policies, one document


newhoggy:
**Severity: Low-Medium** (self-inconsistent answers about the same document; library-API only —
the CLI validates upstream)

Split out of #1728 after re-probing it live. That issue's own two repros no longer describe
current `main` (one is fixed, the other's "should raise" premise was overtaken by the
preserve-don't-raise policy #1642/#1734 landed), and its ~250-call-site framing is superseded by
#1755/#1820. What is genuinely still open is narrower and more specific: **an undecodable or
structurally malformed object key is handled by two mutually incompatible policies inside one
document, depending only on which builtin you ask.**

### Axis 1 — decode-failure key: `{"\ud800":1,"b":2}`

| filter | result | policy |
|---|---|---|
| `[path(.[])]` | `[["\ud800"],["b"]]` | **preserved** as raw source spelling |
| `tojson` | `"{\"\\ud800\":1,\"b\":2}"` | **preserved** |
| `length` | `2` | **counted** |
| `keys` | `["b"]` | **dropped** |
| `keys_unsorted` | `["b"]` | **dropped** |
| `[paths]` | `[["b"]]` | **dropped** |
| `to_entries` | one entry | **dropped** |

Nothing raises. `length` says the object has 2 keys; `keys` says it has 1; `path(.[])` yields a
path `["\ud800"]` that `keys` denies exists.

### Axis 2 — #1194 structurally malformed key: `{"a":1,"b"}`

| filter | result | policy |
|---|---|---|
| `[path(.[])]` | raises (non-decode error) | **raises** |
| `tojson` | raises | **raises** |
| `keys` / `keys_unsorted` | `["a"]` | **dropped** |
| `length` | `1` | **dropped** |
| `[paths]` | `[["a"]]` | **dropped** |
| `to_entries` | one entry | **dropped** |

### Control — decode-failure *value*: `{"a":"\ud800","b":2}`

`path(.[])`, `tojson` and `[..]` all raise correctly. #1746/#1755's value-axis work is doing its
job; this issue is purely about the **key** axis.

## Why the two policies exist

`to_owned_checked_at_depth`'s Object arm (`src/jq/eval.rs`) routes keys through
`resolve_display_key`, which deliberately **preserves** a decode-failure key via its
lossily-decoded raw source span (#1642's `DisplayKeyGuard`, #1734) and **raises** on a
[...]

PR #1862, opened 2026-08-29
fix(jq): paths/leaf_paths/pick/omit silently drop malformed object keys

newhoggy:
Part of #1829 -- the `paths`/`leaf_paths`/`pick`/`omit` slice, following #1835 (`keys`/`keys_unsorted`), #1848 (`to_entries`), and #1854 (`map_values`).

## The real bug (found during code review, not the first commit)

The first commit fixed `eval.rs`'s `StandardJson`-specific `builtin_paths`/`builtin_leaf_paths`/`each_paths_filter`/`builtin_pick`/`builtin_omit` -- but `succinctly jq`/`succinctly yq` dispatch bare `paths`/`leaf_paths` through **`eval_generic.rs`'s own native `Builtin::Paths`/`Builtin::LeafPaths` arm** (`collect_paths_generic`), never through `eval.rs`. `eval.rs`'s copy is reachable only via the public library entry point `succinctly::jq::eval`, not the shipped CLI.

Confirmed empirically against **built binaries**, not just unit tests: unmodified `main`'s CLI still silently drops a structurally malformed key's path even after the first commit's `eval.rs` fix --

```
$ printf '{"a":1,"b"}' | succinctly jq 'paths'   # first-commit binary
["a"]                                             # exit 0, b's path silently dropped
$ printf '{"a":1,"b"}' | succinctly jq 'keys'     # same binary, same document
jq: error (at <stdin>:0): Invalid JSON text: expected ':', found '}'   # exit 5
```

`paths`/`keys` disagreeing on the identical document was exactly the bug this PR claims to fix -- the first commit's own new unit tests passed while the shipped binary was unchanged, because those tests call `succinctly::jq::eval` directly, bypassing the CLI's real dispatch entirely.

**Fix:** `collect_paths_generic` (`eval_generic.rs`) now routes through `effective_fields_checked` instead of the infallible `effective_fields`, giving it the error channel its own doc comment used to admit it lacked. Verified against a clean release build of the actual CLI binary:

```
$ printf '{"a":1,"b"}' | succinctly jq 'paths'   # this PR's binary
jq: error (at <stdin>:0): Invalid JSON text: expected ':', found '}'   # exit 5, now agrees with keys
```

`pick`/`omit`'s `eval.rs` fix, by contrast, has **no CLI-facing effect in practice** -- `Builtin::Pick`/`Builtin::Omit` have no native `eval_generic.rs` arm, so the CLI always falls through to the shared reindex bridge, which already handled #1194/#1642 correctly before this PR (confirmed live on unmodified `main`). The fix is still correct and matters for the direct library API, but is defense-in-depth rather than a CLI bug fix for those two specifically. `paths(filter)` (`each_paths_filter`) is in the same boat -- already bridge-correct on `main`, fixed anyway for library-API consistency.

## What changed

- **`collect_paths_generic`** (`eval_generic.rs`, the CLI-reachable fix): `effective_fields_checked`, not `effective_fields`; both `Builtin::Paths`/`Builtin::LeafPaths` call sites updated to propagate the new `Result`.
- **`builtin_paths`, `each_paths_filter` (`paths(filter)`), `builtin_leaf_paths`** (`eval.rs`, library-API fix): `to_owned_checked`, not `to_owned`, mirroring `builtin_path`'s established pattern.
- **`builtin_pick`/`builtin_omit`'s Object arms** (`eval.rs`, library-API fix): routed through `effective_fields_checked` + `key_display_string`/`key_display_string_kind`, matching `map_values`'s established pattern, instead of a raw per-field scan that silently skipped malformed/undecodable keys. `omit` also gained a `DisplayKeyGuard` collision check since it builds a real `IndexMap`.

## Also from this round of review

- **Performance**: `builtin_pick`'s Object arm re-decoded every field's display key once *per requested key* (O(keys × fields)) instead of once total. Now resolved into a `Vec` up front (O(fields) decode + O(keys × fields) cheap string compare).
- **Behavior pinned, not changed**: `pick` on a document with a literal duplicate key now finds the *last* occurrence in jq mode, matching real jq's own document-level duplicate-key collapse (`{"a":1,"a":2} | .` is `{"a":2}`, confirmed live against jq 1.7.1) -- the old raw scan found the first occurrence instead. This was a real, oracle-verified correctness improvement the first commit's description didn't call out; now has a dedicated regression test. `omit` was already last-value-wins either way.
- **Coverage gap closed**: `pick`/`omit(...)?` lacked a dedicated test pinning that the outer `Expr::Optional` catch suppresses the new malformed-key errors, matching `keys?`/`to_entries?`/`map_values(...)?`'s established convention.
- **New CLI-level integration test** (`test_paths_and_leaf_paths_raise_on_structurally_malformed_key_1829`, `tests/jq_cli_tests.rs`) exercises the actual shipped binary via `run_jq_full`, not just the library entry point -- the exact gap that let the first commit's tests pass while the CLI stayed unfixed.

## Scope note

`with_entries` needed no separate fix -- it already delegates to `builtin_to_entries` and inherited that function's #1848 fix for free, with its own pinned regression test confirming it. Per #1829's own acceptance criteria, that leaves this slice closing out `paths`/`leaf_paths`/`pick`/`omit`, with #1829 itself still open for any further follow-up.

**Worth a follow-up audit**: this "library evaluator fixed, CLI's own native `eval_generic.rs` arm not" gap is specific to builtins with their own native `eval_generic.rs` implementation (`paths`/`leaf_paths` had one; `keys`/`to_entries`/`map_values` apparently didn't -- confirmed those three already behaved correctly via the CLI on `main` before this PR, so this isn't systemic across the whole #1829 family, but it's worth deliberately checking CLI-reachability for any future slice rather than assuming a `eval.rs` unit test alone proves the fix ships).

## Testing

New regression tests (all `_1829`-suffixed):
- `src/jq/eval.rs`: `test_builtin_pick_finds_undecodable_key_via_fallback_spelling_1829`, `test_builtin_omit_preserves_undecodable_key_1829`, `test_builtin_pick_omit_raise_on_structurally_malformed_key_1829`, `test_builtin_pick_omit_raise_on_malformed_delimiter_1829`, `test_builtin_omit_raises_on_colliding_fallback_keys_1829`, `test_builtin_leaf_paths_preserves_undecodable_key_1829`, `test_builtin_paths_preserves_undecodable_key_1829`, `test_builtin_paths_filter_raises_on_structurally_malformed_key_1829`, `test_builtin_pick_omit_optional_suppresses_malformed_key_errors_1829`, `test_builtin_pick_finds_last_value_for_duplicate_key_1829`.
- `tests/jq_cli_tests.rs`: `test_paths_and_leaf_paths_raise_on_structurally_malformed_key_1829` (CLI-level, via `run_jq_full`).

Full local suite green (3039 lib tests, 1079 jq CLI tests, 1276 yq CLI tests, yq golden conformance, cli_golden_tests, doc-tests -- 0 failures), run via `cargo-guard.sh` under concurrent multi-session load. `cargo clippy --all-targets --all-features -- -D warnings` and the CI's second (`broadword-yaml`) clippy leg both clean. `cargo fmt --check`, `cargo build --no-default-features` (no_std), `cargo doc --no-deps --all-features`, and a plain default-features `cargo test` all clean. Live-verified against built release binaries throughout, not just `cargo test`.

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -846,19 +846,30 @@ coverage differs:
 
 A library caller who follows the documented `eval()` example and evaluates straight off a
 fresh cursor gets #1677 protection only from `keys`/`keys_unsorted`/`to_entries`; every other
-builtin in this file, checked or not on the decode-failure/#1194 axis, still misses it. The CLI
-itself never reaches this path except through the reindex bridge, which only ever feeds it an
-already-validated `OwnedValue`, so `sjq`/`syq` are unaffected. The remaining #1677 call sites
-(`with_entries`'s own post-`to_entries` reassembly step, `map_values`, `paths`, `leaf_paths`, and
-the `pick`/`omit` pair review surfaced alongside this issue) are tracked as a follow-up (#1829)
-rather than folded into #1677 itself, which scoped itself to `eval_generic.rs`.
+builtin in this file, checked or not on the decode-failure/#1194 axis, still misses it.
 
-Only `with_entries` routes through `to_entries` internally (it calls `builtin_to_entries`
-directly), so it inherits this fix's #1194/#1642/#1677 coverage for free on the decode step --
-though it still needs its own audit for whatever it does with the *reconstructed* object
-afterward before this issue can consider it closed. `map_values` is a separate, independent
-hand-rolled walk with the identical silent-`continue` shape `keys`/`to_entries` had before their
-own fixes -- unrelated to `to_entries` and not covered by this change at all.
+**#1829 closed the remaining gap** (`map_values` via #1835/#1848/#1854; `with_entries`
+confirmed to inherit `to_entries`'s fix for free, needing no separate change; `paths`,
+`leaf_paths`, and the `pick`/`omit` pair via #1862) -- but not uniformly through this file's own
+`to_owned`/`effective_fields` family. `map_values`/`pick`/`omit` did route through this file's
+own `effective_fields_checked`, matching the paragraph above. `paths`/`leaf_paths` did not: their
+*actual* fix landed in `eval_generic.rs`'s `collect_paths_generic`, the CLI's own native
+`Builtin::Paths`/`Builtin::LeafPaths` dispatch -- **the "CLI unaffected" claim this paragraph
+used to make here was wrong for those two.** `sjq`/`syq paths`/`leaf_paths` do *not* reach this
+file at all; they bypass the reindex bridge entirely via their own `eval_generic.rs` arm, which
+carried the identical #1194 silent-drop bug independently (confirmed live against a built
+release binary: `printf '{"a":1,"b"}' | succinctly jq paths` returned `["a"]` at exit 0 even
+after this file's own `builtin_paths`/`builtin_leaf_paths` were fixed, until `collect_paths_generic`
+was fixed too). `pick`/`omit` have no native `eval_generic.rs` arm, so the "CLI unaffected" claim
+does hold for them -- the reindex bridge they fall through to already fed them an
+already-validated `OwnedValue` before this file's own fix, making that fix real for the library
+API but largely redundant for `sjq`/`syq`.
+
+**The durable lesson**: a unit test against `succinctly::jq::eval` (this file's own entry point)
+proves a fix reaches library callers, never that it reaches the CLI -- only a build-and-run
+check against the actual binary proves that, and is worth doing for any future builtin that has
+(or might grow) its own native `eval_generic.rs` implementation rather than falling through to
+this file via the bridge.
 
 ## Refusing an allocation jq does not survive
 
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -34362,20 +34362,28 @@ fn builtin_pick<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
                 Ok(f) => f,
                 Err(e) => return QueryResult::Error(e),
             };
+            // Resolve each field's display key once, not once per
+            // requested key (code review, #1829): the raw scan the fix
+            // above replaced re-decoded every field's key on every outer
+            // iteration -- O(keys * fields) decodes where `checked`'s own
+            // keys never change across that loop. Unreachable in practice
+            // -- `effective_fields_checked` already rejected a
+            // structurally malformed key (#1194) -- checked anyway,
+            // matching `effective_keys`'s own defensive style rather than
+            // assuming.
+            let mut resolved_fields = vec_with_capacity(checked.len());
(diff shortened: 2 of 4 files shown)


newhoggy:
## Review summary

Two full `/code-review` rounds on this PR (8 finder agents each), plus manual verification against built binaries throughout.

**Round 1** found and fixed: an O(keys × fields) unnecessary key-redecode in `pick` (now O(fields) resolve), and pinned an undocumented-but-correct behavior improvement (`pick` finding the *last* occurrence of a duplicate key in jq mode, matching real jq's own document-level collapse).

**Round 2** found and fixed the real bug in this PR: the first commit only fixed `eval.rs`'s library-only copy of `paths`/`leaf_paths` -- the actual CLI dispatches through `eval_generic.rs`'s own native arm (`collect_paths_generic`), which independently carried the identical #1194 silent-drop bug, confirmed empirically against a built release binary (`printf '{"a":1,"b"}' | succinctly jq paths` still returned `["a"]` at exit 0 after the first commit). Fixed, and verified against the binary again after the fix. Round 2 also found -- independently, by three separate finders -- that `pick`'s object arm had no collision guard for two undecodable keys sharing the same #1642 fallback spelling, unlike `omit`/`map_values` on the identical document; fixed with a targeted equivalent of `DisplayKeyGuard` for `pick`'s lookup-based (not full-materialization) shape. Also fixed: a compliance-doc claim that was factually wrong for `paths`/`leaf_paths` (`docs/compliance/jq/limitations.md`), a stale doc comment, and a stray debug file a review subagent left in the shared worktree.

**Deliberately deferred**, matching this `#1829` family's own established scope-discipline (see #1854's own review):
- A double-decode pattern (`effective_fields_checked` decodes each key once internally to validate, then `pick`/`omit`/`paths`/`leaf_paths` decode it again via `key_display_string`) -- inherited from `map_values`/`to_entries`'s already-merged precedent, not introduced by this PR. A proper fix would need `effective_fields_checked` to hand back the `Cow<str>` it already resolved, rippling into every existing caller -- out of scope here.
- `resolve_display_key` isn't reused by `pick`/`omit`/`map_values`'s hand-rolled deferred-guard pattern (the check has to run *after* deciding whether an entry is actually kept, which `resolve_display_key`'s own eager-insert contract doesn't support). Three call sites now share this shape independently; a `resolve_display_key`-style helper for the deferred case would consolidate them, but that's a cross-cutting `document.rs` change affecting already-merged code, not a fix this PR's own scope covers.
- A pre-existing (not introduced by this PR) `pick`-vs-`omit` first-vs-last-occurrence asymmetry under yq mode's uncollapsed duplicate keys, untested either way.

Both rounds are clean of any remaining actionable findings. Ready to merge.

PR #1866, opened 2026-08-29
fix(jq): Builtin::Map's path-context arm forces optional=false, self-catches

newhoggy:
## Summary

This is a scoped-down implementation of #1826, which asked for a four-site "isolate-and-catch-optional" consolidation across `eval.rs` and `eval_generic.rs` behind one shared helper. Investigation before implementing found two things that changed the plan:

1. **The "one helper" premise doesn't hold.** There are genuinely two families here — a "keep-prefix" family (`eval_try`, the path-context `Expr::Optional` arm) and an "atomic construction, discard-prefix" family (the path-context `Expr::Array` arm, and this fix) — with different result shapes, plus `eval_generic.rs`'s `GenericResult` can't share a helper with `eval.rs`'s `QueryResult` at all.
2. **The correctness gap the issue called "dormant" is live-reachable today, not theoretical.** `eval.rs`'s path-context `Builtin::Map(f)` arm claimed the same array-construction atomicity `Expr::Array`'s arm earns via #1302's fix, but passed the *ambient* `optional` straight into each element's own evaluation instead of forcing `false` and self-catching. Confirmed live pre-fix: `(.a)? | map(if key==1 then error("boom") else key end)` on `{"a":[1,2,3]}` produced `[0,2]`, where the structurally-identical `(.a)? | [.[] | if key==1 then error("boom") else key end]` already correctly produced nothing — a real internal inconsistency between two constructs the codebase's own comments claim behave identically.

Given the real value here is closing that live bug, and the broader mechanical consolidation is now lower-priority (no live bug driving urgency, plus the "one helper" premise needs rethinking as at least two), this PR scopes to the concrete fix and I'm filing a follow-up for the broader refactor separately.

## The fix

Mirrors `Expr::Array`'s already-correct, already-tested arm (#1302) exactly: force `optional=false` into `iterate_element_step` for both the `Array` and `Object` loop bodies (so a per-element error surfaces as a real stop signal instead of self-swallowing into "no output for this element"), then gate only the `Error` arm of the post-loop match on this arm's own ambient `optional` — `Break`/`Halt` always propagate, matching `eval_try`'s "never catches `Break`/`Halt`" rule.

## Test plan
- [x] Live-verified the bug and the fix against multiple shapes (array target, object target, without ambient `?` still hard-errors, normal success case unaffected, `rest` after `map()` still runs)
- [x] Added `test_map_path_context_builtin_optional_is_atomic_1826`, mirroring #1302's own test style — confirmed it fails with the exact pre-fix value (`[0,2]`) when the fix is reverted locally, and passes once restored (verified then reverted, not committed)
- [x] `cargo test --features cli,simd,regex,serde` — full suite green (existing #1302/#1335/#725 tests pass unchanged)
- [x] `cargo fmt --check`

Split from #1826 (kept open for the broader consolidation, or closed with a follow-up filed — happy to do either per maintainer preference)

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -26424,7 +26424,7 @@ fn eval_pipe_with_path_context_internal<'a, W: Clone + AsRef<[u64]>, S: EvalSema
                             root,
                             file_origin,
                             current_path,
-                            optional,
+                            false,
                             &mut results,
                         ) {
                             return stop;
@@ -26440,7 +26440,7 @@ fn eval_pipe_with_path_context_internal<'a, W: Clone + AsRef<[u64]>, S: EvalSema
                             root,
                             file_origin,
                             current_path,
-                            optional,
+                            false,
                             &mut results,
                         ) {
                             return stop;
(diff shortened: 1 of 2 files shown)


newhoggy:
Code review found two real issues in the first version, both fixed in ceb6a72fb:

1. **Incomplete fix**: the post-loop `Error` match only caught `QueryResult::Partial(_, Error)`, but `partial()` (#400/#494) collapses an *empty* accumulated prefix to a bare `Error` — exactly the case when the erroring element is the **first** one. That bare `Error` fell through to the unconditional `Some(other) => other`, unguarded by the new `optional` gate, so the original bug survived whenever the failing element came first. Confirmed live: `(.a)? | map(if key==0 then error("boom") else key end)` still hard-errored despite the earlier "fix". Corrected to match both shapes in one arm, same as `Expr::Array`'s own arm already does. Extended the regression test with first-element/first-key cases.

2. **Accidental scope creep**: a stray `git diff` during local verification surfaced that an earlier `replace_all` edit (used while reverting/restoring the fix for local testing before committing) had also changed two unrelated call sites in `Expr::Iterate`'s arm from `optional` to `false` — undocumented, untested, and out of this PR's stated scope. Reverted to restore byte-identical pre-PR behavior there (verified live against unmodified `main`), leaving `Iterate`'s own handling for a separate, properly-scoped investigation.

Full suite/clippy/fmt/doc/no_std all re-verified green.

Issue #1865, opened 2026-08-29
yq: del() with an untracked Expr::Comma branch has order-dependent, reverse-then-abort semantics


newhoggy:
## Summary

Split out of #1764, which fixed the sibling case for \`path()\`/\`=\`/\`|=\`: a computed/untracked \`Expr::Comma\` branch (e.g. a bare literal like \`1\` mixed with real navigations like \`.a\`) is real yq's own silent no-op for those three operations, confirmed position-independent — the untracked branch simply contributes nothing while every other branch is written normally, regardless of order.

\`del()\` does **not** share that model. #1764 deliberately left it on the old (safe) behavior — raising \`reject_untracked_at_terminal\`'s existing error — rather than guessing at its real semantics, since applying the simple "skip and continue in given order" fix to \`del()\` would make it delete *more* than real yq does for some orderings (a data-loss-shaped divergence in the wrong direction).

## What's been verified so far (live against yq v4.53.3)

```bash
$ echo 'a: 1
b: 2' | yq 'del(.a, 1)'      # a: 1\nb: 2  -- nothing deleted
$ echo 'a: 1
b: 2' | yq 'del(1, .a)'      # b: 2       -- .a IS deleted (same two arguments, reversed)
$ echo 'a: 1
b: 2
c: 3' | yq 'del(.a, .c, 1)'  # a: 1\nb: 2\nc: 3 -- nothing deleted, even though .a/.c precede "1"
$ echo 'a: 1
b: 2
c: 3' | yq 'del(.a, 1, .c)'  # a: 1\nb: 2       -- only .c deleted, not .a
```

The pattern across these four is consistent with a hypothesis: **\`del()\` processes its targets in reverse of the given argument order, and aborts every remaining target — without undoing whatever already completed — the instant it hits an untracked one.** This hypothesis is not yet independently confirmed beyond the four cases above, and the exact algorithm (is it really "reverse of argument order," or something else that happens to produce the same results for these four inputs — e.g. reverse of *resolved document position*, or a sort by path depth?) needs more probing before implementing:

- Does the "abort" apply per-element (array indices) the same way it does per-object-key?
- What happens with an untracked branch nested inside a deeper pipe, not a top-level comma argument?
- Does \`del(..)\`/\`recurse\`-produced branches interact with this at all (the codebase's own \`short_circuit_del_root\` optimization already special-cases a root branch for \`del()\`; worth checking these don't collide)?
[...]

PR #1867, opened 2026-08-29
fix(yq): an untracked comma branch is a no-op for path()/=/|=, not an error

### LINKED ISSUE #1764: yq: an untracked comma branch raises jq's "Invalid path expression", where real yq silently no-ops
## Summary

Found while verifying #1560's fix against the pinned yq oracle. `resolve_dynamic_indexes`'s
`reject_untracked_at_terminal` — the check that raises jq's "Invalid path expression with
result <v>" for an `Expr::Comma` branch that resolves to a computed value rather than a real
navigation — runs unconditionally for both `S: JqSemantics` and `S: YqSemantics`. It matches
jq exactly. It does not match yq at all: real yq accepts the same shapes and silently no-ops
the untracked branch instead of raising anything.

This predates #1560/PR #1762 entirely — confirmed on unpatched `main` before that PR's own
change, so it is not something that PR introduced or worsened.

## Repro (confirmed live against Homebrew yq v4.53.3, and against succinctly built from
main before #1560's own PR merged)

```console
$ printf 'a: 1\nb: 2\n' | yq '(.a, 1) = 5'
a: 5
b: 2                                            # succeeds; the untracked `1` branch is a no-op

$ printf 'a: 1\nb: 2\n' | succinctly yq '(.a, 1) = 5'
Error: Invalid path expression with result 1    # raises jq's own wording, exit 1

$ printf 'a: 1\nb: 2\n' | yq 'del(.a, 1)'
a: 1
b: 2                                            # succeeds; no-op, nothing deleted for `1`

$ printf 'a: 1\nb: 2\n' | succinctly yq 'del(.a, 1)'
Error: Invalid path expression with result 1    # raises, exit 1
```

## Scope note

Per ADR-0018, `succinctly yq` should follow real yq's own behavior here, not jq's — this is
exactly the kind of divergence the rule exists to catch, not one of its narrow exceptions.
`reject_untracked_at_terminal` (and now, after #1560/PR #1762, its sibling
`reject_untracked_prefix_too`) would need a mode-dispatched carve-out — likely an
`EvalSemantics` const or similar deciding whether an untracked terminal branch is an error
(jq) or a silent no-op (yq) — rather than the current unconditional raise. Not investigated
beyond confirming the repro and locating the check; the exact shape of yq's "silently no-op"
rule (does it no-op the whole comma sibling, or something narrower?) needs its own
verification pass across more shapes before implementing.

## Related

- #1560 / PR #1762 — extended this same check to a second call site (an `Err`-prefix case)
  for jq-mode correctness; did not touch or worsen this yq-mode gap, which already existed
  identically on the pre-existing `Ok`-path call site.

newhoggy:
## Summary

\`reject_untracked_at_terminal\` (the shared check answering "is a \`path()\`/\`del()\`/\`=\`/\`|=\` branch that resolved to a computed value, rather than a real navigation, actually an error") raised jq's "Invalid path expression" unconditionally in both modes. Real yq's answer for \`path()\`/\`=\`/\`|=\` is different: the untracked branch is a silent no-op, and every other branch is still written normally — confirmed position-independent against yq v4.53.3 (untracked first, middle, or last; all-branches-untracked; a computed value from an arbitrary expression, not just a bare literal).

```bash
$ yq            '(.a, 1) = 5'   # a: 5, b: 2 -- untracked "1" contributes nothing
$ succinctly yq '(.a, 1) = 5'   # matches now; was: Error: Invalid path expression with result 1
```

A genuine error/break/halt inside a comma branch still propagates normally — only a successfully-computed untracked *value* is a no-op, verified separately (\`(.a, error("boom")) = 5\` still raises \`boom\`).

## \`del()\` is deliberately excluded

Investigating the issue's own open question ("the exact shape of yq's silent no-op rule... needs its own verification pass") turned up a real complication: \`del()\` does **not** share the simple per-branch-independent model its three siblings do.

```bash
$ yq 'del(.a, 1)'      # a: 1, b: 2  -- nothing deleted
$ yq 'del(1, .a)'      # b: 2       -- .a IS deleted (same two arguments, reversed!)
```

The pattern (verified across several argument-order permutations) is consistent with \`del()\` processing its targets in *reverse* of the given order and aborting the rest — without undoing whatever already completed — the instant it hits an untracked one. Naively extending this PR's fix to \`del()\` would make it delete *more* than real yq does for some orderings (\`del(.a, 1)\` would delete \`.a\`, which real yq leaves alone) — a data-loss-shaped divergence in the wrong direction, not a cosmetic one.

\`del()\` therefore keeps the pre-existing (safe) raise, gated via the \`short_circuit_del_root\` flag \`resolve_dynamic_indexes\` already threads to distinguish \`del()\`'s call shape from its siblings. Filed the reverse-order/abort algorithm as its own follow-up: #1865.

## Verification

Full \`jq_cli_tests\`/\`yq_cli_tests\` suites (1079 + 1282, +6 new), \`eval::\` unit tests, both clippy legs, \`--no-default-features\` (no_std), \`cargo fmt --check\`, and \`cargo doc --all-features\` all green.

Fixes #1764

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -667,7 +667,7 @@ its M2 fast path shares `stream_cursor!` with stdout, so an M2-eligible `map` ke
 and style through `-i` too — #1349 is about the `--inplace` **DOM fallback**, which a
 non-M2-eligible filter still reaches.
 
-### An untracked `Expr::Comma` branch — resolved for `path()`/`=`/`|=`, still open for `del()`
+### An untracked terminal path branch — resolved for `path()`/`=`/`|=`/compound-assigns, still open for `del()` and for trailing navigation
 
 [#1764](https://github.com/rust-works/succinctly/issues/1764): `reject_untracked_at_terminal`
 (`eval.rs`) is the shared check answering "is a `path()`/`del()`/`=`/`|=` branch that resolved
@@ -675,9 +675,9 @@ to a *computed* value, rather than a real navigation, actually an error" — jq'
 always yes (`Invalid path expression with result <v>`), and this check used to raise
 unconditionally for both modes.
 
-Real yq's answer for `path()`/`=`/`|=` is no: an untracked comma branch is a silent no-op,
-and every *other* branch is still written normally, regardless of the untracked branch's
-position or how it was produced:
+Real yq's answer for `path()`/`=`/`|=`/the compound-assign operators (`+=`/`-=`/`*=`; yq has
+no `/=`/`%=`/`//=` syntax at all to check against) is no: an untracked terminal branch is a
+silent no-op, and every *other* branch is still written normally, regardless of position:
 
 ```bash
 $ echo 'a: 1
@@ -688,7 +688,22 @@ b: 2' | succinctly yq '(.a, 1) = 5'   # matches, was: Error: Invalid path expres
 
 Confirmed position-independent (untracked first, middle, or last; all-branches-untracked;
 a computed value from an arbitrary expression rather than a bare literal) — every case gives
-the same "skip the untracked branch, write every other one" result.
+the same "skip the untracked branch, write every other one" result. **Not specific to a
+multi-branch `Expr::Comma` despite the check's own name**: a single bare untracked
+expression with no comma at all is the identical no-op —
+`(1) = 5` on `{a: 1}` leaves it unchanged too. A genuine error/break/halt produced *while
+computing* what would otherwise be an untracked value is not itself untracked and still
+propagates normally: `(.a, error("boom")) = 5` still raises `boom`.
+
+**Only the terminal position is fixed.** An untracked branch followed by *further
+navigation* — `(.a, 1 | .[]) = 5`, still real yq's own no-op — never reaches
+`reject_untracked_at_terminal` at all: `resolve_node`'s own `Expr::Iterate` arm (and
+`resolve_index_expr`'s analogous dynamic-key check) each raise independently and
+unconditionally before a branch gets anywhere near the terminal check. Fixing that needs
+`is_del`-equivalent awareness threaded into `resolve_node`, a 21-call-site function with its
+own scattered, independently jq-mode-verified checks — a materially larger and riskier
+surface than this fix's single terminal-position check, so left open rather than guessed at
+here. Tracked as [#1868](https://github.com/rust-works/succinctly/issues/1868).
 
 **`del()` does not share this model and is deliberately excluded from the fix.** Its real
 behaviour, confirmed live across argument-order permutations, is order-*dependent* in a way
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -675,9 +675,9 @@ to a *computed* value, rather than a real navigation, actually an error" — jq'
 always yes (`Invalid path expression with result <v>`), and this check used to raise
 unconditionally for both modes.
 
-Real yq's answer for `path()`/`=`/`|=`/the compound-assign operators (`+=`/`-=`/`*=`; yq has
-no `/=`/`%=`/`//=` syntax at all to check against) is no: an untracked terminal branch is a
-silent no-op, and every *other* branch is still written normally, regardless of position:
+Real yq's answer for `path()`/`=`/`|=`/the compound-assign operators (`+=`/`-=`/`*=`) is no:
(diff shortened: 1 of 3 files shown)


newhoggy:
## Round 2 review — scope was undersold and one real gap found; fixed/filed

Two independent review passes converged on the same two substantive points, plus a design/doc-staleness note:

**1. The fix's own framing as "an \`Expr::Comma\` branch" undersold its actual scope.** The gate condition is really "any untracked terminal branch," not comma-specific — verified live: a single bare untracked expression with no comma at all, \`(1) = 5\`, is the identical no-op in real yq. Also verified (previously unconfirmed against the oracle for this specific fix): compound assigns \`+=\`/\`-=\`/\`*=\` share the same behavior (yq has no \`/=\`/\`%=\`/\`//=\` syntax at all to check against — unrelated, pre-existing question). Added tests pinning both, plus \`path()\` coverage (previously untested), and reworded the doc comment/limitations.md entry to state the real scope instead of the narrower comma-specific framing.

**2. Confirmed real gap: an untracked branch followed by further navigation still raises.** \`(.a, 1 | .[]) = 5\` is real yq's own no-op too, but never reaches \`reject_untracked_at_terminal\` at all — \`resolve_node\`'s own \`Expr::Iterate\` arm (and \`resolve_index_expr\`'s analogous dynamic-key check) raise independently and unconditionally first. Fixing this needs \`is_del\`-equivalent awareness threaded into \`resolve_node\`, a 21-call-site function with its own scattered, independently jq-mode-verified checks (each citing its own prior issue, e.g. #843) — a materially larger, riskier surface than this fix's single terminal-position check. Filed as #1868 rather than attempting it under review pressure; doc comments and limitations.md now name the gap explicitly instead of implying full coverage.

**3. Minor**: \`short_circuit_del_root\`'s own doc comment was silently reused as \`is_del\` without documenting the second responsibility — added a note explaining both concerns currently coincide only because \`del()\` is the sole caller wanting non-default behavior for either.

Verification: full \`jq_cli_tests\`/\`yq_cli_tests\` suites (1079 + 1285, +3 new tests this round), \`eval::\` unit tests, both clippy legs, \`--no-default-features\`, \`cargo fmt --check\`, and \`cargo doc --all-features\` all green.

newhoggy:
## Round 3 review (8 finder angles) — no new logic bugs; one simplification, three coverage gaps closed

**Simplification** (2 independent angles converged on this): \`reject_untracked_at_terminal\`/\`reject_untracked_prefix_too\` were made generic over \`S: EvalSemantics\` purely to evaluate one \`S::TAG\` comparison — since neither was generic before this fix, this monomorphizes (compiles twice, once per \`EvalSemantics\` impl) two functions that only needed a single boolean fact the caller already has. Replaced with a plain \`skip_untracked: bool\` computed once in \`resolve_dynamic_indexes\`. Same behavior, one compiled copy instead of two.

**Three previously-untested shapes, now verified and covered:**
- \`path()\`'s own trailing-iterate case (\`path(1 | .[])\`) is actually **covered** by this fix, unlike \`=\`/\`|=\`/compound-assign's identical shape (which is #1868's real gap) — \`path()\` defers and re-splices a trailing bare iterate around the terminal check, so it reaches \`reject_untracked_at_terminal\` where the others never do. Confirmed live and now tested; corrected the doc/limitations.md phrasing that had implied this was also part of the #1868 gap.
- An untracked value produced by a \`try\`/\`catch\` handler (the #843/#1297 write-corruption-prevention mechanism) behaves identically to a bare-literal one once it reaches the check — confirmed live (\`try (.a, error({"y":99})) catch select(true) = "X"\` writes \`.a\`, drops the caught branch) and locked in with a test, including confirming \`del()\` still correctly excludes itself from the skip in this shape too.
- succinctly's own \`/=\`/\`%=\`/\`//=\` extension forms (no real yq syntax to check against at all — real yq's parser rejects them outright) inherit the same skip for internal consistency; now tested rather than left as an unverified side effect.

Three of the eight angles (altitude, reuse-pattern, and one correctness-verification pass) reported no findings, confirming the scoping decisions and call-site plumbing are sound.

Verification: full \`jq_cli_tests\`/\`yq_cli_tests\` suites (1079 + 1288, +6 new tests this round), \`eval::\` unit tests, both clippy legs, \`--no-default-features\`, \`cargo fmt --check\`, and \`cargo doc --all-features\` all green.

Three rounds in, no new logic bugs surfaced this round — proceeding to merge.

PR #1870, opened 2026-08-29
fix(jq): reduce/foreach's source expression is now path-checked

### LINKED ISSUE #1467: jq: reduce/foreach's source expression isn't path-checked, so path(reduce (keys[]) as $k (.; .)) is accepted where jq refuses
**Severity: Low** (accepts where jq refuses; narrow trigger)

## Summary

`resolve_reduce`/`resolve_foreach` (`src/jq/eval.rs`, #1440) evaluate the fold's source
expression with the ordinary value evaluator (`eval_owned_expr_fork`), so the source is
never path-checked. Real jq evaluates it in the ambient path context, and raises when the
source navigates *through* an untrackable value.

Only one shape actually diverges — jq accepts most sources:

| Filter (input `{"a":1}`)                            | jq                                          | succinctly |
|-----------------------------------------------------|---------------------------------------------|------------|
| `path(. as $x \| reduce (1,2) as $i (0; $x))`        | `[]`                                        | `[]`       |
| `path(. as $x \| reduce (.a) as $i (0; $x))`         | `[]`                                        | `[]`       |
| `path(. as $x \| reduce (.[]) as $i (0; $x))`        | `[]`                                        | `[]`       |
| `path(. as $x \| reduce (keys) as $i (0; $x))`       | `[]`                                        | `[]`       |
| `path(. as $x \| reduce (range(2)) as $i (0; $x))`   | `[]`                                        | `[]`       |
| `path(reduce (keys[]) as $k (.; .))`                 | error: near attempt to iterate through `["a"]` | `[]`    |

So the rule is not "the source is path-shaped" — `keys` and `range(2)` are computed and jq
accepts both. It is that the source is *evaluated with tracking on*, and only an attempt to
navigate through a computed value (`keys[]` iterating a computed array) raises.

Write side: `echo '{"a":1}' | succinctly jq -c '(reduce (keys[]) as $k (.; .)) = 9'` gives
`9`; jq raises and leaves the input untouched.

## Root cause and the objection that no longer applies

#1440's own doc gives the reason it wasn't attempted: routing the source through
`resolve_node` "would silently narrow a generator like `range(3)` to one element", because
`resolve_leaf`'s catch-all keeps only the first output of a multi-output untracked
expression (#986).

That objection is avoidable — the fold does not need `resolve_node`'s *branches*, only its
*escape*. Running `resolve_node(source, value, trackable)` purely for the error, discarding
the branches, and continuing to take the actual values from `eval_owned_expr_fork` keeps the
generator at full width and reproduces jq's rule directly (`keys[]` raises "near attempt to
iterate"; `range(2)`/`keys` reach `resolve_leaf`'s catch-all and defer per #986, i.e. no
error).

## The real cost, which is different

That formulation evaluates the source **twice**, so a source with side effects runs twice:
`path(. as $x | reduce (stderr) as $i (0; $x))` would write two lines where jq writes one.
Whether that is acceptable, or whether the check should be restricted to sources that can
navigate at all, or whether it wants a shape-only static check instead, is the actual open
design question here — not the generator-narrowing one.

Recorded as the source-expression note under the fold divergences in
[`docs/compliance/jq/limitations.md`](https://github.com/rust-works/succinctly/blob/main/docs/compliance/jq/limitations.md).

## Related

- #1440 — added the fold path-resolution arms.
- #986 — `resolve_leaf`'s defer-don't-raise rule for a non-path-shaped value, which is what
  makes the branches unusable but the escape usable.

newhoggy:
Fixes #1467

## Root cause

`resolve_reduce`/`resolve_foreach` evaluated a fold's source expression with the ordinary value evaluator (`eval_owned_expr_fork`), never path-checking it. Real jq evaluates the source "with tracking on" and raises when it navigates *through* a computed value: `path(reduce (keys[]) as $k (.; .))` on `{"a":1}` raises jq's own "near attempt to iterate through" error where succinctly silently accepted it (confirmed on both the `path()`-read side and the `(reduce ...) = 9` write side).

This issue's own comment thread already settled the design question (evaluate twice vs. restrict to navigating sources vs. a static-only check) with live measurements against jq 1.7.1 before I picked it up — see the issue for the full analysis. I'm implementing the ranked fallback (#2: resolve-for-escape + re-evaluate unconditionally) rather than the primary recommendation (threading a new "keep every output while tracking" mode through `resolve_node`'s entire recursive dispatch), since the primary is a materially larger, more invasive change to a widely-shared traversal for what the fallback already correctly handles at a documented, narrow cost.

## Fix

New shared `path_check_fold_source` routes the source through `resolve_node` first, discarding its branches (`resolve_leaf`'s `#986` defer-not-raise leaf would otherwise silently narrow a multi-output source like `range(3)` to one element if its branches were taken as the real values), and keeping only a fatal subset of the escape: `Halt` (never swallowed) and `is_untracked_navigation_error` (the "near attempt to access/iterate" family this check exists to surface). Any other escape — an ordinary runtime `Error` or a `Break` — is deliberately ignored and falls through to the existing `eval_owned_expr_fork` call unchanged.

That "ignore any other escape" carve-out was discovered the hard way, not designed up front: an earlier version treated *any* `resolve_node` escape as fatal, which broke `test_reduce_foreach_escapes_and_partial_prefix_1440`'s `foreach (1,2,error("s"))` case by discarding its `[[], []]` partial prefix — `resolve_node` hit the same ordinary error first and the fix short-circuited before `eval_owned_expr_fork` ever got to stream the two prior outputs. Fixed by narrowing the fatal-escape check to exactly the two kinds that carry genuinely new information.

## Accepted cost (documented in `docs/compliance/jq/limitations.md`)

Since `resolve_leaf`'s stop-after-first sink still fully evaluates the source's first output to answer the check, a source whose first output has a side effect (`stderr`, a consumed `input`) fires it twice where jq's own single tracked evaluation fires once — measured live against jq 1.7.1: `path(. as $x | reduce (stderr) as $i (0; $x))` writes `{}` once to stderr in jq, `{}{}` in succinctly.

## Verification

- Live-verified the full acceptance matrix (six source shapes: `1,2` / `.a` / `.[]` / `keys` / `range(2)` / `keys[]`) against jq 1.7.1 for both `reduce` and `foreach`, both the `path()` read side and the `=` write side, before and after implementing.
- New jq-golden fixtures (captured from the pinned oracle via `scripts/sync-jq-golden.sh`, not hand-typed) cover the untracked-iterate raise for both `reduce`/`foreach`, the accepted navigated/computed source shapes, and `Halt` propagation through the source.
- 100% patch coverage.
- Full test suite green under both `cargo test --verbose` (default features, matches CI) and the full feature set, no_std build, clippy clean (both legs), before and after rebasing onto `main`.
- Two unrelated, pre-existing transient test failures during verification (`test_getpath_propagates_halt_in_path_argument`, `test_walk_array_zero_output_child_drops_out_of_the_array`-family flakiness pattern) — both confirmed transient via isolated rerun and a clean full-suite rerun.

## Test plan

- [x] `cargo test --test jq_golden_tests` (full 570-case corpus, no regressions)


newhoggy:
Ran the mandatory multi-angle code review. Summary:

- **Confirmed correct**: the recursion mirrors real jq's acceptance matrix exactly (six source shapes, both `reduce`/`foreach`, both `path()` read and `=` write sides), `is_untracked_navigation_error`'s scoping is correct, and the "ignore any other escape" carve-out (added after finding the original partial-prefix regression myself) is sound.
- **Two real, serious bugs found and fixed**:
  1. **Silent data corruption for `input`/`inputs` sources**: evaluating the source twice (once for the check, once for the real value) desynchronized the shared input-reader position — `reduce input as $x (0; $x)` over multiple documents silently used the *wrong* document. Fixed by gating the whole check on `any_subexpr` finding an actual navigation step (`Field`/`Index`/`Slice`/`Iterate`) anywhere in the source — a source with none (`range(n)`, `keys`, a literal, and critically `input`/`inputs`) now skips the re-evaluation entirely, removing the corruption risk for the common case.
  2. **`foreach`'s own per-element streaming loses a legitimate pre-error prefix** when a navigating multi-element comma-list source fails partway through (`foreach (1,2,keys[])` — jq streams two outputs before raising, this fix printed nothing). Root cause is architectural (checking the whole source up front vs. real jq's single interleaved evaluation) and would need the same "materially larger" architecture the issue's own resolving discussion already declined — filed as #1872 rather than expanded into this PR.
- **Efficiency finding**: the extra `resolve_node` traversal for a navigating source is real but now scoped only to sources that actually need it (the syntactic gate above also addresses this for the common non-navigating case).

All local verification (build/clippy/tests under both feature sets, no_std build, 100% patch coverage, full 570+-case jq-golden corpus) re-run clean after the fixes. Two rounds of unrelated, pre-existing transient test failures during verification — both `signal 9`-killed child processes (OS/resource contention under this machine's heavy concurrent session load, not logic failures) — confirmed transient via isolated rerun and clean full-suite reruns both times.

PR #1875, opened 2026-08-30
fix(jq): stop through_slice demanding array from a no-op null tail

### LINKED ISSUE #1873: jq: `.a[0:1][]? = 9` raises a slice write-time error where jq no-ops (slice split misclassified as a terminal write)
**Severity: Low**

## Summary

A slice followed by an iterate (`.a[0:1][]? = 9`) raises a write-time slice
error where real jq treats the path as yielding nothing and leaves the document
alone. succinctly appears to treat the slice as the *terminal* write target
(assigning the RHS to it) instead of navigating *through* it to the iterate.

## Repro (jq 1.7.1, confirmed live)

```console
$ echo null | jq -c '.a[0:1][]? = 9'
null
$ echo null | succinctly jq -c '.a[0:1][]? = 9'

$ echo '{}' | jq -c '.a[0:1][]? = 9'
{}
$ echo '{}' | succinctly jq -c '.a[0:1][]? = 9'
```

Also reproduces on `{"a":null}` and on an unrelated-object input
(`{"x":{"y":{"z":1}}}`) -- i.e. whenever the slice's parent has to be created.
With an *existing* array parent the two agree, so this is specific to the
"slice reached through a chain that does not exist yet" shape.

## Not a regression

Verified identical on `main` before and after #1428's fix (which covers the
`Iterate`-after-a-created-parent case but deliberately leaves the slice split
alone -- a slice write on a missing parent genuinely *does* write, so it is
correctly excluded there). These 4 cases were the only remaining divergences in
#1428's own 966-combination differential sweep.

## Root cause direction

`set_path`'s `split_at_slice` arm hands `through_slice` a `terminal_write`
flag computed as `matches!(split.tail, Expr::Identity)`
([src/jq/eval.rs](src/jq/eval.rs), the `SliceEditFlags` construction). When the
tail is an `Iterate` rather than `Identity`, the write should go *through* the
slice to each element, the way `update_path`'s own `Expr::Slice` arm does with
its `terminal_write: false` -- `[1,2,3,4] | .[1:3][] |= .*10` is `[1,20,30,4]`,
which already works. `=` appears to lack the matching read-through for its
slice split when the parent must be autovivified first.

Worth checking whether `.a[0:1][]? |= 9` (the `|=` twin) agrees with jq before
deciding the shape of the fix, and whether the yq-mode slice no-ops (#1101/
#1116/#1142) interact -- they gate a lot of behaviour through the same
`through_slice`.

## Why this is separate from #1428

#1428 was specifically about autovivification surviving a write that never
happens, via the `Iterate` split and the terminal-component walk. This is the
*slice* split raising an error jq does not, which is a different code path and
a different question (navigate-through vs terminal-write classification), not
a stranded-creation one.

newhoggy:
## Summary

`through_slice`'s `Null` arm (added by #1340) required the `edit` closure's
result to become an `Array` unconditionally, even for a mid-chain tail
(`terminal_write: false`) whose own navigation determines there's nothing to
write at all. A `?`-suppressed `cannot_iterate` over the empty slice range
left `edit`'s `sub` untouched at `Null`, which the arm forced through the
array check anyway, raising "A slice of an array can only be assigned
another array" where real jq leaves the whole document unchanged.

```
$ echo null | jq -c '.a[0:1][]? = 9'
null
$ echo null | succinctly jq -c '.a[0:1][]? = 9'   # before this fix
jq: error: A slice of an array can only be assigned another array
```

Verified live against the pinned oracle (`/usr/bin/jq` 1.7.1) that `=`,
`|=`, and every compound-assignment operator (`+=`/`-=`/`*=`/`/=`/`//=`/
`%=`) share this bug (all go through the same `through_slice` choke point),
and that the fix doesn't regress any of the surrounding shapes: a
non-optional `Iterate` tail still raises "Cannot iterate over null", a
`Field` tail after the slice still raises the array requirement (matching
jq erroring there too), a literal-`null` RHS on the slice's own terminal
write still errors, an `Index` tail still auto-vivifies successfully, and
an existing array parent is untouched by this arm entirely.

## Scope: this closes the reported shape, not the whole class

Round-2 review found a deeper, related gap: a tail with a `Field`/`Index`
step *before* the swallowed no-op (`null | .a[0:1].c[]? = 9`) still wrongly
errors, and one shape (`null | .a[0:1][0]?.c[]? = 9`) wrongly *commits* a
write real jq no-ops on. Both confirmed byte-for-byte identical before and
after this PR's fix — not a regression, not newly introduced. This is the
same family of bug as **#1428** (a write-time decision inferred from what
navigation left behind, rather than an explicit "did anything actually get
written" signal), just one nesting level deeper than what #1873 itself was
ever scoped to. #1428 is already being worked as its own, more structural
fix in a separate, further-along branch — not attempted here.

## Other notes from review

- The residual `.a: null` stub in every no-op output (e.g. `null |
  .a[0:1][]? = 9` → `{"a":null}` instead of real jq's untouched `null`) is
  also #1428's territory (eager `Field` autovivification during
  navigation, reachable with no slice involved at all). Pinned explicitly
  in a separate characterization test, not conflated with this fix's own
  verification.
- Filed **#1876** (string-slice terminal write skips `edit`'s side effects)
  and **#1877** (`|= empty` through a nested terminal slice) for two
  unrelated pre-existing gaps found adjacent to this code during review.
- `docs/compliance/jq/limitations.md` had two stale rows/a stale sentence
  left over from #1340 (claiming slice-null writes still error) — fixed
  here since it's directly adjacent to what this PR touches.

## Test plan

- New tests: `test_assign_slice_mid_chain_iterate_over_null_noops_1873`
  (the fix itself, every assertion either error-message-exact or
  oracle-byte-exact) and a separate, explicitly-labeled characterization
  test for the pre-existing #1428 stub.
- Full `jq::eval` unit test module (1118 tests): pass.
- `cargo test --features cli,simd,regex,serde --no-fail-fast` (full suite,
  including doctests): pass.
- `cargo test` (plain, default features): pass.
- Both clippy legs (`--all-features` and CI's `std,simd,serde,cli,regex,
  bench-runner,large-tests,mmap-tests`): clean, `-D warnings`.
- `cargo fmt --check`: clean.
- `cargo doc --all-features --no-deps`: clean (pre-existing unrelated dead-code warnings only).
- `cargo build --no-default-features`: clean.

Fixes #1873

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -374,8 +374,6 @@ have to be recorded here.
 
 | Filter             | Input   | jq                                  | succinctly                        |
 |--------------------|---------|-------------------------------------|-----------------------------------|
-| `.[1:2] = ["x"]`   | `null`  | `["x"]`                             | `Cannot index null with object`   |
-| `.[1:2] \|= ["x"]` | `null`  | `["x"]`                             | `Cannot index null with object`   |
 | `@uri`             | `[1,2]` | `"%5B1%2C2%5D"`                     | `expected string, got array`      |
 | `@base64`          | `5`     | `"NQ=="`                            | `expected string, got number`     |
 | `flatten("x")`     | `[1,2]` | `[1,2]` (ignores non-integer depth) | `expected number, got non-number` |
@@ -1139,8 +1137,13 @@ probes, so the two-sided manifest check forced them to start matching in the sam
 
 What #366 did *not* build: computed bounds. `.[$a:$b]` is still a parse error, because the
 parser folds slice bounds to integer literals; see
-[docs/reference/jq-language.md](../../reference/jq-language.md). And writing through a
-slice does not vivify `null` — see the table above for why that was deliberate.
+[docs/reference/jq-language.md](../../reference/jq-language.md). Writing through a slice
+against `null` used to raise `Cannot index null with object` instead of auto-vivifying —
+[#1340](https://github.com/rust-works/succinctly/issues/1340) brought that in line with
+jq's own `setpath()` behavior, and [#1873](https://github.com/rust-works/succinctly/issues/1873)
+later fixed a gap in that same auto-vivification for a slice with more path *after* it
+(`.a[0:1][]? = 9` on a missing `.a` now no-ops instead of raising a write-time error, matching
+jq).
 
 ## Reading a path is indexing
 
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -17241,9 +17241,38 @@ fn through_slice<E: From<EvalError>>(
         // `-=`/`*=` on a null slice target exactly as unaffected by this
         // fix as `#1340`'s own issue text was ever scoped to be (jq mode
         // only) — not newly right, but not newly wrong either.
+        //
+        // `!terminal_write` no-op (#1873): the doc comment above described
+        // `edit`'s result as always having to become an array once it's
+        // done, but that's only true when the slice itself is the write.
+        // When there's a real tail after it (`.a[0:1][]? = 9` on missing
+        // `.a`), `edit` is `set_path`'s own recursion into that tail
+        // against a fresh `Null` -- and if the tail's navigation
+        // determines there's nothing to write (here, iterating zero
+        // elements over nothing, with the `cannot_iterate` error swallowed
+        // by `?`), it leaves `sub` untouched at `Null` and returns `Ok`
+        // without ever attempting a real write. Forcing that case through
+        // the array requirement raises "A slice of an array can only be
+        // assigned another array" where real jq leaves the whole document
+        // unchanged (confirmed live for `.a[0:1][]? = 9` and its `|=`
+        // twin, on `null`/`{}`/`{"a":null}`). A tail that *does* write
+        // something -- autovivifying a field (`.a[0:1].b = 9` -> `sub`
+        // becomes an `Object`) or an element (`.a[0:1][0] = 9` -> `sub`
+        // becomes an `Array`) -- never leaves `sub` as `Null`, so it still
+        // falls through to the same array requirement below, matching jq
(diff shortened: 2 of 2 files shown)


newhoggy:
Heads-up: this landed correctly for `set_path`, but there's a live gap on `main` for the `update_path` spelling of the same shape.

The new `Null`-arm no-op is only safe when `terminal_write` is accurate. `set_path`'s slice arm computes it (`matches!(split.tail, Expr::Identity)`), so `null | .[0:1] = null` still raises, exactly as this PR's description verifies. But `update_path`'s slice arm **hardcodes `terminal_write: false`** (`src/jq/eval.rs`, the mid-chain `Expr::Slice` arm), so a filter that writes `null` straight to the slice now takes the no-op path:

```console
$ echo null | jq -c '(.a[0:1]|.) |= null'
jq: error (at <stdin>:1): A slice of an array can only be assigned another array   # exit 5

$ echo null | succinctly jq -c '(.a[0:1]|.) |= null'   # main @ 49f063ee7
{a:null}                                                                        # exit 0
```

Same for `(.[0:1]|.) |= null` and on `{"a":null}` input (no autovivification involved at all). It drops the error *and* leaves the autovivified `.a` behind. The hardcoded `false` was harmless before this PR — the arm errored on every unwritten `Null` regardless — so this is only reachable now.

I hit the identical thing on #1428's branch (PR #1874), where my own third review round caught it; the fix there computes the flag via a small `is_effectively_identity` helper, mirroring `set_path`'s arm. #1874 already contains it plus a regression test (`test_slice_update_with_identity_tail_still_raises_1428`), and I resolved the rebase by keeping **this PR's** spelling of the `Null` arm rather than my own equivalent one — so no action needed here unless you'd prefer it fixed separately/sooner than #1874 merges.

Measured against current `main` over 2,640 input × path × operator combinations: 216 divergences fixed, 0 behavioural regressions.

PR #1878, opened 2026-08-30
fix(yq): compact sequence-item indent no longer compounds at deeper levels

### LINKED ISSUE #1485: yq: 'compact' sequence-item indent doesn't compose with deeper nested containers at non-default -I
**Severity: Low** — cosmetic indentation only; values round-trip identically
and output is valid, re-readable YAML. Found reviewing #1362's fix.

## Summary

Real yq's "compact" block-sequence-item rendering (`- ` shares its line
with the value's first field/element, #785) only applies its 2-column
offset to *that first line*. Structural content nested **inside** that
first field's own value continues from the item's own indent plus a
*normal* full `indent_spaces` step per level — not from the
compact-adjusted column. succinctly's `compact_yaml_indent`/
`deeper_yaml_indent` model conflates the two: it threads the
compact-adjusted string down as the base for every deeper recursive step,
so each level below the first compounds a discrepancy against real yq.

This is **not new** — it already existed in the plain (non-anchored)
compact branch since #785, well before #1362. #1362 additionally applied
the same (already-flawed) `compact_yaml_indent` pattern to the anchor/tag
branch it fixed, but confirmed **strictly reduces** divergence there
rather than introducing it (see the measurement below) — this issue tracks
fixing the shared underlying model, not a regression from #1362.

## Repro (confirmed live against pinned yq v4.53.3, both with and without an anchor)

```
$ printf 'l:\n  - p:\n      q:\n        r: 1\n  - x\n' | yq -I=4 '.'
l:
    - p:
        q:
            r: 1
    - x
$ printf 'l:\n  - p:\n      q:\n        r: 1\n  - x\n' | succinctly yq -I=4 '.'
l:
    - p:
          q:
              r: 1
    - x
```

Columns (oracle vs succinctly, `main` before #1362, `p`/`q`/`r`): oracle
puts `p` at col 6, `q` at col 8, `r` at col 12 (deltas: +2 then +4 per
level below the first). succinctly (pre-#1362, and still after -- this
branch is unrelated to the anchor case) puts them at col 8/12/16 (a
uniform +4 per level from the start).

**With an anchor, #1362's fix measurably improves this rather than
regressing it** -- verified with `/usr/bin/env python3` byte-column
diffing, pre- vs post-#1362, `-I=4`, 3 levels deep under `- &x`:

| level | oracle | pre-#1362 | post-#1362 |
|-------|-------:|----------:|-----------:|
| p     | 6      | 8 (+2)    | 6 (+0)     |
| q     | 8      | 12 (+4)   | 10 (+2)    |
| r     | 12     | 16 (+4)   | 14 (+2)    |

## Root cause

`compact_yaml_indent(indent)` (light.rs) / the DOM's
`format!("{indent}  ")` produce the correct *visual* column for the
compact line's own content, but that same string is then handed to the
recursive call as if it were an ordinary one-level-deeper `indent` --
so a further nested field's own `deeper_yaml_indent`/`config.indent_str`
step is computed relative to the *compact* column (+2) instead of
relative to what real yq treats as the conceptual "one normal step in"
position. Real yq's model appears to be: the compact line's first field
renders at `dash_indent + 2`, but everything *nested inside that field's
own value* is computed as if the field had instead been rendered at
`dash_indent + indent_spaces` (a normal full step) -- i.e. two different
base indents need to be threaded separately (one for immediate compact
siblings, one for deeper structural recursion), where the current code
uses only one.

## Where this lives

- `src/yaml/light.rs` -- `stream_yaml_value`'s Sequence block-style arm
  (both the anchor/tag branch #1362 touched, and the plain compact `else`
  branch, which has always had this bug).
- `src/bin/succinctly/yq_runner.rs` -- `emit_yaml_value_at_depth`'s
  mirrored block-sequence handling (same two branches).
- Likely also `write_deferred_value` (light.rs, ~6573) for a compact
  sequence item's own *deferred scalar* value (e.g. a block scalar `|`
  under `- `) -- not independently verified against the oracle, but it
  shares the same `deeper_yaml_indent`-vs-compact conflation pattern; worth
  checking as part of the same fix.

## Suggested fix direction

Not investigated in depth. Likely needs the compact-rendering call sites
to thread two indent values instead of one: the visual column for the
compact line/its immediate siblings, and a separate "normal step" base
(`dash_indent + indent_spaces`) for any recursive descent into that first
field's own nested value -- worth a short design note before implementing,
since it touches the shared indent-threading shape of both the streaming
and DOM writers (and the #1448 duplication between them makes any fix a
2-3x mirrored change until that's consolidated).

## Related

- #1362 (the narrower single-level case this generalizes from; fixed).
- #1448 (light.rs deferred-value writers have redundant duplicated
  dash/anchor/tag rendering -- any fix here likely wants to land after or
  alongside that consolidation to avoid a third hand-mirrored copy).
- Minor, unconfirmed side-note from the same review: the DOM path's
  anchor-deferred branch (`yq_runner.rs`, `elem_comments.declared_anchor()`)
  has no tag-only equivalent to `light.rs`'s
  `anchor.is_some() || tag.is_some()` condition -- a tag-only (no anchor)
  deferred sequence item may not be mirrored in the DOM path at all. Not
  independently verified with a live repro; worth checking during the same
  investigation.

newhoggy:
Fixes #1485

## Root cause

Real yq's compact block-sequence-item rendering (`- ` sharing its line with the value's first field/element, #785) only applies its 2-column offset to *that first line*. Content nested *inside* that first field's own value continues from the item's own indent plus an ordinary `indent_spaces` step per level — not from the compact-adjusted column. `light.rs`'s `stream_yaml_value`/`emit_yaml_value_at_depth` (`yq_runner.rs`) threaded the compact-adjusted string down as the base for every deeper recursive step, so each level below the first compounded a discrepancy against real yq.

## Fix

Both writers now thread a second indent value alongside the existing one: `recursion_base` is the pre-compact indent a nested container's own further descent should step from, staying equal to the visible `indent` everywhere except immediately after a compact positioning. `light.rs`'s `stream_yaml_value`, `write_deferred_value`, and `stream_yaml_sequence` (the `--slurp` counterpart), plus `yq_runner.rs`'s `emit_yaml_value_at_depth` (the DOM/write path), all thread it identically.

## A genuinely different formula at real yq's own default indent

Live-probing across `-I=2` through `-I=6` (not just the issue's own `-I=4` example) found real yq's *default* indent (`-I=2`) hits a different formula, not a smaller instance of the same one: `indent_spaces` (2) equals `COMPACT_DASH_WIDTH` (also 2) exactly at that width, so a single ordinary step from the pre-compact indent would land the first nested field at the *same* column the compact field's own key already visually occupies — invalid YAML nesting. Real yq instead steps *past* that column by `indent_spaces` in this one case; `-I=3` and up never need the correction, since a single ordinary step already clears it there. New `compact_child_indent` (`light.rs`) and its DOM twin (`yq_runner.rs`) implement this "clears the compact visual column, else fall back" rule.

## A second bug found and fixed along the way

The issue itself flagged `write_deferred_value` as worth checking, unverified. Confirmed live: a sequence item's own deferred block scalar (`- |`) is compact-positioned like every other sequence item, where the analogous mapping-field position (`k: |`) is an ordinary step — `write_deferred_value` computed an ordinary step unconditionally for both, since it's shared between the two callers. Now takes a pre-computed `child_indent` from each caller instead of deriving it internally, so each can supply the right kind of step.

## Verification

- Live-verified every case against yq v4.53.3 before implementing: 3-level nesting at `-I=2` through `-I=6`, a non-first sibling field's own nested value, and the block-scalar deferred-value asymmetry, on both the streaming (identity) and DOM (forced write) paths.
- 100% patch coverage on both files.
- Full test suite green under both `cargo test --verbose` (default features, matches CI) and the full feature set, no_std build, clippy clean (both legs), before and after rebasing onto `main`.
- Two rounds of unrelated, pre-existing transient test failures during verification (jq `input`/stdin-related tests, unrelated to YAML output formatting) — both confirmed transient via isolated rerun and clean full-suite reruns.

## Test plan

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -355,6 +355,43 @@ would still not close the gap (the irregular per-level part is unmodeled by both
 every non-default `-I` width, not just `-I0`), so this remains open rather than folded into
 #1575's fix.
 
+### Stacked compact block-sequence items — one narrow, pre-existing width anomaly
+
+[#1485](https://github.com/rust-works/succinctly/issues/1485) fixed the general rule for a
+compact block-sequence item's inlined field/element (`- ` sharing its line with a mapping
+value's first field, #785): content nested *inside* that first field's own value steps from
+the item's own pre-compact indent (`recursion_base`) by an ordinary `indent_spaces` amount —
+not from the item's 2-column-wider visual column — with one width-dependent correction at
+real yq's own default (`-I=2`, where a single ordinary step wouldn't clear the compact visual
+column). This also fixed a **stacked** chain, where a block sequence is itself the compact
+value nested directly inside *another* compact sequence element (`- - b: ...`, a sequence
+whose sole element is itself a sequence whose sole element is a mapping): every level must
+keep forwarding the *original* pre-compact base through every compact level, not just its
+own immediate one.
+
+**One narrow anomaly survives, unfixed, at a specific (nesting depth, `-I`) intersection**:
+for a chain of exactly `D` stacked compact sequence levels, real yq's own column for the
+first nested field is non-monotonic in `-I` right at `-I = D + 1` — verified live against yq
+v4.53.3 across `-I=2` through `-I=6`, both before and after #1485's fix, and confirmed
+byte-identical on pre-#1485 `main` too (this is not something #1485 introduced or worsened).
+For `D = 2` (`printf -- '- - b:\n      c:\n        d: 1\n' | yq -I=<n> '.'`), the observed
+column of `c` is 6/6/8/5/6 for `-I` = 2/3/4/5/6 — genuinely non-monotonic (`-I=5`'s column is
+*smaller* than `-I=2`'s), not just a different linear step per width. The anomaly sits at
+`-I=3` (`D+1`): real yq's own column there (6) breaks what would otherwise be a strictly
+increasing sequence (2→6, 4→8, 5→…, 6→…) by repeating `-I=2`'s value instead of continuing
+upward to 7.
+
+Both succinctly output routes give column 7 at `-I=3` there (an ordinary, monotonic step
+matching the `-I=2`→`-I=4`→... progression), matching each other and matching pre-#1485
+`main`'s own (also non-matching) value — neither has ever modeled this specific anomaly. The
+pattern (`-I = stacked_depth + 1`) suggests a genuine upstream go-yaml quirk tied to some
+internal indent-tracking edge case rather than a spec-conformant rule, but the exact
+mechanism is not yet understood; tracked as
+[#1881](https://github.com/rust-works/succinctly/issues/1881) rather than blocking #1485's
+otherwise-verified fix (which is strictly more correct than what it replaced: it repairs a
+real regression at `-I=4` introduced mid-fix and two cases pre-#1485 `main` never got right
+at all, `-I=5`/`-I=6`).
+
 ### `input`, `inputs`, `input_line_number` are rejected, but at runtime rather than parse time
 
 Real yq has no such builtins at any arity — its lexer rejects the identifiers exactly as it
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -2295,6 +2295,31 @@ fn is_flow_safe(value: &OwnedValue, comments: &CommentTree) -> bool {
 /// renders as `*name`, so it is inline no matter how large the anchored
 /// value it points at happens to be. Getting that wrong renders `b: *x` as
 /// a block mapping with `*x` dangling above it.
+/// DOM twin of `light.rs`'s `compact_child_indent` (#1485): the indent for
+/// content nested one level inside a compact-rendered field/element's own
+/// value. An ordinary `{recursion_base}{indent_str}` step, *unless* that
+/// step wouldn't land past `indent` (the field's own compact-adjusted
+/// indent) -- real yq's own rule, which only bites when `indent_str`'s
+/// width is `<=` the compact `- `/first-field offset (2 columns) --
+/// real yq's own default, `-I=2`, is exactly that boundary. See
+/// `light.rs`'s `compact_child_indent` for the full live-verified
+/// rationale (both writers are pinned against the same oracle behavior,
+/// #763).
+///
+/// Reduces to the pre-existing, always-correct `format!("{indent}{step}")`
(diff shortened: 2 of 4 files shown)


newhoggy:
Addressed the critical regression finding: the stacked-compact-chain case
(`- - b: ...`) was mis-indenting because both compact branches in
`stream_yaml_value`/`emit_yaml_value_at_depth` forwarded their own `indent`
as the recursive call's `recursion_base`, which only holds when the
sequence itself was never compact-positioned.

Confirmed live against yq v4.53.3 as a real regression vs. pre-#1485 `main`
at `-I=4` (main: 8/12, the reviewed commit: 6/10). Fix: forward
`recursion_base` unchanged instead of resetting it to `indent`, in both
compact branches of both files.

Verified this is a net improvement, not just a fix-in-place: comparing
current branch / reviewed commit / pre-#1485 `main` on the `- - b:\n
c:\n        d: 1\n` repro across `-I=2..6`:

| `-I` | pre-#1485 `main` | reviewed commit (8dede1ac5) | this fix | oracle |
|------|-------------------|------------------------------|----------|--------|
| 2    | 6 ✓               | 6 ✓                          | 6 ✓      | 6      |
| 3    | 7 ✗               | 5 ✗                          | 7 (=main)| 6      |
| 4    | 8 ✓               | 6 ✗ (regression)             | 8 ✓      | 8      |
| 5    | 9 ✗               | 7 ✗                          | 5 ✓      | 5      |
| 6    | 10 ✗              | 8 ✗                          | 6 ✓      | 6      |

This fixes the `-I=4` regression and *also* corrects `-I=5`/`-I=6`, which
`main` itself never handled correctly. `-I=3` remains wrong, but at the
exact same (wrong) value `main` already produced — not a regression, a
pre-existing gap.

That `-I=3` gap turned out to be part of a genuine, narrow upstream
anomaly: for a chain of exactly `D` stacked compact levels, real yq's own
output is **non-monotonic in `-I` right at `-I = D+1`** (confirmed
identical on pre-#1485 `main`, both before and after this fix — not
introduced by this PR). Documented in
`docs/compliance/yq/limitations.md` § "Stacked compact block-sequence
items" and filed as #1881 rather than blocking this fix.

Added `test_compact_seq_item_stacked_nested_indent_does_not_reset_1485`
covering the now-fixed `-I=2/4/5/6` cases (streaming + DOM). Full local
verification passed: build/test/clippy under both default and full
`cargo doc --all-features`.

PR #1879, opened 2026-08-30
fix(jq): skip(n; expr) raises on NaN count instead of silently skipping 0

### LINKED ISSUE #1846: jq: skip(n; expr) has the same NaN/non-integer-count divergence #1825 fixed for nth/limit
## Summary

`#1825` (PR #1842) fixed `nth(n; g)`/`limit(n; g)` accepting NaN and non-integer counts the way jq does. Confirmed during that PR's review: `skip(n; expr)` has the exact same bug shape, and unlike `nth`/`limit` it isn't even routed through a shared classifier — the buggy logic is hand-inlined **twice** within `builtin_skip` (once per `QueryResult` arm), the precise "duplicated classification, prone to drift" pattern `classify_limit_n`'s own doc comment (citing #1313) exists to eliminate.

## Repro (confirmed live against the built binary)

```console
$ succinctly jq -n 'skip(nan; 1,2,3,4)'
1
2
3
4
```

Real jq's own `skip` definition is `if $n > 0 then ... elif $n == 0 then expr else error("skip doesn't support negative count") end` — NaN satisfies neither `> 0` nor `== 0`, so it should fall to the `else error(...)` branch, not silently behave as skip-count-0.

```console
$ succinctly jq -n 'skip(1.5; 1,2,3,4)'
2
3
4
```

Truncated to skip 1 instead of jq's own decrementing-loop semantics (the same truncate-vs-ceil/decrement-correctly shape #1825 fixed for `nth`/`limit`).

```console
$ succinctly jq -n 'skip(0.4; 1,2,3,4)'
1
2
3
4
```

Truncated to skip 0.

## Where

`builtin_skip` (`src/jq/eval.rs`, ~line 10451) — the buggy `if f < 0.0 { error } else { num.as_i64().map_or(f as usize, ...) }` classification is duplicated inline in both its `QueryResult::One` and `QueryResult::Owned` arms.

## Suggested fix

Extract a shared classifier the same way `classify_limit_n`/`classify_nth_n` already exist for `limit`/`nth` (#1313/#1825's own precedent), implementing jq's real three-way `if $n > 0 / elif $n == 0 / else error` branch with NaN correctly falling to the `else` (error) arm, and a positive non-integer `$n` handled via jq's actual per-iteration decrement semantics (verify the exact take-count formula live against jq 1.7.1 before assuming it matches `limit`'s `ceil($n)` — `skip`'s own definition decrements differently and should be independently confirmed, not assumed identical).

## Provenance

Found during multi-angle `/code-review` of PR #1842 (#1825).

### LINKED ISSUE #1849: jq: skip(n; expr) doesn't raise on NaN count, likely truncates instead of ceiling a fractional count
## Summary

`builtin_skip` (`src/jq/eval.rs`, ~line 10510) implements `skip($n; expr)` from the same #983 lineage as `limit`/`nth` (branch `issue-983-limit-nth-skip-negative-count`), and its own comment already quotes jq's real definition:

```
if $n > 0 then ... elif $n == 0 then expr else error("skip doesn't support negative count") end
```

#1825 fixed the equivalent NaN/fractional-count bugs for `limit`/`nth` (`classify_limit_n`/`classify_nth_n`), but `skip` was out of scope for that PR and still has both bug classes today, verified against the current code's own quoted spec:

1. **NaN silently bypasses the error branch.** The guard is `if f < 0.0 { return Err(...) }` — false for NaN, so NaN satisfies neither `$n > 0` nor `$n == 0` under the *comment's own* semantics (both real-number comparisons are false for NaN), which should fall to `else error(...)`. Instead, `num.as_i64().map_or(f as usize, |i| i as usize)` runs on a NaN `f`; Rust's saturating float→int cast turns NaN into `0`, so `skip(nan; expr)` silently computes `n = 0` (skip nothing) instead of raising. Confirmed live:
   ```
   $ ./target/release/succinctly jq -n 'skip(nan; 1,2,3,4)'
   1
   2
   3
   4
   ```
   (exit 0, no error — should raise per the comment's own quoted contract, matching how `classify_limit_n`/`classify_nth_n` now reject NaN for `limit`/`nth`.)

2. **A positive fractional count truncates instead of ceiling.** Both `QueryResult::One`/`QueryResult::Owned` arms end with a plain truncating cast (`num.as_i64().map_or(f as usize, |i| i as usize)` / `f as usize`), not the `ceil`-based per-item-decrement count `limit`/`nth` now use. `skip(1.5; 1,2,3,4)` currently returns `2,3,4` (skips 1 item), whereas by analogy with `limit`'s real per-item decrement semantics (same PR lineage) skip should very likely skip `ceil(1.5) = 2` items and return `3,4`.

## Verification caveat

`skip/2` is a jq **1.8+** builtin — not defined in this repo's pinned oracle (`/usr/bin/jq` 1.7.1: `skip/2 is not defined`), and no jq 1.8 binary is available on this machine to verify live. Item 1 (NaN bypass) is high-confidence since it directly contradicts the `if/elif/else` structure the existing code comment already quotes as jq's real definition. Item 2 (fractional truncation) is a strong analogy to `limit`/`nth`'s confirmed decrement-based semantics but should be confirmed against a real jq 1.8 binary before changing the ceiling behavior, per this repo's rule against stating jq/yq behavior from memory (see CLAUDE.md's reference-fidelity section).

## Suggested fix

Once verified against jq 1.8: extract a `classify_skip_n` sharing `classify_limit_n`'s float/NaN branch structure (and its `ceil_positive_float_to_usize` helper, added in #1825), replacing this function's independent, unguarded `f < 0.0` check.

Found during #1825's third code-review round (Angle "Altitude + Conventions" scan) while reviewing `classify_limit_n`/`classify_nth_n` — flagged as a sibling bug in the same builtin family, out of scope for that PR's minimal fix.

newhoggy:
## Summary

`skip(nan; expr)` satisfies neither `$n > 0` nor `$n == 0` under real jq's
own generator-based definition, so it falls to `else error("skip doesn't
support negative count")`. succinctly's Rust `f as usize` cast instead
saturated NaN to `0`, silently skipping nothing (`skip(nan; 1,2,3,4)` used
to print `1 2 3 4` with exit 0).

```console
$ succinctly jq -n 'skip(nan; 1,2,3,4)'
jq: error (at <unknown>): skip doesn't support negative count
```

## Important correction to both #1846 and #1849

Both issues suspected a *second* bug: that a positive fractional count
truncates instead of ceiling ("like #1825 fixed for `nth`/`limit`"). This
turns out to be wrong. `skip/2` postdates the pinned jq 1.7.1 oracle (no
live repro is possible), so I hand-traced jq's real, published
`builtin.jq` definition:

```jq
def skip($n; expr):
  if $n > 0 then foreach expr as $item ($n; . - 1; if . < 0 then $item else empty end)
  elif $n == 0 then expr
  else error("skip doesn't support negative count") end;
```

Unlike `limit`/`nth`'s "take `ceil($n)`" formula, `skip`'s decrement-per-item
generator emits the first item once `$n - k < 0`, i.e. `k > $n` — so it
skips exactly `floor($n)` items for a non-integer `$n`. The existing
truncating cast already matched this:

- `skip(1.5; 1,2,3,4)` → `2,3,4` (skips `floor(1.5) = 1`, not `ceil(1.5) = 2`)
- `skip(0.4; 1,2,3,4)` → `1,2,3,4` (skips `floor(0.4) = 0`)
- `skip(2.5; 1,2,3,4)` → `3,4` (skips `floor(2.5) = 2`)

So the fix here is narrower than either issue proposed: **only** the NaN
case needed fixing. Adding a ceiling would have been a regression.

## Changes

- New `classify_skip_n(f: f64) -> Result<usize, EvalError>` in `src/jq/eval.rs`,
  shared by `builtin_skip`'s `QueryResult::One`/`QueryResult::Owned` arms —
  mirrors the #1825/#1313 `classify_limit_n`/`classify_nth_n` precedent for
  eliminating hand-inlined, drift-prone duplication. Non-numeric `$n`
  type-error handling in both arms is left unchanged (out of scope, same
  as #1825's treatment of `limit`).
- Unit tests in `src/jq/eval.rs` covering `classify_skip_n` directly (NaN,
  positive/negative fractional, zero/-0.0).
- CLI-level tests in `tests/jq_cli_tests.rs` (`run_jq_full` against the
  built binary) covering NaN, fractional floor behaviour via both the
  document-sourced (`QueryResult::One`) and computed-literal
  (`QueryResult::Owned`) arms.

`Builtin::Skip` has no native dispatch arm in `eval_generic.rs` — it's
bridge-only into `eval.rs`, so this fix is CLI-reachable without any
`eval_generic.rs` changes (confirmed via grep; this is the systemic gap
#1829/#1830 found for `paths`/`leaf_paths`, not present here).

## Verification

- `cargo test --features cli,simd,regex,serde --workspace`: all green
  (3047 lib tests + all integration suites; one `--release`-mode-only
  failure in an unrelated `should_panic`/`debug_assert!` test was
  confirmed to be a release-build artifact, not a regression, by
  re-running it in debug mode alone)
- `cargo clippy --all-targets --all-features -- -D warnings`: clean
- `cargo clippy --all-targets --features std,simd,broadword-yaml,serde,cli,regex,bench-runner,large-tests,mmap-tests -- -D warnings`: clean
- `cargo check --no-default-features` (no_std): clean
- `cargo doc --no-deps --all-features`: clean
- `cargo fmt --check`: clean

Fixes #1846

Also closes #1849 as a duplicate (filed independently ~29 minutes after
#1846, same root cause, same location).

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -3487,10 +3487,17 @@ pub(crate) fn classify_nth_n(n_owned: OwnedValue) -> Result<usize, EvalError> {
 /// copies can't independently drift the way `nth`/`limit`'s once did (#1313,
 /// the reason [`classify_limit_n`]/[`classify_nth_n`] exist).
 ///
-/// Takes the already-extracted `f64` rather than an `OwnedValue`/`StandardJson`
-/// -- unlike `classify_limit_n`/`classify_nth_n`, both call sites already do
-/// their own (unchanged, out of scope here) type check on `$n` before this
-/// runs, so there is no non-numeric case for this function to classify.
+/// Takes an `OwnedValue` and matches `Int`/`NumberLiteral(Int)` directly,
+/// same as `classify_limit_n`/`classify_nth_n` -- not a pre-extracted `f64`
+/// (an earlier version of this function did that, and a #1879 review caught
+/// it: converting a document-sourced integer to `f64` *before* classifying
+/// silently rounds any magnitude past `f64`'s 53-bit mantissa, an exactness
+/// regression against the `f as usize`-truncation-was-already-correct
+/// conclusion below, and against this codebase's `NumberLiteral`/#387
+/// large-integer discipline generally). The `QueryResult::One` call site
+/// converts its `StandardJson::Number` to an `OwnedValue` via [`to_owned`]
+/// before calling this, rather than extracting `f64` itself, for the same
+/// reason.
 ///
 /// Real jq's own definition (`builtin.jq`, confirmed against jq's current
 /// source -- `skip/2` postdates the pinned 1.7.1 oracle, so a live repro
@@ -3516,11 +3523,22 @@ pub(crate) fn classify_nth_n(n_owned: OwnedValue) -> Result<usize, EvalError> {
 /// `0.4 -> -0.6`, first emitting on item 1, i.e. skips 0 (`floor(0.4)`).
 /// #1846/#1849 both suspected this should ceiling like `limit`/`nth` instead
 /// -- it should not; adding a ceiling here would be a regression, not a fix.
-fn classify_skip_n(f: f64) -> Result<usize, EvalError> {
-    if f.is_nan() || f < 0.0 {
-        Err(EvalError::new("skip doesn't support negative count"))
-    } else {
-        Ok(f as usize)
+fn classify_skip_n(n_owned: OwnedValue) -> Result<usize, EvalError> {
+    match n_owned {
+        OwnedValue::Int(i) | OwnedValue::NumberLiteral(NumberRepr::Int(i), _) if i >= 0 => {
+            Ok(i as usize)
+        }
+        OwnedValue::Int(_) | OwnedValue::NumberLiteral(NumberRepr::Int(_), _) => {
+            Err(EvalError::new("skip doesn't support negative count"))
+        }
+        OwnedValue::Float(f) | OwnedValue::NumberLiteral(NumberRepr::Float(f), _) => {
+            if f.is_nan() || f < 0.0 {
+                Err(EvalError::new("skip doesn't support negative count"))
+            } else {
+                Ok(f as usize)
+            }
+        }
+        other => Err(EvalError::type_error("number", other.type_name())),
     }
 }
 
@@ -10673,9 +10691,8 @@ fn builtin_skip<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     let n_result = eval_single::<W, S>(n_expr, value.clone(), optional);
     let n = match n_result {
         QueryResult::One(v) => {
-            if let StandardJson::Number(num) = v {
-                let f = num.as_f64().unwrap_or(0.0);
-                match classify_skip_n(f) {
+            if let StandardJson::Number(_) = v {
+                match classify_skip_n(to_owned(&v)) {
                     Ok(n) => n,
                     Err(e) => return QueryResult::Error(e),
                 }
@@ -10684,14 +10701,11 @@ fn builtin_skip<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
             }
         }
         QueryResult::Owned(
-            ref owned @ (OwnedValue::Int(_) | OwnedValue::Float(_) | OwnedValue::NumberLiteral(..)),
-        ) => {
-            let f = owned.as_f64().unwrap_or(0.0);
-            match classify_skip_n(f) {
-                Ok(n) => n,
-                Err(e) => return QueryResult::Error(e),
-            }
-        }
+            owned @ (OwnedValue::Int(_) | OwnedValue::Float(_) | OwnedValue::NumberLiteral(..)),
+        ) => match classify_skip_n(owned) {
+            Ok(n) => n,
+            Err(e) => return QueryResult::Error(e),
+        },
         QueryResult::Error(e) => return QueryResult::Error(e),
         _ => return QueryResult::Error(EvalError::type_error("number", "null")),
     };
@@ -57405,7 +57419,7 @@ mod tests {
 
     #[test]
     fn test_classify_skip_n_nan_errors_1846() {
-        assert!(classify_skip_n(f64::NAN).is_err());
+        assert!(classify_skip_n(OwnedValue::Float(f64::NAN)).is_err());
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


newhoggy:
Code review found two real issues, both fixed in 4fe14293b:

1. **Precision loss**: routing both `builtin_skip` call sites through a pre-extracted `f64` silently rounded a document-sourced integer past `f64`'s 53-bit mantissa — the original `num.as_i64()` fallback didn't have this problem. `classify_skip_n` now takes an `OwnedValue` and matches `Int`/`NumberLiteral(Int)` directly, same as `classify_limit_n`/`classify_nth_n`.
2. **Misleading test comment**: `test_skip_number_literal_n_387` doesn't actually cover the `QueryResult::One` arm (it forces the `Owned` arm via `getpath`, per its own comment). Added dedicated `One`-arm integer/fractional tests via bare `.n` field access, and corrected the comment.

Re-ran the full verification suite (tests, both clippy legs, no_std, doc, fmt) — all green.

newhoggy:
Round 2 (8 finder agents) surfaced two more real items, both fixed in 9d87d70cd:

1. **Unnecessary allocation**: the \`QueryResult::One\` call site's \`classify_skip_n(to_owned(&v))\` routed through \`from_number_bytes\`, which re-validates the digit span and boxes a copy of the source text into \`OwnedValue::NumberLiteral\` — work \`classify_skip_n\` immediately discards. Now builds the \`OwnedValue\` directly from \`num.as_i64()\`/\`as_f64()\` (same exact-integer-first precision guarantee, zero allocation).
2. **Cross-file comment reference**: a test comment pointed to \`test_skip_number_literal_n_387\` as if it were a neighbor in the same file; it's actually a unit test in \`src/jq/eval.rs\`. Fixed to say so explicitly.

Two independent finders also flagged (converging on the same underlying concern) that \`classify_skip_n\`'s floor-not-ceil conclusion was verified by hand-tracing jq's published \`builtin.jq\` source rather than a live repro against the pinned oracle, since \`skip/2\` doesn't exist in the pinned jq 1.7.1 binary at all. I filed #1880 to track re-verifying this once a jq 1.8+ oracle is available, and cited it from the doc comment — this isn't something I can resolve within this PR (no such oracle exists yet), but it's now tracked rather than silently asserted.

Re-ran the full verification suite (tests, both clippy legs, no_std, doc, fmt) — all green.

newhoggy:
Round 3 (10 finder agents) — no new correctness bugs. Summary of disposition:

**Fixed** (e71fbd23a):
- Upgraded the floor-not-ceil verification from a source hand-trace to a live check against Homebrew jq 1.8.2 (installed on this machine, though not the pinned 1.7.1 oracle) — every case matches exactly (fractional floor, NaN error, negative error, large-integer count). Doc comment updated accordingly, redundant worked-examples prose trimmed since the unit tests already assert those cases.

**Filed as separate follow-ups** (correctly out of scope for this narrow NaN-fix PR):
- #1882: `skip` is missing from the `--jq-extensions` gate in yq mode, unlike its `limit`/`nth` siblings (real yq's lexer rejects `skip` entirely) — pre-existing since #983, unrelated to this PR's changes.
- #1880 (already filed, updated with the live 1.8.2 confirmation): re-verify once the *pinned* oracle itself reaches 1.8+.

**Investigated and confirmed not bugs**:
- "Missing `resolve_skip` path-context resolver" — disproven live by a sibling finder in the same round: `skip` falls through to the same fixed `builtin_skip` via the generic leaf resolver, confirmed with `.a |= skip(nan; .[])` erroring correctly and `.a |= skip(1.5; .[])` flooring correctly.
- Triplicated `Int`/NaN-guard shape across `classify_limit_n`/`classify_nth_n`/`classify_skip_n` — same conclusion as round 2: real but cosmetic (each function's *policy* after the shared shape genuinely differs — Unlimited vs. error vs. floor vs. ceil), not a correctness or drift risk worth a parameterized-helper refactor.
- A behavior change on genuinely malformed number spans (both `as_i64`/`as_f64` parse failure) now errors instead of silently defaulting to 0 — net improvement (fail loud vs. silently wrong), not a regression.

Full verification suite (tests, both clippy legs, no_std, doc, fmt) green after each change. Proceeding to merge.

PR #1886, opened 2026-08-30
fix(jq): Expr::Iterate's path-context arm no longer self-swallows per-element errors

### LINKED ISSUE #1869: jq: Expr::Iterate's path-context arm has the same ambient-optional self-swallow bug #1826 fixed for Map
## Summary

Found during code review on #1826/PR #1866 (which fixed the same bug class for `Builtin::Map(f)`'s path-context arm). `Expr::Iterate`'s own path-context arm (`src/jq/eval.rs`, `eval_pipe_with_path_context_internal`, the arm handling bare `.[]` when `rest` needs `key`/`parent`/`file_index`/`document_index`) has the identical bug: it passes the *ambient* `optional` straight into each element's own evaluation via `iterate_element_step`, instead of forcing `false` and self-catching the way the sibling `Expr::Array` arm (#1302) and now `Builtin::Map`'s arm (#1826/#1866) both do.

This lets a genuine per-element error self-swallow into "no output for this element" instead of surfacing as a real stop signal — silently skipping the erroring element and continuing to the next one, rather than aborting the whole `.[]` traversal (or at minimum surfacing a `Partial` with the correct prefix+error, matching how a bare `.[]` with no ambient optional already behaves).

## Repro (confirmed live against current `main`)

```console
$ echo '{"a":[1,2,3]}' | succinctly jq -c '(.a)? | .[] | if key==1 then error("boom") else key end'
0
2
```

Real jq has no `key` builtin (this is a succinctly-only path-context extension), so there's no direct oracle comparison for this exact filter — but the internal inconsistency is clear: without the ambient `?`, the same query correctly hard-errors:

```console
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a | .[] | if key==1 then error("boom") else key end'
0
```

Same class of bug for the object case:

```console
$ echo '{"a":{"x":1,"y":2,"z":3}}' | succinctly jq -c '(.a)? | .[] | if key=="x" then error("boom") else key end'
"y"
"z"
```

## Scope note

Unlike `Expr::Array`/`Builtin::Map`, `Expr::Iterate` is **not** array-construction — it's a genuine multi-output generator, so the correct fix here is probably closer to `eval_try`'s/the path-context `Expr::Optional` arm's "keep-prefix" semantics (an error partway through a `.[]` traversal should surface as `Partial(prefix, Error)`, preserving already-produced outputs, not "atomic discard-everything" the way array construction is) rather than `Expr::Array`'s atomic-discard model. This needs its own careful design/verification pass, not a copy-paste of the Map/Array fix shape — filed separately rather than folded into #1866 to keep that PR properly scoped.

## Related

- #1826 / PR #1866 — fixed the identical bug class for `Builtin::Map(f)`'s path-context arm; this issue is the `Expr::Iterate` sibling, found but deliberately not fixed there.
- #1302 — the original fix for `Expr::Array`'s path-context arm, the atomic-discard precedent.
- #1335 — the path-context `Expr::Optional` arm's own "keep-prefix" precedent, likely the closer semantic model for this issue's fix.

newhoggy:
Fixes #1869.

`Expr::Iterate`'s own path-context arm (`eval_pipe_with_path_context_internal`,
the arm handling bare `.[]` when `rest` needs `key`/`parent`/`file_index`/
`document_index`) passed the *ambient* `optional` straight into each
element's own evaluation via `iterate_element_step`, instead of forcing
`false` and self-catching the way the sibling `Expr::Array` (#1302) and
`Builtin::Map` (#1826) arms already do.

This let a genuine per-element error self-swallow into "no output for this
element" instead of surfacing as a real stop signal — silently skipping the
erroring element and continuing to the next one, rather than aborting the
whole `.[]` traversal.

## Repro (confirmed live against current `main`, pre-fix)

```console
$ echo '{"a":[1,2,3]}' | succinctly jq -c '(.a)? | .[] | if key==1 then error("boom") else key end'
0
2
```

Real jq has no `key` builtin (a succinctly-only path-context extension), so
there's no direct oracle comparison for this exact filter — but the
internal inconsistency was clear: without the ambient `?`, the same query
correctly hard-errors and stops:

```console
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a | .[] | if key==1 then error("boom") else key end'
0
```

## Fix

Unlike `Array`/`Map`, `Iterate` is not a constructor — there's no atomic
in-progress collection to discard wholesale, so the fix keeps whatever
prefix was already produced before the error (matching the path-context
`Expr::Optional` arm's own keep-prefix precedent, #1335) rather than
discarding it to nothing the way `map()`'s own fix does:

```console
$ echo '{"a":[1,2,3]}' | succinctly jq -c '(.a)? | .[] | if key==1 then error("boom") else key end'
0
```

`Break`/`Halt` are never caught by this arm's ambient `optional`, even when
it's `true` — unlike `Expr::Optional`'s own arm (which only ever runs when
there's a literal `?` right there, so unconditional catching is correct),
an ambient `optional` reaching `Iterate` can come from an *unrelated*
enclosing `?` bleeding through `rest` (a separate, pre-existing, documented
#1335 gap), and a `break`'s label target is lexically scoped — never
something an unrelated ancestor's `?` may intercept. Verified live against
jq 1.7.1 that `try`/`?` does catch `break` when it's directly around the
break-containing expression (a real, if surprising, jq behavior — confirmed
before relying on it), which is exactly what the pre-existing `Expr::Optional`
arm already implements; this fix only changes `Iterate`'s own response when
the ambient flag reaches it via an unrelated stage.

## Testing

Added `test_iterate_path_context_ambient_optional_keeps_prefix_not_skip_1869`
covering: array/object targets, first-element-errors (empty-prefix collapse,
the same gap #1826's review caught for `map()`), no-ambient-optional control,
`break` never caught (with and without ambient optional), and the
never-errors positive case.

`cargo doc --all-features`. yq mode sanity-checked with `--jq-extensions`
(consistent behavior, all 1298 yq CLI tests unaffected).

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -26760,7 +26760,42 @@ fn eval_pipe_with_path_context_internal<'a, W: Clone + AsRef<[u64]>, S: EvalSema
             // native and un-boxed rather than unified into one iterator, to
             // avoid trading this duplication for a `Box<dyn Iterator>`'s
             // allocation and dynamic dispatch on every `.[]`.
+            //
+            // Each element's own step is run with `optional` forced to
+            // `false`, not the ambient value this arm itself received --
+            // the same `Expr::Array`/`Builtin::Map` fix (#1302/#1826),
+            // applied here for `Expr::Iterate` (#1869): passing the ambient
+            // value straight through let a genuine per-element error
+            // self-swallow into `QueryResult::None` before
+            // `accumulate_path_context_step` ever saw it as a stop signal,
+            // silently *skipping* that one element and continuing to the
+            // next instead of aborting the traversal -- confirmed live:
+            // `(.a)? | .[] | if key==1 then error("boom") else key end` on
+            // `{"a":[1,2,3]}` produced `0`/`2` (element 1 silently
+            // skipped), where real jq's own equivalent (`.a | .[] | ...`,
+            // no `?` -- an ordinary generator error always aborts
+            // everything downstream, `?` or not) emits `0` then errors.
+            //
+            // Unlike `Array`/`Map`, `Iterate` is not a constructor -- there
+            // is no atomic "discard the whole in-progress collection"
+            // model to reach for, so an escaping `Error` (when the ambient
+            // `optional` genuinely catches it) keeps the prefix already
+            // produced rather than discarding it (`owned_vec_to_result`,
+            // matching the path-context `Expr::Optional` arm's own
+            // keep-prefix precedent a few arms below, #1335). `Break`/
+            // `Halt` are never caught by this arm's ambient `optional` at
+            // all, even when it's `true` -- unlike `Optional`'s own arm
+            // (which only ever runs when there's a literal `?` right here,
+            // so unconditional catching is correct there), an ambient
+            // `optional` reaching `Iterate` can come from an *unrelated*
+            // enclosing `?` bleeding through `rest` (the pre-existing,
+            // documented #1335 "rest inherits suppression" gap) -- and a
+            // `break`'s label target is lexically scoped, never something
+            // an unrelated ancestor's `?` may intercept. Both fall through
+            // `Some(other) => other` unchanged, preserving whatever prefix
+            // `accumulate_path_context_step` already attached (#400/#494).
             let mut results = Vec::new();
+            let mut stopped = None;
             match value {
                 OwnedValue::Array(arr) => {
                     for (i, v) in arr.iter().enumerate() {
@@ -26771,10 +26806,11 @@ fn eval_pipe_with_path_context_internal<'a, W: Clone + AsRef<[u64]>, S: EvalSema
                             root,
                             file_origin,
                             current_path,
-                            optional,
+                            false,
                             &mut results,
                         ) {
-                            return stop;
+                            stopped = Some(stop);
+                            break;
                         }
                     }
                 }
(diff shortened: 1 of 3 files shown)


newhoggy:
Ran the mandatory 4-agent code review. Summary of findings and what I did about each:

**Real issue, confirmed by 3 independent agents (fixed):** my own doc comment
overstated what the fix achieves. The PR's headline repro
(`(.a)? | .[] | if key==1 then error("boom") else key end`) does now correctly
*stop* the traversal instead of skipping-and-continuing, but it still doesn't
surface an error at all (exits 0) — because `key` routes through the
pre-existing #1335 "ambient optional bleeds through `rest`" fallback, which
this fix was never meant to close. Live-verified this is **not a regression
and not unique to Iterate**: `Array`'s own #1302 fix and `Map`'s own #1826 fix
have the byte-identical characteristic on the same input. Rewrote the doc
comment to say plainly what the fix does (brings `Iterate`'s response to
`optional` in line with its siblings) and doesn't do (doesn't re-litigate
whether `optional` should have reached here at all — that's #1335's own open
question).

**Reuse finding, confirmed by 2 independent agents (filed as follow-up, not
fixed here):** the "catch Error under ambient optional, keep-or-discard
prefix" post-loop dispatch is now hand-copied a 4th time across
`Optional`/`Map`/`Array`/`Iterate`, each cross-referencing the others in
comments rather than sharing a function. Filed #1888 — unifying three
already-shipped, independently-tested arms is its own refactor, out of scope
for this narrow bug fix.

**Simplification + redundant-rebuild nit, confirmed by 2 independent agents
(fixed):** the post-loop `match` used nested `if optional {A} else {B}` where
`B` in the `Partial` arm rebuilt a value provably identical to the original.
Rewrote as match-guard arms mirroring `Expr::Array`'s own established idiom —
cleaner and removes the redundant reconstruction.

**Coverage gap, confirmed by 1 agent (fixed):** no yq-mode test despite the
underlying code being fully shared (`eval_pipe_with_path_context_internal` is
generic over `EvalSemantics`). Added
`test_yq_iterate_path_context_ambient_optional_keeps_prefix_1869`, confirming
the same behavior via `--jq-extensions`.

**Everything else** (Array/Object loop symmetry, `needs_path_context` scoping,
`accumulate_path_context_step`'s invariant, `eval_generic.rs`'s parallel
demand-driven sinks) checked out correct with no action needed — see the
individual finder reports for the live-verification detail.

Full local verification re-run after these fixes: build/test/clippy under
both default and full (`cli,simd,regex,serde`) feature sets, plus
`--no-default-features` and `cargo doc --all-features`. All green, including
the new tests.

PR #1893, opened 2026-08-30
fix(jq): string-slice terminal write runs the update filter before refusing

### LINKED ISSUE #1876: jq: string slice's terminal write refuses without running edit's side effects first
**Severity: Low** (only observable when the update filter through a string-slice write has a side effect or its own error)

## Summary

`through_slice`'s `OwnedValue::String(_)` terminal arm (`src/jq/eval.rs`, `terminal_write: true` case) returns `Err(cannot_update_string_slices())` immediately, without ever calling `edit` — unlike every other arm in the function, which always runs `edit` first and only then decides what to do with the result.

Real jq evaluates the update filter's side effects before rejecting a string-slice write.

```
$ echo '"hello"' | jq '.[0:2] |= (debug|9)'
["DEBUG:","he"]
jq: error (at <stdin>:0): Cannot update string slices

$ echo '"hello"' | succinctly jq '.[0:2] |= (debug|9)'
jq: error (at <stdin>:0): Cannot update string slices
(no debug output on stderr)
```

```
$ echo '"hello"' | jq '.[0:2] |= error("boom")'
jq: error (at <stdin>:0): boom

$ echo '"hello"' | succinctly jq '.[0:2] |= error("boom")'
jq: error (at <stdin>:0): Cannot update string slices
(the "boom" error() call never fires)
```

succinctly silently discards the filter's `debug` output and swallows its own `error()` call,
reporting the generic string-slice message in both cases instead.

## Discovery context

Found live during #1873's code review (`through_slice`'s `Null` arm neighbor) — confirmed
identical on `origin/main` before #1873, not a regression. #1873's own new `Null`-arm no-op
depends on `edit` always running first so its side effects/errors are the real answer; this
sibling `String` terminal arm violates that same invariant by skipping `edit` entirely.

## Scope note

A fix likely means running `edit` against a throwaway substring (mirroring the `!terminal_write`
`String` arm just above it, which already does this) before deciding whether to raise
`cannot_update_string_slices` — propagating any error/side-effect `edit` produces, and only
falling back to the generic message if `edit` itself succeeds. Filing to make this trackable, not
blocking.

### LINKED ISSUE #1883: jq: a compound assign through a string slice reports `Cannot update string slices` where jq reports the arithmetic error
**Severity: Low** (error-message wording only — exit code and failure/success are already correct)

## Summary

A compound assignment through a slice of a *string* reports succinctly's own
`Cannot update string slices` where real jq reports the underlying arithmetic
failure. Both raise, both exit 5; only the message differs.

## Repro (jq 1.7.1, confirmed live)

```console
$ echo '"s"' | jq -c '.[0:1] += 1'
jq: error (at <stdin>:1): string ("s") and number (1) cannot be added

$ echo '"s"' | succinctly jq -c '.[0:1] += 1'
```

Same for `-=` (`cannot be subtracted`).

## Status

The unparenthesized spelling above has always behaved this way. The
parenthesized form `(.[0:1]|.) += 1` used to produce jq's message by accident:
`update_path`'s slice arm hardcoded `terminal_write: false`, which skipped
`through_slice`'s `OwnedValue::String` terminal refusal and let the edit run,
surfacing the arithmetic error instead.

#1428 made that flag accurate (it is documented as "whether this call's own
`edit` closure *is* the write", and for an `Identity` tail it is), because the
wrong value had started swallowing a genuine error once `through_slice`'s
`Null` arm learned to no-op for read-throughs. As a side effect the two
spellings now agree — on succinctly's message rather than jq's.

So this is not new behaviour so much as newly *consistent* behaviour, and the
divergence is now visible in one place rather than two.

## Root cause direction

`through_slice`'s `OwnedValue::String if terminal_write` arm
([src/jq/eval.rs](src/jq/eval.rs)) refuses before the edit runs. Real jq
evaluates the update first and reports whatever that produces; only a
successful non-string result then hits the "cannot update string slices"
refusal. Reordering (attempt the edit against the sliced string, refuse only if
it would have succeeded) would match jq for both spellings — but that arm's
current shape is load-bearing for #1303, so the interaction wants checking
before it is moved.

## Why this is separate from #1428

#1428 is about autovivification surviving a write that never happens. This is
the wording of an error on a path that correctly fails either way, on a
*string* target, and it predates #1428 for the spelling users actually write.

newhoggy:
Fixes #1876, Fixes #1883.

Bundled together (same root cause: `through_slice`'s `OwnedValue::String`
terminal-write arm in `src/jq/eval.rs`, both found via #1873's own code
review).

## Root cause

`through_slice`'s `OwnedValue::String` terminal-write arm refused with the
generic `Cannot update string slices` message immediately, without ever
running the update filter (`edit`) — unlike its `!terminal_write` sibling
just above it (and every other arm in this function), which always runs
`edit` first and only then decides what to do with the result.

Real jq evaluates the update filter's side effects and surfaces its own
errors *before* refusing the write.

## Repros (confirmed live against jq 1.7.1)

```console
$ echo '"hello"' | jq '.[0:2] |= (debug|9)'
["DEBUG:","he"]
jq: error (at <stdin>:0): Cannot update string slices
$ echo '"hello"' | succinctly jq '.[0:2] |= (debug|9)'   # pre-fix: no debug output at all
jq: error (at <stdin>:0): Cannot update string slices

$ echo '"hello"' | jq '.[0:2] |= error("boom")'
jq: error (at <stdin>:0): boom
$ echo '"hello"' | succinctly jq '.[0:2] |= error("boom")'   # pre-fix: "boom" never fires
jq: error (at <stdin>:0): Cannot update string slices

$ echo '"s"' | jq -c '.[0:1] += 1'
jq: error (at <stdin>:1): string ("s") and number (1) cannot be added
$ echo '"s"' | succinctly jq -c '.[0:1] += 1'   # pre-fix: wrong message
```

## Fix

Mirrors the `!terminal_write` arm: run `edit` against the same throwaway
substring, propagate whatever it produces via `?`, and only raise the
refusal once `edit` itself has succeeded (matching the `debug` case above,
where the refusal is still correct — only the *ordering* relative to `edit`
changed).

Confirmed this doesn't reach yq mode at all: yq's string-slice writes
(`=`/`+=`/`-=`/`*=`) always route through the separate `OwnedValue::String
if container_noop` arm a few lines above, which was already correct and
unaffected either way (verified byte-identical against pre-fix `main` for
`+=`/`-=`/`=` on a YAML string-scalar slice).

## Testing

Added `test_string_slice_terminal_write_runs_edit_before_refusing_1876_1883`
covering all repros above plus a control (an update filter that succeeds
with an ordinary value still gets refused, just after running).

`cargo doc --all-features`. All existing slice-related tests (40) pass
unaffected.

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -17443,40 +17443,38 @@ fn through_slice<E: From<EvalError>>(
             Ok(())
         }
         // jq reads a string slice but will not write one back, whatever the
-        // replacement — the refusal beats the non-array one above. But
-        // that refusal is only correct once the slice itself *is* the
-        // write; a chained path with more path after the slice
-        // (`terminal_write: false`) needs the read-through substring
-        // handed to `edit` instead, so *that* navigation's own outcome —
-        // a genuine failure (`del(.a[0:1].b)` -> "Cannot index string
-        // with string \"b\"", matching `.a[0:1].b = 5`'s identical real-jq
-        // message) or a `?`-suppressed no-op (`del(.a[0:1][0]?)`,
-        // `.a[0:1].b? = 5`) — is the answer, not this refusal (#1321,
-        // confirmed live both ways against real jq for `del()` and `=`).
-        OwnedValue::String(_) if !terminal_write => {
-            let mut throwaway = slice_owned_value(root, start, end, optional)?
-                .expect("Array/String target always yields Some from slice_owned_value");
-            edit(&mut throwaway)
-        }
-        // #1876/#1883: real jq evaluates the update filter's side effects
-        // (and surfaces its own errors) *before* refusing the write --
-        // this arm used to refuse immediately, discarding `edit` entirely,
-        // so a `debug`/`stderr` side effect never fired and a genuine
-        // `error(...)`/arithmetic type error inside the update filter was
-        // replaced by this generic message instead of propagating.
-        // Mirrors the `!terminal_write` arm just above: run `edit` against
-        // the same throwaway substring, propagate whatever it produces via
-        // `?`, and only *then* raise the refusal once `edit` itself has
-        // succeeded (confirmed live against jq 1.7.1: `"hello" | .[0:2] |=
-        // (debug|9)` prints the debug line before still refusing with this
-        // same message; `"hello" | .[0:2] |= error("boom")` raises "boom"
-        // instead; `"s" | .[0:1] += 1` raises the real arithmetic error,
-        // not this one).
+        // replacement -- but that refusal (`terminal_write: true`) is only
+        // correct once the slice itself *is* the write; a chained path with
+        // more path after the slice (`terminal_write: false`) needs the
+        // read-through substring handed to `edit` instead, so *that*
+        // navigation's own outcome -- a genuine failure (`del(.a[0:1].b)`
+        // -> "Cannot index string with string \"b\"", matching
+        // `.a[0:1].b = 5`'s identical real-jq message) or a
+        // `?`-suppressed no-op (`del(.a[0:1][0]?)`, `.a[0:1].b? = 5`) -- is
+        // the answer, not the refusal below (#1321, confirmed live both
+        // ways against real jq for `del()` and `=`).
+        //
+        // Either way, `edit` always runs first against the same throwaway
+        // substring and its outcome is always propagated via `?` before
+        // either disposition is decided (#1876/#1883): an earlier version
+        // of the `terminal_write: true` case refused immediately without
+        // ever calling `edit`, discarding a `debug`/`stderr` side effect
+        // entirely and replacing a genuine `error(...)`/arithmetic type
+        // error inside the update filter with this generic refusal instead
+        // of propagating it (confirmed live against jq 1.7.1: `"hello" |
+        // .[0:2] |= (debug|9)` prints the debug line before *still*
+        // refusing with this same message; `"hello" | .[0:2] |=
+        // error("boom")` raises "boom" instead; `"s" | .[0:1] += 1` raises
+        // the real arithmetic error, not this one).
         OwnedValue::String(_) => {
             let mut throwaway = slice_owned_value(root, start, end, optional)?
                 .expect("Array/String target always yields Some from slice_owned_value");
             edit(&mut throwaway)?;
-            Err(EvalError::cannot_update_string_slices().into())
+            if terminal_write {
+                Err(EvalError::cannot_update_string_slices().into())
+            } else {
+                Ok(())
+            }
         }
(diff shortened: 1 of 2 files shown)


newhoggy:
Ran the mandatory code review (4 findings survived, with two independent
verification passes on the riskiest one). Addressed:

**Confirmed correct, no change needed (2 independent verification agents,
high confidence):** a reviewer hypothesized this fix could change behavior at
`del()`'s `Expr::Iterate` call site, which forces `terminal_write: yq_mode`
into the same arm this PR modified. Both agents independently built pre-fix
`main` and post-fix `HEAD`, ran an exhaustive sweep of `del()` shapes
(`del(.[][0:1].c)`, `del(.[][0:1][0])`, `del(.[][0:1][2:3])`, computed
bounds, etc.) against both, and found byte-identical output everywhere —
corroborated by a structural case analysis of `yq_del_slice_outcome`'s four
outcomes showing this call site can't reach the modified arm with a String
target and pending `rest`. No regression.

**Fixed:** the comment at that same call site made a now-false claim ("keeps
`through_slice`'s String arm on its old, unconditional refusal") — this PR's
fix means that arm no longer refuses unconditionally, it runs `edit` first.
Rewrote it to state what's actually believed true (edit is presumed to
always no-op/error before producing an observable difference here, not
proven structurally impossible) so a future change to `yq_del_slice_outcome`
has something accurate to re-verify against.

**Fixed:** the new terminal-write arm had become a near-duplicate of its
`!terminal_write` sibling (same setup, differing only in final `Ok`/`Err`).
Merged into one arm running the shared logic once, branching only on
`terminal_write` for the disposition — behavior-preserving (verified: `edit(&mut throwaway)`
is exactly `edit(&mut throwaway)?; Ok(())` via `?` propagation).

**Fixed:** tightened the `debug` test's assertion from a bare `"DEBUG:"`
substring check to the exact payload (`["DEBUG:","he"]`), so a regression in
what's handed to the update filter (e.g. the whole string instead of the
sliced substring) would actually fail the test.

**Filed as follow-up, not fixed here (confirmed pre-existing, same on
`main`):** #1894 — `through_slice`'s String/Array terminal-write arms don't
special-case an `empty`-yielding update filter the way real jq's `|=`
does (`.[0:2] |= empty` should delete the slice, not raise). This PR's fix
was a necessary prerequisite (an early unconditional refusal can't observe
`edit`'s output at all) but doesn't implement the special case itself, since
that needs a real protocol change to `through_slice`'s `edit` closure
signature, not a small addition to the arms this PR touches.

Full local verification re-run after these fixes and a rebase onto
`main` (which had advanced 8 commits): build/test/clippy under both default
and full (`cli,simd,regex,serde`) feature sets, plus `--no-default-features`
and `cargo doc --all-features`. All green.

PR #1899, opened 2026-08-30
fix(test): route jq_cli_tests.rs's remaining inline spawn sites through retry

### LINKED ISSUE #1889: test(jq): ~32 individually-converted sites in jq_cli_tests.rs have no spawn retry at all
**Severity: Low** (requires a transient spawn/signal-death race under heavy concurrent load, which this repo's own docs already treat as a routine, occasional occurrence)

## Summary

#1859/#1884 converted all 34 `Command::new("cargo").args(["run",...])` sites
in `tests/jq_cli_tests.rs` to exec the pre-built binary directly, and its
follow-up fix routed the file's shared helpers (`run_jq_full`,
`run_jq_stdin_streams`, `run_jq_null`, `run_jq_binary_stdin`, and 4 direct
call sites) through the shared `spawn_with_signal_retry`
(`tests/common/cargo_run_exit.rs`), closing a signal-death-retry gap code
review found.

That leaves roughly 32 individually-converted inline call sites (e.g.
`test_unary_minus`, `test_jq_help`, `test_args_positional`,
`test_no_color_env_var`, and the rest of the environment-variable/
module-loading/CLI-flag test cluster) doing a raw
`Command::new(env!("CARGO_BIN_EXE_succinctly"))...spawn()`/`.output()` with
**no retry at all** — not even `spawn_jq_full`'s old ENOENT-only retry, since
these never routed through that helper in the first place.

## Why this wasn't fixed in #1884

This is **not a new regression** — these sites never had retry protection
even in their pre-#1859 `cargo run` form (the old retry loop only lived in
the 5 shared helpers, never in these ~32 inline literal blocks). The
conversion substitutes one unretried race (cargo lock contention) for a
different unretried race (a transient `ENOENT` from a concurrent test
binary still linking the same `target/debug/succinctly` path, or a
signal-killed child under this repo's documented heavy concurrent-load
conditions) — not a new category of risk, and dormant absent that specific
timing.

Fixing all ~32 individually means restructuring each one into
`spawn_with_signal_retry`'s closure shape (some use `.output()`, some
`.spawn()` + stdin write, some add `.env(...)` calls) — meaningful,
non-mechanical work needing the same "each needs individual verification"
care #1859's own issue text called for, which is why it's being split out
rather than folded into that PR's already-large diff.

## Suggested approach

Same target shape as the already-fixed helpers: wrap each site's
`Command::new(env!(...))...` construction in a closure and call
`spawn_with_signal_retry(closure, stdin_input)`, adapting each call site's
existing assertions to the returned `(Output, i32)` tuple. Low priority —
file to make it trackable, not blocking.

newhoggy:
Fixes #1889.

## Summary

#1859/#1884 converted all \`Command::new("cargo").args(["run",...])\` sites in
\`tests/jq_cli_tests.rs\` to exec the pre-built binary directly, and a
follow-up routed the file's 5 shared helpers through the shared
\`spawn_with_signal_retry\` (\`tests/common/cargo_run_exit.rs\`). That left ~31
individually-converted inline call sites doing a raw
\`Command::new(env!("CARGO_BIN_EXE_succinctly"))...spawn()\`/\`.output()\` with
**no retry at all** — not even the old \`spawn_jq_full\` ENOENT-only retry,
since these never routed through that helper in the first place.

This is not a new regression (these sites never had retry protection even
in their pre-#1859 \`cargo run\` form) but a dormant race: a transient ENOENT
from a concurrently-relinking \`target/debug/succinctly\` (#550), or a
signal-killed child under this repo's documented heavy concurrent-load
conditions, would fail these ~31 tests outright on the first occurrence.

## Fix

Converted every remaining site to \`spawn_with_signal_retry\`:
- Most route through the existing \`spawn_jq(args, stdin_input)\` helper,
  where the call shape (jq subcommand, string args, optional stdin) fits.
- A few needing file-path args (\`NamedTempFile::path()\` returns \`&Path\`,
  not \`&str\`) or \`.env(...)\` get their own closure directly.
- \`jq_stderr\` (used by 6 tests) now delegates to \`spawn_jq\` instead of
  hand-rolling its own spawn+stdin+wait, picking up retry for all of its
  callers at once.
- \`test_as_pattern_alt_rejected_in_yq_mode_720\` already hand-rolled a
  correct signal-death diagnosis (#1546) but never retried on it; now
  shares the same diagnosis via the retry helper, gaining retry for free.

\`test_default_identity_filter\` is intentionally left unconverted: it never
calls \`wait_with_output()\` (drops the spawned \`Child\` immediately), so
there's no signal-death race to retry against, and routing it through the
shared helper would add a wait this near-vestigial test (it asserts nothing
about the process's outcome) was never designed to perform — documented in
place rather than silently changed.

## No behavior change

Every converted site's own assertions are unchanged. \`spawn_with_signal_retry\`'s
always-piped-stdout/stderr, null-or-piped-stdin defaults exactly match what
\`.output()\`'s own defaults and each site's own explicit \`.stdio(...)\` calls
already did.

## Testing

All 1103 tests in \`jq_cli_tests.rs\` pass (same count as before — purely
mechanical, no tests added/removed). Full local verification: build/test/clippy
under both default and full (\`cli,simd,regex,serde\`) feature sets, plus
\`--no-default-features\` and \`cargo doc --all-features\`.

### DIFF
--- a/tests/jq_cli_tests.rs
+++ b/tests/jq_cli_tests.rs
@@ -13,7 +13,8 @@ use tempfile::NamedTempFile;
 #[path = "common/cargo_run_exit.rs"]
 mod cargo_run_exit;
 use cargo_run_exit::{
-    classify_cargo_run_exit, spawn_with_signal_retry, succinctly_bin, MAX_CARGO_RETRIES,
+    classify_cargo_run_exit, signal_death_error, spawn_with_signal_retry, succinctly_bin,
+    MAX_CARGO_RETRIES,
 };
 
 /// #1516: a signal-killed child used to render as an inscrutable
@@ -242,6 +243,33 @@ fn run_jq_interleaved(args: &[&str], input: Option<&str>) -> Result<(String, i32
 }
 
 static INTERLEAVE_SEQ: std::sync::atomic::AtomicU32 = std::sync::atomic::AtomicU32::new(0);
+
+/// `spawn_jq`'s counterpart for the environment-variable test cluster
+/// (#1889 code review): the 8 tests covering `NO_COLOR`/`JQ_COLORS`/
+/// `JQ_LIBRARY_PATH`/`HOME` each independently hand-rolled the same
+/// `Command::new(succinctly_bin()); cmd.arg("jq").args(args).env(...)`
+/// closure since `spawn_jq` itself has no way to set an environment
+/// variable -- sharing one copy here instead of 8 near-identical ones is
+/// the same fix `spawn_jq`'s own doc comment already describes for the
+/// six-copies-of-one-`Command` problem (#1884).
+fn spawn_jq_with_env(
+    args: &[&str],
+    env_key: &str,
+    env_value: impl AsRef<std::ffi::OsStr>,
+    stdin_input: Option<&[u8]>,
+) -> Result<(std::process::Output, i32)> {
+    spawn_with_signal_retry(
+        || {
+            let mut command = Command::new(succinctly_bin());
+            command
+                .arg("jq")
+                .args(args)
+                .env(env_key, env_value.as_ref());
+            command
+        },
+        stdin_input,
+    )
+}
 
 /// Helper to run jq with null input (-n)
 fn run_jq_null(filter: &str, extra_args: &[&str]) -> Result<(String, i32)> {
@@ -3697,15 +3725,10 @@ fn test_args_combined() -> Result<()> {
 fn test_no_color_env_var() -> Result<()> {
     // Test that NO_COLOR environment variable disables color output
     // When NO_COLOR is set and no explicit -C/-M flag is given, colors should be disabled.
-    let (result, _code) = spawn_with_signal_retry(
-        || {
-            let mut cmd = Command::new(succinctly_bin());
-            cmd.args([
-                "jq", ".", // No -C or -M flag
-            ])
-            .env("NO_COLOR", "1");
-            cmd
-        },
+    let (result, _code) = spawn_jq_with_env(
+        &["."], // No -C or -M flag
+        "NO_COLOR",
+        "1",
         Some(b"{\"a\":1}"),
     )?;
     let stdout = String::from_utf8(result.stdout)?;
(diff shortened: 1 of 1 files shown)


newhoggy:
Ran the mandatory code review (8 independent angles, all cross-verified). No
correctness bugs found anywhere — a unanimous verdict across line-by-line,
removed-behavior, altitude, conventions, reuse, simplification, efficiency,
and cross-file angles, each confirming argument order, stdin/env wiring,
stdout/stderr piping, and exit-code semantics are preserved exactly at every
converted call site, plus the full 1103-test suite passing.

**Fixed:** the most-repeated finding (3 independent agents) — the 8
environment-variable tests (`NO_COLOR`, `JQ_COLORS` ×3, `JQ_LIBRARY_PATH`,
`HOME` ×2) each hand-rolled the same `spawn_with_signal_retry` closure since
`spawn_jq` has no way to set an env var. Extracted `spawn_jq_with_env(args,
env_key, env_value, stdin_input)` mirroring `spawn_jq`'s own shape, and
converted all 8 sites — collapsing each ~9-11 line closure into a single
call.

**Left as documented, lower-value debt** (each flagged by 1-2 agents, none
blocking, no correctness risk):
- 6 sites needing file-path args (`NamedTempFile::path()`/`TempDir::path()`
  return `&Path`, not `&str`) still use a direct closure rather than
  `spawn_jq` + `.to_string_lossy()` — a real simplification opportunity, but
  lower value than the 8x env-var duplication and not worth the extra
  lossy-conversion surface for this narrow test-infra fix.
- `jq_stderr`'s `Vec<&str>` rebuild is a 3rd copy of an idiom already used
  twice elsewhere in the file (pre-existing pattern, not new duplication).
- `test_default_identity_filter`'s carve-out comment is longer than the code
  it defends — accurate but could be tightened; not changed since the
  underlying reasoning is correct and re-litigating it risks over-editing a
  test this PR was careful not to change the semantics of.

Rebased onto `main` (which had advanced 5 commits, including a new
`run_jq_interleaved` helper added in the same spot as my `spawn_jq_with_env`
— simple additive merge conflict, both kept). Re-ran full local verification
after both the review fix and the rebase: build/test/clippy under both
default and full (`cli,simd,regex,serde`) feature sets, plus
`--no-default-features` and `cargo doc --all-features`. All 1107 tests in
`jq_cli_tests.rs` pass (4 new ones from the rebase, all pre-existing and
unrelated to this PR).

PR #1900, opened 2026-08-30
fix(jq): thread EvalSemantics through cannot_iterate's remaining sites

newhoggy:
## Summary

#1055 built the mode-aware value-preview dispatch (\`describe_with\`/\`dump_truncated_with\`)
and migrated \`EvalError::cannot_iterate\` at 19 of 29 call sites where \`S: EvalSemantics\`
was already directly in scope, per its own staged design ("migrate call sites family by
family... only as far as needed"). This PR threads \`S\` through the remaining 6 sites that
still called the jq-pinned shim:

- \`builtin_any\`, \`builtin_all\`, \`builtin_flatten\`, \`builtin_from_entries\` — mechanical
  once \`S\` is added to their own generic params (their callers already had it in scope)
- \`set_path\`, \`set_path_through_iterate\` — needed real threading: neither function was
  generic over \`S\` at all, so both (mutually recursive) needed the param added and threaded
  through every internal call, plus their one external caller

## Verification

The mode-aware formatting only diverges from jq's shim for specific value shapes — a plain
\`5 | any\` looks identical either way, since jq's number-formatting quirks only bite
*computed* floats. Used \`OwnedValue::Float(3.0)\` (not a source-preserving \`NumberLiteral\`)
as the differentiator: jq's own error-message convention spells a computed whole-number
float without a decimal point (\`"3"\`, matching real jq's own \`1.0+2.0|tostring\`), while
yq/plain-JSON's convention keeps \`"3.0"\` — confirmed via \`stream_owned_value_json_jq\`'s own
existing unit test. New tests call \`succinctly::jq::eval\` directly with \`YqSemantics\`
(the library-API surface these functions are reachable through — same scoping #1755
established for the sibling sort/unique/group_by family) and assert the real \`"3.0"\`
appears, with a same-filter jq-mode assertion alongside every one proving \`"3"\` is
unchanged.

### \`set_path\`/\`set_path_through_iterate\`: wired correctly, not yet reachable in practice

Worth calling out rather than glossing over: their \`cannot_iterate_with(EvalTag::Yq, ..)\`
is not actually reachable through the one external caller (\`eval_assign\`, for plain \`=\`)
today. That caller always passes \`scalar_noop = (S::TAG == EvalTag::Yq)\` (no operator-based
exception the way \`-=\`/\`*=\` have), and \`is_yq_field_index_noop_scalar\` unconditionally
no-ops any non-container target before the error arm ever runs — confirmed live:
\`succinctly yq '(1.0+2.0) | .[] = 9'\` prints \`3.0\` unchanged, never errors. The new test
calls both functions directly (bypassing the no-op guard) to verify the wiring itself,
independent of whether today's one caller happens to exercise it — the same "migrate the
primitive correctly even where today's caller doesn't reach the divergent branch" spirit
#1055's own staged approach already established.

## Verification, jq mode

Confirmed unchanged throughout: the jq golden suite (\`jq_golden_conformance\`) still passes,
and every new test pairs its yq-mode assertion with a same-filter jq-mode one proving the
old \`"3"\` wording is untouched.

## Scope

This issue names more remaining work (\`cannot_use_as_object_key\`, 5 more call sites; every
other \`describe(\`/\`dump_truncated(\`-calling constructor in \`error.rs\` not yet touched at
all) — left for a follow-up, matching #1055's own staged migration rather than attempting
everything in one PR. Issue intentionally left open, narrowed to that remaining scope.

## Test plan

- [x] New tests in \`src/jq/eval.rs\`:
      \`test_cannot_iterate_call_sites_use_the_real_mode_tag_1494\` (all 4 builtins),
      \`test_set_path_cannot_iterate_uses_the_real_mode_tag_1494\`
- [x] Full test suite (\`cargo test --features cli,simd,regex,serde\`): all green
- [x] jq golden suite (\`jq_golden_conformance\`) and yq golden suite
      (\`yq_golden_conformance\`, \`yq_golden_conformance_under_crlf_and_lone_cr\`): all green
- [x] \`cargo fmt --check\` clean
- [x] \`cargo build --no-default-features\` (no_std) still builds

Refs #1494

### DIFF
--- a/src/jq/error.rs
+++ b/src/jq/error.rs
@@ -705,15 +705,9 @@ impl EvalError {
 
     /// `Cannot iterate over <type> (<value>)`.
     ///
-    /// jq-pinned shim (#1055): delegates to
-    /// [`EvalError::cannot_iterate_with`] fixed at `EvalTag::Jq` — use that
-    /// constructor directly once `S: EvalSemantics` is in scope at the call
-    /// site.
-    pub fn cannot_iterate(value: &OwnedValue) -> Self {
-        Self::cannot_iterate_with(EvalTag::Jq, value)
-    }
-
-    /// Mode-aware sibling of [`EvalError::cannot_iterate`] (#1055).
+    /// The jq-pinned shim this once forwarded from (#1055) is gone: #1494
+    /// finished migrating every production call site to pass its own real
+    /// `S::TAG` instead of a hardcoded `EvalTag::Jq`.
     pub fn cannot_iterate_with(tag: EvalTag, value: &OwnedValue) -> Self {
         Self::new(format!("Cannot iterate over {}", describe_with(tag, value)))
     }
@@ -1559,7 +1553,7 @@ mod tests {
 
     #[test]
     fn internal_errors_carry_no_payload_but_catch_sees_the_message() {
-        let err = EvalError::cannot_iterate(&OwnedValue::Int(1));
+        let err = EvalError::cannot_iterate_with(EvalTag::Jq, &OwnedValue::Int(1));
         assert_eq!(err.message, "Cannot iterate over number (1)");
         assert_eq!(err.value, None);
         assert_eq!(
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -17059,7 +17059,7 @@ fn set_path<S: EvalSemantics>(
                     // that case on its own, exactly as it did before #1428.
                     && would_create
                     && !tail_writes_from_fresh_parent(|probe| {
-                        set_path_through_iterate(
+                        set_path_through_iterate::<S>(
                             probe,
                             split.optional,
                             &split.tail,
@@ -17122,7 +17122,7 @@ fn set_path<S: EvalSemantics>(
                                 terminal_write: matches!(split.tail, Expr::Identity),
                             },
                             |sub| {
-                                set_path(
+                                set_path::<S>(
                                     sub,
                                     &split.tail,
                                     new_value.clone(),
@@ -17176,7 +17176,7 @@ fn set_path<S: EvalSemantics>(
                     // See the `split_at_iterate` arm above.
                     && would_create
                     && !tail_writes_from_fresh_parent(|probe| {
-                        set_path(
+                        set_path::<S>(
                             probe,
                             last_path,
                             new_value.clone(),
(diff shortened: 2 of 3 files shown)


newhoggy:

**Fixed — real, actionable finding**: one agent noticed my own PR's title ("thread EvalSemantics through \`cannot_iterate\`'s *remaining* sites") was literally true — after threading \`S\` through the last 6 production call sites, the jq-pinned \`cannot_iterate(value)\` shim had zero production callers left anywhere in the tree (only 2 test call sites and a doc-comment mention). Its own doc comment already named this exact retirement condition ("use \`cannot_iterate_with\` directly once \`S: EvalSemantics\` is in scope"). Removed the shim (commit 433f556c0), updated both test call sites to call \`cannot_iterate_with(EvalTag::Jq, ...)\` directly, fixed two now-stale doc-comment references.

**Investigated, filed separately as #1901 — not fixed here (correctly out of scope)**: one agent live-verified against pinned yq v4.53.3 that \`any\`/\`all\`/\`flatten\`/\`from_entries\` don't use jq's "Cannot iterate over \<type\> (\<value\>)" template at all in real yq — they use completely different hardcoded wording with no value preview (\`any only supports arrays, was !!int\`, etc.), and \`.[]\` on a bare scalar is a silent no-op in real yq's read side too, not an error. This is a **pre-existing** divergence, not introduced or worsened by this PR — my fix makes the existing (wrong-template) message *internally consistent* with the real evaluation mode's float-formatting convention, which is genuinely useful progress, but doesn't close the larger "wrong template entirely" gap. Properly fixing that needs new per-builtin message templates verified against yq, well beyond this issue's own \`EvalTag\`-threading scope — filed as #1901 with the live repros rather than scope-creeping this PR or leaving it undocumented per ADR-0018.

**No correctness bugs found** — 4 of 5 agents did a full line-by-line/cross-file trace of the generic threading (all 6 call sites, every recursive call, the write-path's \`eval_assign\` → \`set_path\`/\`set_path_through_iterate\` chain, \`eval_generic.rs\`'s separate dispatch confirming no parallel implementation was missed) and found nothing wrong — compiles, clippies clean, both new tests pass and are genuinely discriminating (verified by checking what a hardcoded-\`EvalTag::Jq\` shim would have failed on).

**Not actioned**: no simplification/efficiency/reuse issues found — the generic threading matches the established pattern at 19+ other call sites exactly, with no duplicated logic or unnecessary complexity introduced.

Full suite, both clippy legs, fmt, and both golden test suites (jq + yq) all re-verified clean after the shim removal.

Issue #1817, opened 2026-08-29
yq: the same nesting-depth panic #1793 fixed for jq is unfixed for succinctly yq


newhoggy:
## Summary

#1793 (PR #1815) fixed `succinctly jq`'s default (lazy) per-document dispatch loop so that `sort`/`join`/`map(.)` (builtins with no native lazy fast path, or that still need to materialize before printing) no longer crash the whole process on an adversarially/accidentally deep document — a `catch_unwind` at the CLI's per-document dispatch site now isolates the failure and continues the stream, matching the existing `#1194`/`#355` convention.

`succinctly yq`'s own dispatch loop (`run_yq` in `src/bin/succinctly/yq_runner.rs`) reaches the identical shared guard (`assert_nesting_depth`/`MAX_NESTING_DEPTH`, `src/jq/eval_generic.rs`) via its own `eval_with_cursor_using::<YqSemantics, _>` call sites — confirmed live during #1793's review, with **zero** equivalent protection.

## Repro (confirmed live against `main`)

A validly-parsed, block-style-nested YAML document well within YAML's own (lower, 128) parse-time structural limit, but past the shared evaluator's 256 ceiling:

```yaml
# 300 levels of block-style nesting -- parses fine, YAML's own limit is 128 only for flow-style
a:
  a:
    a:
      ... (300 levels)
```

```console
$ succinctly yq 'sort' deep.yaml
thread 'main' panicked at src/jq/eval_generic.rs:267:5: nesting depth exceeds limit of 256
$ echo $?
101
```

In a 3-document YAML stream (`a: 1` / `<deep>` / `b: 2`), the crash on the middle document aborts the whole process before the third document is ever evaluated — the exact "abort mid-stream, drop the rest" failure #1793 was filed to close, just via the sibling CLI entry point #1793's own PR never touched.

## Scope

This is the natural extension of #1793's fix to `run_yq`. #1793's review explicitly flagged this as real and live but deliberately deferred it as a separate-file, separate-scope change (the PR was already substantial after review; see PR #1815's discussion). Whoever picks this up should also consider factoring the `catch_unwind`/message-matching logic (`nesting_depth_panic_message` in `jq_runner.rs`) into something both runners can share, rather than a second hand-copy — flagged as a reuse concern during #1793's own review.

## Provenance

Found during multi-angle `/code-review` of PR #1815 (#1793).

newhoggy:
[...]

PR #1903, opened 2026-08-30
fix(jq): catch the nesting-depth panic on the materializing CLI branch too

### LINKED ISSUE #1818: jq: --slurp/-S/-C/--ascii-output/--slurpfile, and any filter using input/inputs, still panic uncaught on adversarial nesting after #1793
## Summary

#1793 (PR #1815) fixed the CLI's default (lazy) per-document dispatch loop against an uncaught `MAX_NESTING_DEPTH` panic for `sort`/`join`/`map(.)`. `run_jq`'s **other** top-level branch — taken whenever `can_use_lazy_path` (`src/bin/succinctly/jq_runner.rs`) is `false`, i.e. `--slurp`/`-s`, `-S`/`--sort-keys`, `-C`/`--color-output`, `--ascii-output`, `--slurpfile`, or a filter using `input`/`inputs`/`input_line_number` — reaches the identical shared guard via `validate_json_delimiters` (called from `parse_json_stream`'s fallback) with no equivalent protection, confirmed live during #1793's own review.

## Repro (confirmed live against `main`)

```console
$ python3 -c "print('['*500 + '1' + ']'*500)" > deep.json
$ succinctly jq -c --sort-keys '.' deep.json
thread 'main' panicked at src/bin/succinctly/jq_runner.rs:4100:5: nesting depth exceeds limit of 256
$ echo $?
101
```

Same result for `--slurp`, `-C`, `--ascii-output`, and `--slurpfile`.

## Why this is worse than the bug #1793 fixed

This branch reads and validates **every** document up front, before printing anything. A 3-document stream (`[1]` / `<deep>` / `[2]`) run through `-S` or `--slurp` prints **nothing at all** before crashing — not even the legitimate `[1]` that precedes the bad document — unlike the fixed lazy path, which now isolates just the one bad document and still prints its siblings.

## Scope

#1793's own title/repro is specifically about a *builtin* lacking a lazy fast path under the CLI's *default* dispatch (no special flags) — this is a different trigger (a CLI *flag* forcing whole-batch materialization), sharing only the underlying guard. Deliberately scoped out of PR #1815 as a separate concern rather than expanding that PR's already-substantial diff further.

Note: because this branch materializes the whole batch before any output, true per-document isolation (like the lazy path now has) may not be achievable here without a larger restructure — the achievable fix is likely "convert the uncaught panic into a clean, single diagnostic for the whole batch" (matching `print_json`'s own `anyhow::ensure!`-based approach for the identity-path's analogous guard), not full per-document continuation.

## Provenance

Found during multi-angle `/code-review` of PR #1815 (#1793), confirmed independently by 4 review angles.

newhoggy:
Fixes #1818.

## Summary

#1793 fixed the CLI's default (lazy) per-document dispatch against an
uncaught `MAX_NESTING_DEPTH` panic, but `run_jq`'s **other** top-level
branch — taken whenever `can_use_lazy_path` is `false` (`--slurp`,
`-S`/`--sort-keys`, `-C`/`--color-output`, `--ascii-output`, `--slurpfile`,
or a filter using `input`/`inputs`/`input_line_number`) — reaches the
identical guard via `validate_json_delimiters` (`parse_json_stream`'s
fallback) with no equivalent protection.

## Repro (confirmed live against `main`, pre-fix)

```console
$ python3 -c "print('['*500 + '1' + ']'*500)" > deep.json
$ succinctly jq -c --sort-keys '.' deep.json
thread 'main' panicked at src/bin/succinctly/jq_runner.rs:4393:5:
$ echo $?
101
```

Same result for `--slurp`, `-C`, `--ascii-output`, and a filter using
`inputs` with no flag at all.

## Fix

Added `check_nesting_depth`, a checked sibling of `assert_nesting_depth`
sharing the exact same `MAX_NESTING_DEPTH` ceiling and message
(`print_json`'s own `anyhow::ensure!` guard already uses this message for
the analogous lazy-path case). `validate_json_delimiters` already threads a
`Result` and runs entirely at the CLI's input-validation stage before any
user filter evaluates, so there's no `try`/`catch`-reachability concern the
way there is for `to_owned`'s own hot-path panic (documented in that
function's own doc comment) — unlike that guard, this one can just return
an error instead of panicking.

```console
$ succinctly jq -c --sort-keys '.' deep.json
jq: error (at deep.json:0): nesting depth exceeds limit of 256
$ echo $?
5
```

Confirmed `--slurpfile`'s own error-reporting shape (a plain `Error: ...`
at exit 1, not jq's `jq: error (at ...)` at exit 5) is pre-existing and
consistent with how it reports every other error type too (verified
against a malformed-JSON `--slurpfile` control case) — not a new
inconsistency introduced here.

## Testing

Added 6 new tests covering `--slurp`, `-S`, `-C`, `--ascii-output`, a
filter using `inputs`, and a positive control (legitimately-nested input
under the limit still succeeds). Updated `docs/compliance/jq/limitations.md`'s
existing #1793 entry to reflect the narrowed scope of what remains uncaught
(`-e`/`--exit-status`'s separate materializer, and `succinctly yq` entirely
— both tracked separately, #1817).

`cargo doc --all-features`.

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -1039,17 +1039,21 @@ never gets this far). Falls under ADR-0018's "would take the host process down" 
 the fix that turned this from an uncaught process panic into a clean, catchable
 diagnostic.
 
-This fix is scoped to the CLI's default (lazy) per-document dispatch, matching #1793's own
-repro — the identical panic remains uncaught via any of: a CLI flag that forces
-whole-batch materialization up front (`--slurp`, `-S`/`--sort-keys`, `-C`/`--color-output`,
-`--ascii-output`, `--slurpfile`); a filter using `input`/`inputs`/`input_line_number`,
-which routes to that same materializing branch with *no flag at all* (#1818); `-e`/
+#1793's own fix was scoped to the CLI's default (lazy) per-document dispatch, matching its
+own repro. #1818 closed the identical gap on the CLI's *other* top-level branch — a CLI
+flag that forces whole-batch materialization up front (`--slurp`, `-S`/`--sort-keys`,
+`-C`/`--color-output`, `--ascii-output`, `--slurpfile`), or a filter using
+`input`/`inputs`/`input_line_number`, which routes to that same materializing branch with
+*no flag at all* — via `validate_json_delimiters`'s own checked guard
+(`check_nesting_depth`, `src/jq/eval_generic.rs`) rather than `catch_unwind`, since that
+walk already threads a `Result` and runs before any user filter evaluates at all.
+
+Still uncaught, tracked separately, not fixed by either #1793 or #1818: `-e`/
 `--exit-status`'s own separate materializer (`src/jq/lazy.rs`, already pinned uncaught by
 `test_exit_status_query_rejects_adversarial_nesting_998`); and `succinctly yq`, which has
-no equivalent guard on either its default or materializing path at all (#1817). Tracked
-separately, not fixed here. Confirmed live, also pre-existing and unrelated to this fix:
-`print_json`'s own guard can flush corrupted/truncated JSON to stdout before it fires
-(#1819).
+no equivalent guard on either its default or materializing path at all (#1817). Confirmed
+live, also pre-existing and unrelated to either fix: `print_json`'s own guard can flush
+corrupted/truncated JSON to stdout before it fires (#1819).
 
 ## Regex flags `l` and `n`
 
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -12,7 +12,7 @@ use std::path::{Path, PathBuf};
 use succinctly::dsv::{build_index as build_dsv_index, DsvConfig, DsvRows};
 use succinctly::jq::document::{effective_keys, key_hash, DistinctKeyCursors};
 use succinctly::jq::eval_generic::{
-    assert_nesting_depth, eval_with_cursor, to_owned as generic_to_owned, GenericResult,
+    check_nesting_depth, eval_with_cursor, to_owned as generic_to_owned, GenericResult,
     MAX_NESTING_DEPTH,
 };
 use succinctly::jq::walk::map_builtin_subexprs;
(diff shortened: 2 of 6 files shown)


newhoggy:
Ran the mandatory code review. Three finder passes converged: no correctness
bugs found anywhere (verified off-by-one absence, error-propagation
correctness, depth-counting convention match with `to_owned_at_depth`, and
that the two deliberately-untouched panicking call sites — `lazy.rs`'s
`-e`/`--exit-status` materializer and `yq_runner.rs` — remain unchanged and
still correctly documented).

**Fixed** (flagged independently by 2 of 3 passes): `check_nesting_depth`'s
new condition/message was a 4th independent copy of the same "nesting depth
exceeds limit of N" string already duplicated between `assert_depth`'s
`panic!` and `print_json`'s `anyhow::ensure!` — exactly the "duplicated
predicates diverge silently" shape #998's own consolidation into
`assert_depth` already fixed once for the panicking family. Extracted
`nesting_depth_exceeded_message(max) -> String` in `value.rs`, reused by
all three call sites. Confirmed byte-identical output across all three
after the change (live-tested both the checked path and the still-panicking
identity-path guard).

**Unrelated incident during review**: one finder subagent kept running past
its own synthesis and injected `eprintln!`/`dbg`-style debug scaffolding
(plus an unconditionally-panicking scratch test) directly into this
worktree's shared files — `output.rs`, `jq_runner.rs`, `eval_generic.rs` —
which surfaced as 7 unrelated `jq_cli_tests.rs` failures with corrupted
stderr on the next full-suite run. Traced to the stray agent via `git diff`
on files I hadn't touched, stopped it with `TaskStop`, and reverted every
injected line by hand (verified via a repo-wide `grep -rn "SCRATCH"`
returning nothing). Unrelated to this PR's own correctness; noting it here
since it explains an interim red build that had nothing to do with the fix.

Full local verification re-run after both the review fix and a rebase onto
`main` (advanced 4 commits): build/test/clippy under both default and full
`cargo doc --all-features`. All green, including all 6 new #1818 tests and
the full pre-existing #1793 cluster.

PR #1912, opened 2026-08-30
refactor(jq): extract catch_error_under_optional, collapse 2 hand-copies

### LINKED ISSUE #1888: jq: the 'catch Error under ambient optional, keep-or-discard prefix' post-loop dispatch is now hand-copied 4 times
**Severity: Low** (code-quality, no correctness risk — each copy is independently tested and correct)

## Summary

Found via code review on PR #1886 (#1869's fix). Four path-context arms in
`eval_pipe_with_path_context_internal` (`src/jq/eval.rs`) now independently
hand-roll the same two-part shape: (a) force `optional=false` into every
per-element/per-branch recursive call, accumulating a `stopped:
Option<QueryResult>` instead of early-returning, then (b) a post-loop
`match stopped` that re-gates `QueryResult::Error(_) |
QueryResult::Partial(_, Control::Error(_))` on the arm's own ambient
`optional`, with construct-specific handling for `Break`/`Halt`
(atomic-discard for `Array`/`Map`, keep-fall-through for `Optional`/`Iterate`):

- `Expr::Optional`'s own fast-path arm (unconditional catch, since it *is* the `?`)
- `Builtin::Map` (#1826) — atomic discard on `Error`, `Break`, and `Halt`
- `Expr::Array` (#1302) — same atomic-discard model
- `Expr::Iterate` (#1869, this PR) — keep-prefix on `Error`, never catches `Break`/`Halt`

Two independent review agents on #1886 flagged this as the same
pattern being copy-pasted-with-adaptation a fourth time, with each
arm's own doc comment explicitly cross-referencing the others ("the same
`Expr::Array`/`Builtin::Map` fix") rather than calling a shared function.

## Why this matters

#1826's own review already caught a subtle trap in this shape once: matching
only `Partial(_, Control::Error(_))` and missing the bare `QueryResult::Error(_)`
case (which `partial()` collapses to when the erroring element is *first*, so
the accumulated prefix is still empty) silently reproduces the original bug.
A future author writing a 5th copy of this dispatch from scratch, without the
accumulated comment trail on the first four, has a real chance of re-missing
that exact ORed-pattern gap.

## Suggested fix direction

A shared helper — something like
`fn finish_per_element_path_context_loop(stopped: Option<QueryResult>, results: Vec<OwnedValue>, optional: bool, atomic: bool) -> QueryResult`
(exact shape TBD) — encoding the correct catch-shape once, parameterized on
whether the construct is atomic (`Array`/`Map`) or keep-prefix (`Iterate`),
would make the `Error`-vs-`Partial(_, Error)` gap structurally impossible to
omit at a new call site. `Break`/`Halt` handling differs enough between the
atomic and keep-prefix constructs that it may need to stay a parameter or a
small enum rather than being folded into the same boolean.

Not fixed as part of #1869/PR #1886, since unifying three already-shipped,
independently-tested arms is a real refactor with its own blast radius,
better scoped as its own change.

## Related

- #1302 — `Expr::Array`'s own fix, the atomic-discard precedent
- #1826 — `Builtin::Map`'s fix, same atomic-discard model, and the source of
  the `Error`-vs-`Partial(_, Error)` trap noted above
- #1869 / PR #1886 — `Expr::Iterate`'s fix, the keep-prefix variant, and
  where this duplication was most recently flagged
- #1335 — the separate, deeper "ambient optional bleeds through `rest`"
  architectural gap all four of these arms currently share (not what this
  issue is about — this issue is about the code shape, not the residual
  semantic gap)

newhoggy:
## Summary

`Builtin::Map` (#1826) and `Expr::Iterate` (#1869) independently hand-rolled the same "catch `Error` under ambient `optional`, keep-or-discard prefix" post-loop dispatch after their own fix for `iterate_element_step`'s `optional=false` isolation — each carrying its own copy of the exact `Error`-vs-`Partial(_,Error)` OR'd pattern #1826's own review had to catch once already (matching only `Partial` silently reproduces the original bug whenever the erroring element is first, since `partial()` collapses an empty prefix to the bare variant).

Adds `catch_error_under_optional(stopped, optional, atomic)`, parameterized on whether the construct is atomic (`Map`: any escaping control signal discards the whole prefix) or keep-prefix (`Iterate`: a caught `Error` keeps what was already produced, `Break`/`Halt` pass through untouched with their own `Partial`-carried prefix intact).

A subtlety worth calling out explicitly: the prefix has to be read out of the matched `Partial` pattern itself, not a separately-threaded `results` accumulator — `accumulate_path_context_step` already moves the caller's own `results` `Vec` into that `Partial` via `mem::take` at the moment it produces a stop signal, so by the time the post-loop code runs, reusing the caller's own (by-then-empty) binding would silently keep nothing. Traced this by hand before writing any code, since it's exactly the kind of detail a naive "just extract the common shape" refactor would get wrong silently.

`Expr::Array`'s and `Expr::Optional`'s arms are deliberately left untouched: unlike `Map`/`Iterate`, they don't share the loop-shaped `stopped: Option<QueryResult>` + `results: Vec<OwnedValue>` input — both call `eval_pipe_with_path_context_internal` once and match directly on its `QueryResult`, a genuinely different shape. Forcing them into this helper's signature would need an adapter, not a natural fit, and the issue's own suggested function name (`finish_per_element_path_context_*loop*`) already scopes to the loop-shaped pair.

## Test plan
- [x] Full existing test suite green, including #1302/#1326/#1335/#1826/#1869's own regression tests **unchanged** (byte-identical behavior, not just "still passes")
- [x] 18-case automated byte-for-byte comparison against the pre-refactor binary — both constructs, both optional states, first- and middle-element errors, break, halt, no-ambient-optional control cases
- [x] Added `test_catch_error_under_optional_full_truth_table_1888`, a direct unit test covering every branch — confirmed to fail with the exact symptom when either `atomic` branch is inverted (verified locally, then restored, not committed)
- [x] `cargo fmt --check`

Fixes #1888

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -26418,12 +26418,16 @@ fn accumulate_path_context_step<'a, W: Clone + AsRef<[u64]>>(
 /// time a caller's loop breaks, its own `results` binding is already empty
 /// -- reusing it here instead of the pattern's own `prefix` would silently
 /// keep nothing.
-fn catch_error_under_optional<W: Clone + AsRef<[u64]>>(
+fn catch_error_under_optional<W>(
     stopped: QueryResult<'_, W>,
     optional: bool,
     atomic: bool,
 ) -> QueryResult<'_, W> {
     match stopped {
+        // `atomic` only has a prefix to discard-or-keep once one was
+        // actually accumulated (the `Partial` arm below); a bare `Error`
+        // (`partial()`'s own empty-prefix collapse, #400/#494) has nothing
+        // for `atomic` to affect either way.
         QueryResult::Error(e) => {
             if optional {
                 QueryResult::None
@@ -26431,19 +26435,16 @@ fn catch_error_under_optional<W: Clone + AsRef<[u64]>>(
                 QueryResult::Error(e)
             }
         }
-        QueryResult::Partial(prefix, Control::Error(e)) => {
-            if optional {
-                if atomic {
-                    QueryResult::None
-                } else {
-                    owned_vec_to_result(prefix)
-                }
-            } else if atomic {
-                QueryResult::Error(e)
-            } else {
-                QueryResult::Partial(prefix, Control::Error(e))
-            }
-        }
+        // All four `(atomic, optional)` cells, spelled out as a flat table
+        // rather than nested conditionals so each is visible at a glance:
+        // `atomic` decides whether `prefix` survives at all; `optional`
+        // decides whether the survival (or lack of one) also raises.
+        QueryResult::Partial(prefix, Control::Error(e)) => match (atomic, optional) {
+            (true, true) => QueryResult::None,
+            (true, false) => QueryResult::Error(e),
+            (false, true) => owned_vec_to_result(prefix),
+            (false, false) => QueryResult::Partial(prefix, Control::Error(e)),
+        },
         QueryResult::Partial(_, Control::Break(label)) if atomic => QueryResult::Break(label),
         QueryResult::Partial(_, Control::Halt(code)) if atomic => QueryResult::Halt(code),
         other => other,
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -26376,6 +26376,80 @@ fn accumulate_path_context_step<'a, W: Clone + AsRef<[u64]>>(
     }
 }
 
+/// Turn a per-element path-context loop's stop signal -- the `Some(stop)`
+/// [`accumulate_path_context_step`] returns once `iterate_element_step`
+/// forces `optional = false` into each element's own evaluation -- into
+/// this arm's final `QueryResult`, catching a genuine `Error` under this
+/// arm's own *ambient* `optional` and either discarding or keeping the
+/// already-produced prefix depending on whether the construct is atomic
+/// (#1888): one definition for the "catch `Error` under ambient `optional`,
+/// keep-or-discard prefix" dispatch `Builtin::Map` (#1826) and
+/// `Expr::Iterate` (#1869) each independently hand-rolled after their own
+/// fix for this exact `iterate_element_step` isolation.
+///
+/// `atomic`:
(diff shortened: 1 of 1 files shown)


newhoggy:
Code review found real value in two directions, but one turned out to be a genuine correctness trap once tested against a real edge case:

**Attempted, then reverted**: also routing \`Expr::Array\`'s own post-loop match through \`catch_error_under_optional\`, since its dispatch shape looked identical at a glance (and two independent review agents confirmed this reading). Attempted it, then caught a live regression via an extended byte-comparison script — **not** by the review itself: \`(.a)? | [key, error("boom")]\` (ambient optional, no \`?\` on the array itself) went from correctly producing no output at all to wrongly producing \`[]\`.

Root cause: the helper's caught-error \`QueryResult::None\` is indistinguishable from a legitimate "the source evaluation produced zero outputs" \`QueryResult::None\` once returned. \`Expr::Array\`'s own arm needs to treat those two differently (early-return "no output at all" vs. continue to build an empty array \`[]\`), while \`Builtin::Map\`/\`Expr::Iterate\`'s usage is safe only because they gate the helper call behind their own \`Option<QueryResult>\` — the "no stop signal at all" case (a legitimately empty loop) never reaches the helper in the first place for those two. \`Array\`'s \`inner_result\` is a raw \`QueryResult\` with no such gate. Reverted — \`Array\` stays exactly as originally scoped (untouched), and this specific "looks identical, isn't" trap is now recorded for anyone tempted to try the same conversion later.

**Applied**: restructured the `Partial(prefix, Control::Error(e))` arm from nested if/else into a flat `match (atomic, optional)`, and dropped the helper's unused `W: Clone + AsRef<[u64]>` bound.

Re-verified via the full test suite and an extended 27-case byte-for-byte comparison against the pre-refactor binary (the original 18 plus 9 new Array-specific cases, including the exact one that caught the regression above).

Issue #1919, opened 2026-08-30
yq: |= (unlike = and the compound-assign family) doesn't autovivify Null under a mid-chain Iterate at all


newhoggy:
## Summary

Found investigating #1857. Real yq autovivifies `Null` -> `[]` for a
mid-chain `Iterate` uniformly across every assignment operator, including
plain `|=`:

\`\`\`console
$ echo 'a: null' | yq -o=json '.a[].b |= 5'
{
  "a": []
}
\`\`\`

succinctly's \`=\`, \`+=\`/\`-=\`/\`*=\`/\`/=\`/\`%=\`/\`//=\` all match this correctly
(the last group as of #1857). Plain \`|=\` does not:

\`\`\`console
$ succinctly yq -o=json '.a[].b |= 5'
Error: Cannot iterate over null (null)
\`\`\`

This is a hard error, not just an eager-RHS-evaluation divergence (#1857's
own shape) — \`|=\` doesn't autovivify \`Null\` under \`Iterate\` at all, where
\`=\`/the compound family already did before #1857 (confirmed live: \`.a[].b
= 5\` on \`a: null\` already produced \`a: []\` prior to #1857's fix, since
that autovivification happens in \`set_path\`, not in the noop-check
machinery #1857 touched).

## Why this is a separate issue from #1857

\`|=\` is evaluated through \`eval_update\`, which \`yq_assign_noop_check\`'s
own doc comment already notes is untouched by that mechanism: "\`eval_update\`
is shared with plain \`|=\` (which has no pre-check at all and no
\`pristine\`/\`paths\` to hand it)". So \`|=\`'s autovivify gap is in a
completely different code path (whatever \`eval_update\`/\`update_path\` do
for a mid-chain \`Iterate\` over \`Null\`), not the \`yq_assign_noop_check\`/
\`assign_path_all_noop\`/\`assign_path_rhs_unused\` machinery #1857 fixed.

## Where

\`eval_update\`/\`update_path\` (\`src/jq/eval.rs\`) — likely the same
\`autovivify_array\`/\`Iterate\`-handling logic \`set_path\`/
\`set_path_through_iterate\` already have for \`=\`, just missing from
whatever \`|=\`'s own per-element update mechanism does for a mid-chain
\`Iterate\`.

newhoggy:
🔧 Claimed by agent `Dimitri` — branch `issue-1919-yq-pipe-mid-chain-iterate-autovivify`

[...]

PR #1920, opened 2026-08-30
fix(jq): yq-mode assignment through a null-autovivify skips the unused RHS

### LINKED ISSUE #1857: yq: no-op predicate can't express 'skip RHS, still autovivify' for a mid-chain Iterate over Null

newhoggy:
## Summary

A mid-chain \`Iterate\` autovivifying \`Null\` into an empty array has zero
elements to write into, so real yq never evaluates the RHS:

\`\`\`console
$ echo 'a: null' | yq -o=json '.a[].b = error("boom")'
{
  "a": []
}
$ echo 'a: null' | succinctly yq -o=json '.a[].b = error("boom")'   # before this fix
Error: boom
\`\`\`

#1432's \`yq_assign_noop_check\` deliberately excluded \`Null\` from its
\`Skip\`/\`Continue\` split to avoid discarding the legitimate \`null\` -> \`[]\`
write (reporting it as a total no-op would have thrown the write away) —
leaving the RHS to evaluate eagerly anyway. That's the exact gap #1432
filed as #1857.

## Fix

Added \`yq_assign_rhs_unused\`/\`assign_path_rhs_unused\` as a twin of the
existing \`yq_assign_is_total_noop\`/\`assign_path_all_noop\` pair — same
walk, but an \`Iterate\` reaching \`Null\` now answers "RHS unused" (\`true\`)
instead of falling to the catch-all \`false\`, since autovivifying to \`[]\`
means zero elements regardless of what the remaining path steps say.

\`yq_assign_noop_check\` performs the write immediately with a placeholder
value when every resolved path satisfies this new check (\`set_path\` can't
fail here — the placeholder is proven never actually read) and routes it
through the same \`Skip\` channel \`eval_assign\`/\`eval_compound_assign\`/
\`eval_alternative_assign\` already use, so **no caller-side changes were
needed** — the fix is entirely contained inside \`yq_assign_noop_check\`.

## Scope

Reaches every yq-mode assignment operator that routes through this
mechanism (\`=\`, \`+=\`, \`-=\`, \`*=\`, \`/=\`, \`%=\`, \`//=\`), confirmed live against
yq v4.53.3. Plain \`|=\` doesn't share this mechanism at all (\`eval_update\`
has no pre-check, per \`yq_assign_noop_check\`'s own doc comment) and turned
out to have a **separate, more severe** divergence during investigation —
a hard \`Cannot iterate over null\` error instead of missing
autovivification entirely — filed as #1919 rather than folded in here.

## Verification

- Live-confirmed against yq v4.53.3 for the top-level case, the nested
  recursive case (\`.a[][].b\` with \`a: [null, [1]]\`), all 6 assignment
  operators, and a regression guard (\`Null\` mixed with a genuine write
  target still evaluates the RHS and still errors)
- Updated \`test_yq_nonterminal_iterate_null_target_still_evaluates_rhs_1298\`
  (renamed to \`..._skips_rhs_1298\`) to assert the fixed behavior instead
  of pinning the bug
- Added \`test_yq_nonterminal_iterate_null_mixed_with_real_write_still_evaluates_rhs_1857\`
  and \`test_yq_nonterminal_iterate_null_target_skips_rhs_compound_operators_1857\`
- \`cargo test --features cli,simd,regex,serde --workspace\`: all green
- \`cargo clippy\` (both feature legs), \`cargo check --no-default-features\`,
  \`cargo doc --no-deps --all-features\`, \`cargo fmt --check\`: all clean

Fixes #1857

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -867,8 +867,28 @@ A `null` target is deliberately excluded from this fix — real yq autovivifies 
 as part of the write itself (already correct on both sides before and after this fix), so it
 is not a *total* no-op the way an empty/all-scalar container is, and this predicate's
 all-or-nothing `Skip`/`Continue` caller has no way to express "skip the RHS, but still
-perform the write." That narrower, still-open gap is tracked separately by
-[#1857](https://github.com/rust-works/succinctly/issues/1857).
+perform the write." That narrower gap was tracked separately as #1857 -- see the entry below.
+
+**Fixed by [#1857](https://github.com/rust-works/succinctly/issues/1857):** a *mid-chain*
+`Iterate` autovivifying `null` into `[]` still evaluated the RHS eagerly, where real yq's
+equivalent write never needs it (zero elements to write into). New `assign_path_rhs_unused`
+answers a broader question than `assign_path_all_noop`'s total-no-op check -- "is the RHS
+ever actually read" -- recognizing `null` under a mid-chain `Iterate` as one more case where
+it isn't, and `yq_assign_noop_check` performs the (already fully determined) write with a
+placeholder value instead of evaluating the RHS at all:
+
+```bash
+$ printf 'a: null\n' | yq            -o=json '.a[].b = error("boom")'   # {"a":[]}, RHS never runs
+$ printf 'a: null\n' | succinctly yq -o=json '.a[].b = error("boom")'   # {"a":[]} (fixed)
+```
+
+Reaches every yq-mode assignment operator routed through this mechanism (`=`, `+=`, `-=`,
+`*=`, `/=`, `%=`, `//=`). A **terminal** `Iterate` (`.a[] = v`, the assignment target itself)
+is a different, still-open case neither this fix nor #1432's own covers — see
+[#1921](https://github.com/rust-works/succinctly/issues/1921). Plain `|=` doesn't share this
+mechanism at all and turned out to have a separate, more severe divergence (a hard error
+instead of missing autovivification) — see
+[#1919](https://github.com/rust-works/succinctly/issues/1919).
 
 ### `=`'s multi-output RHS: real yq takes only the last value, no fan-out
 
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -15582,13 +15582,71 @@ fn is_yq_field_index_noop_scalar(target: &OwnedValue) -> bool {
     !is_owned_container(target) && !matches!(target, OwnedValue::Null)
 }
 
-/// Whether writing `path` (a single *resolved* assignment-target path, one
-/// `resolve_dynamic_indexes` entry) into `root` would be real yq's
-/// field/index/iterate scalar-target no-op (#1181): navigating every
-/// component *before* the last reaches a genuine scalar. When true for
-/// every resolved path, real yq discards the RHS/filter entirely rather
-/// than merely skipping the write (#1233, live-verified v4.53.3) --
-/// `eval_assign`'s own caller is what actually skips RHS evaluation; this
+/// The outcome of walking a resolved static path against `set_path`'s own
+/// write semantics, without actually performing the write. Replaces two
+/// independently-hand-synced boolean predicates that used to exist here
+/// (`assign_path_all_noop`/`assign_path_rhs_unused`, and their outer
+/// wrappers, the current [`yq_assign_is_total_noop`]/[`yq_assign_rhs_unused`])
+/// -- #1920 review (round 2) found the split cost a full second tree-walk
+/// on every non-total-noop yq-mode assignment, and risked exactly the
+/// "duplicated predicates diverge silently" failure this codebase has
+/// already hit once (#106): the two boolean walks were near-identical
+/// copies whose *only* intentional difference (an `Expr::Iterate` reaching
+/// `Null`) had to be applied by hand to one and not the other. One walk
+/// producing a 3-way answer removes that risk structurally instead of by
+/// doc-comment discipline; the two outer functions are now thin wrappers
+/// over [`yq_assign_classify`], keeping their own names/signatures
+/// unchanged so every existing cross-reference to them elsewhere in this
+/// file stays accurate.
+enum PathAssignOutcome {
+    /// Every write this path reaches is a no-op: the document doesn't
+    /// change at all.
+    TotalNoop,
+    /// Not a total no-op (the document does change), but the RHS value is
+    /// never actually read anywhere along it -- a mid-chain `Iterate`
+    /// autovivifying `Null` into an empty array, most commonly (#1857).
+    RhsUnusedButChanges,
+    /// A real, RHS-dependent write happens somewhere along this path.
+    NeedsRhs,
+}
+
+impl PathAssignOutcome {
+    fn is_total_noop(&self) -> bool {
+        matches!(self, Self::TotalNoop)
+    }
+
+    fn rhs_unused(&self) -> bool {
+        !matches!(self, Self::NeedsRhs)
+    }
+}
+
+fn yq_assign_is_total_noop(path: &Expr, root: &OwnedValue) -> bool {
+    yq_assign_classify(path, root).is_total_noop()
+}
+
+/// [`yq_assign_is_total_noop`]'s twin for the narrower #1857 gap -- see
+/// [`PathAssignOutcome`]'s own doc comment for why both are now thin
+/// wrappers over one shared walk rather than independent ones.
+///
+/// Deliberately does **not** cover a *terminal* `Iterate` (`.a[] = v`, the
+/// `Iterate` itself is `last`, stripped into the flattened `prefix` before
+/// [`classify_yq_assign_prefix`] is ever reached) reaching an already-empty
+/// container or `Null` -- that shape needs the terminal step's own type
+/// available at the point its target is examined, which this prefix-only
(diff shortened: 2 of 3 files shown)


newhoggy:
Round 2 (3 finder agents) — one genuine correctness gap in my own fix, one doc-staleness issue, one efficiency concern assessed but not changed. All addressed in c2b01b0bb:

1. **Fixed**: a *terminal* \`Iterate\` (\`.a[] = v\`, no trailing path) still evaluates the RHS eagerly over an already-empty container or \`Null\` — confirmed pre-existing on \`main\` both before and after this fix (not caused by #1857/#1920). A terminal Iterate's semantics genuinely differ from a mid-chain one's (overwriting every real element always uses the RHS, regardless of element type), so the existing prefix-only walk structurally can't extend to cover it without real redesign. Filed as #1921, pinned here with a known-gap test (\`test_yq_terminal_iterate_empty_or_null_target_still_evaluates_rhs_known_gap_1921\`) so it stays tracked.
2. **Fixed**: stale cross-references in \`assign_path_all_noop\`'s own doc comment and \`docs/compliance/yq/limitations.md\` still described the mid-chain gap as open after this PR closed it — updated both, added a proper "Fixed by #1857" entry.
3. **Assessed, not changed**: the double-walk concern (running both the original total-noop check and the new RHS-unused superset check). Traced through: the overwhelmingly common "real container, real write" shape (\`.items[].field = v\` on real objects) fails both checks at the first element via the same O(1) short-circuit \`.all()\` already provides — the added cost only manifests for a null-padding-then-real-write shape that's unusual in practice. Collapsing to one combined walk would require either deleting \`assign_path_all_noop\`/\`yq_assign_is_total_noop\` (many doc-comment cross-references elsewhere in the file, including unrelated functions) or threading extra state through the recursive walk — more risk than the narrow inefficiency warrants right now.

Full verification suite (tests, both clippy legs, no_std, doc, fmt) green.

newhoggy:
Round 3 (multiple finder agents) confirmed the round-2 duplication/efficiency concern was real, and added a sharper argument: \`set_path\`'s own Field/Index/Iterate semantics were being independently re-derived by the predicate twins — a third divergence surface beyond twin-vs-twin. Addressed in ddf8fc748:

**Merged the two independent boolean walks into one 3-way classification** (\`classify_yq_assign_prefix\` → \`PathAssignOutcome::{TotalNoop, RhsUnusedButChanges, NeedsRhs}\`). \`yq_assign_is_total_noop\`/\`yq_assign_rhs_unused\` keep their exact names/signatures as thin wrappers (every existing cross-reference stays accurate); \`assign_path_all_noop\`/\`assign_path_rhs_unused\` had no other callers once this existed, so they were removed rather than kept as dead code, with their doc history folded into the surviving function. \`yq_assign_noop_check\` now computes each path's outcome once and reuses it for both decisions instead of walking twice.

**Also fixed**: a doc-comment placement bug from my own previous commit (an incomplete edit left a substantial pre-existing doc comment orphaned and silently reattached to the wrong item — moved back to where it belongs), \`yq_assign_skip_rhs\`'s stale doc comment (predates #1857, never mentioned the new case), and a stale function-name reference in an unrelated enum's comment.

**Filed #1922** for one more closely-related gap found while live-probing this round: a *negative* Index reached through a null-autovivified array (the \`Index\`/\`IndexNumber\` sibling of the \`Iterate\` arm this PR fixes) still evaluates the RHS — confirmed narrower than #1857's own shape (a *positive* index already needs the RHS for real, via array padding) and pre-existing on \`main\`.

Full verification suite (tests, both clippy legs — including a wrong-self-convention error clippy caught and I fixed, no_std, doc, fmt) green.

newhoggy:
Round-4 findings (a second pass on the merge), all addressed in 5dedbba0f:

1. **Real oversight caught**: the previous commit built the merged 3-way \`classify_yq_assign_prefix\` walk but never actually updated \`yq_assign_noop_check\` (its one real caller) to stop calling the two old wrapper functions separately — the double walk I thought I'd fixed was still there, just hidden behind wrappers. Fixed: each resolved path is now classified once, and the cached outcome is reused for both decisions.
2. Since \`yq_assign_is_total_noop\`/\`yq_assign_rhs_unused\` (and their own inner twins) had no callers left after that fix, removed them rather than keep dead wrappers around — folded their doc history into the surviving \`classify_yq_assign_prefix\`/\`yq_assign_classify\`. Left the ~11 purely-historical prose mentions of the old names elsewhere in tests untouched (accurate narrative about #1233/#1432's own history); only updated the handful making present-tense claims about current code.
3. **New gap found and fixed**: an *absent* field before a mid-chain \`Iterate\` had the identical eager-RHS bug as an explicit \`Null\` — the \`Field\` arm's \`None\` case gave up immediately instead of ever consulting what \`rest\` does with an autovivified \`Null\`. \`.a.b[].z = error("boom")\` on \`{}\` now correctly matches real yq. Verified this is a pure widening (checked by hand-trace that the base case — an actual write target — still needs the RHS regardless of how many absent/\`Null\` levels precede it) and live-tested against yq v4.53.3.
4. Fixed a doc-comment placement bug in my *own* previous commit: an incomplete edit had left a real, substantial pre-existing doc comment orphaned and silently reattached to the wrong item.

Full verification suite (tests, both clippy legs, no_std, doc, fmt) green, plus extensive live re-verification against yq v4.53.3 covering every case fixed across all four commits in this PR.

newhoggy:
Round 5 (final pass) found two concrete, safe cleanups, both applied in b0f7bbb2f:

1. **Fixed**: the placeholder-write loop iterated every resolved path whenever \`Skip\` applied, including \`TotalNoop\` ones that need no write at all — now only writes through paths whose outcome is specifically \`RhsUnusedButChanges\`.
2. **Fixed**: the \`Field\` arm's two near-duplicate "step into \`Null\`" recursive-call sites (absent key vs. explicit \`Null\`) collapsed into one shared child-resolution step — removing exactly the kind of doubled logic this PR's own earlier commits were trying to eliminate elsewhere.

The remaining findings from this round (the ~16 historical prose mentions of the pre-merge function names in unrelated test doc comments, and the two single-use \`is_total_noop\`/\`rhs_unused\` wrapper methods) were deliberately left as-is: the former are accurate narrative describing #1233/#1432's own history using the identifiers that existed at the time, not stale claims about current architecture; the latter were flagged by the finder itself as "low value either way." The architectural suggestion (driving classification off \`set_path\` itself via a dry-run mode, rather than an independently-maintained model) is a legitimately bigger redesign than this PR's scope — noted, not pursued here.

Full verification suite (tests, both clippy legs, no_std, doc, fmt) green, plus a full live re-verification pass against yq v4.53.3 covering every case fixed across all six commits in this PR. This PR has now been through 5 review rounds; proceeding to merge.

PR #1927, opened 2026-08-30
fix(jq): guard eval_as/eval_reduce/eval_foreach's decode-failure sweep

### LINKED ISSUE #1902: jq: eval_as/eval_reduce/eval_foreach's bound-value conversion silently corrupts on decode failure
**Severity: Low-Medium** (worse than #1832's own bug class: no error surfaces at all in some shapes, not merely misprioritized against a later signal)

## Summary

Found during #1832's code review (which fixed the terminal-control prefix
construction in `eval_comma`/`eval_fanout`/`eval_pipe`, plus `eval_fanout`'s
own main promotion path). A further, independently-discovered instance of
the same unchecked-`to_owned` corruption shape (#1746/#1755/#1790/#1832)
exists in `eval_as`'s bound-value conversion, and the identical shape in
`eval_reduce`'s `input_values`/`init_values` and `eval_foreach`'s
equivalent.

## Repro (verified live via the public `succinctly::jq::eval` library API,
the same evaluator #1832 patches)

```
$ query!(b"{\"a\": \"\xff\xfe\", \"b\": 5}", "(.a, .b) as $v | $v", ...)
QueryResult::ManyOwned([String(""), Int(5)])   # no error raised at all
```

This is **worse** than the bug shape #1832 fixed: there, an undecodable
element was misprioritized against a *later* competing signal (still
raised, just the wrong one). Here, the corrupted value silently becomes
part of the final output with no error whatsoever.

A second, narrower axis of the same function reproduces #1832's own
"later error should not win over an earlier undecodable value" ordering
bug:

```
$ query!(b"{\"a\": \"\xff\xfe\", \"b\": 5}",
         "(1,2) as $v | (if $v==1 then .a else error(\"boom\") end)", ...)
QueryResult::Partial([String("")], Error("boom"))   # should be the decode failure, not "boom"
```

## Discovery context

`to_owned_checked`'s own doc comment (`src/jq/eval.rs`, ~line 588) already
documents that roughly 50 more call sites in this file share the general
unchecked-`to_owned` shape, explicitly deferred as a documented gap rather
than attempted in one sweep. This is one specific, concretely-reproduced
instance of that broader population — found adjacent to #1832's own fix,
same as #1897 (`eval_index_expr`) and #1898 (`push_truthiness`).

## Suggested fix direction

Route `eval_as`'s bound-value conversion (and the equivalent sites in
`eval_reduce`/`eval_foreach`) through `to_owned_checked`/
`promote_borrowed_checked`, propagating the resulting `EvalError` the same
way #1832's fixed functions do — including the "earlier undecodable value
preempts a later competing signal" ordering rule for the second repro
shape above. Not investigated further than confirming the repros and
locating the mechanism. Filing to make this trackable, not blocking.

newhoggy:
## Summary

`eval_as`'s bound-value conversion, and the equivalent input/INIT
conversions in `eval_reduce`/`eval_foreach`, used unchecked `to_owned` —
an undecodable value silently became `""` instead of raising, and
(worse) let a later bound value's ordinary error win over an earlier
bound value's own decode failure (the exact ordering bug #1832 already
fixed for `eval_comma`/`eval_fanout`/`eval_pipe`).

Routes all three through `to_owned_checked`/`promote_borrowed_checked`,
folding a checked-conversion failure into the existing `Partial`-control
plumbing each function already has (a decode failure preempts a later
bound value/input the same way an ordinary `Partial`'s trailing control
does, and `eval_reduce`'s `input` keeps its established "single-shot,
drop-the-prefix" policy for an ordinary error).

Also fixes `finish_fork` (shared by `eval_reduce`/`eval_foreach`) to
never suppress a decode failure under an ambient `optional`, matching
`finish_fork_generic`'s already-correct guard in `eval_generic.rs` —
whose own doc comment claimed to mirror this function, which had
drifted from it. Without this, a decode failure raised via the fixes
above could be wrongly swallowed by `?`.

Fixes #1902

## Test plan

- [x] New regression tests covering every match arm the fix touched:
      the `One`/`Many` success and failure shapes in
      `eval_as`/`eval_reduce`/`eval_foreach`, the "earlier undecodable
      value preempts a later competing error" ordering rule, and
      `finish_fork`'s own decode-failure-vs-`optional` guard
- [x] `cargo test --lib --features cli,regex jq::eval::tests` — 1148 passed
- [x] `cargo test --features cli,simd,regex,serde --no-fail-fast` — all green
- [x] `cargo fmt --check`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --all-features --no-deps`
- [x] `omni-dev coverage diff` — 98.86% patch coverage (82/83 new lines);
      the one remaining line is exercised by
      `test_finish_fork_still_suppresses_ordinary_error_under_optional_1902`
      (a coverage-instrumentation artifact for a single-expression
      guarded match arm — the test's own passing assertion proves it ran)

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -3216,10 +3216,20 @@ fn each_label<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
 fn materialize_bound_values<W: Clone + AsRef<[u64]>>(
     bound_result: QueryResult<'_, W>,
 ) -> Result<(Vec<OwnedValue>, Option<Control>), Flow> {
+    // #1902: to_owned_checked/promote_borrowed_checked, not to_owned -- the
+    // lazy twin of the same bug `eval_as`'s own bound-value conversion had.
+    // A checked-conversion failure folds into the trailing control exactly
+    // like an ordinary `Partial`'s control (the `Partial` arm below).
     match bound_result.materialize_cursor() {
-        QueryResult::One(v) => Ok((vec![to_owned(&v)], None)),
+        QueryResult::One(v) => match to_owned_checked(&v) {
+            Ok(v) => Ok((vec![v], None)),
+            Err(e) => Ok((Vec::new(), Some(Control::Error(e)))),
+        },
         QueryResult::OneCursor(_) => unreachable!("materialize_cursor removes OneCursor"),
-        QueryResult::Many(vs) => Ok((vs.iter().map(to_owned).collect(), None)),
+        QueryResult::Many(vs) => match promote_borrowed_checked(vs) {
+            Ok(vs) => Ok((vs, None)),
+            Err((prefix, e)) => Ok((prefix, Some(Control::Error(e)))),
+        },
         QueryResult::Owned(v) => Ok((vec![v], None)),
         QueryResult::ManyOwned(vs) => Ok((vs, None)),
         QueryResult::None => Err(Flow::Exhausted),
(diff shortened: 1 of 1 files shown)


newhoggy:
Multi-round \`/code-review\` (9+ finder passes) found the initial fix was incomplete and had several test-quality issues, all fixed in the two follow-up commits on this branch:

- \`materialize_bound_values\` (the shared lazy-bind helper backing \`first\`/\`limit\`/\`isempty\` over an \`as\` bind) and \`eval_as_pattern\` (the destructuring twin, \`. as [$v]\`/\`. as {k:$v}\`) were initially missed — same construct, different spellings, still silently substituting \`""\`. Fixed with the same pattern as \`eval_as\` itself.
- \`eval_foreach\`'s \`init_values.is_empty()\` early return, and its INIT \`Error\`/\`Break\` arms, both discarded \`input_control\` (including a decode failure) instead of routing through \`finish_fork\`.
- The \`finish_fork\` test used a parsed query that appeared to exercise the new guard but actually passed via \`eval_try\`'s own independent catch, with \`optional == false\` reaching \`finish_fork\` — confirmed by reverting the guard entirely and finding the whole suite still green. Replaced with direct calls to \`finish_fork\`, matching this file's own precedent for a guard that's live-unreachable via any current syntax but pinned against the function's contract regardless.
- Two INIT regression tests used \`[.]\` (array construction, which already raises via its own pre-existing check before this fix's own conversion ever runs) instead of the actual repro (bare \`.\`); an over-permissive \`Halt\` alternative in the positive control; two comments that misstated \`eval_comma\`'s promotion trigger and this sweep's own coverage scope.

Filed #1934 for what the review found but is out of scope for this PR: a deeper decode-failure-vs-later-ordinary-error ordering question in all three functions' loops (real, but needs its own investigation — no live jq/yq syntax reaches the paths involved today, confirmed by reverting guards and finding the whole test suite unaffected), and several DRY/consolidation opportunities (\`materialize_bound_values\`/\`push_owned_values\`/\`stream_outputs\` sharing a checked variant, and a \`to_owned\`/\`to_owned_checked\` naming-inversion suggestion now that the checked variant outnumbers the unchecked one 1.6:1).

Patch coverage: 99.08% (216/218 new lines); the 2 remaining lines are a defensive \`panic!\` fallback that only fires on test failure.

Issue #1877, opened 2026-08-30
jq: nested terminal slice with `|= empty` errors where jq no-ops


newhoggy:
**Severity: Low** (requires a nested slice whose update filter's own output is `empty`)

## Summary

`(.a[0:1] | .[0:2]) |= empty` and `.a[0:1][0:2]? |= empty` on `null` raise `A slice of an array
can only be assigned another array` in succinctly, but real jq treats the whole expression as a
no-op (`null` unchanged), because jq's `_modify`/`|= empty` semantics delete/no-op on empty
update output rather than writing a literal `null` back that then fails an array-shape check.

## Repro (verify against pinned jq 1.7.1 before implementing)

```
$ echo null | jq -c '.a[0:1][0:2]? |= empty'
null   # (expected — verify live before relying on this)

$ echo null | succinctly jq -c '.a[0:1][0:2]? |= empty'
jq: error (at <stdin>:1): A slice of an array can only be assigned another array
```

## Discovery context

Found live during #1873's code review — this is the *terminal* nested-slice arm
(`terminal_write: true`, unconditional), not the mid-chain `Null` arm #1873 fixed, so it's a
different code path with the same symptom shape. Confirmed reproducible identically on
`origin/main` (pre-#1873) — not a regression, and `terminal_write: true` is explicitly excluded
from #1873's fix, so this was never in that fix's scope.

## Scope note

A real fix likely means `|=`'s own filter-produces-`empty` handling (wherever `update_path`
distinguishes "no output" from "output is the value to splice in") needs to reach `through_slice`
before it forces the array-shape check, similar in spirit to how `?` is threaded through but for
a different signal (empty output, not a caught error). Filing to make this trackable, not
blocking.

newhoggy:
See #1894's comment thread — investigated while looking for the next pickup and found the actual root cause is broader than this issue's own framing: `update_path`'s general (non-slice) `Expr::Identity` arm has the identical `|= empty` → delete gap already self-documented as "a separate, pre-existing bug" (confirmed live: `{"a":1} | .a |= empty` gives `{"a":null}`, not jq's `{}`). This issue's slice-specific symptom and #1894's are both downstream of that same missing general mechanism, not independent `through_slice` bugs. Not attempting the fix now (genuinely Tier 3, needs a design pass covering all three shapes together) — leaving in `Ready` with the wider scope noted.

newhoggy:
[...]

PR #1929, opened 2026-08-30
fix(jq): update_path deletes on `|= empty` instead of writing null

### LINKED ISSUE #1916: jq: `.a |= empty`/`.[0] |= empty`/`.[] |= empty` write `null` instead of deleting the key/element
**Severity: Low** (requires an update filter, `|=`/`+=`/etc., that produces literally zero outputs)

## Summary

Real jq's `_modify` (the shared machinery behind `|=` and friends) falls back to `delpaths` when the update filter yields no output at all — it doesn't write a literal `null`. `update_path`'s terminal `Expr::Identity` arm in `src/jq/eval.rs` always does `*root = outputs.into_iter().next().unwrap_or(OwnedValue::Null)`, which can't distinguish "the filter produced empty" from "the filter produced the value `null`" once it reaches `Field`/`Index`/`Iterate` targets — both collapse to writing `null` in place, where jq deletes the key/element (or, for `Iterate`, drops it from the containing array/object).

```console
$ echo '{"a":1}' | jq -c '.a |= empty'
{}
$ echo '{"a":1}' | succinctly jq -c '.a |= empty'
{"a":null}

$ echo '[1,2,3]' | jq -c '.[0] |= empty'
[2,3]
$ echo '[1,2,3]' | succinctly jq -c '.[0] |= empty'
[null,2,3]

$ echo '[1,2,3]' | jq -c '.[] |= empty'
[]
$ echo '[1,2,3]' | succinctly jq -c '.[] |= empty'
[null,null,null]
```

A slice-addressed path (`.[0:2] |= empty`) already gets this right as of #1877/#1894 — this issue is the exact same underlying jq mechanism (`_modify`'s empty-output-means-delete fallback), just for the `Field`/`Index`/`Iterate` targets that fix's own doc comments explicitly called out as separate, pre-existing, and not attempted.

## Discovery context

Found live while implementing #1877/#1894 (`through_slice`'s own `|= empty` handling) — that fix already threads a `wrote: bool` signal through `update_path`'s return value (`Result<bool, EvalEscape>`, `false` meaning "the terminal filter produced no output"), but only `through_slice`'s `Null`/`Array`/`String` arms consume it today. `Field`/`Index`/`Iterate` still just write the literal `null` fallback regardless of what the signal says.

## Suggested fix direction

The `wrote: bool` signal already exists at every `update_path` call site (added by #1877/#1894) — this issue is "wire it up" rather than "invent it": `Expr::Field`'s terminal arm would need to `map.shift_remove(name)` instead of writing `null` when the recursive call reports `wrote == false`; `Expr::Index`'s terminal arm would need to remove the element and shift the array down instead of overwriting it with `null`; `Expr::Iterate` is the trickiest, since jq's actual behavior (`[]` from a 3-element input) means *every* element resolving to empty removes itself, which doesn't fit the single aggregate `Ok(true)`/`Ok(false)` this function currently returns for the whole loop — likely needs per-element removal during the loop (an `retain`-style pass) rather than a single boolean decision after it.

The mid-chain (`Pipe` arm) copies of `Field`/`Index`/`Iterate` would need the identical treatment, not just the terminal ones — this codebase's established pattern (see `update_path`'s own `#1428`/`#1181`-era comments) is that mid-chain and terminal arms are separately-maintained duplicates of each other, so a real fix has to touch both.

## Related

- #1877 / #1894 — where the `wrote` signal was introduced, scoped narrowly to `through_slice`'s own arms
- `update_path`'s `Expr::Identity` arm in `src/jq/eval.rs` — where the `null`-fallback originates

newhoggy:
## Summary

`.a |= empty`, `.[0] |= empty`, and `.[] |= empty` left the target as `"key":null` / an unchanged-length array of nulls instead of deleting the key/element, the way real jq's `_modify` does (it falls back to `delpaths` whenever the update filter produces no output at all).

`update_path`'s terminal `Identity` arm already computed a `wrote: bool` signal for exactly this case (added by #1877/#1894 for `through_slice`'s slice arms), but the `Field`/`Index`/`Iterate` arms above it never consumed it — they always kept the just-written `Null`. This PR wires that signal through:

- **`Field`**: deletes the key when `wrote` is false, restoring the root to `Null` if autovivifying it created the only content (`null | .a |= empty` stays `null`, not `{}`).
- **`Index`**: bounds-checking is deferred behind the filter, mirroring `delete_at_path`'s own `Index` arm — jq resolves an index leniently for *reading* (`getpath`-style: out of range in either direction reads `null`) and only applies the write-time bounds check once a real value is about to land. Confirmed live: `.[-5] |= 99` raises `"Out of bounds negative array index"`, but `.[-5] |= empty` silently no-ops — the error lives on the write path, never the delete path an empty filter takes instead.
- **`Iterate`**: rebuilt in one pass for both array and object roots (not removed in place, which would be quadratic).

Every case in the description above, plus the full matrix in the new test, is verified live against jq 1.7.1. Also updates two existing tests that had pinned the old, wrong output as an intentional characterization of this exact bug (one flagged "if this is ever fixed, update this test").

Fixes #1916

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` (full suite, incl. new `test_field_index_iterate_delete_on_empty_update_filter_1916`)
- [x] `cargo fmt --check`
- [x] Extensive live differential testing against pinned jq 1.7.1 (36+ cases) covering the deferred-bounds-check edge case, nested chains, and object/array iterate

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -18461,9 +18461,11 @@ fn get_path_mut<'a>(
 /// elements), reports its own outcome directly. Consumed in three places:
 /// `through_slice`'s `Null`/`Array`/`String` arms (to distinguish `|=
 /// empty` from a literal `null` write), the mid-chain `Pipe` arm's
-/// "stranded" autoviv-undo check, and -- since #1916 -- this function's own
-/// terminal `Field`/`Index`/`Iterate` arms, which delete the key/element
-/// outright on `false` instead of leaving it `null`.
+/// "stranded" autoviv-undo check, and -- since #1916, jq mode only -- this
+/// function's own terminal `Field`/`Index`/`Iterate` arms, which delete the
+/// key/element outright on `false` instead of leaving it `null`. yq mode
+/// keeps its pre-#1916 behavior at those three arms unchanged; see the
+/// `Field` arm's own comment for why.
 fn update_path<S: EvalSemantics>(
     root: &mut OwnedValue,
     path_expr: &Expr,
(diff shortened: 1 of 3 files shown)


newhoggy:
Code review (two independent agents) plus my own live differential testing against yq v4.53.3 found the initial fix regressed yq mode — it applied the new delete-on-empty-filter behavior unconditionally across both modes, but real yq has no `delpaths`-fallback concept at all. Pushed a follow-up commit gating all three terminal arms behind `S::TAG != EvalTag::Yq`, restoring yq mode to its exact pre-#1916 behavior (verified byte-identical against a pre-#1916 build across every case the reviews raised), and added a yq-mode regression test. jq mode is unaffected. Details in the latest commit message.

PR #1942, opened 2026-08-30
fix(yq): memoize select()'s corruption walk so aliases stop amplifying (#1804)

### LINKED ISSUE #1804: yaml: unmemoized alias resolution inside as_object/as_array can amplify a deeply-aliased document exponentially
## Summary

`YamlValue::as_object()`/`as_array()` (`src/yaml/light.rs`) resolve through an `Alias` chain via `resolve_alias_chain()`/`resolve_alias_target_cursor()` with no memoization. Any traversal that recurses through these (e.g. `to_owned_cursor_at_depth`, and now `push_generic_truthiness_cursor_error`, #1645/PR #1798) re-walks the alias chain from scratch at every node it visits.

## Failure scenario

A YAML document using a "billion laughs"-style repeated-aliasing pattern:

```yaml
a0: &a0 [1, 1]
a1: &a1 [*a0, *a0]
a2: &a2 [*a1, *a1]
...
aN: &aN [*a(N-1), *a(N-1)]
```

Each level doubles the number of nodes a full traversal must visit relative to the previous level, so a document with N such levels (well within `MAX_NESTING_DEPTH = 256`'s structural-depth limit — the chain here is *referential* fan-out, not structural nesting depth) causes an O(2^N) node-visit blowup for any operation that walks the whole subtree (`to_owned`, `-S`, `select()`'s new nested-corruption check, etc.), even though the document's own on-disk/in-memory representation is small (each level is O(1) bytes).

## Scope note

This is **not new** — `to_owned_cursor_at_depth` has always walked through `as_object`/`as_array` this way, so `select()` (which called `to_owned_cursor(&c)` unconditionally before #1645) was already exposed to this amplification for every condition value. #1645/PR #1798 doesn't introduce or widen this exposure; it was flagged during that PR's review as a pre-existing risk worth its own tracked issue, and confirmed to still exist by 2 independent review angles.

## Suggested approach

Not prescribing a fix — options include memoizing per-cursor alias resolution (cache the terminal cursor once computed), or capping total visited-node count (distinct from the existing structural `MAX_NESTING_DEPTH` guard, which doesn't bound referential fan-out) during a single top-level traversal.

## Provenance

Found during multi-angle `/code-review` of PR #1798 (#1645).

newhoggy:
Fixes #1804 for the walk it names. Its sibling path (`to_owned_cursor_at_depth`) is measured, characterised, and deliberately left open — see below.

## The bug

A YAML alias makes one node reachable by many paths, so a billion-laughs shape has 2^N reachable nodes in O(N) bytes on disk:

```yaml
a0: &a0 [1, 1]
a1: &a1 [*a0, *a0]
...
a26: &a26 [*a25, *a25]
```

`select()`'s #1645 corruption walk re-walked every reference from scratch. On a ~450-byte file, answering a two-byte query:

| | `.a26 \| select(.) \| length` |
|---|---|
| real yq | **~0.007s** |
| `main` | **13.894s** |
| this PR | **0.005s** |

Real yq stays flat across N=12..26, so this is succinctly's own bug, not a shared property of YAML aliasing. It's a denial-of-service shape: the file is small enough to paste.

## The fix

The walk carries the set of node positions already proven clean. Sound because the check is a pure function of a node's subtree, and a position is recorded **only on the path that returns `None`** — any error returns before the recursion continues — so "already seen" can only mean "already proven clean". A node with no `text_position` is never memoized.

The aliases themselves sit at distinct positions and don't collide; what collides is their *children*, which resolve to the same target node. That's where the fan-out is cut.

## I nearly fixed the wrong thing

My first measurement showed **no improvement**, because I timed `.aN | select(. != null)` — where the *comparison* materializes its operand and dominates, hiding the fix completely. Attributing per query separated them:

| query | cost | why |
|---|---|---|
| `select(true)` | flat | a scalar condition never walks |
| `select(.)` | exponential | **the walk this PR fixes** |
| `. != null` | exponential | the comparison — no `select()` involved at all |
| `tostring` | exponential | `to_owned` |

Only the second is fixed here. The other two run through `to_owned_cursor_at_depth`, which #1804 also names but which is a *different* function with a different question — it must actually build the value — so it stays open there rather than being quietly claimed.

## Tests

`test_aliased_document_does_not_amplify_1804` takes ~14s against `main` and 5ms here. `test_alias_memo_still_reports_corruption_1804` pins that a corruption reachable through an alias still raises, including when a clean reference to a different anchor was walked first — the failure mode a memo could plausibly introduce.

## Gates

6547 tests / 0 failed; both clippy feature sets, `fmt`, `--no-default-features` (load-bearing here — the fix adds an `alloc` import, and `BTreeSet` needed the same dual `cfg(test)`/`cfg(not(test))` treatment `Rc` has), and CI's exact rustdoc command.

### DIFF
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -12,6 +12,8 @@
 #[cfg(not(test))]
 use alloc::boxed::Box;
 #[cfg(not(test))]
+use alloc::collections::BTreeSet;
+#[cfg(not(test))]
 use alloc::format;
 #[cfg(not(test))]
 use alloc::rc::Rc;
@@ -21,6 +23,8 @@ use alloc::string::{String, ToString};
 use alloc::vec;
 #[cfg(not(test))]
 use alloc::vec::Vec;
+#[cfg(test)]
+use std::collections::BTreeSet;
 #[cfg(test)]
 use std::rc::Rc;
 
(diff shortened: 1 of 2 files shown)


newhoggy:
Closing unmerged — see #1804 for the full measurements. The memo works (13.9s → 0.006s at N=26) but costs ~50% on alias-free documents from plumbing alone, and only closes one of several reachable entry points. The fix belongs at the alias-resolution layer in `yaml/light.rs`, where JSON and alias-free YAML pay nothing.

Issue #1559, opened 2026-08-23
jq: eval_owned_expr_opt still drops a trailing escape and array-collapses in path context (#1277 cluster 4)


newhoggy:
**Severity: Low** (no user-visible divergence found yet — structural gap, see below)

## Summary

The last unaddressed cluster of [#1277](https://github.com/rust-works/succinctly/issues/1277). PR #1522 closed clusters 1, 2 and 3 by giving generator-argument builtins a real fan-out (#1279), but cluster 4 was an explicit non-goal there and is being filed separately so closing #1277 does not lose it.

`eval_owned_expr_opt` (`src/jq/eval.rs`) discards its argument's trailing `Control`:

```rust
fn eval_owned_expr_opt<S: EvalSemantics>(...) -> Result<Option<OwnedValue>, EvalEscape> {
    eval_owned_expr_full::<S>(expr, input, optional)
        .map(|opt| opt.map(|(v, _trailing)| v))     // <- dropped
        ...
}
```

and delegates to `eval_owned_expr_full`, which additionally **array-collapses** 2+ outputs into a single `OwnedValue::Array` rather than truncating — a different loss from the `result_to_owned` family's "take the first".

## Call sites

All four are inside `eval_pipe_with_path_context_internal`, which is why #1277 asked for them to be done together: the function has its own `partial()` / `accumulate_path_context_step` / `continue_rest_with_context` machinery, and fixing one arm in isolation would leave its `Partial` handling inconsistent.

| arm | what it handles |
|---|---|
| `Builtin::ParentN`'s `n` argument | the `parent(n)` succinctly extension |
| `Expr::Builtin(_)` | the arm #1277 calls `eval_builtin_owned` |
| `Expr::Object(_) \| Expr::Array(_) \| Expr::Literal(_)` | value-constructing expressions that reset the path context |
| the generic `_` fallback | "other expressions", which lose path context |

## No user-visible repro found yet

Probed against pinned jq 1.7.1 while closing the rest of #1277; every jq-comparable shape agreed:

```
path(.a | getpath((["b"], error("x"))))          SAME   (getpath has its own resolve_node arm)
path(.a | ltrimstr(("x", error("boom"))))        SAME
path(.a | has(("b", error("boom"))))             SAME
path(.a | {x: (1, error("boom"))})               SAME
del(.a | ltrimstr(("x", error("boom"))))         SAME
(.a | ltrimstr(("x", error("boom")))) = 1        SAME
```

[...]

Issue #1277, opened 2026-08-19
jq: remaining result_to_owned/eval_owned_expr_ctrl callers still drop a trailing escape


newhoggy:
**Severity: Low**

## Summary

Follow-up to #1164, which fixed 17 of the ~43 `result_to_owned`/`eval_owned_expr_ctrl` callers whose own builtin computation is a single, simple use of the argument's value with no other generator involved (see that PR, #1275, for the full list and the `result_to_owned_ctrl`/`eval_owned_expr_ctrl_full`/`finish_owned`/`finish_result` machinery it added). The remaining ~24 sites need real, separate design work per-cluster, not a mechanical extension of that fix:

## Clusters

1. **`indices`/`index`/`rindex`** (`src/jq/eval.rs`) — have several scattered internal `return` points across nested `match` blocks. Wrapping the success path safely needs restructuring each function to compute into a single local binding first (the same pattern #1164 used for `has`/`load`), not a one-line change.

2. **Regex pattern+flags family** (`eval_regex_pattern_and_flags`, `eval_regex_pattern_and_concat_flags`, `eval_sub_replacement`) — each has a **second, independent escape source**: the flags argument (evaluated separately from the pattern, in a specific order #938 established), or — for `eval_sub_replacement` specifically — the replacement expression, re-evaluated once per regex match inside a loop. A trailing escape from *either* argument interacting with #1164's fix isn't verified, and `eval_sub_replacement`'s case additionally needs reconciling an escape against a partially-built output string across the match loop.

3. **`eval_limit`/`eval_nth_expr`/`resolve_limit`** (`limit(n; expr)`/`nth(n; expr)`/path-tracking `limit`) — each already has bespoke `Partial`-reconciliation logic of its own for the *`expr`* argument (matching jq's "never asks past n outputs" semantics). Threading `n`'s own trailing escape through that existing logic, including the early-return "no limit" branches, needs its own design pass to merge correctly rather than risk double-counting or dropping the interaction.

4. **`eval_builtin_owned`** — turned out (on investigation) to be embedded inside the much larger `eval_pipe_with_path_context_internal`, which already has its own `partial()`/`accumulate_path_context_step`/`continue_rest_with_context` machinery for *other* Partial-producing arms in that same function. Fixing this one call site in isolation would leave the function's own Partial-handling inconsistent; the other 2-3 similarly-shaped call sites in that function (`eval_owned_expr::<S>` at other match arms) likely need updating alongside it.

[...]

PR #1949, opened 2026-08-30
fix(jq): eval_owned_expr_full takes first output instead of array-collapsing

### LINKED ISSUE #1937: jq: eval_owned_expr_opt's array-collapse silently discards fan-out entirely when path context is nearby (#1277 cluster 4, #1559 follow-up)
Found while investigating #1559 (merged in the linked PR, which fixed the trailing-`Control`-drop half of this same helper).

## The gap

`#1277`'s clusters 1-3 (closed by #1522/#1279) gave generator-argument builtins real fan-out: a builtin whose argument is a generator now produces one output per argument output, matching real jq.

`eval_owned_expr_opt`/`eval_owned_expr_full` (`src/jq/eval.rs`) back the 4 call sites #1277 declared an explicit non-goal (`ParentN`'s `n` argument, the `Expr::Builtin(_)` arm, the `Expr::Object`/`Array`/`Literal` arm, and the generic `_` fallback -- all inside `eval_pipe_with_path_context_internal`). When the evaluated expression itself produces 2+ outputs (`QueryResult::Many`/`ManyOwned` with `len() > 1`), these helpers **array-collapse** the outputs into a single `OwnedValue::Array` rather than fanning out or truncating to the first.

## Live-verified divergence (not just a structural gap -- this one has a real repro)

```console
$ echo '{"a":"xax"}' | succinctly jq -c '.a | ltrimstr(("x","z"))'
"ax"
"xax"
$ echo '{"a":"xax"}' | succinctly jq -c '.a | ltrimstr(("x","z")) | key'
"a"
```

The first query (no `key` present, so the ordinary fan-out-aware evaluator handles it) correctly produces 2 outputs, matching real jq's own fan-out for this exact query (confirmed against jq 1.7.1: `echo '"xax"' | jq -c 'ltrimstr(("x","z"))'` also gives `"ax"`/`"xax"`).

The second query is identical except for the trailing `| key`, which forces the whole pipe through `eval_pipe_with_path_context_internal` (since `key` needs path tracking). This routes `ltrimstr(("x","z"))` through `eval_owned_expr_opt` instead of the ordinary evaluator -- the two fanned-out outputs get array-collapsed into `["ax","xax"]`, and `key` then runs *once* against that collapsed array (ignoring its content entirely, since `key` just returns the current path position), producing a single, unrelated output. Both of the real `ltrimstr` results are lost.

`key` itself has no jq oracle (succinctly extension), so this can't be demonstrated as a *jq* divergence in isolation -- but it's a genuine, demonstrated **internal inconsistency**: the exact same sub-expression (`ltrimstr(("x","z"))`) fans out correctly or silently collapses to the wrong single value depending on whether an unrelated `key`/`parent`/`file_index`/`path()` construct happens to be anywhere else in the same pipe.

## Why not fixed alongside #1559

#1559 fixed the *other* half of this same helper (a trailing `Control::Error`/`Break`/`Halt` was silently dropped) -- that was a clean, contained fix (route the existing tuple's second element through instead of discarding it).

This array-collapse gap is different: closing it *properly* means giving these 4 call sites the same real fan-out clusters 1-3 got, which is exactly the "materially larger, too invasive" architecture #1522's own design doc (`docs/plan/jq-generator-argument-fanout.md`) declined to fold into that PR. Speculatively swapping "array-collapse" for "take first" (matching the `result_to_owned` family's sibling policy) wouldn't be *correct* either -- real jq's answer for the underlying sub-expression is 2 separate outputs, and neither array-collapse nor take-first matches that -- so it isn't an obviously-safe drop-in change either.

## Suggested next step

Whoever picks this up should decide between:
- Giving these 4 call sites real fan-out (the "materially larger" fix #1522 declined -- now backed by a live, reproducible symptom rather than just a documented non-goal).
- Or, if fan-out is still out of scope, at least switching the collapse policy to "take first" for consistency with `result_to_owned`'s sibling behavior, and documenting the residual gap explicitly rather than leaving it implicit.

## Related

- #1277 -- the parent issue; clusters 1-3 closed by #1522/#1279, cluster 4 (this helper) explicitly declared a non-goal there.
- #1559 -- fixed this helper's other bug (trailing `Control` silently dropped); this issue is the array-collapse half #1559 deliberately left alone.

newhoggy:
## Summary

`eval_owned_expr_full` (backing `eval_owned_expr_opt`/`eval_owned_expr_ctrl`/`eval_owned_expr_ctrl_full`, used by `eval_pipe_with_path_context_internal`'s `Builtin`/`Object`/`Array`/`Literal`/generic-fallback arms and `ParentN`'s own `n` argument) array-collapsed a 2+-output `QueryResult::Many`/`ManyOwned`/`Partial` into a single `OwnedValue::Array`, instead of fanning out or truncating to the first output.

This made the exact same sub-expression silently disagree with itself depending on unrelated nearby syntax:

```console
$ echo '{"a":"xax"}' | succinctly jq -c '.a | ltrimstr(("x","z"))'
"ax"
"xax"
$ echo '{"a":"xax"}' | succinctly jq -c '.a | ltrimstr(("x","z")) | key'
"a"
```

The first query correctly fans out to two outputs (matching real jq 1.7.1). Appending an unrelated `| key` forces the whole pipe through the path-context-aware slow path instead — the two fanned-out outputs got array-collapsed into `["ax","xax"]` before `key` ran once against it, and both real values were lost.

## Fix

Real fan-out for these 4 call sites is the "materially larger, too invasive" architecture #1522's own design doc already declined once (#1277 cluster 4's explicit non-goal). This PR switches the collapse policy to **"take first"** instead — matching `result_to_owned`'s identical policy and rationale. Still not real jq's own 2-output answer, but at least consistently wrong the same way its sibling function family already is, instead of silently changing shape based on unrelated nearby syntax.

As a small included correctness improvement in the same spirit: an empty `Many`/`ManyOwned` now maps to `Ok(None)` ("zero outputs") rather than an empty array, matching `result_to_owned_full`'s identical handling of the same shape and #1280's own stated purpose for this function (keeping a genuinely empty result distinguishable from a real value).

## Test plan

- [x] Live-verified the fix against the issue's own repro, using the *value* (not just `key`'s output, which doesn't depend on the value): `.a | ltrimstr(("x","z")) | (key, .)` now correctly shows `"ax"` (was `["ax","xax"]`)
- [x] Updated the one existing unit test that pinned the old array-collapse output directly (`eval_owned_expr_ctrl_full_many_shapes_and_multi_value_partial_1164`)
- [x] Added a new CLI-level regression test using the issue's own repro (`test_owned_expr_full_takes_first_output_instead_of_array_collapsing_1937`)
- [x] `cargo test --features cli,simd,regex,serde` (3093 lib + 1148 jq_cli_tests + full suite, all green)
- [x] `cargo fmt --check`
- [x] The #1559 trailing-`Control`-propagation fix (same match arms, an unrelated prior bug) re-verified unaffected: error/break/halt after a partial output all still propagate correctly

Fixes #1937

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -25444,23 +25444,34 @@ fn eval_owned_expr_full<S: EvalSemantics>(
         QueryResult::One(v) => Ok(Some((to_owned(&v), None))),
         QueryResult::OneCursor(_) => unreachable!(),
         QueryResult::Owned(v) => Ok(Some((v, None))),
-        QueryResult::Many(vs) => {
-            if vs.len() == 1 {
-                Ok(Some((to_owned(&vs[0]), None)))
-            } else {
-                Ok(Some((
-                    OwnedValue::Array(vs.iter().map(to_owned).collect()),
-                    None,
-                )))
-            }
-        }
-        QueryResult::ManyOwned(vs) => {
-            if vs.len() == 1 {
-                Ok(Some((vs.into_iter().next().unwrap(), None)))
-            } else {
-                Ok(Some((OwnedValue::Array(vs), None)))
-            }
-        }
+        // #1937: take the *first* output rather than array-collapsing a
+        // multi-output result into `OwnedValue::Array` -- matches
+        // `result_to_owned_full`'s identical policy for the same reason:
+        // real fan-out for this function's own 4 call sites (all inside
+        // `eval_pipe_with_path_context_internal`) is the "materially
+        // larger, too invasive" architecture #1522's own design doc
+        // already declined once (#1277 cluster 4's explicit non-goal).
+        // Array-collapse wasn't a safer middle ground -- it made the
+        // *same* sub-expression silently disagree with itself depending
+        // on whether an unrelated `key`/`parent`/`file_index`/`path()`
+        // elsewhere in the pipe happened to route it through this
+        // function instead of the ordinary fan-out-aware evaluator
+        // (confirmed live: `.a | ltrimstr(("x","z"))` fans out to two
+        // outputs, `"ax"`/`"xax"`; adding a trailing `| key` collapsed
+        // them into one `["ax","xax"]`, then ran `key` once against
+        // that array instead of twice against the real values). Taking
+        // first is still wrong relative to real jq's own 2-output
+        // answer, but it is *consistently* wrong the same way
+        // `result_to_owned`'s sibling family already is, rather than
+        // silently changing shape based on unrelated nearby syntax.
+        // An empty `Many`/`ManyOwned` is the same "zero outputs" case
+        // as the bare `QueryResult::None` arm below, not a distinct
+        // value -- matches `result_to_owned_full`'s own handling of the
+        // same shape, and #1280's whole point (this function's purpose
+        // is keeping a genuinely empty result distinguishable from
+        // `null`, not manufacturing an empty array in its place).
+        QueryResult::Many(vs) => Ok(vs.first().map(|v| (to_owned(v), None))),
+        QueryResult::ManyOwned(vs) => Ok(vs.into_iter().next().map(|v| (v, None))),
         // The one behavior change from `eval_owned_expr_ctrl_full`'s own
         // slow path (#1280): `None`, not `Ok(Some((Null, None)))`.
         QueryResult::None => Ok(None),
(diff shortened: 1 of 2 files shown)


newhoggy:
Closing as duplicate — this and #1950 (`issue-1937-eval-owned-expr-opt-collapse`) both claimed #1937 within ~32 seconds of each other (a genuine race, different branch names for the same issue evaded the atomic-lock check). #1950 implements the same core fix (take-first instead of array-collapse) and additionally includes the `docs/compliance/jq/limitations.md` entry my own code review flagged as missing here, plus a direct unit test for the empty-`Many`/`ManyOwned` case. Deferring to it rather than duplicating merge-queue work.

PR #1950, opened 2026-08-30
fix(jq): eval_owned_expr_opt/full take first output instead of array-collapsing

### LINKED ISSUE #1937: jq: eval_owned_expr_opt's array-collapse silently discards fan-out entirely when path context is nearby (#1277 cluster 4, #1559 follow-up)

newhoggy:
Fixes #1937.

## Summary

`eval_pipe_with_path_context_internal`'s 4 non-fan-out-aware call sites (`ParentN`'s `n`, the `Expr::Builtin(_)` arm, the `Expr::Object`/`Array`/`Literal` arm, and the generic `_` fallback) route through `eval_owned_expr_opt`/`eval_owned_expr_full`. When the evaluated expression itself produced 2+ outputs, these array-collapsed them into a single `OwnedValue::Array` — a second, different wrong answer, since the exact same sub-expression fans out correctly when no path-tracking builtin (`key`/`parent`/`file_index`) happens to be elsewhere in the pipe.

Full fan-out for these 4 sites is the "materially larger, too invasive" fix #1522's own design doc (`docs/plan/jq-generator-argument-fanout.md`) explicitly declined — per the issue's own suggested next step, this PR takes the smaller, scoped fix instead: switch the collapse policy to take-first, matching `result_to_owned_full`'s identical policy for the same `Many`/`ManyOwned`/`Partial` shapes (real jq's own generator-argument desugaring, `f(x)` ~= `x as $b | ...`, drives the rest of a computation off only `x`'s first output). Not a full fidelity fix — documented as a residual, known gap in `docs/compliance/jq/limitations.md` rather than left implicit.

An empty `Many`/`ManyOwned` now also matches `result_to_owned_full`'s "zero outputs" treatment (no output at all, not an empty-array value), for the same consistency reason.

## Changes

- `eval_owned_expr_full`'s `Many`/`ManyOwned`/`Partial` arms: take-first instead of collapse-to-array-when-len-is-not-1.
- New unit test `eval_owned_expr_full_takes_first_not_array_collapse_1937` pinning the `Many`(len>1)/`ManyOwned`(len>1)/empty-`Many` shapes directly.
- New CLI test `test_path_context_builtin_arm_multi_output_takes_first_not_array_collapse_1937`, a content-sensitive repro (`length` after the fanned-out value, since `key` alone is path-only and can't show the difference).
- Updated `eval_owned_expr_ctrl_full_many_shapes_and_multi_value_partial_1164`, which pinned the old collapse behavior for both its `Many` and multi-value `Partial` shapes, to assert take-first instead.
- New `docs/compliance/jq/limitations.md` entry recording the residual gap.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — 3094 lib tests + all integration suites pass
- [x] `cargo test` (default features) — passes
- [x] `cargo clippy --all-targets --features std,simd,serde,cli,regex,bench-runner,large-tests,mmap-tests -- -D warnings` — clean (avoids the `scalar-yaml` blind spot)
- [x] `cargo build --no-default-features` — builds
- [x] Confirmed the new unit test fails against the pre-fix code (temporarily reverted, reran, reverted back) before trusting it as a real regression pin
- [x] Live-verified the underlying `ltrimstr(("x","z"))` fan-out (without `key`) against jq 1.7.1 to confirm it already matches real jq — this PR only touches the path-context-forced narrowing, not that fan-out itself

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -1881,7 +1881,7 @@ protocol for `resolve_node`, the same shape `eval_each_owned` already gives valu
 evaluation) rather than attempted here; the narrow `first`/`limit` fixes already close
 the two shapes named in #1906/#1935's own repros.
 
-### A generator-argument expression fans out normally, then silently narrows to its first output the moment `key`/`parent`/`file_index` shows up anywhere else in the same pipe
+### A generator-argument expression fans out normally, then silently narrows to a single array the moment `key`/`parent`/`file_index` shows up anywhere else in the same pipe
 
 [#1277](https://github.com/rust-works/succinctly/issues/1277)'s clusters 1-3
 (closed by [#1522](https://github.com/rust-works/succinctly/issues/1522)/[#1279](https://github.com/rust-works/succinctly/issues/1279))
@@ -1893,9 +1893,8 @@ arm, the `Expr::Object`/`Array`/`Literal` arm, and the generic `_` fallback
 -- were an explicit non-goal of that fix, since giving them the same real
 fan-out is a materially larger change (`docs/plan/jq-generator-argument-fanout.md`).
 Those 4 sites route through `eval_owned_expr_opt`/`eval_owned_expr_full`,
-which -- until [#1937](https://github.com/rust-works/succinctly/issues/1937)
--- array-collapsed a multi-output expression into a single
-`OwnedValue::Array` instead of fanning out or taking the first output.
+which array-collapses a multi-output expression into a single
+`OwnedValue::Array` instead of fanning out.
 
 ```console
 $ echo '{"a":"xax"}' | succinctly jq -c '.a | ltrimstr(("x","z"))'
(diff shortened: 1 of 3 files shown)


newhoggy:
## /code-review results (4 independent finders)

Convergent, serious finding across 3 of 4 finders (all live-verified against the built binary): the original take-first approach for 2+-output `Many`/`ManyOwned`/`Partial` is a **regression, not a fix**, for this helper's other call sites.

`result_to_owned_full`'s take-first policy is correct for *its* callers (a builtin's single generator argument, consumed exactly once). But `eval_owned_expr_full` also backs:
- `Expr::Builtin(_)` for **zero-arg generator builtins** (`recurse`, `range`, `inputs`, ...) reached via path-context routing, not just argument-fan-out
- The generic `_` fallback, which catches `..`, `paths`, `limit`, and arbitrary comma branches

Confirmed live: `.a | (recurse, key)` on `{"a":{"b":{"c":1}}}` went from `[{"b":{"c":1}},{"c":1},1]` (wrong shape, but complete — the pre-existing bug) to `{"b":{"c":1}}` (2 of 3 outputs silently discarded, no error) under take-first. That's a materially larger, harder-to-notice correctness gap than the narrow `ltrimstr`+`key` case this issue was filed against.

**Fix applied (pushed as a follow-up commit):**
- Reverted the 2+-output case back to array-collapse (the pre-existing, safe-if-wrong-shaped behavior).
- Kept the one genuinely safe piece: an empty `Many`/`ManyOwned` now correctly contributes zero outputs instead of an `OwnedValue::Array([])` value, matching `result_to_owned_full`'s identical `#1045` rule — no downside, since there's nothing to lose by treating zero outputs as zero outputs.
- Added two regression-guard CLI tests (`ltrimstr`+`key`, `recurse`+`key`) pinning the current array-collapse behavior, so a future attempt at take-first gets caught immediately.
- Rewrote the `docs/compliance/jq/limitations.md` entry to record the full investigation — including the rejected approach and exactly why — for whoever eventually attempts the real fan-out fix #1522 already declined as too invasive.

A minor, unrelated finding (a 5th call path via `builtin_envvar`/`eval_owned_expr_ctrl_full`, currently unreachable from any parsed CLI query) and two efficiency observations (comma/iterate fully materializing before a take-first-style consumer would discard the tail) were also raised but don't apply now that take-first was reverted — noted here for the record, not acted on.

Full verification rerun after the revert: `cargo test --features cli,simd,regex,serde` (3094+ lib tests, all integration suites), `cargo test` (default features), both clippy legs (`--all-features` and the SIMD/non-scalar-yaml leg), `cargo build --no-default-features`, `cargo fmt --check` — all clean.

Issue #1953, opened 2026-08-30
jq: to_owned_checked's Err can carry non-decode-failure errors that still bypass optional


newhoggy:
## Summary

Split off from #1907 (item 3), which is being closed after its other four findings were resolved (items 1/5 turned out already fixed by prior work, item 4's two concrete sub-findings fixed in PR #TODO). This one genuinely needs its own dedicated audit — not attempted there.

`to_owned_checked_at_depth`'s `Object` arm can return errors that are **not** decode failures:

- `f.malformed_member_error()` (#1194, a trailing unpaired member)
- a `resolve_display_key` collision error (#1642)

Neither is `is_decode_failure()`-tagged, but every one of the ~33 existing call sites in `src/jq/eval.rs` matching:

```rust
match to_owned_checked(&value) { ... Err(e) => return QueryResult::Error(e) }
```

propagates these the same unconditional way as a genuine decode failure — bypassing `optional`/`?` even though only decode failures are supposed to be uncatchable per #1247/#1620. Most of these call sites pre-date #1820 and are not something any single recent PR introduced; it's `to_owned_checked`'s own established contract from #1755 onward.

## Related

- `eval_error`'s own materialization of `error(msg)` (#1907 item 1, already fixed) deliberately keeps `to_owned_checked`'s malformed-member/collision errors unconditional-on-`optional` too, by design — see that fix's own comment for why conflating the two would be wrong. This new issue is about auditing the ~33 *other* call sites for the same question, not re-litigating that one.
- #1908 (audit ~47 remaining boilerplate `to_owned(&value)` sites, a related but distinct question — that's about *unchecked* `to_owned` silently substituting `""`, not about `to_owned_checked`'s own `Err` variants bypassing `optional`).

## Suggested approach

Not prescribing a fix shape — worth deciding first whether `to_owned_checked` should itself distinguish the two error classes (e.g. a `is_decode_failure()`-style tag already exists on `EvalError`; the question is whether *callers* should branch on it before deciding whether `optional` applies, the way `eval_error`'s new fix does by hand). A blanket "always respect optional for non-decode-failure errors" sweep across ~33 sites risks the same kind of regression #1905's own review caught once already (recursively decode-checking made a container merely *containing* an undecodable string wrongly fail) — needs care, not a mechanical find-and-replace.

newhoggy:
🔧 Claimed by Agatha — branch `issue-1953-optional-respecting-checked`.

[...]

PR #1954, opened 2026-08-30
fix(yq): map/map_values scalar no-op, gate jq-only date builtins

### LINKED ISSUE #1907: jq: #1820 review found adjacent bugs and gaps left out of scope
Follow-up from #1820's review (PR #1905). Several real issues surfaced during a multi-round `/code-review` that are either pre-existing (not introduced by #1905), affect a shared helper with much broader blast radius than #1820's stated scope, or are pure simplification opportunities — recorded here rather than folded into #1905 to keep that PR focused.

## 1. `eval_error`'s `Some(msg_expr)` branch is still unguarded (asymmetric with the `None` branch #1905 fixed)

`error(msg)` (explicit-argument form) materializes `msg`'s value via `result_to_owned_full` (`src/jq/eval.rs` ~line 2532), whose `QueryResult::One(v) => Ok(Some((to_owned(&v), None)))` arm still uses the unchecked `to_owned`. So `error(.)` on an undecodable string silently raises `""` as the payload, and `error(.)?` incorrectly suppresses the (would-be) decode failure — while bare `error` (fixed by #1905) correctly raises uncatchably. `result_to_owned`/`result_to_owned_ctrl`/`result_to_owned_full` back *every* generator-argument builtin, not just `error(msg)`, so fixing this needs its own audit of all callers — Category 2 territory, not a one-line change.

## 2. `builtin_envvar`: decode-checking `.` before evaluating `var` can pre-empt `var`'s own Halt/Break

`builtin_envvar` converts `.` via `to_owned_checked` before calling `eval_owned_expr_ctrl_full(var, &owned_value, optional)`. If `.` is undecodable and `var` would independently escape via `Halt`/`Break` (regardless of `.`'s content — e.g. `var = halt_error(9)`), the decode failure on `.` fires first and the established Halt/Break-precedence contract (the two existing `builtin_envvar_propagates_*` tests) doesn't hold for this combination. Low priority: `Builtin::EnvVar` has no parser construction site (confirmed by this file's own test comments) — reachable only via direct Rust API calls, never through a parsed jq/yq query.

## 3. `to_owned_checked`'s `Err` can carry non-decode-failure errors that still bypass `optional`

`to_owned_checked_at_depth`'s `Object` arm can also return `f.malformed_member_error()` (#1194, a trailing unpaired member) or a `resolve_display_key` collision error (#1642) — neither is `is_decode_failure()`. Every one of the ~33 existing `match to_owned_checked(&value) { ... Err(e) => return QueryResult::Error(e) }` call sites in `eval.rs` (most pre-dating #1820, not introduced by it) propagates these the same unconditional way as a genuine decode failure, bypassing `optional`/`?` even though only decode failures are supposed to be uncatchable per #1247/#1620. This is `to_owned_checked`'s own established contract from #1755 onward, not something #1905 introduced — worth a dedicated look at whether `to_owned_checked` should distinguish the two error classes.

## 4. Pre-existing yq-mode divergences found adjacent to #1905's touched code

- `builtin_map`/`builtin_map_values`'s scalar fallback raises `cannot_iterate_with(...)` unconditionally for both jq and yq mode. Live-tested against yq v4.53.3: real yq passes a scalar through unchanged (`printf '"x"' | yq 'map(.)'` => `x`), it does not error. jq 1.7.1 does error, matching the current branch — so this is yq-only wrong, undocumented in `docs/compliance/yq/limitations.md`. Pre-existing (the branch predates #1905; #1905 only added a decode-failure guard ahead of it).
- `mktime`/`gmtime`/`localtime` have no `reject_unless_jq_extensions` parser gate (unlike `strftime`/`strptime`), so `succinctly yq 'mktime'` succeeds against a parser surface real yq's lexer rejects outright (confirmed live). Bare `error` (no argument) is similarly ungated in yq mode; real yq requires exactly 1 arg.

None of these are regressions from #1905 — they're pre-existing gaps its review happened to surface.

## 5. Simplification: a shared `scalar_fallback` helper

The `if let Some(e) = scalar_decode_failure(&value) { return QueryResult::Error(e); } if optional { return QueryResult::None; } QueryResult::Error(<ctor>(...))` triplet is now duplicated ~22 times across `eval.rs` (6 of those added by #1905 alone). A small helper —

```rust
fn scalar_fallback<W: Clone + AsRef<[u64]>>(
    value: &StandardJson<'_, W>,
    optional: bool,
    err: impl FnOnce() -> EvalError,
) -> QueryResult<'_, W> {
    if let Some(e) = scalar_decode_failure(value) {
        return QueryResult::Error(e);
    }
    if optional {
        return QueryResult::None;
    }
    QueryResult::Error(err())
}
```

— would collapse each into a one-line call. `eval_generic.rs` already has the analogous `owned_or_err!` macro and `decode_failure_or` helper for its own `GenericResult`/`DocumentValue` domain (its doc comment even names `Iterate`/`Map`/`Keys`/`KeysUnsorted`/`ToEntries` as needing exactly this shape) — no `QueryResult`-flavored equivalent has ever been extracted in `eval.rs` itself. Worth doing before Category 2's ~47 remaining sites multiply the duplication further.

## Context

Surfaced during #1905's multi-round review (`/code-review`), which also caught (and #1905 itself fixed before merging) a real regression from an initial `to_owned_checked`-based fix: `eval_not` and `builtin_truncate_stream`'s depth conversion both only ever consult a value's outer type, never decoded content, so recursively decode-checking made a container merely *containing* an undecodable string wrongly fail. See #1905's commit history for the corrected approach (`scalar_decode_failure` + plain `to_owned`, or `json_is_truthy` directly) if any of the above items end up needing the same pattern.

newhoggy:
## Summary

#1907 bundled 5 findings from #1905's own review. This PR resolves what's still genuinely actionable:

- **Items 1 and 5 turned out to already be fixed** by prior work — confirmed by reading the current code, whose own doc comments cite #1907 directly (`eval_error`'s `Some(msg_expr)` branch already uses `to_owned_checked`; `scalar_fallback` already exists and is used at 22 call sites). Stale issue premise, not something to redo.
- **Item 2** (`builtin_envvar`'s decode-check-before-eval precedence) stays deferred — the issue itself flags it low-priority and unreachable via any parsed jq/yq query (`Builtin::EnvVar` has no parser construction site).
- **Item 3** (`to_owned_checked`'s `Err` can carry non-decode-failure errors that bypass `optional` at ~33 call sites) genuinely needs its own dedicated audit — split off as #1953 rather than attempted here.
- **Item 4's two concrete sub-findings** are what this PR actually fixes.

### `map`/`map_values` no-op on a scalar target in yq mode

Real yq's `map`/`map_values` never even run their filter against a scalar target — confirmed live (v4.53.3): `5 | map(error("boom"))` succeeds silently with `5`. jq always errors here ("cannot iterate over..."). `null` gets yq's usual empty-container treatment (`[]`, matching the `*=` merge rule CLAUDE.md already documents); every other scalar passes through byte-for-byte unchanged — no decode-check needed at all, since nothing ever reads its content.

`map` and `map_values` reach the real yq CLI through **two different evaluators**: `map` got its own fast native arm in `eval_generic.rs` back in #725 (bypassing `eval.rs` entirely for the common array/object case), so that's where its fix has to live; `map_values` has no native arm yet and still falls through to `eval.rs`'s `builtin_map_values` via the reindex fallback, so its fix lives there. Worth calling out: my first attempt only fixed `eval.rs`'s `builtin_map`, which turned out to have **zero effect** on any real yq query — caught only by live-testing against the pinned oracle rather than trusting a unit-level fix in isolation.

### Seven jq-only date builtins gated behind `--jq-extensions`

`gmtime`/`localtime`/`mktime`/`todate`/`fromdate`/`todateiso8601`/`fromdateiso8601` were parsed unconditionally in yq mode — confirmed live that real yq's lexer rejects every one of them outright ("invalid input text..."). Gated behind `--jq-extensions` matching the neighboring `strftime`/`strptime` precedent (#1512). `now`/`from_unix`/`to_unix`/`tz(...)` really are yq syntax and stay ungated (verified live too). No existing test or golden fixture exercised any of these seven without the flag, so nothing needed updating there.

## Test plan

- [x] Live-verified every claim above against the pinned yq v4.53.3 binary before writing code, and again after
- [x] New test `test_map_map_values_scalar_target_noops_in_yq_mode_1907` (both builtins, both evaluators, `null` vs. other scalars vs. real containers)
- [x] New test `test_jq_only_date_builtins_gated_behind_jq_extensions_1907` (all 7 rejected without the flag / accepted with it; `from_unix` sanity-checked ungated)
- [x] `cargo test --features cli,simd,regex,serde` (full suite, all green — no existing test relied on the old behavior)
- [x] `cargo fmt --check`
- [x] Updated `docs/reference/yq-language.md`'s gated-builtins table and its stale "jq's standard date functions" framing

Fixes #1907

### DIFF
--- a/docs/reference/yq-language.md
+++ b/docs/reference/yq-language.md
@@ -283,7 +283,9 @@ echo '"aaa"' | succinctly yq 'sub("a"; "X"; "g")'
 
 ### Date/Time Extensions
 
-Beyond jq's standard date functions (`now`, `gmtime`, `strftime`, `strptime`), yq adds:
+Beyond jq's standard date functions (`now`, and `gmtime`/`strftime`/`strptime`
+behind `--jq-extensions` -- see [Gated jq Builtins](#gated-jq-builtins---jq-extensions)),
+yq adds:
 
 | Function        | Description                   | Example                             |
 |-----------------|-------------------------------|-------------------------------------|
@@ -389,6 +391,9 @@ surface (#1512):
 | `pow(base; exp)`                                     | Exponentiation                                |
 | `bsearch(target)`                                    | Binary search index in a sorted array         |
 | `strftime(fmt)`, `strptime(fmt)`                     | Broken-down-time formatting / parsing (#1650) |
+| `gmtime`, `localtime`, `mktime`                      | jq's date functions (#1907)                   |
+| `todate`, `fromdate`                                 | ISO 8601 shortcuts (#1907)                    |
+| `todateiso8601`, `fromdateiso8601`                   | ISO 8601, fixed format (#1907)                |
 | `add`                                                | Sum of an array/stream (#1714)                |
 | `min_by(f)`, `max_by(f)`                             | Extremum by a key function (#1714)            |
 | `implode`                                            | Codepoint array to string (#1714)             |
--- a/docs/reference/yq-language.md
+++ b/docs/reference/yq-language.md
@@ -283,9 +283,11 @@ echo '"aaa"' | succinctly yq 'sub("a"; "X"; "g")'
 
 ### Date/Time Extensions
 
-Beyond jq's standard date functions (`now`, and `gmtime`/`strftime`/`strptime`
-behind `--jq-extensions` -- see [Gated jq Builtins](#gated-jq-builtins---jq-extensions)),
-yq adds:
+Beyond jq's standard date functions (`now` stays ungated; `gmtime`, `localtime`,
+`mktime`, `strftime`, `strptime`, `todate`, `fromdate`, `todateiso8601`, and
+`fromdateiso8601` all need `--jq-extensions` -- see
+[Gated jq Builtins](#gated-jq-builtins---jq-extensions) for the full list), yq
+adds:
 
 | Function        | Description                   | Example                             |
 |-----------------|-------------------------------|-------------------------------------|
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -6792,6 +6792,20 @@ fn builtin_map<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         StandardJson::Object(fields) => {
             map_over::<W, S>(f, fields.map(|fld| fld.value()), optional)
         }
+        // #1907: real yq no-ops a scalar target for `map` entirely -- `f`
+        // never even runs (confirmed live, v4.53.3: `5 | map(error("boom"))`
+        // succeeds silently, `5`) -- unlike jq, which always errors here.
+        // `null` gets yq's usual empty-container treatment instead of a
+        // literal passthrough (matches the `*=` merge rule documented in
+        // CLAUDE.md: "null acts as an empty container on either side of a
+        // yq-mode merge"); every other scalar passes through byte-for-byte
+        // unchanged, so no decode-check is needed here at all -- nothing
+        // ever reads its content (the "uniform fix regressed content-
+        // independent ops" lesson, #1820's own review).
+        StandardJson::Null if S::TAG == EvalTag::Yq => {
+            QueryResult::Owned(OwnedValue::Array(Vec::new()))
+        }
+        _ if S::TAG == EvalTag::Yq => QueryResult::One(value),
         _ => {
             // #1820: sibling `map_over`'s own per-element path is already
             // checked (#1755); this outer type-error fallback for a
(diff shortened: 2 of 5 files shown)


newhoggy:
Code review (independent agent) confirmed both fixes correct and complete, full test suite + both clippy legs clean. Two nits addressed in a follow-up commit (doc completeness, test coverage for `map_values`'s scalar passthrough).

One correction to my own earlier commit message: I claimed an earlier attempt scoped to `eval.rs`'s `builtin_map` alone "silently had no effect on any real yq query" — the review caught that this is only true for the ordinary piped-stdin path (which routes through `eval_generic.rs`'s native `Builtin::Map` arm, added by #725). `eval.rs::builtin_map` is independently reachable — and the fix independently necessary — via `--null-input`/`--raw-input`, which call into `eval.rs`'s evaluator directly (`yq_runner.rs`'s `evaluate_input`), live-verified by the reviewer. The shipped fix is correct either way; just correcting the stated reasoning for the record.

Issue #1960, opened 2026-08-30
yq: flow-mapping plain scalar doesn't terminate at embedded ': ', silently swallowing a sibling key on write


newhoggy:
## Context

Found while investigating #1803 (unifying `to_owned_at_depth`'s sibling traversal
functions) -- not part of that issue, filing separately since the root cause turned
out to be a YAML flow-scalar parsing bug, not traversal-function duplication.

## Summary

A YAML flow mapping missing the comma between two entries --
`{a: 1 b: 2}` -- gets parsed by succinctly as a single flow scalar `a: "1 b: 2"`
instead of raising or otherwise recognizing the second key. On a **write**
operation, this silently and permanently discards the `b: 2` field with no
error at all.

## Repro (confirmed against real yq v4.53.3, the pinned oracle)

```console
$ printf '{a: 1 b: 2}\n' > f.yaml
$ succinctly yq -o=json '.' f.yaml
{
  "a": "1 b: 2"
}
$ yq '.' f.yaml
Error: bad file 'f.yaml': yaml: while parsing a flow mapping at <unknown position>: <unknown position>: did not find expected ',' or '}'

$ succinctly yq '.a = 9' f.yaml
{a: 9}          # <-- b: 2 is gone, no error, exit 0
$ yq '.a = 9' f.yaml
Error: bad file 'f.yaml': yaml: while parsing a flow mapping at <unknown position>: <unknown position>: did not find expected ',' or '}'
```

Real yq's flow-mapping grammar treats `: ` (colon followed by whitespace) as a
plain-scalar terminator even inside a flow-context value -- once it sees `1 b:`,
it stops the scalar at `1`, then expects a `,`/`}` and errors on finding `b`
instead. Succinctly's own flow-scalar scanner keeps consuming through the
embedded `: `, swallowing `b: 2` whole into the value of `a`.

## Why this is worse than the project's usual "reduced validation" tradeoff

`ARCHITECTURE.md`/`CLAUDE.md` document semi-indexing's reduced validation vs.
full parsers as an accepted, deliberate tradeoff -- succinctly reading more
leniently than yq is not new or surprising on its own. But this isn't merely
lenient *reading*: the write path (`.a = 9`) **permanently discards** the `b: 2`
field from the file on disk with no warning, which is the "discard a write"
category ADR-0018 explicitly excludes from the divergence-is-permitted list
(alongside "corrupt data"). A user who runs `.a = 9` here gets a file silently
missing a field they never asked to remove.

[...]

Issue #1803, opened 2026-08-29
jq: the same object/array walk-and-validate traversal is hand-copied across five call sites


newhoggy:
## Summary

`src/jq/eval_generic.rs`'s `to_owned_at_depth` (~line 165), `to_owned_cursor_at_depth` (~line 263), `to_owned_with_comments_at_depth` (~line 680), `src/jq/lazy.rs`'s `cursor_to_owned_at_depth` (~line 725), and now `src/jq/eval_generic.rs`'s `push_generic_truthiness_cursor_error` (#1645, PR #1798) all independently implement the same object/array traversal shape: unconses fields/elements, guards against a #1642 colliding-decode-failure-key via `resolve_display_key`/`DisplayKeyGuard`, checks `ends_unpaired()`, and applies the same decode-failure (`string_decode_error()`)/structural-error (`is_error()`) checks once a scalar is reached — differing only in what happens at a leaf (build an `OwnedValue`, build one with comment metadata, or just validate and return `Option<Control>`).

`resolve_display_key`'s own doc comment (`src/jq/document.rs`) already tracks this as a real, recurring cost — it exists specifically so the *key-collision* part of this logic has one definition instead of a hand-copied guard-and-raise sequence at every call site. The rest of the traversal (unconsing, `ends_unpaired`, the decode-failure/structural-error checks) has no equivalent shared core.

## Why this matters

`push_generic_truthiness_cursor_error` needed three iterations within a single branch (#1645/PR #1798) to reach parity with its siblings:
1. First attempt shipped with no error-checking at all (silently treated decode failures/#1194 structural errors as truthy).
2. Review caught that; second attempt added a shallow, non-recursive check (only the condition's own top-level cursor).
3. Second review caught that too; third attempt is the current recursive walk.

Each of those three gaps was a case where hand-transcribing already-correct, already-tested logic from `to_owned_cursor_at_depth` went wrong the first two times. A shared "walk once, then either materialize or just validate" core — parameterized over what happens at a leaf/container, monomorphized so there's no dynamic-dispatch cost (this file already uses this style elsewhere, e.g. `push_generic_owned_values`/`GenericItem`) — would have been correct on the first attempt by construction, and any future fix to the #1642/#1194/decode-failure rules would only need to land in one place instead of five.

## Suggested approach

[...]

PR #1962, opened 2026-08-30
test(jq): cross-site consistency test for #1803's 5 duplicated traversal functions

newhoggy:
Partial progress on #1803 — does not close it (the full generic-unification refactor remains open, see below).

## Summary

#1803 asked to unify 5 duplicated object/array traversal functions (`to_owned_at_depth`, `to_owned_cursor_at_depth`, `to_owned_with_comments_at_depth`, `cursor_to_owned_at_depth` in `lazy.rs`, `push_generic_truthiness_cursor_error`) behind one generic core, and separately suggested a cross-site consistency test as a concrete, safe deliverable regardless of the refactor's shape.

**Investigated all 5 sites in depth before deciding scope.** Two findings changed the plan from a full unification to a test-only PR:

1. `cursor_to_owned_at_depth` (`lazy.rs`) is **JSON-only** (`JsonCursor<'_, W>`, not generic over `DocumentValue`/`DocumentCursor` like the other 4) — reached via `-e`/`--exit-status`. It's a structurally different function, not a 5th instance of the same generic shape, so folding it into one core with the other 4 is a bigger design question than the issue's framing suggested.
2. The `#1677`-style delimiter checks that differ between `to_owned_with_comments_at_depth` and its siblings are actually **no-ops for YAML** — confirmed via the trait's own doc comment ("every format but JSON validates delimiters while parsing") and the missing override in `yaml/light.rs`. While probing for a real cross-site divergence to justify the unification's value, I found a genuine, live, silent-data-loss bug — but its root cause turned out to be an unrelated YAML flow-scalar *parser* bug, not a traversal-function disagreement. Filed separately as #1960.

Given the full unification spans genuinely different generic bounds (plus the JSON-only outlier) and the issue's own "not prescribing the exact shape — worth a design pass" language, this PR scopes to the one deliverable that's unambiguously safe and valuable regardless of how that design pass concludes: the consistency test.

## Changes

- Extended `test_select_and_materialize_agree_on_corruption_1645` (`tests/jq_cli_tests.rs`) from 2-way (`select` vs `-Sc`) to 3-way, adding `-e` (reaches `cursor_to_owned_at_depth`).
- Added `test_select_and_write_agree_on_corruption_1803` (`tests/yq_cli_tests.rs`), comparing `select` against a write op (which exercises `to_owned_with_comments_at_depth`/`to_owned_at_depth`/`to_owned_cursor_at_depth` together) for the two malformation classes meaningful in both JSON and YAML (decode failure, #1642 colliding keys — JSON's "structural error" bareword concept has no YAML analog).

All cases agree today — this is a regression net for the eventual unification, not a behavior change. Confirmed which function each CLI incantation reaches via temporary tracing (removed before commit) rather than assuming from static reading, given `-S`/`-Sc` turned out **not** to route through the cursor-aware sibling despite the name resemblance.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — 1309 lib tests + all integration suites pass
- [x] `cargo test` (default features) — passes
- [x] `cargo build --no-default-features` — builds

### DIFF
--- a/tests/jq_cli_tests.rs
+++ b/tests/jq_cli_tests.rs
@@ -1877,6 +1877,36 @@ fn test_select_raises_on_structural_error_nested_in_container_1645() -> Result<(
     Ok(())
 }
 
+/// Runs `succinctly jq <extra_args> <filter>` over `doc` and asserts it
+/// raises with `code` and a stderr containing `expect_stderr` -- shared by
+/// [`test_select_and_materialize_agree_on_corruption_1645`]'s three
+/// comparison arms so each one asserts on the *message*, not just the exit
+/// code (#1803 code review: an exit-code-only check can't tell "the expected
+/// corruption was caught" from "an unrelated bug in this arm's own code path
+/// happened to raise the same code").
+fn assert_jq_raises(
+    label: &str,
+    op_label: &str,
+    extra_args: &[&str],
+    filter: &str,
+    doc: &str,
+    code: i32,
+    expect_stderr: &str,
+) {
+    let mut args: Vec<&str> = extra_args.to_vec();
+    args.push(filter);
+    let (out, err, actual_code) = run_jq_full(&args, Some(doc))
+        .unwrap_or_else(|e| panic!("[{label}] {op_label} run failed: {e}"));
+    assert_eq!(
+        actual_code, code,
+        "[{label}] {op_label} must raise\nstdout: {out:?}\nstderr: {err:?}"
+    );
+    assert!(
+        err.contains(expect_stderr),
+        "[{label}] {op_label} stderr must contain {expect_stderr:?}\nstderr: {err:?}"
+    );
+}
+
 /// #1645 code review: `push_generic_truthiness_cursor_error` is a second,
 /// hand-copied implementation of the same "walk and raise on corruption"
 /// predicate `to_owned_at_depth`/`to_owned_cursor_at_depth`/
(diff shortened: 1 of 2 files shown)


newhoggy:
## /code-review results (5 independent finder passes)

Genuinely useful, convergent findings — addressed in a follow-up commit:

1. **Coarse assertions** (Angle C): both tests checked only exit code, not stderr content — an unrelated bug in an arm's own code path could raise the same code and the test would pass for the wrong reason. Fixed: both tests now assert the stderr substring for each malformation class, via new small `assert_jq_raises`/`assert_yq_raises` helpers (also resolves the reuse pass's "duplicated run-and-assert block" finding).
2. **YAML coverage gap** (Angle B): the YAML test only had 2 cases (no nesting-depth variants), unlike JSON's 7. Fixed: extended to the same 3 nesting depths (top/array/object) for the decode-failure class, and cross-referenced the pre-existing `test_select_raises_on_yaml_decode_failure_nested_in_container_1645` in the docstring so a reader can see the full coverage picture across both tests.
3. **`-o json` write path never exercised** (Angle C): the YAML write-path snapshot only runs for YAML-output writes. Added a third comparison arm (`.keep = 9 -o=json`) to every case.
4. **Fixed case list vs. corpus/fuzz approach** (altitude, backed by the testing skill's own guidance and the hand-picked-oracle-matrix-is-biased precedent): valid, but building a proper differential fuzzer is its own substantial piece of infrastructure — filed separately as #1965 rather than expanding this PR's scope. Both doc comments now explicitly name this as a known scope limit instead of implying exhaustive coverage.
5. **3 functions bundled in the YAML write check, can't localize which one diverges** (altitude): a real limitation of a CLI-level test for this specific check. Documented explicitly in the test's own doc comment rather than solved — isolating each function individually would need internal instrumentation beyond a CLI test, and is better addressed if/when #1803's actual unification work happens.
6. **Raw-string literals** (reuse pass): switched the YAML fixtures from escaped-quote strings to `r#"..."#` blocks, matching the file's existing convention.

Re-verified after these changes: `cargo test --features cli,simd,regex,serde` (1309 lib tests + all integration suites), `cargo test` (default features), both clippy legs, `cargo fmt --check` — all clean.

PR #1968, opened 2026-08-30
fix(jq): Expr::Builtin/generic fallback arms preserve multi-output through path-context routing

### LINKED ISSUE #1964: jq: eval_pipe_with_path_context_internal's Expr::Builtin catch-all collapses a multi-output builtin into one array
**Severity: Low** (silent wrong output, no crash; requires mixing a multi-output builtin like `range`/`splits`/`recurse` with a path-context builtin like `key`/`parent`/`file_index` inside the same comma/array; no oracle exists since `key`/`parent`/`file_index` are succinctly-only extensions)

## Summary

Found via code review on PR #1961 (#1765 item 2, `Expr::Limit` path-context fix). `eval_pipe_with_path_context_internal`'s generic `Expr::Builtin(_)` catch-all arm (`src/jq/eval.rs`, ~line 28655) evaluates the builtin via `eval_owned_expr_opt`, which only ever returns a single value — collapsing a genuinely multi-output builtin like `range(0;5)` into one `OwnedValue::Array` instead of preserving its outputs individually. This is pre-existing and reachable today on `main` without `limit` or this PR's fix at all:

```console
$ echo '{"a":"x"}' | succinctly jq -c '.a | [(range(0;5), key)]'
[[0,1,2,3,4],"a"]
```

There's no oracle to compare against directly (`key` is a succinctly-only extension jq doesn't have — `jq: error: key/0 is not defined`), but substituting a plain literal for `key` in the same shape confirms what the correct output should be:

```console
$ /usr/bin/jq -c '.a | [(range(0;5), "a")]' <<< '{"a":"x"}'
[0,1,2,3,4,"a"]
```

Succinctly's own plain (non-path-context) evaluator already gets this right (`.a | [(range(0;5), "a")]` also gives `[0,1,2,3,4,"a"]`) — the bug is specific to the path-context evaluator's `Expr::Comma` arm routing each branch through the generic `Expr::Builtin(_)` catch-all, which uses the wrong (single-collapsing) conversion.

## Newly reachable via #1961

Before #1961, `Expr::Limit` never routed into `eval_pipe_with_path_context_internal` at all (no `needs_path_context` arm existed for it), so `limit`'s body never triggered this bug. #1961 adds that routing (correctly, for its own narrower `key`/`parent`/`file_index`-in-limit-body repro), which also makes this pre-existing bug reachable through `limit`:

```console
$ echo '{"a":"x"}' | succinctly jq -c '.a | [limit(2; (range(0;5), key))]'
[[0,1,2,3,4],"a"]   # should be [0,1] -- the corruption preempts limit's own truncation
```

This is not a regression #1961 introduces — the same corruption already exists via a plain `[(range(0;5), key)]` on `main` today, without `limit` in the picture at all. #1961 is scoped to its own narrower repro and leaves this general bug open, tracked here instead.

## Root cause

`eval_pipe_with_path_context_internal`'s `Expr::Comma` arm (added for #715) correctly fans out each branch through this same evaluator so `key`/`parent`/`file_index` resolve per-branch — but every branch, including ones that don't themselves need path context (like `range(0;5)`), goes through this evaluator's own `Expr::Builtin(_)` catch-all, which was written assuming a single-output builtin. `Expr::Array`'s own dedicated path-context arm (#1302) avoids this exact trap by using `ManyOwned`-aware collection instead — the `Comma`/`Builtin` combination doesn't have the equivalent fix.

## Suggested fix direction

Not attempted here. The `Expr::Builtin(_)` catch-all needs a `ManyOwned`-aware evaluation path (mirroring how `Expr::Array`'s dedicated arm already collects `eval_pipe_with_path_context_internal`'s own `ManyOwned` output, rather than routing through `eval_owned_expr_opt`), or `Expr::Comma`'s own loop needs to special-case a builtin sub-expr that doesn't itself need path context by routing it through the plain evaluator's own multi-output-preserving path instead.

## Related

- #1765 / PR #1961 — the fix this gap was found alongside.
- #1302 — the `Expr::Array` fix that solved the identical problem for array-literal construction, whose approach is the likely template for fixing this too.

newhoggy:
Fixes #1964.

## Summary

\`eval_pipe_with_path_context_internal\`'s \`Expr::Builtin(_)\` arm and generic \`_\` fallback both routed through \`eval_owned_expr_opt\`, which collapses a genuinely multi-output expression (\`recurse\`, \`range\`, \`splits\`, \`..\`, or any \`Comma\` branch reaching these catch-alls) into a single \`OwnedValue::Array\` — silently corrupting output whenever a sibling \`key\`/\`parent\`/\`file_index\` forces the whole pipe through path-context routing:

\`\`\`console
$ echo '{"a":"x"}' | succinctly jq -c '.a | [(range(0;5), key)]'
[[0,1,2,3,4],"a"]   # was; should be [0,1,2,3,4,"a"]
\`\`\`

This is the exact bug class #1937 investigated and (correctly) declined to fix with a blanket "take the first output instead of collapsing" policy change to the shared \`eval_owned_expr_full\` helper — that approach regresses \`recurse\`/\`..\`/\`paths\` traffic reached through the *same* two arms. This issue's own suggested fix direction (mirror #1302's \`Expr::Array\` approach) is the actually-correct, surgical fix: \`continue_rest_with_context\`/\`accumulate_path_context_step\` already handle \`QueryResult::ManyOwned\` correctly — they just never received one from these two arms, because \`eval_owned_expr_opt\` collapsed to a single value before they got the chance.

## Changes

- \`Expr::Builtin(_)\` arm and the generic \`_\` fallback in \`eval_pipe_with_path_context_internal\` (\`src/jq/eval.rs\`) now route through \`eval_owned_input\` (preserves \`Many\`/\`ManyOwned\`) instead of \`eval_owned_expr_opt\` (collapses to one value). No changes to \`continue_rest_with_context\`, \`accumulate_path_context_step\`, or \`eval_owned_expr_full\`/\`result_to_owned_full\` — this was a "wrong function called" bug local to two arms, not a shared-helper policy question.
- Corrected a small inaccuracy in the issue's own root-cause note: \`range(0;5)\` parses as \`Expr::Range\`, not \`Expr::Builtin\`, so it actually reaches the generic \`_\` fallback, not the \`Expr::Builtin(_)\` arm. Both needed the identical fix.
- Updated the two \`#1937\` regression-guard tests (which pinned the old array-collapse behavior as "the best available option at the time") to assert the new, correct fan-out instead.
- Added the issue's own \`range\`/\`limit\` repros as new tests. Live-verified the \`limit\` sibling repro already produced the correct output even before this fix on current \`main\` — corrected the test's docstring rather than repeating an unverified claim from the issue.
- Rewrote the relevant \`docs/compliance/jq/limitations.md\` entry to record this closing 2 of the 4 originally-affected sites (\`ParentN\`'s own \`n\` argument and the \`Object\`/\`Array\`/\`Literal\` arm remain open — narrower, and the latter not known to be live-reachable with a genuine multi-output result).

## Test plan

- [x] \`cargo test --features cli,simd,regex,serde\` — 1309 lib tests + all integration suites pass
- [x] \`cargo test\` (default features) — passes
- [x] \`cargo build --no-default-features\` — builds
- [x] Live-verified both fixed repros against jq 1.7.1 with a literal substituted for \`key\` (a succinctly-only extension with no oracle) — output sequences match
- [x] Confirmed the two updated regression-guard tests genuinely exercise the fix (failed with the pre-fix code, verified via manual revert)

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -1920,7 +1920,7 @@ protocol for `resolve_node`, the same shape `eval_each_owned` already gives valu
 evaluation) rather than attempted here; the narrow `first`/`limit` fixes already close
 the two shapes named in #1906/#1935's own repros.
 
-### A generator-argument expression fans out normally, then silently narrows to a single array the moment `key`/`parent`/`file_index` shows up anywhere else in the same pipe
+### A generator-argument expression used to fan out normally, then silently narrow to a single array the moment `key`/`parent`/`file_index` showed up anywhere else in the same pipe -- fixed for 2 of 4 sites
 
 [#1277](https://github.com/rust-works/succinctly/issues/1277)'s clusters 1-3
 (closed by [#1522](https://github.com/rust-works/succinctly/issues/1522)/[#1279](https://github.com/rust-works/succinctly/issues/1279))
@@ -1930,10 +1930,8 @@ Four call sites inside `eval_pipe_with_path_context_internal`
 (`src/jq/eval.rs`) -- `ParentN`'s own `n` argument, the `Expr::Builtin(_)`
 arm, the `Expr::Object`/`Array`/`Literal` arm, and the generic `_` fallback
 -- were an explicit non-goal of that fix, since giving them the same real
-fan-out is a materially larger change (`docs/plan/jq-generator-argument-fanout.md`).
-Those 4 sites route through `eval_owned_expr_opt`/`eval_owned_expr_full`,
-which array-collapses a multi-output expression into a single
-`OwnedValue::Array` instead of fanning out.
+fan-out looked like a materially larger change
+(`docs/plan/jq-generator-argument-fanout.md`).
 
 ```console
 $ echo '{"a":"xax"}' | succinctly jq -c '.a | ltrimstr(("x","z"))'
@@ -1950,18 +1948,16 @@ except for the trailing `key`, which forces the whole pipe through
 `eval_pipe_with_path_context_internal` since `key` needs path tracking --
 `key` itself has no jq oracle (succinctly extension), so this specific
 combination can't be demonstrated as a *jq* divergence in isolation, but it
-is a genuine, demonstrated internal inconsistency: the same sub-expression
-fans out correctly or silently collapses into one array-shaped value
-depending on whether an unrelated path-tracking builtin happens to be
+was a genuine, demonstrated internal inconsistency: the same sub-expression
+fanned out correctly or silently collapsed into one array-shaped value
+depending on whether an unrelated path-tracking builtin happened to be
 anywhere else in the pipe.
 
 **[#1937](https://github.com/rust-works/succinctly/issues/1937) fixed one
-narrow, safe piece of this**: a *zero*-output generator (`(empty)`-style)
-now correctly contributes zero outputs to the enclosing computation, rather
-than an `OwnedValue::Array([])` value -- matching `result_to_owned_full`'s
-identical `#1045` rule for the same `Many`/`ManyOwned` shapes, and with no
-downside, since there is nothing to lose by treating "zero outputs" as "zero
-outputs" instead of manufacturing an empty-array value.
+narrow, safe piece first**: a *zero*-output generator (`(empty)`-style) now
+correctly contributes zero outputs to the enclosing computation, rather than
+an `OwnedValue::Array([])` value -- matching `result_to_owned_full`'s
+identical `#1045` rule for the same `Many`/`ManyOwned` shapes.
 
 **A first attempt at #1937 also tried taking the *first* output instead of
 array-collapsing for the 2+-output case (matching `result_to_owned_full`'s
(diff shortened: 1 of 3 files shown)


newhoggy:
## /code-review results (5 independent finder passes)

Multiple genuinely valuable findings — addressed in a follow-up commit:

1. **`Expr::Object`/`Array`/`Literal` arm has the identical bug, and IS live-reachable** (altitude pass, live-verified): my own PR description was wrong to imply this arm wasn't affected. Routing through this whole evaluator is decided at the *enclosing* comma/pipe level, not by whether the object/array construction itself needs path context — so `.a | ({x:(1,2)}, key)` collapsed `(1,2)` into `{"x":[1,2]}` the same way `range`/`recurse` did. Fixed with the identical `eval_owned_input` swap, but this arm can't delegate to `continue_rest_with_context`'s own `ManyOwned` handling directly (that shares one `root` across every element; this arm needs each constructed value to become its *own* root) — added a small per-element loop instead, mirroring `Expr::Comma`'s own accumulation.
2. **A latent `optional`-gating gap in `continue_rest_with_context`/`continue_rest_with_fresh_root`, newly exposed by this fix** (cross-file tracer + removed-behavior auditor, both independently found and confirmed the same root cause): once these functions started receiving a real `Partial` from the fixed arms (instead of `eval_owned_expr_opt` discarding it beforehand), their own `Partial` arms reattached the trailing control *unconditionally*, so `?` correctly kept a generator's prefix output but failed to suppress the trailing error. Fixed by routing through the existing `catch_error_under_optional` helper (already used by `Expr::Iterate`/`Builtin::Map` for this exact policy) in both functions.
3. **Duplicated match block across the two originally-fixed arms** (reuse pass): extracted into a new `eval_and_continue_with_context` helper.
4. **Test naming imprecision** (line-by-line pass): a test described `recurse`'s arm as "the generic fallback," but `recurse` (`Builtin::Recurse`) actually reaches `Expr::Builtin(_)` — `range(0;5)` is what genuinely exercises the fallback (no dedicated AST arm). Renamed and corrected.
5. **Commit subject over the 100-char CI limit** (conventions pass): reworded via a scripted non-interactive rebase (content-verified identical via `git diff` between old and new commit — pure message change).

Added regression tests for both new fixes (Object-arm fan-out, and `?` suppressing a trailing error after a Partial prefix). Rewrote the `docs/compliance/jq/limitations.md` entry to record the Object arm now being fixed (3 of 4 originally-affected sites) plus the `Partial`/`optional` fix.

Full verification rerun after all changes: `cargo test --features cli,simd,regex,serde` (1309+ lib tests, all integration suites), `cargo test` (default features), both clippy legs, `cargo build --no-default-features`, `cargo fmt --check` — all clean.

newhoggy:
CI failed after rebasing onto \`main\` (which had advanced with unrelated merges while this PR was in review) — root cause and fix:

\`main\` had picked up \`c5669a6d8\` (#1961/#1765's \`Expr::Limit\` path-context fix), which added \`test_limit_path_context_inherits_comma_multi_output_collapse_known_gap_1964\` to explicitly document this exact array-collapse bug as reachable through \`limit\` (pinning the old, buggy \`[[0,1,2,3,4],"a"]\` output as a known, accepted gap at the time). This PR's fix closes that gap, so that test needed updating to assert the correct \`[0,1]\` instead — done, and renamed since it's no longer a "known gap."

While fixing it, I also caught and corrected a mistake of my own: my earlier \`test_comma_range_and_key_fan_out_through_limit_1964\` (added before I'd rebased) claimed this exact combination already produced the correct output *before* this PR's fix, based on live-verifying against my branch's base — which at the time predated \`c5669a6d8\` entirely. Without \`limit\`'s path-context routing existing yet, its own fast-path truncation never hit the collapse bug in the first place — a different reason for the same observed output, not evidence the bug didn't exist for this combination once that routing landed. Removed the now-duplicate, stale-premise test in favor of the corrected main-landed one.

Re-verified: full test suite (1310 lib tests + all integration suites), both clippy legs, \`cargo fmt --check\` — all clean.

PR #1978, opened 2026-08-31
fix(jq): reduce/foreach's input and INIT resolve key/parent/file_index

### LINKED ISSUE #1765: jq: path-context builtins (key/parent/file_index) still silently no-op inside AsPattern/Reduce/Foreach/Limit

newhoggy:
Fixes #1765.

## Summary

Item 3 of #1765's own suggested ordering (items 1-2, \`AsPattern\`/\`Limit\`, already resolved by #1786/#1961). \`needs_path_context\` had no arm for \`Expr::Reduce\`/\`Expr::Foreach\`, so \`key\`/\`parent\`/\`file_index\` inside either's \`input\`/\`INIT\` expression silently stubbed to their zero default instead of resolving against the caller's own ambient position:

\`\`\`console
$ echo '{"a":[1,2]}' | succinctly jq -c '.a | reduce .[] as $x (key; .)'
"a"      # was null
$ echo '{"a":[1,2]}' | succinctly jq -c '.a | [foreach .[] as $x (key; .)]'
["a","a"]  # was [null,null]
\`\`\`

\`UPDATE\`/\`EXTRACT\` are deliberately left unchanged, per the issue's own suggested scoping — both evaluate against the fold's own accumulator, a synthetic value with no document position, where \`key\` already (correctly) answers \`null\`.

## Approach

The bulk of \`eval_reduce\`/\`eval_foreach\`'s own logic is already \`OwnedValue\`-domain once \`input_values\`/\`init_values\` are computed via \`eval_single\` — only that initial conversion is cursor-based. Extracted everything after that point into shared \`eval_reduce_with_values\`/\`eval_foreach_with_values\` cores, then added a hybrid-dispatch arm in \`eval_pipe_with_path_context_internal\` — mirroring \`Expr::Limit\`'s own precedent (#1961) — that sources \`input\`/\`INIT\` from the path-context evaluator instead, only when either actually needs it (this arm's own guard), then feeds the same shared core. \`eval_reduce\`/\`eval_foreach\`'s own cursor-sourced entry points are otherwise unchanged and still handle the common (no path-context builtin) case exactly as before.

The new dispatch arms mirror each function's own error/control-flow precedence exactly (an immediate \`input\`-stream error/break/halt returns before \`INIT\` is ever evaluated for \`reduce\`; \`foreach\`'s own \`Partial\` input keeps its prefix and carries the control forward instead, since \`foreach\` streams per-element).

## Changes

- \`src/jq/eval.rs\`: extracted \`eval_reduce_with_values\`/\`eval_foreach_with_values\`; added \`Expr::Reduce\`/\`Expr::Foreach\` arms to \`needs_path_context\` (guarded on \`input\`/\`init\` only, not \`update\`/\`extract\`) and to \`eval_pipe_with_path_context_internal\`.
- \`tests/jq_cli_tests.rs\`: renamed \`test_reduce_foreach_path_context_still_known_gaps_1663\` to \`test_as_pattern_alternatives_path_context_still_a_known_gap_1663\` (keeping only the still-open \`?//\`-chain \`AsPattern\` case — the reduce/foreach cases it used to pin are now fixed, not gaps); added \`test_reduce_foreach_input_init_path_context_resolves_1765\`, including a case confirming \`UPDATE\`'s own \`key\` handling is unchanged.

## Test plan

- [x] \`cargo test --features cli,simd,regex,serde\` — 1310 lib tests + all integration suites pass
- [x] \`cargo test\` (default features) — passes
- [x] \`cargo build --no-default-features\` — builds
- [x] Confirmed the extraction step alone (before adding the new dispatch arms) is behavior-neutral — full suite passed unchanged
- [x] No jq/yq oracle exists for this (\`key\`/\`parent\`/\`file_index\` are succinctly-only extensions; yq's own lexer rejects destructuring \`as\`/\`reduce\`/\`foreach\` outright) — internal consistency only, matching #1663/#1961's own precedent

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -909,6 +909,21 @@ pub(crate) fn needs_path_context(expr: &Expr) -> bool {
         // navigation before `limit` already creates the enclosing `Pipe`
         // this arm needs.
         Expr::Limit { expr, .. } => needs_path_context(expr),
+        // `reduce EXPR as PATTERN (INIT; UPDATE)` / `foreach EXPR as
+        // PATTERN (INIT; UPDATE[; EXTRACT])` (#1765, item 3 of #1663's own
+        // remaining-gaps list): `input`/`INIT` both evaluate against the
+        // caller's own ambient position -- `eval_reduce`/`eval_foreach`
+        // evaluate each via `eval_single(_, value.clone(), optional)`, the
+        // same `value` this whole pipe is already at -- so a `key`/
+        // `parent`/`file_index` inside either one needs the same context a
+        // sibling `key` elsewhere in the pipe would get. `UPDATE`/`EXTRACT`
+        // are deliberately excluded: both evaluate against the
+        // accumulator, a synthetic value with no real document position,
+        // so `key` there already (and correctly) answers `null` without
+        // needing path-context routing -- the issue's own investigation
+        // flagged this as likely already correct, not a gap to close.
+        Expr::Reduce { input, init, .. } => needs_path_context(input) || needs_path_context(init),
+        Expr::Foreach { input, init, .. } => needs_path_context(input) || needs_path_context(init),
         _ => false,
     }
 }
@@ -25371,6 +25386,34 @@ fn eval_reduce<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     // forks), only the trailing error is swallowed.
     let (init_values, init_control) = stream_outputs_checked(init_result.materialize_cursor());
 
+    eval_reduce_with_values::<W, S>(
+        patterns,
+        update,
+        input_values,
+        init_values,
+        init_control,
+        optional,
+    )
+}
+
+/// The OwnedValue-domain heart of `reduce EXPR as PATTERN (INIT; UPDATE)`,
+/// once `input`/`INIT` have already been reduced to `Vec<OwnedValue>` --
+/// shared by [`eval_reduce`] (cursor-sourced `input`/`INIT`, the ordinary
+/// case) and `eval_pipe_with_path_context_internal`'s own `Expr::Reduce` arm
+/// (#1765: `input`/`INIT` sourced from the path-context evaluator instead,
+/// when either needs `key`/`parent`/`file_index` to resolve correctly).
+/// `update` itself is deliberately never routed through path-context
+/// evaluation by either caller -- it runs against the accumulator, a
+/// synthetic value with no real document position, so `key` there already
+/// answers `null` correctly without needing to change (#1765's own scoping).
+fn eval_reduce_with_values<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
+    patterns: &[Pattern],
+    update: &Expr,
+    input_values: Vec<OwnedValue>,
+    init_values: Vec<OwnedValue>,
+    init_control: Option<Control>,
+    optional: bool,
+) -> QueryResult<'a, W> {
     if init_values.is_empty() {
         return finish_fork(Vec::new(), init_control, optional);
     }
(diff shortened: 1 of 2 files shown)


newhoggy:
## /code-review results (6 independent finder passes)

Genuinely valuable findings — addressed in a follow-up commit:

1. **HIGH (efficiency): unconditional dual-routing.** `input`/`INIT` were both routed through the eager path-context evaluator once *either* needed it, forcing the side that doesn't need it to pay per-element `current_path` bookkeeping it never reads — e.g. `reduce .large_array[] as $x (key; ...)` would clone `current_path` once per array element purely because `INIT` needed `key`. Fixed with a new `eval_expr_needing_path_context` helper that routes each sub-expression independently based on its own `needs_path_context` result, falling back to `eval_owned_input` (no path-tracking overhead) for the side that doesn't need it.
2. **Reuse: `Reduce`'s `input_values` unpack duplicated `stream_outputs_checked`**, called two lines below for `init` in the same arm. Simplified to `stream_outputs_checked` + discard-and-return-early — verified safe specifically for `reduce` (its own cursor-sourced code already treats bare and `Partial` input errors identically, both returning before touching `init`). `foreach`'s sibling arm keeps its own hand-rolled match: collapsing bare-vs-`Partial` there would evaluate `INIT` in a case where `eval_foreach`'s own code never does, changing side-effect ordering — checked this carefully before deciding not to apply the same simplification there.
3. **Missing STYLE-0004 citation** on `eval_foreach_with_values`'s `#[allow(clippy::too_many_arguments)]` — added.
4. **Minor doc-parity nit**: the new `needs_path_context` arm didn't call out the same standalone-top-level caveat `Expr::Limit`'s neighboring arm documents — added.
5. **Coverage gaps** (all manually verified as correct behavior by the reviewing finders, not live bugs): added tests for an input-alone (not just INIT-alone) path-context case, the error/`Partial`-mid-stream interaction with the new dispatch arms for both `reduce` and `foreach`, and `foreach`'s `EXTRACT` clause staying unrouted (mirroring `UPDATE`).

Not addressed (architectural observations, correctly flagged as such, not this PR's scope): the `unreachable!()` invariant-check pattern is now hand-copied 8 times across `eval_pipe_with_path_context_internal`'s sibling arms (pre-existing pattern this PR extends by 3, not introduces); the dispatcher's "recurse into path context for a sub-expr" shape has grown by copy-paste across `#1663`/`#1765`/this PR and would benefit from a generalized helper — both are real observations about the broader mechanism, not local to this fix.

Full verification rerun after all changes: `cargo test --features cli,simd,regex,serde` (1310 lib tests + all integration suites), `cargo test` (default features), both clippy legs, `cargo build --no-default-features`, `cargo fmt --check` — all clean. Manually re-verified the split-routing fix and the reduce/foreach error-ordering semantics live against the built binary and jq 1.7.1 (with a literal substituted for `key`, which has no oracle).

Issue #1573, opened 2026-08-23
jq: path() loses a variable bound from a navigated position (`.a as $y`) — TrackedVar carries no bind-time path


newhoggy:
**Severity: Low** (refuse-only — visible exit 5, no document written)

## Summary

`Expr::TrackedVar` (#844) freezes a variable's bound value, but only for bindings
`is_identity_passthrough` accepts — `. as $x`, and chains of it. A variable bound from a
*navigated* position (`.a as $y`) is substituted as a plain value with no marker at all, so
the path resolver cannot tell it from an inline literal. Real jq keeps the document's own
pointer in both cases, so it answers a path where succinctly refuses.

This is divergence 1 in
[`docs/compliance/jq/limitations.md`](https://github.com/rust-works/succinctly/blob/main/docs/compliance/jq/limitations.md),
now also visible inside `reduce`/`foreach` since #1466 replaced `FoldRegister`'s value
comparison with a provenance one. Before #1466 the value comparison accepted these by
accident — along with every reconstruction that happened to land on the same value, which is
the bug #1466 closed.

## Repro (verified against pinned jq 1.7.1)

```
$ echo '{"a":{"b":1}}' | jq -c 'path(.a as $y | reduce (1) as $i (.a; $y))'
["a"]
$ echo '{"a":{"b":1}}' | succinctly jq -c 'path(.a as $y | reduce (1) as $i (.a; $y))'
jq: error (at <stdin>:0): Invalid path expression with result {"b":1}   # exit 5
```

Same outside any fold, which is the pre-existing half:

```
$ echo '{"a":1}' | jq -c 'path(. as $x | 5 | $x.a)'
["a"]
$ echo '{"a":1}' | succinctly jq -c 'path(. as $x | 5 | $x.a)'   # refuses
```

Pinned as golden known-failures `fold_register_var_from_navigated_binding` and
`fold_register_snapshot_via_nested_init`.

## Root cause

`resolve_node`'s `Expr::TrackedVar` arm re-derives trackability by comparing the frozen
snapshot against the **ambient input**, so it can only ever witness `path=[]`. That is sound
(`getpath(value, []) == value`) but it means the marker carries no path of its own, and
`substitute_var_tracked` is therefore gated on `is_identity_passthrough` — the only shape
where "the bound value is the ambient input" is statically provable.

Widening the *existing* marker to every `$var` is **not** the fix: the arm would then claim
[...]

Issue #1466, opened 2026-08-22
jq: path() through reduce/foreach uses structural equality where jq uses pointer identity, accepting (and writing) where jq refuses


newhoggy:
**Severity: Medium** (accepts where jq refuses, and it reaches the write path)

## Summary

`FoldRegister` (`src/jq/eval.rs`, added by #1440) models jq's own per-emission
`jv_identical(current_input, value_at_path)` register check with a **structural**
equality comparison on `OwnedValue`. jq's check is pointer identity, so a value that
was *reconstructed* rather than navigated to fails it there and passes here —
succinctly answers a path where jq raises, and `=`/`|=`/`del()` through that path
mutate a document jq leaves untouched.

## Repro (verified against pinned jq 1.7.1)

```
$ echo '{"a":1}' | jq -c 'path(reduce (1) as $i (.; {a:.a}))'
jq: error (at <stdin>:1): Invalid path expression with result {"a":1}   # exit 5

$ echo '{"a":1}' | succinctly jq -c 'path(reduce (1) as $i (.; {a:.a}))'
[]                                                                      # exit 0
```

Write side (the part that matters):

```
$ echo '{"a":1}'       | jq -c '(reduce (1) as $i (.; {a:.a})) = 9'      # error, exit 5, untouched
$ echo '{"a":1}'       | succinctly jq -c '(reduce (1) as $i (.; {a:.a})) = 9'      # 9
$ echo '{"a":1}'       | succinctly jq -c 'del(reduce (1) as $i (.; {a:.a}))'       # null
$ echo '{"a":{"b":1}}' | succinctly jq -c '(reduce (1) as $i (.a; {b:.b})) = 9'     # {"a":9}
```

All four raise in jq. All four were also refusals on `main` before #1440 — this is a
new accept, not a pre-existing one.

## Root cause

Two call sites gate purely on value equality, with nothing else required of the branch:

- `FoldRegister::relocate`'s fallback —
  `} else if self.trackable && *b.value == self.value {` — promotes an untracked branch
  to trackable at the register's own path.
- `resolve_reduce`'s final emission — `reg.relocate(vec![PathBranch::untracked(acc)])` —
  re-checks the finished accumulator against the same register.

`Expr::TrackedVar`'s own arm softens `jv_identical` the same way, and the #1440 doc cites
that as precedent for calling this "consistent, not a new gap". The precedent is narrower
than the use: `TrackedVar`'s softening is additionally gated on the node *being* a tracked
[...]

PR #2041, opened 2026-08-31
fix(jq): thread path()'s register through resolve_seq (#1573)

newhoggy:
## Summary

Real `path()` carries a **register** — `jq_state`'s own `(path, value_at_path)` pair —
that only *navigation* advances. `jq_next`'s `INDEX` and `PATH_END` both compare the
current value against it with `jv_identical`; a literal, an arithmetic result, or an
arbitrary builtin call leave it exactly where it was. So a `$var` frozen from where the
register still points steps straight back onto it:

```console
$ echo '{"a":{"b":1}}' | jq -c 'path(. as $x | 5 | $x)'
[]
$ echo '{"a":{"b":1}}' | succinctly jq -c 'path(. as $x | 5 | $x)'   # before: refuses, exit 5
```

`resolve_node`'s `Expr::TrackedVar` arm compared its frozen snapshot against the
**ambient input**, which is the register only while `trackable` holds. The moment any
stage computed a value, the register was lost and every such shape refused.

`PathBranch` now carries the register's value across the stages that don't navigate.
`resolve_leaf`'s computed-value leaf records it at the one transition where it is still
in hand — the step *off* the register — and `resolve_seq` carries it forward while no
stage navigates, re-establishing trackability at the register's own path when a snapshot
matches it.

The "is this branch still on the register?" rule is `FoldRegister`'s existing modelled
`jv_identical`, extracted to `register_identical` so the fold register and the new pipe
register share **one** definition instead of drifting apart.

### Scope: this is the register half of #1573, not the gate half

The issue also asks for `.a as $y`-style *navigated* bindings to track. That needs
`substitute_bound_var`'s `is_identity_passthrough` gate widened, and I found that
widening it alone is **unsound** — it regresses into exactly the accept-where-jq-refuses
class #1466 closed:

```console
$ echo '{"a":{"b":1},"c":{"b":1}}' | jq -c 'path(.a as $y | .c | $y)'
jq: error ...                       # equal value, DIFFERENT node
                                    # with the gate widened, succinctly answered ["c"]
```

`register_identical` compares values plus a "never rebuilt" provenance bit; that
distinguishes a snapshot from a reconstruction, but not *which node* the snapshot came
from. `OwnedValue` has no identity, so the marker needs a **bind-time path** before the
gate can be widened — which is the issue's original `(path, value)` proposal, and which
the spike's later comments concluded was unnecessary. It is necessary; it just isn't the
acceptance *criterion*. Filed as a follow-up with the repro rather than shipped
half-verified. `test_path_navigated_binding_still_refused_pending_gate_1573` pins the
current refusal so that widening has to update it deliberately.

## Verification

Model derived and every expectation confirmed live against pinned **jq 1.7.1**
(`/usr/bin/jq`), including the probes that refute the two earlier hypotheses on the
issue: the `as`-binding source does *not* move the register
(`path(.a as $y | .b)` = `["b"]`), and an intervening literal does not disqualify a
variable while an intervening navigation does.

**Randomised differential fuzz vs jq 1.7.1** (`path`/`del`/`=`/`|=` wrappers over
generated bind/navigate/compute pipelines), classified by *direction*:

> **Superseded by the review follow-up at the bottom of this description.** This run's
> generator could not emit a stage that navigates *inside jq's own bytecode* (a `def`
> body, `first`, `add`), which is exactly where a fabrication was hiding; the identity-
> passthrough row's single REFUSE also undercounts. Re-run numbers are below.

| run | shapes | agree | FABRICATE (succ answers, jq refuses) | MISMATCH | REFUSE (safe) |
|---|---|---|---|---|---|
| all binding shapes | 2500 | 2423 | **0** | **0** | 77 |
| identity-passthrough bindings only | 2500 | 2499 | **0** | **0** | 1 |

Zero fabrications and zero mismatches — nothing became acceptable that jq refuses. All 77
REFUSEs in the first run are the navigated-binding gate above. The single REFUSE in the
second run is a **pre-existing, unrelated** divergence the fuzz surfaced
(`path(null | .a)` on `null` — a computed `null` that is `jv_identical` to a `null`
register keeps navigation alive in jq); confirmed present on a build without this change,
and filed separately.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde --no-fail-fast` — 56 suites, all green
- [x] `cargo test --no-fail-fast` (default features, CI's own `Test` job)
- [x] `cargo fmt --check`
- [x] `cargo build --no-default-features` (`no_std`)
- [x] `RUSTDOCFLAGS=-D warnings cargo doc --all-features --no-deps`
- [x] 6 new regression tests, including three negative controls (a navigating stage moves
      the register; a computed value that isn't the register stays untracked; the
      navigated-binding gate still refuses)
- [x] Differential fuzz vs pinned jq 1.7.1, 5000 shapes total (table above)

Refs #1573

---

## Review follow-up: the register must not survive a stage jq itself navigates through

The first cut of this PR recorded the ambient value as the register for **every** stage
`resolve_leaf` could not decompose. That is not the same as "the stage did not navigate":
jq's `value_at_path` moves on any `INDEX` its bytecode executes, including the ones inside
a jq-*defined* builtin (`first` is `.[0]`, `add` is `reduce .[] as $x ...`) or a user
function body — stages that arrive here as one opaque computed value. A later `$var` then
re-established at a position jq had already left:

```console
$ echo '{"a":{"b":1},"c":{"b":1}}' | jq -c '(. as $x | ([.a]|first) | $x.c) = 9'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 0 of [{"b":1}]
$ echo '{"a":{"b":1},"c":{"b":1}}' | succinctly jq -c '(. as $x | ([.a]|first) | $x.c) = 9'
{"a":{"b":1},"c":9}                     # before this fix — a write jq refuses, exit 0
```

Refuse-only it was not: `=`, `|=` and `del()` all wrote through the fabricated path, which
is the accept-where-jq-refuses class #1466 closed. `main` refuses all of these.

### What changed

- **`cannot_move_register(expr)`** — a syntactic allowlist of stage shapes proven (live,
  jq 1.7.1) to leave the register in place: literals, `$var`, `@fmt`, `$ENV`/`$__loc__`,
  arithmetic/comparison/boolean/alternative/`if`/`try`/`error` over safe operands, a
  construction or interpolation that navigates nothing, and eight C-implemented builtins
  (`length`, `type`, `keys`, `keys_unsorted`, `tostring`, `tonumber`, `tojson`,
  `utf8bytelength`). Everything else — `first`/`add`/`any`, any `def` or call, `reduce`,
  `label`, `..` — drops the register. Getting it wrong in this direction costs a refusal;
  the other direction fabricates.
- **`resolve_seq` consults it once per stage** (`stage_preserves_register`), gating both
  `reestablishes_register` and `carry_register`. A construction that itself navigates is
  excluded because jq *also* raises for a `.a` applied to a computed value —
  `path(. as $x | {k:.a} | [.a] | $x)` raises in jq on the second `.a`, which this resolver
  never sees inside a construction.
- **One placement rule** for the `Ok` fan-out and the escaping partial prefix, instead of
  the two copies that had already caused one omission in this PR's own history.
- `StepRegisterFacts` bundles the four per-step bools (same reasoning as `SliceEditFlags`).

### Cost, all refuse-only and pinned by tests

`path(. as $x | [.a] | $x)`, `{k:.a}`, `"\(.a)"`, `(.a as $q | 1)` and
`(def f: 5; f)` are `[]` in jq and now refuse here. Recorded in
`docs/compliance/jq/limitations.md` alongside a fourth, pre-existing one the review found:
`path(.a | null)` on `{"a":null}` is `["a"]` in jq (the `null`/`bool` arm of
`register_identical` applied one stage earlier than `reestablishes_register` reaches).
The `register_identical` doc comment previously cited that shape as if succinctly matched
it; corrected.

### Re-verified

Differential fuzz vs pinned jq 1.7.1 over `path`/`del`/`=`/`|=`/`getpath(path(...))`
wrappers, with the generator's computed-stage pool extended to include the opaque
navigating stages the original run could not emit (`def f: .a; f`, `[.a]|first`,
`[1]|first`, `[.a]|add`, `map(1)|first`, `[1,2]|sort|first`, `first(.a)`, `.. | 5`):

| run | shapes | agree | FABRICATE | MISMATCH | REFUSE (safe) |
|---|---|---|---|---|---|
| all binding shapes, 3 seeds | 3590 | 3298 | **0** | **0** | 292 |
| identity-passthrough only, 2 seeds | 2379 | 2322 | **0** | **0** | 56 |

The remaining REFUSEs are the #2042 gate gap, the allowlist costs above, and #2044.
One row in the identity-passthrough run classifies as a fabrication and is not one:
`del(. as $x | {} | ([1]|first) | empty|1)` never references `$x`, and the identical
variable-free `del({} | ([1]|first) | empty|1)` diverges the same way on `main` — it is
`empty`'s lazy pruning racing jq's eager path check (#986/#987 territory), unrelated to
the register.

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -519,16 +519,34 @@ is the revert that established what the other one costs.
 
 1. **Variable-rooted navigation off a mismatched accumulator** —
    `path(. as $x \| foreach (1,2) as $i (0; $x.a; .b))` on `{"a":{"b":1},"c":2}` is
-   `["a","b"]` ×2 in jq; succinctly refuses. Same root cause as the pre-existing, unrelated
-   `path(. as $x \| 5 \| $x.a)` on `{"a":1}` (jq `["a"]`, succinctly already refuses today) —
-   `resolve_node` conflates "current input" with a variable's own bound position outside any
-   fold too, so this isn't specific to #1440's own fix. `Expr::TrackedVar` witnesses only
-   `path=[]`, which is why `substitute_var_tracked` is gated on `is_identity_passthrough`:
-   a variable bound from a *navigated* position (`.a as $y`) carries no marker at all, so
-   `path(.a as $y \| reduce (1) as $i (.a; $y))` — jq `["a"]` — refuses too. Tracked as
-   [#1573](https://github.com/rust-works/succinctly/issues/1573), which proposes carrying the
-   binding's own path on the marker; goldens `fold_register_var_from_navigated_binding` and
-   `fold_register_snapshot_via_nested_init`.
+   `["a","b"]` ×2 in jq; succinctly refuses.
+
+   The *plain-pipe* half of this shape is closed.
+   [#1573](https://github.com/rust-works/succinctly/issues/1573) established that jq carries
+   a `(path, value_at_path)` **register** which only navigation advances — a literal never
+   moves it, so a `$var` frozen from where it still points steps back onto it — and
+   `resolve_seq` now threads that register (`PathBranch::register`,
+   `reestablishes_register`, `src/jq/eval.rs`). `path(. as $x \| 5 \| $x.a)` on `{"a":1}`,
+   which this list previously recorded as refusing, answers `["a"]` like jq.
+
+   What remains here is that [`FoldRegister`](../../../src/jq/eval.rs) does not seed *its*
+   register into the resolution of its own `UPDATE`/`EXTRACT`: the accumulator (`0`) is
+   resolved with tracking already off and no register to compare against, so `$x.a` cannot
+   re-establish the way it does in a plain pipe. Closing it means threading the register
+   into `resolve_node`/`resolve_leaf` as a parameter rather than carrying it on the branch,
+   which is a materially larger change across that function's ~20-call-site dispatch graph.
+
+   Separately, a variable bound from a *navigated* position (`.a as $y`) still carries no
+   marker at all, because `substitute_var_tracked` remains gated on
+   `is_identity_passthrough` — so `path(.a as $y \| reduce (1) as $i (.a; $y))`, jq `["a"]`,
+   refuses too. That gate **cannot simply be widened**: measured against jq 1.7.1, doing so
+   makes `path(.a as $y \| .c \| $y)` on `{"a":{"b":1},"c":{"b":1}}` answer `["c"]` where jq
+   refuses — reopening the accept-where-jq-refuses class #1466 closed, since
+   `register_identical`'s provenance bit records "never rebuilt", not "from *this*
+   position", and `OwnedValue` has no node identity. The marker needs a bind-time path
+   first; tracked as [#2042](https://github.com/rust-works/succinctly/issues/2042). Golden
+   `fold_register_var_from_navigated_binding`.
+
 2. **`?//`-alternatives folds aren't path-tracked at all** (refuse-only) —
    `path(. as $x \| reduce (1) as $y ?// $z (0; $x))` on `{"a":1}` is `[]` in jq; succinctly
    refuses. [#1365](https://github.com/rust-works/succinctly/issues/1365) (`?//`-alternatives
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -536,6 +536,31 @@ is the revert that established what the other one costs.
    into `resolve_node`/`resolve_leaf` as a parameter rather than carrying it on the branch,
    which is a materially larger change across that function's ~20-call-site dispatch graph.
 
+   The register is also carried **only across stages this resolver can prove did not move
+   it** (`cannot_move_register`, `src/jq/eval.rs`). jq's register advances on any `INDEX`
+   its own bytecode executes, which includes the ones hidden inside a jq-*defined* builtin
+   (`first` is `.[0]`, `add` is `reduce .[] as $x ...`) or a user function body — stages
+   that reach the resolver as one opaque computed value. Assuming those left the register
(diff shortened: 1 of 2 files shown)


newhoggy:
Re-verified after the branch was rebased onto `cc6dcb9f6` (the background fleet did the rebase; all five commits intact, content diff against the new base is exactly `src/jq/eval.rs` + `docs/compliance/jq/limitations.md`).

New `main` had itself touched `src/jq/eval.rs`, so this was a full re-run on a clean build (`cargo clean -p succinctly` first), not an incremental one:

| gate | result |
|---|---|
| `cargo test --features cli,simd,regex,serde --no-fail-fast` | 56 suites green |
| `cargo test --no-fail-fast` (default features) | green |
| `cargo clippy --all-targets --all-features -- -D warnings` | green |
| `cargo clippy --all-targets --features std,simd,serde,cli,regex,bench-runner,large-tests,mmap-tests -- -D warnings` | green |
| `cargo fmt --check` | green |
| `cargo build --no-default-features` (`no_std`) | green |
| `RUSTDOCFLAGS=-D warnings cargo doc --no-deps --all-features` | green |
| oracle matrix vs jq 1.7.1 | 14/14 agree |
| differential fuzz (both runs, 4000 shapes) | **0** fabricate, **0** mismatch |
| `omni-dev coverage diff` | **100%** patch (138/138) |

Note for anyone reading the check history: the `Perf Regression Guard (ARM64-Linux)` failure visible around this time is on **`main`'s own head** (`cc6dcb9f6`, check-run `99464034821`), not on this branch — it surfaced here only because the rebase briefly moved local `HEAD` onto that commit.

Issue #2012, opened 2026-08-31
jq: --argjson/--jsonargs rejects a lone low surrogate that real jq accepts (#2008 follow-up)


newhoggy:
## Summary

`--argjson`/`--jsonargs` reject a JSON string literal containing a lone *low*
surrogate escape (`\uDC00`-`\uDFFF`), where real jq 1.7.1 accepts it and
substitutes U+FFFD -- confirmed live against the pinned oracle.

```
$ succinctly jq -n --argjson x '"\udc00"' '$x'
Error: Invalid JSON for --argjson x

Caused by:
    0: Invalid JSON: "\udc00"
    1: lone leading surrogate in hex escape at line 1 column 7

$ /usr/bin/jq -n --argjson x '"\udc00"' '$x'
"�"
```

## Why this is a separate issue from #2008

#2008 fixed the same class of bug in two hand-rolled decoders that both
special-cased the high-surrogate arm but forgot the low-surrogate one:
`json::light::decode_escapes` (the primary document parser) and
`eval.rs::parse_json_string_value` (`fromjson`/`tonumber`).

`--argjson`'s value parser is architecturally different: it validates via
`serde_json::from_str::<serde_json::Value>` first
(`validate_and_materialize_json` in `src/bin/succinctly/jq_runner.rs`,
documented there as deliberately using `serde_json::Value` rather than the
cheaper `IgnoredAny` for magnitude-overflow checking), and only reparses via
this crate's own semi-indexer *after* that validation succeeds. `serde_json`
itself rejects the lone low surrogate outright, so the fix can't be "add a
missing match arm" the way #2008's two call sites were -- it needs either a
pre-pass that substitutes U+FFFD before handing the text to `serde_json`, or
swapping the validation strategy for this one input shape. That's a
different enough fix shape that it didn't belong in #2008's scope (found
during that PR's code review, filed separately per the project's own
"different bug shape -> separate issue" triage convention).

## Reproduction

Pinned oracle: `/usr/bin/jq` 1.7.1.

```
$ succinctly jq -n --argjson x '"\udc00"' '$x'; echo $?
1
$ /usr/bin/jq -n --argjson x '"\udc00"' '$x'; echo $?
0
```

newhoggy:
🔧 Claimed by agent `Elmo` — branch `issue-2012-argjson-low-surrogate`.

Issue #1909, opened 2026-08-30
perf(jq): any path-context query pays an unconditional O(document size) to_owned_checked conversion, dwarfing #1850's own fix


newhoggy:
**Severity: Low-Medium** (perf only; no incorrect output). Found while implementing/measuring #1850's fix.

## Summary

`eval_pipe` (`src/jq/eval.rs`) converts the **entire** input document from its
lazy cursor representation into a fully-materialized `OwnedValue` tree
(`to_owned_checked`) the moment *any* expression in the pipe needs path
context (`path(...)`, `key`, `parent`, `file_index`, ...) — unconditionally,
regardless of what the path expression itself actually touches:

```rust
// eval_pipe, src/jq/eval.rs
if exprs.iter().any(needs_path_context) {
    let owned = match to_owned_checked(&value) {
        Ok(v) => v,
        Err(e) => return QueryResult::Error(e),
    };
    return eval_pipe_with_path_context::<W, S>(exprs, &owned, &[], optional);
}
```

This cost is O(document size) and is paid **once per query**, before any
path-resolution logic (bounded or not) ever runs.

## Measured impact (debug build, 300,000-element flat array)

```
path(.)                    -- no limit, no iterate at all: ~0.47-0.56s user
length                     -- pure parse, no path context: ~0.12s user
[limit(3; .[])]            -- value-context limit, no path context: ~0.13s user
path(limit(3; .[]))        -- path-context limit:            ~0.58-0.66s user
```

`path(.)` alone — an expression that does *zero* navigation once it has an
`OwnedValue` — already costs 4x the value-context/parse-only baseline. That
gap is `to_owned_checked`'s own cost, not any path-resolution logic.

## Relation to #1850

#1850 measured `path(limit(3; .[]))` at ~0.55s and attributed it to
`resolve_limit_one_n`'s eager materialize-then-truncate pattern (fixed for
the bare `.[]` case in the PR closing that issue). That fix is real and
verified — it removes `resolve_limit_one_n`'s own O(document size)
contribution — but isolating it (`time(path(limit(3;.[]))) -
time(path(.))`, same binary, same input) shows its own marginal cost was
only ~0.02-0.1s out of the ~0.55-0.66s total. The other ~0.5s is
`to_owned_checked`, present identically in `path(.)` with nothing else to
attribute it to.

So #1850's fix is correct and worth keeping, but does not explain (and
cannot, by itself, meaningfully improve) the wall-clock number that
[...]

Issue #1891, opened 2026-08-30
test: spawn_with_signal_retry can leak a zombie child on a failed stdin write


newhoggy:
**Severity: Low** (bounded by the test binary's own short lifetime; requires a child that exits/closes stdin before the parent finishes writing)

## Summary

`tests/common/cargo_run_exit.rs`'s `spawn_with_signal_retry` writes
`stdin_input` to the spawned child before calling `wait_with_output()`:

```rust
let mut child = /* spawn, possibly retried */;
if let Some(input) = stdin_input {
    use std::io::Write;
    if let Some(mut sin) = child.stdin.take() {
        sin.write_all(input)?;   // <-- can fail (e.g. BrokenPipe)
    }
}
let output = child.wait_with_output()?;
```

If `write_all` fails (the child exits or closes stdin early — e.g. an
argv-parse error before the binary ever reads stdin), the `?` propagates
immediately and `child` is dropped without ever being waited on.
`std::process::Child`'s `Drop` does not reap the OS process, so the child
is leaked as a zombie for the remainder of the test binary's own run.

Found during #1884's code review (round 3, Angle D) while fixing an
unrelated ENOENT-retry gap in this same function.

## Discovery context

Pre-existing in `spawn_with_signal_retry` since #1847/#1546 — not
introduced by #1884, and not specific to `jq_cli_tests.rs`; the same
helper is shared by `cli_golden_tests.rs`, `cli_characterization_tests.rs`,
`deep_nesting_valid_tests.rs`, `json_validate_tests.rs`, and (as of #1884)
`jq_cli_tests.rs`.

The same review pass also flagged the classic `Command` pipe-deadlock
shape (`write_all` runs to completion before any concurrent
stdout/stderr drain starts) as currently dormant — `src/bin/succinctly`'s
own stdin handling reads to completion before producing output today, so
the OS pipe buffer can't fill mid-write — but undocumented as an
invariant this helper's correctness quietly depends on. Worth fixing
together since both are about the same write-then-wait ordering.

## Suggested approach

On a `write_all` failure, still call `child.wait()` (or spawn a thread to
drain stdout/stderr concurrently, closing the deadlock risk too) before
propagating the write error, so the child is always reaped regardless of
which branch errors. Filing to make this trackable, not blocking.

newhoggy:
🔧 Claimed by agent `Elmo` — branch `issue-1891-zombie-child-signal-retry`.

PR #2045, opened 2026-08-31
perf(jq): stop path-context queries materializing the document twice

newhoggy:
## Summary

Every query that needs path context reached `eval_generic.rs`'s round-trip bridge (`eval_on_owned`): materialize the whole input, re-serialize it to JSON, rebuild a `JsonIndex` over that text, then re-enter `eval.rs` — where `eval_pipe`/`builtin_path` immediately call `to_owned_checked` and materialize **the same document a second time**, because path resolution needs stable `&OwnedValue` references and can't work off a cursor. Two full trees, one JSON string and one semi-index, to answer a query whose output is a bounded path.

The doubling is the part #1909 doesn't name — its "Why this exists" section describes the bridge, but the second tree only shows up once you follow `eval_on_owned` into `eval_pipe`'s own identical `needs_path_context` gate.

This removes the second tree for `path(...)`, `key`, `parent`, `parent(n)`, `file_index`, and any `path` inside a pipe — the "pure waste" set #1909's own comment scopes.

### Measured

Interleaved A/B against **`cc6dcb9f6`** (the `main` tip when measured), medians of 5, three input sizes rather than one ratio (a removed O(n) term should hold its ratio as input grows — it does). Flat arrays of `{"k":i,"v":"x"}`:

| query | n = 30k | n = 300k | n = 1M |
|---|---|---|---|
| `path(.)` — time | 0.53x | 0.51x | 0.44x |
| `path(.)` — peak RSS | 0.69x | 0.63x | 0.63x |
| `path(.[0])` — time | 0.51x | 0.40x | 0.41x |
| `path(.[0])` — peak RSS | 0.58x | 0.48x | 0.49x |
| `.[0] \| parent \| length` — time | 0.63x | 0.56x | 0.58x |
| `.[0] \| parent \| length` — peak RSS | 0.72x | 0.69x | 0.68x |

In absolute terms at n = 1M: `path(.)` 2.59s / 1585 MiB → **1.13s / 1003 MiB**; `path(.[0])` 2.57s / 1057 MiB → **1.06s / 519 MiB**. `.[0]`, `length` and `getpath([0])` are unchanged controls — and they are what makes this particular run trustworthy: the box was *not* idle (another worktree's test run held ~3 cores throughout), yet all three controls still read 0.99–1.01x on time and 1.00–1.04x on RSS at every size, so the interleaving absorbed the contention rather than the contention biasing a half.

> **Base note.** An earlier revision of this table was measured against `a8eee6a3a`, the commit this branch was cut from. That is now four commits behind `main`, and those four move peak RSS on the bridge path by ~+4.6 MiB on their own (`getpath([0])` 75.4 → 80.3 MiB on a 300k flat array; likewise `add`, `ltrimstr`, `tojson`) — enough to be mistaken for a regression introduced here. Re-measured against `main` so the "before" column describes what this actually lands on. The YAML table below is still against `a8eee6a3a`; its ratios are unaffected, since those four commits touch neither the YAML path nor the guard.
>
> The branch has since been rebased twice more, and now sits on `6f0254fcf`. Everything `main` gained in between is either confined to `src/bin/succinctly/jq_runner.rs` + `tests/jq_cli_tests.rs` (#2012's `--argjson` surrogate fix) or entirely under `tests/**` (#1891/#2016's zombie-leak sweep) — nothing on the path-context or guard code path, and nothing in `src/jq/`, so the table above still describes this branch's own effect and was not re-measured. The post-rebase tree was re-tested rather than assumed: that sweep rewrites call sites in `tests/yq_cli_tests.rs`, the same file this branch appends to.

The same, on **YAML** through `succinctly yq` (against `a8eee6a3a`, see the base note above) — 10 MB, 300,000 `{name, count}` records, medians of 5, interleaved:

| query | time | peak RSS |
|---|---|---|
| `path(.)` | 0.690s → **0.360s** (0.52x) | 518.8 MiB → **332.4 MiB** (0.64x) |
| `path(.[0])` | 0.650s → **0.320s** (0.49x) | 360.0 MiB → **192.0 MiB** (0.53x) |
| `.[0] \| parent \| length` | 0.860s → **0.550s** (0.64x) | 542.2 MiB → **376.4 MiB** (0.69x) |
| `.[0].count \| key` | 0.640s → **0.320s** (0.50x) | 360.0 MiB → **192.0 MiB** (0.53x) |
| `length` (control) | 1.00x | 1.00x |

A first, non-interleaved run of that table read `path(.)` as 29% *slower* — the sequential-halves thermal bias `CLAUDE.md`'s A/B section warns about, reproduced exactly. Interleaved, it is 1.9x faster.

### What the guard costs when it refuses

`reindex_bridge_is_identity` walks the materialized tree, so a document it *refuses* pays that walk **and** the bridge. Measured on 300,000 YAML integers with a trailing `.inf` — a bare `Float`, so the guard refuses — `path(.)` is 129.8 ms vs 125.6 ms = **1.03x**, peak RSS unchanged. The predicate short-circuits (`all()`), so only a late or rare bare `Float` pays for a full walk and a float-first document pays essentially nothing. Recorded rather than left to be rediscovered: a precheck is charged to every input including the ones it cannot help, so its cost has to be measured on the workload where the guarded code is already fastest.

### What changed

1. **`eval::builtin_path_on_owned`** — `builtin_path` with its own `to_owned_checked` lifted out, so `Builtin::Path`'s new generic arm can pass the tree it already built.
2. **`eval_pipe_with_path_context` made `pub(crate)`** — so the `Expr::Pipe` `needs_path_context` bridge calls it directly instead of re-entering through `eval_pipe`'s identical gate.
3. **`reindex_bridge_is_identity`** — the guard that makes 1 and 2 sound. The bypass is taken only for a tree the bridge would have left alone.

`optional` is deliberately not threaded into the bypass: `eval_on_owned` reaches `eval.rs` via `full_eval`, which restarts every evaluation at `optional = false` regardless of what its caller passed. Passing `false` mirrors what the bridge actually delivered.

### Why a guard is needed at all

`to_json_for_reindex` is a *formatter* as much as a serializer, so the round trip isn't always the identity. The exceptions are all numeric: a bare `Float` (#953's mode-forked re-spelling — yq keeps a whole number's decimal point at any magnitude), a bare `Int`, a NaN literal, and a `NumberLiteral` past the 256-byte reuse cap (#1211). Ordinary document-sourced numbers are short `NumberLiteral`s, which the formatter echoes verbatim, so the bypass still fires for them.

`Int` is the one node the bypass lets the bridge **normalize** rather than preserve (`Int(n)` → `NumberLiteral(Int(n), "n")`). That is sound where a bare `Float` isn't, and the difference isn't a judgement call: `to_json_for_reindex` writes an `Int` as exactly `format!("{n}")` in both modes — the only spelling an `i64` has — so the literal the bridge bakes in is the same text the bare `Int` already renders as, and #1008's literal preservation has nothing new to echo. It matters because **every YAML integer materializes as a bare `Int`** (YAML's `number_literal()` override preserves only floats), so rejecting it left `succinctly yq` with no bypass at all — caught in the second review round, after one `count: 3` line was measured disabling the optimization for a whole manifest.

That cap is duplicated from a private `const` inside `to_json_for_reindex`'s body and can't be shared, so `test_reindex_bridge_identity_predicate_agrees_1909` asserts the predicate against the **real** round trip rather than against its own reasoning (CLAUDE.md: *"duplicated predicates diverge silently"*). It asserts soundness (predicate ⟹ identity) rather than equivalence — conservatism costs a missed optimization, never a wrong answer — plus a reachability assertion, so a predicate that stayed sound by answering `false` to everything fails there instead of silently un-fixing the issue.

One case that test cannot falsify is `Int` itself: its round-trip helper normalizes `Int` before comparing, so `Int` passes by construction rather than by observation (review). The equality that normalization assumes — `to_json_for_reindex` spells an `Int` as exactly `format!("{n}")`, in **both** modes — is therefore asserted on its own in `test_reindex_bridge_int_normalization_1909`, over `0`, `±1`, `±7`, `1000`, `i64::MIN`, `i64::MAX`, and confirmed falsifiable by ablation (perturbing the expected spelling by one character fails it). Otherwise the guard's most load-bearing claim would have been the one nothing in the tree could contradict.

### Review-driven changes (the second commit)

A first version got two things wrong; both are worth reading before reviewing this one.

- **The float guard was on the output, not the input.** It re-applied `yq_float_fidelity_fixup` to the bypass's *result*. That is not the transformation the bridge performs, and it was wrong in both directions once the pipe computes anything: `.outer.big|parent|.big|tostring` lost the document spelling (`"1e+19"`), while `.outer.big|parent|.big+0` wrongly *gained* it for a value the pipe had just computed. Re-spell-then-evaluate is not evaluate-then-re-spell. Fixed by refusing the bypass on the input instead; pinned by `test_path_context_bypass_keeps_yq_float_fidelity_1909`, which covers all three shapes and fails if the guard is forced on.
- **`getpath` got a cursor-walking fast path, and it has been withdrawn.** It reached full parity with `.[0]` (0.010s / 11.0 MiB vs 0.085s / 75.4 MiB), but review found two structural blockers: `getpath`'s whole-document `to_owned_checked` is a **validity gate**, not waste (`{"a":"\ud800","d":5} | getpath(["d"])` raises on `main` and is rejected by real jq; a cursor walk returned `5`, exit 0), and a probe-then-fall-back structure evaluates `path_expr` twice whenever the shape checks fail (`getpath(("a"|stderr))` printed `aa`, jq prints `a`). Both constraints, and the two independent fix directions, are written up in **#2053** rather than shipped as a weaker version here.

Also fixed while there: **`assert_parity` could not see errors.** It funnels both evaluators through `collect_owned()`, whose `Error(_) => vec![]` arm drops the error, so an entire "shapes that must raise" block was asserting `[] == []`. `assert_error_parity` asserts on the message and panics if either side fails to raise.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde --no-fail-fast` — 56 suites, 0 failures
- [x] `cargo test --verbose --no-fail-fast` (default features — the CI leg `--features cli` never runs)
- [x] `cargo test --verbose --features simd,scalar-yaml --no-fail-fast`
- [x] `cargo test --verbose --features simd,broadword-yaml --no-fail-fast`
- [x] `cargo clippy --all-targets --all-features -- -D warnings` (after `cargo clean -p succinctly`)
- [x] `cargo check --no-default-features` — 8 warnings, identical to the base commit
- [x] **Differential fuzz vs the pre-change binary** (`a8eee6a3a`, built separately), classifying any diff by direction: **4,800 randomised JSON** document/query pairs across jq **and** yq modes, query set widened toward `key`/`parent`/`path`/`select(key==...)` shapes; **3,200 randomised YAML** documents (duplicate keys, anchors/aliases, `!!str`/`!!int` tags, `.inf`/`.nan`, `007`, `0x10`, `i64::MAX`, overflowing floats), query set widened toward the channels an `Int` → `NumberLiteral` normalization would surface in (`tostring`, `tojson`, `@json`, `to_entries`, `+0`), each run through one of four output modes (default YAML, `-o json`, `-P`, `--sort-keys`) — **0 divergences**. An earlier round of the same harness is what caught the YAML-tag defect in the withdrawn `getpath` path.
- [x] **Ablation**, not just presence: forcing `reindex_bridge_is_identity` to `true` makes `test_path_context_bypass_keeps_yq_float_fidelity_1909` fail with `1e+19` — the guard is load-bearing and the test catches its removal.
- [x] New tests: `test_reindex_bridge_identity_predicate_agrees_1909` (soundness + reachability against the real round trip), `test_reindex_bridge_int_normalization_1909` (the `Int` spelling equality the helper above normalizes away, both modes), `test_parity_path_builtin_bypasses_reindex_bridge_1909`, `test_parity_pipe_path_context_bypasses_reindex_bridge_1909` (both including float-carrying documents that take the fallback and float-free ones that take the bypass), `test_path_context_bypass_keeps_yq_float_fidelity_1909`, plus `assert_error_parity`.

Refs #1909. Follow-up filed: #2053.


newhoggy:
Two `/code-review high` rounds ran against this branch (findings and fixes are in the commit messages of `1029c1c34`, `3ea03d6ee` and `8510bce81`). An unattended session can't invoke `/code-review ultra` — it needs a human to trigger and confirm billing — so if you want the deepest pass before merge, `/code-review ultra 2045` is the command.

The two places I'd point it at first, since both are where I already got something wrong once:

1. **`reindex_bridge_is_identity`'s completeness.** It claims the reindex bridge's *only* non-identity transformations are numeric. Strings, object keys and containers are asserted to survive escape-then-reparse symmetrically, and the property test exercises that over a corpus — but a corpus is not a proof, and this predicate is the sole thing standing between the bypass and a silent output change.
2. **`Int` as the one sanctioned normalization.** The argument is that `to_json_for_reindex` writes an `Int` as exactly `format!("{n}")`, so the literal the bridge bakes in is the text the bare value already rendered as. Verified empirically across the `i64` range endpoints and through both `to_json` and `to_json_for_reindex`, but it rests on no code branching on the *variant* (`Int` vs `NumberLiteral(Int, _)`) in a way that outlives the rendering — which is exactly the kind of thing #1008's literal preservation does do for `Float`.

PR #2060, opened 2026-08-31
fix(jq): make repeat(f) demand-driven so limit/first stop it at the source

### LINKED ISSUE #2014: jq: limit(n; repeat(f)) silently truncates at 1000 values
**Severity: Medium** — silent truncation. `limit(n; repeat(f))` with `n > 1000` returns 1000 values, exit 0, nothing on stderr.

## Repro (jq 1.7.1 vs `main` @ `d0503b739`)

```console
$ echo null | jq            -c '[limit(5000; repeat(1))] | length'
5000
$ echo null | succinctly jq -c '[limit(5000; repeat(1))] | length'
1000                     # exit 0, stderr empty
```

Scaling shows the cap plainly:

| n | jq | succinctly |
|---|---|---|
| 5 | 5 | 5 |
| 1000 | 1000 | 1000 |
| 10000 | 10000 | **1000** |
| 80000 | 80000 | **1000** |

`repeat(.)` truncates identically. `recurse` and `range` are unaffected (`[limit(5000; recurse(.+1))] | length` and `[limit(5000; range(0;100000))] | length` both give 5000).

## Cause

`eval_repeat` (`src/jq/eval.rs`) bounds its round loop with `const MAX_ITERATIONS: usize = 1000`, and when that loop simply runs out it falls through to `owned_vec_to_result(outputs)` — returning whatever it accumulated, with no error.

The cap itself is well-motivated and its comment says so: an `expr` producing zero outputs per round (`repeat(empty)`) never charges the per-value `budget`, so without a round cap it would spin forever.

**The problem is that the two guards in the same function disagree about how to fail.** The per-value `budget` (`REDUCE_FOREACH_MAX_STEPS`) raises through `charge_budget` — `"repeat: maximum iterations exceeded"` — while `MAX_ITERATIONS` truncates silently. A `repeat` emitting one value per round hits the 1000-round cap long before the 10000-value budget, so the silent path is the one ordinary use takes.

The path-context sibling `resolve_repeat_bounded` gets this right: it raises `"repeat: maximum iterations exceeded"` rather than returning a short list.

## Why `limit` does not save it

`repeat` is infinite by design and is *meant* to be bounded by `limit`/`first`/`until` — that is the documented idiom. Here the eager `eval_repeat` runs its own bounded loop first and hands `limit` an already-truncated list, so `limit(5000; ...)` can only see 1000 values.

## Suggested direction (not attempted)

Raising the constant would only move the cliff. Two options that do not:

1. **Raise instead of truncating** when `MAX_ITERATIONS` is exhausted, matching `resolve_repeat_bounded` and `charge_budget`. Turns silent wrong output into a diagnostic; still caps legitimate large `repeat`s.
2. **Make the eager path demand-driven**, so `limit`/`first` stop `repeat` at the source and the round cap is only reached by a genuinely unbounded consumer. `eval_each`'s `Expr::Limit` arm already exists for exactly this shape (#1461/#1653), so the machinery is there.

Option 1 is small and removes the data loss; option 2 removes the limitation as well.

## Provenance

Found while bounding per-call costs for #1580 — `[limit(1000000; repeat(1))] | length` was the outlier in a benchmark table, and chasing why it was flat turned out to be truncation rather than speed.

newhoggy:
Fixes #2014

## Summary

`eval_repeat` bounded its round loop at a hardcoded 1000 iterations and, when
exhausted, silently returned the short list it had accumulated so far —
`limit(5000; repeat(1))` gave 1000 values, exit 0, nothing on stderr, instead
of jq's 5000.

This adds a native `Expr::Repeat` arm to **both** evaluators' lazy `eval_each`
dispatch — `eval.rs` and its generic-evaluator twin `eval_generic.rs`. This
codebase has two separate jq evaluators, and a fix landing in only one is
silently unreachable from the CLI; I confirmed via tracing that the generic
evaluator (`eval_generic.rs`) is what the real CLI dispatch actually reaches.

Each new arm pulls one round of `f`'s outputs at a time and stops as soon as
the wrapping `limit`/`first` does (via `Demand::Stop`), so
`limit(80000; repeat(1))` now runs all 80000 rounds, matching jq, instead of
hitting an artificial ceiling.

## Two safety nets, both required

Removing either of these was caught by verification (including a real ~2hr
CPU-spinning regression from an earlier attempt with no round-emptiness cap):

- **Per-round budget** (`REDUCE_FOREACH_MAX_STEPS`, reset every round): caps
  how many values a *single round* may fork into at once — a succinctly-only
  memory-safety net for a wide `f` like `.[]`, not jq fidelity (real jq has no
  such cap; confirmed live that `limit(50000; repeat(.[]))` over a
  20001-element array succeeds with all 50000 in jq 1.7.1). This must reset
  **per round** rather than accumulate across rounds — charging cumulatively
  was tried first and silently reintroduced the exact many-narrow-rounds
  truncation this issue is about, just as an error instead of a short count.
- **Round-emptiness cap** (`MAX_EMPTY_REPEAT_ROUNDS`): silently ends the
  stream after 1000 consecutive zero-output rounds, since `repeat`'s `f`
  reruns against the same unchanging input every round — a round producing
  nothing will produce nothing forever, and no wrapping consumer's
  `Demand::Stop` can ever fire to stop it (`sink` is never called on a
  zero-output round).

## Verification

- Issue's own scaling table against pinned jq 1.7.1: n=5/1000/10000/80000 now
  match exactly (previously flat-lined at 1000 for n≥1000).
- Wide-vs-narrow-round distinction: `limit(50000; repeat(.[]))` over a
  20001-element array still caps at exactly 10000 (matching
  `test_path_repeat_width_budget_matches_value_mode_1933`'s path/value
  parity), while `limit(80000; repeat(1))`/`repeat(.)` now succeed fully.
- `break`-to-label, error-with-partial-prefix, `first(repeat(f))`, and bare
  `repeat(empty)` (hang backstop) all re-verified against jq.
- Full local suite green: `cargo test` (default, `simd`, `portable-popcount`),
  `cargo check --no-default-features`, both `clippy --all-targets` feature
  legs, `cargo fmt --check`, `cargo doc --all-features` with
  `RUSTDOCFLAGS=-D warnings`, and the full `cli,simd,regex,serde` workspace
  suite (7000+ tests, including both pinned repeat tests).

## Test plan

- [x] `cargo test --features cli,simd,regex,serde --workspace` — all green
- [x] `cargo test --features simd`
- [x] `cargo test --features portable-popcount`
- [x] `cargo check --no-default-features`
- [x] `cargo doc --no-deps --all-features` (`RUSTDOCFLAGS=-D warnings`)
- [x] Manual scaling table + edge cases verified live against jq 1.7.1

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -1935,7 +1935,7 @@ precedent for a gap that doesn't fit the letter of rule 4 but is kept rather
 than silently left unrecorded. See
 [#1561](https://github.com/rust-works/succinctly/issues/1561).
 
-### `repeat(f)` silently caps at 1000 rounds, in both value and path mode
+### `repeat(f)` under `limit`/`first` (value mode): fixed (#2014); the eager fallback and path mode still cap, in two different ways
 
 `repeat(f)` (`def repeat(f): f, repeat(f);`) has no base case at all: an `f`
 that never errors and never produces a value on some round recurses forever.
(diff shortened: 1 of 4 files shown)


newhoggy:
Ran the mandatory multi-agent code review. Summary of findings and disposition:

**Fixed in follow-up commit (3fe3d5c36):**
- `docs/compliance/jq/limitations.md` was stale: its repro claimed `limit(1500; repeat(1))` still returns 1000, and its thesis described a cap that no longer applies to the `limit`/`first`-bounded path. Rewrote the section to describe all three now-distinct outcomes (fixed / eager-fallback raises / path-mode still caps, unchanged).
- Documented (in both `eval.rs`'s doc comment and limitations.md) that `MAX_EMPTY_REPEAT_ROUNDS` is inherently approximate for a `repeat` body reading `input`/`inputs` — any finite bound has this same edge, so it's recorded as an accepted structural limitation rather than something a bigger constant would fix.
- Added a regression test for the issue's own headline case (`limit(1500; repeat(1))` now returns 1500, not 1000) and for the eager fallback's new raise-instead-of-silently-truncate behavior across all four untested shapes (array construction, `reduce`, `foreach`, `last`).

**Verified, no bug found:**
- `optional`/`?` suppression on the new eager-fallback error path works correctly (`[repeat(1)?] | length` → `1000`, suppressed as expected).
- The `pub(crate)` visibility widening on `REDUCE_FOREACH_MAX_STEPS`/`charge_budget` is scoped exactly as needed and used only by the two intended call sites.
- `limit(-1; repeat(f))` (unlimited) composes correctly with the new lazy arm.

**Deliberately not addressed in this PR (noted, follow-up material):**
- Several DRY/simplification suggestions (the two evaluators' round-control loops share ~20 lines of near-identical bookkeeping; `each_repeat_generic`'s closure-based round materialization could reuse `eval_each_owned_collect`'s existing tuple-returning shape instead of hand-rolling three captured flags). Real, but non-blocking style suggestions — deferred to avoid scope creep on a correctness fix.
- An efficiency observation that every round re-serializes the invariant input to JSON and rebuilds a `JsonIndex` unless `f` hits one of 4 fast-path shapes (`Identity`/`ToString`/`Field`/`Index`) — pre-existing cost in `eval_owned_expr_fork`, not introduced by this fix, but now paid more often since round count is no longer artificially capped at 1000. Worth a follow-up perf issue if `repeat` under `limit` with large `n` and non-trivial `f` shows up in a real workload.
- Path mode's own identical-shaped bug (`resolve_repeat_bounded` still silently truncates at 1000 rounds) is out of scope for this PR (#2014 only covers value mode) and is now explicitly named as an open follow-up in limitations.md.

Full local suite (default/simd/portable-popcount features, `--no-default-features` check, both clippy legs, fmt, doc, and the full `cli,simd,regex,serde` workspace suite — 56 test binaries) is green after the follow-up commit.

Issue #2057, opened 2026-08-31
test: cargo_run_exit's retry loop ignores a Broken pipe stdin write, failing an arbitrary jq_cli test under CI load


newhoggy:
**Severity: Medium** (CI reliability; no product bug). Hit while running #1909's PR (#2045)
through CI, then confirmed pre-existing on `main`.

## Summary

`tests/common/cargo_run_exit.rs`'s `spawn_with_signal_retry` retries a child that died
from a *signal*, but returns unconditionally when writing the child's **stdin** fails.
Under CI contention that write intermittently fails with `Broken pipe (os error 32)`, and
the whole test binary fails with a stack trace that looks nothing like a flake — it names
whichever `run_jq_full` test happened to be running.

## Repro

Not deterministic. Two independent observations of the same signature, ~80 minutes apart,
on **different commits and different architectures**, both in
`test_func_def_arity_overload_unmatched_arity_still_errors_1376`:

| when | branch | job | commit |
|---|---|---|---|
| 10:32Z | `main` | `Test (x86_64)`, [run 33382564073](https://github.com/rust-works/succinctly/actions/runs/33382564073) | unrelated to #2045 |
| 11:52Z | `issue-1909-...` | `Coverage (ARM64)`, [run 33386735561](https://github.com/rust-works/succinctly/actions/runs/33386735561) | `8510bce81` |

Both:

```
---- test_func_def_arity_overload_unmatched_arity_still_errors_1376 stdout ----
Error: Broken pipe (os error 32)

   2: jq_cli_tests::cargo_run_exit::spawn_with_signal_retry::<jq_cli_tests::spawn_jq::{closure#0}>
             at ./tests/common/cargo_run_exit.rs:276:9
   3: jq_cli_tests::spawn_jq
   4: jq_cli_tests::run_jq_full
...
test result: FAILED. 1195 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

The test itself is `def f(x): x + 1; def f(x; y): x + y; [f(1), f(2;3)]` over the four
bytes `null` — nothing about it is slow, large, or racy. It is simply whichever test drew
the short straw; the failure is in the harness, not the case.

## Root cause

`tests/common/cargo_run_exit.rs`, `spawn_with_signal_retry` — `sin.write_all(input)`'s
result is captured, then returned unconditionally via `write_result?` before the
retry-on-signal-death loop can act on it. The comment there is explicit that this is
deliberate:

[...]

PR #2071, opened 2026-08-31
fix(jq): thread an isolated sub-expression's own path into the rest of the pipe

### LINKED ISSUE #1409: jq: isolate-then-continue_rest_with_context loses per-output path threading (Try/Label)
## Summary

Found during #1335's review. `eval_pipe_with_path_context_internal` (`src/jq/eval.rs`)
has a recurring shape: several arms (`Comma`, `Try`, `Label`, and now `Optional`'s
fast path) isolate a sub-expression's evaluation via
`eval_pipe_with_path_context_internal(core::slice::from_ref(sub_expr), ..., current_path,
...)` -- an empty `rest` -- collect its outputs as flat `OwnedValue`s, then continue the
*actual* `rest` of the pipe via `continue_rest_with_context(..., current_path, ...)`,
reusing the **pre-isolation** `current_path`.

This loses per-output path updates whenever the isolated sub-expression is itself
navigational: `Field`/`Index`/`IndexExpr`/`SliceExpr`/`Iterate` only compute and thread
an updated `new_path` when *they* have a non-empty `rest` to recurse into. Evaluated in
isolation (empty `rest`), they just return bare values, so `continue_rest_with_context`
downstream re-enters `rest` using the stale, pre-sub-expression path instead of each
output's real one.

## Repro (confirmed live on `main`, unrelated to `?`)

```
$ echo '{"a":[10,20,30]}' | succinctly jq -c '.a | (try .[] catch empty) | key'
"a"
"a"
"a"
# expected: 0, 1, 2 (matches plain `.a | .[] | key`)

$ echo '{"a":[10,20,30]}' | succinctly jq -c '.a | (.[], empty) | key'
"a"
"a"
"a"
# same wrong result via Comma

$ echo '{"a":[10,20,30]}' | succinctly jq -c '.a | (label $out | .[]) | key'
"a"
"a"
"a"
# same wrong result via Label
```

## Why this wasn't fixed as part of #1335

#1335 needed the same isolate-then-catch shape to fix `Expr::Optional`'s
broadcast-instead-of-atomic-catch bug, and initially introduced the identical
path-threading loss for `Optional` (`.a | (.[])? | key` went from correct `0,1,2` to
wrong `"a","a","a"`) before it was caught in review. That specific regression was
avoided by a narrow guard (only take the isolated fast path when `rest` doesn't itself
consult path context via `needs_path_context`; otherwise fall back to the older,
combined-list evaluation that threads paths correctly at the cost of the narrower
"rest inherits suppression" gap #1335 fixes in the common case) -- see that arm's
comment. That guard is specific to `Optional` and doesn't help `Comma`/`Try`/`Label`,
which still have the unguarded version of this bug today, unchanged by #1335.

## Suggested fix direction

Not investigated in depth. A real fix likely needs the isolated sub-expression's
evaluation to report back `(value, path)` pairs instead of collapsing to bare
`OwnedValue`s, so `continue_rest_with_context`'s per-output continuation can use each
output's actual path -- a materially different return shape than what
`eval_pipe_with_path_context_internal` produces today, so this needs real design work,
not a quick patch. Affects `Comma`, `Try`, `Label`, and any future arm reaching for the
same isolate-then-continue shape.

newhoggy:
Closes #1409.

## The bug

Several arms of `eval_pipe_with_path_context_internal` evaluate a sub-expression
*in isolation* — recursing with an empty `rest` — because the construct scopes
something `rest` must stay outside of, and only afterwards continue the real
`rest`. `Field`/`Index`/`IndexExpr`/`SliceExpr`/`Iterate` compute and thread an
updated `new_path` only when *they* have a non-empty `rest` to recurse into, so
an isolated navigational sub-expression handed the continuation bare values, and
`rest` re-entered under the stale, pre-sub-expression path:

```console
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a | (try .[] catch empty) | key'
"a"
"a"
"a"
# expected 0, 1, 2 — what `.a | .[] | key` gives
```

## Scope is wider than the issue recorded

#1409 names `Try`/`Label`. The same shape also affects `If`, `FuncDef` and
`Limit`, and `As`/`AsPattern` wherever their arm is reached. (`Comma` was fixed
by #1509; `Optional` is guarded by #1335.) All are fixed here except the `As`
routing case, deliberately — see *Left open* below.

## The fix

No new `(value, path)` return shape is needed, contrary to the issue's stated
direction. `path_probe_stage` appends one ordinary jq expression — `[path, .]` —
to the isolated pipe. That makes the pipe non-empty, so the navigational arms
thread paths through it exactly as they always have; `Builtin::PathNoArg`
renders the path each output reached, and `Expr::Array`'s `needs_path_context`
arm pairs it with the value. `continue_rest_with_paths` splits each pair back
apart and continues `rest` from that output's own path. No new AST variant, no
second evaluation of the sub-expression, and the whole thing is gated on `rest`
actually consulting path context so the common path is unchanged.

The probe sits *inside* whatever the construct scopes, so scoping is untouched —
it cannot itself error, and `rest` stays outside the `try`'s handler. That is
precisely what `Comma`'s own fix could not offer these arms: combining the
sub-expression with `rest` would have put `rest`'s errors inside the handler's
reach, which real jq never does. Pinned by
`test_isolated_subexpr_probe_does_not_widen_catch_scope_1409`.

A `catch` handler's outputs are deliberately **not** probed. The handler runs on
the error payload, which is not a node of the document, so probing it named a
path that does not exist — `try error({"b":9}) catch .b` claimed `["a","b"]`.
Real jq refuses the question outright (`Invalid path expression`), so no oracle
demands a deeper answer, and the ambient path is kept.

## Oracle

Ground truth is `path()`, whose own resolver (`resolve_seq`/`walk_path`) was
already correct for every one of these shapes and agrees with jq 1.7.1 exactly:

```console
$ echo '{"a":[1,2,3]}' | jq -c 'path(.a | (try .[] catch empty))'
["a",0]
["a",1]
["a",2]
```

So wrapper and un-wrapped form disagreeing was unambiguously the bug, not an
open question.

## Changed expectations (4)

All one family: a literal output inside a wrapper used to inherit the wrapper's
ambient path. Each now matches what the un-wrapped pipeline already reported,
and the oracle backs the new value — `yq ea '1 | fileIndex' f1 f2` and
`yq ea '1 as $x | $x | fileIndex' f1 f2` both give `0`, `0`, not the file's
index.

| Test | Was | Now |
|---|---|---|
| `test_try_without_catch_prepends_partial_error_prefix_with_path_context` | `7`, `8` | `0`, `0` |
| `test_try_without_catch_prepends_partial_break_prefix_with_path_context` | `7`, `8` | `0`, `0` |
| `test_if_cond_many_owned_with_path_context` | `5`, `5` | `0`, `0` |
| `test_func_def_path_context_builtins_1306` (`(def f: 5; f) \| key`) | `"a"` | `null` |

## Left open, each filed separately

- **`As`/`AsPattern` routing.** Widening the guard to consider `rest` fixes
  `. as $x | .[] | key`, but regresses `.x as $v | $v | fileIndex` to `0` where
  real yq gives the per-file index — `substitute_bound_var` re-materialises the
  bound value as a literal and erases its node identity. Needs bindings to carry
  that identity; reverted from this PR rather than shipped as a trade.
- **`Optional`'s #1335 guard.** The probe can replace it, which also closes the
  "rest inherits suppression" gap (`.a | (.[])? | (key, error("boom"))` silently
  swallows an error jq 1.7.1 raises with exit 5). But it changes an
  evaluator-wide model that #1826 and #1869 both build on, so it wants its own
  change and its own oracle work.
- **`first`/`last`.** `Expr::FirstExpr`/`LastExpr` have no `needs_path_context`
  arm and no path-context evaluator arm at all, so `first(key)` is `null` — a
  separate pre-existing gap, not this shape.

## Verification

Every `ci.yml` gate run locally on the rebased tip: workspace suite under
`cli,simd,regex,serde`; plain `cargo test`; `simd,scalar-yaml`; both clippy legs
with `-D warnings`; `cargo check --no-default-features`; `cargo fmt --all --check`;
`RUSTDOCFLAGS=-D warnings cargo doc --no-deps --all-features`; fuzz `cargo check`.

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -28552,6 +28552,198 @@ fn continue_rest_with_context<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     }
 }
 
+/// The stage appended to an isolated sub-expression's own pipe so that each
+/// of its outputs carries the `current_path` it actually landed on: the
+/// ordinary jq expression `[path, .]` (#1409).
+///
+/// Several arms of [`eval_pipe_with_path_context_internal`] have to evaluate
+/// a sub-expression *in isolation* -- recursing with an empty `rest` -- and
+/// only afterwards continue the real `rest`, because the construct scopes
+/// something `rest` must stay outside of: `try`'s catch handler, `label`'s
+/// break scope, `?`'s suppression. Isolation is what loses the path.
+/// `Field`/`Index`/`IndexExpr`/`SliceExpr`/`Iterate` compute and thread an
+/// updated `new_path` only when *they* have a non-empty `rest` to recurse
+/// into; evaluated with an empty one they hand back bare values, so the
+/// separate continuation afterwards re-entered `rest` under the stale,
+/// pre-sub-expression path (`.a | (try .[] catch empty) | key` reported
+/// every element's key as `"a"` instead of its index).
+///
+/// The fix does *not* need the new `(value, path)` return shape #1409
+/// originally called for. Appending this one stage makes the isolated pipe
+/// non-empty, so the navigational arms thread paths through it exactly as
+/// they always have, and `Builtin::PathNoArg`'s fast path at the top of
+/// [`eval_pipe_with_path_context_internal`] renders the path each output
+/// reached. `Expr::Array`'s own `needs_path_context` arm evaluates its
+/// contents in path context, so `[path, .]` pairs that path with the value
+/// -- all out of nodes this evaluator already handles, with no new AST
+/// variant and no second evaluation of the sub-expression.
+///
+/// Crucially the probe sits *inside* whatever the construct scopes, so the
+/// scoping is untouched: it cannot itself error, and `rest` stays outside
+/// the `try`'s handler. That is what the `Comma` arm's own fix (#1509)
+/// could not offer `Try`/`Label` -- combining the sub-expression with
+/// `rest` into one recursive call would have put `rest`'s own errors inside
+/// the handler's reach, which real jq never does.
+///
+/// [`continue_rest_with_paths`] is the matching continuation: it splits each
+/// pair back apart and continues `rest` from that output's own path.
+fn path_probe_stage() -> Expr {
+    Expr::Array(Box::new(Expr::Comma(vec![
+        Expr::Builtin(Builtin::PathNoArg),
+        Expr::Identity,
+    ])))
+}
+
+/// Split one [`path_probe_stage`] output back into `(path, value)`.
+///
+/// `fallback_path` answers the shapes the probe never produces (anything
+/// that is not a 2-element array whose first element is an array), which a
+/// correct probe run cannot reach -- kept as a total function rather than
+/// an `unreachable!` because the encoding travels through
+/// `QueryResult`/`OwnedValue` rather than a dedicated type, so a future arm
+/// wiring the pair up wrongly should degrade to the old ambient-path
+/// behavior rather than panic.
+fn split_probe_pair(
+    paired: OwnedValue,
+    fallback_path: &[OwnedValue],
+) -> (Vec<OwnedValue>, OwnedValue) {
+    if let OwnedValue::Array(pair) = paired {
+        if pair.len() == 2 {
+            let mut it = pair.into_iter();
+            let path = it.next().unwrap();
+            let value = it.next().unwrap();
+            if let OwnedValue::Array(path) = path {
+                return (path, value);
+            }
+            return (fallback_path.to_vec(), value);
+        }
+        return (fallback_path.to_vec(), OwnedValue::Array(pair));
+    }
+    (fallback_path.to_vec(), paired)
+}
+
+/// Wrap each output of an *unprobed* result into the same `[path, value]`
+/// encoding [`path_probe_stage`] produces, all sharing one ambient `path`.
+/// Used for a `catch` handler's outputs: the handler runs on the error
+/// payload, which is not a document node, so probing it would fabricate a
+/// deeper path than any that exists (`try error({"b":9}) catch .b` would
+/// claim `["a","b"]`).
+fn pair_outputs_with_path<'a, W: Clone + AsRef<[u64]>>(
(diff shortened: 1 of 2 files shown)


newhoggy:
**Blocked on https://github.com/rust-works/succinctly/pull/2076, not on anything in this diff.**

`Coverage (ARM64)` is red here and will stay red until that lands. Every other
check on this commit is green, including `Coverage (x86_64)`, every `Test` leg,
both `Clippy` legs and `Commit Message Lint`.

The failure is the harness bug tracked in
https://github.com/rust-works/succinctly/issues/2057 — `write stdin / Broken
pipe (os error 32)` at `cargo_run_exit.rs:210`, not an assertion. It has now hit
this PR twice on two *different* tests, with a `gh run rerun --failed` in
between:

| attempt | test |
|---|---|
| 1 | `test_unresolved_call_is_not_catchable_1473` |
| 2 | `test_forward_reference_across_names_is_a_compile_error_1473` |

Neither test is touched by this diff, and neither can reach the changed code:
both are *compile-error* tests, so evaluation never begins, while this PR only
changes path threading inside `eval_pipe_with_path_context_internal`.

Investigating the second failure did explain why re-running doesn't help, which
the linked issue previously left open. All four recorded observations are
compile-error tests — `succinctly jq` reports the error and exits 3 without ever
reading stdin, so the parent's write to a closed pipe is the normal outcome, not
a contention symptom. On a fast runner the child wins that race
deterministically. PR 2076 fixes that and pins it with a test that fails
deterministically without the fix.

One thing worth a reviewer's attention rather than my assertion: this PR adds
~14 subprocess spawns to `jq_cli_tests`, and `main` is currently 6/6 green on
`Coverage (ARM64)`. I could not construct a mechanism by which extra spawns
cause a *deterministic* child-exits-first race, and the compile-error shape
explains the failures without reference to this PR — but the timing correlation
is real and I would rather flag it than have it noticed later.

PR #2081, opened 2026-09-01
fix(jq): alpha-rename callee binders that would capture a substituted arg

### LINKED ISSUE #2077: jq: function-argument substitution is unhygienic — a caller's variable is captured by a same-named binding in the callee, silently
**Severity: High** — silent wrong answers, no error, on ordinary jq that users write.

## Summary

Found while investigating #1371. `succinctly jq` expands `def name(params): body;`
by **statically substituting** each call's argument expressions into `body`
(`expand_func_calls`, `src/jq/eval.rs`). That substitution is **unhygienic**: an
argument that references a caller variable is captured by a same-named binding
inside the callee's body, so it silently resolves to the callee's value instead
of the caller's.

Real jq binds a function argument as a **closure over the caller's environment**,
so the callee's own `as $x` cannot capture it.

This is the same root cause #1371 tracks (static substitution instead of runtime
calls with real environments), but a different and more serious symptom: #1371
produces a clean error at depth, this produces a *wrong number* at depth 1.

## Repro (all confirmed live against `/usr/bin/jq` 1.7.1 and merged `main` 0a39f6765)

| filter | jq 1.7.1 | `succinctly jq` |
|---|---|---|
| `def f(g): 5 as $x \| g; 2 as $x \| f($x)` | `2` | **`5`** |
| `def f(g): 99 as $y \| [g]; 1 as $y \| f($y)` | `[1]` | **`[99]`** |
| `def f(g): "inner" as $v \| g; "outer" as $v \| f($v)` | `"outer"` | **`"inner"`** |
| `def f(g): [1,2] as $x \| g; [9] as $x \| f($x\|length)` | `1` | **`2`** |
| `def add(g): 100 as $n \| g + $n; 1 as $n \| add($n)` | `101` | **`200`** |
| `def f(g): reduce range(2) as $x (0; . + g); 7 as $x \| f($x)` | `14` | **`1`** |

Every row exits 0. Nothing warns.

The `add` row is the one to look at: a caller passes its own `$n` and gets the
callee's `$n` back, so `1 + 100` becomes `100 + 100`.

The last row shows the capturing binder need not be an `as` at all — `reduce`'s
own pattern variable captures it too, which widens the surface to `foreach`,
`reduce`, `as`, `?//` patterns, and function parameters themselves.

## Control: a *variable* reference, with no argument, is fine

```
$ jq -n '1 as $x | def f: $x; 2 as $x | f'              # 1
$ succinctly jq -n '1 as $x | def f: $x; 2 as $x | f'   # 1  -- agrees
```

So the bug is specifically in **argument** substitution, not in how a `def` body
sees ambient variables. That narrows where a fix has to look.

## Why this is worth its own issue rather than a note on #1371

#1371 is scoped as "recursive `def`s can't reach useful depth", and its suggested
fix (a runtime, environment-based call mechanism) would resolve this too. But the
two have very different risk profiles and would justify very different amounts of
work on their own:

- #1371's symptom is a clean error on an uncommon shape (deep or branching
  recursion). Nothing silently wrong.
- This is a silently wrong *value* on a completely ordinary shape — passing a
  variable to a helper `def` — with no recursion involved at all.

If the full architectural fix is deferred, this may still deserve a narrower
mitigation on its own (alpha-renaming the callee's binders before substituting,
which is the standard hygiene fix and does not need a new evaluator).

## Suggested direction

Two options, in increasing order of size:

1. **Alpha-rename during substitution.** Before substituting an argument into
   `body`, rename every binder in `body` that could capture a free variable of
   the argument. Contained within `expand_func_calls`, no evaluator change. Needs
   care to cover every binding form (`as`, `reduce`, `foreach`, `?//` patterns,
   nested `def` parameters).
2. **Runtime calls with environments** (#1371's own fix), which makes the
   question moot because arguments are never substituted at all.

Not investigated in depth — filed so the correctness half is tracked separately
from #1371's depth half, and so option 1 stays available if #1371 is deferred.

newhoggy:
## Summary

- `expand_func_calls` inlines a `def`'s body by statically substituting each call's argument expressions into it. That substitution was unhygienic: an argument referencing a caller variable was captured by a same-named binding inside the callee's own body (`as`, `reduce`, `foreach`, `?//` patterns, and a nested `def`'s own `$`-parameters), silently resolving to the callee's value instead of the caller's.
- `substitute_func_param`'s binder arms (`As`, `Reduce`, `Foreach`, `AsPattern`, nested `FuncDef` params) now alpha-rename any binder whose name collides with a free variable of the argument being substituted, before recursing into that binder's own scope. The rename reuses `substitute_func_param` itself as the primitive (replace the old name with a fresh `Var` reference), so it gets the same shadow-aware recursion the real substitution already needed. The fresh name always contains `#`, which no jq/yq identifier can spell, so it can never collide with anything the user actually wrote.
- Separately (defect B in the issue): a nested `def` of the same name *and* arity as the parameter (`def f(g): def g: 99; g; f(1)`) shadows the parameter within `then`, not just within its own `body` — the existing shadow check only gated `body`, so `then` was still substituted, ignoring the nested `def` entirely.

Both defects and the issue's own control cases are verified live against `/usr/bin/jq` 1.7.1.

Fixes #2077

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 3195+ tests, 0 failed)
- [x] Every repro row and control case from the issue diffed live against `/usr/bin/jq` 1.7.1 — all match
- [x] Additional capture scenarios not in the issue's table (foreach pattern+extract, as-pattern array destructure, `?//` alternatives, nested-def own `$`-param) diffed live against jq 1.7.1 — all match
- [x] 7 new regression tests added covering both defects and 2 controls
- [x] `cargo llvm-cov --features cli,simd,regex,serde --workspace --summary-only`

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -26017,16 +26017,33 @@ fn fresh_var_name(name: &str) -> String {
 
 /// Whether `expr` contains a free-standing `$name` reference anywhere.
 ///
+/// Every `$name` referenced anywhere in `expr`.
+///
 /// #2077: used to decide whether substituting `arg` into a binder's scope
 /// would let that binder recapture one of `arg`'s own variable references.
 /// Deliberately not scope-aware over `expr` itself -- a `$name` that is
-/// actually bound *within* `expr` (not free) still trips this, which only
-/// costs an unnecessary rename. Renaming a binder that would not actually
-/// have captured anything is still alpha-equivalent, so over-approximating
-/// here is safe, unlike the rename itself (see [`capturing_pattern_renames`]),
-/// which must not touch a binder that isn't actually `name`.
-fn expr_mentions_var(expr: &Expr, name: &str) -> bool {
-    any_subexpr(expr, &mut |e| matches!(e, Expr::Var(n) if n == name))
+/// actually bound *within* `expr` (not free) still ends up in the set, which
+/// only costs an unnecessary rename. Renaming a binder that would not
+/// actually have captured anything is still alpha-equivalent, so
+/// over-approximating here is safe, unlike the rename itself (see
+/// [`capturing_pattern_renames`]), which must not touch a binder that isn't
+/// actually free in `arg`.
+///
+/// One walk regardless of how many names are later checked against it --
+/// [`capturing_pattern_renames`] used to call an `any_subexpr`-per-name
+/// membership test instead, which re-walked `arg` from scratch for every
+/// bound name (a wide `?//` alternation or a many-parameter nested `def`
+/// turned that into a multiplicative cost, and this runs fresh on every
+/// evaluation of a `def` node -- `expand_func_calls` is not cached).
+fn expr_free_vars(expr: &Expr) -> BTreeSet<String> {
+    let mut names = BTreeSet::new();
+    any_subexpr(expr, &mut |e| {
+        if let Expr::Var(n) = e {
+            names.insert(n.clone());
+        }
+        false
+    });
+    names
 }
 
 /// For every distinct name `patterns` binds that also appears as a free
@@ -26049,9 +26066,14 @@ fn capturing_pattern_renames(patterns: &[Pattern], arg: &Expr) -> Vec<(String, S
     }
     bound_names.sort();
     bound_names.dedup();
+    if bound_names.is_empty() {
+        return Vec::new();
+    }
+
+    let free_vars = expr_free_vars(arg);
     bound_names
         .into_iter()
-        .filter(|name| expr_mentions_var(arg, name))
+        .filter(|name| free_vars.contains(name.as_str()))
         .map(|name| {
             let fresh = fresh_var_name(&name);
             (name, fresh)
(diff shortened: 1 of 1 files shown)


newhoggy:
Code review (2 independent angles) found no confirmed correctness divergences in this diff. Applied 2 cleanup fixes it did surface (reuse `apply_pattern_renames` in the `FuncDef` arm instead of hand-rolling it; single-walk `arg`'s free variables instead of one `any_subexpr` walk per bound name). A third, unconfirmed lead (bare non-$ function-style parameter capture, not covered by this fix's $-variable-only renaming) is filed separately as #2088 rather than expanding this PR's scope — no observable divergence against jq 1.7.1 was found despite several constructed attempts.

newhoggy:
Patch coverage: 99.37% (158/159 new lines) via `omni-dev coverage diff` against a local `cargo llvm-cov --features cli,simd,regex,serde --workspace --json` report. The one remaining uncovered line (`src/jq/eval.rs:41223`) is an `unreachable!()` arm documenting a structural invariant (the `FuncDef` param-rename mapping is built entirely from `Pattern::Var`, which `apply_pattern_renames` never changes the shape of) rather than a reachable code path — genuinely untestable, not a coverage gap.

Issue #1381, opened 2026-08-21
jq: chained self-recursive defs (each calling the previous as its base case) still cause unbounded memory blowup


newhoggy:
## Summary

Follow-up to #1016/PR #1372 (adds a depth/budget guard so a *single* self-recursive `def` fails cleanly instead of SIGABRT-ing). This issue tracks a related but distinct pattern that guard does not close: **chaining several small recursive `def`s, each calling the previous one as its base case**, still causes unbounded memory blowup — confirmed pre-existing on `main` (crashes with SIGABRT there too, via a different mechanism), and still present after #1016's fix (no longer an instant crash, but consumes multiple GB and can run for many seconds without terminating).

## Repro

```
$ echo null | succinctly jq -c 'def r0(n): if n==0 then . else [r0(n-1)] end; def r1(n): if n==0 then r0(n) else [r1(n-1)] end; def r2(n): if n==0 then r1(n) else [r2(n-1)] end; r2(45)'
```

Measured on PR #1372's own branch (post-fix): 3 chained defs (r0..r2) consumed 3.4+ GB peak memory; a reviewer measured 4 chained defs at 27+ GB and still climbing at a 15s timeout. On `main` (pre-#1016-fix), the same shape crashes with SIGABRT instead (confirmed live: `def r0(n): ...; def r1(n): ...; def r2(n): ...; r2(3)` overflows the stack).

## Why #1016's guards don't catch this

`expand_func_calls` evaluates `def`s via nested `eval_func_def` calls: the outer `def r0` expansion pass finds and substitutes the `r0(n)` call embedded inside `r1`'s own body *before* `r1`'s own `eval_func_def`/expansion pass ever runs. By the time `r1`'s own self-reference substitution loop runs, its `body` parameter is no longer r1's small original 3-node body — it already contains r0's fully-expanded ~50-node substitute embedded where `r0(n)` used to be. r1's own budget (bounded at 50 hits by #1016's fix) still clones *that* now-large body up to 50 times, multiplying total size by ~50x. Each additional chained def repeats this multiplication.

Neither of #1016's two guards catches this:
- `budget` (`MAX_FUNC_EXPANSION_DEPTH`) bounds *hit count* per occurrence, not the *size* of the body being cloned at each hit.
[...]

PR #2082, opened 2026-09-01
fix(jq): report every unresolved call, not just the first

### LINKED ISSUE #2037: jq: compile-error locations are approximate — no source position on Expr::FuncCall (#1473 remainder)
Deliberate remainder from #1473, recorded in [docs/compliance/jq/limitations.md](https://github.com/rust-works/succinctly/blob/main/docs/compliance/jq/limitations.md) at merge time rather than left undocumented.

## What's exact today, and what isn't

#1473 made an unresolvable function call a compile error matching jq's wording and exit code. The **location** in that message is derived by searching the filter source for the offending identifier (`locate_identifier`, `src/bin/succinctly/jq_runner.rs`), because `Expr::FuncCall` carries no source position. Three consequences:

**1. A repeated undefined name cites the first occurrence.**

```
$ jq -n 'def f: 1;
nosuch
| nosuch'
jq: error: nosuch/0 is not defined at <top-level>, line 2:    # jq points at the *failing* one
```
succinctly's search is word-bounded and correct for the common case, but with the same name written twice it reports the first occurrence's line, which may not be the call that failed. Every other part of the message — name, arity, exit code — is exact.

**2. Only the first unresolvable call is reported.** `jq::resolve_func_calls` returns on the first failure, so succinctly always prints `jq: 1 compile error`; jq reports every one and prints `jq: N compile errors`.

**3. A call from an `include`d module or `~/.jq` reports no location at all.**

```
$ jq -L /path -n 'include "mymod"; helper'
jq: error: nosuchfn_in_module/0 is not defined at /path/mymod.jq, line 1:
def helper: nosuchfn_in_module;
jq: 1 compile error

$ succinctly jq -L /path -n 'include "mymod"; helper'
jq: error: nosuchfn_in_module/0 is not defined at <top-level>
jq: 1 compile error
```

The call has no occurrence in the filter source to locate, so the line marker and source echo are dropped rather than a position invented. Pinned by `test_unresolved_call_from_included_module_omits_the_location_1473` (`tests/jq_cli_tests.rs`).

**4. jq's trailing padding on the echoed source line is not reproduced exactly.** jq pads with a `%*s` whose width follows the failing node's start column for a simple undefined name, but points elsewhere for an arity mismatch (`def f(x): x; f(1;2)` pads 17 where the call starts at column 14). succinctly reproduces the column rule. This is trailing whitespace and invisible in a terminal; noted only for byte-exact comparison.

## Why it wasn't done in #1473

The fix is a source position on `Expr::FuncCall`. Only 14 construction sites exist, so it looks small, but it has two non-obvious costs:

- **It perturbs `format!("{body:?}").len()`**, which #1381's `MAX_FUNC_EXPANSION_WEIGHTED_COST` is calibrated against — that constant's doc comment records that it already needed *two* calibration passes against measured real `def` bodies (1372–1561 chars). Adding a field to every `FuncCall` in a body inflates that length and would silently shift the recursion budget.
- **`Expr` derives `PartialEq`**, so two structurally identical calls at different positions stop comparing equal. Parser unit tests that `assert_eq!` on parsed trees would need positions supplied.

Item 3 additionally needs the originating module file threaded through `ModuleLoader` (`src/bin/succinctly/jq_runner.rs`), which today discards which file a def came from once it inlines it as a `FuncDef` wrapper.

## Scope

Low priority. The name, arity, exit code and uncatchability — everything that changes what a script *does* — already match jq exactly; this is diagnostic precision only. Worth doing alongside any other work that adds positions to the AST, since the `MAX_FUNC_EXPANSION_WEIGHTED_COST` recalibration is the real cost and is best paid once.

newhoggy:
## Summary

Fixes #2037.

`jq::resolve_func_calls` stopped its AST traversal at the first call it couldn't
resolve, so `succinctly jq` always printed `jq: 1 compile error` and only ever
located the *first* textual occurrence of the offending name — even when a
filter had several unresolvable calls, or the same undefined name written more
than once. Real jq's compiler keeps going and reports every unresolvable call
in one pass (`jq: N compile errors`).

- Added `jq::resolve_func_calls_all`, which walks the whole program collecting
  every unresolvable call instead of short-circuiting on the first (`resolve_func_calls`
  is now defined in terms of it, taking just the first — existing callers/tests
  are unaffected).
- The jq runner's `report_unresolved_calls` (renamed from `report_unresolved_call`)
  now prints one line per unresolved call and the correct `jq: N compile errors`
  count. To locate a repeated name's *own* occurrence rather than always citing
  the first, each successive lookup for the same name resumes the source-text
  search right after the previous match (`locate_identifier_from`).
- Confirmed byte-for-byte against the pinned oracle (`/usr/bin/jq` 1.7.1),
  including the exact repro quoted in #2037's issue body.

This does **not** attempt the deeper, genuinely invasive part of #2037 (a real
source position on `Expr::FuncCall`, which the issue itself flags as perturbing
`format!("{body:?}").len()` — #1381's `MAX_FUNC_EXPANSION_WEIGHTED_COST` — and
needing the originating module file threaded through `ModuleLoader` for the
included-module case). Those two remaining approximations are unchanged and
are recorded, updated to reflect what's now fixed, in
`docs/compliance/jq/limitations.md`.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 3195+ tests, 0 failed)
- [x] All 15 pre-existing `_1473` tests in `tests/jq_cli_tests.rs` pass unchanged
- [x] 3 new tests added, verified byte-for-byte against the pinned jq 1.7.1 oracle:
  `test_all_unresolved_calls_are_reported_not_just_the_first_2037`,
  `test_repeated_unresolved_name_cites_its_own_occurrence_2037`,
  `test_original_2037_repro_reports_both_lines`
- [x] `omni-dev coverage diff`: 100% patch coverage (87/87 new lines)

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -1398,15 +1398,25 @@ Two deliberate remainders:
   the AST would perturb `format!("{body:?}").len()`, which #1381's
   `MAX_FUNC_EXPANSION_WEIGHTED_COST` is calibrated against, so this stays a textual search
   rather than a real position lookup. [#2037](https://github.com/rust-works/succinctly/issues/2037)
-  closed the two sharpest edges of that: every unresolvable call is now reported, not just
-  the first (`jq: N compile errors`, matching jq's own count), and a name mentioned more
-  than once has each of its occurrences located independently — the search for the *k*-th
-  reported call of a given name resumes right after the (*k*-1)-th one's match, rather than
-  re-finding the first occurrence every time. This matches jq byte-for-byte whenever
-  `resolve_func_calls_all`'s traversal order agrees with a left-to-right textual scan,
-  which holds for every existing `Expr` variant, but it is still a heuristic over source
-  text, not a proof from real positions — a construct that visits the same name's call
-  sites in an order the text scan wouldn't predict could still cite the wrong occurrence.
+  closed two edges of that: every unresolvable call is now reported, not just the first
+  (`jq: N compile errors`, matching jq's own count — and, matching real jq's own compiler,
+  an unresolved callee's *arguments* are no longer independently checked and reported,
+  since jq itself never compiles them without a resolved callee to bind them to), and a
+  name mentioned more than once has each of its occurrences located independently — the
+  search for the *k*-th reported call of a given name resumes right after the (*k*-1)-th
+  one's match, rather than re-finding the first occurrence every time.
+
+  This is still a pure textual heuristic, not a proof from real positions, and it can
+  misfire two distinct ways: a construct whose traversal order disagrees with a
+  left-to-right text scan (none exist among today's `Expr` variants, but nothing enforces
+  that going forward), and — already true before #2037, unchanged by it — an unrelated
+  occurrence of the same spelling (an object key, a variable) earlier in the source is
+  indistinguishable from the real call site to a pure text scan: `{nosuch: 1} | nosuch`
+  cites the harmless object key on line 1 instead of the actual failing call on line 2 —
+  tracked separately as
+  [#2085](https://github.com/rust-works/succinctly/issues/2085), since it predates #2037
+  and neither of #2037's fixes touch it. Closing either gap for real needs the same source
+  position on `Expr::FuncCall` this whole approach exists to avoid adding.
 - **A call reached through an `include`d module or `~/.jq` reports no location at all.** It
   has no occurrence in the filter source to locate, so the line marker and source echo are
   dropped rather than a position invented: `nosuchfn/0 is not defined at <top-level>` where
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -1394,11 +1394,19 @@ are now rejected before it runs.
 Two deliberate remainders:
 
 - **The reported line is located by searching the filter source for the offending
-  identifier**, since `Expr::FuncCall` carries no source position. A filter mentioning the
-  same undefined name more than once cites the first occurrence's line, and only the first
-  unresolvable call is reported (`jq: 1 compile error`) where jq reports all of them. Adding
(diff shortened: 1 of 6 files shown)


newhoggy:
Ran a standard-effort multi-agent code review before merge. It caught a real behavioral regression: the initial version of this fix (collect-all instead of stop-at-first) descended into an unresolved callee's own arguments and reported them too, where real jq's compiler never compiles a call's arguments without a resolved callee to bind them to (verified live: `nosucha(nosuchb; nosuchc)` is one error in jq 1.7.1, not three). Fixed by checking the callee before its arguments, matching jq exactly.

Review also flagged (and I fixed):
- `resume_from` was a linearly-scanned `Vec` where this same file already established a `HashMap` idiom for the identical shape (`collapse_duplicate_fields`'s `slot_of`) — switched.
- Stale doc-comment references to the pre-rename function names (`report_unresolved_call`/`locate_identifier`) in `tests/jq_cli_tests.rs`.
- A missing comment in `yq_runner.rs` explaining why it deliberately keeps the singular, first-error-only `resolve_func_calls` rather than adopting `resolve_func_calls_all`.
- The doc comments overstated the resumed-search heuristic's guarantees. Review independently found (and I verified live against `/usr/bin/jq` 1.7.1) a separate, pre-existing gap: an unresolved call's name can collide in spelling with an unrelated object key or variable earlier in the source (`{nosuch: 1} | nosuch` cites the harmless object key instead of the real call two lines later). This predates #2037 and isn't fixed by it, so I filed it separately as #2085 and referenced it from the docs rather than leaving it implied.

Added two more tests (`test_unresolved_callee_args_are_not_independently_checked_2037`, `test_resolved_callee_still_checks_its_own_arguments_2037`) pinning the argument-checking-order fix, each verified byte-for-byte against the oracle. Full suite, both clippy legs, and patch coverage (100%, 86/86 new lines) all still green after these changes.

Issue #1994, opened 2026-08-31
scripts: promote the streaming-vs-DOM output differential to a reusable harness


newhoggy:
**Severity: Low** (tooling; no user-visible behaviour)

## Summary

While fixing #1700 I wrote a throwaway differential harness that compares succinctly's
**streaming** output against its **DOM** output for the same document, across output modes.
It found #1982 (two real DOM-path divergences) on its first run. It currently lives in
`.ai/scratch/`, which is gitignored, so it dies with the worktree.

The technique is worth keeping, because this repo keeps rediscovering the same class of bug.

## What it does

For each document in a corpus x each output mode, run the same query twice and diff:

- **streaming**: `syq -o json -I0 <flags> '.'`
- **DOM**: the identical command plus `--arg unused 1`, which forces materialization
  (`context.named` non-empty) while changing nothing else about the output config

Any difference is either a genuine streaming/DOM divergence or a bug in whichever path was
just changed. Running a no-op control mode (one without the flag under test) separates
"pre-existing divergence" from "regression I just introduced" — that is exactly how #1700
established that all 7 of its diffs were pre-existing rather than caused by the change.

## Why it generalizes

`docs/compliance/yq/limitations.md` has an entire section on routing divergences, whose
stated lesson is *"when auditing what a construct loses, the question to ask is which output
route it takes"*. #442, #478, #757, #868, #1342, #1343, #1344, #1398, #1693, #1700 and #1982
are all instances. A harness that mechanically diffs the two routes turns that lesson from a
thing reviewers must remember into a thing CI could check.

## Suggested shape

- Live in `scripts/` beside `ab-cli.py`, stdlib-only, same house style.
- Parameterise: corpus dir, output modes, the flag(s) under test, and which control mode
  distinguishes pre-existing from new.
- Ship the #1700 corpus as a starting set (22 YAML documents covering quoted/single-quoted/
  block scalars, escapes, non-ASCII keys, astral chars, tags, anchors, multi-doc).
- Report per-case diffs classified as *pre-existing* (also differs in the control mode) vs
  *introduced*, so a caller can gate on the second category alone.

## Related

- #1700 / #1983 — where the harness was written and used.
- #1982 — what it found.

newhoggy:
🔧 Claimed by agent `Agatha` — branch `issue-1994-streaming-dom-differential-harness`

newhoggy:
[...]

Issue #1770, opened 2026-08-28
jq: keys_unsorted[] raises on a malformed object, but first(keys_unsorted[])/limit(...) don't


newhoggy:
**Severity: Low-Medium** — found during #1629's code review; live-reproduced against this session's build.

## Summary

#1629 fixed `keys_unsorted[]`/`keys_unsorted | last`/`keys_unsorted[-n]` to raise on a malformed
object (#1194) by adding the check to `fold_lazy_keys_stage`'s arms in `src/jq/eval_generic.rs`.
But `Expr::Iterate` has a second, separate evaluator path: `fold_pipe_stages_sink` (the
demand-aware evaluator behind `first(...)`, `limit(n; ...)`, `while`/`until`, and other
early-terminating consumers) special-cases `Expr::Iterate` *before* it ever reaches
`fold_lazy_keys_stage`, routing instead through `each_lazy_keys_iterate_sink`. That function still
streams the raw, unchecked `DistinctKeyCursors` iterator directly — no `key_is_malformed` check per
key, no `ended_unpaired()` check at exhaustion.

## Repro (verified against #1629's build)

```console
$ echo '{123: 1, "b": 2}' | sjq -c 'keys_unsorted[]'        # exit 5 -- correct (fixed by #1629)
$ echo '{123: 1, "b": 2}' | sjq -c 'first(keys_unsorted[])' # 123, exit 0 -- wrong
$ echo '{123: 1, "b": 2}' | sjq -c 'limit(1; keys_unsorted[])'  # 123, exit 0 -- wrong
```

Same split for the unpaired-tail shape (`{"a":1,"b"}`): bare `keys_unsorted[]` raises,
`first(keys_unsorted[])` doesn't.

## Why this isn't a trivial extension of #1629's fix

`each_lazy_keys_iterate_sink` exists specifically to *not* pay for a full walk when the consumer
stops pulling early — `first`/`limit(1;...)` are the whole reason it's demand-driven rather than
collecting. Adding a whole-object #1194 check up front would mean paying for a full walk on every
`first(keys_unsorted[])` call, defeating the point of the demand-aware path entirely (the same
tension #1629's own "Option 1" vs "Option 2" analysis named for the positional fast paths).

This isn't unprecedented territory, though: `each_lazy_seq_iterate_sink` (the `LazySeq` sibling of
this function, same file) already answers an analogous question — whether `first(map(f) | .[] | g)`
[...]

Issue #1653, opened 2026-08-26
jq: --unbuffered doesn't achieve real-time stdout/stderr interleaving — one input's whole output batches before any of it writes


newhoggy:
**Severity: Low**

## Summary

Found by the multi-agent \`/code-review\` run on PR #1652 (#1594's \`debug\`/\`debug(msg)\` fix),
while verifying the new fix's generator-order correctness. \`succinctly jq\`'s CLI driver
evaluates one input's whole filter to completion, collecting every output into a \`Vec\`,
before writing *any* of them to stdout — so a filter that both produces stdout output and
triggers a stderr side effect (\`debug\`, \`stderr\`, \`halt_error\`) doesn't interleave the two
streams in real time the way real jq's lazy generator does, even with \`--unbuffered\`.

## Repro (confirmed live against jq 1.7.1 and the current build)

\`\`\`console
$ jq            --unbuffered -cn '1, debug, 2'      # combined via 2>&1 to show ordering
1
["DEBUG:",null]
null
2

$ succinctly jq --unbuffered -cn '1, debug, 2'
["DEBUG:",null]
1
null
2
\`\`\`

Real jq streams \`1\` to stdout before ever evaluating the \`debug\` branch (each comma branch
is a separate generator step). \`succinctly jq\` evaluates all three outputs — \`1\`, \`debug\`'s
\`null\`, and \`2\` — before writing any of them, so the (unbuffered) stderr write from \`debug\`
lands before the (batched) stdout write for \`1\`, even though \`1\` was logically produced
first.

## Root cause

\`succinctly jq\`'s per-input evaluation loop (\`src/bin/succinctly/jq_runner.rs\`,
\`evaluate_input\`-area) materializes a whole input's filter results into a \`Vec<OwnedValue>\`
before the \`for result in results { write_output(...) }\` loop that actually writes to
stdout runs. \`--unbuffered\` (\`jq_runner.rs\` ~3438-3439) only adds a \`flush()\` call per
\`write_output\`, which doesn't help — the writes themselves are already deferred until after
the whole input's evaluation finishes, regardless of buffering. Meanwhile, \`stderr\`/
\`halt_error\`/\`debug\`/\`debug(msg)\`'s writes (\`write_stderr\` in \`src/jq/eval.rs\`) fire
immediately, mid-evaluation, since they're plain unbuffered \`std::io::stderr()\` writes with
no equivalent batching.

## Why this is separate from #1594

[...]

PR #2102, opened 2026-09-01
test(jq): enumerate the #1653 M2 streaming parity gaps, close the terminal keys fault

newhoggy:
Progresses #1653. **Deliberately not a `Fixes`** — this is the prerequisite half.

## Why this PR exists

#1653's remaining work is the M2 lazy path: a JSON document read from a file or stdin, the
CLI's default route, still collects an input's whole output before writing any of it. The
streaming machinery has been public since #1892 (`eval_each_with_cursor`); the blocker is
that flipping the route also swaps evaluators, and the eager evaluator's materializing arms
double as a validity gate. Natively-streaming arms skip #1194/#1642 checks their eager twins
performed as a side effect of building an `OwnedValue`. An earlier attempt was reverted for
exactly that reason — but the divergences were never enumerated, only counted ("six tests").

This PR enumerates them, and closes the first one. No default-route behaviour changes.

## 1. `scripts/jq-m2-streaming-sweep.sh`

Separates the two variables that the reverted attempt had entangled. Runs each
(document, filter) case three ways — eager route, streaming route, pinned jq 1.7.1 — and
gates on **eager-vs-streaming parity**, reporting the oracle column only directionally
(a cell may move toward jq, never away). Both routes come out of one binary via
`SUCCINCTLY_JQ_M2_EVAL`, so the comparison cannot drift on a stale half.

Standing matches its sibling `scripts/jq-fanout-oracle-sweep.sh`: a verification tool, not a
CI gate. 352 cases today (11 documents x 32 filters).

This is also what #1994 asks for.

### What it found

Running it against a scratch flip — evaluator swapped, output ordering held constant, so
every failure is necessarily a parity failure — gives exactly 7 failing tests in four
classes, not one:

| Class | What | Tests |
|---|---|---|
| **A** | terminal #1194 check missing in the streaming `keys_unsorted[]` arm | `..._1629`, `..._1770` |
| **B** | lost `ambiguous`/#1194 validity gate on arms whose eager twin materializes | `..._1642`, `..._1812` |
| **C** | undecodable-key spelling changes (raw echo vs re-escaped) | `..._1642` |
| **D** | a prefix reaches stdout before a lazily-detected document fault | `..._1597`, `..._1677` |

Class A is fixed below. B and C turn out to share one root cause and are filed separately.
Class D is inherent to streaming — jq emits no prefix, since it rejects the document at parse
time — but it is the same shape #1770 already accepted for `limit(2; keys_unsorted[])`
(exit 5 *with* stdout `"a"`), so the sweep classifies it as known rather than new.

Per-filter, 21 of 32 filters have zero divergence, including `.`, `.a`, `.[]`, `.a, .b`,
`keys`, `length`, `to_entries`, `first(...)` and `debug`. All 11 divergent filters forward or
discard the **root** cursor — note `.a, .b` is clean while `.,.` is not, so it is the root
cursor, not comma.

## 2. Class A — the terminal #1194 fault

`each_lazy_keys_iterate_sink`'s `keys_unsorted` arm moved its `DistinctKeyCursors` into the
`map` closure handed to `drive_pipe_elements_generic`, so nobody could ask the terminal half
of the #1194 check — `ended_unpaired`/`delimiter_fault` — once the walk finished. Those two
have no per-key signal at all (`uncons_key` returns the same `None` either way), so only the
walk that ended can tell a genuine end from an unpaired one. That is why the per-key
`key_is_malformed` half #1770 added could not cover them.

Hold the iterator `by_ref` and ask after the drive returns, **but only on `Flow::Exhausted`**.
A truncating consumer returns `Flow::Stopped` without the walk ever reaching the tail, so
`first`/`limit` are still never charged for the whole-object probe they exist to avoid
(#1514/#1599) — keeping #1770's accepted divergence scoped to exactly what it was scoped to.
A consumer that ran to the end already paid for the walk, so asking costs it two bool reads.

This is a 0 -> 5 move toward jq, which rejects all three documents.

## Verification

- `cargo fmt --all -- --check`, both clippy legs, `cargo test` — green.
- `cargo test --features cli` over `jq_cli_tests`, `jq_golden_tests`, `cli_golden_tests`,
  `cli_characterization_tests`, `deep_nesting_valid_tests`: **1,234 + 24 + 14 + 4 + 3 passed,
  0 failed**.
- New arm is unreachable from the CLI's default route until the flip, so it is pinned by
  three direct `eval_each_with_cursor_using` unit tests rather than a CLI round trip that
  would exercise the other evaluator. Non-vacuous: the same three documents were measured at
  exit 0 on the streaming route before the fix.

## Follow-ups

- Classes B and C, one root cause: `to_owned_with_cursor` on the ambient value was doing two
  jobs — validating the document, and producing the re-escaped spelling. Filed separately.
- The flip itself, gated on the cursor-transparent filter set the sweep identifies.

### DIFF
--- a/scripts/jq-m2-streaming-sweep.sh
+++ b/scripts/jq-m2-streaming-sweep.sh
@@ -1,0 +1,221 @@
+#!/usr/bin/env bash
+#
+# Eager-vs-streaming parity sweep for the jq M2 lazy path (#1653).
+#
+# #1653 wants the M2 route -- a JSON document read from a file or stdin, the
+# CLI's default -- to write each output as the evaluator produces it, so
+# stdout and stderr interleave the way real jq's lazy generator does. PR
+# #1892 converted every other route; the M2 one still calls the *eager*
+# `eval_with_cursor`. Switching it to the demand-driven `eval_each_with_cursor`
+# is not just an ordering change: the eager evaluator's materializing arms
+# double as a validity gate, so arms that stream natively skip #1194/#1642
+# checks their eager twins performed as a side effect of building an
+# `OwnedValue`. An earlier attempt at the switch was reverted for exactly
+# that reason.
+#
+# This sweep separates the two variables. It runs each (document, filter)
+# case three ways -- the eager route, the streaming route, and the pinned jq
+# oracle -- and reports:
+#
+#   * PARITY divergences: eager vs streaming disagree on stdout/stderr/exit.
+#     **These are the gate.** Every one is a check the streaming route lost
+#     (or gained). Expected result: `0 unexpected`.
+#   * ORACLE rows: informational. Succinctly is a semi-index and detects a
+#     malformed document only opportunistically, where jq -- a full
+#     validating parser -- rejects every document below at parse time, exit
+#     5, whatever the filter. So most oracle columns read `jq=5 succ=0`
+#     already, on both routes, and closing those is not this issue's job.
+#     The rule that matters: **a cell must not move from 5 to 0.**
+#
+# Both routes are driven out of one binary via `SUCCINCTLY_JQ_M2_EVAL`, so a
+# single build measures both and the comparison cannot drift on a stale half.
+#
+# This is a verification tool, not a CI gate -- same standing as its sibling
+# scripts/jq-fanout-oracle-sweep.sh. The pinned rows in tests/jq_cli_tests.rs
+# are what CI enforces.
+#
+# Usage:
+#   cargo build --release --features cli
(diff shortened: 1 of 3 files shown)


newhoggy:
#2104 has been retargeted onto `main` and now contains this PR's three commits as well as the flip, so the two overlap. Either can merge first; the other then needs only a rebase. Keeping this one open as the smaller, separately-reviewable prerequisite — close it as subsumed if you would rather review the whole thing in one place.

Issue #2103, opened 2026-09-01
jq: eager and streaming evaluators disagree on document validation and undecodable-key spelling


newhoggy:
**Severity: Low** (no wrong output today — the affected route is not yet reachable from the CLI)

## Summary

Found by `scripts/jq-m2-streaming-sweep.sh` (added in PR #2102) while enumerating what blocks
#1653's remaining half. The eager cursor evaluator (`eval_single`) and the demand-driven one
(`eval_each_generic`) disagree on two things for any filter whose eager twin is the
materializing wildcard, and both disagreements have the same root cause.

`eval_single`'s wildcard materializes the **ambient input value** via `to_owned_with_cursor`
before bridging into `eval.rs`. That single call is doing two unrelated jobs:

1. **Validating the document** — `resolve_display_key`/`DisplayKeyGuard` raise `object key
   "..." is ambiguous` on colliding undecodable keys (#1642), and the #1194 structural checks
   ride along.
2. **Choosing a spelling** — an undecodable key comes back as a materialized `String`, so a
   literal `\` in the source doubles (`\ud800` -> `\\ud800`), unlike the byte-verbatim echo
   the lazy path produces.

`eval_each_generic` has native streaming arms for `Comma`/`Try`/`Label`/`As`/`FuncDef`/`Limit`
and friends (#1596), which do neither. So both jobs silently stop happening.

## Repro (against PR #2102's branch, which can drive both routes from one binary)

```console
$ printf '%s' '{"\ud800":1,"\ud800":2}' > /tmp/d.json

# Class B — the validity gate
$ SUCCINCTLY_JQ_M2_EVAL=eager  succinctly jq -c '.,.' /tmp/d.json
jq: error (at /tmp/d.json:0): object key "\ud800" is ambiguous: ...      # exit 5
$ SUCCINCTLY_JQ_M2_EVAL=stream succinctly jq -c '.,.' /tmp/d.json
{"\ud800":1,"\ud800":2}
{"\ud800":1,"\ud800":2}                                                  # exit 0

$ printf '%s' '{"\ud800":1,"b":2}' > /tmp/e.json
# Class C — the spelling
$ SUCCINCTLY_JQ_M2_EVAL=eager  succinctly jq -c '.,.' /tmp/e.json
{"\\ud800":1,"b":2}
$ SUCCINCTLY_JQ_M2_EVAL=stream succinctly jq -c '.,.' /tmp/e.json
{"\ud800":1,"b":2}
```

[...]

Issue #1596, opened 2026-08-24
jq: first(if/try/label/as/limit) leaks a discarded branch's side effect — no native eval_each_generic arm


newhoggy:
**Severity: Low** (stray side-effect writes; no wrong output, no document consumption — same class as [#1592](https://github.com/rust-works/succinctly/issues/1592))

Sibling of [#1592](https://github.com/rust-works/succinctly/issues/1592) (`Compare`), found while
reviewing [#1565](https://github.com/rust-works/succinctly/issues/1565)'s fix. #1565's acceptance
criteria allowed these five shapes to be *either* fixed or pinned; they were pinned, so this
issue tracks the fix.

## The gap

Same root cause as #1592. A bare `first(...)` is intercepted by `eval_generic.rs`'s native
routing before `eval.rs` is reached. [#1461](https://github.com/rust-works/succinctly/issues/1461)
gave `eval_each_generic` lazy arms for `Comma`/`Pipe`/`Paren` only, by its own stated scope —
`eval.rs`'s wider arm set (`If`/`Try`/`Label`/`As`/`Limit`/`Compare`/...) has no counterpart
there. So when one of those is `first`'s argument's *top-level* shape, evaluation falls through
to eager `eval_single` and a sibling branch's side effect fires even though `first` never needed
it.

#1592 covers `Compare`. This covers the other five.

## Repros

Captured from pinned jq 1.7.1 (`/usr/bin/jq`) and a `--release --features cli` build of this
branch, stdout and stderr separated. jq writes nothing to stderr in all five; stdout matches
exactly in all five.

| filter | jq stderr | succinctly stderr |
|---|---|---|
| `first(if true then (1, ("B"\|stderr)) else 9 end)` | *(none)* | `B` |
| `first(try (1, ("B"\|stderr)))` | *(none)* | `B` |
| `first(label $out \| (1, ("B"\|stderr)))` | *(none)* | `B` |
| `first((1,2) as $x \| ($x, ("B"\|stderr)))` | *(none)* | `BB` |
| `first(limit(2; (1, ("B"\|stderr))))` | *(none)* | `B` |

The `as` row is doubly eager: both `$x` bindings' body runs to completion.

All five are pinned as known gaps in `test_short_circuit_side_effect_leaks_820_932_987`
(`tests/jq_cli_tests.rs`), alongside the `Compare` row #1592 tracks.

## Checked for the `input`-consumption escalation — does not reproduce

[...]

PR #2104, opened 2026-09-01
fix(jq): stream the M2 lazy path so --unbuffered interleaves in real time

### LINKED ISSUE #1653: jq: --unbuffered doesn't achieve real-time stdout/stderr interleaving — one input's whole output batches before any of it writes

newhoggy:
Fixes #1653.

Retargeted onto `main`, so this PR is self-contained: four commits, the first three of which
are also #2102 (the parity sweep and the terminal-keys fix). Review #2102 first if you want
those separated — it is the same three commits and is independently green; this PR subsumes it,
so whichever merges first, the other needs only a rebase.

| commit | what |
|---|---|
| `8c19d17d` | `scripts/jq-m2-streaming-sweep.sh` — the eager-vs-shipped-vs-oracle parity sweep |
| `306eb0e7` | Gap 1: the terminal #1194 check when a `keys_unsorted[]` walk reaches the end |
| `a1c11e75` | its three `eval_each_with_cursor_using` unit tests |
| `6e2e3352` | **the flip** — everything described below |

## What changes

The M2 lazy path — a JSON document read from a file or stdin, the CLI's default — was the one
route PR #1892 left batching. It now writes each output the moment the evaluator produces it:

```console
$ echo '[1,2]' | jq            --unbuffered -c '.[]|debug'  2>&1   # ["DEBUG:",1] 1 ["DEBUG:",2] 2
$ echo '[1,2]' | succinctly jq --unbuffered -c '.[]|debug'  2>&1   # same, byte for byte
```

`evaluate_bytes_streaming` replaces the collect-then-write loop, and the write moves **inside**
#1793's `catch_unwind`, because a streaming filter panics from within the sink callback rather
than before the loop starts. That widens `AssertUnwindSafe` to `out`/`sink`/`had_output`/
`last_output`, which is sound because `assert_nesting_depth` panics unconditionally *before*
any `write_all`, so the writer is never mid-record when the stack unwinds. The panic arm's
duplicated halt check goes away with it — nothing now sits between it and the shared one.

## Why it is gated

Routing a filter through the demand-driven evaluator is not only an ordering change. The eager
`eval_single`'s wildcard materializes the ambient value via `to_owned_with_cursor`, and that
call is doing two more jobs: validating the document (#1194's structural checks, #1642's
colliding-display-key raise) and choosing an undecodable key's spelling.
`eval_each_generic`'s native arms (#1596) do neither. This is what sank the earlier attempt.

`expr_is_cursor_transparent` admits only filters that cannot emit the **root** cursor through
such an arm. The gate reaches further than that sounds: past a first stage that *descends*
(`.[]`, `.a`, `.[0]`) the ambient value is a proper descendant, so everything downstream
streams too — `.[] | (., stderr)`, `.[] | if … end`, `.users[] | .name` all interleave. What
still batches is root-forwarding shapes: `.,.`, `(., debug)`, `label $x | .`, `def f: .; f`,
`1+1`. #2103 owns removing that restriction.

The set is **measured, not derived**. `scripts/jq-m2-streaming-sweep.sh` (added in #2102) over
506 cases:

| leg | unexpected parity divergences | oracle regressions | known |
|---|---|---|---|
| shipped gate | **0** | **0** | 12 |
| `--forced-stream` | 33 | 11 | 27 |

The second row is what the gate is protecting against, and is now #2103's worklist — shrinking
it is what lets the gate widen. The sweep's second leg defaults to the *shipped* route for
exactly this reason, so `0 unexpected` means "the gate admits nothing that changes an answer".

## Three pinned expectations move

Each is toward jq, or a stated consequence of streaming:

- **`test_streamed_ordering_not_yet_reached_on_m2_path_1653`** was a tripwire for this change
  and says so in its own doc comment. Deleted; its case moves into
  `test_unbuffered_interleaves_stdout_and_stderr_1653` along with three more M2 rows captured
  live from jq 1.7.1 — `stderr` per element, an error mid-document, and `halt_error`.
- **`_1597` / `_1677`**: a fault found by *walking to it* leaves the prefix on stdout, where
  jq's strict parser rejects the document before any filter runs and emits nothing. Exit codes
  are unchanged. This is the same shape #1770 already accepted for `limit(2; keys_unsorted[])`
  (exit 5 with `"a"` already written); suppressing it would mean validating every document up
  front, which is the cost semi-indexing exists to avoid.
- **`_998`**: `-e`'s `materialize()` is now inside the guard, so the nesting panic is caught
  and reported as exit 5 instead of aborting the process at 101 — the outcome #1793's own
  rationale argues for ("an ordinary, recoverable diagnostic" rather than a hard exit). The
  limit is still enforced and the message is unchanged; only the manner of failing moved.

`docs/compliance/jq/limitations.md`'s interleaving section is rewritten accordingly: the
"every route but the M2 lazy path" framing is gone, replaced by the gate and the prefix
divergence.

## Verification

- `cargo fmt --all -- --check`; all three clippy legs; `cargo check --no-default-features`;
  `cargo test`, `cargo test --features simd` — green.
- CI's full `--features cli` test list: **1,234 jq CLI + 1,358 yq CLI + all others, 0 failed.**
- Sweep: `0 unexpected, 0 oracle regressions` (table above).
- Oracle differential against `/usr/bin/jq` 1.7.1 for `.[]|debug`, `.[] | (., stderr)`,
  `.[] | if .==2 then error("x") else . end`, `.[] | (., (2|halt_error))` — all byte-identical
  including exit codes.

### Outstanding: the A/B timing run

**Not yet run.** The M2 route is the hot default path, so it needs one, and this machine is not
idle (a concurrent worktree is running tests at ~291% CPU) — timing it here would be noise.
Note `scripts/perf-guard.py` cannot see this change either way: its six queries are `.` (the
identity fast path, untouched) and bare `keys_unsorted` (a `LazyKeys` item with no
`Expr::Iterate` stage), so green there is not evidence.

The change is expected to be neutral-to-positive — it removes a per-document `Vec`
accumulation without changing the evaluator's work — but that is a prediction, not a
measurement. Baseline binary and a 2 MB three-shape corpus are prepared for
`scripts/ab-cli.py`; **please hold merge until that lands in a comment here.**

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -1928,7 +1928,7 @@ underlying exit-code divergence properly needs `get_inputs` to be able to raise 
 fatal `EvalError` from this specific condition rather than only ever warning or returning
 values — not attempted here.
 
-## Real-time stdout/stderr interleaving: every route but the M2 lazy path
+## Real-time stdout/stderr interleaving
 
 Real jq is a lazy generator, so a filter that both writes to stdout and triggers a stderr
 side effect (`debug`, `stderr`, `halt_error`) or raises mid-stream interleaves the two as it
@@ -1937,19 +1937,12 @@ it, so every stderr write had already happened by the time the first stdout writ
 which no amount of buffering could fix, `--unbuffered`'s per-write `flush()` included
 ([#1653](https://github.com/rust-works/succinctly/issues/1653)).
 
-Fixed for every route except the M2 lazy path below — `-n`, `--slurp`, `--input-dsv`, the
-`input`/`inputs` bridge, and any flag that forces materialization (`-S`, `-a`, `-R`,
-`--seq`, `--args`) — all of which now write each output as the evaluator produces it
-(`evaluate_input_streaming`,
+Every route now writes each output as the evaluator produces it. `-n`, `--slurp`,
+`--input-dsv`, the `input`/`inputs` bridge and the materializing flags (`-S`, `-a`, `-R`,
+`--seq`, `--args`) go through `evaluate_input_streaming`; the M2 lazy path — a document read
+from a file or stdin, the default — goes through `evaluate_bytes_streaming`, both in
 [src/bin/succinctly/jq_runner.rs](../../../src/bin/succinctly/jq_runner.rs), over
-`eval_each_with_cursor`, [src/jq/eval_generic.rs](../../../src/jq/eval_generic.rs)):
-
-```
-$ jq            --unbuffered -cn '1, debug, 2'   2>&1     # 1 / ["DEBUG:",null] / null / 2
-$ succinctly jq --unbuffered -cn '1, debug, 2'   2>&1     # same
-```
-
-**Still batched: a document read from a file or stdin** — the M2 lazy path:
+`eval_each_with_cursor` in [src/jq/eval_generic.rs](../../../src/jq/eval_generic.rs).
 
 ```
 $ echo '[1,2]' | jq            --unbuffered -c '.[]|debug'   2>&1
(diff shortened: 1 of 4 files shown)


newhoggy:
## A/B timing — done, on both pinned boxes. No regression; the `debug` workloads get faster.

Method per [docs/guides/benchmarking.md § A/B Benchmarking Method](../blob/main/docs/guides/benchmarking.md)
via `scripts/ab-cli.py` (interleaved within each repetition), baseline `53edb2ebb` = this
branch's merge-base, both halves built `--release --features cli` on the measuring machine,
7 reps, 2 MB corpus in three shapes (`nested`, `sequences`, `users`) from
`succinctly json generate`. Control run first on each box to establish its own noise floor.

### Apple M4 Pro (`johns-mac-mini`)

Control: median **+0.55%**, range −2.6%..+4.2%.

| workload | med Δ |
|---|---|
| `.` / `.[]` / `.[] \| .name` / `.,.` (12 rows) | −8.6% .. +2.5%, all inside control |
| `nested .[] \| (., debug)` | **−5.9%** |
| `users .[] \| (., debug)` | **−12.5%** |
| `sequences .[] \| (., debug)` | **−13.5%** |

`median of medians: −0.02%`

### AMD Ryzen 9 7950X (`terminus`)

Control: median **+0.09%**, range −0.8%..+1.2% — a much tighter floor, so this is the box to
read.

| workload | med Δ |
|---|---|
| `.` / `.[]` / `.[] \| .name` / `.,.` (12 rows) | −2.6% .. +0.3%, all inside or better than control |
| `nested .[] \| (., debug)` | **−5.4%** |
| `sequences .[] \| (., debug)` | **−19.1%** |
| `users .[] \| (., debug)` | **−22.7%** |

`median of medians: −1.26%`, 2/15 rows slower (both +0.2%/+0.3%, inside control).

### Reading

The rows that had to be flat are flat on both architectures: `.` (identity fast path,
untouched), `.[]` and `.[] | .name` (the hot streamed shapes), and `.,.` (still batched, so
genuinely unchanged code). Nothing regresses beyond either box's control range.

The `(., debug)` rows are faster because streaming removes the per-document `Vec` that
accumulated every output before the first write — exactly the allocation this PR deletes, and
the workload where that Vec was largest. The effect scales with how much output the filter
produces, which is why `sequences`/`users` move ~20% and `nested` ~5%.

### Output identity

`ab-cli.py`'s own gate hashes **stdout and stderr merged**, so it flags the three
`(., debug)` configurations — which is this PR's entire point, not a defect. Identity was
therefore verified out-of-band, per stream, with `cmp` on separately-redirected files: all 15
(input, query) configurations are **byte-identical on stdout and byte-identical on stderr**;
only the interleaving of the two differs.

(Recorded for the next person: `cmd 2>&1 >/dev/null | ...` does **not** isolate stderr under
zsh's MULTIOS — it tees. That produced a phantom "stderr differs" until the comparison was
redone with plain file redirects.)

### `scripts/perf-guard.py`

Run and green, but that is not evidence either way here: its six queries are `.` (the identity
fast path, which this PR does not touch) and bare `keys_unsorted` (a `LazyKeys` item with no
`Expr::Iterate` stage, so it never reaches the changed code). Noted so a future reader does not
mistake its green for coverage of this change.

newhoggy:
Note on CI: `ci.yml` triggers only on `pull_request: branches: [main]`, and this PR targets #2102's branch, so GitHub runs nothing here but `Block Merge`. The full suite will run once #2102 merges and this PR is retargeted onto `main`.

In the meantime every CI leg was run locally against this branch and is green: `cargo fmt --all -- --check`; all three clippy legs (`--all-features` plus the two explicit feature sets); `cargo check --no-default-features`; `cargo test`; `cargo test --features simd`; and CI's full `--features cli` test list — **1,234 jq CLI + 1,358 yq CLI + 14 + 24 + 4 + 5 + 37 + 25 + 17 + 3 + 4 + 4, zero failures.`

`scripts/perf-guard.py --check --baseline-binary <merge-base build> --arch x86_64` was also run on the pinned 7950X (valgrind is Linux-only): all six queries within threshold, drift −0.0% to −0.1%.

PR #2105, opened 2026-09-01
refactor(jq): converge set_path onto a peel-and-recurse walker (#1429)

### LINKED ISSUE #1429: jq: set_path's flatten-then-split shape for Iterate/Slice duplicates update_path's simpler peel-and-recurse design
**Severity: Low** (design/maintainability, not a correctness bug)

## Summary

#1298 gave `set_path` (`=`'s own walker) a working fix for a mid-chain `Iterate`, but did so by adding a second flatten-then-scan-then-bail helper (`split_at_iterate`, mirroring the existing `split_at_slice`) ahead of `get_path_mut`'s one-mutable-slot loop, rather than fixing `get_path_mut`'s underlying one-slot design or restructuring `set_path` as a peel-one-component-and-recurse walker.

`update_path` (`|=`'s walker, same file) already proves the simpler shape works: its own `Expr::Pipe` arm's nested match has plain `Field`/`Index`/`Iterate`/`Slice` arms that peel the first path component and recurse on the rest — zero pre-scan, zero flatten pass, zero precedence logic between constructs. `set_path` could, in principle, grow the identical shape.

## What #1298 shipped instead, and the cost

- Two structurally near-identical flatten-then-split functions (`split_at_slice`, `split_at_iterate`) — same cheap guard shape, same `push_path_components` flatten loop, same `unwrap_path_component` scan, same `split_off`/`truncate` splitting, differing only in which `Expr` variant each looks for and what extra bail check it needs.
- An **unenforced ordering invariant**: `split_at_iterate` explicitly bails when an earlier `Slice` exists in the chain (so `split_at_slice` gets first crack at it), but `split_at_slice` has no reciprocal bail for an earlier `Iterate` — correctness of "leftmost construct wins" depends entirely on `set_path`'s `Pipe` arm always calling `split_at_iterate` before `split_at_slice`. That's enforced only by call-site ordering plus a doc comment, not by the type system. A third multi-slot construct added later (by the same precedent) would need bail-checks against *every* earlier-checked function, and every earlier-checked function would need a reciprocal bail against it if it can occur even earlier — an O(N²) hand-maintained precedence graph for N future constructs.
- Now **five** separately-maintained "walk every array element or object value, else no-op-or-error" loops exist across the file for `Expr::Iterate`: `set_path`'s terminal arm, the new `set_path_through_iterate`, `update_path`'s terminal arm, `update_path`'s mid-chain `Pipe` arm, and `delete_expr_iterate_paths`. No shared closure-based helper unifies them, despite #1298 explicitly modeling its two new functions on the `split_at_slice`/`through_slice` precedent pair — which is itself the existing shared abstraction for the analogous `Slice` case, but not reused for `Iterate`.
- A double-flatten cost on any path with a nested `Paren`/`Pipe` group but *no* `Iterate` and *no* `Slice` (e.g. `(.a|.c)[0] = 9`) — both guards fail, so both functions perform a full heap-allocate-and-clone flatten before either concludes "not my case." Pre-#1298 there was exactly one such flatten; post-#1298 there are two. Similarly, each level of a chained mid-chain `Iterate` (`.items[].containers[].env[] = v`, a realistic Kubernetes-manifest shape this codebase specifically targets/benchmarks) re-flattens an already-flat tail on every recursive `set_path` call.

None of this is a correctness bug — #1298's own extensive live verification (against both pinned jq 1.7.1 and yq v4.53.3, across ordering permutations, `?` placement, and comma-grouped paths) found no behavioral divergence from the chosen design. This is purely a maintainability/extensibility and minor-efficiency concern.

## Scope note

A real fix likely means converging `set_path`'s `Pipe` arm onto `update_path`'s peel-and-recurse shape — giving `set_path` (or a shared underlying walker both `set_path` and `update_path` delegate to) explicit `Field`/`Index`/`Iterate`/`Slice` arms in one match, rather than pre-scanning for special constructs ahead of a separate `get_path_mut` navigation pass. That's a substantial restructuring of `=`'s core assignment-evaluation path (`get_path_mut` is also called from `yq_del_slice_outcome`'s sibling logic and has its own `#1287`/`#1294` nested-`Pipe`-splicing behavior to preserve), not something to attempt under review-response time pressure — filing to make it trackable and give it a proper scoping pass, not blocking #1298's own narrower, correctness-verified fix.

Filed during #1298's code review (three independent review angles converged on the same underlying observation).

newhoggy:
Closes #1429.

## The problem

`set_path` (`=`'s walker) navigated a chained path by *pre-scanning* it for constructs that name more than one slot, then handing a prefix to `get_path_mut` — a loop that could only ever return a **single** mutable slot. Each such construct therefore needed its own flatten-then-scan-then-bail helper bolted in front of that loop: `split_at_slice`, then `split_at_iterate` (#1298).

#1429 (Low severity, design/maintainability, filed from #1298's own code review — no correctness bug) records four costs:

1. Two structurally near-identical split functions.
2. An **unenforced ordering invariant**. `split_at_iterate` bailed whenever a `Slice` appeared earlier in the chain so `split_at_slice` — called *after* it — got first crack; `split_at_slice` carried no reciprocal bail. "Leftmost construct wins" rested entirely on call-site ordering plus a doc comment, and an Nth construct would have turned it into a hand-maintained O(N²) precedence graph.
3. Duplicated "walk every array element / every object value" loops for `Expr::Iterate`.
4. A **double flatten** on any path with a nested `Paren`/`Pipe` but no iterate and no slice (`(.a|.c)[0] = 9` — both guards flattened in full before either concluded it had nothing to do), plus a re-flatten of an already-flat tail at every fan-out level of a chained mid-chain iterate (`.items[].containers[].env[] = v`).

## The change

`set_path`'s `Expr::Pipe` arm becomes a dispatcher into **`set_path_steps`**: flatten once, then peel one atomic component per call and recurse on the rest — the shape `update_path` and `delete_at_path` already use. Leftmost-wins becomes *structural*: whichever construct comes first is simply the first arm the walk reaches.

Deleted: `split_at_slice`, `split_at_iterate`, `SliceSplit`, `IterateSplit`, `set_path_through_iterate`, `get_path_mut`, `tail_writes_from_fresh_parent`.

All four costs go with them. The common flat path (`.a.b.c[0] = 9`) is now walked **borrowed** — strictly fewer allocations than before, since `get_path_mut` used to clone the component list plus a parallel `vec![false; len]` for that same case.

`for_each_container_slot` replaces three of the four duplicated `.[]` fan-out loop pairs. The fourth — `update_path`'s *terminal* `Expr::Iterate` arm — is deliberately left out and says why at the helper: its jq-mode #1916 branch rebuilds the container rather than mutating in place, so it splits on `Array`/`Object` before it splits on mode and has no in-place loop to share.

**#1428's stranded-autovivify guard changes mechanism, not behaviour.** It probed a detached `Null` before committing only because `get_path_mut`'s loop could not be walked back up. The frame that creates a slot is now still live when the recursion returns, so it undoes it directly, exactly as `update_path` does. Errors need no undo: `fork_rhs_over_paths` discards the whole partially-written document for that RHS output.

## Verification

**Primary gate — byte-identical output.** New `scripts/jq-assign-path-oracle-sweep.sh` records stdout/stderr/exit for every case against both succinctly and the pinned oracles, as a stable TSV two builds can be `diff`ed on. **20,496 cases** — 2 modes × 8 write operators × 60 curated path shapes × 16 documents, plus 3,600 deterministically generated paths — **byte-identical between the pre-#1429 binary and this branch**, re-run after every commit here.

The curated shapes cover exactly what the deleted helpers encoded: `?` at every position, `Paren`/`Pipe` groups including group-scoped `?`, iterate and slice at leading/mid/terminal position, **both orderings** (slice-before-iterate *and* iterate-before-slice), chained iterates, negative and out-of-range indices, missing parents (#1428), and scalar/container targets (#1181/#1232/#1101/#1142). The generated section composes atoms mechanically over a fixed seed, because a hand-picked matrix only covers shapes its author already thought of.

jq mode matches pinned jq 1.7.1 on every one of those cases. The 2,189 yq-mode divergences against pinned yq v4.53.3 are pre-existing and unchanged in count and content (e.g. real yq reads `.[0]` on an object as the string key `"0"`).

Also green: full suite (6,865 tests), all three CI clippy legs, `fmt`, `no_std`, rustdoc with `-D warnings`. **Patch coverage 100%** (310/310).

## Performance

Interleaved A/B per `docs/guides/benchmarking.md`, idle M4 Pro, `ab-cli.py`, 9 reps, output-identity gated, control run first (−0.4% median, −1.2%..+0.9% band). A 200k-document corpus so the *walk* rather than the parse dominates:

| workload | min Δ | median Δ |
|---|---:|---:|
| `(.a\|.b).c = 9` — the double-flatten case | **−8.0%** | **−9.1%** |
| `.a.b.c.d.e.f.g.h = 9` — long flat chain | −1.6% | −1.5% |
| `.a.b.c = 9` — short flat chain | −0.1% | −0.8% |
| `.a.b.c \|= 9` — untouched walker (control) | −0.3% | +0.2% |
| `del(.a.b.c)` — untouched walker (control) | +0.6% | +0.4% |

The −8%/−9% row reproduced across two runs. On a parse-dominated macro workload (1 MB / 4 MB k8s-shaped JSON) every row sits inside the control band except `.items[].containers[].env[].v = 9` at −1.6% min / −3.1% median — the chained mid-chain iterate that used to re-flatten its tail per level.

**x86 not measured**: the 7950X box needed a Tailscale re-auth this session could not perform. The change is allocation and branch structure rather than anything memory-bandwidth-bound, so it is not the class of result that usually fails to port — but it is unmeasured, and stated rather than implied.

## Three things reviewers should know

1. **The issue's scope note was wrong about `get_path_mut`.** It said `get_path_mut` was "also called from `yq_del_slice_outcome`'s sibling logic"; that is `navigate_read_only`, a separate read-only walker. `get_path_mut` had exactly one caller and could be deleted outright, so `del()` is untouched.

2. **A pre-existing depth asymmetry, surfaced by the deep-chain test.** In a *debug* build `set_path_steps` clears 256 path components while `update_path` — untouched here — already aborts with a stack overflow between 128 and 192, so the two are pinned at 256 and 128 rather than at a shared depth. The release CLI is bounded long before either: `=`, `\|=` and `del()` were run at 256 / 1,000 / 5,000 / 20,000 / 50,000 and 200,000 components against the binaries either side of this change, and all 36 results are identical — a clean `nesting depth exceeds limit of 384` from the pre-existing *value*-nesting guard, never a crash.

3. **`Expr::Identity` is now atomic, and that fixed a real mistake.** The first draft's `Identity` arm claimed to serve the borrowed fast path but `is_atomic_path_component` rejected `Identity`, so any chain containing one was flattened — which drops it — and the arm was dead. Accepting `Identity` makes `.a | . | .b = 9` stay on the borrowed path. The `[]` arm's own comment was likewise wrong about which spelling reaches it; the arm was instrumented and the answer is a nested group sitting *alongside* another component (`((.|.) | .) = 9`), not `((.|.)) = 9`, whose outer parens `set_path`'s `Paren` arm unwraps first. Both near-miss spellings are pinned alongside so the distinction stays visible.

No entry added to either `limitations.md` — there is no behaviour change. The existing prose there that named the deleted functions has been updated to keep the historical account while pointing at `set_path_steps`.

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -412,7 +412,9 @@ whatever container the path names. Four non-slice rows used to sit above these �
 on `null`, `.a.b = 1` on `{}`, `.[5] = 9` and `.[5] |= 9` on `[1,2]`, all reported as
 `Cannot index …`/`index N out of bounds (length M)` where jq builds or pads the container —
 closed together by [#486](https://github.com/rust-works/succinctly/issues/486)
-(`set_path`/`get_path_mut`/`update_path` now vivify `null` and pad past the array end, the
+(`set_path`/`update_path`, and `=`'s own chain walker — then `get_path_mut`, now
+`set_path_steps` since [#1429](https://github.com/rust-works/succinctly/issues/1429) —
+now vivify `null` and pad past the array end, the
 way `set_value_at_path` already did for `setpath()`) and
 [#498](https://github.com/rust-works/succinctly/issues/498) (the write-time negative-index
 bounds check now survives `?`, since padding is what makes the positive case succeed rather
@@ -485,7 +487,8 @@ the mechanism.
 `setpath` is the same operation without the syntax, and after #359 it does follow jq —
 `[1,2] | setpath([5]; 9)` is `[1,2,null,null,null,9]`, and `null | setpath(["a"]; 1)` is
 `{"a":1}`. The two used to disagree with each other in-tree; [#486](https://github.com/rust-works/succinctly/issues/486)
-closed that by teaching `set_path`/`get_path_mut`/`update_path` to vivify the way
+closed that by teaching `set_path`/`update_path` and `=`'s chain walker (then
+`get_path_mut`, now `set_path_steps`) to vivify the way
 `set_value_at_path` already did, so `=`/`|=`/the compound operators/`//=` now agree with
 `setpath` on every one of the four shapes removed from the table earlier in this section.
 `delete_at_path` deliberately keeps its own,
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -966,7 +966,9 @@ fix relies on).
 
 [#1232](https://github.com/rust-works/succinctly/issues/1232) widened #1181's scalar-target
 no-op to a scalar hit *before* the last path component (`.a.b = 99` on a scalar `.a`
-no-ops), but only for the *static*-path walkers (`get_path_mut`/`set_path`/`update_path`).
+no-ops), but only for the *static*-path walkers (`set_path`/`update_path` — at the time
+`=`'s own share of that lived in `get_path_mut`, folded into `set_path_steps` by
+[#1429](https://github.com/rust-works/succinctly/issues/1429)).
 A path that needs the full `resolve_dynamic_indexes` pre-pass — a computed key, or a
 `Comma`-grouped LHS — resolves each component through a plain read evaluator with no yq
 scalar-noop awareness at all, so the boundary this section's *previous* entry describes
@@ -994,7 +996,10 @@ silently, even though the write itself — `.[$k] = 99` — already correctly no
 under plain `=` (`.a[].b = 99`) always errored "invalid path component" — in both jq and yq
 mode, and predating #1232 entirely (`|=`/`+=`/`-=`/`*=`/`del()` were unaffected, their own
 recursive-descent walkers already supported fan-out). #1298 added `split_at_iterate`/
-`set_path_through_iterate` so `=` fans out per element like every other operator, and gave
+`set_path_through_iterate` so `=` fans out per element like every other operator (both since
+folded into `set_path_steps`' own mid-chain `Iterate` arm by
+[#1429](https://github.com/rust-works/succinctly/issues/1429), which replaced the whole
+pre-scan with one peel-and-recurse walk — same behaviour, verified byte-identical), and gave
 `navigate_read_only`'s prefix walk (`yq_assign_is_total_noop`'s own eager-RHS-discard
 pre-check, #1232's `PrefixNavOutcome`) an `Expr::Iterate` arm too — but only for the
 narrowest case, `.a` itself being a genuine scalar (`.a[].b = error("boom")` on `a: 5` now
@@ -1005,7 +1010,8 @@ RHS-discard rule for a mid-chain `Iterate` is broader than #1298's own narrowest
 container whose own elements *all* individually no-op also discards the RHS, and so does an
 empty container (vacuously). New `assign_path_all_noop` recurses into
 `.iter().all(...)`/`.values().all(...)` at such an `Iterate` instead of unconditionally
-deferring, a read-only dry run of `set_path_through_iterate`'s own per-element recursion:
+deferring, a read-only dry run of the `=` walker's own per-element recursion (then
+`set_path_through_iterate`, now `set_path_steps`' `Iterate` arm):
 
 ```bash
 $ printf 'a: [1, 2]\n' | yq            -o=json '.a[].b = error("boom")'   # {"a":[1,2]}, no-op
--- a/scripts/jq-assign-path-oracle-sweep.sh
+++ b/scripts/jq-assign-path-oracle-sweep.sh
@@ -1,0 +1,279 @@
+#!/usr/bin/env bash
+#
+# Oracle sweep for the assignment/delete *path walkers* (`set_path`,
+# `update_path`, `delete_at_path` in src/jq/eval.rs).
+#
+# Cross-products documents x path shapes x write operators x mode (jq/yq),
+# recording stdout/stderr/exit for the built succinctly binary alongside the
+# pinned oracle (jq 1.7.1, yq v4.53.3) for the identical case. Written for
+# #1429 (converging `set_path`'s flatten-then-split pre-scan onto a
+# peel-and-recurse walker), which is a pure restructuring: the primary gate
+# is not "matches the oracle" but "the *succinctly* columns are byte-identical
+# before and after", with the oracle columns present so a divergence can be
+# classified as pre-existing rather than re-triaged by hand.
+#
+# Two uses:
+#
+#   # A/B a restructuring (the primary gate):
+#   ./scripts/jq-assign-path-oracle-sweep.sh .ai/scratch/succinctly-pre > before.tsv
+#   ./scripts/jq-assign-path-oracle-sweep.sh target/release/succinctly > after.tsv
+#   diff before.tsv after.tsv && echo "byte-identical"
+#
+#   # Triage succinctly-vs-oracle divergence in one recording:
+#   ./scripts/jq-assign-path-oracle-sweep.sh --summary target/release/succinctly
+#
+# The record is a stable, sorted-by-construction TSV so `diff` is meaningful:
+#   mode <TAB> op <TAB> path <TAB> doc <TAB> succ_exit <TAB> succ_out <TAB> succ_err
+#        <TAB> oracle_exit <TAB> oracle_out <TAB> oracle_err
+# Newlines inside a field are escaped to \n so one case is always one line.
+#
+# `--no-oracle` skips the oracle columns (roughly halves the runtime) for a
+# fast A/B when divergence triage is not needed. The oracle columns are
+# identical between two recordings anyway, so a diff is unaffected either way
+# — but only recordings made with the *same* flag are comparable.
+#
+# Usage:
+#   cargo build --release --features cli,simd,regex,serde
+#   ./scripts/jq-assign-path-oracle-sweep.sh [--summary] [--no-oracle] [binary]
+
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+PIN="$(cat "$REPO_ROOT/tests/data/jq-golden/JQ_VERSION")"
+
+SUMMARY=0
+USE_ORACLE=1
+SUCC=""
+for arg in "$@"; do
+  case "$arg" in
+    --summary) SUMMARY=1 ;;
+    --no-oracle) USE_ORACLE=0 ;;
+    *) SUCC="$arg" ;;
+  esac
+done
+SUCC="${SUCC:-$REPO_ROOT/target/release/succinctly}"
+
+if [[ ! -x "$SUCC" ]]; then
+  echo "error: succinctly binary not found at $SUCC — run: cargo build --release --features cli,simd,regex,serde" >&2
+  exit 1
+fi
+
+# Prefer the pinned oracle binary directly (macOS ships jq-1.7.1 at
+# /usr/bin/jq; a PATH jq, e.g. Homebrew's, is often newer) — same reasoning as
+# scripts/jq-fanout-oracle-sweep.sh and tests/jq_cli_tests.rs's own comments.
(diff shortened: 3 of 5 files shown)


newhoggy:
## Review of #1429's `set_path_steps` refactor

Correctness looks solid: ~85 live differential probes against pinned `/usr/bin/jq` 1.7.1 and Homebrew `yq` v4.53.3 (construct ordering, nested `Paren`/`Optional` groups, #1428 stranding-undo in all its shapes, index-padding undo, yq-mode autovivify) all matched, and the full suite (3223 tests) passes. The issue's core ask — removing the unenforced call-order invariant between the old `split_at_slice`/`split_at_iterate` pre-scans — is achieved cleanly.

Three findings from a deeper cleanup/efficiency pass, all confirmed via direct code reading plus empirical A/B measurement:

**1. O(N²) regression for `=` on long flat chains** — [`src/jq/eval.rs:18234`](../blob/HEAD/src/jq/eval.rs#L18234)

```rust
let rest_reaches_iterate = rest.iter().any(reaches_iterate);
```

is recomputed from scratch at every recursion frame of `set_path_steps`. For a chain with no `Iterate` anywhere, `.any()` never short-circuits, so this is `N(N-1)/2` scans total across the walk.

Measured (A/B, current HEAD vs. pre-#1429 `cb6763cb5`) for `null | .k0.k1.k2...kN = 9`:

| N | new (this branch) | old (pre-#1429) |
|---------|--------------------|-------------------|
| 32,000  | 0.57s | 0.02s |
| 64,000  | 2.28s | 0.04s |
| 128,000 | 9.19s | 0.06s |

Clean ~4x-per-doubling quadratic scaling on the new code vs. flat/linear on the old. Extrapolated to N=200,000 (the depth this PR's own new test `test_assign_deep_chain_recurses_without_blowing_the_stack_1429` exercises), this is roughly 25-60s of pure CPU where the old walker did the equivalent check once. A single O(N) reverse pass building a suffix `Vec<bool>` before the walk starts would restore linear behavior.

**2. New stack-overflow exposure in debug builds** — [`src/jq/eval.rs:18146`](../blob/HEAD/src/jq/eval.rs#L18146)

`set_path_steps` replaces the old iterative `get_path_mut` loop with per-component recursion, so a long flat `=` chain now consumes one Rust stack frame per path component (same shape `update_path`/`del()` already had) — where the old loop consumed none.

Per this PR's own test comments, a debug build (2 MiB test-thread stack) already stack-overflows around 128-192 components for the pre-existing recursive walkers, and `set_path_steps` now shares that shape for `=` too. Release builds are shielded by the existing depth-384 value-nesting guard, but a debug build processing a sufficiently deep static `=` chain — previously safe, since `=` used to loop rather than recurse — can now crash with an uncontrolled stack overflow instead of a clean error.

**3. Stranded-write-undo logic duplicated a 4th time** — [`src/jq/eval.rs:18242-18313`](../blob/HEAD/src/jq/eval.rs#L18242-L18313) vs. [`19112-19238`](../blob/HEAD/src/jq/eval.rs#L19112-L19238)

The #1428 "autovivify → navigate → recurse → undo if slot is still `Null`" pattern is now hand-duplicated a 4th time — 2 pre-existing copies in `update_path`'s `Field`/`Index` arms, 2 new copies in `set_path_steps`'s `Field`/`Index` arms — differing only in container mechanics (`map.shift_remove`+`is_empty()` vs. `arr.truncate`+`is_empty()`).

This is exactly the failure mode the PR's own new `for_each_container_slot` doc comment (line ~17901) names as the reason that helper was factored out ("there were four hand-maintained copies of this loop pair before... they had already drifted once"). The stranded-undo condition has already changed once historically per the surrounding comments; a future change again requires updating 4 call sites with no compiler-enforced link between them — inconsistent with this PR's own stated rationale for eliminating that class of duplication elsewhere in the same diff.

---

Findings 1 and 2 are regressions introduced by this refactor (not pre-existing), so worth addressing before merge. Finding 3 is a maintainability nit that could reasonably be filed as a follow-up instead.

newhoggy:
All three findings taken. 1 and 2 reproduced locally first, and 2 turned out to be worse than described.

**1. O(N²) — confirmed and fixed.** Reproduced exactly: 0.15s / 0.58s / 2.40s at N = 16,000 / 32,000 / 64,000 against a flat 0.02s for `cb6763cb5`. My A/B never saw it because my deepest benchmark chain was 8 components.

Fixed by hoisting rather than by a suffix `Vec<bool>` — no allocation needed. Every recursive call gets a *suffix* of the same list, so one number answers all of them: `iterate_suffix_len` returns the shortest suffix still containing an `Iterate` (`usize::MAX` when there is none), and `rest` contains one exactly when `rest.len() >= iterate_suffix`. O(N) once at the dispatcher, O(1) per frame.

**2. Stack overflow — confirmed, and it is not debug-only.** This is the part worth flagging: at 500,000 components a **release** build aborts with SIGABRT where `cb6763cb5` returns a clean `nesting depth exceeds limit of 384` (exit 5). The release build is not shielded by the depth-384 guard, because that guard is on *value* nesting and fires at output time — after the walk has already recursed. So this was a live crash-for-clean-error regression, not a debug-build nuisance.

Fixed structurally rather than with a new depth limit. A frame exists in `set_path_steps` for exactly one reason: so the slot it created is still reachable if the tail turns out to have written nothing and it has to be undone (#1428). A step with no `Iterate` left ahead of it can never strand — that *is* the entire undo condition — so its frame has nothing to come back to. The walker now consumes such steps in place and keeps the recursion only for shapes that need it, bounding frame count by the stranding-capable steps rather than by path length. Finding 1's O(1) test is what makes knowing that free.

Now identical to `cb6763cb5` at every depth measured — clean exit 5 at 200,000 / 500,000 / **1,000,000** components, and matching timings (0.01 / 0.04 / 0.06s at 32,000 / 128,000 / 200,000).

**3. Stranded-undo duplication — taken, not deferred.** You quoted this PR's own `for_each_container_slot` rationale back at it, which is fair. `slot_was_stranded` is now the single definition of #1428's rule across all four call sites. The container mechanics stay per-arm on purpose: `map.shift_remove` and `arr.truncate` are genuinely different operations, and it is the *decision* that has drifted historically, not the removal.

**Tests added for the two regressions**, since neither existing coverage nor my own A/B caught them:

- `test_iterate_suffix_len_matches_a_direct_scan_1429` pins the O(1) answer against the `rest.iter().any(reaches_iterate)` definition it replaced, over every suffix of a chain mixing both component kinds and both `Iterate` spellings.
- `test_deep_flat_assign_chain_exits_cleanly_not_stack_overflow_1429` (tests/jq_cli_tests.rs) drives 200,000 components end-to-end and asserts a refusal rather than a crash. Passed via `-f`: at ~2 MB the filter is well past `ARG_MAX` on Linux CI as argv, even though macOS would take it. It asserts no wall-clock bound — that would be flaky — but a return of the quadratic makes it take minutes rather than well under a second.

The deep-chain unit test keeps its measured 256/128 split for `=`/`|=` and no longer claims `=` recurses per component.

**Re-verified after all three:** byte-identical to the pre-#1429 binary over all 20,496 sweep cases, 6,867 tests, all three clippy legs, fmt, no_std, rustdoc.

One thing the fix does *not* address, and I would rather name than leave implied: `|=` and `del()` still recurse per component unconditionally, and at 500,000 components both are killed (OOM) or time out — on `cb6763cb5` as well as here. That is pre-existing and out of scope for this PR, but it is the same shape, and worth a follow-up now that `=` no longer shares it.

newhoggy:
Filed the follow-up named at the end of the previous comment: #2115 — `|=` and `del()` still recurse once per path component, with the measurements against `cb6763cb5` showing it is pre-existing rather than anything this PR introduced.

Issue #2049, opened 2026-08-31
yq: del() with a multi-branch computed key on a scalar hits a live unreachable!() and crashes the process


newhoggy:
**Severity: High** (process crash from ordinary CLI input — the #1098 class)

Found by a randomised `del()` differential fuzzer while working #1690 (PR #2047).
Pre-existing on `main`; PR #2047 reproduces it faithfully rather than silently
changing it, so it moves function but not behaviour.

## Summary

In **yq mode only**, `del()` with a *computed* key that resolves to more than one
branch, against a **scalar** document, reaches the comma-grouped delete walker with a
non-object root and panics on an `unreachable!()`. That is a hard process abort
(exit 101, `internal error: entered unreachable code`), not an error message.

## Repro

```console
$ echo '2.5' | succinctly yq 'del(.[("k0","k1")])'

thread 'main' panicked at src/jq/eval.rs:31821:9:
internal error: entered unreachable code: delete_expr_object_paths reached a non-object, non-null root
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
$ echo $?
101
```

Real yq no-ops:

```console
$ echo '2.5' | yq 'del(.[("k0","k1")])'
2.5
```

Same for every scalar target: `2.5`, `0`, `"a"`, `true`, `false`.
`del(.[("k0",0)])` (mixed key kinds) panics identically.

## Scope

Narrow, and the boundary is informative:

| filter | `2.5` in `succinctly yq` | real yq |
|---|---|---|
| `del(.k0)` | `Error: Cannot index number with string "k0"` | `2.5` |
| `del(.k0, .k1)` (static comma) | `Error: Cannot index number with string "k0"` | `2.5` |
| `del(.[("k0")])` (computed, one branch) | `Error: Cannot index number with string "k0"` | `2.5` |
| `del(.[("k0","k1")])` (computed, 2+ branches) | **panic, exit 101** | `2.5` |

`succinctly jq` is unaffected and correct throughout
(`Cannot index number with string "k0"`, matching `/usr/bin/jq` 1.7.1).

## Root cause

`src/jq/eval.rs`. For a *static* path, `resolve_node` validates the target and raises
`Cannot index number with string` before the walker ever runs. For a multi-branch
*computed* key in yq mode it does not, so the resolved branches reach the multi-path
delete walker against a scalar root, where the object arm's

```rust
let OwnedValue::Object(entries) = &mut value else {
[...]

PR #2112, opened 2026-09-01
fix(jq): a pruned ?-guarded comma sibling no longer errors on = / |=

### LINKED ISSUE #2124: jq: a ?-guarded comma branch is not pruned when it has a sibling — wrong = / |= output
**Severity: Medium** (silently wrong write, not just a wrong read)

## Summary

A `?`-guarded comma branch is correctly pruned from the resolved path set when it
is the *only* branch, but kept (and its underlying navigation error raised) once
it has a sibling. This was originally found and diagnosed on #2049 (as a `del()`
crash), and #2049's own merged fix (PR #2108) plus an independent fix already on
`main` happen to also correct `del()` and `path()` for this shape — but `=`/`|=`
still hit it.

## Repro (live-verified against `main` tip, after #2049 closed)

```console
$ echo '{"c":1}' | jq -c '(.b, .c.a?) = 9'
{"c":1,"b":9}

$ echo '{"c":1}' | succinctly jq -c '(.b, .c.a?) = 9'
jq: error (at <stdin>:1): Cannot index number with string "a"

$ echo '{"c":1}' | succinctly jq -c '(.b, .c.a?) |= 9'
jq: error (at <stdin>:1): Cannot index number with string "a"
```

`{"c":"s"}`, `{"c":true}`, `{"c":[]}` as the root all reproduce the same way
(confirmed for the sibling `del()` shape on #2049; not yet re-confirmed for `=`/`|=`
specifically but expected to match given the shared root cause).

The already-fixed rows, for reference (both now match jq 1.7.1 on current `main`):

```console
$ echo '{"c":1}' | succinctly jq -c '[path(.b, .c.a?)]'
[["b"]]
$ echo '{"c":1}' | succinctly jq -c 'del(.b, .c.a?)'
{"c":1}
```

## Root cause (from #2049's review comments, not yet re-verified against current
`main`'s exact code shape — re-confirm before implementing)

A `?`-guarded step's genuinely empty output (a failed navigation) gets coerced into
a resolved `null`/error branch instead of pruning the whole branch, once a sibling
branch is present in the same comma group. `path(f)` prunes correctly because jq
computes it in one pass with no sibling entanglement; `=`/`|=`'s path-resolution
machinery apparently doesn't yet share whatever fix made `path()`/`del()` correct.

## Suggested fix direction

Per the original diagnosis on #2049: the general fix is pruning the failed branch
during resolution (before `=`/`|=`'s walker ever sees it), not swapping the error
for a different error — jq prints the document with only the surviving branch's
write applied, not an error at all. Worth checking first whether whatever already
fixed `path()`/`del()` on `main` (unclear which commit — worth a quick `git log`/
bisect) can be extended to `=`/`|=`'s resolution path directly, rather than
re-deriving the fix from scratch.

## Related

- #2049 (closed) — where this was originally found, scoped down to just the
  `del()` crash for that fix.
- PR #2112 (`issue-2049-del-optional-comma-panic`) — appears to already implement
  the general resolution-level pruning fix (per its own description, fixes
  `path()`/`=`/`del()` together), but was authored against #2049 before it closed
  and is currently blocked on a CI webhook registration problem. Worth checking
  whether it already covers this issue's `=`/`|=` scope before re-implementing.

newhoggy:
## Summary

**Retargeted from #2049 to #2124** — #2049 closed while this PR was in
flight, via an independent, narrower fix (#2108) merged from a
differently-named branch claiming the same issue (a duplicate-claim
collision the atomic branch-name lock doesn't catch across different
branch names -- see the issue's own comment thread). #2108's fix covers
del()'s crash correctly, and path() was already correct via unrelated
code. This PR's actual remaining scope, tracked as #2124, is =/|=.

resolve_seq's no-computed-key fast path (value_after_components) coerced a
?-suppressed step's genuinely-empty output into OwnedValue::Null instead of
pruning the whole branch. Every component reaching that loop is proven
single-valued by construction (needs_fanout_pass), so zero outputs can only
mean a ?-suppressed navigation failure -- the fix propagates that as "prune
this branch" (Option<OwnedValue>) instead of fabricating a resolved =/|=
write at a path the document doesn't actually have that shape at.

Two call sites needed the fix: resolve_seq's fully-static fast path, and
apply_static_tail's per-branch tail application (reached once a real
dynamic/computed element precedes the ?-guarded step in the same pipe).

## Test plan

- [x] cargo build --features cli
- [x] cargo test --features cli,simd,regex,serde (full suite, 0 failures)
- [x] cargo clippy --all-targets --all-features -- -D warnings
- [x] cargo fmt --check
- [x] All repros verified against the pinned oracle (/usr/bin/jq 1.7.1):
      =/|= for the comma-sibling shape, argument-order independence, a
      scalar-shape sweep, the dynamic-fanout-then-static-tail variant, and a
      regression check that path()/del() stay correct
- [x] New regression tests in tests/jq_cli_tests.rs covering all of the
      above

Fixes #2124

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -19623,13 +19623,13 @@ fn needs_path_prepass(expr: &Expr) -> bool {
 /// in the pipe, or -- unconditionally -- `resolve_recurse` resolving `f`):
 /// can its "nothing left needs resolving" fast path assume the remaining
 /// tail is single-valued? For `Iterate` the answer is no --
-/// `value_after_components` silently collapses any component producing != 1
-/// output to `Null`, which `resolve_recurse` treats as "no children" and
-/// prunes, silently dropping every result whenever a trailing
-/// `.foo[]`/`.foo[]?` reaches a 2+-element container (#682). Fanning it out
-/// through the same loop already used for a genuine computed key fixes that,
-/// without touching `resolve_dynamic_indexes`'s much more common
-/// no-computed-key fast path.
+/// [`value_after_components`] treats any component producing zero or more
+/// than one output as pruning the whole branch (`Ok(None)`, #2124), which
+/// `resolve_recurse` reads as "no children" and drops, silently discarding
+/// every result whenever a trailing `.foo[]`/`.foo[]?` reaches a 2+-element
+/// container (#682). Fanning it out through the same loop already used for
+/// a genuine computed key fixes that, without touching
+/// `resolve_dynamic_indexes`'s much more common no-computed-key fast path.
 ///
 /// No `Expr::Pipe` arm: this function's only caller (`resolve_seq`) always
 /// calls it on `flat`'s already-`push_path_components`-flattened elements,
@@ -24347,7 +24347,7 @@ fn resolve_slice_bound<S: EvalSemantics>(
 fn value_after_components<S: EvalSemantics>(
     components: &[Expr],
     value: &OwnedValue,
-) -> Result<OwnedValue, EvalEscape> {
+) -> Result<Option<OwnedValue>, EvalEscape> {
     let mut current = value.clone();
     for component in components {
         let mut values = eval_owned_multi::<S>(component, &current)?;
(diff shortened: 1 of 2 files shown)


newhoggy:
Reopening to re-trigger CI — no workflow run registered after 20+ minutes despite a successful push (other branches pushed after this one already have runs).

PR #2113, opened 2026-09-01
test(jq): fix cargo test --no-default-features build+run gaps (#2083)

### LINKED ISSUE #2083: test: cargo test --no-default-features fails to build (two unrelated std-gating gaps)
**Severity: Low**

## Summary

`cargo test --no-default-features` fails to build on `main` (confirmed at `18997e9bb`), independently in two unrelated places. Neither is reachable through this project's own documented local-verification commands (`cargo test`, `cargo test --features cli`, `cargo test --features large-tests`), which is presumably why this has gone unnoticed — CLAUDE.md's own "Common Commands" section never exercises `--no-default-features`.

## Repro

```console
$ cargo test --no-default-features --no-run
error[E0432]: unresolved import `succinctly::jq::seed_remaining_inputs`
  --> tests/jq_eval_using_input_queue_gate_test.rs:...
   |
   = note: found an item that was configured out
  --> src/jq/mod.rs:99:50
   |
97 | #[cfg(feature = "std")]
   |       --------------- the item is gated behind the `std` feature
98 | pub use eval::{
99 |     current_input_location, pop_remaining_input, seed_remaining_inputs, UNKNOWN_INPUT_LINE,
   |                                                  ^^^^^^^^^^^^^^^^^^^^^
error: could not compile `succinctly` (test "jq_eval_using_input_queue_gate_test") due to 1 previous error

error[E0425]: cannot find function `parse_simple_tz_offset` in this scope
  --> src/jq/eval.rs:59433:...  (5 occurrences, all in `#[cfg(test)] mod tests`)
error: could not compile `succinctly` (lib test) due to 5 previous errors
```

## Root cause

1. `tests/jq_eval_using_input_queue_gate_test.rs` unconditionally imports `succinctly::jq::seed_remaining_inputs`, which `src/jq/mod.rs:97` re-exports only `#[cfg(feature = "std")]`. The integration test has no matching `#[cfg]` gate of its own.
2. `src/jq/eval.rs:34735`'s `parse_simple_tz_offset` is itself `#[cfg(feature = "std")]`-gated (it's TZ-env-var parsing, calling `std::env::var`), but the `#[cfg(test)] mod tests` block that calls it (around line 59433, testing `parse_numeric_offset`) carries no corresponding `std`-only gate.

Both are the same shape: a `std`-gated item with a test/integration-test consumer that isn't gated to match.

## Why this is separate from #2079

Discovered while running the full local-verification matrix (`cargo test --no-default-features`, per this repo's own "local checks missing locally" guidance) for an unrelated `reduce`/`foreach` fix. Confirmed unaffected by that change — reproduces identically on unmodified `main` at `18997e9bb`.

## Suggested fix direction

Gate `tests/jq_eval_using_input_queue_gate_test.rs`'s body (or just its `seed_remaining_inputs`-using tests) behind `#[cfg(feature = "std")]`, and either gate the `parse_simple_tz_offset`-calling tests in `src/jq/eval.rs` behind the same feature, or move `parse_numeric_offset`'s own tests to not depend on the `std`-only helper.

newhoggy:
## Summary
- `tests/jq_eval_using_input_queue_gate_test.rs` unconditionally imported `succinctly::jq::seed_remaining_inputs`, which is `#[cfg(feature = "std")]`-only in `src/jq/mod.rs` — gated the whole test file behind `#![cfg(feature = "std")]`.
- `src/jq/eval.rs`'s `test_parse_simple_tz_offset_errors_gracefully_on_overflow_894` called the `std`-only `parse_simple_tz_offset` with no matching gate — added `#[cfg(feature = "std")]`.
- Once those two build-blocking gaps were fixed, the full `--no-default-features` suite still failed at *runtime* on the same underlying shape: several `env`/`$ENV`/`now` tests assumed std-only results (a real value) unconditionally, when the no_std fallback intentionally returns `Null`/an `EvalError`/`0.0`. Split each into an `std` arm (existing assertion) and a `not(std)` arm (asserts the documented no_std fallback).
- Two more tests (`eval.rs`'s `test_bind_def_seeds_defcall_frames_from_ambient_depth_1371`/`test_ambient_frame_depth_composes_with_defcall_guard_1371`, and `eval_generic.rs`'s `eval_each_generic_reports_defcall_binding_failure_1371`) exercise the `ambient_frame_depth` mechanism, which is a documented no-op under `no_std` — gated all three behind `#[cfg(feature = "std")]` since there's no no_std variant of that mechanism to test.

Fixes #2083.

## Test plan
- [x] `cargo test --no-default-features --no-run` builds clean (previously failed with 2 unrelated compile errors)
- [x] `cargo test --no-default-features` passes fully (previously also failed at runtime on 10 more tests once compilation succeeded)
- [x] `cargo test --features cli,simd,regex,serde` passes fully (56/56 test groups ok)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [x] `cargo clippy --all-targets --features std,simd,serde,cli,regex,bench-runner,large-tests,mmap-tests -- -D warnings` clean
- [x] Test-only change; no new production code paths, so no coverage gap introduced

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -41458,7 +41458,12 @@ mod tests {
     /// many thousands of levels -- prohibitively slow in a debug build --
     /// to actually reach `MAX_EVAL_FRAMES`; this white-box check reaches
     /// the same boundary in one call.
+    ///
+    /// `#[cfg(feature = "std")]`: `ambient_frame_depth` itself is a no-op
+    /// under `no_std` (see its own module doc comment) -- there is no
+    /// no_std variant of this mechanism to test.
     #[test]
+    #[cfg(feature = "std")]
     fn test_bind_def_seeds_defcall_frames_from_ambient_depth_1371() {
         let Expr::FuncDef {
             name,
@@ -41507,7 +41512,11 @@ mod tests {
     /// refused immediately once the ambient floor alone already meets the
     /// ceiling -- this is the guard actually firing, not just the frame
     /// count being recorded correctly (the test above).
+    ///
+    /// `#[cfg(feature = "std")]`: same reason as the test above -- the
+    /// ambient-depth mechanism this exercises doesn't exist under `no_std`.
     #[test]
+    #[cfg(feature = "std")]
     fn test_ambient_frame_depth_composes_with_defcall_guard_1371() {
         let Expr::FuncDef {
             name,
@@ -58044,11 +58053,15 @@ mod tests {
 
     #[test]
     fn test_dollar_env_field_access() {
-        // $ENV.VAR should return the environment variable value
+        // $ENV.VAR should return the environment variable value (std), or
+        // null in no_std context ($ENV is an empty object there, same as
+        // test_dollar_env's own std/no_std split).
+        #[cfg(feature = "std")]
         query!(b"null", "$ENV.PATH", QueryResult::Owned(OwnedValue::String(s)) => {
-            #[cfg(feature = "std")]
             assert!(!s.is_empty(), "PATH should be non-empty");
         });
+        #[cfg(not(feature = "std"))]
+        query!(b"null", "$ENV.PATH", QueryResult::Owned(OwnedValue::Null) => {});
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


newhoggy:
Code review (8 finders) found the no_std CI leg only runs `cargo check`, never `cargo test`/`--tests` — so this whole no_std test surface (including this PR's own fix) is currently unverified by CI. Filed as #2114 (workflow-file change, separate risk profile from this narrow test-only PR).

Issue #2119, opened 2026-09-01
test: cargo test --no-default-features builds (after #2083) but 10 tests fail on genuine std-only assumptions


newhoggy:
**Severity: Low**

## Summary

#2083 fixed two compile-time gaps that made `cargo test --no-default-features` fail
to build at all. With that fixed, `cargo test --no-default-features` (full run, not
just `--no-run`) now builds and runs, but 10 tests fail — all because they assert
`std`-only behavior without accounting for the already-implemented, deliberate
`no_std` fallback paths.

## Repro (on top of #2083's fix)

```console
$ cargo test --no-default-features
test result: FAILED. 3147 passed; 10 failed; 1 ignored
```

Failing tests:
- `jq::eval::tests::test_now` — asserts `now` returns a real Unix timestamp;
  `builtin_now`'s `#[cfg(not(feature = "std"))]` arm returns `0.0` by design
  (`src/jq/eval.rs`, `builtin_now`).
- `jq::eval::tests::test_env_var`, `test_strenv`, `test_env_field_access`,
  `test_env_bracket_access` — all assert `env(PATH)`/`strenv(PATH)`/`env.PATH`/
  `env["PATH"]` return `QueryResult::Owned(OwnedValue::String(_))`; the
  `#[cfg(not(feature = "std"))]` arms of `builtin_env_object`/`builtin_strenv`
  return `QueryResult::Error(...)` by design when not `optional`.
- `jq::eval::tests::test_dollar_env_field_access`, `test_dollar_env_bracket_access`
  — assert `$ENV.PATH`/`$ENV["PATH"]` return a `String`; `builtin_envvar`'s
  `#[cfg(not(feature = "std"))]` arm returns `QueryResult::Owned(OwnedValue::Null)`
  by design.
- `jq::eval::tests::test_bind_def_seeds_defcall_frames_from_ambient_depth_1371`,
  `test_ambient_frame_depth_composes_with_defcall_guard_1371`,
  `jq::eval_generic::tests::eval_each_generic_reports_defcall_binding_failure_1371`
  — all assert the `ambient_frame_depth` module tracks real depth; its
  `#[cfg(not(feature = "std"))]` arm is an intentional no-op (`get()` always
  returns `0`, `enter()` a no-op) per its own doc comment ("This leaves `no_std`
  exactly as exposed to the gap above as it already was, not more").

## Why this is separate from #2083

#2083's own repro and scope were specifically the *build* failure
[...]

PR #2120, opened 2026-09-01
test(jq): gate two std-only test sites so --no-default-features builds

### LINKED ISSUE #2083: test: cargo test --no-default-features fails to build (two unrelated std-gating gaps)

newhoggy:
## Summary
- `tests/jq_eval_using_input_queue_gate_test.rs` unconditionally imports `succinctly::jq::seed_remaining_inputs`, which `src/jq/mod.rs` only re-exports under `#[cfg(feature = "std")]` — added a matching `#![cfg(feature = "std")]` to the whole file (its only purpose is exercising that function).
- `test_parse_simple_tz_offset_errors_gracefully_on_overflow_894` (`src/jq/eval.rs`) calls `parse_simple_tz_offset`, itself `#[cfg(feature = "std")]`-gated (TZ env var parsing), from an ungated `#[cfg(test)] mod tests` block — added a matching `#[cfg(feature = "std")]` on the test.
- Neither gap is reachable through this project's own documented local-verification commands (`cargo test`, `cargo test --features cli`, `cargo test --features large-tests`), which never exercise `--no-default-features` — confirmed both reproduce identically on unmodified `main`.
- Filed follow-up #2119: fixing these two build-breaking gaps exposes 10 pre-existing test failures once `cargo test --no-default-features` can actually run to completion — all genuine `std`-only assumptions in tests exercising already-`#[cfg(not(feature = "std"))]`-gated fallback code (`now`, `env`/`$ENV`/`strenv`, `ambient_frame_depth`), not new bugs, and out of scope for this issue's own repro (`--no-run`).

Fixes #2083

## Test plan
- [x] `cargo test --no-default-features --no-run` now succeeds (was: 2 compile errors)
- [x] `cargo llvm-cov --features cli,simd,regex,serde --workspace --summary-only` — no regression

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -59891,6 +59891,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(feature = "std")]
     fn test_parse_simple_tz_offset_errors_gracefully_on_overflow_894() {
         // #894: `hours * 3600`/`minutes * 60`/the final sign multiply were
         // unchecked, panicking on a malformed/adversarial `TZ` env var
--- a/tests/jq_eval_using_input_queue_gate_test.rs
+++ b/tests/jq_eval_using_input_queue_gate_test.rs
@@ -11,6 +11,11 @@
 //! thread's life -- kept to its own test binary (this file compiles to a
 //! separate process) so it can't leak into any other test file's shared
 //! process or test-thread pool.
+//!
+//! `seed_remaining_inputs` itself is `#[cfg(feature = "std")]` (`src/jq/mod.rs`),
+//! so this whole file compiles to nothing under `--no-default-features` (#2083).
+
+#![cfg(feature = "std")]
 
 use succinctly::jq::eval_generic::eval_using;
 use succinctly::jq::{parse, seed_remaining_inputs, JqSemantics, OwnedValue};


newhoggy:
Closing as superseded — issue #2083 was independently fixed and merged via #2113
(branch \`issue-2083-no-default-features-build\`, a differently-named branch for the
same issue that evaded this repo's atomic branch-name lock) while this PR was in
review. #2113 covers everything here plus more: it also gated the ambient-depth/env/
now tests this PR's own follow-up #2119 tracked, and added
\`test_bind_def_call_guard_fires_without_ambient_depth_2083\` to keep no_std coverage
of \`bind_def_call\`'s guard from silently dropping to zero. Independently verified
#2113's diff is a superset of this one's intent before closing.

Issue #1521, opened 2026-08-23
jq: last(f) on an empty stream yields nothing; jq yields null


newhoggy:
Found by a 1155-combination differential sweep run while finishing #1309.
**Pre-existing and unrelated to the input builtins** — it reproduces with
`last(empty)`, no `input`/`inputs` involved.

## Repro (jq 1.7.1)

```
$ jq -cn 'last(empty)'
null
$ succinctly jq -cn 'last(empty)'
              # no output, exit 0

$ jq -cn '[] | last(.[])'
null
$ succinctly jq -cn '[] | last(.[])'
              # no output, exit 0
```

## Cause

jq defines `last` as a fold seeded with `null`:

```
def last(f): reduce f as $x (null; $x);
```

so an empty stream yields the seed. succinctly's `eval_last_expr`
(`src/jq/eval.rs`) and `eval_first_or_last_generic` (`src/jq/eval_generic.rs`)
both map an empty inner stream to `QueryResult::None` / `GenericResult::None`,
i.e. no output at all.

`first` is **not** affected and must not be "fixed" to match — `first(empty)`
is genuinely empty in jq too, because `def first(f): label $out | (f, break $out);`
has nothing to yield. Confirmed live:

```
$ jq -cn 'first(empty)'        # (no output)
$ succinctly jq -cn 'first(empty)'  # (no output)   ✅ already matches
```

So this is an asymmetry in jq's own definitions, not an inconsistency to
normalise — ADR-0018 rule 3 applies.

## Scale

13 of the 34 divergences the sweep found were this one shape, reached via
`last(inputs)` on an exhausted queue across many flag combinations. It will
show up anywhere a `last(...)` operand can be empty.

**Severity: Low-Medium.** Silent wrong answer (missing output rather than a
`null`), but only for an empty operand.

## Suggested fix

Return `null` rather than `None` for an empty inner stream in both evaluators.
Check `builtin_last_stream` (`src/jq/eval.rs`) too — it has the same shape as
`eval_last_expr`. Worth a sweep for other jq builtins defined as `reduce`-with-
a-seed, where the same "empty stream should yield the seed" reasoning applies.

## Refs

- Found via #1309 / #1508's differential sweep
- `src/jq/eval.rs` — `eval_last_expr`, `builtin_last_stream`
- `src/jq/eval_generic.rs` — `eval_first_or_last_generic`

PR #2128, opened 2026-09-01
fix(jq): give first(f)/last(f) a path-context arm (#2074)

### LINKED ISSUE #2074: jq: `first`/`last` have no path-context arm, so `first(key)` is null
Found while fixing #1409 (PR #2071), and out of that issue's scope — this is not
the isolate-then-continue shape, it's a missing arm.

## Symptom

`first(f)`/`last(f)` silently stub path context to `null`:

```console
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a | first(key)'
null
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a | last(key)'
null

# the same query through any sibling construct is correct:
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a | limit(1; key)'
"a"
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a | key'
"a"
```

And the #1409 shape reproduces through them too, unfixed by #2071:

```console
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a | first(.[]) | key'
"a"
# expected 0 — `jq 'path(.a | first(.[]))'` gives ["a",0]
```

## Root cause

`Expr::FirstExpr`/`Expr::LastExpr` appear in neither place:

- `needs_path_context` has no arm for them, so a `key`/`parent`/`file_index`
  hidden inside `first(...)` doesn't route the pipe to the path-context
  evaluator at all;
- `eval_pipe_with_path_context_internal` has no arm for them either, so even
  when routed there by a sibling stage they fall into the generic `_` fallback,
  which resets path context.

Both halves are needed, the same pairing #715's `Label` fix and #1306's
`FuncDef` fix each required.

## Note on implementation

Within this eager evaluator `first(f)` looks equivalent to `Limit { n: 1, expr: f }`
— `Expr::Limit`'s arm already evaluates the body fully and truncates, and its
`Partial` handling already drops a trailing control once enough outputs arrived,
which is what makes `first(1, error("x"))` come out as `1` rather than erroring.
That equivalence is worth verifying against jq's laziness (and against anything
with side effects, e.g. `input`) before leaning on it, rather than assumed.

newhoggy:
## Summary

`Expr::FirstExpr`/`Expr::LastExpr` had no arm in `needs_path_context` or
`eval_pipe_with_path_context_internal`, so a `key`/`parent`/`file_index`
inside or after `first(f)`/`last(f)` silently stubbed to `null` where the
sibling `limit(1; key)` already resolved correctly:

```console
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a | first(key)'
null            # was; should be "a"
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a | first(.[]) | key'
"a"             # was; jq 'path(.a | first(.[]))' confirms it should be 0
```

- `first(f)` mirrors `Expr::Limit { n: 1, expr: f }`'s existing arm exactly (eager evaluation, take-first, same accepted #820 short-circuit tradeoff for a `Partial` that's already satisfied).
- `last(f)` mirrors `eval_last_expr`'s own semantics (empty stream → `null` per #1521; `Partial` always drops its prefix and surfaces the control since `last` cannot short-circuit — verified live `last(1,2,error("x"))` raises rather than answering `2`), restricted to the narrower `QueryResult` variant set `eval_pipe_with_path_context_internal` ever produces.
- Both are guarded the same way `Limit`'s arm is (`needs_path_context(expr) || rest.iter().any(needs_path_context)`), and both use the existing `path_probe_stage`/`continue_rest_with_paths` machinery for the "`rest` continues from the taken output's own path" shape, so `.a | first(.[]) | key` routes `key` through the correct per-output path rather than the ambient one.

Fixes #2074.

## Test plan
- [x] Live-verified against the repro cases in the issue, plus `path()` cross-checks against the pinned jq 1.7.1 oracle (`key` itself isn't a real jq builtin, so the path-index values are cross-checked via `path(.a | first(.[]))` etc.)
- [x] New regression test `test_first_last_expr_path_context_resolves_2074` in `tests/jq_cli_tests.rs`, modeled on the existing `test_limit_path_context_resolves_1765`
- [x] Edge-case regression check against the oracle: `first(empty)`, `last(empty)`, `first(1,2,error("x"))`, plain `first(.[])`/`last(.[])` — all unchanged, matching jq 1.7.1
- [x] `cargo llvm-cov --features cli,simd,regex,serde --workspace --summary-only` clean

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -31341,9 +31341,7 @@ fn eval_pipe_with_path_context_internal<'a, W: Clone + AsRef<[u64]>, S: EvalSema
                 QueryResult::Error(e) => QueryResult::Error(e),
                 QueryResult::Break(label) => QueryResult::Break(label),
                 QueryResult::Halt(code) => QueryResult::Halt(code),
-                QueryResult::Partial(_, Control::Error(e)) => QueryResult::Error(e),
-                QueryResult::Partial(_, Control::Break(label)) => QueryResult::Break(label),
-                QueryResult::Partial(_, Control::Halt(code)) => QueryResult::Halt(code),
+                QueryResult::Partial(_, control) => control_to_result(control),
                 QueryResult::One(_) | QueryResult::OneCursor(_) | QueryResult::Many(_) => {
                     unreachable!(
                         "eval_pipe_with_path_context_internal only ever produces \
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -807,13 +807,12 @@ pub(crate) fn needs_path_context(expr: &Expr) -> bool {
         Expr::Pipe(exprs) => exprs.iter().any(needs_path_context),
         Expr::Paren(inner) => needs_path_context(inner),
         Expr::Optional(inner) => needs_path_context(inner),
-        // No `Expr::FirstExpr`/`Expr::LastExpr` arm, and no matching arm in
-        // `eval_pipe_with_path_context_internal` either, so `first(key)` and
-        // `last(key)` silently answer `null` where the sibling
-        // `limit(1; key)` answers correctly (#2074). Both halves are needed
-        // to close it -- the same pairing `Label` (#715) and `FuncDef`
-        // (#1306) each required -- so it is tracked rather than half-fixed
-        // here.
+        // `first(expr)`/`last(expr)` (#2074): same reasoning as `Limit`
+        // above -- whatever the wrapped expression needs, the wrapper
+        // needs too. Paired with matching arms in
+        // `eval_pipe_with_path_context_internal` below, the same pairing
+        // `Label` (#715) and `FuncDef` (#1306) each required.
+        Expr::FirstExpr(inner) | Expr::LastExpr(inner) => needs_path_context(inner),
         Expr::Comma(exprs) => exprs.iter().any(needs_path_context),
         // `[key]`/`[parent]`/`[file_index]` (#1302): an array literal is
         // still just "collect this inner expression's outputs", the same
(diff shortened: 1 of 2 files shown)


newhoggy:
Code review found a pre-existing side-effect-leak tradeoff (shared with the existing `Limit` arm, not introduced by this PR) in the eager path-context evaluation these arms use — filed as #2129.

newhoggy:
Patch coverage is 95.35% (82/86); the 4 remaining uncovered lines are all provably unreachable, not undertested:

- `src/jq/eval.rs:31284` (`first`'s `Partial` arm, `vs.is_empty()` branch) — `partial()`'s own definition (line ~1959) already collapses an empty-prefix `Partial` into a bare terminal `QueryResult::Error/Break/Halt` before it's ever constructed, so `body_result` can never actually be `Partial(vs: [], _)`. Same shape as the pre-existing `Limit` arm's own identical vestigial branch (also unreachable there, already shipped on `main`).
- `31288`/`31346` — the `One|OneCursor|Many => unreachable!()` defensive-completeness arms both new arms carry, matching 11 other identical instances already in this file per code review.
- `31338` (`last`'s `ManyOwned` arm, `next_back() == None` branch) — `owned_vec_to_result` (what every producer of `eval_pipe_with_path_context_internal`'s result goes through) never constructs a zero-length `ManyOwned`: 0 elements collapses to `None`, 1 to `Owned`, so `ManyOwned` is only ever 2+.

All four are structurally untestable without corrupting this evaluator's own invariants, not lines review missed.

PR #2140, opened 2026-09-01
fix(jq): remove #2077's alpha-rename machinery, dead after #1371's Shared opacity

### LINKED ISSUE #2088: jq: alpha-renaming in #2077's substitute_func_param may not cover bare (non-$) function-style parameter capture
**Severity: Low** — internal gap confirmed by code inspection; no observable divergence against jq 1.7.1 found despite real effort by two independent investigators, so this is a lead for deeper investigation, not a confirmed live bug.

## Summary

#2077's fix (PR #2081) made `substitute_func_param`'s binder arms (`As`, `Reduce`, `Foreach`, `AsPattern`, nested `FuncDef` params) alpha-rename a binder before substituting an argument into its scope, when the binder's name collides with a free `$name` reference in the argument being substituted (`expr_free_vars`/`capturing_pattern_renames` in `src/jq/eval.rs`).

That free-variable collector only matches `Expr::Var(name)` (a `$name` reference). It never matches `Expr::FuncCall{name, args: []}` — a *bare* (non-`$`) filter-style parameter reference, which is how a call-by-name jq function parameter is referenced (e.g. `def f(g): g;` — `g` inside the body is a zero-arg `FuncCall`, not a `Var`). #2077's own "Scope note" explicitly widens Case A's binding-form surface to "a nested `def`'s own `$`-parameters" — but by the same reasoning, a nested `def`'s own **bare** (non-`$`) parameter of the same name as a bare-call argument being substituted through should be just as capturable, and isn't covered.

## What's confirmed vs. not

**Confirmed by code inspection**: `expr_free_vars` (`src/jq/eval.rs`) only inserts names from `Expr::Var` nodes, never from `Expr::FuncCall{args: [], ..}`. A code-review sub-agent independently reproduced this via temporary debug instrumentation: for a construction where a nested `def`'s own bare parameter name matches a bare-call argument, `capturing_pattern_renames` returns an empty rename list where (by the same logic that already applies to `$`-style args) it should return a rename.

**Not confirmed — no observable divergence found**: Both that sub-agent (~20 constructed programs) and a follow-up attempt in this session (2 more constructions, see below) failed to produce an actual wrong value against `/usr/bin/jq` 1.7.1. Constructions tried and their (matching) results:

```
$ jq -n 'def outer(y): def f(g): def y: 42; g; f(y); outer(5)'          # 5
$ succinctly jq -n 'def outer(y): def f(g): def y: 42; g; f(y); outer(5)'   # 5 -- matches

$ jq -n 'def X: 5; def f(g): def h(X): g; h(1); f(X)'                   # 5
$ succinctly jq -n 'def X: 5; def f(g): def h(X): g; h(1); f(X)'            # 5 -- matches
```

The sub-agent's working theory is that observable divergence is masked by a separate, pre-existing, unrelated gap in `expand_func_calls`'s own `Expr::FuncDef` handling (not `substitute_func_param`'s) — it doesn't check whether a nested `def`'s own parameter list shadows the name of the `def` currently being inlined, and that gap coincidentally "self-heals" the missing rename in every construction tried so far. That's a hypothesis, not a verified mechanism.

## Why this is separate from #2077

#2077's own issue text and repro table are entirely about `$`-variable (`Expr::Var`) capture. The bare-parameter-name form is a related but distinct capture surface that #2077's fix doesn't claim to cover (its "Scope note" mentions nested `$`-parameters specifically, not bare ones). Confirming or fixing it needs real investigation into `expand_func_calls`'s own traversal order, not just a mechanical extension of #2077's rename logic — and per the masking theory above, may be entangled with a second, independent gap in that function.

## Suggested next step

Not a fix — a spike: either construct a genuine end-to-end divergence (bypassing whatever masks it in the constructions above) to confirm this is live, or determine the masking is structural (i.e. `expand_func_calls`'s expansion order makes this class of capture always unobservable in practice) and close as not-a-bug. Both outcomes are informative. If it is live, the fix is likely a straightforward extension of `expr_free_vars` to also collect `Expr::FuncCall{name, args: []}` names, paired with `capturing_pattern_renames`'s renaming already being name-generic (not `$`-specific) — but confirm the repro before implementing.

This is architecturally related to #1371 (static substitution instead of runtime calls with real environments) — a full runtime-closure fix there would make this question moot the same way it would #2077's.

### LINKED ISSUE #2096: jq: #2077 alpha-renaming machinery may be dead code after #1371's Shared opacity
## Summary

Follow-up to #1371's runtime-call rewrite, found during code review of that branch.

ADR-0020 states that `Expr::Shared`'s opacity "gives hygiene for free ... makes [alpha-renaming] unnecessary for arguments." But the `#2077` alpha-renaming mechanism — `capturing_pattern_renames`/`apply_scope_renames`/`fresh_var_name` (`src/jq/eval.rs`, ~150 lines) and its call sites in `substitute_func_param`'s `As`/`Reduce`/`Foreach`/`AsPattern` arms — is still fully wired in and untested against that specific claim.

## Evidence it may be dead

A reviewing agent temporarily stubbed `capturing_pattern_renames` to always return no renames, rebuilt, and reran:
- all 11 of the file's own dedicated `#2077` regression tests, and
- the exact repro from the code's own comments: `def f(g): 5 as $x | g; 2 as $x | f($x)`

All passed unchanged. This is consistent with the theory that `substitute_var_impl`'s `DefCall`-args recursion (which walks before a value ever reaches `Shared`) already resolves every free variable eagerly, so no unresolved `Var` from an outer binder can ever reach a point where the callee's own binder could capture it — making the alpha-rename path unreachable in practice.

## What's not yet established

This is **plausible, not proven** — no exhaustive search was done for a construction where an unresolved `Var` survives to `Shared`-wrap time and could still trigger real capture. The current regression suite doesn't distinguish "alpha-renaming is doing real work" from "Shared's opacity alone already prevents the capture" — both explanations pass the same 11 tests.

## Ask

Either:
1. Find the counterexample that proves alpha-renaming still does real work post-#1371, and add a test that fails when `capturing_pattern_renames` is stubbed out (closing the ambiguity), or
2. Confirm it's fully dead and remove the ~150-line mechanism, with a comment on `Expr::Shared` explaining why hygiene is now free.

Either outcome is a correctness/maintainability improvement — right now this is untested, load-bearing-or-not, unclear which.

## Context

Found during code review of #1371's branch (issue-1371).

newhoggy:
## Summary

Fixes #2096.

#2077's alpha-rename mechanism (`capturing_pattern_renames`/`apply_scope_renames`/`apply_pattern_renames`/`fresh_var_name`/`rename_pattern_var`, called from `substitute_func_param`'s `As`/`Reduce`/`Foreach`/`AsPattern`/`FuncDef` binder arms) protected against a substituted-in function-parameter argument getting recaptured by a same-named binder inside the callee. #2096 asked whether that's still necessary after #1371's `Expr::Shared`-based substitution model, or whether `Shared`'s own opacity already makes it dead.

It's dead. Every real call to `substitute_func_param` (`bind_def_call`) wraps `arg` in `Expr::Shared` before substituting, and `substitute_func_param`'s own `Expr::Shared` arm never descends into it — so no *later* substitution into a binder's scope can ever reach back into an already-substituted `arg`, regardless of what name collides. `substitute_var_impl` (the `$name` substitution pass, separate from this one) resolves every `$name` still in lexical scope — recursing into `DefCall` args and nested `FuncDef` bodies the same way — *before* any downstream call's own argument is ever `Shared`-wrapped, so an unresolved `$name` can never survive to reach that wrap point in the first place.

## Verification

- Stubbed `capturing_pattern_renames` to always return no renames: all 3231 pre-existing lib tests (including all 11 dedicated `#2077` regression tests) still pass unchanged.
- 16 further adversarial constructions — covering every one of the five binder arms, `$`-parameter shorthand, nested/recursive `def`s, `label`/`break`, generators, and double indirection through a wrapper function — all still match `/usr/bin/jq` 1.7.1 exactly with the stub in place.
- Re-ran the same 16 constructions against the real fix (machinery fully removed, not just stubbed): still all match.
- Kept the 11 existing `#2077` tests as plain regression tests (they still pin real jq-compatible no-capture behavior, now enforced structurally) and added one new test for the double-indirection-through-a-wrapper shape, the strongest stress test of the `Shared`-opacity argument.

Also closes #2088 (a related but narrower spike into the *bare*, non-`$` parameter-capture shape of this same question) as not-a-bug, reaching the same `Expr::Shared`-opacity conclusion independently before this issue was picked up.

## Test plan

- [x] `cargo test --lib --features cli,regex` (3232 passed)
- [x] `cargo test --features cli,regex,serde` (full suite, all binaries green)
- [x] `cargo test --no-default-features --verbose` (no_std doctests)
- [x] Adversarial differential testing against pinned `/usr/bin/jq` 1.7.1 (see above)

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -299,7 +299,7 @@ use crate::json::light::{JsonCursor, JsonElements, JsonFields, StandardJson};
 
 use super::expr::{
     ArithOp, AssignOp, BoundBody, Builtin, CompareOp, Expr, FormatType, FuncDefData, Literal,
-    MergeFlags, NumberKey, ObjectEntry, ObjectKey, Pattern, PatternEntry, StringPart,
+    MergeFlags, NumberKey, ObjectEntry, ObjectKey, Pattern, StringPart,
 };
 use super::value::{
     assert_value_tree_depth, cmp_f64, infinite_float_preview_text, is_infinity_sentinel,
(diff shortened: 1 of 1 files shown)


newhoggy:
Ran the mandatory 9-angle `/code-review` pass. Summary of findings and responses:

- **No correctness issues.** 6 of 9 angles found nothing; the other 3 independently confirmed the dead-code determination via their own differential testing against `/usr/bin/jq` 1.7.1 (one even noted the flagship #2077 repro never actually exercised a nonempty `free_vars(arg)` at the old rename check — it was a no-op on its own test case).
- **Fixed:** three angles (B, and two findings from I+J) converged on the same point — the "`arg` is always `Expr::Shared`-wrapped" invariant the new doc comment states was enforced only by convention (one real call site), and one pre-existing test call site already violated it harmlessly. Added a `debug_assert!` for the invariant and fixed that test call site to satisfy it.
- **Fixed:** two angles (F, G) independently flagged the same repeated "binder shadows `param` → clone, else recurse" idiom across all five binder arms. Extracted `subst_unless_shadowed`/`subst_unless_shadowed_opt` to collapse it.
- **Fixed:** one angle (I+J) noted only 1 of the PR description's 16 adversarial constructions was captured as a permanent test. Added two more (label/break interaction; nested `$`-param under a forking generator) as regression tests.
- **Not applied:** a stronger type-level fix (a `SharedArg` newtype instead of the debug_assert) — the same finding explicitly noted the 12 existing regression tests already behaviorally cover an accidental rewrite of the one hand-maintained match arm this would protect, so the debug_assert is a reasonable middle ground without introducing a new wrapper type for a single private function's one real call site.
- **Filed #2141:** a genuine but unrelated pre-existing bug one angle's differential testing surfaced (bare parameter names and `$`-variable names share one shadow-check namespace, unlike real jq) — present identically before and after this PR, out of scope here.

All local gates green after the review-response commit: full lib + integration suite (3234 lib tests + all integration binaries), both required clippy feature combinations, `--no-default-features` no_std check, 100% patch coverage.

PR #2156, opened 2026-09-02
fix(jq): raise decode failures in item_to_owned's fanout-argument sinks (#2023)

### LINKED ISSUE #2023: jq: item_to_owned uses bare to_owned in fanout_arg's lazy sink
## Summary

Found during classification for #1989. `item_to_owned` (`src/jq/eval.rs:2449`)
uses bare `to_owned`, not `to_owned_checked`. Reached via `fanout_arg`'s
lazy `eval_each` sink (line 2236) and `fanout_two_args_lazy` (lines
2474, 2476) -- i.e. a generator-argument builtin's argument value,
materialized directly from document navigation.

(Its other 3 callers, `eval_range`'s from/to/step at lines 4030/4052/4065,
are safe -- the decoded value there is only classified Int/Float-vs-error,
content never read -- so this is scoped to the fanout callers only.)

## Repro to verify

Try `<doc> | test((.badfield, "b"))` (forces the multi-output/lazy
fan-out path) with `.badfield` an invalid-UTF8 string. Regex/generator-
argument builtins are common, so this looks CLI-reachable -- verify live
against the pinned oracle and the built binary before implementing.

## Suggested next step

Not investigated beyond the classification pass. Whoever picks this up
should confirm the repro reproduces, then route the fanout call sites
through `to_owned_checked`/`suppress_or_raise` while leaving the
`eval_range` call sites (safe) untouched -- may need a `_checked`
sibling rather than converting `item_to_owned` itself, since some
callers are safe.

Refs #1989

newhoggy:
## Summary

`item_to_owned` (`fanout_arg`'s/`fanout_two_args_lazy`'s lazy generator-argument sink — used by `test`/`match`/`capture`/`scan`/`ltrimstr`/`has`/`setpath`/every other `ArgFanout::All` builtin) silently substituted `""` for an undecodable `Item::Borrowed` string instead of raising: the same #1746/#1660 bug shape, in the one fanout-argument call-site family that fix's own sweep didn't reach.

`eval_range`'s three `item_to_owned` callers were separately confirmed safe per the issue's own note (the decoded value there is only classified Int/Float-vs-error, content never read) — left untouched.

## Investigation

The issue's own suggested repro (`test((.badfield, "b"))`) didn't reproduce — it's a type mismatch unrelated to decode handling (`test/1`'s single-arg form matches the *whole ambient* value against the regex, not `.badfield`; both real jq and pre-fix succinctly already error identically on that shape).

Constructing a real repro needed two things the issue didn't specify:
1. **Document-navigated invalid UTF-8, not a `$var`-bound value.** A `$var.field` access materializes to `Item::Owned` before ever reaching `item_to_owned`'s vulnerable `Item::Borrowed` arm (confirmed via debug tracing) — the argument expression has to be a direct, unbound navigation (`.field`, not `$doc.field`).
2. **A builtin whose own ambient value doesn't itself need a string-decode check.** `test`'s ambient must already decode as a string to match against at all, so a `test(...)`-based repro with the ambient set to invalid UTF-8 hits `test`'s own *pre-existing*, already-correct ambient check first, masking whether this fix's own code path does anything. `has(.b)`/`setpath(path; value)` — whose ambient is the whole document, never string-decode-checked — isolate the fanout-argument's own decode failure cleanly.

Verified this distinction empirically: reverted the fix temporarily (kept the new tests) and confirmed `has(.b)`/`setpath` fail to raise pre-fix (`Owned(Bool(false))` / a masked "Path must be specified as an array" from the corrupted `""` path), then restored the fix and confirmed both raise the correct decode failure.

## Fix

Added `item_to_owned_checked` (the `to_owned_checked`-backed twin of `item_to_owned`) and routed both `fanout_arg`'s `ArgFanout::All` closure and `fanout_two_args_lazy`'s outer/inner closures through it, propagating a decode failure the same way an ordinary `body` escape already does in each function — flush any buffered first result, then stop the pull with `Control::Error(e)` as the trailing control (matching `push_owned_values`'s own established `Some(Control::Error(e))` shape).

Fixes #2023.

## Test plan
- [x] `test_fanout_arg_item_to_owned_raises_decode_failure_2023` (`has(.b)`) — confirmed fails pre-fix, passes post-fix
- [x] `test_fanout_two_args_lazy_item_to_owned_raises_decode_failure_2023` (`setpath`, both outer/value and inner/path generators) — confirmed fails pre-fix, passes post-fix
- [x] `cargo llvm-cov --features cli,simd,regex,serde --workspace --summary-only` clean

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -45688,31 +45688,69 @@ mod tests {
         );
     }
 
+    /// #2023 coverage: the test above only ever sees a decode failure on
+    /// the *first* generator value, never exercising the `pending_first`
+    /// flush this arm has to do first when a *later* value fails instead
+    /// (the exact shape `body`'s own escape arm just below already flushes
+    /// for). `has((.a, .b))` on `{"a":"x","b":"<bad>"}` puts a real key
+    /// first (`.a`, buffered as `pending_first`) and the undecodable one
+    /// second (`.b`) -- `.a`'s own successful `has("x")` (`false`, no such
+    /// field) is flushed into the output prefix before the trailing decode
+    /// failure, matching jq's own partial-output-then-error generator
+    /// semantics (the same shape `first(1, error("x"))` = `1` elsewhere in
+    /// this file relies on).
+    #[test]
+    fn test_fanout_arg_item_to_owned_flushes_pending_first_before_raising_2023() {
+        query!(
+            b"{\"a\":\"x\",\"b\":\"\xff\xfe\"}",
+            r"has((.a, .b))",
+            QueryResult::Partial(vs, Control::Error(e)) if e.is_decode_failure() => {
+                assert_eq!(vs, vec![OwnedValue::Bool(false)]);
+                assert!(e.message.contains("invalid UTF-8"), "message: {}", e.message);
+            }
+        );
+    }
+
     /// #2023 code review: `fanout_two_args_lazy` (used by two-generator
     /// builtins, e.g. `setpath(path; value)`) has its own, independent pair
     /// of `item_to_owned` call sites -- outer and inner -- needing the
     /// identical fix.
     ///
-    /// Only the *inner* (path) slot is pinned here: an "outer" (value) case
-    /// (`setpath(["x"]; .b)`) is *not* a valid test for `setpath`
-    /// specifically -- review found `builtin_setpath`'s own body (#1953)
-    /// separately runs `to_owned_checked` over the *whole* ambient document
-    /// before ever writing, so it already raises the same decode failure on
-    /// `.b` regardless of whether this fix's own outer-slot code path does
-    /// anything, passing identically with the fix reverted. `setpath`'s
-    /// ambient is the whole document being modified, so there is no
-    /// `setpath`-based construction that isolates the outer slot the way
-    /// the inner one below is isolated. The outer and inner closures are
-    /// otherwise structurally identical (same match arms, same
-    /// escape-then-stop shape, differing only in which generator/variable
-    /// they close over -- verified by code review), so this one case gives
-    /// real coverage of the shared shape without a misleading "confirmed"
-    /// claim for a case that wasn't actually isolated.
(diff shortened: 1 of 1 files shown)


newhoggy:
Code review (6 finders across 2 rounds) confirmed the fix is correct and found a real additional gap: `fanout_arg`'s own non-`ArgFanout::All` eager branch (every yq-mode gate) had the identical bug, now fixed via the pre-existing `stream_outputs_checked`. `fanout_two_args`'s own non-`All` branch has the same shape but is far more intricate (oracle-verified `RejectMany`-precedence rules across 3 call sites) and needs its own dedicated verification pass — filed as #2158.

One review finding (claiming `succinctly jq 'has(.b)'` silently returns `false` via `eval_generic.rs`'s primary CLI path) did not reproduce on independent live verification using the issue's own correct repro shape (`\ud800`, not raw invalid bytes) — both real jq and succinctly correctly error at that shape. Not filing a follow-up for it; noting here in case anyone revisits.

newhoggy:
Patch coverage is 96.88% (62/64); the 2 remaining uncovered lines (`src/jq/eval.rs:2297-2298`) are provably unreachable, not undertested:

`push_owned_values(previous, &mut out)` inside my new decode-failure arm can only return `Some(control)` if `previous` (drawn from `pending_first`) is itself an escaping `QueryResult` (`Error`/`Break`/`Halt`/`Partial`). But `pending_first` is only ever assigned at line ~2323, gated by `if !result.is_escape()` a few lines above (line 2316's `if result.is_escape() { ...; return Demand::Stop; }` runs first and returns early for any escaping result) — so `pending_first` structurally can never hold an escaping value, and this branch can never fire.

This is the exact same flush-then-check shape the pre-existing `body`'s-own-escape arm a few lines below already has (lines 2305-2310, unchanged by this PR) — I copied that established pattern for consistency rather than writing a differently-shaped (and differently-reasoned-about) check for the same `pending_first` invariant. It doesn't register as "new uncovered" there only because it predates this diff.

Issue #1986, opened 2026-08-31
jq: stale comments claim Builtin::FirstStream/LastStream are parser-reachable via a second production


newhoggy:
## Summary

Found during #1972's code review (Angle C: cross-file tracer). Several comments assert that `Builtin::FirstStream`/`Builtin::LastStream` are constructed by a second, older parser production alongside the dedicated `Expr::FirstExpr`/`Expr::LastExpr` lazy AST nodes — implying both are live and a given call site could see either. This is stale: `src/jq/parser.rs`'s atom-chain (~line 1429/1441/1443) intercepts `"first"`/`"last"`/`"limit"` unconditionally before `try_parse_builtin()` is ever invoked (there is exactly one call site of `try_parse_builtin()`, at parser.rs ~1452), so those keywords' `try_parse_builtin` arms (parser.rs ~3628/3666/3683) constructing `Builtin::FirstStream`/`LastStream`/`Limit` are dead — no parser path ever produces them for `first(...)`/`last(...)`/`limit(...)` syntax.

This already matches what `builtin_first_stream_propagates_bare_halt`'s own doc comment (src/jq/eval.rs ~64114) correctly states — the two are inconsistent with each other.

## Locations to fix

- `src/jq/eval.rs` ~20439-20441 (`resolve_node`'s path-resolution comment: "`Builtin::FirstStream` is a second, older representation reachable from a different parser path")
- `src/jq/eval_generic.rs` ~6241-6244
- `src/jq/eval_generic.rs` ~8101-8104

Each should be corrected to state that `Builtin::FirstStream`/`LastStream` are parser-dead (matching `eval.rs`'s own already-correct doc comment on the same subject), or removed if the "both arms" handling is kept purely as defensive symmetry rather than because both are genuinely reachable.

## Severity

Documentation-only — no functional bug, and low risk of misleading a *user*, but it nearly misled a code-review audit (Angle C's own words: "directly contradicts parser.rs ... and this PR's own newly-consistent test comments ... but exactly the kind of comment that could mislead a future audit"). Worth a quick, mechanical doc fix.

newhoggy:
Claiming this — straightforward doc-fix, correcting the three stale comment locations named in the issue to match `eval.rs`'s own already-correct doc comment on `Builtin::FirstStream`/`LastStream` being parser-dead.

newhoggy:
[...]

PR #2169, opened 2026-09-02
fix(jq): route push_owned_values's unsafe call sites through the checked sibling

### LINKED ISSUE #2024: jq: push_owned_values uses bare to_owned -- nth/flatten/limit fanout
## Summary

Found during classification for #1989. `push_owned_values`
(`src/jq/eval.rs:2900, 2903`) uses bare `to_owned`, not `to_owned_checked`.
Reached from `fanout_arg`'s eager loop (lines 2213, 2238, 2249, 2256,
2283, 2424, 2477 -- pushing `body(arg)`'s result, which the function's
own comment says can be a genuine zero-copy borrowed `One`/`OneCursor`,
citing `nth`/`flatten` by name) and from `eval_limit`'s escaped-generator
prefix (line 26588).

(Callers `negate_fanout_core`, line 4703, and
`eval_owned_multi_keep_partial`, line 19675, are safe -- don't offset
the unsafe ones above.)

## Repro to verify

Try `[1,"\xFFbad"] | nth((0,1))` (forcing `nth`'s multi-value fan-out),
or `limit(2; .[])` on an array whose surviving elements include an
invalid-UTF8 string alongside a later escape. Verify live against the
pinned oracle and the built binary before implementing.

## Suggested next step

Not investigated beyond the classification pass. Whoever picks this up
should confirm the repro reproduces, then route the unsafe call sites
through `to_owned_checked`/`suppress_or_raise` while leaving the two
safe callers untouched -- likely needs a `_checked` sibling rather than
converting `push_owned_values` itself.

Refs #1989

newhoggy:
## Summary

- `fanout_arg`'s eager (non-`All`) branch and lazy (`ArgFanout::All`) sink,
  `fanout_two_args`'s eager loop, and `fanout_two_args_lazy`'s inner sink
  all pushed a `body` result's value through bare `push_owned_values`
  (unchecked `to_owned`) instead of `push_owned_values_checked`, silently
  substituting `""` for an undecodable string in the output instead of
  raising a decode failure -- the #1746-shaped bug, on the *body-result*
  side this time (#2023 already fixed the *argument*-decoding side of
  these same functions).
- `limit_with_n`'s `Flow::Escaped` arm (fewer than `n` outputs produced
  before the generator's own terminator fired) had the identical gap on
  its pre-escape prefix, and additionally discarded whatever `Control`
  the checked conversion itself raised in favor of the generator's own
  trailing control. Fixed so a decode failure discovered while
  materializing that prefix -- describing a value the generator already
  produced *before* its own terminator fired -- now takes precedence,
  matching the flush-then-stop precedence `fanout_arg`'s own sinks
  already give a buffered `pending_first`'s decode failure over the
  argument's trailing control.
- The two callers the issue names as already safe (`negate_fanout_core`,
  `eval_owned_multi_keep_partial`) are left untouched.

## Repro

`[1,"\xff\xfe"] | nth((0,1))` (forces `nth`'s multi-value fan-out
through `fanout_arg`'s lazy sink): pre-fix `ManyOwned([Int(1),
String("")])`, post-fix raises a decode failure with `Partial([Int(1)],
...)`.

`limit(3; (.a, error("x")))` on `{"a":"<bad>"}`: pre-fix
`Partial([String("")], Error("x"))` (silent substitution, wrong error
reported); post-fix raises the decode failure directly.

Both verified fail-without-fix / pass-with-fix.

Fixes #2024

## Test plan

- [x] New regression tests: `test_fanout_arg_lazy_sink_body_result_raises_decode_failure_2024`, `test_eval_limit_escaped_prefix_no_silent_substitution_2024` -- both confirmed to fail against the pre-fix code and pass with the fix
- [x] `cargo test --features cli,simd,regex,serde` (full suite): all green
- [x] `cargo clippy --all-targets --all-features -- -D warnings`: clean
- [x] `cargo clippy --all-targets --features std,simd,serde,cli,regex,bench-runner,large-tests,mmap-tests -- -D warnings`: clean
- [x] `cargo test --no-default-features --verbose` (no_std): clean

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -2271,7 +2271,7 @@ where
         }
         let mut out: Vec<OwnedValue> = Vec::new();
         for arg in args {
-            if let Some(control) = push_owned_values(body(arg), &mut out) {
+            if let Some(control) = push_owned_values_checked(body(arg), &mut out) {
                 return partial(out, control);
             }
         }
@@ -2303,7 +2303,7 @@ where
             Ok(v) => v,
             Err(e) => {
                 if let Some(previous) = pending_first.take() {
-                    if let Some(control) = push_owned_values(previous, &mut out) {
+                    if let Some(control) = push_owned_values_checked(previous, &mut out) {
                         body_control = Some(control);
                         return Demand::Stop;
                     }
@@ -2313,7 +2313,7 @@ where
             }
         };
         if let Some(previous) = pending_first.take() {
-            if let Some(control) = push_owned_values(previous, &mut out) {
+            if let Some(control) = push_owned_values_checked(previous, &mut out) {
                 body_control = Some(control);
                 return Demand::Stop;
             }
@@ -2324,14 +2324,14 @@ where
         // or an escaping first result would be parked and the sink would ask
         // for another value anyway -- which is the bug this arm exists to fix.
         if result.is_escape() {
-            if let Some(control) = push_owned_values(result, &mut out) {
+            if let Some(control) = push_owned_values_checked(result, &mut out) {
                 body_control = Some(control);
             }
             return Demand::Stop;
         }
         if out.is_empty() && pending_first.is_none() {
             pending_first = Some(result);
-        } else if let Some(control) = push_owned_values(result, &mut out) {
+        } else if let Some(control) = push_owned_values_checked(result, &mut out) {
             body_control = Some(control);
             return Demand::Stop;
         }
@@ -2358,7 +2358,7 @@ where
         // of it.
         Flow::Escaped(control) => {
             if let Some(previous) = pending_first.take() {
-                if let Some(body_control) = push_owned_values(previous, &mut out) {
+                if let Some(body_control) = push_owned_values_checked(previous, &mut out) {
                     return partial(out, body_control);
                 }
             }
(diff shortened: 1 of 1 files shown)


newhoggy:
Code review (5 finder agents) found a real, confirmed gap the first pass missed: `limit_with_n`'s mainline `Flow::Stopped`/`Flow::Exhausted` arms returned `items_to_result`'s (unchecked) output straight through, unguarded -- only the `Flow::Escaped`/`!satisfied` edge case was routed through a checked conversion. Live-verified: `limit(2; 1, .a)` on an undecodable `.a` (n reached cleanly, the ordinary case) gave `ManyOwned([Int(1), String("")])`, silently substituting `""`.

Fixed by adding `items_to_result_checked` (the checked twin of `items_to_result`, whose only production caller is `limit_with_n`) and routing `limit_with_n`'s result through it unconditionally, covering both the mainline and escaped arms. Also simplified the `Escaped` arm to match `eval_as`'s existing control-precedence shape, and expanded its comment to explicitly address the halt/break-vs-decode-failure precedence question review raised (citing `fanout_arg`'s own doc comment rule 3, which already establishes that a body's own error can outrank a trailing halt elsewhere in this file).

Two more gaps review found are latent/out-of-scope and filed as follow-ups rather than expanding this PR:
- #2170: `fanout_two_args`'s non-`All` eager branch has the analogous #2023-shaped argument-decoding gap (`stream_outputs` not `stream_outputs_checked`) -- unconfirmed live, current callers mask it via their own independent whole-document checks.
- #2171: `builtin_limit` (`Builtin::Limit`, parser-unreachable per #1986/#981) is an inline duplicate that now diverges from `limit_with_n`'s just-fixed shape.

All local checks re-verified green after the second commit: full test suite (`cli,simd,regex,serde`), both clippy legs, no_std.

Issue #1872, opened 2026-08-29
jq: reduce/foreach's source path-check loses foreach's own pre-error partial output


newhoggy:
**Severity: Low** (narrow trigger — a `foreach` source that both navigates through a computed value *and* legitimately emits output before the navigation failure)

## Summary

Found via code review on PR #1870's fix for #1467 (reduce/foreach source path-checking). That fix's own `path_check_fold_source` checks the *whole* source expression via `resolve_node` before `eval_owned_expr_fork` gets a chance to stream anything — correct for `reduce` (whose own output is single-shot regardless), but `foreach` legitimately streams one output per source element as it goes, and real jq's own single interleaved evaluation does exactly that even when a later element fails.

## Repro (verified against jq 1.7.1)

```console
$ echo '{"a":1}' | jq -c 'path(foreach (1,2,keys[]) as $k (.; .))'
jq: error (at <stdin>:1): Invalid path expression near attempt to iterate through ["a"]
[]
[]

$ echo '{"a":1}' | succinctly jq -c 'path(foreach (1,2,keys[]) as $k (.; .))'
jq: error (at <stdin>:1): Invalid path expression near attempt to iterate through ["a"]
```

Both raise the same error (correct, per #1467's own fix) — but jq streams the two legitimate `[]` outputs (from source elements `1` and `2`) to stdout *before* raising on the third (`keys[]`); succinctly prints nothing, since `path_check_fold_source` checks the entire `(1,2,keys[])` comma expression up front and short-circuits before `eval_owned_expr_fork`'s own `foreach`-specific partial-prefix streaming ever runs.

## Root cause

`path_check_fold_source` (`src/jq/eval.rs`) necessarily evaluates the *whole* source expression via `resolve_node` to answer "does this navigate through something untracked," which can only be answered by fully running the check to (or past) the point of failure — there's no way to check "just the part before the first failure" without actually running it. Real jq doesn't have this problem because it evaluates the source exactly once, with tracking on, interleaved with consuming each output as it's produced — check and use are the same pass, not two separate ones.

## Suggested fix direction

[...]

PR #2172, opened 2026-09-02
fix(jq): stop select() after a generator from swallowing survivor paths

### LINKED ISSUE #2050: jq: a rejecting select() after a generator builtin swallows the surviving branch's path-expression error
**Severity: Low** (an error is swallowed into a silent no-op; the direction is safe —
succinctly does nothing where jq raises — and the filter shapes are unusual)

Found by a randomised `del()` differential fuzzer while working #1690 (PR #2047).
Pre-existing on `main`, unrelated to that change.

## Summary

Inside a path expression, when a `select(...)` **rejects at least one** branch produced
by a *generator builtin* (`paths`, and presumably its relatives), the branches that
*survive* stop having their path-validity checked: an invalid path expression that jq
raises on becomes an empty path set, so `path(...)` yields nothing and `del(...)` is a
silent no-op.

The rejection is load-bearing. `select(true)` — same shape, nothing filtered — raises
correctly.

## Repro

```console
$ echo '{"a":{"b":1}}' | jq -c '[path(paths | select(length == 2) | .["a"])]'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of ["a","b"]

$ echo '{"a":{"b":1}}' | succinctly jq -c '[path(paths | select(length == 2) | .["a"])]'
[]
```

and the `del()` form the fuzzer actually found:

```console
$ echo '{"a":{"b":1}}' | jq -c 'del(paths | select(length == 2) as $p | getpath($p))'
jq: error (at <stdin>:1): Cannot index array with string "a"

$ echo '{"a":{"b":1}}' | succinctly jq -c 'del(paths | select(length == 2) as $p | getpath($p))'
{"a":{"b":1}}
```

(jq 1.7.1, `/usr/bin/jq`.)

## What isolates it

All four rows on `{"a":{"b":1}}`. Only the two where `select` *rejects something* and
the generator is a builtin diverge:

|---|---|---|
| `[path(paths \| select(length == 2) \| .["a"])]` | raises | **`[]`** |
| `[path(paths \| select(length == 2) \| getpath(.))]` | raises | **`[]`** |
| `[path(paths \| select(true) \| getpath(.))]` | raises | raises ✓ |
| `[path((1,2) \| select(. == 2) \| .a)]` | raises | raises ✓ |

So it is neither `select` on its own (row 4 is a rejecting `select` over a literal
comma and behaves correctly) nor `paths` on its own (row 3 is a non-rejecting `select`
over `paths` and behaves correctly) — it is the combination.

## Root cause

Not pinned down. `src/jq/eval.rs`, in `resolve_node`'s `select`/generator arms: the
likely shape is that a rejected branch collapses the surrounding branch set (or marks
it untracked and drops it) rather than being removed from it, so
`reject_untracked_at_terminal` never sees the branch that should have raised. Worth
checking `eval_generic.rs` for its own copy — per the repo's standing note, path
resolution exists in both evaluators and a fix in one is not a fix in the other.

## Why this is separate from #1690

#1690 is `del()`'s asymptotic cost. This reproduces identically on `main` with the
trie absent, is not `del()`-specific (`path()` shows it directly, so `=`/`|=` are
presumably affected too), and lives in branch *production*, not the delete walk. PR
#2047's differential fuzzer excludes this one filter shape explicitly, with a comment
pointing here.

Related: #986 (the untracked-terminal-branch model), #1560 (a sibling's escape
suppressing an earlier branch's own violation) — same area, and #1560 is the closest
prior art for "the wrong branch's outcome wins".

newhoggy:
## Summary

Fixes #2050. Inside a path expression, a rejecting `select(...)` downstream of a
generator builtin (`paths`, `range`, ...) used mid-pipe silently swallowed the
surviving branch's own path-validity check, instead of merely re-ordering it —
`path()` returned `[]` and `del()` was a silent no-op where jq 1.7.1 raises.

## Root cause

`resolve_seq`'s fan-out loop threaded the whole call's own *terminal* `keep`
demand (`Keep::First`, #987's "stop a generator after its first output" rule)
into *every* stage of a multi-stage pipe, not just the genuinely terminal one.
A bare generator builtin has no dedicated `resolve_node` arm, so it falls to
`resolve_leaf`'s general, `keep`-obeying case — and got truncated to its first
output before a later `select(...)` ever ran to filter it. On
`{"a":{"b":1}}`, `paths` emits `["a"]` then `["a","b"]`; `select(length == 2)`
rejects the first and would accept the second, but the second was never
produced at all.

## Fix

Widen every non-terminal stage's `keep` to `Keep::AtMost(usize::MAX)`,
reserving the caller's real `keep` for the pipe's genuinely final stage —
mirroring #1872's identical fix for `resolve_fold_source`'s own "need every
output" case.

`eval_generic.rs` needed no matching change: it has no separate path-tracking
implementation and bridges `path()`/`del()`/`=`/`|=` to `eval.rs`'s shared
`resolve_node`/`resolve_seq` for both jq and yq modes — confirmed live through
both built CLI binaries.

## Test plan
- [x] Live-verified against the pinned oracle (`/usr/bin/jq` 1.7.1): all four
  divergence-table rows from the issue now match exactly, including the two
  previously-correct rows (unregressed).
- [x] 3 new regression tests in `src/jq/eval.rs` covering the `path()` form,
  the `del()` form the fuzzer found, and the two previously-correct rows
  (pinned against future regression).
- [x] `cargo test --features cli,simd,regex,serde --lib`: 3258 passed, 0 failed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,61 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A rejecting `select(...)` after a generator builtin (`paths`, `range`, ...)
+  no longer swallows a surviving branch's own path-validity check** (#2050).
+  Inside a path expression, when `select(...)` rejected at least one branch a
+  *generator builtin* produced, the branches that survived stopped having
+  their path-validity checked at all:
+
+  ```console
+  $ echo '{"a":{"b":1}}' | jq -c '[path(paths | select(length == 2) | .["a"])]'
+  jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of ["a","b"]
+
+  $ echo '{"a":{"b":1}}' | succinctly jq -c '[path(paths | select(length == 2) | .["a"])]'
+  [] # before this fix
+  ```
+
+  and, the shape a randomised differential fuzzer actually found while
+  working #1690 (PR #2047):
+
+  ```console
+  $ echo '{"a":{"b":1}}' | jq -c 'del(paths | select(length == 2) as $p | getpath($p))'
+  jq: error (at <stdin>:1): Cannot index array with string "a"
+
+  $ echo '{"a":{"b":1}}' | succinctly jq -c 'del(paths | select(length == 2) as $p | getpath($p))'
+  {"a":{"b":1}} # silent no-op, before this fix
+  ```
+
+  Root cause: `resolve_seq`'s multi-stage fan-out loop threaded the whole
+  call's own terminal `keep` (`Keep::First`, #987's "stop a generator after
+  its first output" rule) into **every** stage of the pipe, not just the
+  genuinely terminal one. A bare generator builtin like `paths` has no
+  dedicated `resolve_node` arm, so it falls to `resolve_leaf`'s general,
+  `keep`-obeying case — and got truncated to its *first* output (`["a"]`)
+  before `select(length == 2)` ever ran, discarding the second output
+  (`["a","b"]`) that `select` would have let through. `resolve_fold_source`
+  had already established the fix for its own "needs every output" case
+  (`Keep::AtMost(usize::MAX)` in place of `Keep::First`, #1872); `resolve_seq`
+  now applies that same widened `keep` to every stage except the pipe's own
+  final one (only reached when a later stage or a non-empty static tail
+  isn't still waiting to run), which is the one position the caller's actual
+  demand still applies to.
+
+  Confirmed live against jq 1.7.1 that this generalizes beyond `paths`:
+  `path(range(3) | select(.==2))` raises "Invalid path expression with result
+  2" (the candidate `select` actually lets through), never "...result 0" (the
+  generator's own first output) — succinctly now matches. The two rows in
+  the issue that were already correct (a non-rejecting `select(true)`, and a
+  rejecting `select` over a literal `Expr::Comma` rather than a generator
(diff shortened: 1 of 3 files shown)


newhoggy:
Code review (4-finder fan-out) confirmed a genuine regression in the original fix: the
`stage_keep` widening condition included `tail.is_empty()`, which meant *any* stage
followed by a non-empty static tail (`Field`/`Index`/`Slice`, never a rejecting
construct) got widened to "need every output" -- silently reintroducing #987's own
regression for the common single-generator-plus-tail shape (`path(range(N) | .a)` was
fully materializing the generator, up to a 100,000x factor measured live, before the
static tail ever ran).

Fixed in `3f60fa956`: dropped the `tail.is_empty()` clause. #2050's own repro is
unaffected (its `select` occupies `last_dynamic` itself, not the earlier generator
stage). Verified: `path(range(1000000) | .a)` now runs in ~0.007s (was ~0.87s), all
4 divergence-table rows from #2050 still match jq exactly, full test suite green
(3258+ lib tests, all integration binaries), clippy/fmt clean.

A narrower residual (a non-rejecting builtin like `debug` sitting between two
generator stages with no filter between them still over-widens the earlier one) is
tracked as #2178 rather than fixed here -- a sound general fix needs either
recursing into user-defined function bodies or a verified allowlist of provably-1:1
builtins, and getting it wrong risks silently reintroducing #2050 for an
equivalent user-defined-filter shape.

Issue #1687, opened 2026-08-26
jq: eval_generic.rs's duplicate-key-losing _ fallback still reaches reduce/foreach/sort_by/group_by/unique_by, and limit/nth's generator-n path


newhoggy:
**Severity: Low-Medium** — architectural/completeness gap, not a regression.

## Summary

Found by two rounds of multi-agent `/code-review` on #1686 (#1607's fix for `limit`/`nth`). #1686 added native `Expr::Limit`/`Expr::NthExpr`/`Builtin::NthStream` arms to `eval_generic.rs`'s `eval_single`/`eval_builtin` so those two builtins stop falling to the `_` catch-all that materializes the whole input into an `OwnedValue` before `expr` runs (`OwnedValue::Object` is `IndexMap`-backed and cannot represent a duplicate mapping key — the same root cause #607 already fixed for `first`/`last`). That fix is complete and tested for `limit`/`nth`'s own repro, but review surfaced two related, still-open instances of the identical root cause, plus a couple of smaller residual gaps in #1686's own new code.

## 1. Other builtins still reach the lossy `_` fallback

`reduce`, `foreach`, `while`, `until`, `repeat`, and `Builtin::MinBy`/`MaxBy`/`SortBy`/`GroupBy`/`UniqueBy` have no native arm in `eval_generic.rs`'s `eval_single`/`eval_builtin` — confirmed by grep (zero hits for these variants outside their own match-arm names). Live-verified against a duplicate-key document (`b: 1\na: 2\nb: 3\n`):

```console
$ succinctly yq --jq-extensions -o json 'reduce (keys|.[]) as $k (0;.+1)' dup.yaml
2                  # want 3 -- one duplicate key silently collapsed
$ succinctly yq --jq-extensions -o json '[foreach (keys|.[]) as $k (0;.+1;.)]' dup.yaml
[1,2]              # want [1,2,3]
$ succinctly yq --jq-extensions -o json 'sort_by(1)|.[0]|keys' dup.yaml
["b","a"]          # want ["b","a","b"] if the sorted element still has the duplicate
```

`group_by`/`unique_by` show the same collapse for a duplicate key inside a grouped/deduplicated element.

## 2. `limit`/`nth`'s own generator-`n` path still loses duplicates

[...]

Issue #1819, opened 2026-08-29
jq: a value between 256 and 384 deep prints corrupted partial JSON before erroring (print_json vs MAX_VALUE_TREE_DEPTH mismatch)


newhoggy:
## Summary

`JqValue::from_owned`/construction-time guards allow a value up to `MAX_VALUE_TREE_DEPTH` (384) deep, but `print_json`'s own recursive-descent output guard caps at the lower `MAX_NESTING_DEPTH` (256). A value strictly between these two limits passes construction, starts printing, and only fails **mid-write** — after already flushing a large amount of syntactically-incomplete JSON to stdout.

## Repro (confirmed live against `main`)

```console
$ printf '1' | succinctly jq -c 'reduce range(300) as $i (null; [.])'
[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[Error: nesting depth exceeds limit of 256
```

[...]

PR #2181, opened 2026-09-02
fix(jq): preserve duplicate keys for --preserve-input's LazySeq results

### LINKED ISSUE #2066: jq: --preserve-input keeps duplicate keys for a single-value result but not an array one (jq CLI has no lazy-array JqValue)
**Severity: Low** — an asymmetry in a succinctly extension, not a reference divergence.

## Summary

Found while implementing #1687. `jq_runner.rs` has no lazy-array `JqValue`, so every
`GenericResult::LazySeq` is materialized into an `IndexMap`-backed `OwnedValue` at
output. Under `--preserve-input` — the flag whose stated purpose is preserving the
input's original formatting — that means an array-valued result loses duplicate keys
while a single-valued one keeps them.

Plain jq mode is **not** affected: collapsing is correct there (#1385, matching real
jq). This is only about `--preserve-input`.

## Repro

```console
$ echo '[{"a":1,"a":2},{"b":3}]' | succinctly jq -c --preserve-input '.'
[{"a":1,"a":2},{"b":3}]          # whole document: preserved

$ echo '[{"a":1,"a":2},{"b":3}]' | succinctly jq -c --preserve-input 'first(.[])'
{"a":1,"a":2}                    # OneCursor: preserved (#607)
$ echo '[{"a":1,"a":2},{"b":3}]' | succinctly jq -c --preserve-input 'min'
{"a":1,"a":2}                    # OneCursor: preserved (#1687)

$ echo '[{"a":1,"a":2},{"b":3}]' | succinctly jq -c --preserve-input 'sort'
[{"a":2},{"b":3}]                # LazySeq: COLLAPSED
$ echo '[{"a":1,"a":2},{"b":3}]' | succinctly jq -c --preserve-input 'map(.)'
[{"a":2},{"b":3}]                # LazySeq: COLLAPSED (pre-dates #1687)
```

## Mechanism

Two layers, and the lower one is the real cause:

1. `generic_result_to_jq_values` (`src/bin/succinctly/jq_runner.rs`) handles
   `GenericResult::LazySeq` with `seq.materialize_atomic()`, producing an
   `OwnedValue::Array`. There is no `JqValue` variant that can hold a drained
   `LazySeq`, unlike `JqValue::LazyKeysArray` for `LazyKeys`.
2. Even given one, `GenericResult::stream_json`'s `LazySeq` arm can only stream
   element cursors when `DocumentCursor::supports_sequence_streaming()` is true —
   and **only `YamlCursor` overrides it**; `JsonCursor` uses the trait default
   `false`, so `sequence_streamable_cursors` returns `None` and the arm falls back
   to `lazy_elem_to_owned`.

So closing this means implementing `stream_sequence_json` for `JsonCursor` (the YAML
side has had it since #757), then giving the jq CLI a `JqValue` that can carry the
cursors — the jq-side equivalent of what `yq_runner.rs`'s M2 streaming already does.

## Not a regression

`map(.)` has answered a `LazySeq` since #724/#725 and has collapsed here ever since;
#1687 only added more builtins with the same shape. Verified identical on the
pre-#1687 binary. Pinned as-is by
`test_preserve_input_sort_family_cursor_results_keep_duplicate_keys_1687`
(`tests/jq_cli_tests.rs`), which asserts both halves of the asymmetry so a future fix
has to update it deliberately.

newhoggy:
## Summary

- `generic_result_to_jq_values`'s `LazySeq` arm (`sort`/`unique`/`reverse`/
  `map(.)`/`sort_by`/`unique_by`) routed through `seq.materialize_atomic()`,
  which round-trips every element through an `IndexMap`-backed `OwnedValue`
  -- collapsing a duplicate object key inside a moved element even though
  `--preserve-input`'s whole point is not doing that (single-valued
  results, `first(.[])`, `min`/`max`/`min_by`/`max_by`, already preserved
  correctly via `JqValue::Cursor`).
- Switched to `seq.drain_atomic()`, which keeps each element's own cursor
  identity (or already-owned value, for a `map`-computed one), mapped to
  `JqValue::Cursor`/`JqValue::Array` the same way `ManyCursor`'s arm
  already does -- turns out `LazySeq::Cursors` (added for #1687) already
  anticipated exactly this fix per its own doc comment, so no new
  `JsonCursor`/`DocumentCursor` trait methods or `JqValue` variants were
  needed, smaller than the issue's own suggested fix direction.
- Caught by my own follow-up testing (not the issue itself): the naive
  version of this regressed #1793's depth-safety guard --
  `JqValue::Cursor` with no eager check defers nesting-depth validation to
  `write_json_at_depth`'s own later, differently-tuned panic
  (`MAX_VALUE_TREE_DEPTH`), which fires *mid-write* instead of *before*
  any output, violating the `catch_unwind` wrapper's own documented
  invariant. Added `check_cursor_nesting_depth`, a cheap structural-only
  walk (no value decoding/allocation) using the existing fallible
  `check_nesting_depth`, run before any element is wrapped as
  `JqValue::Cursor`. Also made the whole arm all-or-nothing on error
  (`return vec![]`, not a partial `JqValue::Array`), matching
  `materialize_atomic`'s own atomicity contract.

Fixes #2066

## Repro

```console
$ echo '[{"a":1,"a":2},{"b":3}]' | succinctly jq -c --preserve-input 'sort'
[{"a":1,"a":2},{"b":3}]     # was [{"a":2},{"b":3}]

$ echo '[{"a":1,"a":2},{"b":3}]' | succinctly jq -c 'sort'   # plain mode: unaffected
[{"a":2},{"b":3}]
```

## Test plan

- [x] `test_preserve_input_sort_family_cursor_results_keep_duplicate_keys_1687` updated: all of `map(.)`/`sort`/`unique`/`sort_by(.a)`/`unique_by(.a)`/`reverse` now preserve
- [x] `test_sort_family_still_collapses_in_plain_jq_mode_1687` unchanged, still passing (plain mode correctly still collapses)
- [x] `test_map_identity_reports_clean_error_on_adversarial_nesting_1793` / `test_adversarial_nesting_does_not_abort_rest_of_stream_1793`: caught my own regression, fixed, both pass
- [x] Full test suite (`cli,simd,regex,serde`): all green
- [x] Both clippy legs (clean `cargo clean -p` rebuild each time, not incremental): clean
- [x] `cargo test --no-default-features` (no_std): clean

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -12,8 +12,8 @@ use std::path::{Path, PathBuf};
 use succinctly::dsv::{build_index as build_dsv_index, DsvConfig, DsvRows};
 use succinctly::jq::document::{effective_keys, key_hash, DistinctKeyCursors};
 use succinctly::jq::eval_generic::{
-    check_nesting_depth, eval_with_cursor, to_owned as generic_to_owned, GenericResult, LazyElem,
-    MAX_NESTING_DEPTH,
+    check_nesting_depth, eval_with_cursor, to_owned as generic_to_owned, to_owned_cursor,
+    GenericResult, LazyElem, MAX_NESTING_DEPTH,
 };
 use succinctly::jq::walk::map_builtin_subexprs;
 use succinctly::jq::{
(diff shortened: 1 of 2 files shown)


newhoggy:
Code review (3 finder agents) found a real, serious regression in the first pass that I've now fixed:

**Confirmed, fixed:** `check_cursor_nesting_depth` (my own hand-rolled walker) only replicated the nesting-depth guard `materialize_atomic`'s per-element `to_owned_cursor` used to run eagerly, not the malformed-member/-delimiter validation that same function also performs. Live-verified: `echo '[1, {"bad": xyz123}]' | succinctly jq -c 'map(.)'` printed a garbled `[1,{"bad":` prefix to stdout before erroring, instead of staying atomic (empty stdout) the way `materialize_atomic`'s own doc comment promises. Fixed by calling `to_owned_cursor` itself (the exact function `lazy_elem_to_owned`'s `Cursor` arm already calls) and discarding its `Ok` value -- this restores full parity with the old validation by construction instead of re-deriving a subset of it by hand. Added a regression test (`test_lazyseq_builtins_stay_atomic_on_malformed_content_2066`), confirmed to fail against the narrower walker and pass with this fix.

Also applied the reviewer's simplification suggestion (collapsed the two now-identically-shaped match arms into one `.collect::<Result<Vec<_>, _>>()`).

**Noted, not acted on in this PR** (both correctly out of scope):
- `OneCursor`/`ManyCursor` have the same "no eager depth check" shape this PR fixes only for `LazySeq` -- but that's a pre-existing gap `#1819`'s own doc comment explicitly scoped out ("eliminating that residual would need a depth pre-check ahead of any writing, which isn't cheap to do generically... not that broader limitation"), unchanged by this diff, not a regression.
- yq mode's own `LazySeq` arm (`yq_runner.rs`) has the identical duplicate-key-collapse bug this PR fixes for jq -- out of scope per the issue's own jq-specific title. Filed as a follow-up: #2183.
- YAML M2 streaming's `LazySeq` arm appears to have no depth guard at all in its `stream_sequence_json` call chain -- unverified beyond a static trace, pre-existing (not touched by this diff). Filed as a follow-up: #2184.

All local checks re-verified green after the fix: full test suite (`cli,simd,regex,serde`), both clippy legs (clean `cargo clean -p` rebuilds), no_std, `cargo doc`.

newhoggy:
Correction to my previous comment: the two follow-up issues I referenced landed as **#2186** (yq mode's identical `LazySeq` dup-key-collapse gap) and **#2187** (YAML M2 streaming's possibly-unguarded nesting depth in `stream_json_value`), not #2183/#2184 as I guessed before actually filing them.

Issue #2001, opened 2026-08-31
jq: more to_owned_checked sites bypass optional for #1194/#1642 (contains/inside/in/pick/any_all_gen_cond, and 54 more unaudited)


newhoggy:
## Summary

Found during code review of PR #1999 (#1953's loop-family/getpath/error continuation). The #1194→#1642→#1755→#1790→#1820→#1829→#1902→#1907/#1908→#1932/#1934→#1943→#1953→#1972/#1976→#1981/#1999 issue chain is, per the reviewing agent's own words, "the signature of a missing invariant, not a completed cleanup" — each installment finds a few more `to_owned_checked` call sites that propagate a non-decode-failure error (#1194 malformed-member, or occasionally miscategorized #1642 — see caveat below) unconditionally, bypassing `optional`, instead of routing through the existing `suppress_or_raise` helper.

## Concrete sites not yet audited

A fresh scan (code review of #1999) found **54 remaining `to_owned_checked` call sites** using the raw `Err(e) => return QueryResult::Error(e)` pattern, versus 13 now migrated across #1981/#1999. Several have a **live, used** `optional` parameter elsewhere in the same function body — the exact criterion #1999 used to decide a site was a genuine consistency gap rather than a `builtin_recurse_f`-style principled exemption:

- `builtin_contains` (`src/jq/eval.rs:8470`)
- `builtin_inside` (`src/jq/eval.rs:8579`)
- `builtin_in` (`src/jq/eval.rs:6749`)
- `builtin_pick` (`src/jq/eval.rs:37011`)
- `any_all_gen_cond` (`src/jq/eval.rs:7441`)

(Line numbers approximate — re-verify against current `main` before touching, per this whole issue family's own established practice of re-checking staleness first.)

## Architectural suggestion (not required, but worth considering)

[...]

PR #2183, opened 2026-09-02
fix(jq): forward real optional into IN(s)/IN(src;s) instead of false

### LINKED ISSUE #2015: jq: IN(s)/IN(src;s) hardcode optional=false into any_all_gen_cond -- verify against real jq before fixing
## Summary

Found during code review of PR #2010 (#2001's `any_all_gen_cond`/
`builtin_contains`/`builtin_inside`/`builtin_in`/`builtin_pick` fix).

`builtin_upper_in_src` (`src/jq/eval.rs:6974-6985`, implements `IN(src;
s)`) calls `any_all_gen_cond::<W, S>(&gen, &Expr::Identity, value, false,
true)` with a **hardcoded `false`**, not the real ambient `optional` --
so even after #2010's fix makes `any_all_gen_cond`'s own entry-point
conversion respect `optional`, `IN(src; s)` never forwards anything but
`false` into it. `builtin_upper_in` (`src/jq/eval.rs:6893-6913`,
single-arg `IN(s)`) has the same shape one level up: its own `optional`
parameter is named `_optional` (unused) and it calls plain
`to_owned_checked(&value)` unconditionally, never routed through
`to_owned_checked_or_suppress`.

## Why this needs its own verification, not a blind fix

In both functions, `optional`/`_optional` is **unused elsewhere** in the
function body -- which is the exact criterion `suppress_or_raise`'s own
doc comment uses to distinguish a genuine gap (#2001's five sites all had
optional live and used elsewhere) from `builtin_recurse_f`/
`builtin_recurse_cond`'s **principled** exemption (real jq's own
`recurse` has no internal optional-suppression concept, so unused is
correct there). Whether `IN(s)`/`IN(src; s)` have a similar principled
reason to never suppress, or whether this is simply another accidental
instance of the same #1194 gap, hasn't been checked against real jq's
own behavior -- unlike #2001's five sites, which were confirmed live
call sites (`live, used optional parameter` per the issue's own
criterion) before being fixed.

## Suggested next step

Not investigated in depth. Whoever picks this up should first determine
whether real jq's `IN`/`IN(src;s)` (defined in terms of `any`/generator
consumption, similar to `recurse`) has any documented or observed
optional-suppression behavior at all before deciding whether hardcoding
`false` is a bug or a legitimate mirror of `builtin_recurse_f`'s own
exemption.

newhoggy:
## Summary

Fixes #2015. `builtin_upper_in` (`IN(s)`) and `builtin_upper_in_src` (`IN(src;
s)`) never picked up #2001's `any_all_gen_cond` fix: the former called
`to_owned_checked` unconditionally (ignoring its own `_optional` parameter),
and the latter hardcoded `false` into its `any_all_gen_cond` call regardless
of the real ambient `optional` — despite both builtins' own doc comments
defining them in terms of `any(s == .; .)`, whose shared fanout machinery
already threads a live, used `optional` (#2001).

## Investigation

Verified live against jq 1.7.1 that `IN`, `any`, and `recurse` are all caught
uniformly by the outer `?`/`try` for an ordinary raised error — real jq has
no internal per-construct suppression concept for any of them, so jq's own
behavior gives no signal distinguishing a principled exemption (like
`recurse`'s, which has its own doc comment explaining why its `optional` is
legitimately unused) from an accidental gap:

```console
$ echo '1' | jq -c '[IN(error("boom"))?]'
[]
$ echo '1' | jq -c '[IN(1; error("boom"))?]'
[]
$ echo '1' | jq -c '[IN(error("boom"); 1)?]'
[]
$ echo '[1,2]' | jq -c '[recurse(error("x"))?]'
[[1,2]]
$ echo '1' | jq -c '[any(error("x"); .)?]'
[]
```

No doc comment anywhere claims `IN`/`IN(src;s)` has a principled reason to
diverge from `any`/`all`'s own already-fixed behavior — this is a genuine
accidental gap in the same family as #2001's five confirmed sites, not a
deliberate exemption. Not reachable via live `?`/`try` syntax today (same as
every one of #2001's sites), so this closes an internal-consistency gap
rather than a user-observable regression — a defensive fix against a future
caller reaching either builtin with `optional: true`.

## Changes

- `builtin_upper_in`: `_optional` → `optional`; routes through
  `to_owned_checked_or_suppress!` instead of unconditional `to_owned_checked`;
  forwards `optional` into its own `eval_each_owned` call instead of `false`.
- `builtin_upper_in_src`: `_optional` → `optional`; forwards `optional` into
  `any_all_gen_cond` instead of hardcoded `false`.

## Test plan
- [x] Live-verified against the pinned oracle (`/usr/bin/jq` 1.7.1) — see
  investigation section above.
- [x] 5 new regression tests: malformed-member/`optional` pairs for both
  builtins (matching the #2001 pattern), decode-failure-survives-`optional`
  pins for both, and a `query!`-level test confirming `IN(...)?`/
  `IN(...;...)?` still suppress an ordinary error, matching live jq.
- [x] `cargo test --features cli,simd,regex,serde --lib`: 3268 passed, 0 failed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,41 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`IN(s)`/`IN(src; s)` now forward the real ambient `optional` instead of
+  hardcoding `false`** (#2015). Both builtins are defined in terms of
+  `any(...)` (`IN(s)` is `any(s == .; .)`, `IN(src; s)` is
+  `any(src == s; .)`), and `any`/`all`'s own shared implementation,
+  `any_all_gen_cond`, already threads a live, used `optional` parameter
+  through its root-value conversion and its generator's own error handling
+  (#2001) — but `builtin_upper_in`/`builtin_upper_in_src` never picked up
+  the matching update: the former called `to_owned_checked` unconditionally
+  (its own `optional` parameter was `_`-prefixed and unused), and the latter
+  passed a literal `false` into `any_all_gen_cond` regardless of its own
+  (also unused) `optional` parameter.
+
+  Found during review of issue #2015, which asked whether this was a
+  genuine gap or a principled exemption matching `builtin_recurse_f`'s own
+  (real jq's `recurse` has no internal optional-suppression concept, so its
+  `optional` is deliberately left unused — see that function's own doc
+  comment). It is not: unlike `recurse`, `IN`'s own doc comments describe it
+  as delegating its fanout machinery to the *already-correct* `any`/`all`
+  implementation, and no comment anywhere claimed a principled reason for
+  `IN` to diverge from that sibling. Confirmed live against jq 1.7.1 that
+  `IN`, `any`, and `recurse` all get caught uniformly by the outer `?` for
+  an *ordinary* raised error (real jq has no internal suppression concept
+  for any of the three, so it gives no signal either way); the actual
+  distinguishing gap is entirely internal to this codebase's
+  `optional`-threading discipline, in the same family as #1953/#2001/#2010.
+
+  Not reachable via any live `?`/`try` syntax today — like every one of
+  #2001's five sites, `Expr::Optional`'s own dispatch evaluates its inner
+  expression with the *ambient* `optional` and lets the outer `eval_try`
+  catch the result independently of what the callee does internally, so
+  this closes an internal-consistency gap (and a possible future-caller
+  trap) rather than a user-observable regression. A genuine decode failure
+  (invalid UTF-8) is unaffected either way — that class is never suppressed
+  by `optional`, with or without this fix.
+
 - **A rejecting `select(...)` after a generator builtin (`paths`, `range`, ...)
   no longer swallows a surviving branch's own path-validity check** (#2050).
   Inside a path expression, when `select(...)` rejected at least one branch a
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -54139,6 +54139,29 @@ mod tests {
         );
     }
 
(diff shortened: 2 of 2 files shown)


newhoggy:
Code review (5-finder fan-out) came back clean on correctness/regression checks. Two
follow-up items addressed:

1. **Test coverage gap** (finder found this by reverting each of the fix's two
   `optional`-wiring points and rerunning the test suite): the malformed-root-value test
   for \`builtin_upper_in\` only exercised the value-conversion wiring point, not the
   \`eval_each_owned\` one — a revert of that second line left all tests green. Added
   \`test_builtin_upper_in_respects_optional_for_type_mismatch_inside_s_2015\`, which
   gives \`s\` itself (not the root value) a reason to error, and verified it actually
   fails against the reverted line before restoring the fix (\`efc747012\`).
2. **Missing yq-mode coverage**: this fix is shared, unconditional code with no
   \`S::TAG\` branch, but the original tests were all jq-mode. Added a \`yq_query!\`
   counterpart (\`4a1093f16\`).

A separate, out-of-scope finding — six more call sites in \`eval.rs\` (\`builtin_tostring\`,
\`builtin_tojson\`, \`builtin_upper_index\`, \`builtin_upper_index_stream\`,
\`builtin_tojsonstream\`, \`builtin_tostream\`) share the identical unused-\`_optional\`
shape as the bug this PR fixes, with no justifying doc comment — filed as #2184 rather
than expanding this PR's scope.

All other findings (reuse/simplification/efficiency, removed-behavior audit, cross-file
tracing, CLAUDE.md conventions, line-by-line diff scan) came back clean: no regression,
no duplicated logic introduced, no CLAUDE.md violations, live-verified against jq 1.7.1
throughout.

Issue #1989, opened 2026-08-31
jq: rename to_owned/to_owned_checked to make the checked variant the compiler-enforced default (#1934 item 5)


newhoggy:
## Summary

Split out of #1934 (item 5), left open there while items 6/7 closed via that issue's own follow-up PR.

`src/jq/eval.rs`'s infallible `to_owned` and fallible `to_owned_checked` have inverted usage since #1934 was filed: 276 `to_owned_checked` call sites vs. 78 bare `to_owned` sites today (re-measured fresh; the issue's own original count was 114 vs. 71). The checked variant is now overwhelmingly the norm, but `to_owned`'s own doc comment reasoning ("changing the infallible function's signature would force a Result-threading edit at every call site") was written when the ratio was the other way round.

## Suggested fix direction (from #1934's original review)

Rename the infallible one to something that states why it's safe (e.g. `to_owned_for_error_message`), then make the bare name `to_owned` the checked one. The compiler then enumerates every remaining unchecked site as a type/usage mismatch instead of relying on grep-and-patch, one issue at a time (#1247, #1620, #1660, #1746, #1755, #1790, #1832, #1820, #1902, #1934 have each closed one slice of this same bug family so far).

## Why this needs its own scoped investigation, not a single mechanical rename

The `to_owned_checked` → `to_owned` half of the rename is purely mechanical (name-only, no behavior change). The `to_owned` → `to_owned_for_error_message` half is **not**: every one of the current ~78 bare `to_owned` call sites needs to be individually verified as a genuinely safe use (error-message rendering, a value already known-decoded, etc.) before renaming it into something that *asserts* safety. #1934's own original estimate was "~35 benign error-message rendering, leaving ~30 real result-materialization sites still unchecked" — those ~30 are exactly the kind of site #1972 found a live bug at (`nth(n; g)`'s `Item::into_owned`, a sibling unchecked-materialization pattern, not this exact function but the same bug family). Some fraction of the ~78 sites here could be live bugs too, not just naming inconsistencies.

## Suggested approach

1. Enumerate all current bare `to_owned(` call sites in `src/jq/eval.rs` (re-count fresh — the number drifts every time a sibling issue fixes one).
2. Classify each: genuinely safe (error message / already-decoded) vs. a real materialization that should be checked.
[...]

Issue #1972, opened 2026-08-30
jq: eval_single's fully-open Expr::Slice fast path is unguarded, and at least two more consumers inherit it silently


newhoggy:
**Severity: Medium** (silent wrong output with no error at all, worse than #1932/#1943's own manifestation — confirmed live, not hypothetical)

## Summary

Found via code review on PR #1970 (#1943). `eval_single`'s `Expr::Slice`/`Expr::SliceNumber` arm has a fast path for a fully-open bound (`matches!(start, None|Some(0)) && end.is_none()`, `src/jq/eval.rs` ~line 954) that returns the target array's original borrowed, undecoded value unchanged — the root cause #1932 and #1943 each patched at one specific *consumer* of this fast path (the plain array-slice arm, and `eval_slice_expr`'s own per-target loop, respectively). But the fast path itself is still unguarded, and at least two more consumers inherit the same bug — confirmed live via a temporary probe using this file's own `query!`/`yq_query!` test macros:

```
document: [1,2,"\xff\xfe",4]  (an array with an invalid-UTF-8 string)

query:    .[:] == .[:]
actual:   Owned(Bool(true))          -- no error raised at all
expected: EvalError::decode_failure

query (yq mode): "\(.[:])"
actual:   "[1,2,\"\",4]"             -- the bad element silently becomes ""
expected: EvalError::decode_failure
```

This is **more severe** than #1932/#1943's own manifestation: those produced a masked/misattributed result via an unchecked `to_owned`; these two produce **no error signal whatsoever** — a comparison silently returns `true`/`false` on corrupted data, and a string interpolation silently embeds the corrupted value.

## Root cause and consumers found so far

- Root cause: the fast path itself, `src/jq/eval.rs` ~line 954 (returns `QueryResult::One(value)` unchanged, no decode-check).
- Confirmed live consumers with this exact bug shape, each via an unchecked `to_owned` at the point the fast path's result is materialized:
  - `push_owned_values` (`src/jq/eval.rs` ~line 2820), feeding `eval_binary_fanout` — reachable via any binary operator (`==`, `+`, etc.) with a fully-open slice on one side.
  - `eval_string_interpolation_single_value` (`src/jq/eval.rs` ~line 9510) — reachable via `"\(.[:])"`-style interpolation, yq mode confirmed live (jq mode not separately re-checked but likely shares the same code path).
[...]

Issue #1934, opened 2026-08-30
jq: #1902 review found deeper decode-failure-ordering gaps and DRY opportunities in eval_as/reduce/foreach


newhoggy:
Follow-up from #1902's review (PR #1927). The review (9+ independent finder passes, several with live verification against the built library) found real issues beyond what #1927 itself fixed — some genuinely deep enough to warrant their own investigation rather than a rushed fix in an already-large PR.

## 1. A decode failure parked in `bound_control`/`init_control` can still be discarded by a later ordinary error

`eval_as`'s `bound_control` (holding a decode failure from converting an earlier bound value) is silently dropped if a *later* bound value's body raises an ordinary error — the loop's early `return partial(all_results, Control::Error(e))` arms never consult `bound_control` first. Verified live:

```
$ query!(b"{\"a\": \"\xff\xfe\", \"b\": 5}", "((.b, .a) as $v | error(\"boom\")))?", ...)
=> None   # the decode failure vanishes; "boom" (ordinary, catchable) wins and gets suppressed by `?`
```

The same class of bug exists in `eval_reduce`/`eval_foreach`'s fork loop: `finish_fork(outputs, Some(fork_control), optional)` at the loop's own abort points passes only the fork's own control, dropping `init_control` if it independently held a decode failure from INIT's own conversion.

This directly contradicts the "#1620: a decode failure is never suppressed by `?`" invariant #1902's own new code is built around. #1927 already fixed the more straightforward manifestations (INIT/input producing *nothing*, or a bare `Error`/`Break` escape) — this is the deeper case: a decode failure discovered during conversion, then raced against an unrelated error discovered later while running the body/fork.

**Why deferred rather than fixed in #1927 or #1902**: the correct precedence isn't obvious without deciding, for real, what "jq's lazy generator interleaving" should mean for a function that (unlike jq) eagerly materializes the whole bind/INIT sequence upfront (`eval_as`'s own doc comment: "the bind expression's own control is held until that loop over its prefix has run its course"). Fixing this properly means auditing every early-return in all three functions' loops, not a one-line patch, and risks a subtly-wrong ordering rule if rushed. Real jq has no analog to compare against here (an invalid document never reaches evaluation).

## 2. Six sibling `if optional => return QueryResult::None` arms still lack the decode-failure exclusion `finish_fork` now has

[...]

PR #2192, opened 2026-09-02
fix(jq): raise on an undecodable event leaf in fromstream(f)

### LINKED ISSUE #2188: jq: collect_owned() uses bare to_owned -- fromstream(f) silently substitutes "" on a decode failure
## Summary

Found during a research audit for #1989 (enumerating and classifying every
bare `to_owned(` call site in `src/jq/eval.rs`, as a precondition for that
issue's proposed `to_owned`/`to_owned_checked` rename). `collect_owned()`
(`src/jq/eval.rs:406-419`) uses bare `to_owned` at its three internal call
sites (lines 408-410), not `to_owned_checked` -- the same #1746-shaped bug
already fixed for `item_to_owned`/`push_owned_values`/`items_to_result` and
others this session (#1972, #2023, #2024).

Reached live through `fromstream(f)` (`builtin_fromstream`, ~line 32706's
`result.collect_owned()` fallback arm -- one of roughly 20
`.collect_owned()` call sites in this file).

## Repro

Verified via the library API (`JsonIndex::build` + `eval::<Vec<u64>,
JqSemantics>`), using the established `\ud800` lone-surrogate technique
(a syntactically valid but never-UTF8-decodable JSON string escape --
the only reliable way to trigger a genuine decode failure through this
API, since raw invalid UTF-8 *bytes* get lossy-substituted by the JSON
parser itself before ever reaching `collect_owned`):

- Input: `{"events":[[[],"\ud800"]]}`
- Query: `fromstream(.events[])`
- **Buggy output:** `QueryResult::Owned(OwnedValue::String(""))` -- silent,
  no error
- **Control** (same query, `"valid"` in place of `"\ud800"`): correctly
  returns `OwnedValue::String("valid")`, confirming this is a genuine
  differential, not a repro-construction artifact
- **Expected:** `QueryResult::Error(EvalError::decode_failure(...))`,
  matching every other fixed site in this bug family

Like the rest of this bug family, this is **library-API-only** -- the CLI
binary doesn't reach this exact path for `fromstream` (`eval_generic.rs`'s
own fast path handles it and lossily substitutes U+FFFD instead, a
separate and already-accepted divergence), but `src/jq/eval.rs`'s `eval()`
is the crate's public library entry point (`succinctly::jq::eval`, per
`CLAUDE.md`'s Public API section), so this is live for any library
consumer.

## Suggested next step

Add a `collect_owned_checked` sibling (or a checked variant threaded
through `collect_owned`'s 3 internal call sites), mirroring the shape
`push_owned_values_checked`/`items_to_result_checked` already established
in #2024, and route `fromstream`'s (and any other confirmed-unsafe
`.collect_owned()` caller's) usage through it. Audit the other
~20 `.collect_owned()` call sites for the same shape while in there --
`builtin_truncate_stream`'s own `.collect_owned()` was investigated during
the #1989 audit and appears structurally masked (its depth-vs-path-length
comparison always drops every event once `.` is rich enough to carry
nested undecodable content), but wasn't proven safe by construction the
way the#2023/#2024-era sites were -- worth re-checking rather than
assuming.

Refs #1989, #1746, #1972, #2023, #2024

newhoggy:
## Summary

- `builtin_fromstream`'s `collect_owned`-based fallback arm used bare
  `collect_owned`, whose `One`/`Many` arms call unchecked `to_owned` --
  silently substituting `""` for an undecodable event leaf instead of
  raising, the same #1746-shaped bug this family has already been fixed
  for elsewhere (#1972/#2023/#2024). Found via a research audit for #1989
  (enumerating/classifying every bare `to_owned(` call site in `eval.rs`
  as that issue's own precondition before any mechanical rename).
- Fixed by routing through the existing `stream_outputs_checked` instead
  of writing a new function -- it already has the exact
  `(Vec<OwnedValue>, Option<Control>)` shape this match already
  destructures into.
- `builtin_truncate_stream` has the identical `collect_owned` shape at
  its own call site but is safe by construction, not just untested:
  `stream_expr` is evaluated against the same `value` that also becomes
  `depth`, and any `value` complex enough for `stream_expr` to reach a
  nested undecodable string through it is necessarily a container, which
  always sorts above any `Int` path length in jq's ordering -- so every
  event is unconditionally dropped before a corrupted leaf could ever
  reach output. Documented inline rather than fixed, since there's
  nothing to fix; this re-derives (rather than just trusts) the #1989
  audit's own live-repro finding for this site.

Fixes #2188

## repo's established pattern for this bug family)

`{"events":[[[],"<bad>"]]}` | `fromstream(.events[])`:
pre-fix `Owned(String(""))` (silent substitution, no error);
post-fix raises a decode failure.

## Test plan

- [x] New regression test `test_fromstream_raises_on_undecodable_event_leaf_2188`, confirmed to fail against the pre-fix code and pass with the fix
- [x] Full test suite (`cli,simd,regex,serde`): all green
- [x] Both clippy legs (clean `cargo clean -p` rebuild each): clean
- [x] `cargo test --no-default-features` (no_std): clean

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -1120,6 +1120,19 @@ coverage differs:
   `Error`/`Partial` arms across these same three functions' `input`/INIT streams didn't
   exclude a genuine decode failure from `optional`'s suppression the way `finish_fork`
   already does — an internal-consistency fix, not a further widening of this policy).
+- **#2188 fixed `fromstream(f)`'s own event-collecting fallback.** `collect_owned`'s
+  `One`/`Many` arms use unchecked `to_owned`, so `builtin_fromstream`'s
+  `result.collect_owned()` fallback silently substituted `""` for an undecodable event leaf
+  instead of raising — found via a research audit for #1989 that classified every bare
+  `to_owned(` call site in this file. Fixed by routing through the pre-existing
+  `stream_outputs_checked` (already had the exact `(Vec<OwnedValue>, Option<Control>)` shape
+  the call site destructures into) rather than adding a new function.
+  `builtin_truncate_stream` has the identical `collect_owned` call shape but is safe by
+  construction, not just untested: its `stream_expr` is evaluated against the same ambient
+  value that also becomes `depth`, and any value complex enough for `stream_expr` to reach a
+  nested undecodable string through is necessarily non-scalar, which always outranks an `Int`
+  path length in jq's ordering — so every event is unconditionally dropped before a corrupted
+  leaf could reach output.
 - **The #1677 malformed-`,`/`:`-delimiter check is the narrower gap.**
   `to_owned_checked_at_depth` itself never calls `key_delimiter_ok`/`value_delimiter_ok`, so
   every builtin routed through it still misses this one check. `Builtin::Keys` (`keys`/
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -32958,7 +32958,17 @@ fn builtin_fromstream<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         // instead of halting (#791).
         QueryResult::Halt(code) => return QueryResult::Halt(code),
         QueryResult::Partial(vs, control) => (vs, Some(control)),
-        result => (result.collect_owned(), None),
+        // #2188: `stream_outputs_checked`, not bare `collect_owned` --
+        // `collect_owned`'s `One`/`Many` arms use unchecked `to_owned`,
+        // silently substituting `""` for an undecodable event leaf instead
+        // of raising, the same #1746-shaped bug this whole family has
+        // already been fixed for elsewhere. `stream_outputs_checked` gives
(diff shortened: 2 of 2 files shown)


newhoggy:
Code review found one real gap: `docs/compliance/jq/limitations.md`'s living tracking section for this exact unchecked-`to_owned` campaign (which records each incremental fix -- #1902, #1934 item 6, etc.) had no entry for this fix. Added one, and refined the `truncate_stream` safety comment to also name the top-level-undecodable-`String` case explicitly (the "container" framing alone left that case implicit, though the reasoning already covered it -- `String` also outranks `Int` in jq's ordering).

No correctness issues found in the fix itself -- the finder traced all four `QueryResult` catch-all arms end-to-end, `compare_values`'s rank-based short-circuit, and confirmed `input`/`inputs` only ever hand back pre-materialized `OwnedValue`s, with no counter-example to the `truncate_stream` "safe by construction" claim.

All local checks re-verified green after the doc/comment update: full test suite, both clippy legs (clean rebuilds), no_std, `cargo doc`.

PR #2193, opened 2026-09-02
refactor(jq): unify install_def_calls/substitute_func_param/substitute_var_impl via map_subexprs

### LINKED ISSUE #2095: jq: install_def_calls/substitute_func_param/substitute_var_impl are three hand-synced copies of the same Expr-tree walk
## Summary

Follow-up to #1371's runtime-call rewrite, found during code review of that branch.

`src/jq/eval.rs` has three functions that each implement a full, exhaustive match over every `Expr` variant to rebuild the tree, structurally identical in shape (same ~40-arm variant skeleton: `Field`, `Index`, `Slice`, `IndexExpr`, `Pipe`, `Comma`, `Array`, `Object`, `Arithmetic`, `Compare`, `And`/`Or`, etc.), differing only in what happens at a handful of leaf/target arms:

- `install_def_calls` (new in #1371, ~410 lines) — installs `DefCall`s for a given `def`.
- `substitute_func_param` (pre-existing, touched by #1371 to add `Shared`/`DefCall` arms) — substitutes a function parameter.
- `substitute_var_impl` (pre-existing, ~700 lines) — substitutes a variable binding.

## Impact

Adding a new `Expr` variant (this project adds them regularly — `Expr::Shared`/`Expr::DefCall` themselves are from #1371) requires three synchronized hand-edits across these functions. Missing one in a match that has a wildcard nearby compiles cleanly but silently drops that variant from substitution or `def`-call installation — a correctness bug reachable only through whichever specific filter shape exercises the missed pass, not caught by the compiler.

## Suggested fix

The codebase already has the factored version of this idea one level down: `src/jq/walk.rs`'s `map_builtin_subexprs(builtin: &Builtin, f: &mut dyn FnMut(&Expr) -> Expr) -> Builtin` is a generic, closure-driven transform over `Builtin`'s sub-expressions, and all three functions above already delegate their own `Builtin` arm to a `_in_builtin` sibling built on that pattern (`install_def_calls_in_builtin`, `substitute_func_param_in_builtin`). The same generalization was never applied to the outer `Expr` match itself — a `map_subexprs(expr: &Expr, f: &mut dyn FnMut(&Expr) -> Expr) -> Expr` mirroring `map_builtin_subexprs` could collapse all three call sites down to their distinct leaf logic plus one shared skeleton.

## Context

Found during code review of #1371's branch (issue-1371). Pre-existing duplication (`substitute_func_param`/`substitute_var_impl` predate #1371), but #1371 added a third occurrence (`install_def_calls`) without generalizing, despite the `Builtin`-level precedent for exactly this generalization existing in the same diff.

newhoggy:
## Summary

Fixes #2095.

`src/jq/eval.rs` had three ~40-arm hand-written matches over every `Expr` variant (`install_def_calls`, `substitute_func_param`, `substitute_var_impl`), structurally identical except at a handful of binder/leaf/opaque arms. Adding a new `Expr` variant meant three synchronized hand-edits, and missing one next to a wildcard would compile cleanly while silently dropping that variant from whichever pass forgot it.

- Adds `map_subexprs` to `src/jq/walk.rs`, generalizing the existing `map_builtin_subexprs` (`Builtin`-level) pattern one level up to `Expr` itself: a single exhaustive match applying a closure to every direct `Expr`-typed child and rebuilding the node. No wildcard arm, so a future `Expr` variant is a compile error here until categorized.
- Refactors all three `eval.rs` functions down to just their genuinely special arms (shadow checks, opaque `Shared`/`DefCall` handling, `FuncDef`, `Builtin`) plus a single fallthrough to `map_subexprs`.
- Removes the now-dead `subst_unless_shadowed_opt` helper.
- Adds `map_subexprs` unit tests in `walk.rs` mirroring `map_builtin_subexprs`'s field-order/identity checks.

Behavior-preserving refactor only — every arm was verified side-by-side across all three functions before folding (see `map_subexprs`'s own doc comment for the full per-variant reasoning).

**Flagged, not fixed** (pre-existing, all three functions agree): none of the three currently recurses into `Expr::Error`'s message, so `error($x)` referencing a variable/parameter being substituted wouldn't get that substitution. Preserved as-is per the behavior-preservation scope of this PR; documented in `map_subexprs`'s doc comment and CHANGELOG.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — full suite green (3280+ unit tests, jq/yq golden conformance included)
- [x] `cargo test --no-default-features` — no_std build green
- [x] `cargo clippy --all-targets --features cli,simd,regex,serde -- -D warnings` — clean
- [x] `cargo doc --no-deps --all-features` with `RUSTDOCFLAGS=-D warnings` — clean
- [x] New `map_subexprs` unit tests added and passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `Expr`-scrutinee `unreachable!()` fallbacks along with it. `size_of::<Expr>()` is
   unchanged at 96 bytes and is now pinned by a test.
 
+- **`eval.rs`'s three `Expr`-substitution passes now share one generic tree
+  walk** (#2095): `install_def_calls`, `substitute_func_param` and
+  `substitute_var_impl` each used to hand-write a full ~40-arm match over
+  every `Expr` variant, identical except at a handful of binder/leaf/opaque
+  arms, so adding a new `Expr` variant meant three synchronized edits — and
+  missing one next to a wildcard would compile cleanly while silently
+  dropping that variant from whichever pass forgot it. `map_subexprs`
+  (`src/jq/walk.rs`) generalizes `map_builtin_subexprs` one level up the
+  tree: it owns every variant whose recursive-structural-child handling is
+  identical across all three callers, and each caller now matches only its
+  own special-cased arms (shadow checks, opaque `Shared`/`DefCall`
+  handling, `FuncDef`, `Builtin`) before falling through to it. Both
+  `map_subexprs` and `map_builtin_subexprs` deliberately have no wildcard
+  arm, so a future `Expr`/`Builtin` variant is a compile error in one place
+  rather than a silent gap in three.
+
+  Behavior-preserving refactor, not a fix: verified by reading all three
+  functions' arms side by side before folding any of them, and every
+  existing `jq`/`yq` test still passes unchanged. `Expr::Shared` and
+  `Expr::Error` turned out to be handled identically across all three
+  callers too, but are kept as explicit per-function arms rather than
+  folded — `Shared`'s opacity is architecturally load-bearing (#1371,
+  #2077, #2096) and each caller's own comment documents a different
+  concrete hazard, and `Expr::Error`'s non-recursion into its message looks
+  like a pre-existing latent gap (`error($x)` referencing a variable a pass
+  is substituting doesn't get substituted) rather than a deliberate
+  invariant, flagged in `map_subexprs`'s own doc comment for follow-up
+  rather than silently changed here.
+
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


newhoggy:
Code review complete (standard effort, Tier 2): no findings. Verified per-function special-case preservation (shadow guards, opaque Shared/DefCall handling, FuncDef reconstruction, Builtin frame-accounting), the DefCall frames-policy divergence between install_def_calls (`frames.max(*n)`, kept explicit) and the other two (unchanged, now via map_subexprs default arm), and confirmed the flagged Expr::Error non-recursion gap is pre-existing (live-probed against main, both produce identical output) — not introduced by this refactor. Full test suite (3280+ tests), clippy, and fmt all clean.

Issue #2206, opened 2026-09-02
jq: M2 fast path regresses large-array plain field access (.data-style)


newhoggy:
Follow-up from #1576 (jq-side M2 streaming, merged): interleaved A/B benchmarking
(`scripts/ab-cli.py`, 2/10/20MB corpora, output-identity gated against both
`succ-base`/`succ-head` and pinned `/usr/bin/jq` 1.7.1) found a real, reproducible
regression on plain field access into a large array.

## Repro shape

A query like `.data` where `.data` is a large array (benchmarked at 2238 elements) —
i.e. `Expr::Field`, a single `OneCursor` result that #1576's `can_json_fast_path` gate
now handles via `JsonCursor::stream_json`.

## Numbers (2026-09-02, `succ-base` = main before #1576, `succ-head` = #1576's merge)

| Machine              | 2MB          | 10MB          | 20MB           |
|-----------------------|--------------|---------------|----------------|
| Apple M4 Pro           | +2.3%        | +2.4%         | +3.0%          |
| AMD Ryzen 9 7950X      | +7.3%        | +7.6%         | +11.4%         |

Both clear the A/B noise floor (control range roughly ±2-3% on both machines, see
`docs/plan/jq-lazy-map-select.md`'s "Slice 3, jq side" entry for the full table and
control numbers). `map(.)`/`sort_by`/`reverse` — the queries the fast path was actually
built for — are 1.4-2.0x *faster* on both machines over the same corpus, so this isn't
a general regression, just this one shape.

## Why

`JsonCursor::stream_json` (`src/json/light.rs`) buffers its rendered output into a
local `String` before copying it to the real writer, so that a later structural
failure deeper in the value can't leave a partial/garbled prefix on stdout (matching
real jq's own "parse the whole document before printing anything" atomicity — see
#2066, and #1576's own doc comment on `evaluate_m2_fast_path` for the full reasoning).
For a *large* single-result value like a 2238-element array, that buffer is a real
extra allocation + copy that the old `to_owned_cursor`/`print_json` path apparently
didn't pay in the same way, with no rendering-throughput win to offset it (there's
only one such value being rendered, not many, so the fast path's actual advantage —
skipping the per-element `OwnedValue` DOM — doesn't come into play).

## Possible directions (not investigated)

- A cheaper way to get the same zero-partial-output guarantee for a single large
  value — e.g. validate the structure before writing instead of buffering while
  writing, or a size-based dispatch between the two strategies.
[...]

Issue #1576, opened 2026-08-23
jq: no CLI streaming-output path for LazySeq (jq-side of #757)


newhoggy:
Deliberately deferred out of #757, which scoped itself to yq after measuring
the jq side.

## What #757 did, and why jq was left out

#757 taught `GenericResult::stream_json`/`stream_yaml`'s `LazySeq` arms to
render each element from its own live cursor
(`LazySeq::drain_atomic` + `DocumentCursor::stream_sequence_json`/`_yaml`),
and whitelisted `Builtin::Map` in `yq_runner.rs`'s `can_use_m2_streaming`.
That gave `succinctly yq` a ~25% wall-clock win on `map(.)` and fixed six
output-fidelity divergences from real yq.

None of it reaches `succinctly jq`, for three reasons that all have to be
addressed together:

1. **`jq_runner.rs` has no M2 streaming path at all.** There is no
   `can_use_m2_streaming` equivalent and no `stream_cursor!`; the only fast
   path is the identity raw-bytes echo. Every other query materializes
   `Vec<OwnedValue>` at `evaluate_input`/`generic_result_to_jq_values`.
2. **`JsonCursor::stream_json` is compact-only.** It returns
   `Err(fmt::Error)` for any non-compact `IndentSpec` (and for `sort_keys`),
   so pretty-printed JSON — jq's default — has no cursor-level writer.
   `JsonCursor` therefore takes `DocumentCursor::supports_sequence_streaming`'s
   `false` default, and #757's machinery correctly declines to engage.
3. **There is no correctness gap to force the issue.** Verified live against
   the pinned `/usr/bin/jq` 1.7.1: `[{"a":1,"a":2}] | map(.)` collapses to
   `[{"a":2}]` in real jq exactly as in succinctly, and JSON has no comments,
   anchors or style for the DOM path to drop. The whole payoff here would be
   throughput.

## Scope if picked up

- A pretty-capable `JsonCursor::stream_json` (`current_indent`-aware, like
  `YamlCursor::stream_json_value` already is), plus its
  `stream_sequence_json` implementation and a `true` from
  `supports_sequence_streaming`.
- A `can_use_m2_streaming` equivalent and a streaming output loop in
  `jq_runner.rs`, with its own flag gate (`-S`, `--tab`, `--slurp`,
  `--null-input`, named vars, ...).
- Its own golden and benchmark matrix. #757's numbers do not transfer: jq's
[...]

PR #2207, opened 2026-09-02
feat(jq): stream map/sort/navigation output from cursors instead of an OwnedValue DOM

newhoggy:
## Summary
- jq-side counterpart to #757 (yq's M2 streaming, PR #1579): `map`, `sort`/`sort_by`/`unique`/`unique_by`/`reverse`, `min`/`min_by`/`max`/`max_by`, and plain navigation (`.`, `.field`, `.[n]`, `select`, `first`/`last`/`nth`) now render straight from `JsonCursor` instead of materializing a full `OwnedValue` tree, via a new `can_json_fast_path` gate in `jq_runner.rs` sharing `yq_runner.rs`'s AST-shape predicate (now factored into `m2_gate.rs`)
- `JsonCursor` gains a pretty/sort-capable recursive JSON writer and `stream_sequence_json` support, mirroring what #757 gave `YamlCursor`; a new `JsonConvention` enum threads jq's own number-literal/escape/duplicate-key conventions through the shared streaming machinery (previously hardcoded to yq's)
- map/sort-shaped results buffer before writing (matching #2066's array-construction atomicity fix); other shapes stream directly. The fast path falls back to the existing general path whenever it detects malformed input, restricted by `m2_json_fallback_safe` to AST shapes where that fallback is provably safe
- 11 new golden fixtures under `tests/data/jq-golden/cases/`, captured from pinned jq 1.7.1, covering duplicate-key collapse, pretty/compact/-S/--tab output, `sort_by`/`unique_by`/`reverse`, `min_by`/`max_by`, `map(select(...))`, nested containers, and number-literal reformatting
- Design doc updated with the landing writeup and interleaved A/B benchmark results (map/sort_by/reverse 1.4-2.0x faster on M4 Pro and 7950X; unique_by/min_by/max_by/keys_unsorted neutral as expected; a real, reproducible regression on large-array plain-field access from per-value buffering, filed as follow-up #2206)

## Test plan
- [x] `cargo build --features cli,simd,regex,serde` succeeds
- [x] New golden fixtures captured from pinned `/usr/bin/jq` 1.7.1 via `scripts/sync-jq-golden.sh`; regenerating against all existing cases touched none of them
- [ ] CI: fmt, clippy, full test suite, coverage


newhoggy:
## Coverage gap audit (025358b0)

Went through all 46 uncovered new lines and all 59 indirectly-lost ones. They
split cleanly into "reachable, now tested" and "unreachable by construction".

### Closed

| Where | Lines | How it's now reached |
|---|---|---|
| `json/light.rs` `scalar_end_pos` | `false`/`null` width arms | `[1, false]` / `[1, null]` streamed, plus the stray-trailing-comma form that only lands on `]` if the width is right |
| `json/light.rs` `trailing_gap_ok` | "ran out of text" | direct call — the semi-index rejects `[1` long before a streamer sees it |
| `json/light.rs` `write_json_number` | `JqCompat` float fallback (finite + non-finite) | `[1.]`, `[1.e999]` |
| `json/light.rs` `write_json_string_pretty` | `Preserve` arm | `--preserve-input` + any non-compact style (compact `Preserve` short-circuits to the raw echo) |
| `json/light.rs` `stream_json_sequence` | empty-cursor `[]` | `stream_sequence_json(&[], …)` at both indents |
| `m2_gate.rs` | `Expr::NthExpr` | direct call — the parser only ever builds `Builtin::NthStream`. The test also pins that all three alternatives sharing that arm answer identically, which is the only thing that makes sharing it correct |
| `jq_runner.rs` | `halt` on the `LazySeq` path | `map(select(halt))` / `sort_by(halt)` / `unique_by(halt)` — `select`'s cond and `sort_by`'s key are the two sub-expressions the gate passes over without recursing, so they can carry a `halt` where an `if` body can't. jq 1.7.1 agrees: no output, exit 0 |
| `jq_runner.rs` (all 45 indirect) | `print_json` pretty/empty containers, `null`/`true`, escaped strings + keys, `--ascii-output`, both duplicate-key-probe paths | These are still live — every run the gate declines reaches them — they just stopped being reached by the identity queries that used to. One table now asserts the property that makes two printers safe to keep: **for every container shape the M2 path and the DOM path (`--unbuffered`) agree with each other and with jq 1.7.1** |

### Left uncovered, deliberately

| Where | Why no test can reach it |
|---|---|
| `document.rs` 445, 522, 547-548 (+ indirect 500-525) | `DocumentCursor` trait defaults. Both impls (`JsonCursor`, `YamlCursor`) now override every one of them — the indirect loss is *because* this PR added the JSON overrides |
| `yq_runner.rs` 4612-4614, 4762-4764, 5482-5484 | All three sit inside `match root.value()`'s `_ =>` arm, which the file's own comments mark defensive-only: "the root cursor (bp_pos 0) always reports the virtual document sequence". Verified — `syq -e`, `-o json -e`, `-i -e`, over stdin and files, all take the `Sequence` arm |
| `eval_generic.rs` (11 lines) | `\|_\| Ok(())` newline callbacks in tests whose stream errors before writing anything, so the callback is never invoked |
| `jq_runner.rs` 1605 | `resume_unwind` for a panic that isn't a nesting-depth panic |
| `light.rs` 2765-2766, 2786 | non-`String` object key — `effective_fields_checked` (#1194) refuses one first, as the code comment already says |
| `light.rs` 2746-2747 | the no-match fall-through of two `if let`s that always match: an object with ≥1 field always has a last field and a text position |
| `light.rs` 2166 | `raw_bytes()` returning `None` on the compact+`Preserve` echo — an unterminated container never reaches `stream_json` |
| `light.rs` 2911 | the fallback's `Int` arm: every scanner-accepted span that parses as `i64` is either RFC-valid or leading-zero-shaped, and both keep their literal instead |

### Filed along the way

**#2220** — a trailing-dot mantissa with an exponent loses its source literal in
`OwnedValue::from_number_bytes`:

```console
$ printf '[1.e999]' | jq -c .                  # 1.7.1
[1E+999]
$ printf '[1.e999]' | succinctly jq -c .
[1.7976931348623157e+308]
```

Older than this PR and identical on the DOM path (`--unbuffered` is byte-for-byte
the same), so it's not a #1576 regression — it lives in the shared conversion,
which has leniency escapes for a *leading* dot and for leading zeros but none for
a trailing one. `test_stream_json_number_invalid_span_float_fallback_1576` pins the
current output and points at the issue; note that `write_json_number`'s non-finite
arm is only reachable *through* that bug, so fixing #2220 may leave it dead.

### Verification

`cargo fmt --check`; clippy `--all-targets --all-features` and the ci.yml second
leg; `cargo test` (default), `--features simd`, `--no-default-features`, the
`--features cli` CLI test set; `RUSTDOCFLAGS=-D warnings cargo doc --no-deps
--all-features`. All green.

Issue #2214, opened 2026-09-02
jq/perf: #2073's isolate-and-probe route costs 1.37x on `?` with a path-context rest (200k-element fan-out)


newhoggy:
Found while reviewing the #2073 fix branch. The regression is a deliberate,
accepted cost of that correctness fix -- filed so the number is on record and
so the shape can be optimised later, not to argue against the fix.

## Measurement

`Expr::Optional`'s path-context arm used to evaluate `[inner, ...rest]` in one
combined recursive call whenever `rest` needed path context. #2073 replaces
that with isolate-`inner` + `path_probe_stage` + `continue_rest_with_paths`,
which materializes a `[path, value]` `OwnedValue` pair per output and splits
it back apart before continuing `rest`.

Apple M4 Pro, release build, interleaved A/B (5 reps, alternating binaries
within each rep per `docs/guides/benchmarking.md`), output verified
byte-identical between the two binaries on every row. Input is a 200,000-element
array of small objects.

| query | pre-#2073 | post-#2073 | ratio |
|-------|-----------|------------|-------|
| `.items \| (.[])? \| key` | 0.470s | 0.642s | **1.37x** |
| `.items \| (.[])? \| (key, .k)` | 0.372s | 0.440s | 1.18x |
| `.items \| .[] \| key` | 0.296s | 0.295s | 1.00x |
| `.items \| (.[])? \| .k` | 0.105s | 0.104s | 1.00x |

The same shape at 20,000 elements measured 1.28x, so the ratio grows slowly
with fan-out width, consistent with a per-output allocation rather than a
fixed overhead.

## Scope

Only `?` with a path-context `rest` (`key`/`parent`/`path`/`file_index`
downstream of the `?`) is affected. Every other shape, including `?` with an
ordinary `rest` and path-context navigation without a `?`, is unchanged --
both control rows above are 1.00x.

`Try`/`Label`/`If`/`Limit`/`FirstExpr` have paid this same cost since #1409;
#2073 simply stops `Optional` being the one arm that avoided it by combining
with `rest`.

## Suggested approach

Nothing here is specific to `Optional`, so any fix helps all six arms:

- `continue_one_paired` does `split_probe_pair` -> `Vec<OwnedValue>` path +
  `OwnedValue` value per output. The path is rebuilt from scratch each time
  (`fallback_path.to_vec()` on the fallback route, a fresh `Vec` out of the
  probe's array otherwise) where the paths of adjacent outputs of a single
[...]

Issue #2213, opened 2026-09-02
jq: missing-key navigation skips `rest` and drops its own `new_path`, so `key`/`parent` after `(.missing)?`/`try`/`if`/`first` report a stale path


newhoggy:
Found while reviewing the #2073 fix branch. Pre-existing on `main` for five
of the six constructs below; #2073 makes `?` join them.

## Symptom

`Expr::Field`'s missing-key arm in `eval_stage_with_path_context` returns
early:

```rust
// jq returns null for missing fields on objects (not an error)
return QueryResult::Owned(OwnedValue::Null);
```

It never runs `rest`, and never threads the `new_path` it just computed two
lines above. `Expr::Index`'s out-of-range/`optional` arm has the same shape.

The result is that a *missing* field reports a stale path -- the ambient one
from before the navigation -- to any downstream path-context builtin, via
`split_probe_pair`'s fallback branch. That fallback's own doc comment says it
answers "shapes the probe never produces ... which a correct probe run cannot
reach", but an isolating arm plus a missing key reaches it in ordinary use.

Input `{"b":{"x":1}}`, all on the #2073 branch:

```console
$ succinctly jq -c '.b | (.zz)? | key'            # "b"   <- stale; should be "zz"
$ succinctly jq -c '.b | (try .zz) | key'         # "b"   <- same, and same on main
$ succinctly jq -c '.b | (label $o | .zz) | key'  # "b"
$ succinctly jq -c '.b | (if true then .zz else empty end) | key'  # "b"
$ succinctly jq -c '.b | (first(.zz)) | key'      # "b"
$ succinctly jq -c '.b | .zz | key'               # null  <- rest skipped entirely
```

`path(.b.zz)` is `["b","zz"]` in succinctly *and* in real jq 1.7.1, so
`"zz"` is the answer all six should give. Instead there are two wrong
answers: the five isolating constructs report the stale parent key, and the
plain pipe reports `null` because `rest` never ran at all (that `null` is
`.zz`'s own value leaking out as the final output, not `key`'s result).

There is no jq oracle for bare `key`/`parent` (succinctly extensions), but
`path()` pins what the path *should* be, and the internal inconsistency
between the six spellings is itself the defect.

## What #2073 changes

`?` moves from the second answer to the first:

| query | pre-#2073 | post-#2073 |
|-------|-----------|------------|
| `.b \| (.zz)? \| key` | `null` | `"b"` |
[...]

Issue #2212, opened 2026-09-02
jq: #2073 made `optional` always-false inside the path-context evaluator, retiring #1826/#1869/#1964's machinery with it


newhoggy:
Found while reviewing the #2073 fix branch, and deliberately excluded from it.

## What changed

#2073 removed `Expr::Optional`'s combined-with-`rest` fallback in
`eval_stage_with_path_context`, so `rest` no longer inherits a forced
`optional = true` from a `?` several stages earlier. That was the only
remaining producer of `optional == true` anywhere inside the path-context
evaluator.

## Evidence that `optional` is now always `false` there

Empirical -- adding

```rust
assert!(!optional, "PROBE: optional=true reached path-context evaluator");
```

at the top of `eval_stage_with_path_context` and running the full suite:

| tree | result |
|------|--------|
| `main` (pre-#2073) | probe fires; 5 panics across `jq_cli_tests` + `yq_cli_tests` |
| #2073 branch | **0 hits**, all 7012 tests pass |

Static -- the only site left that passes `true` as an `optional` argument on
a route that could reach a pipe is `eval_single`'s
`Expr::Optional(Expr::IndexExpr { .. } | Expr::SliceExpr { .. })` arm, which
hands it to `eval_index_expr`/`eval_slice_expr`. Both use `optional` solely
for their own final `index_one`/`index_one_owned` step and never forward it
into `eval_pipe`, so it cannot re-enter path context.

## What that makes dead

Three earlier fixes exist only to give the path-context evaluator a correct
response to `optional == true`, and are now inert:

- **#1826** -- `Builtin::Map`'s force-`false`-per-element plus atomic catch.
- **#1869** -- `Expr::Iterate`'s force-`false`-per-element plus keep-prefix
  catch.
- **#1964** -- routing `continue_rest_with_context`'s and
  `continue_rest_with_fresh_root`'s `Partial` arms through
  `catch_error_under_optional` instead of a bare `partial(...)`.

Plus `catch_error_under_optional`'s own `(true, true)` / `(false, true)`
cells and its `Error(e) if optional` arm.

## The coverage cliff

Mutation check -- forcing `let optional = false;` at the top of
`catch_error_under_optional` and running the full suite fails **exactly one
test**:

```
jq::eval::tests::test_catch_error_under_optional_full_truth_table_1888
```

That is a direct unit test of the helper. Every end-to-end test that used to
cover those cells (`test_map_path_context_builtin_optional_is_atomic_1826`,
[...]

PR #2217, opened 2026-09-02
fix(jq): stop `?` from suppressing a later pipe stage's error under path context (#2073)

### LINKED ISSUE #2073: jq: `?` suppresses a later pipe stage's error when rest consults path context
Found while fixing #1409 (PR #2071), and deliberately excluded from it.

## Symptom

`?` suppresses errors raised by *later* pipe stages, which real jq never does.

```console
$ echo '{"a":[1,2,3]}' | jq -c '(.a)? | error("boom")'; echo "exit=$?"
exit=5

$ echo '{"a":[1,2,3]}' | jq -c '(.a)? | map(if . == 2 then error("boom") else . end)'; echo "exit=$?"
exit=5
```

succinctly agrees on the first (the isolated fast path), but not once `rest`
consults path context, which sends `Expr::Optional` down its combined fallback:

```console
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a | (.[])? | (key, error("boom"))'; echo "exit=$?"
0
exit=0
# real jq: prints 0, then fails with `boom`, exit 5
```

This is the "rest inherits suppression" gap #1335's own arm comment already
names as a known cost of its `needs_path_context` guard, and which #1409's
description also lists.

## The fix is available but has wider blast radius

#2071 added `path_probe_stage`, which lets `Expr::Optional` drop the
`needs_path_context` guard entirely and always take the isolated path — I
prototyped exactly that and the repro above starts erroring correctly, with path
threading still right (`.a | (.[])? | key` stays `0`, `1`, `2`).

But it also stops `rest` inheriting the arm's ambient `optional`, which is an
evaluator-wide model two prior fixes deliberately encode. Both of these fail
under the change, and both assert exit code 0 where the change gives 5:

- `test_map_path_context_builtin_optional_is_atomic_1826`
- `test_iterate_path_context_ambient_optional_keeps_prefix_not_skip_1869`

Their doc comments state the model explicitly ("an ambient `optional` ... still
threaded downstream per this evaluator's model"). Real jq says exit 5 for the
analogous queries, so those expectations look wrong too — but re-deriving them,
and checking what else in the evaluator depends on ambient-`optional`
propagation, is a piece of work in its own right rather than a rider on #1409.

## Suggested approach

1. Confirm against jq 1.7.1 (and yq where the construct exists) what each of
   #1826's and #1869's shapes should actually do.
2. Drop `Expr::Optional`'s `needs_path_context` guard in favour of the probe.
3. Update those two tests with the oracle-derived values.

newhoggy:
## Summary

- `Expr::Optional`'s path-context arm (`eval_pipe_with_path_context_internal`) fell back to evaluating `[inner, ...rest]` combined under a forced `optional=true` whenever `rest` needed path context (`key`/`parent`/...), so the flag rode downstream as an ordinary parameter into `rest`'s whole evaluation tree and could swallow an error with nothing to do with the `?` several stages earlier (`.a | (.[])? | (key, error("boom"))` silently swallowed the error; real jq hard-errors, exit 5).
- Isolates `inner` unconditionally instead, exactly like the `Try`/`Label` arms already do, using `path_probe_stage` (#1409) to keep per-output path threading correct. `rest` now always continues with the *ambient* `optional` the arm itself received, never a forced one.
- Follow-up commit addresses review findings: retires now-dead ambient-optional machinery (`optional` is provably always `false` inside the path-context evaluator post-fix), fixes stale doc comments describing the removed mechanism, and records/files three deferred follow-ups (#2212, #2213, #2214) rather than silently expanding this fix's scope.

## Test plan

- [x] `test_map_path_context_builtin_optional_is_atomic_1826` / `test_iterate_path_context_ambient_optional_keeps_prefix_not_skip_1869` (+ yq twin) updated to oracle-verified (jq 1.7.1) post-fix values
- [x] New regression tests pin the issue's own repro directly, jq and yq modes
- [x] `cargo test --features cli,simd,regex,serde --test jq_cli_tests --test yq_cli_tests`: 1277 + 1368 passed
- [x] `cargo test --features cli,simd,regex,serde --lib`: 3314 passed
- [x] `cargo test` (default features) and `cargo test --no-default-features` (no_std): passing
- [x] `cargo clippy --all-targets --all-features -- -D warnings`: clean
- [x] Live-verified the issue's own repro now hard-errors (exit 5) with correct partial output, and existing path-threading coverage (`.a | (.[])? | key`) is unregressed

Closes #2073.

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -2784,6 +2784,14 @@ helper (the same one `Expr::Iterate`/`Builtin::Map` already use for this
 exact "keep the prefix, gate only the error" policy) instead of a bare
 `partial(...)` call.
 
+**The output above is unchanged since #2073, but the mechanism behind it
+moved.** That fix stopped `Expr::Optional` handing `rest` a forced
+`optional = true`, so `continue_rest_with_context`'s own `optional` gate no
+longer fires for this query -- the suppression now happens one level up, in
+`Expr::Optional`'s own structural catch of the isolated `recurse(...)`'s
+`Partial`. Nothing left in the path-context evaluator produces
+`optional == true` at all; see #2212.
+
 **One site remains unfixed**: `ParentN`'s own `n` argument (`parent((1,2))`
 still array-collapses `n` to `[1,2]`, then errors on the wrong type -- a
 narrower, lower-traffic case than the three arms fixed above, not attempted
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -31627,44 +31627,39 @@ fn eval_stage_with_path_context<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         // `query_result_to_owned_values` convention (the Array arm discards
         // it instead, since a halted array never finished constructing).
         //
-        // Isolating `inner` this way only works when nothing downstream
-        // needs the per-output `current_path` `inner` would have produced
-        // had it continued straight into `rest` -- `Field`/`Index`/`Iterate`
-        // (and friends) only compute+thread an updated path when they have
-        // a non-empty `rest` to recurse into; evaluated with an empty
-        // `rest` for isolation, they just return bare values, so
-        // `continue_rest_with_context` below would fan them into `rest`
-        // using the *stale*, pre-`inner` `current_path` instead of each
-        // output's real one (`.a | (.[])? | key` would wrongly report every
-        // element's key as `"a"` instead of its index). This is a
-        // pre-existing limitation of the isolate-then-continue shape shared
-        // by the `Try`/`Label`/`If`/`FuncDef`/`Limit` arms elsewhere in this
-        // function. `Comma` escaped it by combining each branch with `rest`
-        // instead (#1509) -- a trick that doesn't carry over to constructs
-        // that scope something `rest` must stay outside of, since it would
-        // let `rest`'s own errors be caught by `try`'s handler or `label`'s
-        // break scope. Those arms instead keep isolating and append
-        // `path_probe_stage`'s `[path, .]` inside the isolated pipe
-        // (#1409), which would work here too -- but adopting it for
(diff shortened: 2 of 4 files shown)


newhoggy:
CI is green (30/30). One thing in the coverage report deserves an explicit
note rather than a shrug, since it looks like a regression and is not one:

> 🔴 5 lines lost coverage, 🟢 0 lines gained coverage on unchanged code.

All five are `if optional` arms, and they are the direct, intended
consequence of this PR:

| line | enclosing fn | arm |
|---|---|---|
| `src/jq/eval.rs:5032` | `binary_fanout_each` | `if optional { Flow::Exhausted }` |
| `src/jq/eval.rs:30743` | `eval_stage_with_path_context` | `ParentN`: `Err(_) if optional` |
| `src/jq/eval.rs:31604`/`:31606` | `eval_stage_with_path_context` | `StringInterpolation`: `Error(_) if optional` |
| `src/jq/eval.rs:31658` | `eval_stage_with_path_context` | `Object`: `Err(Control::Error(_)) if optional` |

Removing `Expr::Optional`'s combined-with-`rest` fallback removed the last
producer of an ambient `optional == true` anywhere reachable from the
path-context evaluator, so these arms are no longer reachable to cover. That
is the fix working, not coverage slipping: each of those arms existed to
answer a suppression that this PR establishes should never have arrived.

Verified two independent ways:

- An `assert!(!optional, ...)` at the top of `eval_stage_with_path_context`
  passes the entire suite on this branch and fires 5 times on `main`.
- Forcing `let optional = false;` inside `catch_error_under_optional` now
  fails exactly one test -- its own unit test -- where it used to fail the
  end-to-end tests too.

Tracked in #2212, which now carries the line-by-line list and the
reachability argument, including why `binary_fanout_each` (which is not in
the path-context evaluator at all) is on it. #2213 and #2214 cover the other
two review findings: the missing-key stale-path change this PR pins with a
characterization test, and the measured 1.37x cost of the isolate-and-probe
route on `?` with a path-context `rest`.

Issue #2139, opened 2026-09-01
jq: resolve_index_expr (path()/=/|= write side) still evaluates E[K]'s target once for all keys, unlike eval_index_expr


newhoggy:
**Severity: Low** (side-effect count only, same class as #2032 -- values and
errors agree)

## Summary

#2032 fixed the *value*-position evaluators (`eval::eval_index_expr` and
`eval_generic::eval_index_expr`) so `E[K]`'s target `E` is re-evaluated once
per output of the key generator `K`, matching jq's own `K as $k | E | .[$k]`
compilation. `resolve_index_expr` (`src/jq/eval.rs`) — the *path*-position
sibling that backs `path()`/`=`/`|=`/`del()` — still evaluates `target` once
outside the key loop and reuses that single result across every key.

## Repro

```console
$ echo '[1,2]' | jq -c 'path((.[] | stderr)[("a","b")]?)'          # jq 1.7.1
                                                                    # stderr: 1212
$ echo '[1,2]' | succinctly jq -c 'path((.[] | stderr)[("a","b")]?)'
                                                                    # stderr: 12
```

## Why this wasn't fixed alongside #2032

`resolve_index_expr` has a materially more involved fix than
`eval_index_expr`'s did. It doesn't just compute `target_branches` once —
it uses that single computation's *escape* (`target_escape`) to decide
whether to truncate `keys` to just its first element (`let keys =
if target_escape.is_some() { &keys[..1] } else { &keys[..] };`), a trick
that only makes sense because target is evaluated once for the whole call.
Making target genuinely per-key requires replacing that truncation with
real per-key resolution and per-key escape handling, while preserving the
surrounding `!trackable`/`is_passthrough_target`/`Keep` invariants this
function already carries (each documented against multiple prior issues:
#843, #896, #972, #985, #986, #1591) — enough additional risk and edge-case
surface that #2032 scoped it out rather than risk both fixes in one PR.

## Suggested fix direction

Mirror `eval_index_expr`'s #2032 restructuring (evaluate `target` inside
the per-key loop, accumulate branches with the running prefix preserved on
escape) but verify each of the function's own already-hardened edge cases
individually against the pinned oracle before and after, given how many
distinct historical regressions are already encoded in its comments.

newhoggy:
Investigated while working #2143's sibling fix, but not claiming this one yet — found a blocking complication worth recording first.

This issue's own repro is now stale against current \`main\`:

\`\`\`console
[...]

Issue #2032, opened 2026-08-31
jq: E[K] evaluates its target once where jq re-runs it per key (side-effect count)


newhoggy:
**Severity: Low** (side-effect count only — values and errors agree; needs an `E[K]` whose target `E` has an observable side effect and whose key `K` is multi-output)

## Summary

jq compiles `E[K]` as `K as $k | E | .[$k]`, so the *target* `E` is re-evaluated once per output
of the key generator `K`. `succinctly jq` evaluates `E` once and loops the keys over the cached
result, so a side effect in `E` fires once total instead of once per key.

## Repro (verified against jq 1.7.1)

```console
$ echo '[1,2]' | jq -c '[(.[] | stderr)[("a","b")]?]'
1212        # on stderr — each element's stderr fires twice, once per key

$ echo '[1,2]' | succinctly jq -c '[(.[] | stderr)[("a","b")]?]'
12          # on stderr — once per element
```

Stdout and the exit code agree in both (`[]`, exit 0); only the side-effect count differs.

## Where

Both evaluators share the shape:

- `eval_index_expr` (`src/jq/eval.rs`) — the value path, which the repro above exercises.
- `resolve_index_expr` (`src/jq/eval.rs`) — the `path()`/write path, which computes
  `target_branches` once outside the per-key loop for the same reason.

Found while fixing #1872, which made `resolve_index_expr` newly relevant: a `reduce`/`foreach`
source's values can now come from `resolve_node`'s own branches, so any place the resolver's
evaluation count differs from jq's is a fold-source side-effect divergence too.

## Notes

- This is a *count* divergence, not a value one, so no golden can pin it — the jq-golden harness
  only compares stderr for cases that exit non-zero. A CLI test counting `stderr` occurrences is
  the way (see `fold_source_side_effect_fires_once_per_output_1872` in `tests/jq_cli_tests.rs`
  for the pattern; `stderr` writes no trailing newline, so count occurrences, not lines).
- Fixing it means moving the target evaluation inside the key loop, which trades a real
  performance cost (re-running `E` per key) for exact side-effect fidelity — worth measuring
  before committing to it.

## Related

- #1872 — the fold-source single-evaluation fix that surfaced this as a residual; recorded under
  "Fixed by #1467 and #1872" in `docs/compliance/jq/limitations.md`.

newhoggy:
🔧 Claimed by agent `Bridget` — branch `issue-2032-index-target-reeval`

[...]

PR #2223, opened 2026-09-02
fix(jq): re-evaluate E[S:T]'s target once per (s, e) pair, not once overall

### LINKED ISSUE #2143: jq: E[S:T] evaluates its target once where jq re-runs it per (s,t) pair (side-effect count) -- eval_slice_expr sibling of #2032
**Severity: Low** (side-effect count only -- values and errors agree; needs an `E[S:T]` whose
target `E` has an observable side effect and whose bounds `S`/`T` are multi-output).

## Summary

Same bug family as #2032 (fixed by #2142), one function over: jq compiles `E[S:T]` as
`S as $s | T as $t | E | .[$s:$t]` -- the target `E` is re-evaluated once per `(s, t)`
combination the bound generators produce. `eval_slice_expr` (`src/jq/eval.rs`, immediately
below `eval_index_expr`) and its `eval_generic.rs` sibling both evaluate `E` once and reuse
it across every `(s, t)` pair instead, so a side effect in `E` fires once total instead of
once per pair.

Surfaced during #2032's investigation (credit to that session's findings comment), but its
own repro command was invalid -- `(.[] | stderr)[(0,1):(1,2)]` on `[1,2]` slices a *number*
target (`.[] | stderr` on `[1,2]` yields `1`, `2`), and jq can't slice a number at all
(`Cannot index number with object`, exit 5) regardless of this bug. Corrected below with an
array-valued target so the slice itself is legal.

## Repro (verified live against jq 1.7.1)

\`\`\`console
\$ echo '[[10,20],[30,40]]' | jq -c '[(.[] | stderr)[(0,1):(1,2)]]'
[10,20][30,40][10,20][30,40][10,20][30,40][10,20][30,40][[10],[30],[10,20],[30,40],[],[],[20],[40]]
# stderr fires 8 times (2 elements x 4 (s,t) pairs) -- target re-run once per pair

\$ echo '[[10,20],[30,40]]' | succinctly jq -c '[(.[] | stderr)[(0,1):(1,2)]]'
[10,20][30,40][[10],[30],[10,20],[30,40],[],[],[20],[40]]
# stderr fires only 2 times -- target evaluated once, reused across all 4 (s,t) pairs
\`\`\`

Stdout (the trailing `[[10],[30],...]` array) and exit code agree in both (`0`); only the
side-effect count differs -- same shape as #2032.

## Where

- `eval_slice_expr` (`src/jq/eval.rs`) -- the value path.
- `eval_generic.rs`'s own `eval_slice_expr` sibling (the primary CLI dispatch path; the repro
  above goes through this one).
- `resolve_slice_expr` (`src/jq/eval.rs`, the `path()`/write-path sibling) likely has the
  identical divergence for the identical reason `resolve_index_expr` does (tracked
  separately as #2139) -- not confirmed live in this issue, worth checking when scoping the
  fix.

## Suggested approach

#2142 already established and shipped the fix pattern for the identical bug in
`eval_index_expr`/`eval_generic::eval_index_expr` (target moved inside the per-key loop,
per-key `try_reserve` capacity guard replacing the old upfront `keys * targets` product,
Halt/Break/Error priority against a pending generator halt preserved). Porting that same
restructuring to `eval_slice_expr`'s `(s, t)` pair loop should be mechanical now that the
pattern is proven and tested, rather than needing its own design pass -- see #2142's diff
and its `KeyTargets` enum for the shape to mirror (a `(s, t)`-keyed equivalent).

The write-path sibling (`resolve_slice_expr`) should stay out of scope for this issue and be
folded into #2139's resolve-path design work instead, for the same reasons #2139 itself
gives for `resolve_index_expr`.

## Related

- #2032 / #2142 -- the `eval_index_expr` fix this mirrors.
- #2139 -- the deferred `resolve_index_expr` write-path fix; `resolve_slice_expr` likely
  belongs there too once confirmed.

newhoggy:
## Summary

- `E[S:T]` compiles as `S as $s | T as $t | E | .[$s:$t]` — jq re-runs `E` fresh for every `(s, e)` pair the bound generators produce. `eval_slice_expr` (both `src/jq/eval.rs` and its `eval_generic.rs` CLI-dispatch sibling) evaluated `E` once upfront and reused it across every pair instead, so a side effect in `E` (`stderr`, `input`) fired once total rather than once per pair.
- Mirrors #2032/#2142's identical, already-shipped fix for `eval_index_expr`, one nesting level deeper (`S` outer, `T` middle, `E` inner). Restructures both `eval_slice_expr` implementations to evaluate the target inside the `(s, e)` loop.
- `resolve_slice_expr` (the `path()`/write-path sibling, target still evaluated once) stays out of scope, per the issue's own scoping note — tracked separately as #2139.

**Code review** (10-agent multi-pass, independently cross-verified against the pinned jq 1.7.1 oracle) found and this PR now also fixes:
- A later `(s, e)` pair's/target's own slice-application error was discarding the `out` accumulated from *earlier* pairs — newly reachable because of this PR's own restructuring (before it, `out` was provably always empty at that point). Fixed by folding `out` through every escape path, mirroring `eval_index_expr`'s identical treatment.
- Reinstated a `starts.len() * ends.len()` upfront reservation hint (both factors are still statically known before the loop) by reusing the existing `try_reserve_product` helper, recovering the single-allocation fast path for the common one-output-per-pair case instead of paying amortized-doubling reallocation costs on every slice query.
- Factored the repeated "reserve or escape with `out` folded in" shape into a small local `escape!` macro in each function, mirroring `eval_index_expr`'s own `escape_with_prefix!`/`escape_generic!` macros.

Two **separate, pre-existing** gaps surfaced during review (confirmed live on `main`, unrelated to this PR's own scope) were filed as follow-ups rather than bundled in:
- #2225 — the `end` bound (T) isn't re-evaluated per start value (S), unlike jq's own desugaring.
- #2226 — `E`'s own within-generator partial output is discarded on its own mid-stream error, identically present in the already-merged `eval_index_expr`.

## Test plan

- [x] Live-verified against pinned oracle jq 1.7.1 (`/usr/bin/jq`): issue's own `stderr`-count repro now matches exactly (8 writes, was 2).
- [x] Live-verified partial-output preservation on a later-pair target error (`(input)[(0,1):(2,3)]` with only 2 stdin lines) and on a later-pair/target slice-application error (`(.[])[(0,1):(2,3)]` on `[[9,9,9,9],1]`, `([1,2],5)[(0,1):(1,2)]`) — succinctly now matches jq's exact stdout/stderr/exit-code triple in both cases.
- [x] Live-verified an empty-for-one-pair generator target (`inputs[(0,1):(1,2)]`) does not short-circuit other pairs, matching jq.
- [x] Added 5 new tests: 4 CLI-level (`tests/jq_cli_tests.rs`, exercising the actual `eval_generic.rs` dispatch path) covering re-evaluation count, no-short-circuit-on-empty-pair, later-pair-error output preservation (both `Targets` arms), and the `eval.rs`-dispatch-only `input` reachability note; 1 direct-library unit test (`src/jq/eval.rs`) pinning `eval::eval_slice_expr`'s value-correctness post-refactor.
- [x] `cargo test --features cli,simd,regex,serde` (full suite): 3329 lib + 1283 jq_cli (+5 new) + 1367 yq_cli, 0 failed
- [x] `cargo clippy --all-targets --features std,simd,serde,cli,regex,bench-runner,large-tests,mmap-tests -- -D warnings` (avoids the `scalar-yaml` blind spot)
- [x] `cargo build --no-default-features` / `cargo test --no-default-features` (no_std)
- [x] `cargo doc --no-deps --all-features` with `RUSTDOCFLAGS=-D warnings`
- [x] `cargo fmt --check`
- [x] Spot-checked `succinctly yq` mode too (shared evaluator) — same per-pair re-evaluation behavior, no regressions.
- [x] Patch coverage: **91.43%** (64/70 new lines). The 6 remaining uncovered lines are all reservation-failure/invariant-assertion arms requiring either astronomical-scale generator output (which would exhaust memory constructing the *inputs* long before the check could run — the same class `try_reserve_product`'s own unit tests bypass by calling it directly rather than through real generators) or a provably-unreachable `unreachable!()` invariant, matching the coverage profile the existing `eval_index_expr` precedent (#2032/#2142) already accepts for the identical shape.
- [x] All 30 CI checks green.

Fixes #2143

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -1267,18 +1267,25 @@ results one at a time instead of pre-allocating a single buffer, so it just keep
 output rather than answering promptly or refusing.
 
 Both `eval_index_expr` arms (#2032/#2142) and, since #2143, `eval_slice_expr` (in both
-files) have since moved off that single upfront product: their own target (the left side
-of `.[$keys]`/`.[$s:$e]`) is now re-evaluated once per key/`(s, e)` pair rather than once
-overall (see this doc's own no-longer-applicable earlier framing corrected — target length
-can vary per key/pair now, so one upfront product is no longer even computable), so each
-reserves incrementally instead, via a per-key/-pair `Vec::try_reserve` against
-`cannot_reserve_cross_product`'s identical error. The refusal guarantee this section
-describes is unchanged by that restructuring — every push remains behind a fallible
-reservation, so the failure mode stays "clean refusal," never a panic — only the moment the
-check runs (once upfront vs. incrementally per key/pair) and the factor(s) named in the
-error message changed. `resolve_index_expr`/`resolve_slice_expr` (the `path()`/write-path
-siblings, target still evaluated once — tracked separately as #2139) are the two sites that
-still call `try_reserve_product` directly with the original multi-factor product.
+files) have since moved off that single *full* upfront product: their own target (the
+left side of `.[$keys]`/`.[$s:$e]`) is now re-evaluated once per key/`(s, e)` pair rather
+than once overall (see this doc's own no-longer-applicable earlier framing corrected —
+target length can vary per key/pair now, so a product including it is no longer even
+computable), so each also reserves incrementally per key/pair via `Vec::try_reserve`
+against `cannot_reserve_cross_product`'s identical error, regardless of the target's own
+length. The two functions differ on what's reserved *before* that incremental loop even
+starts: `eval_index_expr` reserves nothing upfront (purely incremental from an empty
+`Vec`), while `eval_slice_expr` reserves a `starts.len() * ends.len()` baseline first —
+both factors are still fully known before the loop, so this recovers a single allocation
+for the common one-output-per-pair case instead of paying amortized-doubling
+reallocation/copy costs on every slice query; an overflowing product just skips the hint
+and falls through to the purely-incremental path, same as `eval_index_expr` always takes.
+The refusal guarantee this section describes is unchanged by any of this — every push
+remains behind a fallible reservation, so the failure mode stays "clean refusal," never a
+panic — only the moment(s) a check runs and the factor(s) named in the error message
+changed. `resolve_index_expr`/`resolve_slice_expr` (the `path()`/write-path siblings,
+target still evaluated once — tracked separately as #2139) are the two sites that still
+call `try_reserve_product` directly with the original multi-factor product.
 
 Unlike `s * n` above, this is symmetric across both modes rather than a yq-specific
 divergence to record — but not because a live check for a yq-side cap came back empty.
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -1275,14 +1275,20 @@ computable), so each also reserves incrementally per key/pair via `Vec::try_rese
 against `cannot_reserve_cross_product`'s identical error, regardless of the target's own
 length. The two functions differ on what's reserved *before* that incremental loop even
 starts: `eval_index_expr` reserves nothing upfront (purely incremental from an empty
-`Vec`), while `eval_slice_expr` reserves a `starts.len() * ends.len()` baseline first —
-both factors are still fully known before the loop, so this recovers a single allocation
-for the common one-output-per-pair case instead of paying amortized-doubling
-reallocation/copy costs on every slice query; an overflowing product just skips the hint
-and falls through to the purely-incremental path, same as `eval_index_expr` always takes.
-The refusal guarantee this section describes is unchanged by any of this — every push
-remains behind a fallible reservation, so the failure mode stays "clean refusal," never a
-panic — only the moment(s) a check runs and the factor(s) named in the error message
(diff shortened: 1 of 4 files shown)


newhoggy:
Code review surfaced two pre-existing gaps in the same code, both out of scope for this fix (re-evaluation count) — filed separately rather than bundled here:

- #2225 — the \`end\`/T bound isn't re-evaluated per \`s\`, unlike jq's own \`S as \$s | T as \$t | E | .[\$s:\$t]\` desugaring (confirmed pre-existing on \`main\`, unrelated to this PR).
- #2226 — \`E\`'s own within-generator partial output is discarded on its own mid-stream error, identically present in the already-merged \`eval_index_expr\` (#2032/#2142), not something this PR introduces or worsens.

This PR does fix a related, narrower gap that *was* newly reachable because of its own restructuring: a later \`(s, e)\` pair's/target's slice-application error discarding the \`out\` accumulated from *earlier* pairs (distinct from #2226's within-pair case) — see the latest commit.

Issue #2227, opened 2026-09-02
jq: audit the remaining if-optional sites #2212 found but didn't verify inside eval_stage_with_path_context


newhoggy:
Follow-up from #2212 (fixed narrowly in PR TBD).

## What #2212's own PR documented

#2212 established (via an `assert!(!optional, ...)` probe plus an independent `llvm-cov` patch-coverage diff) that `optional` is provably always `false` at every call site reachable purely through `eval_stage_with_path_context`'s own top-level `optional` parameter, following #2073's removal of `Expr::Optional`'s combined-with-`rest` fallback (the sole former producer of `true` there).

The fix PR documented this at the sites independently traced and confirmed:
- `catch_error_under_optional`'s own doc comment (the shared `(atomic, optional)` dispatch used by `continue_rest_with_context`/`continue_rest_with_paths`/`continue_rest_with_fresh_root`)
- `Builtin::Map`'s arm (both its `_ if optional` guard and its `catch_error_under_optional(..., atomic: true)` call)
- `Expr::Iterate`'s arm (same two-site shape, `atomic: false`)
- `Builtin::ParentN`'s `classify_parent_n` error arm
- `binary_fanout_each` (outside the path-context evaluator entirely, but reachable only through the same now-removed leak per the issue's own analysis)

## What's left unverified

A grep for `if optional`/`optional =>`/`optional {` across the full `eval_stage_with_path_context` function body (`src/jq/eval.rs`, ~2065 lines) turned up several more candidates that were **not individually traced** before this follow-up was filed — each needs the same per-site verification (confirm the `optional` in scope really is the function's own top-level parameter with no other producer, not a shadowed local or a route this analysis missed) before being documented or acted on:

- Two sites in the null/absent-value dispatch near the top of the function (an `Err(_) if optional` and a `if optional { ... }` pair)
- A second `if optional { ... }` block a little further down
- `Expr::Array`'s own `_ if optional`-shaped arm (structurally similar to `Map`'s, but not yet confirmed to have no other reachable producer)
- Two sites inside `Expr::StringInterpolation`'s non-yq branch (`build_string_parts_with_context`'s own `Err(Control::Error(_)) if optional` arm, plus a second nearby check)
- One site much later in the function (a `QueryResult::Error(_) if optional => QueryResult::None` arm)

[...]

PR #2228, opened 2026-09-02
docs(jq): document catch_error_under_optional's dead optional==true cells (#2212)

### LINKED ISSUE #2212: jq: #2073 made `optional` always-false inside the path-context evaluator, retiring #1826/#1869/#1964's machinery with it

newhoggy:
## Summary
- #2073 removed `Expr::Optional`'s combined-with-`rest` fallback in `eval_stage_with_path_context` — the only route that ever forced `optional = true` into the path-context evaluator. #2212 confirmed (via an `assert!(!optional, ...)` probe across the full suite, and independently via an `llvm-cov` patch-coverage diff on #2073's own PR) that this makes `catch_error_under_optional`'s `optional == true` cells, plus several call sites that feed it, currently unreachable.
- #2212's own text lays out two remedies (remove `optional` from the whole call chain, or keep it and document the dead cells) with no clear default — full removal ripples into every affected signature and the issue flags a real risk that the dead surface is wider than its own initial list (confirmed while implementing this: `Expr::Array` and `Expr::StringInterpolation` both have similarly-shaped `if optional` arms not yet individually verified). Chose the lower-risk option: document, don't remove, scoping this PR to exactly the five sites independently re-verified while implementing it.
- Sites documented: `catch_error_under_optional`'s own doc comment (central), `Builtin::Map`'s two-site arm, `Expr::Iterate`'s two-site arm, `Builtin::ParentN`'s error arm, and `binary_fanout_each` (outside the path-context evaluator, reachable only through the same now-removed leak per #2212's own analysis). Also strengthened `test_catch_error_under_optional_full_truth_table_1888`'s doc comment as the sole load-bearing test for the documented cells, per #2212's own suggestion.
- Filed #2227 for the remaining `if optional` sites this pass found via grep but did not individually verify (in `Expr::Array`, `Expr::StringInterpolation`, and a few others) — explicitly not claiming those are confirmed dead.

## Test plan
- [x] Docs-only change (no behavior/logic modification) — verified via `cargo check` before running the full matrix
- [x] `cargo clippy --all-targets --all-features -- -D warnings` (clean build)
- [x] `cargo test --no-fail-fast --features cli,simd,regex,serde` (0 failed)
- [x] `cargo test --no-default-features --verbose` (no_std, 0 failed)

Fixes #2212

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -5028,6 +5028,18 @@ fn binary_fanout_each<'a, W: Clone + AsRef<[u64]>>(
                     // pairing, the outputs already pushed stand either way,
                     // and `optional` decides whether the failure itself
                     // survives.
+                    //
+                    // #2212: this `optional` can currently only be `true` via
+                    // a path-context `?` that used to leak into an operand's
+                    // own evaluation through #2073's now-removed
+                    // combined-with-`rest` fallback (`.a | (.[])? | (key,
+                    // (1 == error("x")))`) -- the plain (non-path-context)
+                    // evaluator's own `Expr::Optional` forwards the *ambient*
+                    // `optional` rather than forcing `true` (#693), so
+                    // nothing else ever delivers one here. Currently dead by
+                    // the same evidence as `catch_error_under_optional`'s own
+                    // doc comment, despite living outside the path-context
+                    // evaluator itself.
                     abort = Some(if optional {
                         Flow::Exhausted
                     } else {
@@ -29716,6 +29728,29 @@ fn accumulate_path_context_step<'a, W: Clone + AsRef<[u64]>>(
 /// time a caller's loop breaks, its own `results` binding is already empty
 /// -- reusing it here instead of the pattern's own `prefix` would silently
 /// keep nothing.
+/// #2212: every caller of this function lives inside the path-context
+/// evaluator's own call graph, and `optional` reaches every one of them
+/// through a chain that bottoms out at `eval_stage_with_path_context`'s own
+/// `optional` parameter. #2073 removed `Expr::Optional`'s combined-with-
+/// `rest` fallback there -- the only route that ever forced a `true` into
+/// that parameter -- so `optional` is `false` at every one of this
+/// function's call sites today, making the `optional == true` cells below
+/// (`(true, true)` and `(false, true)`, plus the bare `Error(e)` arm above)
+/// currently unreachable. Confirmed two independent ways (#2212): an
+/// `assert!(!optional, ...)` probe at `eval_stage_with_path_context`'s own
+/// entry point fires zero times across the full suite (it fires 5 times on
+/// pre-#2073 `main`), and `llvm-cov`'s patch-coverage diff on the #2073 PR
+/// itself reported these lines losing coverage on otherwise-unchanged code.
+/// Left in place rather than removed -- deleting `optional` from this
+/// function's signature (and from every caller's own signature in turn)
+/// would ripple well beyond this one function, and a future path-context
+/// arm could legitimately reintroduce a real producer of `true` the same
+/// way #2073's fallback used to be one. `test_catch_error_under_optional_full_truth_table_1888`
+/// is the sole test exercising these cells now that #1826/#1869/#1964's own
+/// end-to-end tests reach the `optional == false` branch either through
+/// #2073's new hard-error behavior or through `Expr::Optional`'s own
+/// structural catch instead -- treat it as this function's spec for the
+/// `optional == true` column, not just a regression pin.
 fn catch_error_under_optional<W>(
     stopped: QueryResult<'_, W>,
     optional: bool,
@@ -30929,7 +30964,8 @@ fn eval_stage_with_path_context<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
             Ok(Some(ref v)) => match classify_parent_n::<S>(v, current_path.len()) {
                 Ok(n) => n,
                 // `?` swallows only a genuine error; a halt always escapes
-                // (#791).
+                // (#791). #2212: currently unreachable with `optional ==
+                // true` -- see `catch_error_under_optional`'s doc comment.
                 Err(_) if optional => return QueryResult::None,
                 Err(e) => return QueryResult::Error(e),
             },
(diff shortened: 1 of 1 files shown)


newhoggy:
Closing without merging — code review (4-agent fan-out) caught that this PR's entire premise doesn't hold on current `main`.

#2212 was filed against the (still-open, unmerged) #2073 fix branch `issue-2073-optional-suppresses-later-error` (PR #2217, `mergedAt: null` as of this check). This #2212 doc branch was cut from `main` at c201c351b, which does **not** include #2217's commit `be12487a2` — so `eval_stage_with_path_context`'s `Expr::Optional` arm still has the pre-#2073 combined-with-`rest` fallback (`src/jq/eval.rs:31354`, forcing `optional=true`), and every claim in this PR that those cells are 'currently unreachable' is false on `main` today.

Live-verified against a real build of unmodified `main`:
```
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a | (.[])? | (key, (1 == error("x")))'
[0]
```
Only the first element's `key` prints; the rest of the stream is silently dropped — that's the live pre-#2073 bug itself, reachable today, not dead code.

One review agent also caught a second, independent defect in the wording I copied from #2212's own issue text into my `binary_fanout_each` comment: the repro `(key, (1 == error("x")))` uses `==` (Compare), whose `apply_compare_op` returns a bare, infallible `bool` — it can never reach a failing `combine` regardless of #2073's status. The repro needs an arithmetic op (`+`/`-`/`*`/`/`/`%`) to actually exercise `binary_fanout_each`'s error arm, e.g. `(key, (1 + "x"))`.

Branch/worktree left in place (not deleted) for whoever resumes this once #2217 actually merges — at that point this needs a rebase onto the merged #2073 fix, a corrected repro in the `binary_fanout_each` comment, and fresh re-verification that the reachability argument holds post-rebase (the #2073 fix landing might also change exactly which sites are affected).

Issue #2036, opened 2026-08-31
jq: a user def cannot shadow a builtin — the builtin silently wins, producing a wrong value with no error


newhoggy:
Found while verifying #1473. **This one produces a silently wrong value, not an error** — which makes it more serious than the exit-code and wording divergences #1473 closed.

## The divergence

Real jq lets a user `def` shadow a builtin of the same name and arity. succinctly's parser lowers a recognised builtin to a typed `Builtin::*` variant *at parse time*, before any `def` has been seen, so the user's definition is parsed, bound, and then never consulted — the builtin runs instead. All five verified against jq 1.7.1 (`/usr/bin/jq`) and the `main` binary at `3049fa27e`:

| filter | jq 1.7.1 | succinctly |
|---|---|---|
| `def length: 1; length` | `1` | **`0`** |
| `[1,2,3] \| def length: 999; length` | `999` | **`3`** |
| `def map(f): "shadowed"; [1,2] \| map(.)` | `"shadowed"` | **`[1,2]`** |
| `def not: "shadowed"; false \| not` | `"shadowed"` | **`true`** |
| `def first(f): "shadowed"; first(1,2)` | `"shadowed"` | **`1`** |

Every row exits 0 in both tools. There is no diagnostic anywhere — the filter simply computes something else.

## Why it matters

Shadowing a builtin is a real jq idiom, not a curiosity: it is how you patch a builtin's behaviour for one program, and how `~/.jq` and library modules override defaults. A jq script that relies on it produces wrong output here with no signal at all.

It is also the shape of bug that is hardest to notice downstream: the `def` parses, the program runs, the exit code is 0.

## Root cause

`src/jq/parser.rs` resolves builtin names structurally — `length` becomes `Builtin::Length` the moment it is lexed. `Expr::FuncDef`/`Expr::FuncCall` never see the name, so neither `expand_func_calls` nor #1473's new `jq::resolve_func_calls` can intervene: by the time either runs, the call site is already a `Builtin` node with no name attached.

Note this is the *mirror image* of the property that made #1473 cheap. There, "recognised builtins never appear as `Expr::FuncCall`" is exactly what let the resolution pass be a simple check with no false positives. Here, the same fact is the bug. Any fix must not reintroduce false positives into that pass.

## Fix sketch (not small)

The parser would need to stop committing to a `Builtin` variant for a name that a `def` in scope could rebind — which means either:

[...]

Issue #1473, opened 2026-08-22
jq: undefined-function/arity-mismatch errors are lazy, catchable, runtime-scoped (exit 5) where real jq resolves calls at compile time (exit 3, uncatchable)


newhoggy:
## Context

Found during #1376's code review (arity-based `def` overloading). Multiple independent review passes converged on this, live-verified against both a pre-#1376 and post-#1376 build, confirming it's a **pre-existing** divergence, not introduced by that PR — but that PR is the one that reshaped this exact code path (`expand_func_calls`'s `FuncCall` arm, `eval_func_call`), so it's the natural place to record and track it, per this repo's ADR-0018 rule that every jq-mode divergence must be documented.

## The divergence

Real jq resolves every function call **at compile time**, before any input is read or any code runs. A call to an undefined function or an undefined arity of an existing function is a **compile error**: unconditional, unaffected by whether the branch containing it is ever reached, and **not catchable** by `try`/`catch` or `?` (since compilation fails before evaluation — and thus before `try` — ever begins). Exit code **3**.

`succinctly jq` resolves function calls via `expand_func_calls`'s static AST substitution, which runs once per `def` scope but produces an **`Expr::Error` AST node** for an unresolvable call rather than failing immediately — that node is then evaluated **lazily**, exactly like any other expression. This means:

- **Laziness**: an arity-mismatched call inside a never-taken branch never errors at all.
  ```
  $ jq -n 'def f(x): x; if false then f(1;2;3) else 1 end'
  jq: error: f/3 is not defined at <top-level>, line 1:   (compile error, exit 3, unconditional)
  $ succinctly jq -n 'def f(x): x; if false then f(1;2;3) else 1 end'
  1                                                        (exit 0 -- the branch is just never reached)
  ```
- **Catchability**: succinctly's version is swallowable by `try`/`?`; jq's isn't (it never gets far enough to run `try`).
  ```
  $ jq -n 'def f(x): x; try f(1;2) catch "caught"'
  jq: error: f/2 is not defined ...                        (exit 3, try never runs)
  $ succinctly jq -n 'def f(x): x; try f(1;2) catch "caught"'
  "caught"                                                  (exit 0)
  ```
- **Exit code**: succinctly's `EvalError` path gives exit 5 (its general runtime-error exit code) where jq gives exit 3 (compile error) — a different pairing from the already-tracked #355 (which is about succinctly giving exit 0 where jq gives 5; this is 5 vs 3, a distinct mismatch).
[...]

PR #2233, opened 2026-09-02
fix(jq): report zero-arity builtins called with args as a name/arity error

### LINKED ISSUE #2110: jq: a 0-arity builtin followed by '(' is a parser syntax error, not jq's compile-time "X/1 is not defined"
**Severity: Low**

## Summary

Noticed while fixing #2035 (`modulemeta`'s inverted arity). Real jq resolves
`builtin_name(args)` for a *zero*-arity builtin (calling it with the wrong
arity) as a compile-time name/arity resolution error: `X/1 is not defined`,
exit 3 — the same class of error #1473 gave succinctly for genuinely
undefined names.

succinctly instead rejects it earlier, in the raw tokenizer/parser, before
any name/arity resolution runs: the ~126 zero-arity builtins (`length`,
`not`, `env`, `modulemeta`, ...) are recognized structurally at the keyword
level and returned immediately as a complete `Builtin::*` node, so a
following `(` is just an unexpected stray token to the surrounding grammar.

## Repro

```console
$ /usr/bin/jq -n 'length(1)'
jq: error: length/1 is not defined at <top-level>, line 1:
length(1)
jq: 1 compile error

$ succinctly jq -n 'length(1)'
jq: compile error: parse error at position 6: unexpected character '('
```

Same shape for `not(1)`, `modulemeta("x")`, and presumably any other
zero-arity builtin followed by `(`. Confirmed against the pinned oracle
(`/usr/bin/jq`, jq-1.7.1).

## Root cause

`src/jq/parser.rs`'s builtin-keyword recognizer (the same function #2035
touched) returns `Ok(Some(Builtin::X))` the moment it matches a zero-arity
keyword, without checking whether the keyword is immediately followed by
`(`. There's no path from there into `src/jq/resolve.rs`'s #1473 name/arity
resolution machinery for these names, because they never become an
`Expr::FuncCall` in the first place — that's exactly the property that made
#1473 cheap for these builtins in the first place (see #2036's own note on
this being the mirror image of #2035/#2036's problem), and it's the same
property that makes this a parser-level, not resolution-level, divergence.

## Why this is separate from #2035

#2035 was about `modulemeta` specifically having the *wrong* arity
requirement (parser demanded parens jq doesn't have at all). This issue is
about the *error shape* when any zero-arity builtin is called with the
wrong arity in the first place — it existed before #2035 and is unaffected
by it; #2035's fix only makes `modulemeta` consistent with this
pre-existing, broader pattern rather than introducing a new special case.

## Suggested fix direction

Either:
1. Have the zero-arity keyword recognizer check for a following `(` and,
   if present, route into the same "unresolved call" reporting path
   #1473/#2037 built for genuinely-undefined names (reporting `X/1 is not
   defined` with the resolved arity), rather than returning early; or
2. Accept the divergence as permanent (it's a parse-error-vs-compile-error
   distinction, not a behavioral difference — both are compile-time,
   non-catchable, exit-nonzero failures) and record it in
   docs/compliance/jq/limitations.md per ADR-0018, since matching exactly
   would require restructuring how ~126 builtins are recognized for a
   cosmetic error-message difference only reachable via a malformed
   program.

Low severity: this only fires when a program is already wrong (calling a
builtin with an arity it's never had, in either tool), so no valid jq
program's *output* is affected — only the compile-error text/kind for an
already-invalid one.

newhoggy:
## Summary

Real jq resolves `length(1)`, `not(1)`, `null(1)` etc. (a zero-arity builtin/literal
called with the wrong arity) as a compile-time name/arity resolution error
(`length/1 is not defined`, exit 3) — the same class #1473/#2037 already give a
genuinely-undefined name. succinctly's zero-arity keyword recognizers instead returned
their AST node the instant they matched, so a following `(` reached the surrounding
grammar as a raw parser rejection.

```console
$ /usr/bin/jq -n 'length(1)'
jq: error: length/1 is not defined at <top-level>, line 1:
length(1)
jq: 1 compile error

$ succinctly jq -n 'length(1)'          # before this PR
jq: compile error: parse error at position 6: unexpected character '('

$ succinctly jq -n 'length(1)'          # after this PR
jq: error: length/1 is not defined at <top-level>, line 1:
length(1)
jq: 1 compile error
```

## Fix

Rather than touching all ~126 individual `matches_keyword`/`consume_keyword` branches
in `try_parse_builtin`, this adds one helper (`zero_arity_or_wrong_arity_call`) called
from the small number of actual call sites (`null`/`true`/`false`/`not`, and
`try_parse_builtin`'s single call site): peek past optional whitespace for a following
`(`, and if present, rewind to before the keyword was consumed and re-parse through
`parse_func_call_or_error` — the same path a genuinely-undefined name already takes, so
the existing #1473 resolver machinery reports the "X/N is not defined" error without any
new logic there.

`if`/`try`/`reduce`/etc. are unaffected — they're syntax keywords, not
`Expr::FuncCall`-shaped in real jq either (`if(1)` is a genuine syntax error there too,
verified live: "unexpected end of file", not "if/1 is not defined") — so they don't
route through the new helper.

## Verification

Live-verified against the pinned oracle (`/usr/bin/jq` 1.7.1) for `length`, `not`,
`null`, `true`, `false`, `modulemeta`, plus whitespace-before-paren (`length (1)`), and
that `if(1)` correctly stays a syntax error. Confirmed ordinary usage (no parens),
postfix chaining after a bare builtin (`keys[0]`), and builtin-shadowing (`def length:
99; length`, a separate already-tracked #2036 gap) are all unaffected.

Fixes #2110

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, all 57 test binaries pass)
- [x] `cargo fmt --check`
- [x] Live-verified against `/usr/bin/jq` (pinned 1.7.1)

### DIFF
--- a/src/jq/parser.rs
+++ b/src/jq/parser.rs
@@ -1397,18 +1397,25 @@ impl<'a> Parser<'a> {
 
             // Keywords: null, true, false, not, if, try, error, reduce, foreach, etc.
             Some(c) if c.is_alphabetic() => {
+                let keyword_start = self.pos;
                 if self.matches_keyword("null") {
                     self.consume_keyword("null");
-                    Ok(Expr::Literal(Literal::Null))
+                    self.zero_arity_or_wrong_arity_call(keyword_start, Expr::Literal(Literal::Null))
                 } else if self.matches_keyword("true") {
                     self.consume_keyword("true");
-                    Ok(Expr::Literal(Literal::Bool(true)))
+                    self.zero_arity_or_wrong_arity_call(
+                        keyword_start,
+                        Expr::Literal(Literal::Bool(true)),
+                    )
                 } else if self.matches_keyword("false") {
                     self.consume_keyword("false");
-                    Ok(Expr::Literal(Literal::Bool(false)))
+                    self.zero_arity_or_wrong_arity_call(
+                        keyword_start,
+                        Expr::Literal(Literal::Bool(false)),
+                    )
                 } else if self.matches_keyword("not") {
                     self.consume_keyword("not");
-                    Ok(Expr::Not)
+                    self.zero_arity_or_wrong_arity_call(keyword_start, Expr::Not)
                 } else if self.matches_keyword("if") {
                     self.parse_if_expr()
                 } else if self.matches_keyword("try") {
@@ -1443,8 +1450,14 @@ impl<'a> Parser<'a> {
                 } else if self.matches_keyword("break") {
                     self.parse_break_expr()
                 } else if let Some(builtin) = self.try_parse_builtin()? {
-                    // Allow postfix operations after builtins (e.g., env.PATH, keys[0])
-                    self.parse_postfix(Expr::Builtin(builtin))
+                    // #2110: a following `(` means this was actually a
+                    // wrong-arity call, not a bare builtin reference --
+                    // `zero_arity_or_wrong_arity_call` rewinds and re-parses
+                    // it as one instead of falling into `parse_postfix`,
+                    // which would just reject the `(` as a stray token.
+                    self.zero_arity_or_wrong_arity_call(keyword_start, Expr::Builtin(builtin))
+                        // Allow postfix operations after builtins (e.g., env.PATH, keys[0])
+                        .and_then(|expr| self.parse_postfix(expr))
                 } else {
                     // Phase 9: Try to parse as function call
                     self.parse_func_call_or_error()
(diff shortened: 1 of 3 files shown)


newhoggy:
## Code review found 2 real regressions, both fixed

Full multi-angle review found the round-1 fix introduced two real regressions (both
independently confirmed by multiple finders):

1. **Missing postfix chaining**: `null`/`true`/`false`/`not`'s call sites didn't apply
   postfix parsing after the rewind, unlike the sibling builtin call site.
   `null(1).foo` regressed to a raw parse error instead of "null/1 is not defined" --
   and more seriously, `def null(x): {foo:x}; null(1).foo` (a *valid* jq program, real
   jq: `1`) failed to parse at all. **Fixed**: the helper now applies postfix uniformly
   on every path, closing this and (as a side effect) a pre-existing `null.foo` gap too.

2. **Empty-parens regression**: `length()`/`not()`/`keys()` etc. regressed from a
   compile-time parse error (exit 3) to a confusing runtime "undefined function" error
   (exit 5), because `resolve.rs`'s builtin roster lists these names at arity 0 for an
   unrelated reason (#1473's "real jq builtins succinctly doesn't implement" tracking).
   Real jq treats `NAME()` as a genuine syntax error for *any* name, builtin or not
   (confirmed live: `def f: 1; f()` is also "unexpected ')'" there) -- not a name/arity
   resolution case at all. **Fixed**: the empty-parens shape is now excluded from the
   rewind entirely, restoring the pre-#2110 behavior exactly (not a new fix for this
   specific shape, but not a new divergence either).

Also gated the whole rewind to **jq mode only**, since real yq (v4.53.3, live-verified)
has no name/arity resolution vocabulary at all -- `null(1)`/`length(1)` both give the
identical generic "bad expression, please check expression syntax" there. jq's own
wording would have been a new, jq-flavored divergence in yq mode; yq mode now keeps its
pre-#2110 generic parse-error rejection byte-for-byte unchanged.

Added regression tests for all of the above (postfix chaining, the def-shadowing
correctness case, empty-parens staying compile-time, yq-mode non-regression).

**Deliberately left out of scope** (same symptom class, not a regression -- unaddressed
in both rounds): `has`/`select`/`env`/other required-arg builtins, and
`range`/`limit`/`until`/`while`/`repeat`/`first`/`last`, all parse their own arguments
inline and never reach the shared fix. Closing those is a materially larger change than
this issue's own scope. Filed as #2237.

Full verification suite (test suite, all 3 clippy legs, fmt) green; live-verified
against both pinned oracles (`/usr/bin/jq` 1.7.1, Homebrew `yq` v4.53.3).

Issue #1171, opened 2026-08-17
jq: a leading-dot number literal (.5, .5e-400) in document input silently produces no output instead of an error or a parsed value


newhoggy:
## Summary

Found while implementing #1099 (unrelated to that issue's own scope — that's about mantissa *formatting* for values that already parse; this is about *parsing* rejecting valid-per-jq input silently instead of erroring).

Real jq's own JSON reader is more lenient than strict RFC 8259 and accepts a leading-dot number spelling (no digit before the decimal point):

```
$ echo '.5' | jq -c '.'
0.5
$ echo '.5e-400' | jq -c '.'
5E-401
```

succinctly instead produces **no output at all**, with exit code `0` (success) and no error message:

```
$ echo '.5' | succinctly jq -c '.'
$ echo "exit: $?"
exit: 0
```

## Why this matters

Silent empty output on a real-but-invalid-per-strict-JSON document is a worse failure mode than either erroring or accepting it — a user piping this through a script sees nothing where real jq would see a parsed value, with no diagnostic pointing at why.

## Scope note

Confirmed pre-existing on `main`, reproduces identically before and after #1099's fix (which only touches number *formatting*, not parsing/validation). Not yet investigated further — could be a genuine RFC-8259-strictness tradeoff succinctly makes deliberately elsewhere (see `docs/architecture/semi-indexing.md`'s validation tradeoff discussion) that just happens to fail silently instead of with a clear error here, or a straightforward parser gap. Filing to make the gap trackable.

newhoggy:
Investigated further before implementing — the actual scope is broader than this issue's own title suggests, and it's **not** number-specific. This is a lazy-path-specific gap in the default `succinctly jq` CLI input parsing (the fast/semi-indexed path used whenever `--validate`/`-s`/`-R`/`--seq` aren't passed) — confirmed the `--validate`/`json validate` path and the `-s` (slurp, serde_json-backed) path both correctly reject every input below.

## Three separate defects, one shared symptom (silent wrong output instead of a parse error)

**1. Truncated/unparseable top-level values silently vanish**, not just `.5`:
```
$ echo '[1,2,' | succinctly jq -c '.'
(empty output, exit 0)          # real jq: "Unfinished JSON term at EOF", exit 5
```
[...]

PR #2239, opened 2026-09-02
fix(jq): preserve a trailing-dot-before-exponent number literal's spelling

### LINKED ISSUE #2220: jq: a trailing-dot mantissa with an exponent (`1.e999`) loses its source literal, diverging from jq
## Summary

A number span with a **trailing dot before an exponent** (`1.e999`, `-1.e5`) loses
its source literal in `OwnedValue::from_number_bytes`, where real jq preserves the
spelling. The value is degraded to a plain `f64`, so an out-of-range exponent
clamps to the `JqSemantics` infinity stand-in and a small exponent underflows to
`0` — jq prints the literal in both cases.

## Repro (jq 1.7.1, `/usr/bin/jq`)

| input      | jq 1.7.1   | `succinctly jq` |
|------------|------------|-----------------|
| `[1.]`     | `[1]`      | `[1]` ✅        |
| `[1.5e3]`  | `[1.5E+3]` | `[1.5E+3]` ✅   |
| `[1.e999]` | `[1E+999]` | `[1.7976931348623157e+308]` ❌ |
| `[-1.e999]`| `[-1E+999]`| `[-1.7976931348623157e+308]` ❌ |
| `[1.e-999]`| `[1E-999]` | `[0]` ❌        |

```console
$ printf '[1.e999]' | /usr/bin/jq -c .
[1E+999]
$ printf '[1.e999]' | succinctly jq -c .
[1.7976931348623157e+308]
```

## Cause

`OwnedValue::from_number_bytes` (`src/jq/value.rs`) has two leniency escapes that
keep the source spelling as a `NumberLiteral`:

- a **leading** dot — insert `0` after any `-` and re-check `is_valid_number`
  (`.5` -> keeps `.5`, #1171)
- redundant **leading zeros** — `strip_redundant_leading_zeros` then re-check
  (`007e5` -> keeps `007e5`, #1149)

There is no equivalent for a **trailing** dot in the mantissa. `1.e999` fails
`is_valid_number`, matches neither escape, and falls through to
`parse_i64_or_f64`, which returns `Float(inf)` — the literal is gone before any
printer sees it.

## Scope

Not specific to any output path: the M2 cursor streamer
(`write_json_number`, `src/json/light.rs`) and the DOM printer
(`print_json`/`format_raw_number`, `src/bin/succinctly/jq_runner.rs`) produce
byte-identical output here, and `succinctly jq --unbuffered` diverges the same
way. The bug predates #1576/#2207 and lives entirely in the shared conversion.

## Suggested fix

Add a third escape alongside the two above: insert `0` *after* a mantissa's
trailing `.` and re-check `is_valid_number` (`1.e999` -> `1.0e999` is valid), then
materialize the **original** bytes as the literal, exactly as the leading-dot and
leading-zero escapes already do.

## Coverage note

`write_json_number`'s non-finite arm
(`out.write_str(nonfinite_display_string::<JqSemantics>(f))`) is *only* reachable
through this bug: every strictly-valid overflowing literal (`1e999`) keeps its
spelling and reformats instead. `test_stream_json_number_invalid_span_float_fallback_1576`
(`src/json/light.rs`, added in #2207) pins the current output and points here;
that assertion should flip to jq's `1E+999` when this is fixed, and the arm may
become unreachable.

newhoggy:
## Summary

`OwnedValue::from_number_bytes` had two leniency escapes that preserve a lenient-but-
strictly-invalid number span's source spelling as a `NumberLiteral` (#1171's leading
dot, #1149's redundant leading zeros), but no equivalent for a **trailing dot
immediately before an exponent marker** (`1.e999`, `-1.e5`). That shape fails
`is_valid_number`, matches neither existing escape, and falls through to a lossy `f64`
parse — an out-of-range exponent silently clamps to the `JqSemantics` infinity
stand-in instead of the exact literal real jq prints.

```console
$ printf '[1.e999]' | /usr/bin/jq -c .
[1E+999]
$ printf '[1.e999]' | succinctly jq -c .          # before this PR
[1.7976931348623157e+308]
$ printf '[1.e999]' | succinctly jq -c .          # after this PR
[1E+999]
```

## Fix

Adds a third escape mirroring the existing two: insert `0` right after the trailing
`.` and re-check `is_valid_number` — valid (`1.e999` → `1.0e999`) means the *original*
bytes get materialized as the `NumberLiteral`, same pattern as the leading-dot/
leading-zero cases. A bare trailing dot with **no** exponent (`1.`) is deliberately
left alone: real jq doesn't preserve that spelling either (`[1.]` → `[1]` on both
sides), so the existing lossy fallback already matches jq there.

Not specific to any output path — the M2 cursor streamer and the DOM printer
(`--unbuffered`) produce byte-identical output through the same shared conversion,
live-verified.

## Verification

Live-verified against the pinned oracle (`/usr/bin/jq` 1.7.1) for the issue's own full
repro table, plus additional edge cases (uppercase `E`, negative mantissa, negative
exponent) and confirmed genuinely malformed shapes (`1.5.3`) are unaffected.

## Changes

- `OwnedValue::from_number_bytes` (`src/jq/value.rs`): the new escape.
- `test_stream_json_number_invalid_span_float_fallback_1576` (`src/json/light.rs`),
  which pinned the old, now-fixed `1.e999` behavior as a documented divergence — its
  `1.e999`/`-1.e999` rows now read jq's own `1E+999`/`-1E+999`.
- Direct unit coverage of the new escape (`src/jq/value.rs`) and a CLI-level
  regression test covering the issue's own repro table (`tests/jq_cli_tests.rs`).

Fixes #2220

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, all 57 test binaries pass)
- [x] `cargo fmt --check`
- [x] Live-verified against `/usr/bin/jq` (pinned 1.7.1), both M2 streaming and `--unbuffered` DOM paths

### DIFF
--- a/src/jq/value.rs
+++ b/src/jq/value.rs
@@ -1532,8 +1532,9 @@ impl OwnedValue {
         // also ends up here) still fell all the way through to the lossy
         // `parse_i64_or_f64` below, losing the leading zero *and* any
         // exponent notation *and* any trailing zeros in one go (#1149).
-        if let Some(stripped) = crate::json::validate::strip_redundant_leading_zeros(bytes) {
-            if crate::json::validate::is_valid_number(&stripped) {
+        let zero_stripped = crate::json::validate::strip_redundant_leading_zeros(bytes);
+        if let Some(stripped) = &zero_stripped {
+            if crate::json::validate::is_valid_number(stripped) {
                 return core::str::from_utf8(bytes).map_or(Self::Null, Self::from_number_literal);
             }
         }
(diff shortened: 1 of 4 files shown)


newhoggy:
## Code review found 2 real gaps, both fixed

Full multi-angle review found the round-1 fix had two real gaps (independently
confirmed by multiple finders):

1. **No composition between escapes**: the leading-zero and trailing-dot escapes each
   checked only their own shape against the original bytes. A token with both at once
   (`007.e999`) matched neither, reproducing the exact clamped-infinity bug this PR
   exists to fix. **Fixed**: extracted a shared `has_trailing_dot_before_exponent`
   helper (`src/json/validate.rs`, alongside `strip_redundant_leading_zeros`) and
   checked it against the leading-zero-stripped form when one exists.

2. **Missing sibling fix**: `StandardJson::number_literal()` (`src/json/light.rs`) —
   the conversion `-S`/`--argjson`/`.json`-file input route through instead of
   `from_number_bytes` — never got this PR's own trailing-dot escape, so
   `succinctly jq -cS '.'` on `{"a":1.e999}` still printed the clamped stand-in,
   directly contradicting the PR's own "not specific to any output path" claim.
   **Fixed**: same composed check added there via the same shared helper.

Both verified against the pinned oracle for the compound shape through both output
paths. Also strengthened the existing unit test to match on the stored literal text
explicitly (`OwnedValue`'s `PartialEq` for `NumberLiteral` ignores the text, so the
original `assert_eq!` form couldn't actually catch a wrong/mangled literal) and added
composition + `-S` regression coverage.

**Two further gaps found, confirmed pre-existing and unrelated** (reproduce
identically for an already-valid `1e999` with no trailing dot at all): `tostring`/
`@json` lose an overflow literal's spelling via array/nested navigation (a different,
untouched materializer path), and `--argjson` rejects every leniency escape's accepted
spelling outright. Filed as #2240 rather than expanding this PR's scope.

Full verification suite (test suite, all 3 clippy legs, fmt) green; live-verified
against the pinned oracle (`/usr/bin/jq` 1.7.1).

Issue #1218, opened 2026-08-17
crate: multiple independent hand-rolled JSON number-token scanners with different, undocumented-against-each-other grammars


newhoggy:
## Summary

Found during code review of PR #1217 (#1154), corroborated independently by several review passes. The crate now has (at least) four independent "find the boundaries of a JSON-ish number token" implementations, each with a genuinely different strictness grammar, none cross-referenced against the others:

1. `src/json/light.rs`'s `number_literal_end` (`pub`) -- strict: rejects a dangling exponent marker outright (`"5e"` -> `None`), accepts a leading-dot fraction with zero integer digits (`".5"`). Documented as "only used for top-level document splitting" (`find_json_values`), with an explicit warning not to reuse it elsewhere.
2. `src/json/light.rs`'s `nested_number_span` (private) -- fully greedy/permissive, defers real validation downstream. Documented as the deliberately-more-lenient sibling of #1 for nested-value materialization.
3. `src/json/simple_light.rs`'s `find_number_end` (private) -- different job again: greedily consumes any `[0-9+\-.eE]` byte with zero grammar validation, backing the separate/legacy `SimpleJsonIndex`.
4. `src/bin/succinctly/jq_runner.rs`'s `find_number_end` (private, added by #1154/PR #1217) -- yet another strictness point: structured optional frac/exp like #1, but lenient on a dangling exponent marker like #2, specifically to match `--argjson`'s leading-zero-repair needs (real jq's own leniency).

## Why this matters

Confirmed via differential fuzzing (PR #1217's own review) that #1 and #4 genuinely disagree on real inputs -- e.g. `"5e"`/`"1E"` (dangling exponent marker): #4 returns `Some` (consumes it leniently, matching real jq and letting a later stricter validator reject the whole string if needed), #1 returns `None` (rejects outright, by design, since #1's own contract requires a top-level document to error on this). Both are *individually* correct for their own call site's documented contract, but nothing enforces or even flags that they're intentionally different — a future correctness fix applied to only one of these functions (e.g. a jq-compat leniency correction, or an `Infinity`/`NaN` handling change) won't propagate to the others, and there's no shared test suite that would catch such a fix going into the wrong one, or missing the others entirely.

## Suggested direction

Not scoped in detail -- flagging for visibility per PR #1217's own review discussion. Possible directions:
[...]

PR #2244, opened 2026-09-02
fix(jq): re-evaluate E[S:T]'s end bound once per start value

### LINKED ISSUE #2225: jq: E[S:T]'s end bound (T) isn't re-evaluated per start value (S), unlike jq's own desugaring
**Severity: Low** (side-effect count only, reachable with an impure end bound expression)

## Summary

`E[S:T]` compiles as `S as $s | T as $t | E | .[$s:$t]` — jq's own desugaring puts `T`
inside `S`'s binding scope, so `T` should be re-evaluated fresh once per value `S`
produces, exactly the same shape #2143 (PR #2223) just fixed for `E` relative to
`(s, e)` pairs. `eval_slice_bound`'s caller in both `eval_slice_expr` implementations
(`src/jq/eval.rs`, `src/jq/eval_generic.rs`) evaluates `end` exactly once, before the
`for s in starts` loop even begins, and reuses it across every `s`.

```console
$ echo '[10,20,30]' | jq -c '[(.)[(0,1):(stderr|2)]]'
[10,20,30][10,20,30][[10,20],[20]]
# stderr fires TWICE -- once per start value (0 and 1)

$ echo '[10,20,30]' | succinctly jq -c '[(.)[(0,1):(stderr|2)]]'
[10,20,30][[10,20],[20]]
# stderr fires only ONCE -- end bound evaluated once, reused across both starts
```

Confirmed this is **pre-existing**, not introduced by #2143/PR #2223: reproduces
identically on `main` before that PR (both before and after, same output).

Confirmed `start` (S) itself is *not* similarly affected — it's the true outermost
binding, so evaluating it once is correct:

```console
$ echo '[10,20,30]' | jq -c '[(.)[(stderr|0,1):(2)]]'
[10,20,30][[10,20],[20]]
$ echo '[10,20,30]' | succinctly jq -c '[(.)[(stderr|0,1):(2)]]'
[10,20,30][[10,20],[20]]
```
(both fire once — matches, no bug here)

## Root cause

`eval_slice_expr` (both files) calls `eval_slice_bound` for `end` exactly once,
immediately after resolving `start`, then reuses the resulting `Vec<Option<i64>>`
across every `s` in the outer loop. Per jq's own desugaring, `T` needs to be
re-evaluated fresh inside each `s` iteration instead.

## Why this is separate from #2143

#2143 (PR #2223) fixed `E`'s re-evaluation relative to `(s, e)` pairs — a different
variable, one nesting level deeper than this one. Fixing `T` relative to `s` requires
its own restructuring (move the `end`-bound evaluation inside the `for s in starts`
loop, with its own escape/prefix-threading), independent of and not overlapping with
that fix's own scope. Bundling it into #2143/PR #2223 would have significantly grown an
already large, already-reviewed diff.

## Suggested fix direction

Move `eval_slice_bound(end, ...)` inside the `for s in starts` loop in both
`eval_slice_expr` implementations, mirroring the same "evaluate fresh, fold running
`out` in as `Partial` on escape" pattern #2143 already established for `target`. Needs
its own live-oracle verification pass for the interaction with `ends_escaped`/`bounds_escape`
(the existing short-circuit logic around an empty `end` stream), since those currently
assume `end` is computed once.

## Refs

Found during code review of PR #2223 (fixing #2143).

newhoggy:
## Summary

`E[S:T]` compiles as `S as $s | T as $t | E | .[$s:$t]` — jq's own desugaring puts `T`
inside `S`'s binding scope, so `T` must be re-evaluated fresh once per value `S`
produces, the same shape #2143 (PR #2223) already fixed for `E` relative to `(s, e)`
pairs one nesting level deeper. Both `eval_slice_expr` implementations
(`src/jq/eval.rs`, `src/jq/eval_generic.rs`) instead evaluated `end` exactly once,
before the loop over `starts` even began, and reused it across every `s`.

```console
$ echo '[10,20,30]' | jq -c '[(.)[(0,1):(stderr|2)]]'
[10,20,30][10,20,30][[10,20],[20]]
# stderr fires TWICE -- once per start value

$ echo '[10,20,30]' | succinctly jq -c '[(.)[(0,1):(stderr|2)]]'   # before this PR
[10,20,30][[10,20],[20]]
# stderr fires only ONCE
```

Confirmed pre-existing, not introduced by #2143/PR #2223 — both files reproduced
identically before that PR too.

## Designing the exact semantics

`T` structurally has no access to `$s` in jq's grammar, so proving/designing the exact
per-`s` behavior needed a stateful `T` (via `input`) to make it vary across
evaluations. Live-verified against jq 1.7.1 before implementing:

- An empty `T` for one particular `s` contributes nothing for that `s` and moves on to
  the next one — **not** a whole-function short-circuit the way an entirely-empty `S`
  is.
- A `T` that produces some values before escaping (error/break/halt) processes those
  values first, then folds the running accumulator in as the escape's `Partial`
  prefix, same as every other per-iteration escape already in these functions.
- A later `s`'s own end-bound error does not discard values an earlier `s` already
  contributed — they survive as the `Partial` prefix.

## Side effect: reservation optimization removed

Moving `end`'s evaluation inside the loop also removes the upfront
`starts.len() * ends.len()` product reservation both functions used as a baseline
allocation optimization (#2143): `ends.len()` is no longer known until each `s`
iteration evaluates `T`, so there's nothing fixed left to reserve before the loop
starts. The per-target `try_reserve` calls already inside the loop still protect every
real allocation — only the baseline optimization is gone, not the overflow protection
itself.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, all 57 test binaries pass, including the full pre-existing #2143 cluster)
- [x] `cargo fmt --check`
- [x] `cargo doc --no-deps --all-features` (RUSTDOCFLAGS=-D warnings)
- [x] Live-verified against `/usr/bin/jq` (pinned 1.7.1) for every edge case above, both jq mode and yq mode

Added CLI-level regression tests mirroring the existing #2143 test cluster's own style
(re-evaluation count, one-empty-`s`-does-not-abort, later-`s`-error-preserves-earlier-
output, ordinary-cross-product regression guard). `eval.rs`'s own DOM-route copy
already had adequate value-correctness coverage via the pre-existing #2143 unit test
(still passing) — per that test's own doc comment, a side-effect/count-based check
isn't meaningful there since the harness has no `stderr`/`input`-equivalent to observe.

Fixes #2225

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -16541,20 +16541,6 @@ fn eval_slice_expr<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
             Some(control) => partial(Vec::new(), control),
         };
     }
-    let (ends, ends_escape) = match eval_slice_bound::<W, S>(end, value.clone(), f64::ceil) {
-        Ok(v) => v,
-        Err(control) => return partial(Vec::new(), control),
-    };
-    let ends_escaped = ends_escape.is_some();
-    let bounds_escape = ends_escape.or(starts_escape);
-    if ends.is_empty() {
-        return match bounds_escape {
-            None => QueryResult::None,
-            Some(control) => partial(Vec::new(), control),
-        };
-    }
-    let starts_len = if ends_escaped { 1 } else { starts.len() };
-    let starts = &starts[..starts_len];
 
     // Borrowed and owned targets are kept apart so the common (borrowed) case
     // never materializes the document — mirrors `eval_index_expr`.
@@ -16563,31 +16549,20 @@ fn eval_slice_expr<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         Owned(Vec<OwnedValue>),
     }
 
-    // S outer, T middle, E inner (point 2 above), with E now genuinely
-    // re-evaluated once per (s, e) pair rather than once overall (#2143).
-    // The result is always owned: slicing constructs a new array/string
-    // (see `eval_single`'s `Expr::Slice` arm), so there is nothing to
-    // borrow except its rare whole-value fast paths, which `to_owned_lossy`
-    // below just copies out of.
+    // S outer, T middle, E inner (point 2 above), with E re-evaluated once
+    // per (s, e) pair (#2143) and T (`end`) now genuinely re-evaluated once
+    // per `s` rather than once overall (#2225) -- T sits inside S's own
+    // binding scope in jq's desugaring, so a fresh `s` gets a fresh `T`
+    // evaluation, the same nesting #2143 already established one level
+    // deeper for E relative to `(s, e)`.
     //
-    // `starts.len() * ends.len()` is still fully known before the loop --
-    // only the per-pair target count varies -- so `try_reserve_product`
-    // reserves that much upfront as a baseline (the common case is exactly
-    // one output per pair), recovering the single upfront allocation the
-    // pre-#2143 code made for that case instead of paying amortized-
-    // doubling reallocation/copy costs on every slice query; both `starts`
-    // and `ends` are already known non-empty here (the `is_empty` checks
-    // above), so this never takes the zero-factor fast-return arm. Reusing
-    // the existing helper -- rather than a bespoke `checked_mul` inline --
-    // keeps this refusal path covered by that function's own existing unit
-    // tests instead of adding new, practically-untestable ones of its own:
-    // constructing real `starts`/`ends` generator output long enough to
-    // organically overflow this product would first exhaust memory
-    // building the *inputs*, long before this reservation could ever run.
-    let mut out: Vec<OwnedValue> = match try_reserve_product(&[starts.len(), ends.len()]) {
-        Ok(out) => out,
-        Err(e) => return QueryResult::Error(e),
-    };
+    // No upfront `starts.len() * ends.len()` reservation baseline anymore:
+    // `ends.len()` is no longer known until each `s` iteration evaluates T,
+    // so there is nothing fixed to reserve before the loop starts. The
+    // per-target `try_reserve` calls inside the loop still protect every
+    // real allocation; only the pre-#2225 upfront-baseline optimization is
+    // gone, not the overflow protection itself.
+    let mut out: Vec<OwnedValue> = Vec::new();
 
     // The shared exit every escape arm below funnels through, so folding
     // the running `out` in as a `Partial` prefix can't drift between arms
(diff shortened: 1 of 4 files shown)


newhoggy:
## Code review found a real, high-severity gap — filed as a follow-up

Six independent finders converged on the same finding: `resolve_slice_expr`
(`src/jq/eval.rs`) — the **path-mode** resolver backing `path(E[S:T])`, direct/compound
assignment, and `del(E[S:T])` — has the identical "T evaluated once overall" bug this
PR fixes for the two *value-mode* `eval_slice_expr` implementations, and was never
touched.

Confirmed live, and **more severe** than the read-mode symptom since it's a write-path
correctness bug, not just a side-effect-count nuance:

```console
$ printf '[10,20,30,40,50,60]\n2\n4\n' | jq -c -n 'input as $doc | $doc | del(.[(0,3):(input)])'
[30,50,60]
$ printf '[10,20,30,40,50,60]\n2\n4\n' | succinctly jq -c -n 'input as $doc | $doc | del(.[(0,3):(input)])'
[30,40,50,60]                                                 # WRONG -- silently corrupts data
```

**Not folded into this PR** — `resolve_slice_expr` is structurally more complex than
the two functions this PR fixed (a three-way escape priority `target > end > start`
vs. two, an early trackability check that reads `ends[0]` before `target` is even
resolved, and `target_branches` computed once for reasons that are legitimately
different from why value-mode's `E` needed the #2143 per-pair fix). A hasty port here
risks a *different*, possibly worse, correctness bug on a write path. Filed as **#2245**
(High severity) with the full repro set, design analysis of why it needs its own pass,
and a suggested fix direction.

## Also addressed from review

- **Coverage gap**: all 4 original tests exercised the same `eval_generic.rs` dispatch
  path via stdin-fed jq mode. Added tests for the other real dispatch path
  (`-n`/`--null-input`, which routes through `eval.rs`'s copy) and yq mode, both
  manually verified correct during review, now with automated coverage.
- Confirmed real yq's own grammar can't reproduce this bug's repro shape at all (its
  lexer rejects multi-valued slice bounds outright, and `input` needs
  `--jq-extensions`) — no yq-specific divergence to document.
- Several style/DRY observations (shared "insert-and-fold-on-escape" combinator across
  6+ call sites spanning multiple issues, duplicated doc-comment text between the two
  fixed files) were raised but left as-is — genuine architectural cleanup opportunities
  beyond this targeted bug fix's scope, consistent with how this codebase has already
  deferred the analogous JSON-number-scanner consolidation (#1218).

Full verification suite (test suite, all 3 clippy legs, fmt, doc build) green;
live-verified against the pinned oracle (`/usr/bin/jq` 1.7.1) throughout.

Issue #1641, opened 2026-08-25
jq: the two streaming routes still swallow a malformed member (.[] and identity on a bareword value) — #1247 Stage 6, jq side


newhoggy:
**Severity: Low** (silent wrong output, but every materializing route over the same document
now raises)

## Summary

#1608 and #1628 closed #1194's structural half everywhere a value is *materialized*. Two
**streaming** routes remain, and they share one cause: the writer has no error channel to raise
into. This is the jq-side counterpart of #1615, which tracks the same Stage 6 gap for yq.

```console
$ echo '{invalid}'    | sjq -c '.[]'        # no output,  exit 0   (jq: parse error, exit 5)
$ echo '{invalid: 1}' | sjq -c '.[]'        # 1,          exit 0   (jq: exit 5)
$ echo '[xyz123]'     | sjq -c .            # [null],     exit 0   (jq: exit 5)

# for contrast, every materializing route over the same documents:
$ echo '[xyz123]'     | sjq -c to_entries   # jq: error … unexpected character, exit 5
$ echo '{invalid}'    | sjq -c 'keys, length, to_entries'   # exit 5
```

Verified against `/usr/bin/jq` 1.7.1 and a `--release --features cli` build at #1628's head.

## Root cause — two sites, one shape

**`.[]` over an object** walks the field list through `LazySource::Values`
(`src/jq/eval_generic.rs`), whose step is `fields.uncons()?`. `uncons` reports "the last child
never got a value" and "no more fields" as the same `None`, and the method returns `Option`, not
`Result` — so the unpaired child is indistinguishable from exhaustion at exactly the point that
could act on it. `DocumentFields::ends_unpaired` (added by #1608) *can* tell them apart, but
there is nowhere to put the answer.

**`[xyz123] | .`** reaches `print_json`'s `StandardJson::Error(_) => out.write_all(b"null")` arm
(`src/bin/succinctly/jq_runner.rs`). #1247's design doc records that this arm was deliberately
reverted rather than made to raise: it walks child cursors lazily and streams as it goes, so by
the time a nested error is reached the opening bracket is already out, and bailing produced a
truncated document plus a generic exit 1 — worse on every axis than the `null` it replaced.

## Why this is its own issue, not #1194's remainder

#1194 is about the semi-index's member *pairing*, and its fix is complete for every route that
can carry an error. What is left is not a missing check — the check exists and is cheap — but a
**writer that cannot report**. That is the same constraint #1615 describes on the yq side and
[...]

PR #2310, opened 2026-09-03
docs(jq): pin evaluate_bytes_streaming's own prefix-leak as the established trade-off (#2210)

### LINKED ISSUE #2210: jq: evaluate_bytes_streaming leaks a partial prefix to stdout before reporting a decode failure
## What's wrong

`succinctly jq`'s `evaluate_bytes_streaming` (the non-M2, non-`-S` general
path, reached for a cursor-transparent expression once #1576's M2 fast path
declines on a malformed document) leaks partial output to real stdout
before reporting a decode failure -- unlike both the M2 path (fixed in
#1576's own follow-up, see the commit fixing the nested-empty-container gap)
and the fully materializing path used under `-S`, both of which write
nothing before erroring on the same input.

## Repro

```
$ printf '{"a": [,]}' > /tmp/t.json
$ succinctly jq -c '.' /tmp/t.json
{"a":jq: error (at /tmp/t.json:0): Invalid JSON text: expected JSON value, found ','
$ echo $?
5

$ succinctly jq -c -S '.' /tmp/t.json    # -S forces the fully materializing path
jq: error (at /tmp/t.json:0): Invalid JSON text: expected JSON value, found ','
$ echo $?
5
```

Both exit 5 with the same diagnostic, but the first leaks `{"a":` to stdout
before the error is reported (confirmed with stdout/stderr separated, not a
terminal-interleaving artifact) and the second writes nothing.

## Why this matters now

This is a pre-existing property of `write_output_jq_value`/
`evaluate_bytes_streaming` -- neither function was touched by #1576's diff.
It was effectively unreachable for this exact malformed shape before,
because the M2 fast path used to *silently accept* a stray comma in a
nested empty container (`{"a": [,]}` -> `{"a":[]}`, exit 0, no error at
all -- that bug is now fixed). Now that M2 correctly declines and falls
back to the general path for this shape, the general path's own older,
separate non-atomicity surfaces for the first time.

## Suggested fix

Either make `evaluate_bytes_streaming`'s single-document write buffer
before committing to real stdout (mirroring the buffer-then-commit pattern
`JsonCursor::stream_json`/`evaluate_m2_fast_path` already use), or
explicitly document this as an accepted "keep the prefix, diagnose later"
trade-off the way the LazySeq cursor-streaming path's own trade-off is
pinned by a test and doc comment elsewhere in this codebase -- right now it
is neither fixed nor documented as intentional.

## Found via

Code review of #1576 (jq-side M2 streaming), while verifying the fix for
the nested-empty-container-stray-comma bug on the M2 path.

newhoggy:
## Summary

- `evaluate_bytes_streaming` (the general, non-M2, non-`-S` path, reached once M2's own fast path declines a malformed document) leaks a container's already-written prefix to stdout before its walk reaches a nested fault — `{"a": [,]}` under bare `.` writes `{"a":` before raising on the stray comma inside the empty array. This surfaced for the first time only because #1576's own M2 fix now correctly declines this shape (it used to silently accept it as `{"a":[]}`, exit 0, never falling back to this path at all).
- Investigated both options the issue named: rearchitecting `evaluate_bytes_streaming` to buffer-then-commit, or documenting the existing behavior as an accepted trade-off. Chose the latter — `docs/compliance/jq/limitations.md` already pins the *identical* shape for a missing delimiter (`[1,,3] | .[]`, `test_jq_missing_delimiter_raises_through_nonreserializing_filters_1677`), for the same reason: "suppressing it would mean validating the whole document up front, which is the cost semi-indexing exists to avoid." Buffering this one path specifically would create an inconsistency with that already-established, deliberate architectural convention rather than fix a real one.
- Extends that doc section with this new confirmed instance and adds a regression test pinning the exact repro (live-verified against both the general path and `-S`, which validates upfront and so writes nothing on the same input).

Fixes #2210

## Test plan

- [x] New regression test `test_jq_general_streaming_path_leaks_prefix_before_nested_comma_fault_2210` pins both the general path's prefix-then-error behavior and `-S`'s clean-error behavior on the same input
- [x] Live-verified the exact repro from the issue against the built binary before and after
- [x] Full local test suite (`cargo test --features cli,simd,regex,serde`), both `cargo clippy` legs, `cargo fmt --check`, `cargo test --no-default-features`, and `cargo doc --no-deps --all-features` (`RUSTDOCFLAGS=-D warnings`) all pass

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -2819,16 +2819,23 @@ Pinned by `test_unbuffered_interleaves_stdout_and_stderr_1653`,
 `test_jq_missing_delimiter_raises_through_nonreserializing_filters_1677`
 ([tests/jq_cli_tests.rs](../../../tests/jq_cli_tests.rs)).
 
-The identical trade-off covers a *stray-comma* fault the same way it covers a *missing*
-delimiter above — `{"a": [,]}` under bare `.` writes `{"a":` before the walk reaches the
-empty array's stray comma and raises (#2210). `-S` forces the fully materializing path,
-which validates the whole document before printing anything, so it emits nothing on the
-same input — the exit code still agrees, only the prefix differs, same as every other case
-in this section. This shape was unreachable through the general (non-`-S`) path before
-#1576's own M2 fast-path fix started correctly declining a nested empty container's stray
-comma instead of silently accepting it as `[]`; only once M2 declines and falls back here
-does this path's own pre-existing, already-accepted non-atomicity apply to it for the first
-time. Pinned by
+The same trade-off also covers a fault found *partway through writing a single result's
+own value*, not just one discovered by a later top-level result's generator advance —
+`print_json`/`write_output_jq_value` stream byte-by-byte with no rewind, so `{"a": [,]}`
+under bare `.` writes `{"a":` before its own recursive walk reaches the stray comma nested
+inside the empty array and raises (#2210), the identical shape `test_jq_identity_on_
+malformed_array_element_errors_1641` above already pins for `[xyz123]`/`[tru]`/
+`[1,zzz,3]`/`{"a": xyz123}` — a bareword-garbage token instead of a stray comma, reaching
+the fault through the same writer the same way. A first pass at this fix buffered
+`write_output_jq_value`'s per-call output and only committed it to real stdout once known
+good (mirroring `evaluate_m2_fast_path`'s own identical contract for its own single-result
+case) — reverted once the existing #1641 test above caught it as a real regression against
+an already-established, deliberately tested contract for this exact writer, not a
+previously-undocumented gap. It would also have needed carving out `--unbuffered`, whose
+own flush (`write_terminator`) is embedded inside this same call tree keyed on whatever
+writer it's given — buffering would make that flush a silent no-op on a temporary buffer
+instead of the real, immediate per-value flush `--unbuffered` promises, breaking its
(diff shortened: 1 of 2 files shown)


newhoggy:
Update: code review argued the initial docs-only approach understated the case, since this fault is found partway through a *single* result's own write (unlike the delimiter-fault precedent I originally cited, which is a genuinely different mechanism — a later result's own generator advance finding nothing at all to buffer). Implemented a targeted fix mirroring `evaluate_m2_fast_path`'s own buffer-then-commit contract for exactly this single-result shape.

That fix turned out to regress two pre-existing, independently-authored tests (`test_jq_identity_on_malformed_array_element_errors_1641`, predating this PR by commit `7fb69832a`, and `test_jq_lazy_keys_array_trailing_comma_leaves_truncated_bracket_2261`) — both already pin a truncated-prefix contract for `print_json`/`write_output_jq_value`'s own byte-by-byte writer on the identical single-result shape. The review's premise that this was a previously-undocumented gap was itself mistaken: #1641 already decided not to buffer this writer.

Reverted the production change, kept `--unbuffered`'s own separate real bug (its embedded flush would've silently no-op'd on a temporary buffer) as evidence the fix direction was riskier than it looked, and rewrote the docs/test to cite #1641 directly as the closer, definitive precedent instead. Back to a docs+test-only PR — full local verification (build, both clippy legs, full test suite including the two previously-broken tests, fmt, no_std, doc) all green.

Issue #2305, opened 2026-09-03
yq: del() with a positive out-of-range index extends the array with nulls instead of no-op


newhoggy:
## What's wrong

Found investigating #2268 (negative out-of-range index handling for
`del()`/`delpaths()`). Real yq's `del()` on a **positive** out-of-range
array index doesn't leave the array unchanged (a no-op) the way jq does --
it extends the array with `null` up to and including that position, the
same way assignment (`.[N] = v`) already extends an array to fit a
too-large index, just without actually setting a value at the end:

```console
$ echo '[1,2]' | yq -o=json 'del(.[10])'
[1, 2, null, null, null, null, null, null, null, null]
$ echo '[1,2]' | yq -o=json 'del(.[5])'
[1, 2, null, null, null]
$ echo '[1,2]' | succinctly yq 'del(.[5])' -o json
[1, 2]                                          # no-op, current behavior
```

For comparison, real jq (and succinctly, matching jq) treats this as a
genuine no-op:

```console
$ echo '[1,2]' | jq -c 'del(.[5])'
[1,2]
```

Confirmed pre-existing (not introduced by #2268's own fix, which only
touches the negative-index case) via a stash-based A/B on the pre-#2268
binary.

## Scope

Purely about a **positive**, out-of-range index -- distinct from #2268's
own negative-index scope, and distinct from the already-known #2106
scalar-root no-op behavior. `delete_at_path`'s `Expr::Index` array arm
(`src/jq/eval.rs`) is the relevant call site; whether `delpaths()`'s own
array arms (`delete_paths_under`/`delete_keys`) share this exact
null-extension behavior for a positive OOB index hasn't been checked yet
either.

## Suggested fix

Investigate whether this null-extension shares any existing helper with
assignment's own array-extend logic (`resolve_setpath_index` or similar,
per its own doc comment referenced near `resolve_read_index`), or whether
it needs its own. Given how unusual this yq behavior is (deleting an index
that doesn't exist *grows* the array), worth double-checking the exact
extension length rule (does `del(.[N])` always extend to length `N+1`
regardless of how far out of range, matching the two examples above?) and
whether `delpaths()`'s multi-index batch shares it before implementing.

newhoggy:
🔧 Claimed by agent `Cadence` — branch `issue-2305-yq-del-positive-oob-extends`

Issue #2268, opened 2026-09-03
yq: del()/delpaths() silently no-op on negative out-of-range index instead of raising


newhoggy:
Follow-up to #2254 (merged) and #2264 — split out separately because this one
is more severe than #2264's own "low-traffic, message-wording" framing covers.

## What #2254 fixed, and what it didn't

#2254 made every ordinary *read* (`.a[-N]`, `.a[$n]`) raise
`index [N] out of range, array size is M` in yq mode when a negative index's
magnitude still exceeds the array length after resolving against it — the
same case real yq (v4.53.3) raises on, where jq (and yq's own
positive-out-of-range case) both answer `null`.

Compound assignment (`=`, `|=`, `+=`, ...) already raises unconditionally on
this shape too — but with jq's own message (`"Out of bounds negative array
index"`) instead of real yq's (`"index [N] out of range, array size is M"`).
That's a genuine divergence, but a cosmetic one (wrong wording, still an
error, still aborts the write) — confirmed live:

```bash
$ echo '{"a":[1,2]}' | yq -o=json '.a[-5] = 9'
Error: index [-5] out of range, array size is 2
$ echo '{"a":[1,2]}' | succinctly yq '.a[-5] = 9' -o json
Error: Out of bounds negative array index
```

## The actual bug: `del()`/`delpaths()` silently no-op instead of raising

Unlike assignment, the whole `del()`/`delpaths()` family currently does not
raise *at all* on this shape — it silently treats the out-of-range deletion
as a no-op and returns the document unchanged, where real yq aborts the
operation with an error. Confirmed live against yq v4.53.3:

```bash
$ echo '{"a":[1,2]}' | yq -o=json 'del(.a[-5])'
Error: index [-5] out of range, array size is 2
$ echo '{"a":[1,2]}' | succinctly yq 'del(.a[-5])' -o json
{
  "a": [1, 2]
}
```

```bash
$ echo '{"a":[1,2]}' | yq -o=json 'delpaths([["a",-5]])'
Error: index [-5] out of range, array size is 2
$ echo '{"a":[1,2]}' | succinctly yq 'delpaths([["a",-5]])' -o json
{
  "a": [1, 2]
}
```

This is worse than a missing error alone for the comma-grouped case: real yq
aborts the *entire* `del(...)` on the first bad component (jq/yq's
source-order, die-on-first-bad-path rule — the same rule
`delete_trie_array`'s own doc comment already cites for the non-array-target
[...]

Issue #527, opened 2026-07-31
jq: del() errors on a missing intermediate field inside a comma group, where jq treats it as a no-op


newhoggy:
**Severity: Low** — wrong error where jq gives a silent no-op, not a crash.

```
$ echo '{"a":{"x":1}}' | jq -c 'del(.a.b.c, .a.b.d)'
{"a":{"x":1}}

$ echo '{"a":{"x":1}}' | sjq -c 'del(.a.b.c, .a.b.d)'
jq: error: field 'b' not found
```

Deleting a nonexistent field is a no-op in jq, including through a missing intermediate: `del(.a.b)` on `{"a":{"x":1}}` correctly returns the input unchanged in both jq and succinctly. But once two sibling paths in a comma group share a missing intermediate field (`.a.b.c` and `.a.b.d`, both continuing through `.a.b`), succinctly's grouped-deletion code errors instead.

### Where

`delete_expr_object_paths` in `src/jq/eval.rs` groups comma-sibling paths by field name and recurses. When a *non-terminal* group's field is missing from the object, it raises `EvalError::field_not_found` unless every path in that group was itself marked `?`-optional:

```rust
match entries.get_mut(*name) {
    Some(slot) => { ... }
    None if group.iter().any(|p| p[start].optional) => {}
    None => return Err(EvalError::field_not_found(name)),
}
```

This path is only reachable when a field name appears in more than one *continuing* sibling (i.e., the grouped/multi-path code, added for #424/#475/#492) — the single-path walker (`delete_at_path`) has no equivalent check and correctly treats a missing intermediate as a no-op, matching jq. The grouped code should presumably do the same: a missing intermediate field means there's nothing to delete under it for any sibling in the group, not an error.

### Reproduction

Verified against the pinned jq oracle (jq-1.7.1) and confirmed pre-existing (reproduces identically on `e3efd067`, prior to #505's fix — that fix only touched the top-level identity/exhaustion check, not this field-name-group code).

Found incidentally while reviewing #505's fix; filed separately since it's unrelated to that change.

PR #2322, opened 2026-09-03
fix(yq): del()/delpaths() extend a mid-chain positive out-of-range index too (#2314)

### LINKED ISSUE #2314: yq: del()/delpaths() on a positive out-of-range *mid-chain* index still no-ops instead of extending (#2305's terminal-only fix)
## What's wrong

Follow-up to #2305 (terminal single-index `del()`/`delpaths()` positive-out-of-range
null-extension) -- deliberately out of that fix's own scope. Found during #2305's
own code review.

Real yq's positive-out-of-range extension isn't limited to a *terminal* delete
index (`del(.[N])`) -- it also fires when the out-of-range index is a **mid-chain**
step, i.e. there's more path after it (`del(.[N].a)`, `del(.[N][])`, or the
equivalent `delpaths([[N, "a"]])`). Confirmed live against yq v4.53.3, all
three still no-op in succinctly:

```console
$ echo '[1,2]' | yq -o=json 'del(.[5].a)'
[1, 2, null, null, null, null]            # length 6 = index + 1
$ echo '[1,2]' | succinctly yq 'del(.[5].a)' -o json
[1, 2]                                    # unchanged -- bug

$ echo '[1,2]' | yq -o=json 'delpaths([[5,"a"]])'
[1, 2, null, null, null, null]            # same shape, delpaths' own recursion
$ echo '[1,2]' | succinctly yq 'delpaths([[5,"a"]])' -o json
[1, 2]                                    # unchanged -- bug

$ echo '[1,2]' | yq -o=json 'del(.[5][])'
[1, 2, null, null, null, []]              # auto-vivifies `[]` at the tail,
                                           # not `null` -- a genuinely
                                           # different shape than the plain
                                           # object-field case above
```

## Why this needs its own fix, not just reusing #2305's helper

The extension target here is `index + 1` (making the out-of-range index
itself exist, so the *rest* of the path can navigate into it), not `index`
(the terminal case's own rule, #2305) -- `extend_array_with_nulls_for_delete`
cannot be reused as-is; this needs `pad_with_nulls`'s own `index + 1`
contract instead (or a direct call to it). The `del(.[5][])` case is a
further, third shape: the newly-created slot at `index` needs to be
auto-vivified to `[]` (or presumably `{}` for a field-navigation tail?
unconfirmed) rather than left as `null`, matching how `setpath`/assignment
already auto-vivifies a `null` into whatever container the next path
component needs.

## Scope

Three known-affected sites (found via code review of #2305, not yet fixed):

- `delete_path_steps`'s `Expr::Index` arm (`src/jq/eval.rs`) -- the AST-walk
  mid-chain case, e.g. `del(.[5].a)`.
- `delete_trie_array`'s mid-chain `ArrayStep::Index` arm (`src/jq/eval.rs`)
  -- the comma-separated multi-target case, e.g. `del(.[5].a, .[0])`.
- `delete_paths_under`'s array `Int`/`Float`/`NumberLiteral` arm
  (`src/jq/eval.rs`) -- `delpaths()`'s own mid-path recursion, a distinct
  implementation from the trie one above.

All three currently call the unmodified `resolve_read_index` and silently
no-op on a positive out-of-range mid-chain index -- unaffected by #2305,
which only touched the two *terminal* single-index call sites
(`delete_at_path`'s `Expr::Index` arm and `delete_keys`'s own single-key
case).

## Suggested fix

For each site: on `resolve_read_index` returning `None` for a positive
out-of-range index (distinguish from negative-OOB/NaN the same way
`resolve_delete_index`/`DeleteIndexResolution` does for #2305), grow the
array to `index + 1` via `pad_with_nulls` before recursing into the newly
created (currently-`null`) slot with the rest of the path, letting the
existing null-to-container auto-vivification (if any exists on this write
path) handle turning that `null` into whatever the next step needs -- confirm
whether that auto-vivification already exists here or needs its own work,
and confirm the `{}`-vs-`[]` tail-vivification rule against the oracle for
a field-navigation tail (`del(.[5].a)`), not just the `del(.[5][])` case
above.

newhoggy:
## Summary

#2305 fixed real yq's positive-out-of-range array-extension rule for `del()`/`delpaths()`, but scoped it to the *terminal* delete index (`del(.[N])`) only. A **mid-chain** positive out-of-range index — `del(.[N].a)`, `del(.[N][])`, `delpaths([[N,"a"]])` — still silently no-opped, unaffected by that fix.

Confirmed live against yq v4.53.3 (all three, matching the issue exactly):

```console
$ echo '[1,2]' | yq -o=json 'del(.[5].a)'
[1, 2, null, null, null, null]
$ echo '[1,2]' | succinctly yq 'del(.[5].a)' -o json   # before this fix
[1, 2]
```

## What changed

The extension target for a mid-chain index is `index + 1` (`pad_with_nulls`'s own contract, same as `setpath`), not #2305's `target_len` — the index itself must exist so the rest of the path has something to navigate into. `del(.[N][])` is a further nuance: the freshly-created slot can't stay `null` since `.[]` never tolerates one (#527), so it's auto-vivified straight to `[]` instead, matching real yq.

Fixed across the three sites the issue named:
- `delete_path_steps`'s `Expr::Index` arm (`src/jq/eval.rs`) — the plain AST-walk mid-chain case.
- `delete_trie_array`'s `ArrayStep::Index` arm — the comma-grouped multi-target case (`del(.[5].a, .[0])`).
- `delete_paths_under`'s array-key arm — `delpaths()`'s own recursion.

The trie walk needs no `[]`-auto-vivify case: a bare `.[]` can never reach it as a resolved step (it's always fanned out against the real document before the trie is built, and there's nothing at an out-of-range index to fan out over), and `delpaths`' path components are always concrete keys, never a `.[]` wildcard.

## A pre-existing test's premise didn't survive this

`test_1116_chained_scalar_slice_del_navigator_out_of_bounds_unaffected` asserted `{"a":[{"b":5}]} | del(.a[99].b[0:1])` was a total no-op. Live-checked against real yq: it's *not* — `.a` still grows to length 100, even though everything after the extension (`.b[0:1]` against the freshly-created `null`) is itself a no-op. The array mutation happens before the tail is even reached. Updated the test's expected output to match (confirmed live).

## Docs

Folded #2305's own `limitations.md` entry's "not yet implemented" note about this gap into a description of the fix, since this closes a real yq-compliance gap rather than opening a new divergence.

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` — full suite green (found and fixed one stale pre-existing test along the way)
- [x] `cargo test` (default features) — green
- [x] Live-verified all repro shapes + the fix against `yq` v4.53.3, byte-for-byte match
- [x] Confirmed jq mode stays a no-op (real jq has no such rule)
- [x] Confirmed #2268's negative-index error is unaffected

Fixes #2314

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -1880,12 +1880,29 @@ $ echo '[1,2]' | yq -o=json 'del(.[5][])'   # [1,2,null,null,null,[]]
 
 Closed across the three sites that resolve a mid-chain array index for a delete --
 `delete_path_steps`'s `Expr::Index` arm (the plain AST walk), `delete_trie_array`'s
-`ArrayStep::Index` arm (the comma-grouped multi-target walk), and `delete_paths_under`'s
-array-key arm (`delpaths`' own recursion) -- all in `src/jq/eval.rs`. The trie walk has no
-`[]`-vivify case to mirror: a bare `.[]` can never reach it as a resolved step (a
-comma-grouped `del()`'s own `.[]` is always fanned out against the real document before the
-trie is built, and an out-of-range mid-chain index has nothing there to fan out over), and
-`delpaths`' own path components are always concrete keys, never a `.[]` wildcard.
+`ArrayStep::Index` arm (the comma-grouped multi-target walk, once that walk's own
+read-based fan-out has already resolved a concrete branch reaching it -- see the residual
+gap below for when it hasn't), and `delete_paths_under`'s array-key arm (`delpaths`' own
+recursion) -- all in `src/jq/eval.rs`. `delpaths`' own path components are always concrete
+keys, never a `.[]` wildcard, so there is no `[]`-vivify case to mirror at that site.
+
+**Two residual gaps found during this fix's own review, not chased here:**
+
+- A **comma-grouped** target whose own `.[]` fan-out crosses the out-of-range index
+  (`del(.[5][], .[0])`) never reaches `delete_trie_array` at all -- the read-based
+  validation that enumerates concrete branches for the trie reads `.[5]` as a plain
+  `null` (correct for an ordinary read) and raises iterating `.[]` over it, rather than
+  extending+vivifying the way the direct, non-comma-grouped form now does. Tracked as
+  [#2324](https://github.com/rust-works/succinctly/issues/2324).
+- The `[]`-vivify rule implemented here is narrowly scoped to a slot this fix *itself*
+  just padded, checked only against a literal `Expr::Iterate` as the very next step.
+  Real yq's actual rule is broader and predates this fix entirely: `null` auto-vivifies
+  into `[]` for **any** subsequent `Index`/`Iterate` del() step, freshly padded or not
+  (`null | del(.[2])` is `[null,null]` in real yq; `{"x":null} | del(.x[2])` is
+  `{"x":[null,null]}`) -- `succinctly yq` gets both wrong today, along with a
+  multi-level chain like `del(.[3][2].a)` (real yq recurses the same rule at each level;
+  this fix's own narrow check only covers one). Tracked as
+  [#2323](https://github.com/rust-works/succinctly/issues/2323).
 
 ### Other categories
 
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -1857,13 +1857,35 @@ path-array form) shares the same order-dependence.
 A negative out-of-range index is untouched by this fix either way -- it raises instead
 (#2268, the section above, landed separately and already covers it).
 
-Also scoped to a *terminal* index only -- a positive out-of-range index with more path
-after it (`del(.[5].a)`, `del(.[5][])`) extends too in real yq, to `index + 1` rather than
-`index` (there *is* something at the out-of-range position once the rest of the path needs
-to navigate into it), and `del(.[5][])` specifically auto-vivifies `[]` there rather than
-`null` -- a different shape from this fix's own terminal-only rule, confirmed live but not
-implemented; tracked separately as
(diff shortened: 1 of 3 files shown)


newhoggy:
Code review verified all three fixed sites against the live yq oracle and caught 4 real gaps, all confirmed live:

1. **Comma-grouped `.[]` fan-out** (`del(.[5][], .[0])`) still raises — this doesn't reach any of the three sites this PR fixed; the read-based branch-enumeration that builds the comma-grouped target set (`resolve_node`) reads the out-of-range index as a plain `null` and raises iterating `.[]` over it. Filed as #2324, since a proper fix needs to touch shared path-resolution machinery carefully, not just this PR's own three sites.
2. **General null-vivify rule is broader than this fix's own narrow case** — real yq vivifies `null` → `[]` for *any* subsequent `Index`/`Iterate` step, not just a slot this fix itself just padded (`null | del(.[2])` → `[null,null]` in real yq, no out-of-range extension involved at all). This predates #2314 entirely and is reachable via `delete_at_path`'s terminal arms too, not just the mid-chain loop this PR touches. Filed as #2323.
3. One of the review's reported cases (`del(.[3][2].a)`, chained Index-after-padded-null) is a symptom of #2323's general rule, not a separate bug.
4. One of the review's reported cases (`del(.[5] | (.[].a))`, a Pipe-wrapped iterate) is also a symptom of #2323 — the existing Pipe-splice logic already reduces it to a literal `Iterate` against a `null` root by the time it matters, so #2323's fix (once implemented) covers it for free; no separate lookahead hack needed.

Corrected `docs/compliance/yq/limitations.md`'s own entry, which had overclaimed the three sites left nothing else to fix. This PR's own scope — the three call sites and exact repro shapes issue #2314 named — is implemented and verified correct; the residual gaps above are pre-existing, broader, and now tracked separately.

PR #2325, opened 2026-09-03
fix(jq): E[K]/E[S:T] keeps E's own partial prefix on its mid-stream error

### LINKED ISSUE #2226: jq: E[K]/E[S:T]'s target's own within-generator Partial prefix is discarded on its own mid-stream error
**Severity: Low** (side-effect/streamed-value count only, reachable when the computed
target `E` is itself a multi-output generator that errors partway through)

## Summary

`E[K]`/`E[S:T]` re-evaluate `E` once per key/`(s, e)` pair (`eval_index_expr`, fixed by
#2032/#2142; `eval_slice_expr`, fixed by #2143/PR #2223). Both fixes correctly fold the
*cross-key/cross-pair* accumulator (`out`) into a `Partial` result when a later
key's/pair's own evaluation of `E` escapes with `Error`/`Break`/`Halt` — so an earlier
key's/pair's already-produced output now survives.

What neither fix (nor the original pre-#2032 code) handles: if `E` **itself** is a
multi-output generator and its evaluation for the *current* key/pair produces some
values before erroring, that within-this-iteration prefix is still discarded. Both
`eval_index_expr` and `eval_slice_expr` (all four functions, `src/jq/eval.rs` and
`src/jq/eval_generic.rs`) match `E`'s result as `QueryResult::Partial(_, control)` /
`GenericResult::Partial(_, control)` and discard the first field (the prefix) entirely,
keeping only `control`.

Index case (pre-existing on `main`, unrelated to #2143):

```console
$ echo null | jq -c '0 as $k | ([1,2],[3,4],error("x"))[$k]'
1
3
jq: error (at <stdin>:1): x

$ echo null | succinctly jq -c '0 as $k | ([1,2],[3,4],error("x"))[$k]'
jq: error (at <stdin>:1): x
# 1 and 3 are lost -- E's own generator produced them before erroring
```

Slice case (same shape, via PR #2223's new `eval_slice_expr`):

```console
$ echo null | jq -c '([1,2],[3,4],error("x"))[(0,1):(1,2)]'
[1]
[3]
jq: error (at <stdin>:1): x

$ echo null | succinctly jq -c '([1,2],[3,4],error("x"))[(0,1):(1,2)]'
jq: error (at <stdin>:1): x
# [1] and [3] (sliced from the target's own prefix) are lost
```

Note the *literal*-index form (`[0]` rather than a genuinely computed `0 as $k | ...[$k]`)
does **not** exhibit this — it dispatches through a different, literal-key code path
(`Expr::Index`) that isn't affected. Use a genuinely computed key/bounds to reproduce.

## Root cause

In all four functions, the target-evaluation match has:

```rust
QueryResult::Partial(_, control) => escape!(control),      // eval.rs (both fns)
GenericResult::Partial(_, control) => escape!(control),    // eval_generic.rs (both fns)
```

The underscored first field — `E`'s own already-produced values for *this* key/pair,
before its own mid-stream escape — is discarded rather than being indexed/sliced (same
per-key/-pair operation the successful branches already apply) and folded into the
running accumulator alongside whatever escape/control eventually propagates.

## Why this is separate from #2143/#2032

Both are about re-evaluating `E` at the right granularity (once per key, once per
`(s, e)` pair) — a different question from what happens to `E`'s *own* partial output
within a single already-correct evaluation. This gap exists identically in the
already-merged `eval_index_expr` (#2032/#2142), so it isn't something #2143 introduced
or worsened; #2143's own review confirmed the two are orthogonal.

## Suggested fix direction

In the `Partial(prefix, control)` arm, apply the per-key/-pair operation (`index_one`/
`index_one_owned`/`slice_one_generic`/`slice_owned_value_read`/`eval_single(&slice_expr,
...)` — whichever the surrounding arm already uses for a full result) to each of
`prefix`'s values, push the successes into the running accumulator the same way the
success arms do, *then* escape with `control` — mirroring the reserve-and-push loop the
surrounding code already has, just seeded from `prefix` instead of a fresh `Vec` pulled
out of `Many`/`ManyOwned`. Needs its own capacity-reservation care (this adds another
unbounded-length source to reserve against) and live-oracle verification across both
`eval_index_expr` and `eval_slice_expr`, in both `eval.rs` and `eval_generic.rs` (4 call
sites total).

## Refs

Found during code review of PR #2223 (fixing #2143); also present on `main` prior to
that PR via `eval_index_expr`.

newhoggy:
## Summary
- `eval_index_expr`/`eval_slice_expr` (both `src/jq/eval.rs` and `src/jq/eval_generic.rs`, 4 call sites total) re-evaluate the target `E` once per key/`(s, e)` pair, but discarded `E`'s own already-produced values whenever `E` itself is a multi-output generator that errors/breaks/halts partway through its own stream.
- Applies the per-key index/slice operation to each already-produced value before propagating the held control — mirroring the pre-existing successful-target arms — matching jq 1.7.1 exactly (live-verified against the pinned `/usr/bin/jq` 1.7.1 oracle for every repro below).
- Corrects several pre-existing tests that pinned the old (buggy, some explicitly documented-as-divergence) behavior as expected, including two `group_by`-key-fn cases whose "exits 0, no output" expectation was itself never actually verified against real jq (indexing/slicing a *number* is never valid regardless of the trailing halt, so real jq raises before the halt is ever reached).

Fixes #2226

## Follow-up filed
A structurally similar but distinct gap survives out of scope: `eval_index_expr`'s **key**-stream (not target-stream) `Partial` arm also discards its own prefix on `Error`/`Break` (confirmed live: `jq -c '.[(1,error("x"))]'` on `[10,20,30]` prints `20` then fails; succinctly still collapses to a bare error). Filed as a follow-up issue.

## Test plan
- [x] `cargo fmt --check`
- [x] All 4 fixed call sites live-verified against `/usr/bin/jq` 1.7.1, including confirming (via temporary debug probes) which repro shape actually routes through which of the 4 functions

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -16837,6 +16837,25 @@ fn eval_index_expr<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
                             // #1832); `vs` is never attempted here either
                             // way, matching that helper's own `extra` never
                             // being touched on its `Err` branch.
+                            //
+                            // Not live-repro-tested (review): `target` is one
+                            // fixed expression re-evaluated fresh per key
+                            // against the same document, so whether a given
+                            // key's evaluation is a clean success (populating
+                            // `borrowed`) or a `Partial` (reaching this arm)
+                            // is a deterministic function of `(target, value)`
+                            // alone, not of which key is current -- an
+                            // *earlier* key populating `borrowed` and a
+                            // *later* key hitting this arm therefore requires
+                            // `target` itself to behave non-deterministically
+                            // across the two calls. The one stateful builtin
+                            // that could do that, `input`, always returns an
+                            // owned value (`builtin_input`'s own
+                            // `owned_vec_to_result`), so it can never
+                            // populate `borrowed` in the first place. Kept
+                            // anyway, matching `resolve_terminal_prefix`'s
+                            // own identical defensive handling, in case a
+                            // future stateful mechanism changes this.
                             Err((prefix, e)) => {
                                 let control = match control {
                                     Control::Halt(_) => control,
@@ -52702,6 +52721,23 @@ mod tests {
         );
     }
 
+    /// #2226 review (patch coverage): the target's own `Partial` arm's
+    /// `Ok(None)` sub-case -- an earlier-produced value that legitimately
+    /// isn't indexable, suppressed by `optional` rather than erroring
+    /// (`true` here; `?` only gates *this* per-value type mismatch, not
+    /// `target`'s own trailing `error("x")`). Verified against jq 1.7.1:
+    /// `0 as $k | (true, [7], error("x"))[$k]?` prints `7` then still
+    /// raises `x`.
+    #[test]
+    fn test_index_expr_target_own_partial_prefix_ok_none_suppressed_by_optional_2226() {
+        query!(b"null", r#"0 as $k | (true, [7], error("x"))[$k]?"#,
+            QueryResult::Partial(vs, Control::Error(e)) => {
+                assert_eq!(vs, vec![OwnedValue::Int(7)]);
+                assert_eq!(e.message, "x");
+            }
+        );
+    }
+
     /// #2226 sibling: the same target-own-prefix fix for `eval_slice_expr`.
     /// Both bounds are `(N+0)` rather than bare literals -- a slice whose
     /// bounds both fully fold to constants never reaches `eval_slice_expr`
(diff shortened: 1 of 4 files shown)


newhoggy:
## Code review results (ultra multi-angle review, 8 finder agents)

Highest-severity finding, confirmed by 3 independent agents and by me directly: the
original fix applied unconditionally via the shared `S: EvalSemantics` generic, but real
yq does **not** stream a target's own escaped generator's prefix at all (unlike jq).
Pre-#2226, `succinctly yq` already matched real yq here by coincidence (the old code
discarded everything); the fix broke that match for 4 call sites. Fixed by gating all 4
sites to jq mode only (`S::TAG == EvalTag::Yq` keeps the pre-#2226 conservative discard),
matching ADR-0018's "the mode decides, never the input format" rule — live-verified
against yq v4.53.3 for all 4 sites, both before and after the gate.

Also fixed:
- Two `Partial` arms (`eval.rs`'s and `eval_generic.rs`'s `eval_index_expr`) could
  silently downgrade a pending `Control::Halt` into a decode `Control::Error` when
  promoting a prior key's still-unvalidated prefix failed — violating this codebase's
  established "an uncatchable halt must not be downgraded" rule (#987/#1832), already
  enforced elsewhere in the same functions via `resolve_terminal_prefix`/`escape_generic!`.
- A new test (`test_slice_expr_target_own_partial_prefix_preserved_2226`) accidentally
  exercised the wrong code path — a slice with two constant-folding literal bounds never
  reaches `eval_slice_expr` at all (#1326) — fixed to use non-foldable bounds, and an
  inline code comment with the identical mistake was corrected the same way.
- Several doc comments that described the now-fixed gap as still open/deliberately
  unaddressed were updated.

New tests added: yq-mode carve-out coverage for all 4 sites (2 in `eval.rs`'s lib tests
via `yq_query!`, 2 CLI-level in `tests/yq_cli_tests.rs`).

Filed a follow-up for a structurally identical, still-open gap the review surfaced:
`eval_index_expr_with_path_context`/`eval_slice_expr_with_path_context` (the
assignment/`del()`/`path()`-context evaluators) still discard the target's own prefix —
out of #2226's own scope (plain-read evaluators only), tracked as #2328. Also corrected
a now-stale comment on that pair claiming behavior it no longer matches.

All local verification (build/clippy ×2/fmt/doc/test ×3 feature tiers) re-run clean after
these fixes, rebased onto latest `main` with no conflicts.

Issue #2324, opened 2026-09-03
yq: comma-grouped del()/delpaths() targets whose .[] fan-out crosses a positive out-of-range index still raise (#2314 sibling)


newhoggy:
## What's wrong

Found during #2314's code review. #2314 fixed a mid-chain positive out-of-range array
index for the single-target `del()`/`delpaths()` forms (`del(.[N].a)`, `del(.[N][])`,
`delpaths([[N,"a"]])`), reusing `pad_with_nulls`/`resolve_delete_index` across
`delete_path_steps`, `delete_trie_array`, and `delete_paths_under`.

The **comma-grouped** form of the exact same `del(.[N][])` shape still raises, both
alone and combined with an unrelated in-range target -- confirmed live against yq
v4.53.3:

```console
$ echo '[1,2]' | yq -o=json 'del(.[5][], .[0])'
[2, null, null, null, []]
$ echo '[1,2]' | succinctly yq 'del(.[5][], .[0])' -o json
Error: Cannot iterate over null (null)
```

The `?` (optional) variant is worse -- it silently *drops* the extension entirely
instead of raising:

```console
$ echo '[1,2]' | yq -o=json 'del(.[5][]?, .[0])'
[2, null, null, null, []]
$ echo '[1,2]' | succinctly yq 'del(.[5][]?, .[0])' -o json
[2]
```

## Root cause

This doesn't reach `delete_trie_array` at all -- the failure happens earlier, during
the comma-grouped target's *read-based* fan-out validation (`resolve_node`, via
whatever builds the resolved branch set the delete trie is constructed from). That
validation evaluates `.[5][]` as an ordinary read to enumerate concrete branches:
reading `.[5]` on `[1,2]` answers `null` (index 5 is out of bounds), and iterating a
`null` for fan-out purposes raises the same "Cannot iterate over null" a genuine read
would -- correctly, for a *read*, but not for a `del()` target where index 5 is
supposed to extend the array first.

`resolve_node` is shared, general-purpose path-resolution machinery used well beyond
`del()`'s own comma-grouped fan-out, so a fix here needs to be scoped carefully to the
delete-target caller rather than changing read semantics generally -- likely a
delete-specific pre-pass or a way to thread "this positive-out-of-range read should
extend+vivify, not raise" through only the call sites that build a `del()`/`delpaths()`
target set, without touching `resolve_node`'s many other (genuinely read-only) callers.

## Scope

[...]

Issue #1863, opened 2026-08-29
jq: navigate_read_only/PrefixNavOutcome duplication, plus 3 undiscovered yq write-path divergences found alongside #1432


newhoggy:
**Severity: Low** (code-quality + narrow correctness divergences, no data-loss risk)

Found via code review on PR #1861 (#1432's fix). Bundling three related-but-separate findings rather than filing three issues, since they surfaced from the same review pass and share a root area (`src/jq/eval.rs`'s `Field`/`Index` mid-chain handling).

## 1. `navigate_read_only`/`PrefixNavOutcome` is now over-engineered for its one remaining caller

Before #1432, `navigate_read_only` had two callers: `yq_assign_is_total_noop` (needed the full `Resolved(parent)`/`HitScalar`/`Absent` 3-way distinction) and `yq_del_slice_outcome` (only ever checked `matches!(_, Resolved(_))`, folding `HitScalar` back into "doesn't apply" same as `Absent`). #1432 replaced the first caller with a new recursive `assign_path_all_noop`, so `navigate_read_only` now has exactly one caller left, which only needs a boolean "did every step resolve to an existing value" answer.

This leaves:
- `PrefixNavOutcome::Resolved`'s payload already dropped to a unit variant (dead once the terminal-scalar check moved into `assign_path_all_noop`'s own base case).
- `assign_path_all_noop` and `navigate_read_only` now duplicate the `Field`/`Index`/`IndexNumber` match arms (and their "defensive, not reachable" catch-alls) almost verbatim -- the only real difference is `assign_path_all_noop`'s `Iterate` arm recursing instead of deferring.

Suggested fix: simplify `navigate_read_only` to a plain `fn(&OwnedValue, &[Expr]) -> bool` (drop `PrefixNavOutcome` entirely), or have `yq_del_slice_outcome` call `assign_path_all_noop`-style logic instead if its own `Iterate`-in-a-del-prefix semantics allow it (not verified whether `del()`'s own slice prefixes can ever contain an `Iterate` -- check before doing this to avoid silently changing `del()`'s behavior).

## 2. `let len = arr.len() as i64; ...` negative-index-normalization idiom duplicated 7x

[...]

PR #2331, opened 2026-09-03
fix(yq): del()/delpaths() vivify a null into [] ahead of an Index/Iterate step (#2323)

### LINKED ISSUE #2323: yq: null never vivifies into [] for a following del()/delpaths() Index/Iterate step (broader than #2314)
## What's wrong

Found during #2314's code review. Confirmed live against yq v4.53.3: in yq mode, a
`null` value reached mid-`del()` (or as the whole document) auto-vivifies into an empty
array `[]` when the *next* path component is `Index` or `Iterate` — the same auto-vivify
philosophy `setpath` already applies for assignment, but never wired into `del()`'s own
null-handling.

```console
$ echo 'null' | yq -o=json 'del(.[2])'
[null, null]
$ echo 'null' | yq -o=json 'del(.[])'
[]
$ echo '{"x":null}' | yq -o=json 'del(.x[])'
{"x": []}
$ echo '{"x":null}' | yq -o=json 'del(.x[2])'
{"x": [null, null]}
$ echo '[1,2]' | yq -o=json 'del(.[3][2].a)'
[1, 2, null, [null, null, null]]
```

`succinctly yq` currently gets every one of these wrong -- either stays a plain `null`
(`del(.[2])`, `del(.x[2])`), raises `Cannot iterate over null (null)` (`del(.[])`,
`del(.x[])`), or under-vivifies past the first extended level (`del(.[3][2].a)` gives
`[1,2,null,null]`, not `[1,2,null,[null,null,null]]`).

## Scope -- confirmed both ways live

**Applies to:** `Index`/`Iterate` steps only, both bare/terminal
(`delete_at_path`, `src/jq/eval.rs`) and mid-chain (`delete_path_steps`), and
presumably `del()`'s comma-grouped form too (`delete_trie_array`/`delete_trie_object`'s
own null-skip branches) -- not yet confirmed live for the comma-grouped case
specifically, worth checking when this is picked up.

**Does NOT apply to:**
- `Field` steps -- `{"x":null} | del(.x.a)` stays `{"x":null}` (confirmed live).
- `Slice` steps -- `null | del(.[0:2])` stays `null` (confirmed live), matching jq.
- `delpaths()` at all -- `{"x":null} | delpaths([["x",2]])` stays `{"x":null}`
  (confirmed live) -- this rule is specific to `del()`'s own expression-based target,
  not `delpaths()`'s path-array form.
- jq mode -- unaffected either way; jq has no such rule (`null | del(.[2])` stays `null`
  in real jq 1.7.1, `null | del(.[])` raises the same "Cannot iterate over null" it
  already does in succinctly).

## Relationship to #2314

Pre-existing and broader than #2314's own fix -- reproduces on a bare `null | del(.[2])`
with no positive-out-of-range index or array involved at all, so it is not something
#2314's own three sites (which only handle *extending an existing array*) could have
covered. #2314's own narrow fix for `del(.[N][])` (vivify a freshly-*padded* slot to `[]`
when the very next step is a literal `Expr::Iterate`) is a correct special case of this
same broader rule, kept as-is rather than folded into a fuller fix here.

## Suggested approach

A general helper invoked wherever `Expr::Index`/`Expr::Iterate` currently matches
`OwnedValue::Null` in yq mode: replace the value with `OwnedValue::Array(Vec::new())`
first, then fall through to (or re-enter) the existing `OwnedValue::Array` handling for
that same step, so the array-growth/iteration logic doesn't need duplicating. In
`delete_path_steps`'s loop this can likely be a `*root = OwnedValue::Array(Vec::new());
continue;` right before the existing Null arm's current `steps = rest; continue;`, gated
on `yq_mode` -- letting the loop re-match the *same* `component` against the freshly
vivified array. `delete_at_path`'s terminal arms would need the equivalent inline.

newhoggy:
## Summary

Found during #2314's own review. Real yq's actual `null`-vivify rule for `del()` is broader than, and predates, #2314's own "extend a slot this fix just padded" case: **any** `null` a `del()` walk reaches auto-vivifies into `[]` the moment the next step is `Index` or `Iterate` — the same auto-vivify philosophy `=`/`|=`'s own write path already applies (there's existing precedent: `test_yq_iterate_pipe_chain_null_autovivifies_under_update_assign_1919`).

```console
$ echo 'null' | yq -o=json 'del(.[2])'          # [null, null] -- no out-of-range extension involved at all
$ echo 'null' | yq -o=json 'del(.[])'           # []
$ echo '{"x":null}' | yq -o=json 'del(.x[2])'   # {"x": [null, null]}
$ echo '[1,2]' | yq -o=json 'del(.[3][2].a)'    # [1, 2, null, [null, null, null]] -- recurses at each level
```

## What changed

Implemented in three places, all `src/jq/eval.rs`:
- `delete_at_path`'s `Index`/`Iterate` arms (the terminal, single-component form) — vivify `null` to `Array(Vec::new())` right before the existing container-handling match, so the existing array logic (including #2305/#2314's own extension) runs unchanged.
- `delete_path_steps`'s `Index`/`Iterate` arms (the mid-chain AST walk) — same idea, but implemented as `*root = Array(vec![]); continue;` *without* advancing `steps`, so the very same path component re-matches against the freshly vivified array on the next loop pass. This is what makes a **chained** out-of-range index (`del(.[3][2].a)`) recurse correctly for free — no explicit recursion or lookahead needed.
- `delete_trie_array`'s null-skip branch (the comma-grouped multi-target walk) — same vivify, gated to when the step is array-level (`delete_trie_object`'s own sibling branch, for `Field` steps, is untouched — `Field` never vivifies).

**Scope, all confirmed live:**
- Does **not** apply to `Field` steps or `Slice` steps (both stay `null`, matching jq and matching `=`/`|='s own precedent).
- Does **not** apply to `delpaths()` at all (`{"x":null} | delpaths([["x",2]])` stays `{"x":null}` — the rule is specific to `del()`'s expression-based target).
- `?` does not suppress the vivify.
- A `Pipe`-wrapped continuation (`del(.[5] | (.[].a))`) is covered for free — it reduces to a literal `Expr::Iterate` against the padded `null` once the existing `Expr::Pipe` splice logic runs.
- jq mode is entirely unaffected (no such rule exists in jq at all).

One surprising, **already-correct, pre-existing and unrelated** oddity surfaced while checking the comma-grouped case, worth flagging rather than silently working around: `{"x":null,"y":1} | del(.x[0:2], .y)` is `{}` in real yq — not just `x` cleared, `y` gone too. `succinctly yq` already reproduces this identically both before and after this fix, via the pre-existing #1116/#1219 chained-slice-parent-drop rule elsewhere in the same trie walk. Verified this fix's own vivify doesn't disturb it.

## A pre-existing test's premise didn't survive this

`test_yq_del_slice_outcome_iterate_prefix_scalar_and_null_1432` had pinned the `Null` case as *known-divergent* (a documented gap from #1863). Updated its expected result to the correct, now-matching output; the test's *other* half (a genuine scalar hit, e.g. `a: 5`) is a different, still-open gap, unaffected by this fix, and stays as-is.

## Not in scope (tracked separately)

#2324's own gap — a comma-grouped target whose `.[]` fan-out crosses an out-of-range index (`del(.[5][], .[0])`) — is unrelated: that one fails earlier, in the read-based validation that builds the trie in the first place, before `delete_trie_array` is ever reached. This PR doesn't touch that path.

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` — full suite green (found and corrected one stale pre-existing test)
- [x] `cargo test` (default features) — green
- [x] Live-verified every repro shape (bare null, field-reached null, chained/recursive, pipe-wrapped, comma-grouped, the Slice-only and mixed comma-grouped cases) against `yq` v4.53.3, byte-for-byte match
- [x] Confirmed jq mode and `delpaths()` are both unaffected

Fixes #2323

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -1886,23 +1886,73 @@ gap below for when it hasn't), and `delete_paths_under`'s array-key arm (`delpat
 recursion) -- all in `src/jq/eval.rs`. `delpaths`' own path components are always concrete
 keys, never a `.[]` wildcard, so there is no `[]`-vivify case to mirror at that site.
 
-**Two residual gaps found during this fix's own review, not chased here:**
+**One residual gap found during this fix's own review, not chased here:** a
+**comma-grouped** target whose own `.[]` fan-out crosses the out-of-range index
+(`del(.[5][], .[0])`) never reaches `delete_trie_array` at all -- the read-based
+validation that enumerates concrete branches for the trie reads `.[5]` as a plain `null`
+(correct for an ordinary read) and raises iterating `.[]` over it, rather than
+extending+vivifying the way the direct, non-comma-grouped form now does. Tracked as
+[#2324](https://github.com/rust-works/succinctly/issues/2324).
 
-- A **comma-grouped** target whose own `.[]` fan-out crosses the out-of-range index
-  (`del(.[5][], .[0])`) never reaches `delete_trie_array` at all -- the read-based
-  validation that enumerates concrete branches for the trie reads `.[5]` as a plain
-  `null` (correct for an ordinary read) and raises iterating `.[]` over it, rather than
-  extending+vivifying the way the direct, non-comma-grouped form now does. Tracked as
-  [#2324](https://github.com/rust-works/succinctly/issues/2324).
-- The `[]`-vivify rule implemented here is narrowly scoped to a slot this fix *itself*
-  just padded, checked only against a literal `Expr::Iterate` as the very next step.
-  Real yq's actual rule is broader and predates this fix entirely: `null` auto-vivifies
-  into `[]` for **any** subsequent `Index`/`Iterate` del() step, freshly padded or not
-  (`null | del(.[2])` is `[null,null]` in real yq; `{"x":null} | del(.x[2])` is
-  `{"x":[null,null]}`) -- `succinctly yq` gets both wrong today, along with a
-  multi-level chain like `del(.[3][2].a)` (real yq recurses the same rule at each level;
-  this fix's own narrow check only covers one). Tracked as
-  [#2323](https://github.com/rust-works/succinctly/issues/2323).
+(A second gap found during the same review -- `null` never vivifying into `[]` ahead of
+*any* `Index`/`Iterate` del() step, not just a slot #2314 itself just padded -- is now
+closed; see the next section.)
+
+### `del()` auto-vivifies a `null` into `[]` ahead of an `Index`/`Iterate` step, matching `setpath`
+
+[#2323](https://github.com/rust-works/succinctly/issues/2323), found during #2314's own
+review. Real yq's actual null-vivify rule is broader than -- and predates -- #2314's own
+"a slot this fix just padded" case: **any** `null` a `del()` walk reaches, freshly padded
+or not, vivifies into `[]` the moment the next step is `Index` or `Iterate`, the same
+auto-vivify philosophy `=`/`|=`'s own write path already applies (`.a[].b = v` on
+`{"a":null}` is `{"a":[]}`, confirmed pre-existing precedent -- see
+`test_yq_iterate_pipe_chain_null_autovivifies_under_update_assign_1919`):
+
+```bash
+$ echo 'null' | yq -o=json 'del(.[2])'          # [null, null] -- no out-of-range extension involved at all
+$ echo 'null' | yq -o=json 'del(.[])'           # []
+$ echo '{"x":null}' | yq -o=json 'del(.x[2])'   # {"x": [null, null]}
+$ echo '[1,2]' | yq -o=json 'del(.[3][2].a)'    # [1, 2, null, [null, null, null]] -- recurses at each level
+```
+
+Applies to `del()`'s terminal form (`delete_at_path`'s own `Index`/`Iterate` arms) and its
+mid-chain form (`delete_path_steps`'s own `Index`/`Iterate` arms) alike. Both are
+implemented the same way: mutate the `null` root into `Array(Vec::new())` immediately
+before the existing container-handling match, rather than duplicating that match's own
+logic -- in `delete_path_steps`'s loop this means `continue`ing without advancing `steps`,
+so the very same path component re-matches against the freshly vivified array on the next
+pass, which is what makes the recursive `del(.[3][2].a)` case above fall out for free
+(each level's own out-of-range index re-triggers #2314's existing extension logic against
+an array that used to be `null`) without any explicit lookahead or recursion of its own.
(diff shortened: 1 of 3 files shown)


newhoggy:
Code review found 3 real regressions in the initial fix, all confirmed live before pushing this update:

1. **\`delete_at_path_through_absent\` leaked the real \`yq_mode\`** into its synthetic scratch walk, so the new vivify fired on a value the function's own contract says stays untouched — \`{} | del(.missing[-1])\` went from a clean no-op to a raised out-of-range error. Fixed by hardcoding \`yq_mode=false\` there, matching \`delete_trie_through_absent\`'s already-established sibling pattern (and dropping the now-pointless \`yq_mode\` parameter entirely, matching that sibling's own signature).

2. **The vivify only checked "is root currently null," not how it got there.** A \`null\` reached by *tolerating* a \`Field\`/\`Slice\` step (#476/#527) turns out not to be vivify-eligible in real yq, even when it sits in a real (possibly #2314-padded) slot: \`[1,2] | del(.[5].missing[2])\` incorrectly vivified \`.missing\` into an array where real yq leaves it plain \`null\`. Fixed by threading a \`real_slot: bool\` through \`delete_at_path\`/\`delete_path_steps\` — true from a real navigation (found key, in-range/padded element), false the moment a null-tolerant \`Field\`/\`Slice\` arm fires, never flipped back except by this fix's own vivify.

3. **The comma-grouped vivify ignored \`delete_keys\`'s own #2305/#2306 single-key-only extension gate**, so a 2+-terminal-key batch traded one wrong answer (staying \`null\`) for a *different*, newly wrong one (\`[]\`) instead of real yq's actual order-dependent multi-key answer (out of scope, #2306). Fixed by gating the vivify on \`terminal_index_count <= 1\` (continuations excluded from the count), which is a genuinely different, already-correct shape.

Added regression tests for all three, and corrected \`docs/compliance/yq/limitations.md\`'s own scope description to match. Full local verification (build, both clippy legs, full test suite, no_std, doc) all green.

newhoggy:
Re-review round 2 verified the \`real_slot\` threading and \`terminal_index_count\` gate are correctly wired end-to-end (no missed or mis-valued call sites), and surfaced 2 more findings — both confirmed **pre-existing** (identical output on the commit before this PR, not a regression from it):

- Comma-grouped \`del()\` with a bare \`.[]\` branch against a genuinely-existing \`null\` (not an out-of-range index) hits the same \`resolve_node\`-fan-out gap #2324 already tracks — added as a confirmed additional case there.
- \`yq_del_slice_outcome\`'s trailing-scalar-slice-run Noop classification silently no-ops regardless of target type, where real yq only does that for a string (arrays/null should raise) — unrelated root cause, filed as #2333.

Neither blocks this PR — both are pre-existing gaps in surrounding code this fix happens to sit near, not something #2323's own vivify logic touches or introduces. Proceeding to merge.

Issue #2016, opened 2026-08-31
test: same zombie-child-leak pattern as #1891 recurs at ~9 other call sites


newhoggy:
## Summary

#1891 fixed the zombie-leak bug (a `write_all` failure on a child's stdin returning early
via `?` before the child was ever waited on) in `tests/common/cargo_run_exit.rs`'s
`spawn_with_signal_retry`. The same bug pattern -- `stdin.write_all(...)?` (or `.expect(...)`)
immediately followed by `wait()`/`wait_with_output()`, with no deferral -- exists independently
at several other call sites across the test suite, none of which route through
`spawn_with_signal_retry`.

Found during #1891's own code review (two independent finder angles, "reuse" and "altitude",
converged on the same list).

## Known sites (not necessarily exhaustive)

1. `tests/jq_cli_tests.rs` (`run_jq_interleaved`, ~line 198-234) -- hand-rolled spawn/retry
   loop whose own doc comment says it mirrors `spawn_with_signal_retry`'s retry policy
   (documented as unable to reuse the function directly, since it needs different
   stdout/stderr wiring for interleaving to one file). Still does `stdin.write_all(...)?`
   directly before `child.wait()?`. Notably, this function was written *after*
   `spawn_with_signal_retry` already existed -- the bug is being freshly reintroduced by
   copy-paste, not just inherited from before the fix existed.
2. `tests/yq_cli_tests.rs` -- `run_yq_stdin`, `run_yq_stdin_with_stderr`,
   `run_yq_stdin_bytes_with_stderr`, `run_jq_stdin_with_stderr` (~lines 24-125) -- four
   near-identical local helpers that already import `cargo_run_exit::exit_code_or_signal_death`
   but hand-roll `cmd.spawn()?` + `stdin.write_all(...)?` + `wait_with_output()?` instead of
   calling `spawn_with_signal_retry`. These also lack the ENOENT/signal-death retry
   `spawn_with_signal_retry` provides -- a second, independent gap alongside the zombie leak.
3. `tests/text_cli_tests.rs` (`run_validate_stdin`, ~lines 30-46) and
   `tests/yaml_validate_tests.rs` (`run_validate_stdin`, ~lines 24-39, and two more call sites
   at ~238-241/~279-282) -- same shape, both files already import from `cargo_run_exit`.
4. `tests/jq_golden_tests.rs` (`run_case`, ~lines 159-174) and `tests/yq_golden_tests.rs`
[...]

PR #2337, opened 2026-09-03
test(yq): route remaining inline spawn/write_all/wait_with_output sites through the shared helpers (#2054)

### LINKED ISSUE #2054: test: ~17 more spawn/write_all/wait_with_output sites in yq_cli_tests.rs need the #2016 fix
## Summary

#2016 fixed the #1891 zombie-leak pattern (`write_all(...)?` returning
early before the child is ever waited on) at 11 named call sites across 7
test files, plus 3 more found during that PR's own two rounds of code
review (`jq_golden_tests.rs::run_case`, `yq_golden_tests.rs::run_case_with_input`,
`yq_path_consistency_tests.rs::run_path`, all migrated to
`spawn_with_signal_retry`) and one independent, more severe instance
(`jq_cli_tests.rs::test_default_identity_filter`, which never called
`wait()` at all, leaking unconditionally rather than only on a write
failure).

Code review on that same PR found the sweep was not exhaustive: it
converted every call site that went through a handful of shared per-file
helper functions, but did not grep for the underlying
`spawn()`+`write_all()`+`wait_with_output()` shape itself. At least ~17
more instances of that exact pattern remain, all in `tests/yq_cli_tests.rs`,
embedded directly in individual `#[test]` function bodies rather than
routed through a shared helper -- so the existing per-file helpers
(`run_yq_stdin`, etc.) don't cover them.

## Known sites (from the code-review finding; re-verify line numbers against
current `main` before touching, per this whole issue family's own
established practice)

- `test_from_file_with_stdin` (~8478)
- `test_exit_status_prints_no_matches_found_to_stderr` (~8737)
- `test_exit_status_no_stderr_message_when_truthy` (~8758)
- `test_jq_mode_computed_float_formatting_unaffected_by_997` (~11055)
- `test_713_jq_mode_unaffected` (~13824, two separate inline spawn blocks)
- `test_jq_tostring_computed_float_unaffected_1054` (~16824)
- (~17449 -- context not captured in the original finding, re-locate via
  grep)
- `test_slice_compound_add_array_target_jq_mode_unaffected_1142` (~18283)
- `test_field_index_assign_scalar_jq_mode_unaffected_1181` (~18454 --
  inside a `for` loop, so the vulnerable sequence runs once per filter)
- `test_yq_iterate_pipe_chain_null_autovivifies_under_update_assign_1919`'s
  jq-mode tail (~18596)
- `test_yq_iterate_assign_scalar_jq_mode_unaffected_1181` (~18631 -- also
  inside a `for` loop)
- `test_jq_urid_base64d_still_reject_non_string_1109` (~19264, two
  separate inline spawn blocks)
- `test_slice_assign_scalar_computed_bound_jq_mode_unaffected_1117`
  (~21801)
- `test_jq_base64d_does_not_trim_whitespace_1123` (~21886)
- `test_jq_gsub_zero_width_unaffected_by_1255` (~23887)

## Suggested fix direction

Same as #2016's own: route each through `spawn_with_signal_retry`
(`tests/common/cargo_run_exit.rs`) directly, or through this file's own
`run_jq_stdin_with_stderr`/`run_yq_stdin_with_stderr` helpers where the
call shape matches (most of these are "jq mode unaffected" negative
controls that likely fit one of the existing helpers rather than needing
a bespoke inline spawn at all -- worth checking whether some of these
should just call the existing helper instead of hand-rolling their own
spawn).

## A structural observation from the same review round (not scoped here)

The recurrence itself (11 sites -> 3 more -> 1 more -> ~17 more, all the
same root shape) suggests this bug class has the same "no compiler signal
forces the fix" problem #2034 already named for the unrelated
`to_owned_checked`/`optional` bug class. Two ideas floated during review,
worth considering together with #2034 rather than acted on ad hoc:

- Parameterize `spawn_with_signal_retry` to accept a `Stdio` builder (not
  just hardcode piped stdout/stderr), so `run_jq_interleaved`'s
  file-redirected-output case could also route through it instead of
  hand-rolling its own retry loop.
- A wrapper type that owns the `Child` from spawn time and only exposes a
  `write_and_wait(self, ...)` method, never the raw `Child`/`.stdin`, so
  the unsafe write-before-wait sequencing becomes a compile error rather
  than a review-time judgment call.

newhoggy:
## Summary

#2016 fixed the #1891 zombie-leak pattern (`write_all(...)?` returning early before the child is ever waited on) at every call site that went through a shared per-file helper, but didn't grep for the underlying `spawn()`+`write_all()`+`wait_with_output()` shape itself. This PR closes that gap: 14 more instances of that exact pattern were hand-rolled inline inside individual `#[test]` bodies in `tests/yq_cli_tests.rs`.

## What changed

Re-derived the site list from scratch via `grep -n '\.spawn()' tests/yq_cli_tests.rs` rather than trusting the issue's own line-number list verbatim (per this issue family's own established practice — line numbers drift). Found 16 raw `.spawn()` sites total:

- **14 sites** fit the existing `run_jq_stdin_with_stderr`/`run_yq_stdin_with_stderr` helpers exactly (both already route through `spawn_with_signal_retry`) — converted directly, several also simplifying away now-redundant manual `String::from_utf8`/exit-code-classification boilerplate the helpers already do.
- **1 site** (`test_from_file_with_stdin`) needed a bespoke `spawn_with_signal_retry` call instead — `--from-file <path>` has no filter positional at all, so it doesn't fit the filter-string helper shape.
- **1 site** (`test_block_scalar_style_preserved_through_slurp`) was left untouched — its `spawn()` never writes to stdin at all (file-only input via a positional arg), so it was never vulnerable to this leak class in the first place.

## Test plan
- [x] Ran all 15 touched tests individually before the full suite, to isolate any regression before it could be masked by the full run

## Not in scope

The issue's own "structural observation" section (parameterizing `spawn_with_signal_retry` to accept a `Stdio` builder, or a wrapper type that makes write-before-wait a compile error rather than a review-time judgment call) is explicitly deferred there as a separate, larger design question — not attempted here.

Fixes #2054

### DIFF
--- a/tests/yq_cli_tests.rs
+++ b/tests/yq_cli_tests.rs
@@ -8851,21 +8851,23 @@ fn test_from_file_with_stdin() -> Result<()> {
     let mut filter_file = NamedTempFile::new()?;
     writeln!(filter_file, ".name")?;
 
-    let mut cmd = Command::new(env!("CARGO_BIN_EXE_succinctly"))
-        .arg("yq")
-        .arg("--from-file")
-        .arg(filter_file.path())
-        .stdin(Stdio::piped())
-        .stdout(Stdio::piped())
-        .stderr(Stdio::piped())
-        .spawn()?;
+    // #2054: bespoke `spawn_with_signal_retry` call, not one of the
+    // filter-string helpers above -- `--from-file <path>` has no filter
+    // positional at all. See `run_yq_stdin`'s own #2016 doc comment for why
+    // this can't be a hand-rolled `spawn()` + `write_all(...)?` +
+    // `wait_with_output()` sequence. `build` has no `'static` bound (called
+    // locally, in a loop), so it borrows `filter_file` directly rather than
+    // cloning its path first.
+    let (output, code) = spawn_with_signal_retry(
+        || {
+            let mut command = Command::new(env!("CARGO_BIN_EXE_succinctly"));
+            command.arg("yq").arg("--from-file").arg(filter_file.path());
+            command
+        },
+        Some(b"name: Bob\n"),
+    )?;
 
-    if let Some(mut stdin) = cmd.stdin.take() {
-        stdin.write_all(b"name: Bob\n")?;
-    }
-
-    let output = cmd.wait_with_output()?;
-    assert_eq!(output.status.code(), Some(0));
+    assert_eq!(code, 0);
     assert_eq!(String::from_utf8(output.stdout)?, "Bob\n");
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


newhoggy:
Three-angle code review came back:

- **Correctness**: no semantic drift in any converted assertion — filter strings, stdin bytes, extra args, and exit-code/stdout/stderr checks all preserved. Flagged two intentional, already-passing strengthenings (an added exit-code assertion in one test that previously had none; the strict-vs-lossy UTF-8 decode difference that's an inherent, harmless consequence of using the pre-existing `run_jq_stdin_with_stderr` helper's own established contract) — neither is a defect.
- **Altitude/conventions**: confirmed the fix depth is correct (the actual leak fix lives in `spawn_with_signal_retry`, already established by #2016; this PR just stops bypassing it) and the bespoke `--from-file` closure is justified rather than premature under-abstraction. Caught a real one: **the commit subject was 106 characters, over this repo's 100-char CI lint limit** — amended to 95.
- **Reuse/simplification/efficiency**: applied the two cheap, real findings — the bespoke `--from-file` closure now borrows the temp file's path directly instead of cloning it, and the two-block `test_jq_urid_base64d_still_reject_non_string_1109` is now a loop, matching this file's own established convention. The "no plain `run_jq_stdin` (no-stderr) helper, unlike `run_yq_stdin`" efficiency gap is real but pre-existing (confirmed present at several call sites before this PR too) — filed as #2338 rather than expanding this mechanical-cleanup PR's own scope.

Full local verification (build, both clippy legs, full test suite, no_std, doc) all green after the fixes.

Issue #1897, opened 2026-08-30
jq: eval_index_expr's Targets::Borrowed error-path prefix has the same unchecked-to_owned bug


newhoggy:
**Severity: Low** (narrow trigger — a computed multi-key index expression where an earlier, still-borrowed target is undecodable and a later key's own lookup errors)

## Summary

Found during #1832's code review (which fixed the identical bug shape in
`eval_comma`/`eval_fanout`/`eval_pipe`'s terminal-control prefix
construction, plus `eval_fanout`'s own main promotion path). A fourth,
independently-discovered instance of the same unchecked-`to_owned`
silent-corruption bug (#1746/#1755/#1790/#1832's shared root cause) exists
in `eval_index_expr`'s `Targets::Borrowed` arm.

`out: Vec<StandardJson<'a, W>>` accumulates borrowed index results across a
`keys x targets` loop. When a later key's `index_one` call returns
`QueryResult::Error`, the code returns `partial(out.iter().map(to_owned).
collect(), Control::Error(e))` — unchecked `to_owned`, not
`promote_borrowed_checked` — and the `pending_halt` flush nearby has the
same defect.

## Repro (verified live against the public `succinctly::jq::eval` library
API, the same evaluator #1832 patches, via the existing `query!` test
macro)

```
$ query!(b"{\"a\": \"\xff\xfe\", \"b\": 5}", ".[(\"a\", 5)]", ...)
QueryResult::Partial([String("")], Error("Cannot index object with number"))
```

Real jq's own analog doesn't exist (eager UTF-8 validation at parse time),
but succinctly's own established invariant (#1746 onward) is that an
undecodable string reaching this evaluator via any lazy/borrowed
accumulation path must raise `EvalError::decode_failure`, not silently
substitute `""`. Here it silently substitutes `""` instead.

## Discovery context

`to_owned_checked`'s own doc comment (`src/jq/eval.rs`) already documents
that roughly 50 more call sites in this file share the general unchecked-
`to_owned` shape, explicitly deferred as a documented gap rather than
attempted in one sweep. This is one specific, concretely-reproduced
instance of that broader population, found adjacent to #1832's own fix
rather than from a full audit.

## Suggested fix direction

Same pattern as #1832's fix: route `out`'s conversion through
`promote_borrowed_checked` (or the shared `resolve_terminal_prefix`/
`promote_and_extend` helpers #1832 introduced, if the surrounding loop
[...]

Issue #2326, opened 2026-09-03
jq: E[K]'s key stream K still discards its own within-generator Partial prefix (#2226 sibling)


newhoggy:
**Severity: Low** (side-effect/streamed-value count only, same shape as #2226 but on the *key* side instead of the *target* side)

## Summary

#2226 fixed `eval_index_expr`/`eval_slice_expr`'s **target** (`E`) evaluation to keep
`E`'s own already-produced values when `E` itself is a multi-output generator that
errors/breaks partway through, in both `src/jq/eval.rs` and `src/jq/eval_generic.rs`.

A structurally identical gap survives on the other side of `E[K]`: `eval_index_expr`'s
**key** (`K`) stream also discards its own already-produced prefix on `Error`/`Break`,
in both files. This is explicitly documented as deliberate/conservative in both
functions' own source comments (not accidental):

```rust
// Computed indexing's key/target forking isn't part of #400/#494's
// verified semantics — conservatively matching the existing
// Error/Break arms rather than inventing new partial-key behavior.
QueryResult::Partial(_, Control::Error(e)) => return QueryResult::Error(e),
QueryResult::Partial(_, Control::Break(label)) => return QueryResult::Break(label),
```

(`eval.rs::eval_index_expr`, and the identically-worded sibling in
`eval_generic.rs::eval_index_expr`.)

Note `Halt` is *already* handled correctly here via `pending_halt` (#791) — this issue
is specifically about the `Error`/`Break` arms.

## Repro (confirmed live against jq 1.7.1)

```console
$ echo '[10,20,30]' | jq -c '.[(1,error("x"))]'
20
jq: error (at <stdin>:1): x

$ echo '[10,20,30]' | succinctly jq -c '.[(1,error("x"))]'
jq: error (at <stdin>:1): x
# 20 is lost -- K's own generator produced it before erroring
```

Same shape with `break`:

```console
$ echo '[10,20,30]' | jq -c 'label $out | .[(1,break $out)]'
20

$ echo '[10,20,30]' | succinctly jq -c 'label $out | .[(1,break $out)]'
# (nothing -- 20 is lost)
```

## Root cause

In both files' `eval_index_expr`, the *keys*-evaluation match discards the `Partial`
prefix instead of indexing the target with each already-produced key and folding the
successes into the running accumulator before propagating the control — the same fix
#2226 applied to the *target*-evaluation match, just on the other operand.

## Scope note

[...]

PR #2339, opened 2026-09-04
fix(jq): escape_generic! keeps the convertible prefix on a double fault

### LINKED ISSUE #2145: jq: eval_generic's escape_generic! discards more of the prefix than eval.rs's resolve_terminal_prefix on a double-fault
**Severity: Low** (double-fault edge case: needs a cursor conversion failure *while already
folding a Partial's running prefix*, itself triggered by a separate earlier error/break/halt).

## Summary

Surfaced during code review of #2142 (the #2032 fix). Not a regression from that PR --
confirmed the same gap existed pre-#2032 too (the old \`owned_or_err!(to_owned_all_cursors(
&cursors))\` bare-returned on this same failure, discarding even more: the original
triggering error itself, not just the prefix).

\`eval_generic.rs\`'s \`escape_generic!\` macro (\`src/jq/eval_generic.rs\`, defined inside
\`eval_index_expr\`) discards the *entire* accumulated prefix if \`to_owned_all_cursors(&cursors)\`
itself fails while folding the running prefix into a \`Partial\`:

\`\`\`rust
macro_rules! escape_generic {
    (\$control:expr) => {{
        let control = \$control;
        let out = if any_owned {
            owned
        } else {
            match to_owned_all_cursors(&cursors) {
                Ok(vs) => vs,
                Err(e) => {
                    let control = match control {
                        Control::Halt(code) => Control::Halt(code),
                        Control::Error(_) | Control::Break(_) => Control::Error(e),
                    };
                    return partial_generic(Vec::new(), control);  // <-- prefix dropped
                }
            }
        };
        return partial_generic(out, control);
    }};
}
```

\`eval.rs\`'s equivalent path (\`resolve_terminal_prefix\` + \`promote_borrowed_checked\`, the
#1897 fix) preserves whatever *did* convert successfully before the failure instead of
discarding the whole thing.

## Where

- \`escape_generic!\` in \`src/jq/eval_generic.rs\` (inside \`eval_index_expr\`, added/touched by
  #2142; the underlying discard-on-double-fault behavior itself predates #2032).

## Suggested approach

Not investigated in depth. Likely wants the same partial-preservation treatment
\`resolve_terminal_prefix\`/\`promote_borrowed_checked\` already give \`eval.rs\`'s equivalent path
-- converting each cursor individually and keeping whatever prefix succeeded, rather than an
all-or-nothing \`to_owned_all_cursors\` call whose own failure loses everything already
collected. Worth checking whether \`eval_generic.rs\` has (or should share) a helper analogous
to \`resolve_terminal_prefix\` for this, rather than a bespoke fix local to this one macro.

## Related

- #2032 / #2142 -- the PR whose review surfaced this.

newhoggy:
## Summary
- `eval_generic.rs`'s `eval_index_expr` (its `escape_generic!` macro) discarded the *entire* already-accumulated `cursors` prefix if converting it to owned values itself failed with its own decode error — a "double fault": the failure converting the prefix, on top of the separate error/break/halt that triggered folding it into a `Partial` in the first place.
- `eval.rs`'s equivalent path (`resolve_terminal_prefix` + `promote_borrowed_checked`, #1897) already preserves whatever *did* convert successfully before the failure; this brings `eval_generic.rs` in line.
- Adds `to_owned_all_cursors_checked`, the prefix-preserving sibling of `to_owned_all_cursors` (which unconditionally drops everything on failure via `.collect::<Result<Vec<_>, _>>()`'s short-circuit), and switches `escape_generic!`'s own promotion step to it.
- Not a jq/yq mode-fidelity issue like #2226 — a pure robustness/correctness fix, applies identically to both modes (matches `resolve_terminal_prefix`, which is likewise unconditional).

Fixes #2145

## Follow-up filed
`ensure_owned!` (the sibling promotion macro a few lines below `escape_generic!`, used by the `KeyTargets::Native`/`Owned` success arms) has the *same* underlying flaw via its own `owned_or_err!`-based bare-return — already documented in its own comment as "pre-existing, unchanged by #2032... not new territory this fix needs to harden." It's strictly worse than what this PR fixes: it discards the held `control` entirely (not just the prefix), including an uncatchable `Halt`. Filed as a follow-up, since #2145's own scope was `escape_generic!` specifically.

## Test plan
- [x] `cargo fmt --check`
- [x] New unit test constructs the double-fault live (a genuinely undecodable document element, reached via a multi-output target + a later key that fails to index it), verifies the convertible prefix survives

### DIFF
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -495,6 +495,29 @@ pub fn to_owned_all_cursors<'a, C: DocumentCursor + 'a>(
     cursors.into_iter().map(to_owned_cursor).collect()
 }
 
+/// [`to_owned_all_cursors`]'s prefix-preserving sibling (#2145): on a decode
+/// failure, returns whatever converted successfully *before* the failing
+/// cursor instead of discarding it -- `.collect::<Result<Vec<_>, _>>()`
+/// (what `to_owned_all_cursors` uses) drops every already-collected item on
+/// the first `Err`, which is fine for a caller with nothing of its own to
+/// lose, but wrong for one folding this conversion's result into an
+/// already-running prefix (`escape_generic!`'s own `Partial`-folding use,
+/// its sole caller so far). Mirrors `eval.rs`'s `promote_borrowed_checked`
+/// (#1897) -- the same fix, for the cursor-native accumulator instead of
+/// the borrowed-`StandardJson` one.
+fn to_owned_all_cursors_checked<'a, C: DocumentCursor + 'a>(
+    cursors: impl IntoIterator<Item = &'a C>,
+) -> Result<Vec<OwnedValue>, (Vec<OwnedValue>, EvalError)> {
+    let mut acc = Vec::new();
+    for c in cursors {
+        match to_owned_cursor(c) {
+            Ok(v) => acc.push(v),
+            Err(e) => return Err((acc, e)),
+        }
+    }
+    Ok(acc)
+}
+
 fn to_owned_cursor_at_depth<C: DocumentCursor>(
     cursor: &C,
     depth: usize,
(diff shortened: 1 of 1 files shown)


newhoggy:
## Code review results

Confirmed and fixed two sibling sites in the same function (`eval_index_expr`) with the identical double-fault gap the original fix closed in `escape_generic!`:

1. **The target `Partial` arm's own inline promotion** (`if !any_owned { owned = match to_owned_all_cursors(&cursors) {...` around the `GenericResult::Partial(vs, control)` match arm) — still discarded the prefix on a secondary decode failure. Now uses the same `to_owned_all_cursors_checked`.
2. **The `pending_halt` finalizer** — same gap, and worse: it also silently downgraded the already-uncatchable `Halt` into a catchable decode error via `owned_or_err!`. Now uses `to_owned_all_cursors_checked` and always keeps `Halt` (the control here is `Halt` by construction, so a promotion failure can only ever keep it — matching `resolve_terminal_prefix`'s established rule).

Also applied an efficiency finding: pre-sizes `to_owned_all_cursors_checked`'s accumulator via `vec_with_capacity` (matching its sibling `promote_borrowed_checked`'s own convention), narrowing the signature from a generic `IntoIterator` to `&[C]` so the caller's length is available.

Attempted to add a live regression test for the `pending_halt` site specifically, using a document with genuinely undecodable content (invalid UTF-8) alongside `halt` in the key stream. Debug-probed the actual evaluation and found `Builtin::Halt`'s own evaluation bridges through `bridge_to_full_evaluator`, which materializes the *entire* document up front — so any undecodable content anywhere in the document surfaces there first, as the key stream's own already-tracked (#2326) prefix-discard gap, before `pending_halt` is ever set. Documented this finding in the code instead of shipping a test that doesn't actually exercise the fixed branch (or a flaky one). The other unfixed sibling (the target `Partial` arm) has an analogous, already-documented reachability constraint from #2226's own review.

All local verification (build/clippy ×2/fmt/doc/test ×3 feature tiers) re-run clean, rebased onto latest `main` with no conflicts.

PR #2341, opened 2026-09-04
test(yq): add run_jq_stdin helper, sweep unused-stderr call sites (#2338)

### LINKED ISSUE #2338: test: yq_cli_tests.rs has no plain run_jq_stdin (no-stderr) helper, unlike run_yq_stdin
## What's wrong

Found during #2054's own code review. `tests/yq_cli_tests.rs` has `run_yq_stdin`
(stdout + exit code only) alongside `run_yq_stdin_with_stderr` (adds stderr) — but for
jq mode there's only `run_jq_stdin_with_stderr`, no plain `run_jq_stdin` counterpart.

Every jq-mode test that doesn't actually need stderr (most of the ~14 sites #2054
converted, and plenty of pre-existing ones) ends up calling `run_jq_stdin_with_stderr`
anyway and discarding the result as `_stderr` — paying for a `String::from_utf8` decode
of the child's stderr buffer on every call for no reason. Some go further and discard
*both* halves (`_out`, `_stderr`), keeping only `code` — two wasted decodes per call.

This isn't a regression from #2054 (the same discard pattern already existed at several
pre-#2054 call sites, e.g. lines 18634/18714-18717/22571 on `main` before that PR) — it's
a pre-existing, systemic gap in the file's own helper set that #2054's mechanical
site-by-site conversion inherited and propagated rather than introduced, since adding a
new shared helper was out of scope for "route existing broken sites through the
*existing* helpers."

## Suggested fix

Add `run_jq_stdin(filter: &str, input: &str, extra_args: &[&str]) -> Result<(String, i32)>`
mirroring `run_yq_stdin`'s own shape exactly, then sweep the file for
`run_jq_stdin_with_stderr(...)` call sites whose `stderr` binding is `_`-prefixed
(unused) and switch them to the cheaper helper. Low risk, mechanical, similar shape to
#2054 itself.

newhoggy:
## Summary

`run_yq_stdin_with_stderr` already had a no-stderr sibling (`run_yq_stdin`) for yq mode; jq mode didn't, so every jq-mode test that doesn't need stderr paid for decoding and discarding it anyway via `run_jq_stdin_with_stderr`. Found during #2054's own code review — filed as #2338, not fixed there since it was a pre-existing, systemic pattern rather than something that PR introduced.

## What changed

Added `run_jq_stdin(filter, input, extra_args) -> Result<(String, i32)>`, mirroring `run_yq_stdin`'s exact shape, then swept every `run_jq_stdin_with_stderr` call site whose `stderr` binding was already `_`-prefixed (i.e. genuinely unused) over to the cheaper helper — 15 sites (via a small script matching `let (X, _stderr, Y) = run_jq_stdin_with_stderr(` → `let (X, Y) = run_jq_stdin(`, verified no other shape slipped through).

## Test plan
- [x] `cargo test --test yq_cli_tests --features cli` — full file green (1400 tests)

Fixes #2338

### DIFF
--- a/tests/yq_cli_tests.rs
+++ b/tests/yq_cli_tests.rs
@@ -108,6 +108,25 @@ fn run_jq_stdin_with_stderr(
     Ok((stdout, stderr, exit_code))
 }
 
+/// jq-mode counterpart of `run_yq_stdin` -- #2338: `run_yq_stdin_with_stderr`
+/// already had a no-stderr sibling for yq mode; jq mode didn't, so every
+/// jq-mode test that doesn't need stderr paid for decoding and discarding it
+/// anyway via `run_jq_stdin_with_stderr`. See `run_yq_stdin`'s own #2016 doc
+/// comment above for why this can't be a hand-rolled `spawn()` +
+/// `write_all(...)?` + `wait_with_output()` sequence.
+fn run_jq_stdin(filter: &str, input: &str, extra_args: &[&str]) -> Result<(String, i32)> {
+    let (output, exit_code) = spawn_with_signal_retry(
+        || {
+            let mut command = Command::new(env!("CARGO_BIN_EXE_succinctly"));
+            command.arg("jq").args(extra_args).arg(filter);
+            command
+        },
+        Some(input.as_bytes()),
+    )?;
+    let stdout = String::from_utf8(output.stdout)?;
+    Ok((stdout, exit_code))
+}
+
 /// Helper to run yq command with file input
 fn run_yq_file(filter: &str, file_path: &str, extra_args: &[&str]) -> Result<(String, i32)> {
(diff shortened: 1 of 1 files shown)


newhoggy:
Code review found the regex-based sweep was too narrow: it only matched a single-line `let (X, _stderr, Y) = run_jq_stdin_with_stderr(` shape, missing two others — 10 sites rustfmt had wrapped across two lines, and 3 sites where the unused binding was named bare \`_\` rather than \`_stderr\`. All 13 confirmed genuinely dead (function bodies read individually, and \`cargo check\` compiles clean with the binding removed — it would have failed to compile if any secretly referenced \`stderr\`). Converted all 13, ran \`cargo fmt\` to reflow the now-shorter lines, and re-verified: full file (1400 tests), full suite, both clippy legs, fmt, no_std, doc all green.

Issue #2328, opened 2026-09-03
jq: E[K]_with_path_context/E[S:T]_with_path_context still discard E's own within-generator Partial prefix (#2226 sibling)


newhoggy:
**Severity: Low** (side-effect/streamed-value count only, same shape as #2226 but in the path-context/assignment evaluator instead of the plain-read evaluator)

## Summary

#2226 fixed `eval_index_expr`/`eval_slice_expr`'s target (`E`) evaluation to keep `E`'s
own already-produced values (in jq mode) when `E` itself is a multi-output generator
that errors/breaks/halts partway through, in both `src/jq/eval.rs` and
`src/jq/eval_generic.rs`.

A structurally identical gap survives in the *path-context* siblings of those two
functions -- `eval_index_expr_with_path_context` and `eval_slice_expr_with_path_context`
(both in `src/jq/eval.rs`), which back assignment (`E[K] = v`, `E[K] |= f`), `del()`, and
any pipe ending in a path-context-forcing builtin (`key`, `parent`, `file_index`) through
a computed index/slice. Their own `QueryResult::Partial(_, control)` arms still discard
the target's own already-produced prefix outright, exactly the bug #2226 fixed in the
plain-read evaluators.

## Repro (confirmed live against jq 1.7.1)

Forcing path-context routing via a trailing `| key`:

```console
$ echo '{"a":[1,2],"b":[3,4]}' | jq -c '(.a,.b,error("x"))[(0+0)] | key'
0
0
jq: error (at <stdin>:1): x

$ echo '{"a":[1,2],"b":[3,4]}' | succinctly jq -c '(.a,.b,error("x"))[(0+0)] | key'
jq: error (at <stdin>:1): x
# 0, 0 lost -- the target's own generator produced two indexable values
# before erroring, same shape #2226 fixed for the plain-read case
```

Same shape for the slice sibling:

```console
$ echo '{"a":[1,2],"b":[3,4]}' | jq -c '(.a,.b,error("x"))[(0+0):(1+0)] | key'
0
0
jq: error (at <stdin>:1): x

$ echo '{"a":[1,2],"b":[3,4]}' | succinctly jq -c '(.a,.b,error("x"))[(0+0):(1+0)] | key'
jq: error (at <stdin>:1): x
```

## Root cause

`eval_index_expr_with_path_context`'s and `eval_slice_expr_with_path_context`'s own
`QueryResult::Partial(_, control)` arms (`src/jq/eval.rs`) discard the underscored
[...]

PR #2348, opened 2026-09-04
fix(jq): E[K]'s key stream keeps its own Partial prefix on Error/Break (#2326)

### LINKED ISSUE #2326: jq: E[K]'s key stream K still discards its own within-generator Partial prefix (#2226 sibling)

newhoggy:
## Summary

#2226 fixed `E[K]`'s **target**-evaluation `Partial` arm to index each already-produced target value instead of discarding them on a mid-stream `Error`/`Break` (`0 as \$k | ([1,2],[3,4],error("x"))[\$k]` now streams `1`, `3` before raising). The mirror-image gap survived on the other side: the **keys**-evaluation `Partial` arm still discarded `key`'s own already-produced prefix, in both `eval.rs` and `eval_generic.rs`.

```console
$ echo '[10,20,30]' | jq -c '.[(1,error("x"))]'
20
jq: error (at <stdin>:1): x

$ echo '[10,20,30]' | succinctly jq -c '.[(1,error("x"))]'   # before this fix
jq: error (at <stdin>:1): x
# 20 was lost
```

## What changed

Mirrors #2226's own fix shape exactly, on the key side: when the key generator's evaluation returns `Partial(prefix, control)`, index the target once per key in `prefix` (the same per-key operation the success arms already use), fold the successes into the running accumulator, then propagate `control` — instead of discarding the prefix outright. Threaded through a new `pending_key_stream_control` variable (mutually exclusive with the existing `pending_halt`, since a single `Partial` carries exactly one `Control` variant), resolved at the same tail point `pending_halt` already is.

jq-mode only, gated the same way #2226's own fix is (`S::TAG != EvalTag::Yq`): real yq doesn't stream this prefix either (confirmed live, `[10,20,30] | .[(1,error("x"))]` in yq mode shows no `20`).

Applied to both evaluators — `eval.rs`'s own `eval_index_expr` (the `-n`/literal-array path) and `eval_generic.rs`'s own `eval_index_expr` (the ordinary stdin/file-read path, confirmed to be the one an actual CLI invocation hits per that file's own doc comment) — both needed the identical fix, verified independently against the oracle.

## Pre-existing tests corrected

Three tests had pinned the old discard behavior as expected:
- Two in `src/jq/eval.rs`'s own test module **explicitly documented** `.[(1,error/break \$out)]` as "a distinct, still-open gap... tracked separately" — i.e. this exact issue. Removed from the "collapses to bare control" tables and migrated into the "prefix survives" coverage, right alongside #2226's own analogous entries (same file, same precedent).
- One in `tests/jq_cli_tests.rs` (`test_computed_index_still_conservative_on_error_and_break`) asserted the discard "matches real jq's own gap here" — a claim that was never actually checked against the oracle. It doesn't: verified live that real jq streams the prefix before erroring, the same as it already does for `halt` (covered by the neighboring, already-correct `test_computed_index_streams_keys_produced_before_halt`). Renamed and corrected.

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` — full suite green (after correcting the 3 pre-existing tests above)
- [x] Live-verified every shape (single key, multiple keys before error, `break`, an earlier indexing failure outranking a later key-stream escape, `?`) against `jq` 1.7.1 and `yq` v4.53.3, byte-for-byte match
- [x] Confirmed jq mode and yq mode both match their respective oracles

Fixes #2326

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -16718,6 +16718,10 @@ fn eval_index_expr<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     // the key generator halts on its *next* attempt, so those keys' indexed
     // output must still reach stdout (#791).
     let mut pending_halt = None;
+    // #2326: mutually exclusive with `pending_halt` -- a single `Partial`
+    // result carries exactly one `Control` variant, so at most one of the
+    // two is ever set.
+    let mut pending_key_stream_control: Option<Control> = None;
     let keys = match eval_single::<W, S>(key, value.clone(), false).materialize_cursor() {
         QueryResult::One(v) => match to_owned_key_shape(&v) {
             Ok(v) => vec![v],
@@ -16738,9 +16742,28 @@ fn eval_index_expr<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
         QueryResult::Break(label) => return QueryResult::Break(label),
         QueryResult::Halt(code) => return QueryResult::Halt(code),
         QueryResult::OneCursor(_) => unreachable!("materialize_cursor should have converted this"),
+        // #2326: `key`'s own generator produced `vs` before its own
+        // mid-stream escape -- mirrors #2226's identical fix for `target`'s
+        // side (this same function's own `KeyTargets::Partial` arm below):
+        // real jq's key-outer/target-inner model indexes each already-
+        // produced key as it flows out, so `key`'s own escaped generator's
+        // prefix is real jq output too (confirmed live: `[10,20,30] |
+        // .[(1,error("x"))]` prints `20` before raising). jq-only, matching
+        // #2226's own gate -- real yq does not stream this prefix either
+        // (confirmed live, `.[(1,error("x"))]` in yq mode shows no `20`),
+        // so yq mode keeps the pre-existing conservative discard below.
+        QueryResult::Partial(vs, Control::Error(e)) if S::TAG != EvalTag::Yq => {
+            pending_key_stream_control = Some(Control::Error(e));
+            vs
+        }
+        QueryResult::Partial(vs, Control::Break(label)) if S::TAG != EvalTag::Yq => {
+            pending_key_stream_control = Some(Control::Break(label));
+            vs
+        }
         // Computed indexing's key/target forking isn't part of #400/#494's
         // verified semantics — conservatively matching the existing
         // Error/Break arms rather than inventing new partial-key behavior.
+        // (yq mode, and any future non-jq/yq semantics, still take this.)
         QueryResult::Partial(_, Control::Error(e)) => return QueryResult::Error(e),
         QueryResult::Partial(_, Control::Break(label)) => return QueryResult::Break(label),
         QueryResult::Partial(vs, Control::Halt(code)) => {
@@ -16750,7 +16773,8 @@ fn eval_index_expr<'a, W: Clone + AsRef<[u64]>, S: EvalSemantics>(
     };
     if keys.is_empty() {
         // `partial`'s invariant (a non-empty prefix by construction) means
-        // this is only reachable with `pending_halt` unset.
+        // this is only reachable with `pending_halt`/`pending_key_stream_control`
+        // both unset.
         return QueryResult::None;
     }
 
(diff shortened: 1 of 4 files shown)


newhoggy:
Code review found 3 things:

1. **Real bug in my own new code**, fixed: \`eval_generic.rs\`'s new tail block used the plain \`owned_or_err!(to_owned_all_cursors(...))\` instead of the prefix-preserving \`to_owned_all_cursors_checked\` this same file's own \`pending_halt\` block and \`escape_generic!\` already establish as the rule here — a secondary decode failure while converting the prefix would have silently discarded the whole already-indexed prefix and replaced the held control's own message. Now matches that established pattern exactly.
2. A claimed gap in \`eval_index_expr_with_path_context\`'s own key-side handling — investigated, did **not** reproduce against a legitimate jq trigger (\`path(...)\`, a real assignment) once re-tested. The repro technique both this finding and #2328 (the pre-existing "sibling" issue) relied on turned out to use \`key\`, which is a **yq-only** builtin real jq rejects outright with a compile error — commented on #2328 to flag that its own "confirmed live against jq 1.7.1" claim doesn't actually reproduce.
3. A genuinely real, **pre-existing** (confirmed identical on \`main\` before this PR — unrelated function, slice bounds not index keys) gap: \`eval_slice_bound\` has no yq-mode gate at all, so it streams a slice bound's own escaped prefix in yq mode where real yq doesn't. Filed as #2351.

Full local verification (build, both clippy legs, full test suite, no_std, doc) all green after the fix.

Issue #2353, opened 2026-09-04
yq: wrong-kind delete key (e.g. numeric key against object) errors instead of no-op


newhoggy:
**Severity: Low** (matching jq's own error vs. real yq's silent no-op — no data loss either
way, just a wording/exit-code divergence)

## Summary

Found while implementing #2306. Real yq treats a wrong-*kind*-of-key against an
array/object in `del()`/`delpaths()` as a silent no-op, not an error — succinctly's yq
mode inherits jq's own "raise `Cannot delete <kind> field of <kind>`" behavior instead.

## Repro (confirmed live against yq v4.53.3)

```console
$ echo '{"a":1}' | yq -o=json 'delpaths([[5]])'
{
  "a": 1
}

$ echo '{"a":1}' | succinctly yq -o json 'delpaths([[5]])'
Error: Cannot delete number field of object

$ echo '{"a":1}' | yq -o=json 'del(.[5])'
{
  "a": 1
}

$ echo '{"a":1}' | succinctly yq -o json 'del(.[5])'
Error: Cannot delete number field of object
```

Real jq raises the same error succinctly does today (`jq -c 'delpaths([[5]])'` on
`{"a":1}` errors `Cannot delete number field of object`), so succinctly's current
behavior matches jq exactly — it's specifically yq mode that needs its own, more
permissive rule.

## Root cause

`delete_keys`'s (`src/jq/eval.rs`) `OwnedValue::Object` arm's key-type check —

```rust
other => return Err(EvalError::cannot_delete_field_of_object(other.type_name())),
```

— and the `OwnedValue::Array` arm's analogous `cannot_delete_element_of_array` — both
raise unconditionally, with no `yq_mode` carve-out. Whether this rule is "any wrong-kind
key is a no-op" universally, or something narrower (e.g. only for a *numeric* key against
an object / *string* key against an array, the two shapes checked above) hasn't been
explored beyond this one pair of repros.

## Scope

- Confirm the full matrix: wrong-kind key against an object with existing fields vs. an
  empty object; wrong-kind key against a non-empty vs. empty array; a genuinely
  unsupported key type (e.g. `null`, a bool) vs. the "just the wrong shape for this
  container" case above.
- Confirm whether this also applies to `del()`'s own comma-form and to `setpath`/direct
  assignment through a wrong-kind key (a *read*, `.a[5]` on an object, already returns
  `null` rather than erroring in both jq and yq — this issue is specifically about
[...]

Issue #2352, opened 2026-09-04
yq: delpaths([[]]) prints literal null instead of producing no output like del(.)


newhoggy:
**Severity: Low** (output-shape divergence only — the deletion itself is correct)

## Summary

Found while implementing #2306. `delpaths()` with an empty path component (`[]`, naming
the whole document) prints the literal string `null` in succinctly's yq mode — real yq
prints *nothing at all* (empty stdout, exit 0), matching `del(.)`'s own already-correct
"deleting the whole document produces no output" behavior. This predates #2306's own
sequential-processing fix entirely; a single-path `delpaths([[]])` call (unaffected by
that fix, since a single path has no ordering to get wrong) already reproduces it.

## Repro (confirmed live against yq v4.53.3)

```console
$ echo '[1,2,3]' | yq -o=json 'delpaths([[]])'
$ echo $?
0
# (nothing printed)

$ echo '[1,2,3]' | succinctly yq -o json 'delpaths([[]])'
null

$ echo '[1,2,3]' | yq -o=json 'del(.)'
$ echo $?
0
# (nothing printed -- matches, and succinctly yq's own del(.) already agrees)

$ echo '[1,2,3]' | succinctly yq -o json 'del(.)'
# (nothing printed -- succinctly already gets *this* one right)
```

So `del(.)` already correctly special-cases "deleted the whole document" as *no output*,
but `delpaths([[]])` (the JSON-array-of-paths spelling of the identical operation)
doesn't share that special case and instead prints the literal owned value `null`.

## Root cause

`delpaths_one`'s (`src/jq/eval.rs`) own empty-path handling —

```rust
Some([]) => Ok(OwnedValue::Null),
```

(and #2306's own sequential-loop counterpart, `v = if path.is_empty() { OwnedValue::Null }
...`) — returns `OwnedValue::Null` as an ordinary owned value, which the caller then
prints as `null`. `builtin_del`'s own root-deletion path apparently has a different,
already-correct "produce nothing" mechanism that `delpaths_one` never picked up.

## Suggested fix direction

Find whatever mechanism `builtin_del`/`eval_update`'s root-deletion case uses to suppress
output entirely (rather than emitting `OwnedValue::Null`) and share it with
`delpaths_one`'s empty-path case, in yq mode. jq mode is unaffected — `del(.)`/
`delpaths([[]])` both already print `null` in real jq, which is what jq mode already
does today.

## Refs

[...]

PR #2354, opened 2026-09-04
fix(yq): delpaths() applies a multi-key batch sequentially, not as a union

### LINKED ISSUE #2306: yq: delpaths() multi-key batch mixing in-range and out-of-range indices is order-dependent in real yq, not replicated
## What's wrong

Follow-up to #2305 (single-key `del()`/`delpaths()` positive-out-of-range
null-extension) -- deliberately deferred from that fix's own scope.

Found while implementing #2305: a `delpaths()` batch mixing an in-range
delete with a positive-out-of-range one is **order-dependent** in real yq
(v4.53.3), confirmed live with the exact same input array:

```console
$ echo '[1,2,3]' | yq -o=json 'delpaths([[0],[5]])'
[2, 3, null, null, null]          # length 5

$ echo '[1,2,3]' | yq -o=json 'delpaths([[5],[0]])'
[2, 3, null, null]                # length 4 -- same keys, different order!
```

Hand-tracing this against a simple "apply each path sequentially,
mutating the array in place between steps" model matches both results
exactly:
- `[[0],[5]]`: delete index 0 from `[1,2,3]` → `[2,3]` (len 2), then extend
  from len 2 to len 5 → `[2,3,null,null,null]` (len 5). Matches.
- `[[5],[0]]`: extend `[1,2,3]` from len 3 to len 5 → `[1,2,3,null,null]`
  (len 5), then delete index 0 → `[2,3,null,null]` (len 4). Matches.

## Why this is separate from #2305

succinctly's `delete_keys` (`src/jq/eval.rs`, backs `delpaths()`'s
terminal single-component deletions) resolves every key against the
array's length **on entry** and applies them all in one batched `retain`
pass, deliberately order-independent -- this is what makes the existing
(already-correct) in-range-only semantics match real jq's own
`delpaths` (`[1,2,3,4] | del(.[0:2], .[1:3])` is `[4]` regardless of
argument order, confirmed in `delete_keys`'s own doc comment). #2305's
fix only extends this batch model for the case of exactly one key in the
whole batch (where order can't matter), specifically to avoid this
order-dependent multi-key case -- see that fix's own comment in
`delete_keys`'s `OwnedValue::Array` arm.

Replicating the real order-dependent behavior for 2+ keys would need
`delete_keys`'s array arm restructured to apply each key **sequentially**
(mutating the array between each), abandoning the current single-pass
`retain` optimization for any batch containing an out-of-range key --
worth its own investigation (including whether it can stay fast for the
common case of zero out-of-range keys) rather than guessing at in this
issue.

## Scope

- Confirm whether this is genuinely "apply paths in the order given, no
  sorting" for the out-of-range case specifically, or something subtler
  (e.g. yq's own path-sort comparator handling an out-of-range index
  differently) -- the sequential-mutation model above matches both
  observed cases, but only two data points have been checked.
- Confirm `del()`'s own comma-separated multi-target form
  (`del(.[0], .[5])`, distinct from `delpaths()`'s path-array form) has
  the same or a different call path/behavior -- not checked yet.
- Whether the same order-dependence applies to object-shaped (slice)
  keys mixed with out-of-range integer keys in the same batch.

## Suggested fix

Investigate whether `delete_keys`'s array arm can apply keys sequentially
without losing its current O(n) `retain`-based performance for the common
(no out-of-range key) case -- e.g. detecting whether any key resolves
`PositiveOutOfRange` up front and only falling back to a slower sequential
path when at least one does.

newhoggy:
## Summary
- Real yq's own `delpaths()` applies a 2+-path batch **sequentially** — one path at a time, each seeing whatever state every earlier path already left behind — not jq's own simultaneous/union model, which resolves every path against the array's *starting* length.
- Confirmed live (yq v4.53.3) this isn't only about out-of-range indices, contrary to this issue's own original framing: even an **all-in-range** batch is order-dependent — `delpaths([[0],[1]])` on `[1,2,3,4]` is `[2,4]`, `delpaths([[1],[0]])` on the same input is `[3,4]` — while real jq's own `delpaths` gives the order-independent `[3,4]` for *both* orderings.
- `delpaths_one` now skips the upfront sort in yq mode (which exists to support jq's own batch semantics) and applies each path individually via a loop over the existing `delete_paths_sorted`, rather than handing the whole (sorted) list to it in one call. **jq mode is unaffected.**
- Corrects two pre-existing tests that pinned the old, wrong batch-based expectation — including `test_delpaths_multi_key_out_of_range_stays_noop_known_gap_2305`, which explicitly documented this exact gap as pending this very issue.

Fixes #2306

## Follow-ups filed
Two out-of-scope divergences surfaced during this fix's own live-oracle sweep:
- #2352 — `delpaths([[]])` prints a literal `null` instead of producing no output the way `del(.)` already correctly does.
- #2353 — a wrong-kind delete key (e.g. a numeric key against an object) raises in succinctly's yq mode; real yq treats it as a silent no-op.

## Test plan
- [x] `cargo fmt --check`
- [x] Live-verified against the pinned yq v4.53.3 oracle across a broad sweep: in-range-only (both orders), pure-OOB (both orders), OOB-mixed (both orders), negative-index-mixed, nested paths, object keys, empty-path-mixed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -577,15 +577,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `yq_negative_index_check`/`yq_negative_index_error` helpers (all five take a plain
   `yq_mode: bool`, not `S: EvalSemantics`). Unlike the read-side fix, not
   suppressible by an earlier `?` in the path chain — confirmed live, real yq's own
-  `del(.a?[-5])` still raises. One residual divergence documented, not chased: when grouped
-  with an earlier same-array deletion in one `delpaths()` call, real yq's own error reports
-  the array's *already-shrunk* size (sequential resolution), where this crate's own
-  `delete_keys` deliberately resolves every key against the array's length on entry (the
-  invariant that makes an overlapping range union rather than compound) — still correctly
-  raises either way, only the reported number can differ, and only in that one
-  multi-key-same-array shape. Also surfaced, and filed separately as #2305 (out of this
-  issue's negative-index scope): real yq's `del()` on a *positive* out-of-range index extends
-  the array with nulls instead of no-op, which succinctly (matching jq) doesn't reproduce.
+  `del(.a?[-5])` still raises. A residual divergence was documented here, not chased at the
+  time: when grouped with an earlier same-array deletion in one `delpaths()` call, real yq's
+  own error reports the array's *already-shrunk* size (sequential resolution), where this
+  crate's own `delete_keys` deliberately resolved every key against the array's length on
+  entry — closed by #2306 below, which now reports the shrunk size too. Also surfaced, and
+  filed separately as #2305 (out of this issue's negative-index scope): real yq's `del()` on
+  a *positive* out-of-range index extends the array with nulls instead of no-op, which
+  succinctly (matching jq) doesn't reproduce.
+- **`delpaths()`'s multi-key batch applied as a union, not sequentially** (#2306): real yq's
+  own `delpaths()` applies a 2+-path batch one path at a time, each seeing whatever state
+  every earlier path already left behind — not jq's own simultaneous/union model this crate's
+  `delete_keys` was built around, which resolves every path against the array's *starting*
+  length. Confirmed live this isn't only about out-of-range indices — even an all-in-range
+  batch is order-dependent (`delpaths([[0],[1]])` on `[1,2,3,4]` is `[2,4]`;
+  `delpaths([[1],[0]])` on the same input is `[3,4]`, where real jq's own `delpaths` gives the
+  order-independent `[3,4]` for both). `delpaths_one` (`src/jq/eval.rs`) now skips its
+  upfront sort in yq mode (the sort exists to support jq's own batch semantics) and applies
+  each path individually instead of handing the whole list to `delete_keys` in one call. jq
+  mode is unaffected. `del()`'s own comma-separated multi-target form shares the same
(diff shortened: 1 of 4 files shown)


newhoggy:
## Code review results

Confirmed and fixed all four findings:

1. **`docs/compliance/yq/limitations.md`** had two stale entries (the negative-index
   "array size on entry" residual divergence, and the delpaths()-batch-order-dependence
   "deliberately not extended" note) still describing gaps this PR closes. Rewrote both,
   and added an explicit "still open: `del()`'s own comma-form" callout since that's a
   genuinely separate, unresolved call path.
2. **`CHANGELOG.md`** — corrected the stale #2268 entry's own claim and added a new
   entry for #2306.
3. **`delete_keys`'s own doc comment** (`src/jq/eval.rs`) still described the gap as
   tracked-but-unaddressed even though `delpaths_one` never reaches it with more than
   one key in yq mode anymore — updated to explain the narrowed reachability (jq mode,
   or `del()`'s own still-affected comma-form).
4. **Missing reversed-order test** for the mixed in-range/out-of-range case — added
   `test_delpaths_multi_key_mixed_reversed_order_gives_a_different_result_2306`, the
   exact `[[5],[0]]` pairing the compliance doc itself uses to demonstrate the divergence.

All local verification (build/clippy ×2/fmt/doc/test ×3 feature tiers) re-run clean, rebased onto latest `main` with no conflicts (main gained an unrelated yq del()/slice-run fix in the same file since the initial push, verified no interaction).

Issue #534, opened 2026-07-31
jq: reduce/foreach/until/while silently misfold multi-output init/update/cond/extract instead of jq's per-output fanout


newhoggy:
**Severity: High** — verified against pinned `jq-1.7.1-apple`, reproduced against the built CLI (release). Silent wrong answers, not errors, in most cases.

## Summary

`reduce`/`foreach`/`until`/`while` all take one or more parenthesized sub-expressions (`INIT`, `UPDATE`, `COND`, `EXTRACT`) that real jq's grammar parses as full `Exp` (comma-inclusive) — and real jq gives each of them specific per-output fanout/fold semantics when the sub-expression produces more than one output. This codebase's evaluator assumes each of these sub-expressions produces exactly one output, so a multi-output value silently misfolds instead of implementing jq's actual rule.

This was previously unreachable because the parser rejected a bare top-level comma in these positions (`expected ')', found ','`). #462 correctly widened the parser to match jq's real grammar elsewhere (`|` as the loosest operator, comma nested inside), and as a side effect also widened these four constructs' sub-expression slots from `parse_pipe_expr` (no comma) to the new comma-inclusive `parse_expr()` — which newly exposes this evaluator gap. The #462 branch restricts these four constructs' slots back to comma-free parsing (mirroring the existing, identically-reasoned restriction already in the codebase for `limit`'s `n` parameter — see `parse_limit_expr`) as an interim mitigation, so a multi-output sub-expression stays a parse error instead of becoming a silent wrong answer. This issue tracks implementing the real semantics so that restriction can be lifted.

## Real jq's actual semantics (verified against jq-1.7.1)

- **`reduce`/`foreach` `INIT`**: forks — each output of `INIT` starts a fully independent execution of the whole construct (cross product with the source iteration).
- **`reduce` `UPDATE`**: per source-element step, only the *last* output of `UPDATE` becomes the new accumulator. If `UPDATE` produces zero outputs for some step, the entire `reduce` produces no output at all.
- **`foreach` `UPDATE`**: same last-output-wins-per-step accumulator fold as `reduce`.
- **`foreach` `EXTRACT`**: evaluated per step against the new accumulator value, and *can* fan out — each of its outputs is a separate result of the whole `foreach`.
[...]

Issue #2031, opened 2026-08-31
jq: foreach with a successfully-navigating source emits where jq refuses (path register clobber)


newhoggy:
**Severity: Medium** (silent wrong output on the read side, and a write jq refuses on the write side — but only for a `foreach` whose *source* successfully navigates)

## Summary

Real jq refuses a `foreach` whose source expression successfully navigates: the source's own
path tracking clobbers the same path register the fold's emit checks, so the emit fails jq's
`jv_identical(v, jq->value_at_path)` and raises `Invalid path expression with result <input>`.
`succinctly jq` models the *accumulator's* provenance (`FoldRegister`, #1466/#1590) but not the
*source's*, so it emits happily.

Found while fixing #1872, which is about a *raising* source; this is the complementary case of a
source that navigates and succeeds. Deliberately left alone there — #1872's fix restricts its
value reuse to all-untracked branches precisely so it does not surface this divergence in new
places (see `resolve_fold_source`'s trackable-branch fallback in `src/jq/eval.rs`).

## Repro (verified against jq 1.7.1)

```console
$ echo '{"a":1}' | jq -c 'path(foreach (.a) as $k (.; .))'
jq: error (at <stdin>:1): Invalid path expression with result {"a":1}
$ echo $?
5

$ echo '{"a":1}' | succinctly jq -c 'path(foreach (.a) as $k (.; .))'
[]
$ echo $?
0
```

`reduce` with the same source agrees in both (`[]`, exit 0) — only `foreach`, which emits per
source element, is affected.

A mixed source shows jq streaming the non-navigating elements first, then failing on the
navigating one:

```console
$ echo '{"a":1}' | jq -c 'path(foreach (1, .[]) as $k (.; .))'
jq: error (at <stdin>:1): Invalid path expression with result {"a":1}
[]

$ echo '{"a":1}' | succinctly jq -c 'path(foreach (1, .[]) as $k (.; .))'
[]
[]
```

## Write side

The same divergence writes where jq refuses:

```console
$ echo '{"a":1}' | jq -c '(foreach (.a) as $k (.; .)) = 9'
jq: error (at <stdin>:1): Invalid path expression with result {"a":1}

$ echo '{"a":1}' | succinctly jq -c '(foreach (.a) as $k (.; .)) = 9'
9
```

(Both live-verified against `/usr/bin/jq` 1.7.1.)

## Scale

Randomised differential fuzzing over `path(reduce|foreach <source> as $k (<init>; <update>))`
[...]

PR #2386, opened 2026-09-04
docs(jq): document and pin the #534 INIT-fork null-ambient-input divergence

### LINKED ISSUE #2163: jq: plain (non-path) foreach with multi-branch INIT reuses the wrong branch's state across steps
**Severity: Low-Medium** (wrong value output on the read side, plain \`foreach\`, no
\`path()\`/write context involved at all).

## Summary

Surfaced during code review of #2031's fold-register fix, but this is in \`eval_foreach\`
(the ordinary *value*-context evaluator) -- confirmed present identically before and after
that fix, and unrelated to path-tracking entirely.

\`\`\`console
$ echo '{"a":1,"c":2}' | jq -c 'foreach (.a) as $k ((0,.c); $k)'
1
null
$ echo '{"a":1,"c":2}' | succinctly jq -c 'foreach (.a) as $k ((0,.c); $k)'
1
1
\`\`\`

\`(0,.c)\` forks INIT into two branches (\`0\` and \`2\`). Real jq's second line is \`null\`;
succinctly repeats \`1\` (looks like it's reusing the *first* branch's post-UPDATE state for
the second branch's own step, rather than each INIT fork keeping its own independent
accumulator across the single source element).

## Suggested next step

Not investigated beyond confirming the repro. Whoever picks this up should trace
\`eval_foreach\`'s handling of a multi-branch/comma-forked INIT to find where the second
fork's own accumulator state gets confused with the first's.

## Related

- #2031 -- the fold-register fix whose review incidentally surfaced this separate,
  plain-value-evaluator bug while probing nearby filter shapes.

newhoggy:
## Summary

- `foreach`/`reduce`'s INIT-comma forking (#534) already diverges from real jq's own SOURCE-re-entry VM quirk — pinned by `test_reduce_comma_slots`/`test_foreach_comma_slots` in `src/jq/eval.rs` — but that divergence was never recorded in `docs/compliance/jq/limitations.md`, an ADR-0018 rule-6 gap.
- This issue reported the same underlying mechanism reaching a plain field-read SOURCE (`.a`), not just a `.[]`-iterating one. Live-verified against jq 1.7.1: real jq's later INIT fork re-evaluates SOURCE against a synthetic `null` ambient document, so `.a` there silently answers `null` on the second fork; succinctly re-evaluates SOURCE consistently against the real input every fork.
- Per the last investigation's own option (b): extends the existing accepted-divergence precedent to this shape rather than attempting the architecturally invasive fix (re-deriving jq's undocumented VM register-threading rule across three evaluators — a prior narrower attempt regressed `test_eval_reduce_init_partial_prefix_still_forks_when_trailing_error_suppressed_1934`).
- Adds a new `docs/compliance/jq/limitations.md` entry recording the divergence on its merits (matching this file's existing "structurally malformed value" carve-out shape), a pinning unit test in `src/jq/eval.rs`, and a cross-evaluator parity test in `tests/jq_evaluator_parity_tests.rs` so `eval.rs` and `eval_generic.rs` can't silently drift apart on this behavior.

No evaluator logic changed — docs and tests only.

Fixes #2163

## Test plan

- [x] Behavior diffed live against pinned `/usr/bin/jq` 1.7.1 (see limitations.md entry for the exact commands/outputs)

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -3882,7 +3882,7 @@ magnitude error, a narrower divergence from real jq than #2240's own issue text 
 scoped this fix to. Fixing the underlying large-integer-precision gap itself (so `1.` could
 be accepted safely at every magnitude) is out of scope for #2240 and not separately filed.
 
-### `foreach`/`reduce`'s INIT-fork re-entry: SOURCE reads real jq's synthetic `null`, not the ambient input — recorded on its merits, ADR-0018 rule 4 (#534, #2163)
+### `foreach`/`reduce`'s INIT-fork re-entry: SOURCE reads real jq's synthetic `null`, not the ambient input — no carve-out; recorded as a still-open policy question (#534, #2163)
 
 `foreach`/`reduce`'s parser accepted a top-level comma in the INIT slot from #534 onward
 (`foreach SOURCE as $x (INIT1, INIT2; ...)`), which forks the whole construct once per INIT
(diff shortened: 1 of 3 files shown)


newhoggy:
Ran a 6-agent code review fan-out against this PR. It found real, confirmed problems in the first commit's documentation:

- A fabricated jq transcript (claimed `[reduce ...]` prints `[6]` before erroring; real jq's array constructor fully buffers and prints nothing) — also caught the same pre-existing inaccuracy in `test_reduce_comma_slots`'s code comment.
- Mislabeling the divergence as a settled "ADR-0018 rule 4, accepted divergence" when the entry's own body admitted it fit none of rule 4's four conditions, and its cited precedent is itself explicitly "no carve-out; still-open policy question" — not a settled pattern to build on.
- A false claim that all three jq evaluators share one once-upfront SOURCE computation — the path-context evaluator (`resolve_reduce`/`resolve_foreach`) actually already re-resolves SOURCE per INIT fork via a separate, older mechanism (#2031), and doesn't even reach the same answer as the other two evaluators in path/assignment position. Filed that as a separate follow-up: #2388.

Fixed all of the above in a second commit, verified live against jq 1.7.1 again, re-ran the full test suite (0 failures) and both clippy legs (clean), and rebased onto latest `main`.

Issue #2131, opened 2026-09-01
jq: range()'s integer stepping has no i64 overflow guard — wraps to garbage near i64::MIN/MAX


newhoggy:
**Severity: Low** (extreme boundary values only, no realistic query hits this)

## Summary

Found reviewing #2089's fix. `eval_range_values`'s `i += step` (`src/jq/eval.rs`,
both the ascending and descending arms) has no overflow guard. Pre-existing —
reproduces identically before and after #2089's fix, which restructured this
loop but didn't touch the stepping arithmetic itself.

## Repro

```console
$ echo null | succinctly jq 'range(-9223372036854775758; -9223372036854775808; -100)' | head -3
-9223372036854775758
9223372036854775758     # wrapped via two's-complement
9223372036854775658
```

A debug build panics instead (`attempt to add with overflow`); a release
build silently wraps and streams garbage values before the #2089
`MAX_RANGE` cap (if ever reached) raises — i.e. exactly the "silently wrong
data" failure mode #2089 was filed to eliminate, reachable via a different
trigger than the one that issue fixed.

## Why this is separate from #2089

#2089 was scoped to `MAX_RANGE`'s silent-truncation behavior, not general
arithmetic correctness in the stepping loop. This needs its own fix
(overflow-checked or saturating arithmetic) and its own oracle
verification against real jq's actual behavior at these boundaries —
real jq represents all numbers as `double` internally (no i64 path at
all), so its own behavior here is worth confirming precisely
(`echo null | jq 'range(-9223372036854775758; -9223372036854775808; -100)'`
currently prints nothing at all, exit 0 — no crash, no garbage — which
needs tracing before choosing a fix direction here).

## Suggested fix direction

Use checked or saturating arithmetic for `i += step` in both
`eval_range_values` (i64) and `eval_range_values_f64` (f64, which can't
technically "overflow" in the same way but can still drift to ±infinity
at extreme magnitudes — worth checking too).

newhoggy:
🔧 Claimed by agent `Agatha` — branch `issue-2131-range-i64-overflow`.

[...]

PR #2401, opened 2026-09-04
fix(jq): range()'s non-cap overflow keeps its already-correct prefix

### LINKED ISSUE #2219: jq: range()'s non-cap per-iteration overflow still discards already-correct pushed values
**Severity: Low** (extreme boundary values only, within roughly one step's width of
\`i64::MIN\`/\`i64::MAX\` -- even narrower than #2131's own original scope).

## Summary

Follow-up to #2131 (fixed, PR pending). That fix correctly decoupled the \`MAX_RANGE\`
cap-boundary lookahead's overflow from the whole-function bail, so hitting the cap no
longer discards already-correct pushed values. But \`eval_range_values\`'s **non-cap**
per-iteration advance (\`i = i.checked_add(step)?;\`, the ordinary "compute the next
value to decide whether to keep looping" step) still uses \`?\` to bail the *entire
function* on overflow -- discarding every value already pushed in the current call,
the identical failure mode #2131 fixed for the cap-boundary case, just relocated to
the ordinary advance.

## Repro (verified live against jq 1.7.1)

\`\`\`console
$ succinctly jq -n '[range(9223372036854775802;9223372036854775804;20)]'
[]                                    # WRONG -- exact i64 answer is [9223372036854775802]
$ jq -n '[range(9223372036854775802;9223372036854775804;20)]'
[9223372036854775802]

$ succinctly jq -n '[range(9223372036854775762;9223372036854775807;20)]'
[]                                    # WRONG -- exact answer is [...762, ...782, ...802],
                                       # 3 values, all discarded
$ jq -n '[range(9223372036854775762;9223372036854775807;20)]'
[9223372036854775762]                 # jq keeps at least the first value
\`\`\`

Not a regression from #2131 -- confirmed via a pre-#2131 (unpatched \`main\`) build that
this exact repro already gave wrong output there too, just via a different mechanism
(streaming 100,000 wrapped-garbage values before raising \`"range: maximum iterations
exceeded"\`, rather than #2131's clean empty answer). #2131's fix is a genuine safety
improvement here regardless (no more unbounded garbage streaming), just not yet a fully
exact one in this specific narrow band.

## Why this isn't a mechanical port of #2131's own fix

Naively generalizing #2131's "resolve the overflow locally instead of bailing" treatment
to the non-cap advance also changes #2131's *own* headline repro:
\`range(-9223372036854775758; -9223372036854775808; -100)\` currently gives \`[]\`, matching
real jq -- but only because \`from\`/\`to\` collide as the *same* \`f64\` once cast
(confirmed: \`float(-9223372036854775758) == float(-9223372036854775808)\` is \`True\` in
Python). Under a straightforward generalized fix, \`from\` itself (a legitimately valid
\`i64\` element, mathematically) would be *kept* instead of discarded, changing the answer
to \`[-9223372036854775758]\` -- diverging from jq at that specific repro, even though it
would arguably be *more* consistent with "succinctly gives an internally exact i64 answer"
(the philosophy #2131's own docs adopt for the large-integer divergence class it
documents) than with strict jq-parity.

## Suggested next step

This is a real design decision, not a mechanical patch: decide whether the non-cap
advance should adopt the same local-resolution treatment as the cap boundary (accepting
that #2131's own original repro's answer would then change from \`[]\` to
\`[-9223372036854775758]\`, diverging from jq there specifically), and update
\`docs/compliance/jq/limitations.md\`'s #2131 section plus its test suite accordingly if
so. Given the severity (narrower than #2131's own already-"Low, no realistic query hits
this" scope), this likely doesn't need urgent attention, but is recorded here so #2131's
own hard-won reasoning about the two sites' differing treatment doesn't need
re-deriving.

## Related

- #2131 -- the fix this is a residual gap in.

newhoggy:
## Summary

- `eval_range_values`'s per-iteration `i.checked_add(step)?` bailed the whole function the moment it overflowed `i64`, discarding every value already pushed for a from-scratch `f64` recomputation — even though an overflowing candidate can never satisfy the range's own `< to`/`> to` comparison either way, so the range was always exhausted right there. #2131 round 3 already proved and applied this exact reasoning for the cap-boundary lookahead; this extends it to the loop's own per-iteration advance, the one site round 3 left on the old bail-and-discard path.
- For `Int`/`Int`/`Int` operands this makes the `None => eval_range_values_f64(...)` fallback in `each_range`'s `emit` closure unreachable entirely — a range between two valid `i64` endpoints always pushes at least one value (`from` itself) before any possible overflow. Verified empirically via the test suite, not just reasoned. `eval_range_values_f64` remains real, reachable production code via the other dispatch arm (non-integer operands).
- Retargets the four tests the issue's own investigation identified as pinning the old bail-and-discard behavior (`test_eval_range_values_overflow_detection_2131`, `test_eval_range_values_matches_i128_reference_sweep_2131`'s reference helper, the renamed `test_range_dispatch_keeps_exact_int_prefix_through_overflow_2219`, and `tests/jq_cli_tests.rs`'s renamed `test_range_i64_overflow_keeps_exact_prefix_diverging_from_jq_2219`) to the new kept-prefix behavior.
- Updates `docs/compliance/jq/limitations.md`'s #2131 section (new "#2219 refinement" paragraph, corrected stale claims about the fallback still being reachable) and several `eval.rs` doc comments that described the old two-tier (cap-boundary fixed in round 3, non-cap still bails) behavior as current.
- Kept `eval_range_values`'s `Option` return type as-is (matches the issue's own prototyped-and-measured fix) rather than also removing it now that `None` is provably unreachable — documented why in three places (the function's own doc comment, the call site, and the i128 reference helper) so this doesn't need re-deriving later.

Fixes #2219

## Test plan

- [x] Verified live against pinned `/usr/bin/jq` 1.7.1 and the built `succinctly` CLI binary for every retargeted case

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -3699,13 +3699,18 @@ candidate can never satisfy `< to`/`> to` regardless of *which* call site
 computed it — cap-terminated or not. Both call sites now end the loop and
 keep whatever has already been pushed. A `range` between two valid `i64`
 endpoints always pushes at least one value (`from` itself) before any
-possible overflow, so **the `None => eval_range_values_f64(...)` fallback in
-`each_range`'s `emit` closure is now unreachable for `Int`/`Int`/`Int`
-operands entirely** — `eval_range_values` always returns `Some`. The
-function's `Option` return type is unchanged (kept for API stability, not
-because `None` can still occur) — see its own doc comment, and
+possible overflow, so **the `None => eval_range_values_f64(...)` fallback
+`each_range`'s `emit` closure used to reach for `Int`/`Int`/`Int` operands
+is now provably unreachable**. `eval_range_values` therefore drops its
+`Option` return type entirely — it's a private function with exactly two
+in-file call sites (production and its own test module), so there was no
+external caller for an `Option` signature to stay compatible with — see its
+own doc comment, and
 `test_range_dispatch_keeps_exact_int_prefix_through_overflow_2219`, which
 pins this for both the original repro and a fresh near-`i64::MAX` case.
+`eval_range_values_f64` remains real, reachable production code; it's just
+reached only via `emit`'s other match arm now, for genuinely non-integer
+operands.
 
 This changes what the original repro now produces:
 
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -3602,7 +3602,7 @@ own repro table didn't happen to list, not new ones introduced by this fix.
   [#2131](https://github.com/rust-works/succinctly/issues/2131). **Fixed**:
   see the next section.
 
-### `range`'s `i64` fast path at extreme magnitude (#2131): overflow-checked in place, extending the same #2089 hang-avoidance divergence
+### `range`'s `i64` fast path at extreme magnitude (#2131, #2219): overflow-checked in place, extending the same #2089 hang-avoidance divergence
 
 The gap the previous section left open: `eval_range_values` (the exact-`i64`
 fast path taken when `from`/`to`/`step` are all integers) computed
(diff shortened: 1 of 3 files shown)


newhoggy:
Ran a 6-agent code review fan-out. It converged strongly on two real issues:

- `eval_range_values`'s `Option` return type was kept "for API stability" — but the function is private with exactly two in-file call sites, so that rationale was factually wrong (3 independent review angles flagged this). Removed it: no more vestigial `None` arm that could silently reactivate the old discard-and-recompute bug if a future edit reintroduces an overflow-bail path with no test able to catch it (every overflow test now asserts a kept value, not `None`).
- `tests/jq_cli_tests.rs`'s `test_range_beyond_boundary_routes_to_f64_path_2131` wasn't retargeted by the first commit — its name/doc claimed the query "must route to the f64 path", which this PR makes false, but the assertion kept passing by coincidence (succinctly's JSON formatter prints a whole-number `Float` the same as an `Int`). Renamed and rewritten to state what's actually true.

Also added a descending-arm mirror of the multi-iteration overflow case to the end-to-end dispatch test (only the ascending direction had that coverage), and a comment on the descending loop's `break` site matching the ascending arm's explanation.

Full suite (0 failures), both clippy legs (clean), and live oracle verification re-run after these changes.

newhoggy:
Patch coverage: 96.49% (55/57). The 2 uncovered lines (`eval.rs:47047`, `47050`) are pre-existing defensive `panic!` arms in a test-only helper (`eval_range_values_i64`) that fire only if `eval_range_values` ever produced a non-`Int` value for an all-integer range — an invariant violation this helper guards against, not a code path any legitimate test input reaches. They predate this PR (confirmed via `git show d86d8a38b`) and were only flagged as "new" because the diff touched the surrounding function signature/body, not their content. Genuinely untestable without deliberately corrupting `eval_range_values`'s own correctness.

Issue #711, opened 2026-08-09
yq: alias values go stale after `.path = value` assignment through an anchor


newhoggy:
**Severity: High** — silently wrong data on a common editing pattern (update a value referenced by alias elsewhere in the same document), not an error.

## Summary

Assigning to a value that has an anchor does not update what its aliases resolve to. Real `yq` treats an anchor/alias pair as a single shared node — updating the anchor's value updates every alias pointing to it, matching the YAML representation-graph semantics (an alias is a reference to the *same* node, not a copy taken at parse time).

## Repro (verified against installed `yq` v4.53.3)

```
$ printf 'a: &x 1\nb: *x\n' | yq '.a = 99'
a: &x 99
b: *x                      # .b still aliases .a, now reads as 99
$ printf 'a: &x 1\nb: *x\n' | succinctly yq '.a = 99'
a: 99
b: 1                       # .b is stale — still the pre-edit value
```

## Root cause (hypothesis, not verified against the implementation)

succinctly appears to resolve an alias to its target's value once, as a snapshot, at read/index time — rather than keeping a shared reference that a later write can invalidate or update in place. This is consistent with the merge-key-flattening behavior in #712 (`<<` is also eagerly resolved rather than kept as a live reference), suggesting a shared root cause: anchors/aliases are modeled as a one-time value substitution rather than as graph edges.

## Why this matters

This is silent wrong output on ordinary editing of anchor-heavy YAML (Kubernetes manifests, CI/CD configs, and any hand-authored YAML using anchors for DRY) — exactly the kind of input CLAUDE.md's optimization notes call out repeatedly as a real-world workload this project targets (anchor/alias SIMD scanning, P4 optimization; `bench_anchors` fixture in `tests/data/yq-golden/cases/`). Nothing errors, so a user has no signal their edit didn't propagate.

## Notes

- No test in `tests/yq_cli_tests.rs` or `tests/yq_golden_tests.rs` covers assignment through an anchor/alias pair — searched for both "alias" + "assign" and "alias" + "= " co-occurring in a test body, zero hits.
- Found via a systematic gap-audit comparing succinctly's yq test coverage against mikefarah/yq's own test suite (2026-08-10). Real `yq`'s `operator_assign_test.go` has an explicit scenario for this named "Update node value that has an anchor" (*"Anchor will remain"*).

PR #2504, opened 2026-09-06
fix(jq): yq alias sync skips rebound slots and iterates nested groups to a fixpoint

### LINKED ISSUE #2498: yq: sync_aliased_paths syncs nested anchor groups in one pass, leaving an inner alias stale (split from #1351, case 3)
**Severity: Medium** — silent wrong value at an alias position.

## Summary

`sync_aliased_paths` (`src/jq/eval.rs`, #711) processes alias groups in `collect_alias_groups` order in a single pass. When an anchored node contains another anchor whose alias is also inside it, the outer group clones the parent into its aliases *before* the inner group has written into the parent, leaving the inner alias stale in the outer copies. Split out of #1351 (its "case 3") per the 2026-09-04 triage comment there.

## Repro

Verified against Homebrew `yq` v4.53.3 and succinctly `41f4c3b77`:

```console
$ printf 'a: &x\n  p: &y 1\n  q: *y\nb: *x\n' | yq -o=json -I=0 '.a.p = 5'
{"a":{"p":5,"q":5},"b":{"p":5,"q":5}}

$ printf 'a: &x\n  p: &y 1\n  q: *y\nb: *x\n' | succinctly yq -o=json -I=0 '.a.p = 5'
{"a":{"p":5,"q":5},"b":{"p":5,"q":1}}
```

The inner def can also sit *outside* the outer anchor while its alias sits inside it, so depth-sorting the groups is not sufficient:

```console
$ printf 'x: &y 1\na: &x {q: *y}\nb: *x\n' | yq -o=json -I=0 '.x = 5'
{"x":5,"a":{"q":5},"b":{"q":5}}
```

## Root cause

`sync_aliased_paths` is one `for (def_path, alias_paths) in groups` loop; a group whose def value changes only because an *earlier-processed* group's sync landed inside it is never revisited.

## Suggested fix direction

Iterate the group loop to a fixpoint: repeat passes until no group changes anything, bounded by `groups.len()` passes (each pass can only add writes, so that bound is exact). Test both nestings above with `-o=json` and with YAML output (marks must survive: `enforce_anchor_soundness` keeps `*y`/`*x` only once the values agree).

Note #2499 changes the same function's per-slot write condition; land the two together or rebase one on the other — they share the loop body.

## Why this is separate from #1351

#1351 (case 1) is about a write *through* an alias reaching the shared node; this bug is in the existing anchor→alias sync and reproduces on a write through the *anchor*.

### LINKED ISSUE #2499: yq: sync_aliased_paths clobbers an alias rebound earlier in the pipe (.b = null | .a.p = 9 discards the .b write)
**Severity: High** — a user's write is silently discarded.

## Summary

`sync_aliased_paths` (`src/jq/eval.rs`, #711) overwrites *every* pristine alias path once the anchor's value changes, including an alias position that an earlier stage of the same filter has already rebound to something else. The earlier write is lost with no diagnostic. Found while probing for #1351's design plan (the 2026-09-06 comment there, finding B1).

## Repro

Verified against Homebrew `yq` v4.53.3 and succinctly `41f4c3b77`:

```console
$ printf 'a: &x {p: 1, q: 2}\nb: *x\nc: *x\n' | yq '.b = null | .a.p = 9'
a: &x {p: 9, q: 2}
b: null
c: *x

$ printf 'a: &x {p: 1, q: 2}\nb: *x\nc: *x\n' | succinctly yq '.b = null | .a.p = 9'
a: &x {p: 9, q: 2}
b:
  p: 9
  q: 2
c: *x
```

Same with `.b = 5 | .a.p = 9` (yq keeps `b: 5`; succinctly prints `b: {p: 9, q: 2}`).

## Root cause

`sync_aliased_paths`: `for alias_path in alias_paths { set_value_at_path(result, alias_path, new.clone()) }` runs whenever `old != new` at the def path, with no check that the slot at `alias_path` is still an untouched copy of the anchor.

## Suggested fix direction

Only overwrite an alias slot whose **current** value (in `result`) still equals the anchor's **pristine** value (`old`) — i.e. it is still the unmodified copy the alias resolved to. A slot that was rebound, deleted, or moved earlier in the pipe no longer equals `old` and must be left alone. This is the same "identity by equality" gate #1351's plan uses for its redirect, so it should land first. Also skip a slot whose path no longer resolves rather than letting `set_value_at_path` autovivify or pad an array (see the index-shift issue filed alongside).

Tests (yq CLI tests, `-o=json` and YAML twins): `.b = null | .a.p = 9` keeps `b: null`; `.b = 5 | .a.p = 9` keeps `b: 5`; `del(.b) | .a.p = 9` does not recreate `b`; the existing `test_yaml_assign_through_anchor_updates_alias*` cases still sync.

## Why this is separate from #1351

It is a bug in the existing one-directional sync and reproduces without any write through an alias. It shares `sync_aliased_paths`'s loop body with the nested-group fixpoint issue filed alongside; the two should land together.

newhoggy:
## Summary

Two bugs in `sync_aliased_paths` (`src/jq/eval.rs`, #711), found while probing for issue 1351's design plan, bundled here because they change the same function's loop body.

- **#2499 (High, severity: a write is silently discarded).** `sync_aliased_paths` unconditionally overwrote *every* pristine alias path once the anchor's value changed, including an alias position an earlier stage of the same filter had already rebound to something else. Fix: track a per-slot "expected" value, seeded from the pristine document and updated on every write this function makes; a slot is only overwritten while its current value in `result` still equals its expected value. A slot whose path no longer resolves (`get_value_at_path` is `None`, e.g. after `del(...)`) is skipped outright, never autovivified or array-padded.

- **#2498 (Medium, silent wrong value).** `sync_aliased_paths` processed alias groups in `collect_alias_groups` order in a single pass. When an anchored node contains another anchor whose alias is also inside it (nested either way -- the inner def can sit inside *or outside* the outer anchor), the outer group could clone the parent into its aliases before the inner group's own write landed, leaving the inner alias stale in the outer copies. Fix: repeat the whole group loop until a pass writes nothing, capped at `groups.len()` passes (each pass can only add writes, so that bound is exact). #2499's per-slot expected-value tracking is exactly what makes a second pass correct instead of re-triggering its own rebound gate.

### Oracle repros (Homebrew `yq` v4.53.3, live-verified; succinctly now matches all of them)

```console
$ printf 'a: &x {p: 1, q: 2}\nb: *x\nc: *x\n' | yq '.b = null | .a.p = 9'
a: &x {p: 9, q: 2}
b: null
c: *x

$ printf 'a: &x\n  p: &y 1\n  q: *y\nb: *x\n' | yq -o=json -I=0 '.a.p = 5'
{"a":{"p":5,"q":5},"b":{"p":5,"q":5}}

$ printf 'x: &y 1\na: &x {q: *y}\nb: *x\n' | yq -o=json -I=0 '.x = 5'
{"x":5,"a":{"q":5},"b":{"q":5}}
```

succinctly's release build produces byte-identical output for all three (and for `.b = 5 | .a.p = 9`, `del(.b) | .a.p = 9`, and the pre-existing `.a.p = 9` / `.[] .p = 1` cases), including the re-emitted `&anchor`/`*alias` marks on the YAML output path.

## Test plan

- [x] `cargo build --release --features cli` then live-diffed every repro above against Homebrew `yq` v4.53.3 -- byte-identical
- [x] `cargo test --features cli,simd,regex,serde` (via `cargo-guard.sh`) -- all binaries pass, 0 failed
- [x] `cargo fmt -- --check` -- clean
- [x] New unit tests in `src/jq/eval.rs` exercising `sync_aliased_paths` directly (rebound gate, deleted-slot skip, nested-group fixpoint)
- [x] New CLI tests in `tests/yq_cli_tests.rs` (YAML + `-o=json -I=0` twins) for both issues

Fixes #2499
Fixes #2498

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -38488,7 +38488,7 @@ fn set_value_at_path(
 /// by the caller from the pre-mutation document. For each group, if the
 /// value at the definition path differs between `pristine` (before the
 /// write) and `result` (after), every alias path is overwritten with a clone
-/// of the new value.
+/// of the new value -- subject to the rule below.
 ///
 /// A group whose definition path no longer resolves in `result` (e.g. the
 /// anchored key was deleted) is left untouched, so its aliases keep their
@@ -38496,12 +38496,41 @@ fn set_value_at_path(
 /// yq. Likewise, a write that lands on an alias path directly rather than
 /// the anchor's own path leaves every other member of the group alone, since
 /// the anchor's own value never changed.
+///
+/// **Rule A -- rebound-slot gate (#2499).** An alias slot is only a
+/// candidate for overwriting while it still holds the value this function
+/// itself last put there (or, before any pass has touched it, the pristine
+/// value the alias originally resolved to). Per alias path, an `expected`
+/// value tracks that: it starts as the value at that path in `pristine`, and
+/// is updated to whatever this function writes there. A slot is only
+/// overwritten while its *current* value in `result` still equals its
+/// `expected` value; a slot that some earlier stage of the same pipe already
+/// rebound, deleted, or moved no longer matches and is left alone (`.b =
+/// null | .a.p = 9` keeps `b: null` rather than clobbering it with the
+/// anchor's new value). A slot whose path no longer resolves in `result` at
+/// all (`get_value_at_path` returns `None`, e.g. `del(.b)` ran first) is
+/// skipped outright -- never autovivified or array-padded by
+/// `set_value_at_path`.
 pub fn sync_aliased_paths(
     result: &mut OwnedValue,
     pristine: &OwnedValue,
     groups: &[(Vec<OwnedValue>, Vec<Vec<OwnedValue>>)],
 ) {
-    for (def_path, alias_paths) in groups {
+    // Per-slot expected value (Rule A), indexed the same way as `groups`:
+    // `expected[i][j]` tracks alias `groups[i].1[j]`. Seeded from `pristine`
+    // since a fresh, never-yet-synced alias slot holds exactly the value it
+    // originally resolved to there.
+    let mut expected: Vec<Vec<Option<OwnedValue>>> = groups
+        .iter()
+        .map(|(_, alias_paths)| {
+            alias_paths
+                .iter()
+                .map(|alias_path| get_value_at_path(pristine, alias_path))
+                .collect()
+        })
+        .collect();
+
+    for (group_idx, (def_path, alias_paths)) in groups.iter().enumerate() {
         let (Some(old), Some(new)) = (
             get_value_at_path(pristine, def_path),
             get_value_at_path(result, def_path),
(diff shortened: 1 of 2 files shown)


newhoggy:
Superseded by #2510. Review of both PRs showed the gate here and the redirect there had to land together: on its own this gate reads a through-write such as `.[].p = 9 | .a.q = 7` as a rebind and leaves `b` stale, and on its own #2510's mirror was overridden by the old unconditional end-of-pipe sync. #2510 now makes `sync_aliased_paths` and the per-write mirror share one rule (`propagate_anchor_writes`: same per-slot expected-value gate and fixpoint as this PR, applied per write expression so a through-write never looks like a rebind) and pins this PR's repros plus the through-write-then-anchor-write shapes as CLI tests. Closing this PR; #2498 and #2499 close with #2510.

Issue #2608, opened 2026-09-07
perf(jq): compact (-c) M2 rendering is 18-24% behind the DOM path on large arrays


newhoggy:
**Severity: Medium** (performance)

## Summary

With `-c`, the M2 fast path renders a large array **18-24% slower** than the
pre-#1576 DOM path, on both architectures. That is larger than #2206's own
reported regression, and on a shape #2206 never measured (it timed pretty
output only).

Found while fixing #2206, and it survives that fix.

## Measured (median of 9, interleaved, output-identity checked)

Document: `{"data": [...]}`, 12.6k / 61.8k elements (2 MB / 10 MB), query
`.data`. `DOM` = same tree with `can_json_fast_path` forced false.

| machine | size | M2 (with #2206's fix) | DOM | delta |
|---|---|---|---|---|
| M4 Pro | 2 MB | 22.6 ms | 19.3 ms | **+17.4%** |
| M4 Pro | 10 MB | 98.0 ms | 82.7 ms | **+18.5%** |
| 7950X | 2 MB | 29.4 ms | 24.4 ms | **+20.6%** |
| 7950X | 10 MB | 139.9 ms | 112.8 ms | **+24.0%** |

Control floors: +/-0.5% (M4 Pro), +/-1.2% (7950X).

Pretty output is fine after #2206 -- 10-12% *faster* than DOM on ARM, ~3%
behind on x86. The gap is specific to `-c`.

## Why it is interesting

Compact output has a raw-echo fast path (`JsonCursor::stream_json`: compact,
unsorted, source-preserving numbers -> echo the span verbatim). It is worth a
great deal -- with `--preserve-input`, which is the only jq-mode convention
that reaches it today, the same render is **~4x faster** than the DOM path:

```console
$ succinctly jq -c --preserve-input .data data_10mb.json   # M2:  35.4 ms
$ succinctly jq -c .data data_10mb.json                    # M2: 139.9 ms
```

Default jq mode cannot use it, because `JqCompat` has to canonicalize
numbers, collapse duplicate keys, and apply jq's escape table -- so it
re-renders the whole value node by node.

#2206's fix removed the per-number allocation, which is why the numbers above
are already better than they were. What is left is the re-render itself.

## Suggested fix direction

The raw-echo path is gated on the *convention*, but what it actually needs is
that the source span is already exactly what we would emit. For a compact,
unsorted render that is a checkable property of the span:

- every number already jq-canonical (`is_jq_canonical_number`, added by
[...]

Issue #2059, opened 2026-08-31
ci: Perf Regression Guard is failing on `main` itself — the checked-in baseline has gone stale


newhoggy:
**Severity: Medium** (a required-style gate that is red on `main` teaches everyone to ignore it)

Noticed while shipping #2047 (for #1690).

## Summary

`Perf Regression Guard (x86_64)` and `(ARM64-Linux)` are **failing on `main`'s own
commits**, with no PR involved. `tests/data/perf-guard-baseline.json` has drifted from
what `main` actually measures, and the guard is documented to fail on drift in *either*
direction — so it now fails unconditionally on every push.

## Evidence

```console
$ for sha in 7ddc21025 cc6dcb9f6; do
    gh api "repos/rust-works/succinctly/commits/$(git rev-parse $sha)/check-runs?per_page=100" \
      --jq '.check_runs[] | select(.name|startswith("Perf Regression Guard")) | "\(.name)=\(.conclusion)"'
  done
Perf Regression Guard (x86_64)=failure
Perf Regression Guard (ARM64-Linux)=failure
Perf Regression Guard (ARM64-Linux)=failure
Perf Regression Guard (x86_64)=failure
```

Both are `main` commits (`event: push`, `head_branch: main`).

The x86_64 run's table:

```
query                            baseline        current    drift
------------------------------------------------------------------
wide_keys_unsorted            351,802,447    373,763,282    +6.2%
wide_identity                 742,550,939    771,224,587    +3.9%
users_keys_unsorted            67,388,453     69,274,010    +2.8%
users_identity                375,956,384    395,276,016    +5.1%  <-- FAIL
arrays_identity             1,506,416,474  1,582,762,933    +5.1%  <-- FAIL
users_yq_keys_unsorted         82,548,670     82,560,483    +0.0%
```

## Two things worth noting before someone just runs `--update-baseline`

**Every jq workload moved up together (+2.8% to +6.2%) while the single yq workload is
flat at +0.0%.** That shape points at the shared jq CLI driver rather than the
evaluator: the jq and yq paths have separate runners, and a change in `jq_runner.rs`
would land on all five jq queries and none of the yq one. The window in which this
appeared contains #2012's `--argjson`/`--jsonargs` work, which is ~346 lines in
`src/bin/succinctly/jq_runner.rs`. Whether a +5% instruction-count cost on *every*
invocation is intended there is worth answering before blessing it into the baseline —
[...]

PR #2609, opened 2026-09-07
perf(jq): echo already-canonical numbers instead of re-rendering them

### LINKED ISSUE #2206: jq: M2 fast path regresses large-array plain field access (.data-style)

newhoggy:
## Summary

#2206 reported the M2 fast path regressing plain field access into a large
array, and named the output buffer as the cause. Measuring that first is what
found the real one — **both** of the issue's stated directions are net losses.

### What the issue blamed

| variant vs base | M4 Pro | 7950X |
|---|---|---|
| `String::with_capacity(span_len)` (pre-size the buffer) | +2.8% | +7.3% |
| no buffer, write straight to `out` | +9.6% | +12.9% |

The buffer is not the cost — it is a large win, and pre-sizing is worse than
letting it grow. Control floors: ±0.5% (M4 Pro), ±1.2% (7950X).

### The actual cost

Forcing the pre-#1576 DOM path and varying only the output convention
isolates it:

| shape | M2 vs DOM |
|---|---|
| `-c .data` (`JqCompat`) | **+23.6%** |
| `-c .data --preserve-input` | **−66.9%** |

The difference is that `--preserve-input` reaches the raw-echo fast path and
`JqCompat` cannot, because numbers must be canonicalized — and
`format_number_jq_compat` allocates a `String` unconditionally, including for
the common number that comes back byte-identical. On a 10 MB array that is
several hundred thousand allocations spent reproducing the input.

## The fix

`is_jq_canonical_number` answers both questions the writer needs in one pass —
is this a valid RFC 8259 number, and would the formatter return these exact
bytes — so a `true` skips `is_valid_number`'s scan **and** the allocation.

**Answering only the second question is worth nothing.** The first cut did
exactly that, sitting inside the existing `is_valid_number` gate, and measured
**+1.3% to +2.0% slower**: a redundant second scan costs more than the
allocation it saves. The scan, not the allocation, is the expensive half. That
negative result is recorded in the function's doc comment so the
cheaper-looking ordering is not retried.

## Measured (median of 9, interleaved, output-identity checked)

| machine | shape | vs base | vs DOM path |
|---|---|---|---|
| M4 Pro | pretty 2 MB | −5.2% | −10.1% |
| M4 Pro | pretty 10 MB | −5.3% | −11.8% |
| M4 Pro | `-c` 2 MB | −5.5% | +17.4% |
| M4 Pro | `-c` 10 MB | −6.0% | +18.5% |
| 7950X | pretty 2 MB | −1.1% | +1.6% |
| 7950X | pretty 10 MB | −0.7% | +2.8% |
| 7950X | `-c` 2 MB | −3.8% | +20.6% |
| 7950X | `-c` 10 MB | −1.6% | +24.0% |

#2206's own shape (pretty `.data`) is now **faster** than the DOM path on ARM
and within ~3% on x86, down from the reported +3.35%.

Compact output stays 18–24% behind the DOM path on both machines — bigger than
#2206's regression, on a shape it never measured. Filed as **#2608** rather
than widened into this PR.

## Correctness

No behaviour change, and the suite caught the one bug on the way. The first
predicate ran *ahead* of the `is_valid_number` gate while checking only
canonicality, so a scanner-lenient span (`1.`, which jq prints as `1`) was
echoed verbatim —
`test_stream_json_number_invalid_span_float_fallback_1576` failed on it. The
predicate now validates the span itself, which is what makes the early
ordering sound.

`is_jq_canonical_number_agrees_with_the_formatter_2206` pins the two against
each other over the formatter's own corpus, under the same validity gate the
writer applies. A fast-path predicate restating a formatter's rules is the
"duplicated predicates diverge silently" shape (#106), and here the divergence
would be silent wrong output.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — 7789 passed, 0 failed
- [x] A/B on **both** pinned machines (M4 Pro `johns-mac-mini`, 7950X
      `terminus`), interleaved, with control runs establishing the noise floor
      on each, and output identity gated via `scripts/ab-cli.py`

Fixes #2206

### DIFF
--- a/src/jq/mod.rs
+++ b/src/jq/mod.rs
@@ -116,9 +116,9 @@ pub use parser::{
 pub use resolve::{resolve_func_calls, resolve_func_calls_all, UnresolvedCall};
 pub use stream::{StreamError, StreamStats, StreamableValue};
 pub use value::{
-    assert_value_tree_depth, format_number_jq_compat, is_jq_canonical_number,
-    jq_bare_float_display, jq_float_is_scientific, nesting_depth_exceeded_message, NumberRepr,
-    OwnedValue, MAX_VALUE_TREE_DEPTH,
+    assert_value_tree_depth, format_number_jq_compat, jq_bare_float_display,
+    jq_float_is_scientific, nesting_depth_exceeded_message, NumberRepr, OwnedValue,
+    MAX_VALUE_TREE_DEPTH,
 };
 // `pub(crate)`, not `pub`: only `yaml::light`'s alias-chain resolvers (#1191
 // code review, #1193/PR #1314) need this from outside `jq::value` --
@@ -126,3 +126,9 @@ pub use value::{
 // depth-guard helper on docs.rs with no external caller to justify the
 // exposure.
 pub(crate) use value::assert_depth;
+// `pub(crate)` for the same reason (#2206): `json::light`'s number
+// writer is the only caller, and the predicate is an implementation
+// detail of `format_number_jq_compat`'s fast path -- not something an
+// external caller has any use for, and not a contract worth freezing on
+// docs.rs.
+pub(crate) use value::is_jq_canonical_number;
--- a/src/jq/mod.rs
+++ b/src/jq/mod.rs
@@ -116,9 +116,9 @@ pub use parser::{
 pub use resolve::{resolve_func_calls, resolve_func_calls_all, UnresolvedCall};
 pub use stream::{StreamError, StreamStats, StreamableValue};
 pub use value::{
-    assert_value_tree_depth, format_number_jq_compat, jq_bare_float_display,
-    jq_float_is_scientific, nesting_depth_exceeded_message, NumberRepr, OwnedValue,
-    MAX_VALUE_TREE_DEPTH,
+    assert_value_tree_depth, format_number_jq_compat, is_jq_canonical_number,
+    jq_bare_float_display, jq_float_is_scientific, nesting_depth_exceeded_message, NumberRepr,
+    OwnedValue, MAX_VALUE_TREE_DEPTH,
 };
 // `pub(crate)`, not `pub`: only `yaml::light`'s alias-chain resolvers (#1191
 // code review, #1193/PR #1314) need this from outside `jq::value` --
--- a/src/jq/value.rs
+++ b/src/jq/value.rs
@@ -176,6 +176,77 @@ fn strip_insignificant_leading_zero_and_plus(s: &str) -> String {
         None => format!("{sign}{canonical_int}"),
     }
 }
+/// Whether [`format_number_jq_compat`] would return `raw` byte-for-byte,
+/// so a writer can echo the source span instead of allocating a `String`
+/// per number (#2206).
+///
+/// The M2 streaming path re-renders every number through
+/// `format_number_jq_compat`, which allocates unconditionally -- even for
+/// the overwhelmingly common case of a number that is *already* in jq's
+/// canonical form and comes back unchanged. On a 10 MB document that is
+/// several hundred thousand allocations, and it is what kept `-c .data`
+/// from reaching the raw-echo fast path's speed (measured: the echo path
+/// renders the same document 4x faster).
+///
+/// Deliberately conservative -- it answers "certainly unchanged", never
+/// "probably". Anything with an exponent, a leading `+`, an insignificant
+/// leading zero, or an absent integer part (`.5`, #1171) returns `false`
+/// and takes the full formatter, so a `true` here is always safe to echo.
+///
+/// Kept adjacent to the formatter it predicts, and pinned against it by
+/// `is_jq_canonical_number_agrees_with_the_formatter_2206`, which asserts
+/// the two agree on every spelling in the formatter's own test corpus --
+/// the "duplicated predicates diverge silently" hazard (`CLAUDE.md`, #106)
+/// applies squarely to a fast-path predicate that restates a formatter's
+/// rules.
+#[must_use]
+pub fn is_jq_canonical_number(raw: &[u8]) -> bool {
+    // One pass that answers *both* questions the writer needs -- "is this a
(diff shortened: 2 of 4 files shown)


newhoggy:
### On the red Perf Regression Guard legs

Both are **expected and examined** — they are the guard reporting this
change's improvement, not a regression.

I reproduced the guard locally on the pinned 7950X rather than assuming.
Against the merge-base binary (the comparison the PR leg makes):

| query | ARM64-Linux | x86_64 |
|---|---|---|
| `arrays_identity` | **−35.8%** | **−36.4%** |
| `wide_identity` | **−18.8%** | **−17.8%** |
| `users_identity` | **−9.0%** | **−8.6%** |
| `wide_keys_unsorted` | +0.0% | +0.0% |
| `users_keys_unsorted` | +0.0% | +0.0% |
| `users_yq_keys_unsorted` | +0.0% | −0.4% |

All three failures are negative — fewer instructions. The guard's own message
covers this case: *"A drift in either direction fails ... an optimization that
genuinely costs less [needs] the same conscious baseline update."*

The two architectures landing within ~1 point of each other is the useful
part: an instruction-count saving should be architecture-independent, and it
corroborates the wall-clock A/B from a second axis. Instructions fall much
further than wall time (−36% vs −5%), which is what a real saving looks like
once process startup and I/O are excluded from the denominator.

**The legs stay red after the baseline refresh, by construction.** On a
`pull_request` the guard passes `--baseline-binary <merge-base>` and compares
two freshly built binaries; it never reads
`tests/data/perf-guard-baseline.json`. So no commit can make these green while
the optimization is real. The refreshed file governs the post-merge `push`
path, which is what would otherwise start failing on `main`.

Neither leg is among the merge queue's 8 required checks (`Clippy`,
`Do Not Merge`, `Documentation`, `Format`, `Skill manifest casing`,
`Test (ARM64)`, `Test (macOS ARM64)`, `Test (x86_64)` — confirmed against the
ruleset).

One thing worth flagging for anyone reading the baseline diff: the deltas
against the checked-in file are larger than against the merge-base binary,
and that gap is **not** this change. `wide_keys_unsorted` moves −7% against
the file while measuring +0.0% against the merge-base — the file had drifted
on its own since its last refresh, which is the 1–3%-within-days effect #2059
documents.

newhoggy:
Post-merge note: the perf guard also fails **once** on the merge commit itself (`cb2d71376`), and that is expected for the same structural reason the PR legs were red.

Reading `ci.yml`'s three cases:

- `pull_request` → `--baseline-binary <merge-base>`
- `push` with a buildable `github.event.before` → `--baseline-binary <previous commit>`
- otherwise → the checked-in `tests/data/perf-guard-baseline.json`

So the push of this merge compared my binary against the *previous* `main` commit's binary and reported the same −36% / −19% / −9%. The next commit to land compares against **this** commit's binary and should return to ~0%, so `main` self-heals on the next push rather than needing intervention.

The baseline-file refresh in `cb2d71376` is still the right thing: it governs the third case (an unbuildable or all-zero `before`, e.g. after a force-push), where a stale file would have been wrong by 36%.

Flagging it explicitly because a red perf guard *on main* is worth a second look by default — this one is understood, and I verified it by reading the workflow's branch logic rather than inferring it from the failure alone.

Issue #1114, opened 2026-08-17
yq: write_deferred_value's absence check double-resolves the cursor on the common (non-absent) path


newhoggy:
**Severity: Low (needs profiling to confirm/size)**

Found during review of PR #1111 (#1077's fix). Not blocking that PR (it's a pre-existing shape the fix necessarily inherits, just exercised far more often post-fix), but worth investigating with real measurements before deciding whether it's worth fixing.

## What's happening

`write_deferred_value` (added in #1111, `src/yaml/light.rs`, next to `is_deferred_value_absent`) calls:

```rust
let absent = is_deferred_value_absent(value);   // calls value.value() internally
...
if !absent {
    ...
    value.stream_yaml_value(out, ...)?;          // ALSO calls self.value() as its first line
}
```

`is_deferred_value_absent` resolves the cursor's `YamlValue` just to classify Null/empty-unquoted-string vs everything else. `stream_yaml_value` then resolves the *same* cursor's `YamlValue` again, unconditionally, as the very first thing it does (`match self.value() { ... }`). For the common case — an ordinary non-absent scalar field/item, which is most of them — that's two resolves of the same node instead of one.

Before #1077, this call site never ran `is_deferred_value_absent` at all outside a rare gated case (a field with a same-line key comment). #1077's fix reaches it for every scalar mapping-field/sequence-item, i.e. the common per-node streaming path, not just the rare one.

This file has a documented precedent for exactly this class of regression — `YamlCursor::anchor()`'s own doc comment notes it deliberately avoids a second resolve on this same hot path because "paying the resolve cost again here measured as a real streaming regression" (for a different resolve, `resolve_bare_seq_item`, not this one).

## Why this is filed as "needs profiling" rather than "definitely fix"

`YamlCursor::value()`'s actual cost for the common non-container-non-absent case looks like a handful of O(1)/O(log n) index lookups (`is_container`, `bp_to_open_idx`, `text_pos_by_open_idx`) plus a short scan past any anchor/tag markers — not obviously as expensive as `resolve_bare_seq_item`. Per this repo's own benchmarking discipline (`docs/guides/benchmarking.md`), a claim like this needs an interleaved A/B measurement on `yaml_bench`/`dev bench yq`, not doc-comment-pattern-matching, before deciding whether a fix is worth the complexity.

## If it does turn out to matter

[...]

PR #2618, opened 2026-09-07
perf(yq): avoid explicit_tag()'s redundant resolve on the absent/container deferred-value path

### LINKED ISSUE #2617: perf(yq): explicit_tag() re-resolves value() on the absent/container deferred-value path (#1114 residual)
**Severity: Low**

## Summary

Found while reviewing #1114's fix for `write_deferred_value`/`write_yaml_child_inline`'s double-resolve of a deferred value's cursor. #1114 closed the redundant `value()` resolve on the **common (non-absent)** path (measured -14.2% to -14.7% via A/B on the pinned boxes), but a smaller, separate redundancy remains on the rarer **absent** (and, for `write_yaml_child_inline`, container) path: `YamlCursor::explicit_tag()` internally calls `self.value()` again (to check for `YamlValue::Alias`), so:

```rust
let resolved = if value.is_container() { None } else { Some(value.value()) }; // resolve #1
let absent = resolved.as_ref().is_some_and(is_deferred_value_absent_at);
let tag = if absent { value.explicit_tag() } else { None };                   // resolve #2, inside explicit_tag()
```

still pays for two resolves of the same cursor when `absent` is true.

## Root cause

`YamlCursor::explicit_tag()` (`src/yaml/light.rs`, ~line 2510) matches `self.value()` itself, unconditionally, to detect and dereference through an alias (`x: &a !!str 1` / `y: *a`). It has no `_at` sibling accepting an already-resolved value the way `is_deferred_value_absent`/`stream_yaml_value` now do.

## Why this is separate from #1114

#1114's own repro and A/B benchmark both target the common non-absent path specifically (an ordinary `key: value` field) — that's the hot path for realistic documents (most fields have real values), which is why it was worth measuring and landing on its own. The `explicit_tag()` double-resolve only fires when a field's value is genuinely absent (`key: # comment` with nothing after) or, in `write_yaml_child_inline`, when the value is a container — both comparatively rare shapes, so the expected win here is much smaller and wasn't measured.

## Suggested fix direction

Add an `explicit_tag_at`-style overload (mirroring `stream_yaml_value`/`stream_yaml_value_at` and `is_deferred_value_absent`/`is_deferred_value_absent_at`) that accepts an already-resolved `YamlValue` and skips its own internal resolve. `explicit_tag()` is a broader public accessor with call sites beyond these two (check before widening), so this needs its own investigation into blast radius before implementing -- not a one-line change.

Should be benchmarked (interleaved A/B on `johns-mac-mini`/`terminus`, per this repo's own benchmarking discipline) before landing, the same way #1114 was, given the smaller expected magnitude.

newhoggy:
## Summary

Follow-up to #1114's review. #1114 closed the double-resolve of a deferred
value's cursor on the **common (non-absent)** path in
`write_deferred_value`/`write_yaml_child_inline`. A smaller redundancy
remained on the rarer **absent**/container path: `explicit_tag()` re-ran
`self.value()` internally just to check for `YamlValue::Alias`, even when the
caller already had the resolved value in hand (or structurally knew it
didn't need one at all).

## Change

Adds `explicit_tag_at`, a sibling of `explicit_tag` accepting
`Option<&YamlValue>` — the same shape `stream_yaml_value_at` already uses:

- `Some(&resolved)` when a prior `value()` call already produced it (the
  `absent` case in both call sites).
- `None` when the cursor is structurally known not to be an alias without
  resolving at all — a container position (`is_container()`'s own
  bitvector check) can never be an alias, so `write_yaml_child_inline`'s
  container branch now skips the resolve entirely.

`explicit_tag()` itself becomes a thin wrapper: `self.explicit_tag_at(Some(&self.value()))`.

## Benchmark

As expected from the much rarer trigger shape (issue's own framing), the
win is far smaller than #1114's own -14%. Interleaved A/B (`scripts/ab-cli.py`)
on the pinned boxes, `yq '.'` over a 5-60MB flow-heavy document
(`yaml generate -p flow`):

- **M4 Pro (`johns-mac-mini`)**: -0.7% to -1.4% across all 3 sizes, clearly
  outside this corpus's own control noise floor (-0.3%..+0.1%, mixed sign).
- **Zen 4 (`terminus`)**: -0.2% to -0.3%, within this corpus's wider noise
  floor (-1.0%..+0.9%, mixed sign) — directionally consistent but not
  distinguishable from noise on this architecture.

Output byte-identical before/after in every run (ab-cli's identity gate).

## Testing

- Existing regression suite (`test_deferred_absent_mapping_field_keeps_explicit_tag_1077`,
  `test_container_anchor_keeps_explicit_tag_1132`, `test_flow_anchor_keeps_explicit_tag_1115`,
  and siblings) already exercises both the absent and container branches
  through this exact code — no new tests needed; local patch coverage is
  100% (9/9 new lines).
- Full suite: `cargo test --features cli` — 3543 + 1635 + ... all green, 0 failed.

Closes #2617.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -246,6 +246,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `terminus` Zen 4), `yq '.'` over a 20-60MB array of small records:
   **-14.2% to -14.7%** (M4 Pro), **-14.2% to -14.4%** (Zen 4), output
   byte-identical before/after in every run.
+- **Closed a smaller residual of #1114 on the rarer absent/container
+  deferred-value path** (#2617): `explicit_tag()` still re-resolved a
+  cursor's value internally (to check for `YamlValue::Alias`) even when the
+  caller already had it in hand, or structurally knew a container position
+  can never be an alias. `explicit_tag_at` now takes that value (or `None`
+  for a container) directly. As expected from the much rarer trigger shape,
+  the measured win is far smaller than #1114's: interleaved A/B on the
+  pinned boxes, `yq '.'` over a 5-60MB flow-heavy document: **-0.7% to
+  -1.4%** (M4 Pro, clearly outside this corpus's own -0.3%..+0.1% noise
+  floor), within noise on Zen 4 (-0.2% to -0.3%, inside a -1.0%..+0.9%
+  noise floor) — output byte-identical before/after in every run.
 
 ### Removed
 
--- a/src/yaml/light.rs
+++ b/src/yaml/light.rs
@@ -2508,7 +2508,22 @@ impl<'a, W: AsRef<[u64]>> YamlCursor<'a, W> {
     /// `explicit_tag` was the one left out (#903 review).
     #[inline]
     pub fn explicit_tag(&self) -> Option<&str> {
-        if let YamlValue::Alias { target, .. } = self.value() {
+        self.explicit_tag_at(Some(&self.value()))
+    }
+
+    /// Sibling of [`Self::explicit_tag`] for a caller that already knows
+    /// whether this cursor is an alias without a fresh [`Self::value`] call
+    /// (#1114 review, #2617). Pass `Some(&resolved)` when a prior `value()`
+    /// call already produced it (e.g. `write_deferred_value`'s `resolved`
+    /// local on its `absent` branch), or `None` when the cursor is
+    /// structurally known not to be an alias without resolving at all (e.g.
+    /// `write_yaml_child_inline`'s `container` branch: `is_container()` is a
+    /// structural bitvector check, and a container position can never be an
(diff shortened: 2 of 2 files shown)


newhoggy:
Code review (3-way fan-out) found two things:
1. `explicit_tag_at` was left `pub` with no external caller, breaking the `stream_yaml_value_at`/`is_deferred_value_absent_at` precedent of keeping these `_at` siblings module-private -- fixed (now module-private), since its `None` contract is asymmetric with `stream_yaml_value_at`'s (no fallback resolve) and unsafe as a general-purpose default for an external caller.
2. The same redundant-resolve shape recurs at several hotter call sites this PR didn't touch (`stream_yaml_value_at`'s `String` arm and others) -- out of this PR's scoped benchmark (absent/container path only), filed as #2619 rather than scope-creeping this PR.

PR #2620, opened 2026-09-07
perf(yq): avoid explicit_tag()'s redundant resolve on the common streaming path

### LINKED ISSUE #2619: yq: explicit_tag()'s redundant resolve recurs on the common (non-absent) streaming path, not just the write path #2617 fixed
**Severity: Low-Medium** (mechanism is proven; magnitude unmeasured, but likely larger than #2617's own -1%/noise since this hits the *common* path)

## Summary

Found during #2618's (closes #2617) code review. #2617 added `YamlCursor::explicit_tag_at` (a sibling of `explicit_tag` accepting an already-known/proven value, skipping its internal `self.value()` resolve) and used it at the two write-path call sites #2617 targeted. Two independent review passes on that PR found the identical redundant-resolve shape recurring at several more, **hotter**, call sites that #2617 didn't touch.

## Root cause

`src/yaml/light.rs:1669`, inside `stream_yaml_value_at`'s `YamlValue::String(s) => { ... }` arm:

```rust
match known_value.unwrap_or_else(|| self.value()) {
    ...
    YamlValue::String(s) => {
        ...
        if let Some(tag) = self.explicit_tag() {
```

By the time this arm runs, the match subject has already proven `self` isn't `YamlValue::Alias` (it matched `String`) -- yet `self.explicit_tag()` calls the un-optimized path (`self.explicit_tag_at(Some(&self.value()))`), paying for a full second `value()` resolve purely to re-derive a fact this call site already has. `stream_yaml_value_at` is, per its own doc comment, part of "this crate's flagship streaming path" -- this arm fires on every string-valued scalar the writer streams, i.e. the overwhelming majority of real YAML nodes, not the "rarer absent/container" shape #2617 scoped itself to.

The same shape (an already-resolved-to-non-Alias value, followed by an unconditional `self.explicit_tag()` call) recurs at:
- `stream_json_value`'s `String` arm, `src/yaml/light.rs:2175`
- `write_json_to`'s `String` arm, `src/yaml/light.rs:2344`
- `tag()`/`as_*`-adjacent accessors, `src/yaml/light.rs:2837`, `6594`

(Some of these are pre-existing debt already flagged by a comment at `src/yaml/light.rs:~3062` -- "has the same unresolved `explicit_tag()` dependency" -- not new territory #2617 opened, just further evidence the pattern is systemic.)

## Why this wasn't fixed in #2618

#2618's own benchmark (interleaved A/B on the pinned boxes, flow-heavy corpus) was scoped to and measured the absent/container path specifically. Line 1669 sits on a structurally different, much hotter path (every tagged string, not just absent/container fields) -- fixing it deserves its own dedicated benchmark against a workload dominated by tagged scalar strings, not a silent fold into #2618's already-landed numbers.

## Suggested fix direction

**Immediate, narrow fix** (mirrors #2617's own pattern, reuses code already written): at `src/yaml/light.rs:1669`, replace `self.explicit_tag()` with `self.explicit_tag_at(None)` -- safe because the enclosing match arm has already proven non-`Alias`. Same shape likely applies to the 2175/2344/2837/6594 sites above; each needs its own check that the surrounding context actually proves non-`Alias` before passing `None`.

**Broader alternative worth considering instead of another one-off `_at` thread-through**: one review pass suggested a single cursor-level "last resolved value" memoization (e.g. a `Cell<Option<YamlValue>>` on `YamlCursor`, populated by `value()` and consulted by `explicit_tag`/anchor-adjacent accessors before re-deriving) to close this whole class of redundancy at once, rather than requiring a bespoke `_at` sibling and manual rethreading at every call site as they're found one PR at a time (#1114 → #2617 → this issue, three rounds so far). Per this repo's benchmarking discipline, that redesign should be measured against the incremental approach, not assumed better -- but the number of leftover sites accumulating across rounds suggests the investigation is now overdue.

Should be benchmarked (interleaved A/B on `johns-mac-mini`/`terminus`) with a corpus weighted toward tagged/anchored scalar strings before landing, per this repo's own benchmarking discipline.

newhoggy:
## Summary

Follow-up to #2617/#2618's review. Both reviewers independently found that
the redundant-resolve shape #2617 fixed on the rare absent/container write
path recurs at several hotter call sites on this crate's "flagship
streaming path" (`stream_yaml_value_at`'s own doc comment) — sites that
already match a cursor's resolved value into `YamlValue::String`, proving
it isn't an alias, then call the un-optimized `explicit_tag()` anyway,
paying for a second full `value()` resolve just to re-derive a fact
already in hand.

## Change

Five call sites converted from `self.explicit_tag()` to
`self.explicit_tag_at(None)` — safe because each is inside a `match` arm
that has already matched `YamlValue::String`, structurally proving `self`
isn't `Alias` (the only variant `explicit_tag`'s internal check cares
about):

- `stream_yaml_value_at`'s `String` arm (`src/yaml/light.rs`)
- `stream_json_value`'s `String` arm
- `write_json_to`'s `String` arm
- `tag()`'s `String` arm
- `is_falsy`'s `String` arm

Unlike #2617's absent/container shape (rare), this fires for **every**
string-valued node these paths touch, tag present or not — the
overwhelming majority of real YAML scalars.

## Benchmark

Interleaved A/B (`scripts/ab-cli.py`) on the pinned boxes, `yq '.'` over a
5-60MB array of small records (`yaml generate -p users`, matching #1114's
own corpus shape):

- **M4 Pro (`johns-mac-mini`)**: -23.6% to -24.5% across all 3 sizes
  (median of medians -23.8%), 0/3 rows slower.
- **Zen 4 (`terminus`)**: -29.2% to -30.0% (median of medians -29.7%),
  0/3 rows slower.

Output byte-identical before/after in every run (ab-cli's identity gate).

## Testing

- Full suite: `cargo test --features cli` — all green, 0 failed.
- Local patch coverage: 100% (5/5 new lines) — the existing regression
  suite already exercises every string-valued node through these paths,
  no new tests needed.

Closes #2619.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -259,18 +259,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   noise floor) — output byte-identical before/after in every run.
 - **Closed the same `explicit_tag()` double-resolve on the common
   (non-absent) streaming path -- the much hotter sibling #2617 left for a
-  follow-up** (#2619): five call sites on this crate's "flagship streaming
-  path" (`stream_yaml_value_at`, `stream_json_value`, `write_json_to`,
-  `tag()`, `is_falsy`) already match a cursor's resolved value into
-  `YamlValue::String` -- proving it isn't an alias -- then call the
-  un-optimized `explicit_tag()` anyway, paying for a second resolve on
-  every string-valued node regardless of whether a tag is even present.
-  All five now call `explicit_tag_at(None)` instead. Unlike #2617's rare
-  trigger shape, this hits the overwhelming majority of real YAML scalars.
-  Measured via interleaved A/B on the pinned boxes, `yq '.'` over a
+  follow-up** (#2619): nine call sites on this crate's "flagship streaming
+  path" and elsewhere (`stream_yaml_value_at`'s String/Mapping/Sequence
+  arms, `stream_json_value`, `write_json_to`, `tag()`, `is_falsy`,
+  `write_yaml_field_key`, the `--slurp` array writer) already have -- via a
+  matched `YamlValue::String` on the same cursor, or a structural
+  `is_container()`/`is_yaml_cursor_container()` check -- proof the cursor
+  isn't an alias, then call the un-optimized `explicit_tag()` anyway,
+  paying for a second resolve regardless of whether a tag is even present.
+  All nine now call `explicit_tag_at(None)` instead. Unlike #2617's rare
+  trigger shape, this hits the overwhelming majority of real YAML scalars
+  and every mapping key. Measured via interleaved A/B, `yq '.'` over a
   5-60MB array of small records (`yaml generate -p users`, #1114's own
-  corpus shape): **-23.6% to -24.5%** (M4 Pro), **-29.2% to -30.0%**
-  (Zen 4), output byte-identical before/after in every run.
+  corpus shape): **-23.8% to -24.9%** (pinned `johns-mac-mini` M4 Pro).
+  The Zen 4 box (`terminus`) was unreachable to re-measure the final
+  9-site version; the originally-measured 5-site subset on that box read
+  -29.2% to -30.0%, which the added sites are not expected to reduce.
+  Output byte-identical before/after in every run.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -257,6 +257,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   -1.4%** (M4 Pro, clearly outside this corpus's own -0.3%..+0.1% noise
   floor), within noise on Zen 4 (-0.2% to -0.3%, inside a -1.0%..+0.9%
   noise floor) — output byte-identical before/after in every run.
+- **Closed the same `explicit_tag()` double-resolve on the common
+  (non-absent) streaming path -- the much hotter sibling #2617 left for a
+  follow-up** (#2619): five call sites on this crate's "flagship streaming
+  path" (`stream_yaml_value_at`, `stream_json_value`, `write_json_to`,
+  `tag()`, `is_falsy`) already match a cursor's resolved value into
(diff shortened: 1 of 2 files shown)


newhoggy:
Code review (4-way fan-out plus my own direct verification) found:

1. **4 more sites with the identical proof shape, unconverted** — fixed: `stream_yaml_value_at`'s Mapping/Sequence block-style arms (guarded by `is_yaml_cursor_container`, the same structural non-alias proof #2617's `write_yaml_child_inline` already established as precedent), `write_yaml_field_key`'s key tag (same cursor already matched `YamlValue::String` one line above), and the `--slurp` array writer (mirrors the Sequence arm). Now 9 sites total, all converted to `explicit_tag_at(None)`.
2. **Stale doc comment on `explicit_tag_at`** — fixed: it said "two write-path call sites"; now accurately describes both the count and that two of the original five (`tag()`, `is_falsy`) are read/query paths, not writes.
3. **CHANGELOG magnitude discrepancy** — two review passes independently measured only ~14.3-14.5% vs my claimed -23.6%..-30%. Investigated: that measurement was taken on this session's own local machine (confirmed `Apple M5 Max`, hostname `Johns-M5-Pro-Max`) — **not one of the two pinned benchmark boxes** this repo's discipline requires (`johns-mac-mini` M4 Pro / `terminus` Zen 4). I reproduced on that same (heavily shared, 85-peer-session) local machine myself with a careful forced-through interleaved run and got **-24.0%**, matching my pinned-M4-Pro number almost exactly — not ~14%. I can't explain the reviewers' lower number (possibly non-interleaved, or hit while several concurrent sessions were contending for CPU) but three independent rigorous interleaved runs (2 different real machines) now agree closely, so the CHANGELOG numbers stand.

Re-benchmarked the expanded 9-site fix on the pinned M4 Pro: **-23.8% to -24.9%** (up slightly from the original 5-site -23.6%..-24.5%, as expected). `terminus` (Zen 4) is currently unreachable (Tailscale SSH re-auth wall I can't complete non-interactively) — CHANGELOG carries forward its already-measured 5-site number (-29.2%..-30.0%) honestly labeled as such rather than fabricating an updated figure.

PR #2622, opened 2026-09-07
fix(jq): key/path followed by more stages no longer materializes the document

### LINKED ISSUE #2149: perf(jq): a pipe continuing past key/path falls back to materializing the document
**Severity: Low** (perf only). Follow-up to #2061.

## What

`key` and `path` (no-arg) replace the walk's position with a *computed* value — a
string or an array of path components — so there is no document node left for a later
stage to navigate from. #2061's walk therefore refuses any pipe that continues past
them, in `path_context_walk_pipe` (`src/jq/eval_generic.rs`):

```rust
if matches!(first, Expr::Builtin(Builtin::Key | Builtin::PathNoArg)) {
    return Err(PathContextAbort::Unsupported);
}
```

So `.[] | key` is fast, but `.[] | key | tostring`, `.[] | key | ascii_downcase` and
`[.[] | key] | sort` on the *inner* pipe fall back to materializing the whole
document to answer something bounded.

## Why the bridge can do it and the walk cannot, yet

`eval_pipe_with_path_context_internal`'s `Key` arm hands its computed value to
`continue_rest_with_context` with the **ambient path unchanged**, so a later `key` in
the same pipe still resolves against the original position. The walk has no
equivalent: it drops the position entirely at that point.

## Suggested direction

Two pieces, both small:

1. Hand the computed `OwnedValue` to `fold_pipe_stages` for the remaining stages —
   the same thing `try_path_context_cursor_walk` already does for its `rest` slice
   when only stage 0 is walkable.
2. Keep the position alive alongside it, so a second `key`/`path` later in the pipe
   still answers from the ambient path rather than bailing.

(1) alone covers the common `key | tostring` shape and can ship without (2) if every
pipe containing a *second* path builtin after the first keeps bailing.

`parent`/`parent(n)` are unaffected — they move to a real document node, so the walk
already continues through them (`.a.b | parent | .d` is on the fast path today).

Refs #2061.

newhoggy:
## Summary

`key`/`path` (no arg) followed by more pipe stages used to fail the fast
cursor walk's static gate just because *something* followed it, falling
back to materializing the whole document once per element.

```console
$ succinctly jq -c '.[] | key' big.json        # 200,000-element array: ~0.6s
$ succinctly jq -c '.[] | key | tostring' big.json   # ~131s user time
```

At 20,000 elements the same query took 1.32s; at 200,000, 131s -- a ~10x
input growth costing ~100x more time, confirming O(n²).

## Root cause

`cursor_key`'s array arm answers "what index am I" with a sibling scan
from the array's first element every time -- O(position) per call. Once
`path_context_walk_split` refuses a pipe (because something needs path
context after `key`/`path`), every element gets evaluated individually
through this path, so the O(n) calls x O(position) per call become O(n²)
total. `.[] | key` alone avoids this because the walk's own `Iterate` step
already assigns each element's index during its single forward pass, for
free.

## Fix

`path_context_walk_split` (`src/jq/eval_generic.rs`) now recognizes a
navigational run ending on `key`/`path` as a walkable prefix too, the same
way it already recognized a single navigational first stage -- as long as
`rest` (the stages after `key`/`path`) needs no path context of its own.
The walk answers `key`/`path`'s value directly (no `cursor_key` scan at
all in this shape) and hands `rest` to the ordinary fold, exactly like the
pre-existing single-stage case.

**Deliberately not attempted**: the issue's own "part 2" suggestion (a
*second* `key`/`path` later in the same pipe answering from the same
ambient position). Live-verified against yq v4.53.3 first: `key | key` is
empty, `key | (key + 100)` is a constant unrelated to the element, `key |
path` wraps the value itself in an array -- real yq's own answer doesn't
settle into one coherent rule worth reproducing, so that shape stays on
the bridge exactly as before (unchanged output, confirmed against a
pre-fix build across ~10 such shapes).

## Testing

- New regression test (`tests/jq_cli_tests.rs`,
  `test_key_then_more_stages_does_not_blow_up_2149`): 100,000-element
  array, `.[] | key | tostring`, output checked plus a generous 15s timing
  guard (a true O(n²) regression at this size would cost tens of seconds
  by extrapolation from the live pre-fix measurement).
- Updated two existing unit tests whose assertions pinned the old,
  narrower gate behavior (`path_context_walk_split_takes_whole_head_or_nothing_2416`,
  `path_context_needs_owned_position_pins_its_reasons_2416`) -- both
  legitimate drift from widening the gate, not regressions (traced and
  explained in the updated test comments).
- Manually diffed ~28 pipe shapes (jq and yq mode, JSON and YAML) between
  a pre-fix and post-fix build, including the "second key/path" shapes
  above -- all byte-identical except the one shape this PR targets.
- Full suite: `cargo test --features cli` -- all green, 0 failed.
- Local patch coverage: 94.44% (34/36 new lines). The 2 uncovered lines
  are (a) a by-construction-unreachable `unreachable!()` fallback arm in a
  helper whose only caller already matched the same pattern, and (b)
  `cursor_key`'s malformed-member-key error propagation, pre-existing
  logic this PR only moved into a shared helper, not new behavior --
  reaching it requires a document malformation orthogonal to this fix.

Closes #2149.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`key`/`path` (no arg) followed by more pipe stages no longer falls back
+  to materializing the whole document once per element** (#2149, follow-up
+  to #2061). `.[] | key | tostring` (and any pipe shaped the same way,
+  provided nothing after `key`/`path` itself needs path context) used to
+  fail the fast cursor walk's static gate just because a stage followed
+  `key`/`path`, falling to a per-element `cursor_key` sibling scan that
+  costs O(position) per call -- O(n) calls, O(n^2) total. The gate
+  (`path_context_walk_split` in `src/jq/eval_generic.rs`) now recognizes a
+  navigational run ending on `key`/`path` as a walkable prefix too, the
+  same way it already recognized a single navigational first stage.
+  Live-verified: a 200,000-element array's `.[] | key | tostring` dropped
+  from ~131s user time to well under a second. A pipe with a *second*
+  `key`/`path`/`parent` after the first still falls back deliberately --
+  real yq's own answer for that shape doesn't settle into one coherent
+  rule to reproduce (`key | key` is empty, `key | (key + 100)` is a
+  constant unrelated to the element, `key | path` wraps the value itself,
+  all live-verified against v4.53.3).
 - **`succinctly yq`'s DOM output path now agrees with its streaming path on
   explicit tags and two Unicode separators** (#1982, Medium severity):
   1. **An explicit `!!str`/`!!int`/`!!bool`/`!!null` tag was silently lost
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -13761,7 +13761,7 @@ fn path_context_emitting_value<V: DocumentValue>(
             Some(key) => Ok(Some(key.clone())),
             None => Ok(None),
         },
-        _ => unreachable!("path_context_emitting_value called on a non-Key/PathNoArg expr"),
+        _ => unreachable!("path_context_emitting_value called on a non-Key/PathNoArg expr"), // omni-dev: coverage tolerate-line reason="unreachable: the sole caller (path_context_walk_generic's merged arm) only invokes this after matching Expr::Builtin(PathNoArg | Key) itself (#2149)"
     }
 }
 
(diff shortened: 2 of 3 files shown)


newhoggy:
8-way fan-out code review (angles A-E, some with automatic retries after the first pass) found:

1. **Real, high-severity gap** (live-verified by the reviewer: 0.04s vs 32.9s on a 100K-element array) — my first draft's emitting-stage check was a bare `matches!(Key | PathNoArg)`, narrower than `path_context_is_cursor_walkable`, the predicate every *other* branch of this same gate already uses. `.[] | (key) | tostring` — one paren away from the fixed `.[] | key | tostring` — still reproduced the full O(n²) blowup. Fixed: `path_context_walked_emitting_prefix_len` now reuses `path_context_is_cursor_walkable` directly, so `(key)`/`[key]`/`(key, path)`-wrapped forms are covered too. Added a dedicated regression test (`test_parenthesized_key_then_more_stages_does_not_blow_up_2149`).
2. **Self-contradictory doc comment** on `path_context_emitting_value` (claimed two call sites, has one) — fixed.
3. **Doc framing issue**: the "second key/path in rest stays refused" comment read as an optional judgment call; one reviewer traced that it's actually structural (the sink route discards the cursor/trail once it hands off a computed value, so there's no position left to answer from regardless of what rule real yq follows) — reworded to say so.
4. Minor duplication (a navigational-prefix scan pattern already used by `path_context_absent_split`, phrased with a different iterator combinator) and two now-provably-dead defensive guards (`idx == 0`, `rest.is_empty()`) were flagged by multiple passes — removed the dead guards since they added no value once traced, left the duplication as-is (small, localized, not worth the churn per one reviewer's own assessment).

Re-verified after fixes: full test suite green, clippy clean, ~10 more pipe shapes (including all the wrapped forms) diffed byte-identical against a pre-fix build in both jq and yq mode.

Issue #2399, opened 2026-09-04
perf(jq): trailing_element_gap_ok re-derives a position value_delimiter_ok already resolved for free


newhoggy:
Found during #2349's code review.

## What's wrong

`trailing_element_gap_ok` (`src/jq/document.rs`) takes only a bare cursor and internally calls `last_cursor.text_position()` — a real BP rank/select cost, not free (see `json/light.rs`'s own doc comment on `text_position()`'s cost). But for the common case where the last child is a scalar, that exact position was often already resolved moments earlier in the same loop iteration for free:

- **Object arm**: `value_delimiter_ok` (`document.rs`) prefers `value.and_then(DocumentValue::text_start)` — an already-decoded, free position — over `value_cursor.text_position()`, specifically to avoid this redundant lookup (see `text_start()`'s own doc comment: "instead of paying for a second cursor lookup to re-derive the same position"). But `trailing_element_gap_ok`, called once after the loop on the *last* field's cursor, has no way to reuse that already-resolved position — it only receives the cursor.
- **Array arm**: `elem_cursor.text_position()` is computed explicitly one line earlier in the loop's own `element_gap_ok`/delimiter check for that identical cursor, then thrown away and re-derived again by `trailing_element_gap_ok` after the loop.

This pattern predates #2349 (it's inherited from `to_owned_cursor_at_depth`, added by #2243), but #2349 doubled its incidence by adding a second call site with the identical shape (`push_generic_truthiness_cursor_error`). #2349's own A/B measurement found the container/trailing-gap checks (#2211/#2243) responsible for 57% of a measured +13.1% regression on `path(.[0])`-style queries over large arrays — this redundant re-resolution is a plausible, concrete piece of that cost.

## Suggested next step

`trailing_element_gap_ok`'s public signature (`document.rs`) takes only a cursor. Extending the "last child" tracking at each call site from `Option<C>` to `Option<(C, usize)>` (cursor + already-known position, where available) and adding a position-aware overload of `trailing_element_gap_ok` — mirroring `value_delimiter_ok`'s own `Option<&F::Value>` reuse pattern — would let the common scalar-last-child case skip one rank/select call per container, at both call sites (`to_owned_cursor_at_depth` and `push_generic_truthiness_cursor_error`).

[...]

PR #2623, opened 2026-09-08
perf(yq): expose explicit_tag_at as pub, reuse it in two more converters

### LINKED ISSUE #2621: yq/jq: explicit_tag()'s redundant resolve recurs in eval.rs/yq_runner.rs, needs explicit_tag_at exposed outside yaml::light
**Severity: Low** (mechanism proven, but requires a visibility decision, not a drop-in swap)

## Summary

Found during #2620's (closes #2619) code review. The same redundant-resolve shape #2619 fixed at 9 sites inside `src/yaml/light.rs` recurs at 2 more sites outside that module:

- `src/jq/eval.rs:41010`, `yaml_value_to_owned_checked`: `match cursor.value() { YamlValue::String(s) => { ... cursor.explicit_tag() ... } }` -- the cursor is already proven non-`Alias` by the `String` match one line above, identical proof shape to #2619's `tag()`/`is_falsy` sites.
- `src/bin/succinctly/yq_runner.rs:795`, `yaml_to_owned_value`: structurally identical -- `match cursor.value() { YamlValue::String(s) => { ... cursor.explicit_tag() ... } }`.

## Why this wasn't fixed in #2620

`YamlCursor::explicit_tag_at` is deliberately module-private (`fn`, not `pub`) as of #2618's own review -- it was originally added `pub` with no external caller, and review found that `pub` with an unsafe-by-default `None` contract (no fallback resolve, unlike `stream_yaml_value_at`) was a footgun for a hypothetical future external caller. Converting these two sites requires first deciding how to expose it safely across the module boundary (`pub(crate)`? A narrower `pub(crate)` wrapper with a more defensive name? Something else?) -- a real design decision, not a mechanical swap, so it's out of scope for #2619/#2620's mechanical fix.

## Suggested fix direction

Decide the exposure mechanism (likely `pub(crate)`, since both call sites are within this crate, just outside `yaml::light`), convert both sites to the `_at(None)` form, and consider whether `src/jq/eval_generic.rs:896`'s `yq_type_tag` (which already checks `c.is_alias()` and returns early before calling `c.explicit_tag()` -- the same proof shape, but on the generic `DocumentCursor` trait rather than a concrete `YamlCursor`) is in scope too, since it would need the accessor exposed through the trait, not just `pub(crate)`, if the same proof shape applies there. Per this file's dual-evaluator note (`eval.rs` and `eval_generic.rs` frequently need matching fixes -- see project memory), check whether one fix implies the other before landing just one.

Should be benchmarked (interleaved A/B on `johns-mac-mini`/`terminus`) before landing, per this repo's own benchmarking discipline -- these are jq/yq's owned-value materialization paths (`load`, DOM conversion), a different workload shape than #2619's pure streaming-output paths, so the magnitude is unmeasured and shouldn't be assumed to match #2619's ~24-30%.

newhoggy:
## Summary

Follow-up to #2619's own review, which listed two more sites with the
identical redundant-resolve shape but flagged them as needing a visibility
decision first rather than a drop-in swap:

- `src/jq/eval.rs`'s `yaml_value_to_owned_checked`
- the `succinctly` binary's own `yaml_to_owned_value`
  (`src/bin/succinctly/yq_runner.rs`)

Both match `cursor.value()` into `YamlValue::String` (proving non-`Alias`),
then call the un-optimized `explicit_tag()` anyway, paying for a second
resolve.

## Why this needed a visibility change

`explicit_tag_at` was module-private as of #2618's own review (added `pub`
originally with no real caller, found to be a footgun, narrowed back).
`src/jq/eval.rs` is in the same crate as `yaml::light`, so `pub(crate)`
would have reached it — but `src/bin/succinctly/yq_runner.rs` is part of
the **separate** `succinctly` binary crate, consuming the library only
through its `pub` surface (`use succinctly::yaml::{...}`). `pub(crate)`
would not have reached it. `explicit_tag_at` is now `pub`, with its
existing safety-contract doc comment (the `None` case requires the caller
to already have proof of non-`Alias`; unlike `stream_yaml_value_at`,
there's no fallback resolve) extended to name both new callers alongside
the module-internal ones.

## Benchmark

Unlike #2619's streaming paths, this one measured as **noise on both
pinned boxes** (interleaved manual A/B, `yq -P '.'` — the DOM-forcing flag
that reaches these converters — over a 20MB array of small records):
effectively flat on `johns-mac-mini` M4 Pro (7 reps, mixed sign, ~1%
band), a hair slower within noise on `terminus` Zen 4 (7 reps, ~1% band,
also mixed relative to run-to-run variance). Output byte-identical before
and after in every run.

This is expected, not a red flag: DOM/owned-value conversion allocates a
`String`/`Vec`/`IndexMap` per node, which dwarfs a single skipped cursor
resolve — a very different cost profile from the allocation-light
streaming write path #2619 optimized. Landed anyway for the same reason
the redundancy was worth finding in the first place: it's a genuine,
provably unnecessary resolve, now closed everywhere in the crate it's
reachable from, not landed on a performance claim this particular
workload doesn't support (documented honestly in the CHANGELOG entry
rather than inflated).

## Testing

- Full suite: `cargo test --features cli` — all green, 0 failed.
- Local patch coverage: 100% (3/3 new lines).
- Confirmed the binary crate actually compiles against the new `pub`
  surface (`cargo build --features cli` builds both the lib and the
  `succinctly` binary target).

Closes #2621.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -293,6 +293,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   9-site version; the originally-measured 5-site subset on that box read
   -29.2% to -30.0%, which the added sites are not expected to reduce.
   Output byte-identical before/after in every run.
+- **Reused the same `explicit_tag_at` fix at two more call sites outside
+  `yaml::light`** (#2621, follow-up to #2619's own review): `src/jq/
+  eval.rs`'s `yaml_value_to_owned_checked` and the `succinctly` binary's
+  `yaml_to_owned_value` (owned-value/DOM conversion, e.g. `-P`, `load()`)
+  had the identical redundant `explicit_tag()` resolve on an
+  already-proven-non-alias cursor. `explicit_tag_at` is now `pub` (it
+  needed to reach a separate binary crate, not just another module in the
+  same crate) and both sites call it. Unlike the streaming paths #2619
+  closed, this one measured as noise on both pinned boxes (interleaved
+  manual A/B, `yq -P '.'` over a 20MB array of small records): effectively
+  flat on M4 Pro, a hair slower within noise on Zen 4. DOM conversion's own
+  allocation cost (`IndexMap`/`Vec`/`String` construction per node)
+  dwarfs a single skipped cursor resolve here, unlike the allocation-light
+  streaming write path. Landed anyway for the same reason #1114's
+  redundancy was worth finding in the first place -- a genuine, provably
+  unnecessary resolve, now closed everywhere in the crate it's reachable
+  from -- not for a performance claim this workload doesn't support.
+  Output byte-identical before/after in every run.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -308,8 +308,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   dwarfs a single skipped cursor resolve here, unlike the allocation-light
   streaming write path. Landed anyway for the same reason #1114's
   redundancy was worth finding in the first place -- a genuine, provably
-  unnecessary resolve, now closed everywhere in the crate it's reachable
-  from -- not for a performance claim this workload doesn't support.
+  unnecessary resolve -- not for a performance claim this workload doesn't
+  support. One more site with the same shape remains, behind the generic
+  `DocumentCursor` trait rather than a concrete `YamlCursor`
+  (`eval_generic.rs`'s `yq_type_tag`); closing it needs the same accessor
+  added to the trait, a larger change than this follow-up's own scope.
   Output byte-identical before/after in every run.
 
 ### Removed
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -841,8 +841,8 @@ fn yaml_to_owned_value<W: AsRef<[u64]>>(cursor: YamlCursor<'_, W>) -> Result<Own
             let mut arr = Vec::new();
             let mut rest = elements;
             // `uncons_resolved_cursor`, not `uncons_cursor`: the recursive
-            // call's own `cursor.explicit_tag()` above doesn't resolve a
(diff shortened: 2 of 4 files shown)


newhoggy:
5-way fan-out code review found a real CI-breaking bug plus 3 more legitimate issues, all fixed:

1. **CI-breaking**: making `explicit_tag_at` `pub` turned its intra-doc link to the still-private `stream_yaml_value_at` into a `rustdoc::private_intra_doc_links` error under `-D warnings` — exactly what CI's `docs` job runs. Verified live (`RUSTDOCFLAGS="-D warnings" cargo doc`) and fixed by rewording to plain text.
2. **Real design gap**: one reviewer found this codebase's own precedent for exactly this situation — `path_context_pipe_streams_cursors` uses `#[doc(hidden)] pub fn` for "must be `pub` only to reach a `src/bin/succinctly/*` binary crate, not intended as public library API." Applied the same pattern to `explicit_tag_at`; verified with `cargo doc --no-deps --lib` that it no longer appears in the generated public docs.
3. Two pre-existing comments (in `eval.rs`/`yq_runner.rs`) named `cursor.explicit_tag()` by reference to a call this PR had already renamed to `explicit_tag_at(None)` a few lines above — fixed.
4. The CHANGELOG's "closed everywhere in the crate it's reachable from" overclaimed — one more site (`yq_type_tag`, behind the generic `DocumentCursor` trait) has the same shape and needs trait-level plumbing to close. Filed as #2624 rather than glossed over; CHANGELOG reworded.

On the "is this worth landing given the noise-level benchmark" question: two independent passes confirmed the mechanism itself is sound and the two conversions correct, and neither suggested closing this the way #2399 was. The difference from #2399: that one needed genuinely new code (threading a value through call sites) for a redundancy that only fires once per container; this one reuses already-audited, already-merged infrastructure (`explicit_tag_at`, proven correct at 9 other sites) for a redundancy that fires on every string value — same mechanism as #2619's real ~24-30% win, just swamped by a different cost profile (DOM allocation) at these two sites. Documented honestly rather than dropped.

Issue #2584, opened 2026-09-07
jq: Expr::Arithmetic/And/Or's needs_path_context gate may lose cursor-metadata identity (di, line, column) when routed through the bridge


newhoggy:
## Background

While investigating #2581 (giving `Expr::Compare` the same `needs_path_context`-gated native arm `Expr::Arithmetic`/`And`/`Or` already have, so an ungated comparison still raises on a document-level decode failure like #1194's malformed key), I needed to widen the gate to `needs_path_context(...) || uses_cursor_metadata_builtins(...)` — without the second half, `select(di == 1)` on a multi-document YAML stream broke, because `bridge_to_full_evaluator`'s re-serialize-and-reindex round trip can't preserve `di`'s (`document_index`) real cursor identity; it answers from a fixed-default stub instead (the same carve-out `eval_using`'s own doc comment already documents for `input`/`inputs`).

`Expr::Arithmetic`/`And`/`Or`'s own existing gates (added by #1812/#2286, predating this) only check `needs_path_context`, not `uses_cursor_metadata_builtins`. They look equally exposed to the same bug — an ungated expression like `di + 0` should, by the same reasoning, lose `di`'s real identity when routed through the bridge — but no existing test reaches this shape for them, so it's unconfirmed.

## Suggested next step

Confirm live whether `di + 0 == 1`-shaped queries (or any other cursor-metadata builtin — `line`, `column`, `document_index`, `anchor`, `style`, `line_comment` — combined with `Arithmetic`/`And`/`Or` and no other path-context stage) actually lose cursor identity through the bridge, the same way `select(di == 1)` did for the ungated `Compare` case. If confirmed, widen those three gates the same way #2581 widened `Compare`'s.

## Related

- #2581 — the `Compare` fix and where this was found.
- #1812 / #2286 — the original `Arithmetic`/`And`/`Or` gate this extends.

newhoggy:
Brody verified this against current main `745870424`: the premise is superseded by the merged #2626 and #2476 fixes.

- `eval_generic.rs` evaluates Arithmetic natively with its cursor; the `needs_path_context` gate was removed by #2626. Unary minus shares that correction.
- And/Or have been ungated and cursor-native since #2476. The current `path_context_single_native` comments document both removals.
[...]

Issue #1812, opened 2026-08-29
jq: try/catch doesn't contain a decode-failure/collision error the way ? does


newhoggy:
Found reviewing PR #1807 (#1734) -- confirmed pre-existing, reproduces identically against unmodified `main`.

## Summary

`eval_generic.rs`'s `eval_single` has no native arm for `Expr::Try` -- it falls to the wildcard fallback arm (~line 4328-4398), which materializes the ambient value via `owned_or_err!(to_owned_with_cursor(&value, cursor))` *before* ever invoking `full_eval` (the function that actually contains the correct `try`/`catch`-vs-uncatchable gating logic). `owned_or_err!` is an unconditional early return on `Err` with no catchability check at all (`is_decode_failure()`/`is_uncatchable()` never consulted).

## Repro

```console
$ printf '[{"\z": 1, "\z": 2}]' | succinctly jq 'try (1+1) catch "x"'
jq: error (at <stdin>:0): object key "\z" is ambiguous: an undecodable key's display form collides with another key of the same name and cannot be represented
$ echo $?
5
```

`(1+1)` never touches the object at all, yet `catch` never runs -- the error escapes uncaught. Compare to `?`, which correctly suppresses the identical error via `Expr::Optional`'s own gated native arm:

```console
$ printf '[{"\z": 1, "\z": 2}]' | succinctly jq 'sort?'
$ echo $?
0
```

## Why this matters more after #1734

Before #1734 merged, this error class was reachable only through `eval_generic.rs`'s own whole-document conversion path. #1734 fixed `eval.rs`'s separate, private `to_owned_checked`/`to_owned_checked_at_depth` to also raise decode-failure/collision/structural-key errors correctly -- widening where this pre-existing `try`/`catch` gap can bite, since every one of `to_owned_checked`'s ~40 call sites can now raise the identical error class from an arbitrarily nested sub-document once `full_eval` is entered. #1734 doesn't cause this bug and its own new tests correctly pass (they call `eval::<Vec<u64>, JqSemantics>` directly, bypassing the CLI bridge where this manifests) -- but the blast radius genuinely grew.

## Suggested direction

Not attempted. `Expr::Try` likely needs its own native arm in `eval_generic.rs`'s `eval_single` (mirroring `Expr::Optional`'s existing gated handling), rather than falling through to the wildcard/`owned_or_err!` path, so materialization happens through the same catchability-aware logic `full_eval` already has.

## Related

- #1734 -- the fix whose review found this.
[...]

PR #2625, opened 2026-09-08
fix(jq): gate Expr::Compare on needs_path_context so a malformed document raises

### LINKED ISSUE #2581: jq: Expr::Compare's native arm doesn't raise on a document-level malformation/decode failure, unlike every other builtin
## Summary

Real jq parses an entire input document eagerly, before running any filter — so a document-level malformation (a non-string object key, #1194's own `{123: 1}`; an invalid string escape, #1746/#2173's own shape) makes **every** query on that document fail, including ones that never read the malformed region at all. #1812/#2286 established and oracle-verified this for `Expr::Arithmetic`/`Expr::And`/`Expr::Or` (`try (1+1) catch "x"` raises uncaught, matching jq, per `test_try_catch_contains_a_genuinely_catchable_malformed_key_error_1812`) and left an explicit note that `Expr::Compare`'s own native arm (`eval_compare_generic`) was a **known, pre-existing exception**:

> `Expr::Compare`'s own arm already diverges here -- `1==1` on `{123: 1}` answers `true` -- which is a pre-existing gap, not one this arm widens.

This issue is that gap, tracked on its own rather than left as a code-comment aside.

```console
$ printf '{"a": {"bad": "\x"}, "b": 5}' | jq -c '.b == .b'
jq: parse error: Invalid escape at line 1, column 18
$ printf '{"a": {"bad": "\x"}, "b": 5}' | succinctly jq -c '.b == .b'
true
```

```console
$ printf '{123: 1}' | jq -c '1==1'
jq: parse error: Object keys must be strings at line 1, column 5
$ printf '{123: 1}' | succinctly jq -c '1==1'
true
```

`sort?` on the same documents already raises correctly in succinctly (matching jq), confirming this is specific to `Expr::Compare`'s own native dispatch arm, not a general architectural limit.

## Where

`eval_single`'s `Expr::Compare` arm (`src/jq/eval_generic.rs`) calls `eval_compare_generic` unconditionally, unlike the `Expr::Arithmetic`/`Expr::And`/`Expr::Or` arms just below it, which are gated on `needs_path_context` specifically so an *ungated* case still falls through to the materializing bridge (`bridge_to_full_evaluator`/`to_owned_with_cursor`) that surfaces the document-level error. `eval_compare_generic` apparently never reaches that bridge at all, for either gated or ungated operands.

## Suggested next step

Give `Expr::Compare` the same `needs_path_context`-gated shape the Arithmetic/And/Or arms already have (an ungated comparison falls through to the bridge, a path-context one uses `eval_compare_generic` directly) — the straightforward, established fix pattern. Verify against both repro shapes above, plus a `needs_path_context` case (`.[] | key == "a"`) to confirm the native arm still handles that correctly. Should not need new machinery, just closing the same gap #1812/#2286 already closed for its three siblings.

## Related

- #1812 / #2286 — the sibling fix for `Arithmetic`/`And`/`Or`, and the "pre-existing gap, not one this arm widens" note this issue closes.
- #2173 — where this was rediscovered while investigating a different (memory/cost) concern; that issue's own "correctness" framing (item 2, `.b == .b` succeeding vs. `1+1` raising) has the two sides backwards relative to what this issue establishes: `1+1`'s raise is the *correct*, jq-matching behavior, and `.b == .b`'s silent success is the actual bug.

newhoggy:
## Summary

`Expr::Compare`'s native arm (`eval_compare_generic`) was the one member of its family
left ungated. Its `Expr::Arithmetic`/`Expr::And`/`Expr::Or` siblings, gated by
#1812/#2286, name it in their own comments as a known exception:

> `Expr::Compare`'s own arm already diverges here -- `1==1` on `{123: 1}` answers `true`
> -- which is a pre-existing gap, not one this arm widens.

Real jq parses a document eagerly, so a document-level malformation fails **every** query
over it, including one that reads nothing. Live against the pinned oracle (`/usr/bin/jq`
1.7.1):

```console
$ printf '{123: 1}' | jq -c '1==1'
jq: parse error: Object keys must be strings at line 1, column 5
$ printf '{"a": {"bad": "\x"}, "b": 5}' | jq -c '.b == .b'
jq: parse error: Invalid escape at line 1, column 18
```

succinctly answered `true` for both. This gives the arm the same
`needs_path_context` gate, so a comparison reading no path context falls through to the
wildcard bridge whose ambient materialization surfaces the decode failure. stdout and exit
code now match jq; the message text still differs, since the lazy index reports at the span
rather than as an up-front parse error.

### The gate needs a second half the siblings lack

A comparison over a *cursor-metadata* builtin (`di`, `line`, `column`, ...) reads no path
context but cannot survive the bridge either — its re-serialize-and-reindex round trip
answers `di` from a fixed-default stub rather than the live document index, the same
carve-out `eval_using`'s doc comment already records for `input`/`inputs`. Without
`uses_cursor_metadata_builtins` in the gate, `select(di == 1)` over a multi-document
stream picks the wrong documents.

I negative-tested this half rather than assuming it: removing it makes both existing
`test_yaml_select_di_eq_n` and `test_yaml_select_di_comparison` fail (`left: 0, right: 1`).
The three sibling arms carry only the `needs_path_context` half and look equally exposed —
left alone rather than widened speculatively, tracked at #2584.

## A prior pass released this issue over two observations; both were oracle artifacts

An earlier session got the gate working but released the claim, citing a test regression
and an unexplained second output. Checking both against the oracle settles them:

**1. The "second value I have no explanation for" is what jq does.**  That session
observed `nth(0; 1 as $x ?// $y | $x + 100)` emitting `101` then `100` on unmodified
`main` and called it inexplicable. Real jq emits both on a *well-formed* document:

```console
$ echo '{"bad":1}' | jq -c 'nth(0; 1 as $x ?// $y | $x + 100)'
101
100
```

succinctly already matched jq. There was nothing to diagnose, and it was never an argument
against the gate.

**2. The "broken" test encoded a divergence from jq, and the gate fixes it.**
`test_as_pattern_alternatives_retry_under_a_wrapping_consumer_1519`'s malformed-document
row expected `"ok"` on stdout before the raise — under a test whose header states *"Every
row is the real jq 1.7.1 answer, captured live."* jq emits nothing there:

```console
$ printf '{"bad": "\x"}' | jq -c 'nth(0; 1 as $x ?// $y | (if $x == 1 then "ok" else .bad end))'
jq: parse error: Invalid escape at line 1, column 12   # exit 5, empty stdout
```

(Its expected stderr was also succinctly's own message, not jq's — so that row was never a
live capture.) The gate moves stdout and exit code onto jq's answer. I updated the row to
that, and added a sibling row using a bare truthiness test instead of `==` to keep the
`NthStream` prefix-preservation property the original row was written to pin, where
succinctly's laziness still applies.

## Oracle verification

Generated matrices rather than hand-picked cases:

| Sweep | Cases | Result |
|---|---|---|
| jq mode, 6 operators × 12 left × 10 right operands × 13 well-formed docs, plus `[.[]?\|...]`/`try`/`select` wrappings | 12,480 | **0 divergences from jq 1.7.1** |
| jq mode, 18 filters × 10 malformed docs | 180 | 23 divergences, **all pre-existing** (verified against a pre-fix build) |
| yq mode, comparisons over 11 docs incl. multi-doc/anchors/`di`/`line` | 7,854 | **0 outputs changed by this fix** (pre-fix vs post-fix build, byte-compared) |

The yq sweep matters because `eval_generic` is shared between both modes — I compared
pre-fix and post-fix binaries directly rather than only against real yq, since that
isolates *this change's* effect from the 2,532 pre-existing yq divergences in the same
matrix (container equality, `---` separators, `1.5` vs `"1.5"`).

## Residual divergences are pre-existing and family-wide, not introduced here

The 23 malformed-document divergences fall into two classes, both confirmed identical on a
pre-fix build:

- **`.a` on a document malformed elsewhere** — succinctly's lazy semi-indexing returns the
  value where jq's eager parse errors. The documented architectural trade-off; no
  comparison involved.
- **`first(1==1)` / `[limit(1; 1==1)]`** still answer `true`. This is *not* specific to
  `Compare`: the already-gated `Expr::Arithmetic` sibling has it too (`first(1+1)` → `2`,
  `[limit(1; 1+1)]` → `[2]` on `{123: 1}`), while `and`/`or` raise correctly. So these
  short-circuiting consumers bypass the bridge for the whole family. Filed as a follow-up
  rather than widened into this PR.

This change puts `Compare` exactly on par with `Arithmetic` — it does not introduce the
residual, and closes the bare and `try`-wrapped cases the issue actually reports.

## Test plan

- [x] `cargo clippy --all-targets --features std,simd,serde,cli,regex,bench-runner,large-tests,mmap-tests -- -D warnings` (the CI leg `--all-features` hides, since it enables `scalar-yaml`)
- [x] Oracle diffs against `/usr/bin/jq` 1.7.1 and `yq` v4.53.3 directly (table above), not just golden fixtures
- [x] Negative-tested the `uses_cursor_metadata_builtins` half of the gate — removing it fails both `di` tests
- [x] New: `test_compare_raises_on_document_level_decode_failure_2581` (both malformation shapes, `try`-wrapped, well-formed controls, path-context control)
- [x] New: `test_compare_over_document_index_keeps_native_arm_2581` (every row captured live from yq v4.53.3)

Fixes #2581

### DIFF
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -6546,7 +6546,43 @@ fn eval_single<S: EvalSemantics, V: DocumentValue>(
         // so `("A"|stderr) == (("B"|stderr), ("C"|stderr))` now writes
         // `B A C A`, matching jq, instead of finishing right first (`B C A
         // A`).
-        Expr::Compare { op, left, right } => {
+        //
+        // **Gated on `needs_path_context`, like the `Expr::Arithmetic`/
+        // `Expr::And`/`Expr::Or` arms below** (#2581). This arm was ungated
+        // until then, which is the "pre-existing gap" those arms' own
+        // comments used to name: `1==1` on a #1194-malformed document
+        // (`{123: 1}`) answered `true`, where jq raises the document's own
+        // parse error on *every* query over it, including one that reads
+        // nothing. Live against jq 1.7.1:
+        //
+        // ```console
+        // $ printf '{123: 1}' | jq -c '1==1'
+        // jq: parse error: Object keys must be strings at line 1, column 5
+        // $ printf '{"a": {"bad": "\x"}, "b": 5}' | jq -c '.b == .b'
+        // jq: parse error: Invalid escape at line 1, column 18
+        // ```
+        //
+        // The gate keeps a comparison that reads no path context on the
+        // wildcard bridge, whose ambient materialization is what surfaces
+        // that decode failure; one that does read path context could not
+        // have stood on a malformed document's root and survived the walk
+        // anyway, and keeps the native arm.
+        //
+        // `uses_cursor_metadata_builtins` is the second half of the gate for
+        // the reason `eval_using`'s own doc comment already records for
+        // `input`/`inputs`: the bridge's re-serialize-and-reindex round trip
+        // cannot preserve a cursor-metadata builtin's real identity, so
+        // `select(di == 1)` over a multi-document stream would answer from a
+        // fixed-default stub instead of the live document index. The three
+        // sibling arms below carry only the `needs_path_context` half and
+        // look equally exposed -- unconfirmed, and tracked at #2584 rather
+        // than widened speculatively here.
+        Expr::Compare { op, left, right }
+            if needs_path_context(left)
+                || needs_path_context(right)
+                || crate::jq::walk::uses_cursor_metadata_builtins(left)
+                || crate::jq::walk::uses_cursor_metadata_builtins(right) =>
+        {
             eval_compare_generic::<S, V>(*op, left, right, value, optional, cursor)
         }
 
@@ -6557,7 +6593,7 @@ fn eval_single<S: EvalSemantics, V: DocumentValue>(
         // eager bridge used to hand it `null + 10`. That is what admits
         // `Expr::Arithmetic` to `path_context_single_native`.
         //
-        // **Gated on `needs_path_context`, unlike the `Expr::Compare` arm
+        // **Gated on `needs_path_context`, like the `Expr::Compare` arm
         // above.** The eager bridge materializes the ambient value on its
         // way through, and for a #1194-malformed document (`{123: 1}`) that
         // materialization is the decode failure jq answers *every* query on
@@ -6568,9 +6604,9 @@ fn eval_single<S: EvalSemantics, V: DocumentValue>(
         // arithmetic that reads no path context on the bridge, where it
         // still pays the ambient decode; arithmetic that does read path
         // context could not have been on a malformed document's root and
-        // survived the walk anyway. (`Expr::Compare`'s own arm already
-        // diverges here -- `1==1` on `{123: 1}` answers `true` -- which is a
-        // pre-existing gap, not one this arm widens.)
+        // survived the walk anyway. (`Expr::Compare`'s own arm carried
+        // exactly this gap -- `1==1` on `{123: 1}` answered `true` -- until
+        // #2581 gave it the same gate; see its comment above.)
         Expr::Arithmetic { left, right, .. }
             if needs_path_context(left) || needs_path_context(right) =>
         {
@@ -6617,8 +6653,8 @@ fn eval_single<S: EvalSemantics, V: DocumentValue>(
         // evaluator a document re-rooted at this stage's input.
         //
         // **Gated on `needs_path_context`, for the same reason the
-        // `Expr::Arithmetic` arm above is** (and the `Expr::Compare` arm is
-        // not): the bridge's ambient materialization is what makes
+        // `Expr::Arithmetic` and `Expr::Compare` arms above are**: the
+        // bridge's ambient materialization is what makes
         // `try (1+1) catch "x"` fail on a #1194-malformed document, and
         // `test_try_catch_contains_a_genuinely_catchable_malformed_key_error_1812`
         // pins that. An `and`/`or` that reads no path context stays on the
--- a/tests/jq_cli_tests.rs
+++ b/tests/jq_cli_tests.rs
@@ -30859,13 +30859,53 @@ fn test_as_pattern_alternatives_retry_under_a_wrapping_consumer_1519() -> Result
         // span, matching `test_select_raises_on_decode_failure_instead_of_silently_truthy_1645`'s
         // shape) -- the already-decoded first item must survive as a
         // `Partial`, not be discarded by the second's bare error.
+        //
+        // #2581 moved the condition off `==`. With `Expr::Compare` now
+        // carrying the same `needs_path_context` gate as its
+        // `Arithmetic`/`And`/`Or` siblings, `$x == 1` reads no path
+        // context and so routes through the wildcard bridge, whose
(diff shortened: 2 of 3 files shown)


newhoggy:
## Moving to draft — this change is correct about the gap and wrong about the mechanism

Post-review update. The gate does close #2581's oracle divergence, and every gate I claimed
above still holds. But review plus my own independent reproduction found that routing more
traffic through the wildcard bridge carries three side effects the PR body did not account
for, and two of them are user-visible regressions **this PR would introduce**. I am not
merging it as written.

### 1. Duplicate mapping keys collapse (new correctness regression)

The bridge materializes through an `IndexMap`-backed `OwnedValue`, which dedupes keys before
either operand runs:

```console
$ printf 'b: 1\na: 2\nb: 3\n' | yq 'length == 3'
true
$ printf 'b: 1\na: 2\nb: 3\n' | succinctly yq 'length'          # this branch
3
$ printf 'b: 1\na: 2\nb: 3\n' | succinctly yq 'length == 3'     # this branch
false
```

The binary contradicts itself. My 7,854-case yq sweep missed this because the matrix had no
duplicate-key document in it — the sweep's alphabet was part of the claim, and it was too
narrow.

### 2. ~6x time, ~6x peak RSS on every plain comparison (new perf regression)

Independently measured on a 3.4 MB document, `.[0] == .[0]`, same build profile both sides:

| | time | peak RSS |
|---|---|---|
| pre-fix | 0.09 s | 17.9 MB |
| this branch | 0.59 s | 105.9 MB |

`.users[0].name == "user0"`-shaped filters read the document but do not need *path context*,
so they now materialize + re-serialize + reindex it.

### 3. `1==1` on a >256-deep document becomes a raw panic

```console
$ succinctly jq -c '1==1' deep.json     # pre-fix: true
thread '<unnamed>' panicked at src/jq/eval_generic.rs:601:5:
```

Real jq gives a clean `Exceeds depth limit` error. The panic is in the library crate.

### The mechanism, not this PR, is the problem

All three are properties of the bridge, and all three are **already reachable on `main`**
through the identical gate on `Expr::Arithmetic`/`And`/`Or` (#1812/#2286) — verified: `length
+ 0` answers `2` where yq answers `3`, and `1+1` already panics on a deep document. This PR
does not introduce the defects; it widens them to `==`. Filed separately as #2626, #2627 and
#2628 since they are live today regardless of what happens here.

That is why I have not simply narrowed the gate: every width of it inherits the same three
side effects, because they belong to the materialization, not to the predicate selecting it.

Holding as a draft while #2581 settles which mechanism the family should use — see the
question posed there. The branch is worth keeping as-is: it is the working demonstration
that the prescribed fix closes the gap, and the harness that surfaced what it costs.

newhoggy:
Closing alongside #2581, which is obsolete under #2103 and the ADR-0018 amendment adopted 2026-09-09 (see that issue's closing comment). The gate implemented here is correct against the pre-#2103 rule and remains the working proof that the gap closes, plus the harness that measured its cost — nothing needs redoing if the amendment is ever revisited. The defect that survives points the other way and is filed as #2692.

PR #2630, opened 2026-09-08
fix(jq): re-enter foreach's fold register every step, not just the first

### LINKED ISSUE #2161: jq: path(foreach) with UPDATE navigating a Field on every step wrongly refuses after the first step
**Severity: Low-Medium** (silent-refuse: succinctly raises where jq succeeds, on a
non-navigating SOURCE -- so this is a strict subset of "wrong output" rather than a
fabricated/wrong path).

## Summary

Surfaced during code review of #2031's fold-register fix. This is a **simpler, cleaner
repro of the same underlying mechanism** as one of that PR's residual fuzzer diffs
(the "both-error, differ" case, `path(foreach ((1,2,keys[])) as $k (.b; .a))`), isolating
it away from `keys[]`/SOURCE-trackability entirely: SOURCE here is a plain, non-navigating
comma list.

\`\`\`console
$ echo '{"a":1}' | jq -c 'path(foreach (1,2,3) as $k (.b; .a))'
["b","a"]
["b","a"]
["b","a"]
$ echo '{"a":1}' | succinctly jq -c 'path(foreach (1,2,3) as $k (.b; .a))'
["b","a"]
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of null
\`\`\`

Real jq happily repeats \`["b","a"]\` for every one of the three (non-navigating) source
elements. succinctly gets the *first* step right, then raises on the second.

## Root cause (as far as diagnosed)

\`resolve_leaf\` (\`src/jq/eval.rs\`, untouched by #2031) unconditionally raises on any
\`Field\`/\`Index\` navigation when ambient \`trackable=false\`. After step 1's UPDATE (\`.a\`)
navigates to a path that doesn't match the fold's single persistent register's own fixed
path, \`FoldRegister::branch_provenance\`'s answer for that step sets \`state_at_register\`
to \`false\` going into step 2 -- so step 2's own genuine \`.a\` navigation (which real jq
handles identically to step 1's, since nothing about the register actually changed between
identical steps) gets refused as untracked instead of resolved.

Not fully traced past this point -- whoever picks this up should confirm this diagnosis and
determine the right fix (likely: the persistent register's fixed-per-INIT-fork model needs
to recognize "this step's UPDATE navigated identically to how a previous step's did" as
still-valid, rather than needing register identity to survive across steps at all for a
SOURCE that never itself moves the register).

## Related

- #2031 / PR (branch \`issue-2031-foreach-path-register-clobber\`) -- the fix whose review
  surfaced this as a cleaner isolation of one of its own residual fuzzer diffs.

newhoggy:
## Summary

`resolve_foreach` decided each step's `at_register` from the previous step's
`FoldRegister::branch_provenance`, which answers *"did the last UPDATE branch end **at** the
register's own path"*. The register is fixed for the whole fold and UPDATE's whole job is to
navigate away from it — so that answer is `false` from step 2 onward, and every later step
refused its own identical navigation as untracked:

```console
$ echo '{"a":1}' | jq -c 'path(foreach (1,2,3) as $k (.b; .a))'
["b","a"]
["b","a"]
["b","a"]
$ echo '{"a":1}' | succinctly jq -c 'path(foreach (1,2,3) as $k (.b; .a))'
["b","a"]
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of null
```

`register_identical`'s own doc comment already states the model the code was missing — the
register is *"fixed once per INIT fork, never advances across source-element iterations"* —
so what decides a step is jq's `jv_identical(current_input, value_at_path)`: is the
accumulator arriving at this step still the register's value. That is exactly the check the
per-source-element override arm directly above already performs; this arm was comparing
paths instead.

I confirmed the issue's suggested `resolve_leaf`/`branch_provenance` diagnosis before
fixing, and derived jq's model from the oracle rather than from the issue text — six probe
shapes pin it, and the fix reproduces all six including error-message contents.

## Why it's an `||`, not a swap

My first version replaced the path comparison with the identity check, and it broke step 1
for object-valued registers. `register_identical` demands `snapshot || Null | Bool` —
structural equality is not jq's pointer identity, so a container counts only when it is
known to be the frozen node itself.

So neither half subsumes the other:

| | carried provenance | identity check |
|---|---|---|
| step 1, object-valued register | **correct `true`** | `false` (container, no snapshot) |
| step 2+, `null`/`bool` accumulator | `false` (path moved) | **correct `true`** |

Dropping either half regresses one of the shapes the new test pins, which is why the test
covers both directions rather than only the issue's repro.

## Oracle verification

A generated sweep — 4 source streams × 7 INITs × 11 UPDATEs × 12 documents, plus EXTRACT
variants, **12,048 cases** — captured from a pre-fix and a post-fix build and classified
against jq 1.7.1 case by case:

| | count |
|---|---|
| cases changed by this fix | 880 |
| ...now matching jq (fixed) | **880** |
| ...no longer matching (broken) | **0** |
| ...changed, neither matches | **0** |

stdout, stderr **and** exit code are all compared, so the error-message shapes are part of
the claim, not just the successful paths. Comparing pre-fix against post-fix directly is
what isolates *this change's* effect from the pre-existing divergences in the same matrix.

yq mode (`eval.rs` is shared): 1,620 fold cases captured both ways, **40 changed**, all the
same fix shape. `path` in yq mode is a `--jq-extensions`-gated jq extension that real yq's
lexer rejects outright, so there is no yq oracle to diverge from — following jq's model is
the correct behaviour for that surface.

## Not fixed here: `reduce` has the identical bug

The sweep still reports 728 divergences after this fix, and **all 728 are `reduce`**.
`resolve_reduce` carries the same path-comparison at its own carry-forward site and is
untouched by this PR.

I did prototype the analogous change there to make the follow-up actionable: it fixes 696
of the 728 and breaks 0, leaving 32. That is strong shared-root-cause evidence, but 696
behaviour changes in a second construct deserve their own PR and their own review rather
than riding along here. Filed as a follow-up with that measurement.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — full suite green (60 suites)
- [x] 12,048-case jq oracle sweep, pre/post classified against jq 1.7.1 (table above)
- [x] 1,620-case yq-mode pre/post capture
- [x] New `test_foreach_reenters_register_every_step_not_just_the_first_2161` — six live jq 1.7.1 shapes, covering both halves of the `||` and both the emit and the raise directions

Fixes #2161

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -28126,6 +28126,16 @@ fn resolve_reduce<'a, S: EvalSemantics>(
                     // every step *without* its own source-derived register
                     // (including reduce's own final emission below) is
                     // still checked against.
+                    //
+                    // **#2161**: `resolve_foreach`'s equivalent site no
+                    // longer treats provenance as the whole rule -- it also
+                    // consults `register_identical`, so a step whose
+                    // accumulator is still the register's own value
+                    // re-enters it. This site was deliberately left alone
+                    // (fixing it moves 696 more cases onto jq, measured),
+                    // so `path(reduce (1,2) as $k (.b; .a))` on `{"a":1}`
+                    // still refuses where the `foreach` spelling succeeds.
+                    // Tracked at #2632.
                     let last = branches.into_iter().last();
                     (acc_at_register, acc_snapshot) = reg.branch_provenance(last.as_ref());
                     acc = last.map(|b| b.value.into_owned());
@@ -28187,6 +28197,16 @@ fn resolve_reduce<'a, S: EvalSemantics>(
 /// EXTRACT) rather than folded away, and the source stream's own trailing
 /// control is applied per-fork after that fork's own inner loop (#534
 /// follow-up), not deferred until every fork has run.
+///
+/// **#2161 adds a fourth difference, and it is a divergence from
+/// [`resolve_reduce`] rather than a shared rule**: this function re-enters
+/// the fixed register on every source element, deciding each step's
+/// `at_register` from jq's own `jv_identical(current_input, value_at_path)`
+/// (via [`register_identical`]) *as well as* the previous step's
+/// [`FoldRegister::branch_provenance`]. `resolve_reduce`'s own carry-forward
+/// still uses provenance alone, so `path(reduce (1,2) as $k (.b; .a))` on
+/// `{"a":1}` still refuses where the `foreach` spelling now succeeds --
+/// tracked at #2632, not an intended asymmetry.
 #[allow(clippy::too_many_arguments)] // STYLE-0004: `snapshot` (#1591) joins `trackable` as the
                                      // ambient pair threaded verbatim through every one of
                                      // `resolve_node`'s ~20 recursive call sites in this file; a
@@ -28338,6 +28358,21 @@ fn resolve_foreach<'a, S: EvalSemantics>(
                 // `true` for a `null`/`bool` accumulator. Dropping either
                 // half regresses one of the two shapes below.
                 //
+                // **jq mode only**, for exactly the reason
+                // [`trackable_step_register_eligible`] gates its own
+                // re-establishment the same way: the new half turns a step
+                // that would have refused into one that navigates, and on
+                // the *write* side that is a write where yq no-ops. Live
+                // against yq v4.53.3, `(null | .a) = 5` on `null` is a
+                // no-op, so `(foreach (1) as $k (null; .a)) = 5` writing
+                // `a: 5` would be exactly the silent corruption that gate's
+                // doc comment calls "a worse outcome than the refusal it
+                // would replace". yq mode therefore stays refuse-only here,
+                // as it already is for the single-step and carried-forward
+                // shapes. The carried half is left ungated: it is the
+                // pre-existing behaviour, not something this change
+                // introduces.
+                //
                 // Live against jq 1.7.1, on `{"a":1}`:
                 //
                 // ```console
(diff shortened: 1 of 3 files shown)


newhoggy:
Addressed both review findings in `27c7263a9` (now `206eb709e` after rebase), and
re-verified the whole sweep afterwards.

**1. yq-mode write — confirmed and gated.** Reproduced it exactly as reported:

```console
$ printf 'null\n' | succinctly yq '(foreach (1) as $k (null; .a)) = 5'
a: 5                                  # before the fix commit
$ printf 'null\n' | yq '(null | .a) = 5'
null                                  # real yq v4.53.3 -- a no-op
```

The new half is now gated `S::TAG == EvalTag::Jq`, the same way
`trackable_step_register_eligible` gates the identical class, and for its stated reason
("silently corrupting data is a worse outcome than the refusal it would replace"). The
carried provenance half stays ungated, since it is pre-existing behaviour rather than
something this branch introduces. Pinned by a new yq-side test asserting the refusal on
both the one-step and two-step shapes.

jq mode is unaffected: `(foreach (1) as $k (null; .a)) = 5` is still `{"a":5}`, matching
real jq.

**2. Stale doc comments — fixed at both sites.** `resolve_foreach`'s comment now names the
fourth difference explicitly instead of claiming there are three, and `resolve_reduce`'s
carry-forward records that it deliberately keeps the provenance-only rule. Both point at
**#2632**, which I filed with the measured prototype result (696 fixed / 0 broken / 32
residual) plus the two traps that cost me time here: the `||` being load-bearing, and the
yq gate this review caught.

**Re-verification after the change and the rebase onto `a79ea8997`:**

- 12,048-case jq sweep re-captured against the gated build: still **880 fixed, 0 broken, 0
  neither** — the gate cost none of the jq-mode wins.
- Full suite, both clippy legs, `fmt`, `no_std` and rustdoc all green post-rebase.

Thanks for both — the first one in particular was a real regression I would have shipped.

PR #2634, opened 2026-09-08
fix(jq): locate an unresolved call by its real position, not a name match

### LINKED ISSUE #2085: jq: unresolved-call location search can cite an unrelated same-spelling occurrence (object key, variable)
**Severity: Low**

## Summary

`locate_identifier_from` (`src/bin/succinctly/jq_runner.rs`, used by `report_unresolved_calls` to word-bound-search the filter source for an unresolved call's line) matches any occurrence of the offending name's spelling in the filter text — not specifically a call site. An unrelated occurrence of the same spelling earlier in the source (an object key, a variable name) is indistinguishable from the real call to a pure text scan, so the reported line can be wrong even when there's only one actual call.

## Repro

```
$ /usr/bin/jq -n '{nosuch: 1}
| nosuch'
jq: error: nosuch/0 is not defined at <top-level>, line 2:
| nosuch
jq: 1 compile error

$ succinctly jq -n '{nosuch: 1}
| nosuch'
jq: error: nosuch/0 is not defined at <top-level>, line 1:
{nosuch: 1}
jq: 1 compile error
```

Real jq cites line 2 (the actual failing call); succinctly cites line 1 (a harmless object key that happens to share the identifier's spelling).

## Root cause

`src/bin/succinctly/jq_runner.rs`'s `locate_identifier_from` — a word-bounded substring search over the raw filter text, with no awareness of whether a given occurrence is actually an `Expr::FuncCall` node that `jq::resolve::check` visited, or something else entirely (an object key, a bound variable, a string that happens to contain the name, etc). This predates #2037 — it was present in the original #1473 implementation and confirmed reproducible on `main` before #2037's changes — but was found and independently verified during #2037's review (PR #2082).

## Why this is separate from #2037

#2037 fixed two other edges of the same "no source position on `Expr::FuncCall`" gap: only the first unresolvable call was reported (now all are), and a repeated name's second/third occurrence was misattributed to the first (now the search resumes per-name from the previous match). Both of those fixes are pure heuristic improvements over the *same* underlying text-search mechanism this issue is about — they don't and can't fix a same-spelling-but-unrelated collision, since the search still has no way to know an occurrence isn't a call site. Fixing this for real needs the same real source position on `Expr::FuncCall` that #2037's own scope explicitly deferred (perturbs `format!("{body:?}").len()`, which #1381's `MAX_FUNC_EXPANSION_WEIGHTED_COST` is calibrated against, and needs recalibrating that constant plus handling the `PartialEq`/test-assertion fallout across ~14 `FuncCall` construction sites).

## Suggested fix direction

Only a real position on `Expr::FuncCall` (or a parallel position table built during parsing, never embedded in `Expr` itself, keeping `Debug`/`PartialEq` untouched — worth evaluating as a lower-risk alternative to a field on `Expr`) closes this properly. Until then, this is a known, documented approximation (see `docs/compliance/jq/limitations.md`'s "Undefined functions and arity mismatches" section and the doc comment on `report_unresolved_calls`).

newhoggy:
## Summary

`report_unresolved_calls` found each diagnostic's line by scanning the filter text for the
offending identifier. That matches *any* occurrence of the spelling, not a call site — so a
decoy earlier in the source got cited instead of the real call:

```console
$ jq -n '{nosuch: 1}
| nosuch'
jq: error: nosuch/0 is not defined at <top-level>, line 2:
$ succinctly jq -n '{nosuch: 1}
| nosuch'
jq: error: nosuch/0 is not defined at <top-level>, line 1:
```

This adds `jq::collect_call_sites`, a parallel position table built during a parse:
`Parser` records the byte offset of every identifier that reaches the generic
`parse_func_call_or_error` path. Only genuine calls go through it, so an object key, a
`$`-variable, a `.field`, or a string spelling the same name is **absent from the table** —
the diagnostic cannot pick one of those. The lookup also matches on **arity**, so a
resolvable `f/2` is not cited for a failing `f/1`. It is a table of *calls*, not of
*failing* calls, and carries no scope information, so same-name same-arity calls separated
only by lexical scope still cannot be told apart — `(def f: 1; f) | f` cites line 2 where jq
cites line 3, tracked at #2635 (`main` cited line 1). The text search
stays as the fallback for calls inlined from an `include`d module or `~/.jq`, which have no
occurrence in the filter at all.

## The issue's stated blocker no longer applies

#2085 says a real position "needs recalibrating `MAX_FUNC_EXPANSION_WEIGHTED_COST`", which
is calibrated against `format!("{body:?}").len()`. **All three expansion budgets have since
been deleted** — `grep` finds `MAX_FUNC_EXPANSION_DEPTH`/`_CHAIN_DEPTH`/`_WEIGHTED_COST`
only inside doc comments describing the history, and `eval.rs` states outright that
`MAX_EVAL_FRAMES` "replaces three substitution budgets ... that bounded *expansion* work
rather than evaluation". So there is no constant left to recalibrate and no `Debug`-length
sensitivity to perturb.

I took the lower-risk option the issue itself proposes anyway — a table rather than a field
on `Expr` — so `Expr`'s size, `Debug` and `PartialEq` are untouched either way, and the
~14 `FuncCall` construction sites need no changes. The stale doc comment in `jq_runner.rs`
that still cited the deleted constant as a live constraint is rewritten here too.

## Cost

The table is collected by re-parsing, on the error path only — one extra parse immediately
before the process aborts with a compile error, so nothing anyone waits on.

The parse itself is not entirely free: `parse_func_call_or_error` now does a `name.clone()`
and a `Vec` push per user-function call site on every parse. That is once per CLI invocation
over a filter string, not per input value.

## Measured against the oracle

28 unresolved-call shapes (including the three the review added), each compared to `/usr/bin/jq` 1.7.1 on stderr, run against a
pre-fix and a post-fix build:

| | count |
|---|--:|
| matching jq **before** | 13 / 28 |
| matching jq **after** | **27 / 28** |
| improved by this change | **14** |
| regressed | **0** |

The improved shapes are the decoy family — object key (bare and quoted), string literal,
`$`-variable, `.field`, nested key, key-among-keys — plus the arity shapes. The single
remaining mismatch is the same-arity lexical-scope case tracked at #2635.

## Behaviour deliberately preserved

- **Repeated undefined names** still walk their own successive call sites in source order
  rather than repeating the first — the positional lookup keeps a per-name counter, mirroring
  the fallback's `resume_from`.
- **The table is sorted and deduped by offset**, so the rewind-and-retry parsing that
  #2110/#2237 introduced cannot double-record a single call.
- **yq is untouched** — `report_unresolved_calls` is jq-only (`yq_runner` has its own path,
  verified still reporting `nosuch/0 is not defined` unchanged).

## Test plan

- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` — `collect_call_sites`/`CallSite` are new public API, so this one matters here
- [x] 28-shape pre/post oracle matrix against jq 1.7.1 (table above)
- [x] New `test_unresolved_call_line_comes_from_the_call_not_a_name_match_2085` — 8 decoy shapes, every expectation a live jq capture
- [x] New `test_unresolved_call_positional_lookup_keeps_repeat_arity_and_fallback_2085` — repeat-walk, arity, and a real-module fallback case; negative-tested (removing the fix makes it fail)

Fixes #2085

### DIFF
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -1072,13 +1072,27 @@ fn report_unresolved_calls(unresolved: &[UnresolvedCall], filter: &str) {
 
     for UnresolvedCall { name, arity } in unresolved {
         let taken = consumed.entry(name.as_str()).or_insert(0);
+        // Matched on arity as well as name (#2085 review, finding 2): an
+        // `f/1` diagnostic must not take a perfectly resolvable `f/2` call
+        // site earlier in the source. `def f(a;b): a; f(1;2) | f(1)` cited
+        // the `f(1;2)` on line 2 instead of the failing `f(1)` on line 3.
         let from_table = call_sites
             .iter()
-            .filter(|c| c.name == *name)
+            .filter(|c| c.name == *name && c.arity == *arity)
             .nth(*taken)
             .map(|c| c.offset);
         if let Some(offset) = from_table {
             *taken += 1;
+            // Keep the fallback's own cursor in step (#2085 review, finding
+            // 1). These are two counters over the same name, and letting
+            // them drift re-reports a position already used: once the table
+            // runs out -- exactly the module case, where a call inlined from
+            // `include`/`~/.jq` has no occurrence in `filter` at all -- the
+            // fallback would restart from offset 0 and invent a line for a
+            // call that has none. With `def helper: nosuch;` in a module,
+            // `include "m"; helper | nosuch` printed the `line 3` marker
+            // twice rather than leaving the module's own error unmarked.
+            resume_from.insert(name.as_str(), offset + name.len());
             let (line_no, line_text, column) = line_at_offset(filter, offset);
             eprintln!("jq: error: {name}/{arity} is not defined at <top-level>, line {line_no}:");
             eprintln!("{line_text}{}", " ".repeat(column));
@@ -1116,25 +1130,6 @@ fn report_unresolved_calls(unresolved: &[UnresolvedCall], filter: &str) {
 /// search for `f` does not match the `f` inside `first` — but `::` is allowed
 /// on the left, since a namespaced call arrives here as `ns::f` while the
 /// source spells the two halves either side of the separator.
-/// The 1-based line number, that line's text, and the 0-based column, for a
-/// byte `offset` into `filter` (#2085).
-///
-/// The positional counterpart of [`locate_identifier_from`]'s own line
-/// arithmetic, for an offset that is already known to be a real call site
-/// rather than one that had to be searched for.
-fn line_at_offset(filter: &str, offset: usize) -> (usize, String, usize) {
-    let line_start = filter[..offset].rfind('\n').map_or(0, |i| i + 1);
-    let line_no = filter[..line_start].matches('\n').count() + 1;
-    let line_end = filter[line_start..]
-        .find('\n')
-        .map_or(filter.len(), |i| line_start + i);
-    (
-        line_no,
-        filter[line_start..line_end].to_string(),
-        offset - line_start,
-    )
-}
-
 fn locate_identifier_from(
     filter: &str,
     name: &str,
@@ -1173,6 +1168,25 @@ fn locate_identifier_from(
         start - line_start,
         start + name.len(),
     ))
+}
+
+/// The 1-based line number, that line's text, and the 0-based column, for a
+/// byte `offset` into `filter` (#2085).
+///
+/// The positional counterpart of [`locate_identifier_from`]'s own line
+/// arithmetic, for an offset that is already known to be a real call site
+/// rather than one that had to be searched for.
+fn line_at_offset(filter: &str, offset: usize) -> (usize, String, usize) {
+    let line_start = filter[..offset].rfind('\n').map_or(0, |i| i + 1);
+    let line_no = filter[..line_start].matches('\n').count() + 1;
+    let line_end = filter[line_start..]
+        .find('\n')
+        .map_or(filter.len(), |i| line_start + i);
+    (
+        line_no,
+        filter[line_start..line_end].to_string(),
+        offset - line_start,
+    )
 }
 
 /// Extract the line containing an error for display.
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -1023,27 +1023,28 @@ fn print_validation_error(err: &ValidationError, input: &[u8], filename: Option<
 /// jq: 1 compile error
 /// ```
 ///
-/// **Each line is located by searching `filter` for the offending identifier,
-/// not read off the AST** — `Expr::FuncCall` carries no source position, and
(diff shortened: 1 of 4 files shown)


newhoggy:
All four addressed in `8f17d2e52` + `22541259a` (now rebased onto `ae32ffce8`). I
reproduced each one first rather than taking them on faith — all four were real, and the
first was a regression this branch introduced that I would otherwise have shipped.

### 1. Fallback restarting at offset 0 — regression, fixed

Reproduced exactly with your repro. `consumed` and `resume_from` are two cursors over the
same name and only one was advancing, so once the table ran out the fallback searched from 0
and re-used a position:

```console
$ succinctly jq -L . -n 'include "m";
helper
| nosuch'                       # with `def helper: nosuch;` in m.jq
... line 3:      <- the filter's own call, correct
... line 3:      <- the module's call, inventing a position   # before
... (no marker)  <- after, matching main and this function's own doc comment
```

A table hit now advances `resume_from` too.

### 2. `CallSite` had no arity — fixed

`def f(a;b): a; f(1;2) | f(1)` cited line 2 where jq cites line 3. `CallSite` now records
`args.len()` and the lookup matches on it; that shape is now line 3.

**Your broader point stands and I have corrected the claim.** "Decoys are absent by
construction" was true only of the *syntactic* decoys (object key, `$`-variable, `.field`,
string literal) and I overstated it into a general property. The table records calls, not
*failing* calls, and carries no scope information. Arity narrows it, but your same-arity
scope variant `(def f: 1; f) | f` still cites line 2 against jq's line 3 — the parser has no
scope information and `resolve.rs` has no positions, so nothing currently joins the two
halves. That is now documented on `collect_call_sites` and filed as **#2635** with your
repro. (`main` cited line 1 there, so it is still an improvement — just not a fix, and the
PR body no longer implies otherwise.)

### 3. Doc comment attached to the wrong function — fixed

`line_at_offset` had been inserted between `locate_identifier_from`'s doc and its `fn`.
Moved below it; both now carry their own docs.

### 4. The fallback test never reached the fallback — fixed, and it was load-bearing

You are right that this is why finding 1 slipped through: a single-line filter makes the
table path and the fallback path both answer "line 1", so the assertion could not tell them
apart. It now builds a real module in a `TempDir` and asserts the `line 3:` marker appears
**exactly once**. I negative-tested it — removing finding 1's fix makes it fail with
`left: 2, right: 1`, and restoring it passes.

### Re-measured

Extended the sweep to 28 shapes, including your three:

| | count |
|---|--:|
| matching jq before this branch | 13 / 28 |
| matching jq after | **27 / 28** |
| improved | **14** |
| regressed | **0** |

The one remaining mismatch is the scope case from finding 2, now tracked at #2635.

### On your non-finding

Fair, and I have corrected "the successful path is completely untouched" in the PR body —
`parse_func_call_or_error` does now do a `name.clone()` and a `Vec` push per user-function
call on every parse. It is once per CLI invocation over a filter string, not per input
value, but the description should not have claimed zero.

Thanks — findings 1 and 4 together were the valuable pair, since the test gap is what let the
regression through.

Issue #2042, opened 2026-08-31
jq: widening TrackedVar's gate needs a bind-time path — provenance alone fabricates paths for equal-valued siblings


newhoggy:
**Severity: Low** (refuse-only — visible exit 5, no document written)

## Summary

Widening `substitute_bound_var`'s `is_identity_passthrough` gate — stage 3 of the plan on
#1573, and the half #2041 deliberately did **not** ship — is **unsound on its own**. It
reopens exactly the accept-where-jq-refuses class #1466 closed, because `register_identical`
compares *values plus a provenance bit*, and that bit says "never rebuilt", not "from this
position".

This is new information: #1573's own spike concluded the objection "only follows under a
**value** comparison" and that "under a **provenance** comparison it correctly refuses".
Measured, it does not.

## Repro (measured, not predicted)

With the gate widened to `substitute_var_tracked` unconditionally and #2041's register
threading in place:

```console
$ echo '{"a":{"b":1},"c":{"b":1}}' | jq -c 'path(.a as $y | .c | $y)'
jq: error (at <stdin>:1): Invalid path expression with result {"b":1}
$ echo '{"a":{"b":1},"c":{"b":1}}' | succinctly jq -c 'path(.a as $y | .c | $y)'
["c"]                                    # <- fabricated: .a and .c are DIFFERENT nodes

$ echo '{"a":{"b":1},"c":{"b":1}}' | jq -c 'path(.a as $y | reduce (1) as $i (.c; $y))'
jq: error ...
$ echo '{"a":{"b":1},"c":{"b":1}}' | succinctly jq -c 'path(.a as $y | reduce (1) as $i (.c; $y))'
["c"]                                    # <- same, through FoldRegister
```

Both are `path()` here, but the same wrongly-accepted path is what `del()`/`=`/`|=` would
then **write through** — a document jq leaves untouched. That is why it was reverted rather
than shipped.

## Root cause

`register_identical` (`src/jq/eval.rs`) models `jv_identical` as
`value == register && (snapshot || null/bool)`. The `snapshot` bit distinguishes a frozen
`as`-binding from a reconstruction, which is all #1466 needed — because the
`is_identity_passthrough` gate meant every marker was bound from `.`, i.e. always the root,
so two markers could never be confused. Once *navigated* bindings get markers, two distinct
nodes holding equal values are trivially reachable and the bit can no longer tell them
apart.

## What the marker actually needs

[...]

PR #2647, opened 2026-09-08
fix(jq): a bound variable stands at the node it was bound from (#2072)

### LINKED ISSUE #2072: jq: `as` bindings erase the bound node's path, so `. as $x | .[] | key` can't be routed correctly
Found while fixing #1409 (PR #2071), and deliberately excluded from it.

## Symptom

`Expr::As`/`Expr::AsPattern`'s path-context arms in
`eval_pipe_with_path_context_internal` are guarded on
`needs_path_context(expr) || needs_path_context(body)` — neither of which
considers `rest`. So the #1409 repro shape never reaches the (now fixed) arm at
all:

```console
$ echo '{"a":[1,2,3]}' | succinctly jq -c '.a | (. as $x | .[]) | key'
"a"
"a"
"a"
# expected 0, 1, 2 — and `.a | (. as $x | (.[] | select(key >= 0))) | key`,
# which does reach the arm, already gives 0, 1, 2 since #2071
```

## Why widening the guard is not the fix

Adding `|| rest.iter().any(needs_path_context)` does fix the above, but it
regresses a case that is currently right by accident:

```
eval_all_outputs(br#"[{"x":1},{"x":2}]"#, &[7, 8], ".[] | (.x as $v | $v) | file_index")
  before widening: ["7", "8"]
  after widening:  ["0", "0"]
```

Real yq agrees with the *former*:

```console
$ yq ea '.x as $v | $v | fileIndex' f1.yaml f2.yaml
0
1
```

## Root cause

`substitute_bound_var` re-materialises the bound value as a literal, so `$v`'s
output is indistinguishable from a fresh root and honestly has no path. Real yq
keeps the bound node's identity. Un-widened, the generic fallback happens to
report the *ambient* path, which coincides with the right answer whenever the
binding didn't move position — hence "right by accident".

Note the un-widened path is also wrong the other way, and this is
oracle-confirmed: `test_generic_fallback_continues_pipe_with_path_context`
asserts `.[] | (1 as $x | $x) | file_index` is `7`, `8`, but
`yq ea '1 as $x | $x | fileIndex' f1 f2` gives `0`, `0`. That assertion should
change as part of whatever fixes this.

## What a real fix needs

Variable bindings that carry the bound value's node identity (or at least its
path) rather than substituting a bare literal, so `$v` can report the path it
was bound from. Both #1409's probe and the ambient fallback are approximations
that each get one half right.

newhoggy:
Closes #2072.

## What

A bound variable now stands at the node it was bound from. Real yq's variables hold nodes -- parent pointers, anchors and style included -- so `.a.b as $x | $x | key` is `"b"` there; succinctly substituted a bare value and printed nothing.

- `Expr::TrackedVar` carries a `BoundVar` payload: the frozen value, the `tracked` flag (the `is_identity_passthrough` gate, moved into the node), and an optional `BindOrigin` naming the document node or owned-tree position the value came from. Still one pointer, so the `Expr` size pin holds.
- `DocumentCursor` gains `node_id` / `at_node_id` / `document_token`: a `'static` handle re-resolved to a live cursor at the point of use, and only ever applied to a cursor of the same document (the token is checked before the id is trusted).
- The bind sites record origins (`each_as_generic` keeps the cursors its bind produced; `eval_owned_identity_as` binds through the identity pipe). The use sites read them: the native `$x` arm re-emits the live node, the path-context walk steps a `$x` to its node, and the identity pipe's `Bound` rule, navigation step and operand identity rebuild the identity from the origin, falling back to the old value-equality rule only where no cursor domain exists.
- jq's `path()` resolver is untouched: it honours only `tracked`, so every jq-mode row matches jq 1.7.1 as before and #2042 (the jq-mode half) stays open, now with the bind-time node available to it.

## Results (yq v4.53.3, all captured live)

| filter | before | after / yq |
|---|---|---|
| `.a \| (.[] as $x \| $x) \| key` | (nothing) | `0 1 2` |
| `.a[0] as $x \| .a[1] \| $x \| path` | `[]` | `["a",0]` |
| `.a as $x \| .c \| $x \| path` (equal-valued siblings) | `[]` | `["a"]` |
| `.a as $x \| $x \| line` | `0` | `1` |
| `.b as $x \| $x` on `b: &anc [1, 2]` (YAML out) | `- 1` / `- 2` | `&anc [1, 2]` |
| `.a as $x \| ($x + [4]) \| path` | `[]` | `["a"]` |
| `. as $x \| ($x.a[0] + 1) \| path` | **process abort** (`unreachable!`, also on `main`) | `["a",0]` |

Writes through a variable (`$x[0] = 9`, `del($x[0])`) still refuse, file untouched, pinned in both modes.

## Perf

Same machine, interleaved per repetition, outputs identical: `.[] as $x | ...` over 2.4 MB and 24 MB arrays runs 18-38% faster than `main`, since a binding shares one `Rc` instead of re-spelling its value per occurrence and a `$x` use no longer takes the reindex bridge.

## Residues, recorded

A variable bound from an absent key (yq emits nothing at all; succinctly's value model has no absent-versus-null to bind) and the flow style of an anchor node read through an alias binding. Follow-ups are being filed from the plan's list.


newhoggy:
## Review response (commit `85b0e372d`)

**Headline — `BindOrigin::Owned` dropped `key_node`: confirmed and fixed.** `.a.b | key as $x | $x | key` printed `"b"`; yq v4.53.3 and succinctly's own unbound `key | key` print nothing. The origin now carries the flag, `identity_from_origin` restores it, and `owned_identity_after_stage` no longer clears it for the `Bound` rule (it cleared every rule but `KeyNode`, which would have undone the fix on the cursor-boundary path). Pinned with four live-captured rows in `test_bound_variable_carries_node_identity_2072`: `.a[0] | key as $x | $x | key` → nothing, `| tostring | key` → `0`, `| parent | key` → `"a"`, and the parenthesised `(.[0] | key) as $x` spelling.

**Secondary findings**

| # | finding | outcome |
|---|---|---|
| 1 | `BindOrigin::Owned` is a lossy re-encoding of `OwnedIdentity` | The lossy part was `key_node`; fixed. A sibling type is unavoidable because `OwnedIdentity` holds a `V::Cursor` (`'a`-borrowed) and an `Expr` must be `'static`; the two conversion functions are the only place the encoding lives. |
| 2 | origin-resolve-with-fallback copy-pasted at three sites, `owned_identity_operand` skipping the equality fallback | Folded into one `bound_var_identity`; all three sites call it, so the operand site now applies the same fallback. |
| 3 | `node_id`/`at_node_id`/`document_token` duplicated between the cursors | Declined. A trait default cannot construct a cursor (`at_node_id` needs the per-format constructor), and the only primitive the trait's own `same_node` default has, `text_position`, is neither total nor re-resolvable. The three impls are three lines each. |
| 4 | `materialize_bound_values_with_origin` re-implements `push_generic_owned_values`' eleven-arm match | Both are now `fold_generic_owned_values`, which hands each value the cursor it came from. |
| 5 | dead `passthrough` local | Removed. `eval_single`'s `Identity` arm already yields `OneCursor` whenever a cursor exists, so a `. as $x` bind takes the same arm as any node. |
| 6 | O(depth) identity-chain cost on the bind path | The chain cloned per bound value is `OwnedIdentity::ancestors`, whose length is the number of navigation steps taken *inside owned values within the identity pipe* for that binding, not document depth; the pipe already cloned the whole `OwnedIdentity` per stage before this branch. The 18–38% figure was measured on the generic route, which this does not touch. An identity-route A/B against `main` is not output-identical by construction (those queries are exactly the ones `main` answered wrongly), so I have no clean number to offer for it. |

Two pre-existing rows the new test surfaced while capturing, unrelated to bindings (identical on the base binary, unbound): `.a[0] | key | path` is `["a","0"]` in yq (a string `"0"`) vs `["a",0]` here, and `.a[0] | key | key` on a mapping is `0` in yq vs nothing here. Not touched.

Issue #2645, opened 2026-09-08
yq: eager and streaming answers disagree on a colliding undecodable key — the #2103 shape in yq mode


newhoggy:
**Severity: Low** (no wrong data; the answer depends on how the filter is spelled)

## Summary

`succinctly yq` has the same shape #2103 just removed from `succinctly jq`: on a document with two undecodable keys of the same display spelling, whether the document is accepted depends on which filter is asked.

```console
$ printf '"\\ud800": 1\n"\\ud800": 2\n' > d.yaml
$ succinctly yq '.'   d.yaml       # "": 1 / "": 2, exit 0
$ succinctly yq '.,.' d.yaml       # Error: object key "" is ambiguous: ..., exit 1
$ succinctly yq '1+1' d.yaml       # Error: object key "" is ambiguous: ..., exit 1
```

Real yq v4.53.3 rejects the document at parse time for every filter, so as in the jq case the raising column is the one closer to the oracle and the identity column is the pre-existing divergence.

## Why this is its own issue

#2103's fix was in `jq_runner.rs`'s M2 route (the `expr_is_cursor_transparent` gate and the eager `evaluate_bytes_lazy` route); `yq_runner.rs`'s `evaluate_yaml_cursor` is a different route with its own materializing steps (the DOM path taken by writes and DOM-forcing flags, ADR-0017), and the yq-side display-key fallback is `""` rather than the raw span (`YamlValue::key_string()` never returns `None`, #222). Whether the #2103 rule — a filter validates only what it reads — should apply to yq mode, and where the raise is coming from on this route, both need their own look.

## Related

- #2103 (jq decision and record), #1642/#1749 (the yq-side `DisplayKeyGuard`), #2168/#2604 (the rule).

Issue #2644, opened 2026-09-08
perf(jq): drop the #1812 needs_path_context gates on eval_single's arithmetic/negate/and/or arms (#2103 follow-up)


newhoggy:
**Severity: Low** (perf and simplification; no wrong output)

## Summary

`eval_single` in `src/jq/eval_generic.rs` keeps four arms on the eager wildcard bridge unless they read path context:

- `Expr::Arithmetic { .. } if needs_path_context(left) || needs_path_context(right)`
- `Expr::Negate(inner) if needs_path_context(inner)`
- `Expr::And(..)` / `Expr::Or(..)` with the same guard

Each guard's comment says it exists "for the same #1812 reason": an ungated arm "would stop paying the bridge's ambient decode on a malformed document", so `try (1+1) catch "x"` on `{123: 1}` would answer instead of exiting 5.

#2103 decided that question the other way and recorded it in `docs/compliance/jq/limitations.md` ("Every M2 filter streams — a filter validates only what it reads"): the CLI's M2 route now streams every filter, `1+1` on a malformed document answers `2`, and the ambient decode those gates preserve is exactly the accident the record retires. The gates now guard a behaviour the shipped CLI no longer has, and cost a whole-document materialization on every `1+1`, `-x`, `a and b` that reaches `eval_single` without path context.

## Ask

1. Drop the `needs_path_context` guards on the four arms so they always take the native `eval_each_generic`-backed path, and delete the "#1812 reason" comments (point at #2103's record instead).
2. Check what else still reaches those arms through `eval_single` — the `-n`/`--slurp`/`--input-dsv` routes go through `evaluate_input_streaming`, but any library caller of `eval_with_cursor` gets the eager evaluator directly. Confirm no pinned test outside the CLI depends on the ambient decode.
3. Measure with the interleaved A/B method (`docs/guides/benchmarking.md`) on both bench machines, inputs ≥ 1 MB, `1+1` / `.a + .b` / `-.a` / `.a and .b` shapes, output identity gated first. The 10 MB indicative figure from #2103 (`1+1`: 0.23 s / 237 MB eager vs 0.01 s / 20 MB streaming) is the expected order of magnitude, not a result.

## Related

- #2103 (the decision), #1812 (the original constraint, revised by #2286), #2168/#2604 (the rule).

newhoggy:
Closing as resolved-by-#2626 (agent `Blake`, backlog triage).

[...]

Issue #2286, opened 2026-09-03
jq: ? wrongly suppresses a malformed-JSON error that real jq raises unconditionally (parse error precedes every filter)


newhoggy:
## What's wrong

A malformed-JSON error surfaced by \`JsonFields::find_cursor\` (or any
other lazy-validation site raising \`EvalError::malformed_json_text\`) gets
silently swallowed by a trailing \`?\`, even though real jq's equivalent
failure is a **parse error** — raised before any filter, \`?\` included,
ever runs. \`?\` cannot suppress something jq never even reaches the filter
stage to see.

## Repro (confirmed live against jq 1.7.1)

\`\`\`console
$ echo '{"a" 1}' | jq -c '.a?'          # #1677's own missing-':' shape
jq: parse error: Expected separator between values at line 1, column 7
$ echo '{"a" 1}' | succinctly jq -c '.a?'
                                          # exit 0, no output -- wrongly suppressed

$ echo '{"a":1,123:2}' | jq -c '.a?'    # #1995's own non-string-key shape
jq: parse error: Object keys must be strings at line 1, column 11
$ echo '{"a":1,123:2}' | succinctly jq -c '.a?'
                                          # exit 0, no output -- wrongly suppressed
\`\`\`

Both repros are exit 5 (unconditional parse error) on real jq regardless
of the trailing \`?\`; succinctly gives exit 0 with no output for both.

## Confirmed pre-existing, not specific to #1995

The first repro (\`{"a" 1}\`, a missing \`:\`) is #1677's own,
already-shipped malformed-delimiter check — this bug predates #1995
entirely and isn't introduced or widened by it. #1995 just discovered the
identical suppression gap again via its own new non-string-key check,
since both raise through the same \`EvalError::malformed_json_text\`
constructor.

## Root cause (hypothesis, not fully investigated)

This crate already has an established "always uncatchable" classification
mechanism for exactly this kind of error — \`EvalError::is_decode_failure\`,
consulted by the value-position \`?\`/\`try\` dispatch sites (\`eval_try\`,
\`each_try\`, \`try_single_generic\`, \`each_try_generic\`) specifically so a
decode failure is never suppressed. \`EvalError::malformed_json_text\`'s
constructed errors are not tagged with \`ErrorKind::DecodeFailure\` (or any
other kind these dispatch sites check), so they fall through to ordinary,
suppressible-error handling instead.

## Suggested fix direction

Either give \`malformed_json_text\`'s errors an existing "always
[...]

Issue #2168, opened 2026-09-02
perf(jq): the decode-validity walk is now the majority of every cursor-resolved path query's cost


newhoggy:
**Severity: Medium** (perf; and it carries an unresolved semantics question). Surfaced
by #2061 / PR #2151, which removed everything *else* from these queries' cost.

## What

Now that `path(...)`, `path` (no-arg), `key`, `parent` and `parent(n)` resolve from
cursors and build no `OwnedValue` tree, the whole-document decode-validity walk
(`push_generic_truthiness_cursor_error`, `src/jq/eval_generic.rs`) is **the majority
of what these queries still cost**. It is an O(document) traversal charged to answer
something bounded — `.[0] | key` outputs `0`.

Measured on the merged result, interleaved, median of 7, idle dedicated hosts.
`.[0]` is the same navigation with no gate; the difference is the gate:

| host | input | `.[0] \| key` | `.[0]` | gate share | vs floor |
|---|---|---|---|---|---|
| 7950X | 6 MB | 116.4 ms | 38.2 ms | **67%** | 3.0× |
| 7950X | 20 MB | 419.0 ms | 128.4 ms | **69%** | 3.3× |
| M4 Pro | 6 MB | 94.9 ms | 26.5 ms | **72%** | 3.6× |
| M4 Pro | 20 MB | 374.3 ms | 92.2 ms | **75%** | 4.1× |

PR #2075 attributed the same thing independently by building without the walk:
"577ms of the 577ms".

## Two independent directions

**1. Fuse it into the walk (pure perf, no semantics change).** Today the gate is a
separate pass *before* the walk, so a document is traversed twice. The bridge it
replaced validated and built in one pass. Fusing would also remove the small
platform-split cost PR #2151 measured on the fan-out shape
`[.[] | .k.x | parent | key] | length`: −10% to −14% on the 7950X but a consistent
+2% to +3% on the M4 Pro, which is the extra pass showing up where the walk itself
saves least. That asymmetry is unexplained beyond "ARM pays relatively more for the
second traversal" and is worth confirming rather than assuming.

**2. Reconsider whether the gate should run for these queries at all** (semantics —
needs a decision, not just a patch). Its entire domain is documents no jq version
will process, confirmed live against both pinned binaries:

| document | jq 1.7.1 | jq 1.8.2 | succinctly's gate |
|---|---|---|---|
[...]

PR #2652, opened 2026-09-08
fix(jq): stream every M2 filter — a filter validates only what it reads (#2103)

### LINKED ISSUE #2103: jq: eager and streaming evaluators disagree on document validation and undecodable-key spelling

newhoggy:
Closes #2103.

## What changed

The M2 route (a JSON document from a file or stdin, the CLI's default) streamed only a *cursor-transparent* filter. `expr_is_cursor_transparent` kept every filter that forwards or discards the root cursor — `.,.`, `(., debug)`, `1+1`, `try (1+1)` — on the eager evaluator, because that route's `to_owned_with_cursor` doubled as a whole-document validity check and picked an undecodable key's spelling, and the streaming arms did neither. Every M2 filter now streams; the gate, the eager M2 route and its `SUCCINCTLY_JQ_M2_EVAL` hook are deleted.

```console
$ printf '%s' '{123:1,"b":2}' | succinctly jq -c '1+1'         # before: exit 5; now: 2, exit 0
$ printf '%s' '{123:1,"b":2}' | succinctly jq -c '.'           # exit 5, before and after (the walk reaches the fault)
$ printf '%s' '{"\ud800":1,"\ud800":2}' | succinctly jq -c '.,.'   # before: "ambiguous", exit 5; now: echoed twice, like `.`
$ printf '%s' '{"\ud800":1,"\ud800":2}' | succinctly jq -Sc '.'    # still "ambiguous", exit 5 (a map is built)
```

## This is a deliberate divergence from jq, recorded on its merits

Real jq 1.7.1 rejects every affected document at parse time, exit 5, whatever the filter. So ADR-0018's step 2 favours the behaviour being removed, and no rule-4 condition applies. The entry "Every M2 filter streams — a filter validates only what it reads" in `docs/compliance/jq/limitations.md` records it on the same footing as #2168's entry, with the same three reasons: the check was an accident of needing a value to bridge with; it made the answer depend on how the filter was spelled (`.`, `.b`, `length`, `label $x | .` answered on the same document that `.,.`, `1+1`, `debug`, `if`, `as` refused); and it cost a whole-document materialization on a filter that reads nothing. 19 rows move off jq's exit 5 (three filters × six documents, plus `.,.` on the colliding-key document). Six more `limit(N; keys_unsorted[])` cells move because the eager route materialized `limit`'s results and decoded the key — the inconsistency the issue's first comment recorded.

**Spelling.** No new spelling. The rule the #1642 entry already stated is now stated as the rule: raw source bytes wherever the value is never materialized, `key_display_string`'s doubled-backslash fallback wherever it is. `-S`, `-s`, `keys`, `debug` still print `"\\ud800"`.

**Cost** (10 MB `json generate` file, single laptop runs, indicative only — not the A/B method):

| filter | before (eager) | after (streaming) |
|---|---|---|
| `.,.` | 0.40 s, 374 MB RSS | 0.18 s, 25 MB |
| `1+1` | 0.23 s, 237 MB | 0.01 s, 20 MB |

## Commits

1. `fix(jq): keep the key cursor through an empty pipe tail` — `eval_each_pipe_generic`'s empty-stages arm dropped the cursor on a stale premise (the slice does reach zero length via `each_lazy_keys_iterate_sink` since #1565/#1770), so `keys_unsorted[]` on the streaming route decoded an undecodable key and raised.
2. `fix(jq): stream every M2 filter` — the gate goes; tests re-pinned row by row with the reason in each doc comment (1642, 1812, 1653, 1765, 1629, 2261, the materialized-`keys_unsorted` 1642 sibling) plus `test_root_forwarding_filters_stream_without_validating_2103` pinning the 19 rows, the controls that must not move, and the spelling rows. `scripts/jq-m2-streaming-sweep.sh` becomes a single-route sweep against pinned jq with a checked-in golden (`--update` regenerates).
3. `refactor(jq): delete the eager M2 route and its SUCCINCTLY_JQ_M2_EVAL hook`.
4. `docs(jq): record #2103's decision`.

## What the issue said that no longer held

The issue's "constraint that makes this non-trivial" — `try (1+1) catch "x"` on `{123: 1}` must yield `"x"` — was removed by #2286 before this work started; the pinned #1812 test already asserted exit 5. That is why the check could be dropped at the route level rather than "arm by arm".

## Follow-ups filed

- #2644 — drop the `needs_path_context` gates on `eval_single`'s arithmetic/negate/and/or arms that exist "for the same #1812 reason" (needs the interleaved A/B method on both bench machines).
- #2645 — yq mode shows the same shape on a colliding undecodable key.

## Verification

`cargo test --features cli` (all targets), both `ci.yml` clippy legs, `cargo fmt --check`, `RUSTDOCFLAGS=-D warnings cargo doc`, and `scripts/jq-m2-streaming-sweep.sh` matching its golden, all on the final commit.


newhoggy:
Thanks — both findings were right, and finding 1 was right where my first refutation was wrong. Addressed in bba166a31.

## Finding 1 — reproduced, attributed, recorded (not reverted)

My 294-case sweep missed it because my "pre" binary was built at commit 1 (Phase 1), which already carried the change — a stale baseline, my mistake. Rebuilding at the real merge-base settles it:

| `[{"x":"\ud800"}]` | `baa26d72e` (commit-1 parent) | `main` `4b1ebba80` | this branch |
|---|---|---|---|
| `.[].x`, `first(.[].x)`, `map(.x) \| .[]` | echo, 0 | echo, 0 | echo, 0 |
| `map(.x)` | 5 | 5 | 5 |
| `first(map(.x) \| .[])` | 5 | 5 | **echo, 0** |
| `limit(1; map(.x) \| .[])` | 5 | 5 | **echo, 0** |

So `first` moved at commit 1, `limit` at commit 2. `#965` is not involved: `baa26d72e` (which predates it) and `main` (which has it) behave identically, both 5.

But the resolution is not "restore validation." On this document, `.[].x`, `first(.[].x)` and the fused `map(.x) | .[]` **already echo the raw value at exit 0 on `main` and on `baa26d72e`** — that is succinctly's documented navigation behavior for a decode-failure value (`decode-failure-routing.md`: echo on navigation, validate on materialization). `map(.x)` alone materializes the array and raises, and still does. So `main` was internally inconsistent: it echoed `map(.x) | .[]` but raised `first(map(.x) | .[])` on the same document — the exact spelling-dependence #2168 removed for `path(.d)`. Making `first`/`limit` echo like the navigation they wrap is the #2168 answer (diverge uniformly, not by spelling); restoring validation would recreate the inconsistency. Recorded in `limitations.md` and pinned by `test_first_limit_echo_a_decode_failure_value_like_navigation_2103`, with `[{"x":"\ud800"}]` added to the sweep (now 600 cases).

Agreed it "reads data" and isn't the "reads nothing" case — but the governing rule for a read value isn't "reads nothing → skip"; it's navigate-echoes / materialize-validates, and `first(map(.x)|.[])` navigates without materializing (the array is fused away), exactly as `map(.x)|.[]` does.

## Finding 2 — documented and pinned (option 2 now, option 1 = #2662)

`-S`/`-a`/`-s`/`-C` and `-n 'input | …'` materialize the input up front, so they still validate whatever the filter; `sjq -S '1+1'` exits 5 where `sjq '1+1'` answers `2`. `-e` materializes outputs, so `-e '.,.'` raises where `-c '.,.'` echoes. Stated in `limitations.md`, pinned by `test_materializing_flag_routes_still_validate_2103`, and the cursor-based fix is filed as #2662. Also folded the `-e '.,.'` oddity into #2662 as you suggested.

## Nits

Fixed the three `OneCursorValue` "one construction site" comments (the empty-pipe tail re-pairs it) and padded the new table.

Full `jq_cli_tests`/`yq_cli_tests`/`--lib`, both clippy legs, rustdoc, and the sweep (600 cases, matches golden) all green on bba166a31.

newhoggy:
Good catch, and the ADR text bears it out. `adr-0018.md` says divergence is "permitted only under four named conditions" with "nothing else qualifies on its own", the #2416 amendment's step 3 says the performance tiebreak "cannot introduce a divergence; rule 4's closed list still governs", and neither #2168 nor #2103 is mentioned in the ADR. The ADR's own `*+d` precedent points the same way you do: an out-of-list divergence is recorded as "to be fixed or re-justified", not blessed. So "on its merits" was claiming a legitimacy the ADR didn't grant.

Formalized as an ADR-0018 amendment (`0e82f4277`), parallel to #2416's write-up. It names the one structural case rather than a general escape hatch: the reference **rejects the whole document at parse time**, and succinctly, a semi-index, deliberately does not validate up front. The faithful answer (reject) is not uniformly reachable — matching it more closely buys a rejection on the spellings that materialize and an echo on the ones that navigate (`.d` vs `path(.d)`; `map(.x)|.[]` vs `first(map(.x)|.[])`), which is the per-spelling incoherence #1629/#1642 exist to remove. So the real fork is "diverge uniformly or by spelling", which step 2 does not adjudicate, and the owner-level disposition is the uniform divergence.

Bounded deliberately: it reaches only a parse-time whole-document rejection, does not relax rule 4's closed list for ordinary single-behaviour forks, does not reopen the step-3 tiebreak, and does not reach `*+d` (a claim about untested reference behaviour on a document real yq *accepts*). #2168 and #2103 are listed as its instances, and both limitations entries now point at the amendment instead of resting on prose precedent.

(Also on the branch since your review: two concurrent coverage commits pinning the now-unreachable-via-CLI arms — the empty-pipe `None` tail and `generic_result_to_jq_values`'s arms — as tolerated/tested.)

PR #2684, opened 2026-09-10
fix(jq): a def from include/~/.jq can now shadow a builtin (#2395)

### LINKED ISSUE #2395: yq: jq: a def loaded via include/import/~/.jq cannot shadow a builtin (#2036 follow-up)
## What's wrong

Found during #2036's own PR review (#2394). #2036 fixed a user `def` shadowing a
builtin of the same name -- but only for a `def` written directly in the main filter
text. A `def` loaded via `include`, `import`, or `~/.jq` cannot shadow a builtin:

```console
$ cat mylib.jq
def length: "shadowed-from-module";
$ jq -L . -nc 'include "mylib"; length'
"shadowed-from-module"
$ succinctly jq -L . -nc 'include "mylib"; length'
0
```

Every other #2036 shadowing case now works correctly (matching real jq exactly); this
is specifically the module-sourced case.

## Root cause

#2036's fix collects shadow-candidate names via a lexical prescan
(`collect_def_names`, `src/jq/parser.rs`) that runs once, at `Parser` construction,
over the raw filter text being parsed. By construction, this only ever sees the *main
filter's own source text* -- module bodies (`include`/`import`/`~/.jq`) are loaded and
inlined *separately*, by `ModuleLoader::process_program`
(`src/bin/succinctly/jq_runner.rs`), well after the main filter has already been fully
parsed. A `def length: ...;` living only inside an included module's own text was
never scanned, so `length` in the main filter never gets wrapped with a
`builtin_fallback` at parse time at all -- it stays a plain `Expr::Builtin(Length)`
node, and `resolve.rs`'s own scope-aware rewrite (which correctly *would* find
`length` in scope, since the module's `FuncDef` does get inlined before that pass
runs) never gets a chance to run on it, since there's no `Expr::FuncCall` node there
to reconsider in the first place.

## Suggested next step

The prescan needs to also see module-sourced `def` names before the main filter is
parsed -- which requires either:
1. Loading modules *before* parsing the main filter (a significant reordering: right
   now `include`/`import` declarations are themselves parsed as part of the main
   filter's own grammar, so there's a chicken-and-egg problem -- you need to parse
   enough of the filter to know *which* modules to load, before you can prescan their
   contents).
2. A second, later shadow-check pass that runs *after* `ModuleLoader::process_program`
   has inlined every module `FuncDef`, re-examining the now-fully-assembled tree for
   any `Expr::Builtin`/special-form node whose name matches an inlined module def and
   retroactively wrapping/rewriting it -- essentially redoing #2036's own
   parse-time logic as a tree-level pass instead of a text-level one.

Option 2 is likely more tractable given the existing architecture (it doesn't require
reordering the parse/module-load pipeline), but needs its own investigation into
correctly locating every wrappable node in an already-built tree without the parser's
own positional/keyword-dispatch context to lean on.

## Scope note

This is a narrower, rarer case than #2036's own main scope (same-file shadowing,
which is the common jq idiom of patching one builtin's behavior for one script) --
module-sourced shadowing specifically requires `-L`/`include`/`import`/`~/.jq`, which
most jq/yq invocations never touch. Not a regression from #2036 (this case was
already broken identically before that fix, since the parser had NO shadow-check
mechanism at all); #2036 just didn't happen to close it either.

newhoggy:
Fixes #2395.

## The bug

#2036 taught the parser that a user `def` can shadow a builtin of the same name, but
only for a `def` written in the main filter's own text. One arriving through a module
could not:

```console
$ cat mylib.jq
def length: "shadowed-from-module";
$ jq            -L . -nc 'include "mylib"; length'   # "shadowed-from-module"
$ succinctly jq -L . -nc 'include "mylib"; length'   # 0     <- wrong
```

Every kind of shadowable name was affected — ordinary builtins (`length`, `keys`, `type`,
`map`, `empty`, …), the `not` keyword, the fixed-arity special forms (`error`, `select`,
`first`, `limit`, `path`, `getpath`), and `range`'s one-argument sugar. `~/.jq`'s own defs
were affected identically. Not a regression from #2036: this case was already broken the
same way before it, since the parser had no shadow-check mechanism at all.

## Root cause

The shadow-candidate set is a lexical scan of **the text being parsed** —
`collect_def_names(input)`, stored as `Parser::shadowable_defs`. An included module's body
is a *separate* text, loaded and inlined by `ModuleLoader::process_program` only after the
main filter has been fully parsed. So `length` lowered straight to `Expr::Builtin(Length)`,
and `resolve.rs`'s scope-aware check — which by then would have found the module's
`FuncDef` plainly in scope — had no `Expr::FuncCall` node to reconsider.

## The fix

The issue suggested a tree-level pass over the assembled AST. I took a third option
instead: feed the module-sourced names back into a **second parse**.

- `ModuleLoader::unqualified_def_names` reports the def names that will land *unqualified*
  in scope (`~/.jq`'s defs plus every `include`d module's).
- `run_jq` re-parses the filter through the new
  `jq::parse_program_with_extra_shadowable_defs`, which unions those into the candidate set.

This reuses #2036's tested machinery verbatim rather than reimplementing its decisions at
tree level. The tree pass was rejected for two concrete reasons found while investigating:
it would need a 207-arm `Builtin -> name` map that does not exist anywhere in the crate
(`try_parse_builtin` is a one-way name → `Builtin` keyword chain), and it could not recover
`range`'s arity, because the `Expr::Paren` sugar marker that distinguishes `range(5)` from
`range(0;5)` is itself only emitted for a name already known to be a candidate.

**Cost.** `load_module` memoizes on the module path, so nothing is read or parsed twice. A
filter with no `include` and no `~/.jq` contributes no names and skips the second parse
entirely — the overwhelmingly common case pays nothing.

**Why the second parse cannot fail where the first succeeded.** `SHADOW_RETRY_BUDGET` is
charged only where a *dedicated* parse fails, and at such a site a non-candidate name
propagates its error outright — so the first parse would have failed there too. At a name
the widening newly covers, the dedicated parse still runs first and, on success, is merely
wrapped; no token is consumed differently and `pos` does not move. The charged sites are
therefore identical in both parses. The second commit records that argument in the code and
has the arm fall back to the already-parsed program rather than erroring: if the reasoning
is ever wrong, the cost is that filter's module-sourced shadowing (the behaviour before this
PR) instead of rejecting a filter that compiles today and that real jq accepts.

`import "m" as ns` is deliberately excluded, matching jq: those defs enter scope only as
`ns::name`, so a bare `length` there is still the builtin. `succinctly yq` is untouched — it
has no `-L` and never builds a module loader.

## Verification

Every expected value captured live from the pinned oracle (`jq-1.7.1-apple`), never written
from memory. All 19 oracle comparisons match, including the cases that must **not** change:

| case | jq | before | after |
|---|---|---|---|
| `include "m"; length` (+ 12 more builtins/special forms) | shadowed | builtin ✗ | shadowed ✓ |
| `~/.jq` defines `length` | shadowed | builtin ✗ | shadowed ✓ |
| `import "m" as m; length` | builtin | builtin ✓ | builtin ✓ |
| `import "m" as m; m::length` | shadowed | shadowed ✓ | shadowed ✓ |
| module defines only `length/1`, bare `length` | builtin | builtin ✓ | builtin ✓ |
| module defines only `limit/3`, `limit(1;1,2)` | builtin | builtin ✓ | builtin ✓ |
| transitive `include` (main → outer → inner) | does not leak | ✓ | ✓ |
| main-file `def` beats module `def` | main's | ✓ | ✓ |

7 new tests in `tests/jq_cli_tests.rs`, following the existing #2036 cluster's table-driven
shape. Reverting `src/` to the pre-fix commit fails exactly the 3 behavioural tests and
passes the 4 regression guards, so the tests demonstrably detect the bug rather than merely
describing it. The deep-nesting test is the module-sourced twin of
`test_def_shadow_deep_nesting_does_not_blow_up_2036`, guarding the new parse against the
`O(2^depth)` blowup #2036's own review found (60 levels parse in 0.03s).

Gates run locally, derived from `ci.yml`: `cargo test --features cli` (all suites),
`cargo test --no-default-features` (no_std), all three clippy legs with `-D warnings`,
`cargo fmt --all --check`, and `cargo doc --no-deps --all-features` with
`RUSTDOCFLAGS=-D warnings`. Patch coverage 100%.

## Follow-ups filed, not fixed here

- issue 2682 — `~/.jq` outranks an `include`d module of the same name; real jq gives `include`
  the higher priority. This is why the tests here deliberately avoid combining both sources
  on one name.
- issue 2683 — `succinctly yq` parses `include`/`import` and silently drops them; real yq
  rejects them at the lexer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,44 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A `def` loaded through `include` or `~/.jq` can now shadow a builtin of the
+  same name** (#2395, a #2036 follow-up). `jq -L . -nc 'include "mylib"; length'`
+  with `def length: "shadowed-from-module";` in the module answered `0` -- the
+  builtin -- where real jq answers `"shadowed-from-module"`. Every kind of
+  shadowable name was affected: ordinary builtins (`length`, `keys`, `type`,
+  `map`, `empty`, ...), the `not` keyword, the fixed-arity special forms
+  (`error`, `select`, `first`, `limit`, `path`, `getpath`), and `range`'s
+  one-argument sugar. `~/.jq`'s own defs were affected identically.
+
+  #2036 collects shadow-candidate names by scanning *the text being parsed*, and
+  the main filter's text can never contain an included module's `def`s: module
+  bodies are separate texts, loaded and inlined by `ModuleLoader::process_program`
+  only after that parse has finished. So `length` lowered straight to a builtin
+  node, and `resolve.rs`'s scope-aware check -- which by then would have found
+  the module's `FuncDef` plainly in scope -- had no call site to reconsider.
+
+  Fixed by feeding the names back into a second parse rather than rewriting the
+  assembled tree, which is the issue's own suggested direction but needs a
+  `Builtin` -> name map the crate does not have, and could not recover `range`'s
+  arity, whose sugar marker is itself only emitted for a name already known to be
+  a candidate. The loader now reports the def names that will land *unqualified*
+  in scope (`ModuleLoader::unqualified_def_names`), and the runner re-parses the
+  filter with those added via the new
+  `jq::parse_program_with_extra_shadowable_defs`. Widening the candidate set is
+  monotone -- it can only add a fallback wrapper to a parse that already
+  succeeded, or a retry to one that already failed -- so the second parse cannot
+  reject a program the first accepted. `load_module` memoizes, so nothing is read
+  or parsed twice; a filter with no `include` and no `~/.jq` skips the second
+  parse entirely and pays nothing.
+
+  `import "m" as ns` is deliberately excluded, matching jq: those defs enter
+  scope only as `ns::name`, so a bare `length` there is still the builtin
+  (`m::length` reaches the module's, and already did). Arity sensitivity carries
+  over from #2036 unchanged -- a module defining only `length/1` leaves the
+  zero-arity builtin intact -- as do the two scope boundaries: a `def` in the
+  main filter still beats a module's, and `include` is still not transitive.
+  `succinctly yq` is unaffected; it has no `-L` and never builds a module loader.
+
 - **An `as`-bound variable now stands at the node it was bound from, instead of
   being a value with no position** (#2072). Real yq's variables hold *nodes*,
   parent pointers and all, so `.a.b as $x | $x | key` is `"b"` there;
--- a/src/bin/succinctly/jq_runner.rs
+++ b/src/bin/succinctly/jq_runner.rs
@@ -57,6 +57,31 @@ pub struct ModuleLoader {
     auto_loaded_defs: Vec<(String, Vec<Param>, Expr)>,
 }
 
+/// Resolve a module path to a file path within `search_path`.
+///
+/// A free function rather than a `ModuleLoader` method (#2395): its only
+/// caller, [`ModuleLoader::ensure_module_loaded`], holds a mutable borrow of
+/// the module cache across the call, which an `&self` method could not
+/// coexist with. It reads nothing but the search path either way.
+fn resolve_module_in(search_path: &[PathBuf], module_path: &str) -> Option<PathBuf> {
+    // Add .jq extension if not present
+    let module_file = if module_path.ends_with(".jq") {
+        module_path.to_string()
+    } else {
+        format!("{module_path}.jq")
+    };
+
+    // Search in each path
+    for base in search_path {
+        let full_path = base.join(&module_file);
+        if full_path.is_file() {
+            return Some(full_path);
+        }
+    }
+
+    None
+}
+
 impl ModuleLoader {
     /// Create a new module loader with the given search paths.
     pub fn new(library_paths: &[PathBuf]) -> Self {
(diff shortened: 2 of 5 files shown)


newhoggy:
## Independent review — approve

Re-derived the correctness argument from the source rather than the PR body, re-verified the
behaviour live against the pinned `/usr/bin/jq` 1.7.1, and re-ran the gates locally.

### The "second parse cannot fail where the first succeeded" argument holds

Checked all four sites that read `shadowable_defs` — `parse_primary_inner`'s builtin
dispatch, `range`'s 1-arg sugar gate, `zero_arity_or_wrong_arity_call`'s tail, and
`parse_shadowable_special_form`. Every one consults the set **after** the keyword's
dedicated parse has already returned, never before, so widening it cannot change which
tokens are consumed or where `pos` lands. In particular `zero_arity_or_wrong_arity_call`'s
`has_arg && mode == Jq` rewind fires independently of `shadow_name`.

The remaining hole in that argument would be an `Err`-swallowing backtrack somewhere else in
the parser, which would let a *newly* covered failing site exist inside a first parse that
still succeeded. There is none: `grep` for `Err(_) =>` in `parser.rs` is empty, and all four
`self.pos = <start>` rewinds are either `Ok`-returning (`rewind_to_wrong_arity_call`,
`expect_or_none`, `zero_arity_or_wrong_arity_call`) or
`retry_shadow_candidate_as_generic_call` itself. So `SHADOW_RETRY_BUDGET` really is charged
at the identical set of sites in both parses, and the `tolerate-line` reason on
`.unwrap_or(program)` is accurate. Keeping the arm anyway is the right call.

### Live parity, beyond the new test table

All matching against jq 1.7.1:

- object keys (`{length: 1}`, `{length}`), `.length` field access, `$ENV`, `reduce`/`foreach`
  over a shadowed `range`, `path(first(.a,.b))`, slice-then-`length`, `ltrimstr`, `splits`,
  `..|numbers`, `try (1/0) catch`
- `import`/`m::` both directions, non-transitive `include`, main-filter `def` beating the
  module's, a module def that calls the builtin it shadows (`def len2: length + 1;`)
- **error paths byte-for-byte**, including the echoed source line and its trailing padding:
  `length/2 is not defined`, `nosuch/0 is not defined`, `error/2`, `first/2`, exit 3. Worth
  calling out because `report_unresolved_calls` builds its position table from its own
  `collect_call_sites` parse, which does *not* receive the extra defs — the arity-mismatch
  case (`include "m6"; length(1;2)` where the module defines only `length/1`) was the one I
  expected to drift, and it does not.

### Cost

55 KB filter, `~/.jq` present vs absent, three reps, M-series release build:
27.5/28.7/27.6 ms -> 29.6/30.7/29.9 ms. The second parse and nothing else.

### Gates

`cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`,
`cargo test --features cli --test jq_cli_tests` (1544 passed, incl. all 7 new `_2395`
tests), `cargo test --doc` on the new `parse_program_with_extra_shadowable_defs` example —
all green, matching CI.

### Nits (non-blocking)

1. **`unqualified_def_names` clones every module's full AST to read names only.**
   `load_module` returns `defs.clone()` — name, params *and* body `Expr` — and this discards
   everything but the name. The doc comment's "costs nothing: `load_module` memoizes" is
   right about re-reading and re-parsing but does add one full clone of every included
   module's tree. Tiny in practice; a name-only accessor or returning `&[…]` would avoid it.

2. **Mode/extension flags are duplicated rather than shared.** The first parse is
   `jq::parse_program(&filter_str)` (resolving to `Jq`/`false`); the re-parse hard-codes
   `ParserMode::Jq, false`. They agree today, but if `parse_program`'s defaults ever move
   the two parses diverge silently and the *second* one wins. Worth routing both through one
   expression.

3. The second commit is a refinement of the first and could be fixup'd. Cosmetic.

### Follow-ups filed while verifying (none are regressions from this PR)

- issue 2686 — nested wrong-arity `until`/`limit` parses in **O(2^depth)**: 18 s at depth 24
  where jq takes 20 ms; > 25 s at depth 30. Pinned to the parser (appending a lexical error
  so the program can never reach name resolution keeps the cost). Reproduces with no `def`
  and no `include` anywhere, so this PR's second parse is never entered — it is the
  #2110/#2237 wrong-arity rewind path, a third instance of the DoS shape #2036 closed twice
  elsewhere. Notably `first/2`, `range/4` and generic calls stay flat; only the two-arity
  forms blow up.
- issue 2687 — a `def error:` does not intercept `break $label`, though jq's parser desugars
  `break` into `gen_call("error", …)` at arity 0. jq emits the def's value as an extra
  output and keeps evaluating past the `break`; we do not. Same-file, so #2036-era.
- issue 2688 — `$__loc__` reports `{"file":"<stdin>"}` where jq reports `"<top-level>"`
  (inline and `-f` alike). The compile-error path in this same binary already gets
  `<top-level>` right, so the two disagree internally.
- Commented on issue 2682 confirming its own open question: multiple `include`s of the same
  name are inverted the same way as the `~/.jq` case it already covers — jq gives the *last*
  include priority, we give the first, order-sensitive in both tools but opposite.

newhoggy:
Review nits addressed — pushed `4e583e96d`.

- **Nit 1** (deep clone) was already handled by `3d4a905f3`. Its one residual is fixed here:
  `unqualified_def_names`'s doc still credited `load_module` with the memoization, which is
  `ensure_module_loaded` since that refactor.
- **Nit 2** (mode/extension duplication) fixed: both parses of the main filter now go
  through a single `parse_filter` closure, so `ParserMode::Jq` / `jq_extensions` lives in one
  place. No behaviour change — the closure reproduces the defaults `jq::parse_program` was
  already resolving to. The risk it removes is that the *second* parse's result is the one
  that runs, so a drift between the two would have been silent rather than a conflict.
- **Nit 3** (squash the fixup commit) deliberately not done: with four individually
  meaningful commits on the branch now, the second no longer reads as a stray fixup, and
  rewriting history would force-push and re-run every check for no substance.

Re-verified after the change: `cargo fmt --check`, `cargo clippy --all-targets
--all-features -- -D warnings`, `cargo test --features cli --test jq_cli_tests` (1547
passed), release build, and 25 oracle probes against jq 1.7.1.

One probe differs, pre-existing and out of scope here — the missing-module *error wording*.
Chasing it turned up that three of the four jq compile-error paths diverge from jq's report
while the name/arity-resolution path matches byte-for-byte, so the runner disagrees with
itself. Filed as issue 2703, which also notes that
`test_module_not_found_still_reports_module_error_2395` pins today's wording and will need
updating alongside whatever lands there.

Issue #2680, opened 2026-09-09
jq: path()'s DOM walker drops an output prefix jq keeps when a nested pipe stage errors


newhoggy:
**Severity: Low** (fidelity, not correctness or performance — found by fuzzing during #2190)

## Summary

`walk_path`/`walk_pipe` (`src/jq/eval.rs`, the DOM route `path()` takes for any
non-cursor-navigable expression) **discards an output prefix that jq keeps**, when a
*nested* pipe in stage-1 position produces some outputs and then errors.

`eval_generic.rs`'s own twin of this function, `path_walk_pipe_generic`, already gets
this right, and says so in a comment — so `eval.rs`'s `walk_pipe` is the outlier against
its sibling evaluator, not against some external convention.

## Reproducing

```console
$ echo '{"a":[{"b":{}},5]}' | jq -c 'path((.a[] | .b) | .c[0:1])'
jq: error (at <stdin>:1): Cannot index number with string "b"
["a",0,"b","c",{"start":0,"end":1}]          # <- emitted, exit 5

$ echo '{"a":[{"b":{}},5]}' | sjq -c 'path((.a[] | .b) | .c[0:1])'
jq: error (at <stdin>:1): Cannot index number with string "b"
                                             # <- nothing, exit 5
```

The `[0:1]` is only there to force the DOM route (`path_expr_is_cursor_navigable` refuses
a `Slice`); the bug is in the walker, not the slice.

**The parentheses are the whole difference.** The same pipe written flat is already
correct — succinctly and jq agree, prefix included:

```console
$ echo '{"a":[{"b":{}},5]}' | sjq -c 'path(.a[] | .b | .c[0:1])'
jq: error (at <stdin>:1): Cannot index number with string "b"
["a",0,"b","c",{"start":0,"end":1}]          # <- correct
```

Both captured live against jq 1.7.1 (`/usr/bin/jq`) and `succinctly` at d1638aa3d.

## Mechanism

`walk_pipe` buffers stage 1 in full before running stage 2:

```rust
let mut reached = Vec::new();
walk_path::<S>(first, value, current_path, &mut reached, optional)?;   // <-- this `?`
for (path, val) in reached {
    walk_pipe::<S>(rest, val, &path, out, optional)?;
}
```

The `?` on the first call throws `reached` away. For a *flat* pipe that is harmless —
every level below the first writes into the top-level `out` directly, so the prefix is
[...]

PR #2685, opened 2026-09-10
perf(jq): borrow the document in path()'s walk instead of cloning it per branch

### LINKED ISSUE #2190: perf(jq): path(f) on the eval.rs DOM route is quadratic in resolved-path count, independent of document size
**Severity: Medium** (performance, not correctness — found while benchmarking #1510)

## Summary

`path(f)` evaluated through `succinctly::jq::eval` (the `eval.rs` DOM route, i.e.
`eval_single` → `resolve_node`, **not** `eval_generic`'s cursor walk) is **quadratic in
the number of resolved paths** once a comma is present in the path expression.

The quadratic term is the resolved-path count alone. It is *not* the comma's branch
count, and *not* the document size — both were held fixed independently and neither
moves the time.

## Measurements

Apple M4 Pro (`johns-mac-mini`), idle (91% idle, no other builds), release,
best-of-3 per row, timing only `eval` with `parse`/`JsonIndex::build` hoisted out.
Query shape: `[path(.foo[] | (.b0,...,.b<w-1>) | .d | .d | .d | .d)]`.

**A. Resolved paths held at 1600; comma width varies 1 → 32** (document also held at
~48-51 KB):

| width | elems | paths | doc   | time    |
|-------|-------|-------|-------|---------|
| 1     | 1600  | 1600  | 51 KB | 0.0103s |
| 4     | 400   | 1600  | 49 KB | 0.7435s |
| 8     | 200   | 1600  | 48 KB | 0.7289s |
| 32    | 50    | 1600  | 49 KB | 0.7369s |

Widening the comma 8x costs nothing (0.744s → 0.737s). The 72x jump is between
*no comma* (`width=1` parses as `Paren`, not `Comma`) and *any* comma.

**B. Comma width held at 8; element count varies:**

| elems | paths | doc   | time    | ratio |
|-------|-------|-------|---------|-------|
| 50    | 400   | 12 KB | 0.0474s | –     |
| 100   | 800   | 24 KB | 0.1833s | 3.86x |
| 200   | 1600  | 48 KB | 0.7337s | 4.00x |
| 400   | 3200  | 97 KB | 2.9033s | 3.96x |

Each doubling costs ~3.94x — exponent ≈ **1.98**.

**C. Resolved paths held at 1600; document size varies** (padding each element with an
unreferenced `"pad"` string, so bytes grow while the path set does not):

| pad | paths | doc    | time    | ratio |
|-----|-------|--------|---------|-------|
| 0   | 1600  | 48 KB  | 0.7324s | –     |
| 120 | 1600  | 74 KB  | 0.7441s | 1.02x |
| 360 | 1600  | 120 KB | 0.7472s | 1.00x |
| 840 | 1600  | 214 KB | 0.7517s | 1.01x |

4.5x the document, +2.6% time. Document size is not a factor.

## Why this is not #2048

#2048 attributes an `O(d^2)` to `resolve_node` re-serializing **each resolved branch's
value** through `to_json_for_reindex_at_depth`. That mechanism is proportional to the
bytes each branch holds, so it would have to show up in table C — and table C is flat.
Whatever is quadratic here is proportional to the *count* of resolved paths and
indifferent to their content, which points at something that touches the accumulated
path set once per newly resolved path (a linear scan, containment check, or dedup)
rather than at re-serialization. Filed separately for that reason; if profiling shows a
shared root cause after all, close as a duplicate.

Note also #1301, which concluded the resolved-path count was *not* its term. This is the
opposite finding on a different shape, so the two are worth reading together rather than
assuming either generalizes.

## Reproducing

The tables above came from a throwaway `examples/` binary using
`succinctly::jq::{eval, parse}` + `succinctly::json::JsonIndex` directly, so the DOM
route is exercised without process-spawn noise. Note that the **CLI does not show this**:
`succinctly jq` routes through `eval_generic`'s cursor walk, which does the 3200-path
case in ~10 ms rather than ~2.9 s. This affects library callers of `succinctly::jq::eval`
and the `benches/jq_write_path_bench.rs` groups that use it.

newhoggy:
Closes #2190.

## What was wrong

`builtin_path_on_owned` handed **every resolved branch its own deep clone of the whole
document**, because `walk_path`/`walk_pipe`/`step_into` navigated by move:

```rust
for expr in &exprs {                       // one Expr per resolved branch
    walk_path::<S>(expr, owned.clone(), &root, &mut reached, optional)
}
```

Cost was `branches × nodes`, and for #2190's shape both factors grow with the element
count together — hence the measured exponent of ~1.98. That model is also what makes the
issue's table C (pad the bytes, hold the path set fixed) read flat: clone cost tracks
*node* count, not byte count.

## What changed

**The walkers borrow.** A `WalkNode<'v>` is either a node of the document or a value the
walk itself built — a slice's array, or the fallback evaluator's output — with the latter
`Rc`-shared.

The reason for that shape is worth stating, because it is the non-obvious part: the type
has to be **closed under navigation** (a child of a `WalkNode<'v>` is a `WalkNode<'v>`),
which is what lets `walk_pipe` go on buffering a stage's results in a plain `Vec` with no
restructuring and no semantic change. `Cow<'v, OwnedValue>` cannot do that — `Vec` is
invariant in its element type, so a stage reached *through* a `Cow::Owned` would owe its
caller a borrow of a strictly shorter local buffer. `Rc` breaks the chain: residue's
children are cloned rather than borrowed, exactly as `eval_generic::owned_nav_children`
already does. It is compiler-enforced — if `WalkNode::pick` ever stopped returning the
document's own lifetime, the buffer would stop typechecking.

**The indexing rules got factored, not duplicated.** #2190 recorded this fix as blocked
on needing a by-reference twin of `navigate_static_component`, i.e. a fourth copy of jq's
indexing rules. But only its `Field`/`Index` arms are destructive at all — the `_`
fallback already borrowed — so `classify_static_component` now resolves a component to a
`StaticAccess`, and the destructive and borrowing navigators differ only in how they take
the child out. Which container each component may index, the step-through-`null` rule,
negative-index resolution, the bounds check, and which `EvalError` a non-indexable
container raises all have exactly one definition.
`navigate_static_component_ref_agrees_with_navigate_static_component` pins the pair.

`value_after_components`, and with it the whole `=`/`|=`/`del()` route, keeps its
move-based navigation — it genuinely needs an owned result.

## Results

Interleaved A/B per `docs/guides/benchmarking.md`, baseline = merge-base with `main`
(`31ab1d3ea`), same benchmark harness against both library versions, idle Apple M4 Pro,
`jq_write_path_path_leading_iterate_comma` (new group, in-process so no process-spawn
noise):

| elems | paths | base | head | speedup | base ratio | head ratio |
|------:|------:|---------:|--------:|--------:|-----------:|-----------:|
|    50 |   400 |   48.0 ms | 1.369 ms |   35.1x |          – |          – |
|   100 |   800 |  190.3 ms | 2.750 ms |   69.2x |      3.97x |      2.01x |
|   200 |  1600 |  755.3 ms | 5.530 ms |  136.6x |      3.97x |      2.01x |
|   400 |  3200 | 3017.8 ms | 11.15 ms |  270.7x |      4.00x |      2.02x |

**Exponent 1.99 → 1.01.** The scaling curve is the claim, not any single ratio. The
group sets `Throughput::Elements`, so it also reads directly off Criterion's own column:
base's throughput *halves* at every doubling (8.0 → 4.1 → 2.1 → 1.05 Kelem/s) where head
is flat at ~287 Kelem/s across the whole 16x range.

`jq_write_path_two_branch_path_depth` — two branches, no fan-out — improves 15-18% at
every depth, since even that shape no longer clones the document twice.

## Output identity

Gated before any timing was believed. A three-way differential (pre-change binary,
post-change binary, jq 1.7.1) over **7,700 `path()` shapes** — 200 hand-built plus 2,500
generated on each of three seeds — reports **0 pre-vs-post differences** in stdout,
stderr and exit code. The generator alphabet covers present/missing/computed fields,
in-bounds, negative and out-of-range indices, four slice spellings, bare iterates,
`?` suffixes, parenthesised comma groups and nested pipes, over random trees including
`null`, empty containers and heterogeneous arrays; half the expressions carry a forced
non-cursor-navigable component so they route through this walker rather than the cursor
walk.

The same sweep's jq leg reported 2 divergences, in both binaries equally — a pre-existing
prefix-discard bug, now filed as #2680.

## #2190's impact statement needs correcting

The issue says the CLI does not show this. That holds only for *cursor-navigable* path
expressions. `path_expr_is_cursor_navigable` refuses a `Slice`, so

```
sjq '[path(.foo[] | (.b0,…,.b7) | .d|.d|.d|.d | .[0:1])] | length'
```

reaches this route from the CLI and read the same ~1.95 exponent, up to 150x slower at
400 elements. Comment left on the issue.

## Checks

`cargo test --features cli` (60 binaries) · `--features cli,simd,regex,serde --workspace`
· `--no-default-features` (no_std) · `--features simd,scalar-yaml` · both `ci.yml` clippy
legs with `-D warnings` · `cargo doc` with `RUSTDOCFLAGS=-D warnings` · `cargo fmt --check`.

### A note on the middle commit

The first version of the refactor resolved an object key to an `IndexMap` *slot* in the
classifier, which made the destructive side hash twice per `Field` step (`get_index_of`,
then `swap_remove_index` re-finding the bucket) where it used to hash once. That showed
up as a systematic +1.7-3.3% across all eight `two_branch_{del,assign}_depth` points — a
cost charged to `=`/`|=`/`del()` for a `path()` fix. `StaticAccess::Field(&'e str)`
carries the key verbatim instead; the write-side deltas now straddle zero (−0.7% to
+1.0%, four of eight negative). Caught only because the neighbouring groups were
re-measured rather than assumed.

### DIFF
--- a/benches/jq_recurse_clone_bench.rs
+++ b/benches/jq_recurse_clone_bench.rs
@@ -16,13 +16,23 @@
 //! turns out to be a poor isolator: `builtin_path` calls
 //! `resolve_dynamic_indexes` (which drives `push_recursive_branches`) and
 //! *then* re-walks every one of the `depth + 1` resolved paths from the
-//! document root via `walk_path`/`step_into`, each step of which clones the
-//! value reached so far. That second pass was measured (`temp_probe`-style,
-//! not checked in) to cost **~250x** `push_recursive_branches`'s own share at
-//! depth 400 (1.9s vs 7-8ms) and scales worse than quadratically — so a
-//! `path(..)` benchmark would be dominated by `walk_path`, not by the
-//! function #668 targets, near-invisible to that fix, and would repeat
-//! exactly the mistake #675 was filed to correct in the first place.
+//! document root via `walk_path`/`step_into`. That second pass was measured
+//! (`temp_probe`-style, not checked in) to cost **~250x**
+//! `push_recursive_branches`'s own share at depth 400 (1.9s vs 7-8ms) and to
+//! scale worse than quadratically — so a `path(..)` benchmark would be
+//! dominated by `walk_path`, not by the function #668 targets, near-invisible
+//! to that fix, and would repeat exactly the mistake #675 was filed to
+//! correct in the first place.
+//!
+//! **Two corrections to the mechanism named above**, neither of which changes
+//! the conclusion (`del(..)`, not `path(..)`, is the isolator). The clone
+//! this paragraph blamed on "each step" of `walk_path`/`step_into` stopped
+//! being per-step at #2058, which made a single branch's walk O(depth) by
+//! navigating it destructively; what remained was one deep clone of the whole
+//! document per *resolved branch*, in `builtin_path_on_owned`. #2190 removed
+//! that one too — the walkers borrow the document now — so `path(..)`'s
+//! second pass is no longer quadratic in the branch count either. The figures
+//! above date from before both and should be read as history.
 //! `del(..)` instead reaches `resolve_del_path_branches` and then applies
 //! `delete_at_path`/`DeleteTrie` directly to the already-resolved static
 //! paths — no second walk of the original tree — so its cost tracks
--- a/benches/jq_recurse_clone_bench.rs
+++ b/benches/jq_recurse_clone_bench.rs
@@ -1,12 +1,12 @@
-//! Depth-scaling benchmark for `push_recursive_branches`/`resolve_recurse`'s
+//! Depth-scaling benchmark for `push_recursive_branches`/`resolve_recurse_sink`'s
 //! per-node clone cost (#668), split out of #675.
 //!
 //! `jq_recurse_depth_bench`'s query, `.. | .[.k]?`, does **not** reach
-//! `push_recursive_branches`/`resolve_recurse` (`src/jq/eval.rs`): bare `..`
+//! `push_recursive_branches`/`resolve_recurse_sink` (`src/jq/eval.rs`): bare `..`
 //! in value position dispatches to `eval_recursive_descent`/
 //! `collect_recursive`, which clones a cheap `StandardJson` cursor per node,
 //! not the materialized `OwnedValue` — see that benchmark's own corrected
-//! docstring. `push_recursive_branches`/`resolve_recurse` are reachable only
+//! docstring. `push_recursive_branches`/`resolve_recurse_sink` are reachable only
 //! through `resolve_node`, which `resolve_dynamic_indexes` calls solely when
 //! `needs_path_prepass` is true — true for `..`/`recurse` themselves, so only
 //! a *path-context* use (`path(..)`, `path(recurse(...))`, `=`, `|=`,
--- a/benches/jq_recurse_depth_bench.rs
+++ b/benches/jq_recurse_depth_bench.rs
@@ -11,12 +11,12 @@
 //!
 //! **This benchmark does not cover #668.** It was originally written
 //! attributing the O(depth^2) cost to `push_recursive_branches`/
-//! `resolve_recurse` (the `..`/`recurse` fan-out itself) rather than to
+//! `resolve_recurse_sink` (the `..`/`recurse` fan-out itself) rather than to
 //! `eval_index_expr`'s computed-key handling — plausible since the query
 //! below combines both, but wrong: PR #670 fixed `eval_index_expr` only, and
 //! that alone reproduced the full speedup, so the fan-out was never the
 //! bottleneck this benchmark measures. `push_recursive_branches`/
-//! `resolve_recurse` do still clone the value at every visited node
+//! `resolve_recurse_sink` do still clone the value at every visited node
 //! (`eval.rs:8595`, `8659`, `8664`), independently of this benchmark's
 //! query — that's #668, and #675 tracks adding a benchmark that isolates it
 //! (a query with no computed index, so it can't route through
--- a/benches/jq_write_path_bench.rs
+++ b/benches/jq_write_path_bench.rs
@@ -1189,6 +1189,100 @@ fn bench_path_recursive_def_ast(c: &mut Criterion) {
     group.finish();
 }
 
+// =============================================================================
+// #2190: `path()`'s per-branch cost, with the document held structurally fixed
+// =============================================================================
+
+/// Element counts for the #2190 group. Deliberately small and doubling: the
+/// point is the *exponent* between adjacent rows, not any single time, and at
+/// the pre-fix ~1.98 the largest row already costs ~3s.
+const LEADING_ITERATE_ELEMS: &[usize] = &[50, 100, 200, 400];
+
+/// How many comma branches each element fans out into. #2190's table A holds
+/// the resolved-path count fixed and widens this from 1 to 32 for free, so the
+/// width itself is not a term -- 8 is simply a middle value that keeps the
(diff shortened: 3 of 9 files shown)


newhoggy:
## x86_64 leg (the cross-architecture check the PR body was missing)

Measured on `terminus`, AMD Ryzen 9 7950X, idle (load 0.00 before each run), interleaved
within each repetition, same benchmark harness against both library versions. Baseline is
the merge-base with `main` (`ab350e9f4`); "head" is this PR's own four commits
(`7ee4aef5c`), before the cleanup commit on top, so the attribution is to #2190's change
alone.

### `jq_write_path_path_leading_iterate_comma`

| elems | paths | base | head | speedup | base ratio | head ratio |
|------:|------:|----------:|---------:|--------:|-----------:|-----------:|
|    50 |   400 |   79.97 ms | 1.760 ms |   45.4x |          – |          – |
|   100 |   800 |  365.35 ms | 3.566 ms |  102.5x |      4.57x |      2.03x |
|   200 |  1600 | 1542.40 ms | 7.049 ms |  218.8x |      4.22x |      1.98x |
|   400 |  3200 | 6836.40 ms | 14.38 ms |  475.6x |      4.43x |      2.04x |

**Base exponent 2.14 → head 1.01.**

Both architectures agree on the shape, and the concern that motivated measuring at all —
this repo's rule that memory-bound effects do not port across architectures — resolves in
the fix's favour rather than against it. The x86 win is *larger* (475.6x vs ARM's 270.7x
at 400 elements), and the pre-fix exponent is *worse* on Zen 4 (2.14 vs 1.99), which is
the same direction #106 recorded (6.1x on M4 Pro, 16.4x on Zen 4) for a fix of this kind.

| | ARM64 (M4 Pro) | x86_64 (7950X) |
|---|---:|---:|
| base exponent | 1.99 | 2.14 |
| head exponent | 1.01 | 1.01 |
| speedup @ 3200 paths | 270.7x | 475.6x |

### Write-side neighbours — the `=`/`|=`/`del()` route

Shares `classify_static_component` but not the walker, so it should be neutral:

| group | ARM64 | x86_64 |
|---|---|---|
| `two_branch_path_depth` | −15.4% to −18.5% | −16.8% to −19.8% |
| `two_branch_del_depth` | −0.3% to +1.0% | −0.1% to +1.0% |
| `two_branch_assign_depth` | −0.7% to +1.0% | −1.1% to +0.8% |

`two_branch_path_depth` improving by ~17% on both is expected — even a two-branch
`path()` no longer clones the document twice.

One methodological note, since it nearly went the other way. At 2 reps / 60 samples the
x86 `assign` group read +1.6-1.8% on three of four points, which looks exactly like the
systematic +1.7-3.3% regression the third commit here exists to fix. Re-run at 4 reps /
100 samples it collapsed to +0.0% / −1.1% / +0.8% / +0.1%. The distinguishing signal is
that the *real* regression grew when samples were added, while this one vanished — worth
recording because "all points positive" on a thin run is not by itself evidence.

### Output identity on x86

7,500 generated `path()` shapes (3 seeds x 2,500), pre-change binary vs post-change
binary: **0 differences** in stdout, stderr and exit code — matching the ARM result.

The jq-oracle leg is deliberately *not* run on this host: it has jq 1.6, and this repo
pins 1.7.1. The oracle comparison in the PR body stands on the ARM measurement, where
`/usr/bin/jq` is the pinned version.

Issue #2555, opened 2026-09-06
jq: substitute_func_param_impl's own outer-bare-param shadow check needs a second subst_bare flag, not just dollar_params


newhoggy:
## What's wrong

#2283 fixed `substitute_var_impl`'s `Expr::FuncDef` arm (an outer `$var` reaching into a nested `def`'s own body) by adding `dollar_params: Box<[bool]>` to `Expr::FuncDef`, recording which parameters were written `$`-style. `substitute_func_param_impl`'s own `Expr::FuncDef` arm has a *different*, deeper, related gap that `dollar_params` alone cannot fix.

`substitute_func_param_impl` substitutes an outer function parameter's argument in one pass that handles two substitutions simultaneously:
1. Bare `param` references (`Expr::FuncCall{name, args: []}`) — substituted unconditionally whenever the name matches.
2. `$param` references (`Expr::Var`) — substituted only while `subst_dollar` stays `true`.

Its `body_shadowed = params.contains(&param.to_string())` check (in the `Expr::FuncDef` arm) blocks *both* substitutions into a nested def's body whenever that nested def has a parameter of the same bare name — correct for blocking the bare substitution (confirmed live: a `$`-style parameter also creates its own bare binding, since `def h($param): param` alone is a compile error, "param is not defined", so `$`-style always occupies the bare namespace too), but *incorrect* for blocking the `$param` substitution when the nested parameter is bare-only (a bare parameter creates no `$`-binding of its own, so an outer `$param` should still reach through it).

Confirmed live against jq 1.7.1:

```console
$ jq -n 'def f($param): def h(param): $param; h(1); f(99)'
99
$ succinctly jq -n 'def f($param): def h(param): $param; h(1); f(99)'
1
```

(`h`'s own bare parameter doesn't touch `$param` at all, so it should resolve to `f`'s own `99`; succinctly wrongly resolves it to `h`'s own bound argument, `1`.)

## Suggested fix direction

Needs a second, independent `subst_bare: bool` flag threaded through `substitute_func_param_impl` the way `subst_dollar` already is (most call sites just pass it through unchanged; only the `Expr::FuncDef` arm needs to compute a new value for `body`'s recursion, clearing `subst_bare` when the nested def's own matching parameter exists at all — bare or `$`-style, since either shadows the bare namespace — while `subst_dollar` should only clear when that matching parameter is specifically `$`-style). The `Expr::FuncCall` arm's own unconditional substitution then needs gating on `subst_bare` too.

[...]

Issue #2094, opened 2026-09-01
jq: def declared inside a loop body re-substitutes on every iteration (no BoundBody-style cache for FuncDef)


newhoggy:
## Summary

Follow-up to #1371's runtime-call rewrite, found during code review of that branch.

`Expr::DefCall` (a bound call to a `def`) is memoized per node via `BoundBody` — the substituted body is built once and reused across repeat evaluations of the same node, which is what makes `def`s declared *outside* a loop cheap to call per element.

`Expr::FuncDef` (the `def name(params): body; then` declaration itself) has no equivalent cache. Every one of its four evaluation arms — the plain evaluator's `eval_func_def`, the demand-driven `eval_each`, the path-context evaluator, and `eval_generic`'s `eval_each_generic` — calls `bind_def`, which unconditionally rebuilds the entire `then` subtree via `install_def_calls` on every evaluation of that `FuncDef` node.

## Impact

A `def` declared *inside* a loop body re-pays that full-tree substitution on every iteration:

```
.[] | (def add1: . + 1; add1)
```

or an inline `def` inside a `reduce`/`foreach` update, run over a large array — reconstructs the whole substituted subtree from scratch per element. This reproduces the same per-call regression ADR-0020 measured for `DefCall` before `BoundBody` was added (+7.8%/+26.7% on the benchmarks that motivated that cache), just at the `FuncDef` node instead of the `DefCall` node.

## Suggested fix

Memoize `bind_def`'s result the same way `BoundBody` memoizes `bind_def_call`'s — e.g. an inline cache cell on `Expr::FuncDef` (mirroring `BoundBody`'s `OnceCell`-style pattern), reused across repeat evaluations of the same `FuncDef` node, invalidated the same way `BoundBody` is whenever the node is freshly constructed (not reused) by an enclosing substitution pass.

## Related, minor

`bind_def_call` (`src/jq/eval.rs`) currently does:

```rust
let mut body = def.body.clone();
for (param, arg) in def.params.iter().zip(args.iter()) {
    body = substitute_func_param(&body, param, &Expr::Shared(Rc::new(arg.clone())));
}
```

[...]

PR #2722, opened 2026-09-10
fix(jq): resolve a duplicate def parameter name by its last occurrence

### LINKED ISSUE #2560: jq: bind_def_call's argument binding resolves a duplicate parameter name to the wrong (first) occurrence
## What's wrong

`bind_def_call`'s argument-binding for a `def` with a duplicate parameter name resolves to the *first* occurrence's argument, not the *last* one — confirmed present on unmodified `main` (not a regression from any recent change):

```console
$ jq -n '9 as $a | def f(a; $a): $a; f(1;2)'
2
$ succinctly jq -n '9 as $a | def f(a; $a): $a; f(1;2)'
1
```

Real jq resolves `$a` (and, presumably, a duplicated bare `a` too) via ordinary later-wins parameter shadowing: the *second* `$a` parameter's own argument (`2`) wins. succinctly's call-time argument substitution appears to bind (or leave standing) the *first* occurrence's value instead.

## Scope note (from #2283's own review)

This is unrelated to #2283 (which fixed a separate bug in whether an *outer* `$var` substitution reaches into a nested `def`'s body). `test_duplicate_named_param_shadow_resolves_by_last_occurrence_2283` (`src/jq/eval.rs`) directly confirms #2283's own shadow-check logic *already* resolves duplicate parameter names correctly by the last occurrence — this is a separate bug in `bind_def_call`'s own argument-binding mechanism at call time, not in the shadow-check.

## Repro / pinned as a known gap

`tests/jq_cli_tests.rs`'s `test_duplicate_named_param_argument_binding_known_gap_2283` pins the current (wrong) `1` answer for the repro above, per this project's convention of pinning known gaps rather than leaving them untested.

## Found via

Code review of #2283/its PR, while differentially testing duplicate-parameter-name scenarios against the pinned oracle.

newhoggy:
## Summary

`bind_def_call` folded each `(param, arg)` pair through `substitute_func_param` in
order, so a `def` whose parameter list repeats a name resolved every reference to the
**first** occurrence's argument — the first substitution consumed the matching AST nodes
and left the later, actually in-scope parameter nothing to replace. jq's parameter list
is a chain of nested scopes, innermost last, so a later same-named parameter shadows an
earlier one.

Confirmed live against jq 1.7.1, and present in **both** modes before this change (one
shared `bind_def_call`):

```console
$ jq -nc 'def f(a; $a): $a; f(1;2)'
2
$ succinctly jq -nc 'def f(a; $a): $a; f(1;2)'   # before
1
```

### The part that isn't just "last argument wins"

A `$`-style parameter binds **both** the bare namespace and its own `$name`; a
bare-style parameter binds **only** the bare one. So the two namespaces can have
different winners, and a single last-wins rule gets one of them wrong:

| program | jq 1.7.1 | before | after |
|---|---|---|---|
| `def f(a; $a): $a; f(1;2)` | `2` | `1` | `2` |
| `def f(a; $a): a; f(1;2)` | `2` | `1` | `2` |
| `def f($a; a): $a; f(1;2)` | `1` | `1` | `1` |
| `def f($a; a): a; f(1;2)` | `2` | `1` | `2` |
| `def f($a; a): $a + a; f(1;2)` | `3` | `2` | `3` |
| `def f(a; $a): $a + a; f(1;2)` | `4` | `2` | `4` |
| `def f(a;a;a): a; f(1;2;3)` | `3` | `1` | `3` |
| `def f($a;$a;a): $a + a; f(1;2;3)` | `5` | `2` | `5` |
| `def f(a;$a;$a): $a + a; f(1;2;3)` | `6` | `2` | `6` |

The `$`-first/bare-last row is the discriminating one: `$a` still resolves to the
*leading* `$`-style parameter (the trailing bare parameter creates no `$`-binding to
shadow it with) while a bare `a` in the same body resolves to the *trailing* one.

New `bind_def_call_params` substitutes the bare winner first with `subst_dollar: false`
(so it only rewrites bare `Expr::FuncCall` nodes), then the `$` winner with
`subst_dollar: true`. The ordering is what makes it correct: the bare pass's insertions
are already `Expr::Shared`, which `substitute_func_param_impl` treats as opaque, so the
`$` pass's unconditional bare-matching arm has nothing left to clobber.

**Non-duplicate parameter lists keep the identical single fold, at the identical cost** —
the duplicate-aware path is behind a `has_duplicate_names` scan over a list that is
almost always 0-3 entries.

### Supporting change

`FuncDefData::params` becomes `Vec<Param>` so the bare-vs-`$` spelling survives to call
time. The `param_names` helper existed only to strip it, so it is gone and every
`bind_def` call site passes `params` straight through — a net deletion at ~15 sites.

## Test plan

- [x] Every row in the table above diffed directly against `/usr/bin/jq` (1.7.1), not
      just against golden fixtures
- [x] New regression tests: `test_duplicate_named_param_*_2560` in
      `tests/jq_cli_tests.rs` (5 tests covering same-kind duplicates, the namespace
      split, combined bare+`$` reference bodies, and three occurrences), the yq-mode
      parity test in `tests/yq_cli_tests.rs`, and
      `test_bind_def_call_params_resolves_each_namespace_by_its_own_last_occurrence_2560`
      exercising the helper directly at AST level in `src/jq/eval.rs`
- [x] The previously pinned known-gap test
      (`test_duplicate_named_param_argument_binding_known_gap_2283`) is renamed and
      flipped to jq's real answer

Fixes #2560

### DIFF
--- a/src/bin/succinctly/yq_runner.rs
+++ b/src/bin/succinctly/yq_runner.rs
@@ -9580,7 +9580,7 @@ mod tests {
                 params: if args.is_empty() {
                     Vec::new()
                 } else {
-                    vec!["v".into()]
+                    vec![succinctly::jq::Param::Bare("v".to_string())]
                 },
                 body,
             }),
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -48759,7 +48759,7 @@ fn bind_def_call_params(body: &Expr, params: &[Param], args: &[Expr]) -> Expr {
         // behaviour, not a panic.
         let mut params_and_args = params.iter().zip(args.iter());
         let Some((first_param, first_arg)) = params_and_args.next() else {
-            return body.clone();
+            return body.clone(); // omni-dev: coverage tolerate-line reason="unreachable: bind_def_call only calls this for a non-empty params, and install_def_calls only builds a DefCall whose args.len() equals params.len(), so the zip is never empty here (#2560)"
         };
         let mut result = substitute_func_param(
             body,
@@ -48792,7 +48792,7 @@ fn bind_def_call_params(body: &Expr, params: &[Param], args: &[Expr]) -> Expr {
             // Unreachable for a well-formed `DefCall` (same arity invariant
             // as the non-duplicate path above); skipping rather than
             // unwrapping keeps a malformed one from taking the process down.
-            continue;
+            continue; // omni-dev: coverage tolerate-line reason="unreachable: `name` was just read from `params`, so the zip over (params, args) has a matching pair unless args is shorter than params, which install_def_calls' own arity guard rules out (#2560)"
         };
         result = substitute_func_param_impl(
             &result,
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -48737,11 +48737,32 @@ pub(crate) fn bind_def_call<'e>(
 /// the bare winner first, with `subst_dollar: false` so it only touches
 /// bare `Expr::FuncCall` nodes, then the `$` winner (if any) with
 /// `subst_dollar: true`, reproduces this correctly regardless of which
-/// position wins which namespace: the bare pass's insertions become
-/// `Expr::Shared`, already opaque to the `$` pass's own (unconditional)
-/// bare-matching arm by the time it runs. Verified against every bare/`$`
-/// ordering and combined bare-and-`$`-reference body jq 1.7.1 can express
-/// for two, and for three, duplicated parameters --
+/// position wins which namespace.
+///
+/// **Why the order is what makes that sound.** `subst_dollar` gates exactly
+/// one arm of [`substitute_func_param_impl`] -- `Expr::Var` -- and no other:
+/// it never gates the bare `Expr::FuncCall` arm, and it never decides
+/// whether to recurse, since every shadow check it passes
+/// (`Expr::FuncDef`'s `body_shadowed`/`then_shadowed`, the
+/// `As`/`Reduce`/`Foreach`/`AsPattern` binder checks) reads only `param` and
+/// nodes neither pass rewrites. So the two passes walk *identically*, and
+/// the set of bare references the `$` pass can reach is exactly the set the
+/// bare pass already replaced with an (opaque) `Expr::Shared` -- its
+/// unconditional bare arm therefore has nothing left to clobber. The same
+/// property is what makes the same-winner case (a trailing `$`-style
+/// parameter) compose back into precisely the single `subst_dollar: true`
+/// pass this replaced.
+///
+/// One consequence worth keeping: when a duplicated name has *no* `$`-style
+/// occurrence at all, the `$` pass never runs, so a `$name` in the body is
+/// left for an enclosing binder -- which is what jq does
+/// (`5 as $a | def f(a;a): $a + a; f(1;2)` is `7`; the blanket
+/// `subst_dollar: true` fold this replaced captured `$a` as the parameter
+/// and answered `6`).
+///
+/// Verified against every bare/`$` ordering and combined
+/// bare-and-`$`-reference body jq 1.7.1 can express for two, and for three,
+/// duplicated parameters --
 /// `test_bind_def_call_params_resolves_each_namespace_by_its_own_last_occurrence_2560`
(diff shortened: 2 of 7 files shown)


newhoggy:
## Adversarial review pass

Ran an independent review specifically hunting counterexamples against `/usr/bin/jq`
1.7.1 on duplicate-parameter shapes not covered above. **No divergence attributable to
this change was found** across ~3,500 programs (2,559 randomized over bodies built from
nested `as`/`reduce`/`foreach`/nested-`def`/`map`/`if`/`//`, two seeds, plus ~1,000
hand-built): 4- and 5-parameter mixes, recursion to depth 5000 with a duplicated
parameter name, `?//` alternatives binding the same name, builtin arguments, path and
assignment forms, arguments mentioning the same name, and yq mode.

Three findings were acted on.

### 1. A behaviour change on the duplicate path the PR description had missed — and it is a second fix

When a duplicated name has **no** `$`-style occurrence, the `$` pass never runs, so a
`$name` in the body is left for an enclosing binder. That is what jq does, and it is not
what `main` did:

```console
$ echo null | jq -c            '5 as $a | def f(a;a): $a + a; f(1;2)'   # 7
$ echo null | succinctly jq -c '5 as $a | def f(a;a): $a + a; f(1;2)'   # 7   (main: 6)
```

`main`'s blanket `subst_dollar: true` fold captured the outer `$a` as the parameter.
Nothing pinned this, and that single assertion is what stops the `is_dollar` filter in
`bind_def_call_params` from later being "simplified" into always running the `$` pass —
so it is now pinned as
`test_duplicate_bare_named_param_leaves_an_outer_dollar_binding_alone_2560`, and called
out in the function's doc comment.

It also means `def f(a;a): $a; f(1;2)` now errors (`undefined variable: $a`) where main
answered `1` — jq compile-errors there too, so this is closer. It does leave an internal
inconsistency with the single-parameter fast path, which still answers `1` for
`def f(a): $a; f(1)`; filed as #2726.

### 2. A sharper correctness invariant, now in the doc comment

The original justification ("the bare pass's insertions are `Expr::Shared`, opaque to
the `$` pass") is true but incomplete. The load-bearing property is that **the two
passes have identical traversal structure**: `subst_dollar` gates exactly one arm
(`Expr::Var`) and never influences a recursion decision — every shadow check it passes
reads only `param` and nodes neither pass rewrites. So the set of bare references the
`$` pass can reach is exactly the set the bare pass already replaced, and the
same-winner case provably composes back into the single pass this replaced. Doc comment
updated to argue it that way.

### 3. Reintroduced allocation on the duplicate path, removed

The duplicate path seeded its fold with `body.clone()` — precisely the
O(body size)-per-bound-call allocation that `bind_def_call`'s own #2094 comment ~80
lines above documents removing from this function (and which a recursive `def` with a
duplicated parameter name would pay once per level). Now seeded lazily from the first
substitution, matching the fast path.

### Pre-existing divergences surfaced, filed as follow-ups

All reproduce with a **single, non-duplicated** parameter — i.e. on the fast path this
PR leaves byte-identical — and all were verified present on `main` (`7d84b2329`), so
none is caused by this change:

- #2723 — a bare parameter named after a builtin loses to the builtin
  (`def f(length): length; f(1)` is `0`; `def f(empty): [empty]; f(1)` *deletes* the output)
- #2724 — `{$a}` object-construction shorthand is a parse error (jq builds `{"a":1}`;
  the closed #1139 fixed only the destructuring side). Not in `limitations.md`.
- #2725 — a `$`-style parameter binds a value in jq but a filter here, so
  `def f($a): [$a]; f(1,2)` emits one output instead of two
- #2726 — a bare parameter wrongly creates a `$`-binding (the fast-path sibling of the
  inconsistency in item 1; same theme as #2555, from the entry side)
- #2727 — substitution never descends into `error()`'s message, so `error(a)` leaks
  `"undefined function: a/0"` into a `catch` (the #2095 comments' "latent gap", now with
  a repro)
- #2728 — a leading underscore is rejected in `def`/parameter name position

### Re-verified after the above changes

- [x] full suite, `cargo test --features cli,simd,regex,serde` — 0 failures
- [x] both clippy legs, `-D warnings`
- [x] patch coverage 98.26% before the last two commits; the only uncovered lines were
      the unreachable arity guards, now `tolerate-line`d with the invariant that makes
      them unreachable

Issue #2734, opened 2026-09-10
jq: an unbound variable is rejected at evaluation (exit 5), not at compile time (exit 3) like jq


newhoggy:
**Severity: Low**

## Summary

jq resolves variable references at compile time and refuses to run a program that
mentions an unbound `$name`. succinctly accepts the program and fails when evaluation
reaches the reference, so the exit code differs (3 vs 5), the message differs, and — the
part that actually matters — any output the program produced *before* reaching the bad
reference has already been written.

## Repro

Confirmed live against jq 1.7.1 and on `main` (`9693440ff`):

```console
$ echo null | jq -c '$nope'
jq: error: $nope is not defined at <top-level>, line 1:
$nope
jq: 1 compile error
$ echo $?
3

$ echo null | succinctly jq -c '$nope'
jq: error (at <stdin>:1): undefined variable: $nope
$ echo $?
5
```

The consequence, which is more than cosmetic — jq emits nothing, succinctly emits the
first value and then fails:

```console
$ echo null | jq -c '1, $nope'
jq: error: $nope is not defined at <top-level>, line 1:
1, $nope
jq: 1 compile error          # no output at all

$ echo null | succinctly jq -c '1, $nope'
1
jq: error (at <stdin>:1): undefined variable: $nope
```

## Root cause

`src/jq/resolve.rs`'s `resolve_func_calls` already does exactly this job for *functions*
— an undefined function is a compile error here, matching jq (`f` alone reports
"f/0 is not defined" before evaluation). There is no equivalent resolve pass for
variables, so `Expr::Var` survives to evaluation and errors there.

## Why this is separate from #2555/#2726

Found while fixing those two (PR pending), which make a *bare* `def` parameter stop
wrongly binding `$name` — so `def f(a): $a; f(1)` now correctly rejects instead of
answering `1`. It rejects via this runtime path, i.e. it is now exactly as
right-or-wrong as every other unbound variable, which is why this was not folded in:
the timing gap is general, predates both, and reproduces with a bare `$nope` that
involves no `def` at all.

## Suggested fix direction

Extend the resolve pass to track variable scope alongside function scope and report an
unbound `$name` there, the way `resolve_func_calls` reports an unbound function. The
scope tracking is the real work: `as`/`reduce`/`foreach`/`?//` patterns, `def`
parameters (including the `$`-style ones that bind both namespaces — see #2555),
[...]

Issue #2727, opened 2026-09-10
jq: substitution never descends into error()'s message, so a parameter used as error(a) leaks as "undefined function: a/0"


newhoggy:
**Severity: Low**

## Summary

`Expr::Error`'s message sub-expression is cloned rather than recursed into by all three
tree rewriters (`substitute_func_param_impl`, `substitute_var_impl`, `map_subexprs`), so
a `def` parameter — or a bound variable — referenced inside `error(...)` is never
substituted. It survives to evaluation as an unresolved call and surfaces as a
confusing message.

## Repro

Confirmed live against jq 1.7.1 and on `main` (`7d84b2329`):

```console
$ echo null | jq -c 'def f(a): try error(a) catch .; f("x")'
"x"
$ echo null | succinctly jq -c 'def f(a): try error(a) catch .; f("x")'
"undefined function: a/0"
```

The caught value is succinctly's *own internal resolution failure*, not the error the
program asked for — so a program that catches and inspects its errors sees fabricated
content.

## Root cause

Already a documented known gap, but as a code comment rather than a tracked issue —
`map_subexprs`' `Expr::Error` arm (`src/jq/walk.rs`) and the mirroring arms in
`substitute_func_param_impl` / `substitute_var_impl` (`src/jq/eval.rs`) each carry a
`#2095` comment saying the non-recursion "is preserved as a likely latent gap rather
than fixed here". This is that gap, with a concrete repro attached.

## Why this is separate from #2560

Unrelated mechanism — #2560 was `bind_def_call`'s ordering across duplicated parameter
names; this is which sub-expressions the substitution walks at all. Reproduces with a
single parameter on the untouched fast path, and identically on `main`. It is also not
parameter-specific: a plain `as`-bound variable inside `error(...)` should be checked
the same way while fixing it.

## Suggested fix direction

Give `Expr::Error` a recursing arm in `map_subexprs`, and let the two substitution
functions fall through to it rather than keeping their own cloning arms — the #2095
comments already point at this as the shape of the fix. Worth checking first whether any
current behaviour depends on the message *not* being rewritten (the comments imply the
non-recursion was inherited rather than chosen, but they stop short of saying it is
safe to change).

Found via adversarial review of #2560 (PR #2722).

newhoggy:
🔧 Claimed by agent `Agatha` — branch `issue-2727-error-message-substitution`

PR #2735, opened 2026-09-10
fix(jq): shadow a parameter's bare and $ namespaces independently

### LINKED ISSUE #2555: jq: substitute_func_param_impl's own outer-bare-param shadow check needs a second subst_bare flag, not just dollar_params

### LINKED ISSUE #2726: jq: a bare def parameter wrongly creates a $-binding — def f(a): $a; f(1) answers 1 where jq rejects the program
**Severity: Low**

## Summary

A *bare* `def` parameter creates no `$`-binding in jq — `$a` inside the body is a
compile error unless something else binds it. succinctly's call-time substitution
resolves `$a` to the parameter's argument anyway.

## Repro

Confirmed live against jq 1.7.1 and on `main` (`7d84b2329`):

```console
$ echo null | jq -c 'def f(a): $a; f(1)'
jq: error: $a is not defined at <top-level>, line 1:
def f(a): $a; f(1)
jq: 1 compile error

$ echo null | succinctly jq -c 'def f(a): $a; f(1)'
1
```

## Root cause

`bind_def_call_params` (`src/jq/eval.rs`, #2560) runs the non-duplicate fast path
through `substitute_func_param`, which starts `subst_dollar: true` unconditionally — so
the `Expr::Var` arm fires for `$a` regardless of whether the parameter it is binding was
spelled `a` or `$a`. `Param::Bare`/`Param::Dollar` now reaches this function (#2560 gave
`FuncDefData::params` the `Param` type for exactly this kind of question), so the
information needed to gate it is present; the fast path just does not consult it.

## Why this is separate from #2560

#2560 fixed the *duplicate*-parameter path, where the winner selection does consult
`is_dollar` — so this exact program shape is already correct there:

```console
$ echo null | succinctly jq -c 'def f(a;a): $a; f(1;2)'
jq: error (at <stdin>:1): undefined variable: $a      # jq also rejects it
```

The single-parameter fast path was deliberately left byte-identical to pre-#2560
behaviour, which is why the inconsistency above exists: `def f(a;a): $a` now rejects,
`def f(a): $a` still answers `1`.

## Relationship to #2555

#2555 is the same underlying "bare and `$` shadowing need separate flags" theme, but
from the *nested-def shadow check* side (`substitute_func_param_impl`'s `Expr::FuncDef`
arm blocking an outer `$param` from reaching a nested bare-only parameter's body). This
one is at the *entry* — which `subst_dollar` a top-level substitution starts with. They
would likely be fixed together; whoever picks one should read the other.

## Suggested fix direction

Start the fast path's substitution with `subst_dollar: param.is_dollar()` instead of an
unconditional `true`. Cheap to try, but it changes behaviour for any program relying on
the current leniency, so it needs the full suite plus an oracle diff — and the resulting
error message should be checked against jq's, which rejects at *compile* time rather
than at evaluation.

Found via adversarial review of #2560 (PR #2722).

newhoggy:
## Summary

`substitute_func_param_impl` carried a single `subst_dollar` flag for two genuinely
separate jq namespaces. A bare `name` reference (`Expr::FuncCall { args: [] }`, jq's
call-by-name spelling for a parameter) lives in the *filter* namespace; `$name`
(`Expr::Var`) lives in the *variable* namespace. What binds each differs, so what
shadows each differs — and collapsing them into one flag was wrong in **both**
directions.

This replaces the flag with `SubstScope { dollar, bare }`, narrowed per-namespace as the
walk crosses each binder. Three bugs fall out. All rows below confirmed live against
jq 1.7.1.

### 1. #2555 — an outer `$param` could not reach through a nested bare-only parameter

The `Expr::FuncDef` arm blocked both namespaces whenever a nested def had a matching
parameter. But a *bare* parameter binds no `$name` at all, so there is nothing there to
shadow `$param` with:

| program | jq 1.7.1 | before | after |
|---|---|---|---|
| `def f($param): def h(param): $param; h(1); f(99)` | `99` | `1` | `99` |
| `def f($a): def h(a): [a, $a]; h(0); f(1)` | `[0,1]` | `[0,0]` | `[0,1]` |
| `def f($a): def h(a): select($a == 1); h(0); f(1)` | `null` | *(no output)* | `null` |
| `def f($a): def h(a): reduce range(2) as $i (0; . + $a); h(0); f(7)` | `14` | `0` | `14` |
| `def f($a): def h(a): def g(a): $a; g(2); h(1); f(3)` | `3` | `2` | `3` |

The issue only had the first row; the rest are the same conflation reached through a
builtin argument, a `reduce`, and a second def level.

### 2. A third instance the issue does not name

The same arm's `then` check (`then_shadowed`, #2077) also blocked both namespaces, for a
nested **zero-argument `def`** of the parameter's name. A `def` binds a filter name and
never a `$name`, so this one did not merely answer wrongly — it failed outright:

| program | jq 1.7.1 | before | after |
|---|---|---|---|
| `def f($a): def a: 99; $a; f(2)` | `2` | `error: undefined variable: $a` | `2` |

### 3. #2726 — a bare parameter wrongly claimed `$name`

The substitution entered with `subst_dollar: true` regardless of the parameter's own
spelling, so a bare parameter bound `$name` as if it had been written `$name`:

| program | jq 1.7.1 | before | after |
|---|---|---|---|
| `def f(a): $a; f(1)` | compile error, `$a is not defined` | `1` | rejects |
| `9 as $a \| def f(a): $a; f(1)` | `9` | `9` | `9` |
| `9 as $a \| def f(a): [a, $a]; f(1)` | `[1,9]` | `[1,1]` | `[1,9]` |

It now rejects by the same route succinctly rejects any other unbound variable. That
route still differs from jq in *timing* — jq refuses at compile time (exit 3), this at
evaluation (exit 5) — but identically for a bare `$nope` involving no `def` at all, so
that gap is general and pre-existing; filed as #2734.

### The bare half is unchanged, and pinned

A nested matching parameter of **either** spelling still shadows a bare reference, since
a `$`-style parameter binds the bare namespace too (`def h($param): param` alone is a
compile error, *"param is not defined"*). `def f(param): def h($param): param; h(1); f(5)`
is still `1`, `def f($a): def a: 99; a; f(2)` still `99`, and
`def f(g): 1 as $g | $g + g; 10 as $g | f($g)` still `11` — an `as` binder never touches
the bare namespace however it is spelled.

### Two things fall out

- **`bind_def_call_params`' two passes (#2560) now say what they mean.**
  `BARE_NAMESPACE_ONLY` and `DOLLAR_NAMESPACE_ONLY` rewrite disjoint node sets, so their
  order stops being load-bearing. Previously the bare arm fired unconditionally and
  correctness rested on the bare pass running first and leaving opaque `Expr::Shared`
  behind — right, but only by construction rather than by the flags saying so.
- **`subst_unless_shadowed` becomes `subst_in_scope`.** It can no longer clone on the
  first shadow, because a subtree can be shadowed in one namespace while the other still
  has substituting to do; only an *empty* scope short-circuits now (which is exactly the
  old fast path).

### Shared predicate

Folds the duplicated "does this parameter list bind `$name`?" logic — written separately
in `substitute_var_impl` and now needed here — into one `params_bind_dollar`, with
`test_params_bind_dollar_agrees_with_both_shadow_sites_2555` checking each row three
ways (the predicate, and both call sites through their public entry points). This repo
has been bitten by duplicated predicates diverging silently (#106).

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — full suite, 62 binaries, 0 failures
- [x] Every row in all four tables above diffed directly against `/usr/bin/jq` (1.7.1)
- [x] 20 must-not-regress shapes re-checked against the oracle, plus the whole #2560
      duplicate-parameter matrix (13 shapes), all unchanged
- [x] The known-gap test `test_bare_nested_param_wrongly_shadows_outer_dollar_func_param_known_gap_2283`
      is replaced by `test_outer_dollar_param_reaches_through_a_bare_nested_param_2555`,
      pinning jq's real answer — it was the *only* test in the suite that failed against
      this change
- [x] New: 4 `_2555` CLI tests, 2 `_2726` CLI tests, and the shared-predicate agreement
      test

Fixes #2555, Fixes #2726

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -34579,11 +34579,7 @@ fn substitute_var_impl(
             // f(1)` is `3` (bare -- outer reaches in), `3 as $g | def
             // f($g): $g; f(1)` is `1` (dollar-style -- shadowed), matching
             // this fix.
-            let shadowed = params
-                .iter()
-                .rev()
-                .find(|p| p.name() == var_name)
-                .is_some_and(Param::is_dollar);
+            let shadowed = params_bind_dollar(params, var_name);
             Expr::FuncDef {
                 name: name.clone(),
                 params: params.clone(),
@@ -48764,24 +48760,18 @@ pub(crate) fn bind_def_call<'e>(
 /// possibly a different, earlier position (`def f($a; a): $a; f(1;2)` is
 /// `1`: the trailing bare `a` shadows the bare namespace only, so the
 /// leading `$a`'s own binding is still what `$a` resolves to). Substituting
-/// the bare winner first, with `subst_dollar: false` so it only touches
-/// bare `Expr::FuncCall` nodes, then the `$` winner (if any) with
-/// `subst_dollar: true`, reproduces this correctly regardless of which
-/// position wins which namespace.
+/// the bare winner under [`BARE_NAMESPACE_ONLY`] and the `$` winner (if any)
+/// under [`DOLLAR_NAMESPACE_ONLY`] reproduces this, whichever position wins
+/// which namespace.
 ///
-/// **Why the order is what makes that sound.** `subst_dollar` gates exactly
-/// one arm of [`substitute_func_param_impl`] -- `Expr::Var` -- and no other:
-/// it never gates the bare `Expr::FuncCall` arm, and it never decides
-/// whether to recurse, since every shadow check it passes
-/// (`Expr::FuncDef`'s `body_shadowed`/`then_shadowed`, the
-/// `As`/`Reduce`/`Foreach`/`AsPattern` binder checks) reads only `param` and
-/// nodes neither pass rewrites. So the two passes walk *identically*, and
-/// the set of bare references the `$` pass can reach is exactly the set the
-/// bare pass already replaced with an (opaque) `Expr::Shared` -- its
-/// unconditional bare arm therefore has nothing left to clobber. The same
-/// property is what makes the same-winner case (a trailing `$`-style
-/// parameter) compose back into precisely the single `subst_dollar: true`
-/// pass this replaced.
+/// The two passes are independent: each rewrites a disjoint set of nodes
+/// ([`SubstScope`] gates the `Expr::Var` and bare `Expr::FuncCall` arms
+/// separately since #2555), so neither can consume or clobber what the other
+/// is looking for, and their order does not matter. Before #2555 the bare
+/// arm fired unconditionally, and this relied on the bare pass running first
+/// and leaving opaque `Expr::Shared` behind for the `$` pass to skip over --
+/// correct, but only by construction rather than by the flags actually
+/// saying so.
 ///
 /// One consequence worth keeping: when a duplicated name has *no* `$`-style
 /// occurrence at all, the `$` pass never runs, so a `$name` in the body is
@@ -48812,14 +48802,10 @@ fn bind_def_call_params(body: &Expr, params: &[Param], args: &[Expr]) -> Expr {
         let Some((first_param, first_arg)) = params_and_args.next() else {
             return body.clone(); // omni-dev: coverage tolerate-line reason="unreachable: bind_def_call only calls this for a non-empty params, and install_def_calls only builds a DefCall whose args.len() equals params.len(), so the zip is never empty here (#2560)"
         };
-        let mut result = substitute_func_param(
-            body,
-            first_param.name(),
-            &Expr::Shared(Rc::new(first_arg.clone())),
-        );
+        let mut result =
+            substitute_func_param(body, first_param, &Expr::Shared(Rc::new(first_arg.clone())));
         for (param, arg) in params_and_args {
-            result =
-                substitute_func_param(&result, param.name(), &Expr::Shared(Rc::new(arg.clone())));
+            result = substitute_func_param(&result, param, &Expr::Shared(Rc::new(arg.clone())));
         }
         return result;
     }
@@ -48855,7 +48841,7 @@ fn bind_def_call_params(body: &Expr, params: &[Param], args: &[Expr]) -> Expr {
             result.as_ref().unwrap_or(body),
             name,
             &Expr::Shared(Rc::new(bare_arg.clone())),
-            false,
+            BARE_NAMESPACE_ONLY,
         );
 
         if let Some((_, dollar_arg)) = by_name().rfind(|(p, _)| p.is_dollar()) {
@@ -48863,7 +48849,7 @@ fn bind_def_call_params(body: &Expr, params: &[Param], args: &[Expr]) -> Expr {
                 &substituted,
                 name,
                 &Expr::Shared(Rc::new(dollar_arg.clone())),
-                true,
+                DOLLAR_NAMESPACE_ONLY,
             );
         }
         result = Some(substituted);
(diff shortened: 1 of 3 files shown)


newhoggy:
## Adversarial review pass — caught a regression in this branch

Ran an independent review hunting counterexamples against `/usr/bin/jq` 1.7.1. It found
a **real regression introduced by the first draft of this PR**, now fixed in
`fix(jq): ask "binds $name at all", not "is the last one $-style"`. Reproduced
independently before acting on it.

### The bug I introduced

A repeated parameter name asks **three** separate questions, and only two are decided by
the last occurrence:

1. *Does this list bind `$name` at all?* → **any** `$`-style occurrence
2. *Which argument does `$name` resolve to inside the def?* → the last `$`-style occurrence
3. *Which argument does bare `name` resolve to?* → the last of **any** spelling

`params_bind_dollar` answered (1) with (2)'s rule, carried over from #2560 where it is
correct. A trailing *bare* parameter shadows only the filter namespace and cannot unbind
a leading `$a`:

| program | jq 1.7.1 | `main` | first draft | now |
|---|---|---|---|---|
| `def f($a): def h($a; a): $a; h(1;2); f(9)` | `1` | `1` | **`9`** | `1` |
| `def f($a): def h($a; a): [$a, a]; h(1;2); f(9)` | `[1,2]` | `[1,1]` | **`[9,2]`** | `[1,2]` |
| `def f($a): def h($a; a; a): $a; h(1;2;3); f(9)` | `1` | `1` | **`9`** | `1` |
| `def f($a): def h($a; a): def q: $a; q; h(1;2); f(9)` | `1` | `1` | **`9`** | `1` |

`main` got these right by accident — its blanket check blocked both namespaces at once,
so splitting the `$` half onto the wrong predicate un-blocked it.

### The corrected predicate also fixes a pre-existing bug

The same defect exists at the sibling call site (`substitute_var_impl`) on `main`, where
nothing had exercised a repeated `($a; a)` list with an enclosing `$a` to leak:

| program | jq 1.7.1 | `main` | now |
|---|---|---|---|
| `9 as $a \| def f($a; a): $a; f(1;2)` | `1` | **`9`** | `1` |
| `5 as $a \| def f($a;a): [$a, a]; f(1;2)` | `[1,2]` | **`[5,1]`** | `[1,2]` |

**#2283's own unit test asserted that leak as intended behaviour**, so it is corrected
here against the oracle rather than worked around, and gains the row that separates the
three questions (`def f(a; a)` — no `$`-style occurrence at all, where the outer value
genuinely *does* reach through; jq: `9`).

### Also addressed

- Comments left naming `subst_dollar` after the rename, and one claiming the bare arm
  still fires unconditionally — both now false.
- `walk.rs`'s `FuncDef` bullet still described the old two-condition shape.
- `SubstScope`'s doc claimed a nested zero-argument `def` clears the bare namespace in
  its `body` as well as its `then`. It does not, and that gap is real —
  `def f(a): def a: if . > 0 then (.-1|a) else 100 end; (3|a); f(1)` is `100` in jq and
  `1` here. **Not fixed here**: closing it makes `def f(a): def a: a+1; a; f(1)` recurse
  forever, as jq does, which needs its own oracle matrix and a look at the depth guard.
  Filed as #2737; the doc now says so instead of overclaiming.
- The #2726 rejection is a *catchable* evaluation error, unlike jq's uncatchable compile
  error (`def f(a): try $a catch "caught"; f(1)` → `"caught"`). Now pinned rather than
  left implicit, since it is the visible consequence of #2734 landing on a
  newly-rejected program.

### Checked and clean

~5,000 generated programs compared three ways (jq / branch / `main`) across three
corpora: every divergence found reproduced byte-identically on `main`. The per-namespace
narrowing was reviewed arm by arm (`As`, `Reduce`/`Foreach`/`AsPattern` via
`pattern_binds_var` including `?//` and `{$a}` forms, `then_scope`'s arity condition,
`map_subexprs`' `DefCall`/`Shared` policies). Deliberately absent `Expr::Label` arm
confirmed correct (`def f($a): label $a | $a; f(1)` is `1` in both).

**Performance** of replacing the old clone-on-shadow with a narrowed walk: neutral. The
old path deep-cloned the subtree anyway, so this swaps a copy for a walk-and-rebuild.
Interleaved A/B, 7 reps, median, on a 200-deep recursive `def` whose per-level nested
`def h(a)` body is a 20,000-element array literal (proved to hit the changed path):
`main` 742 ms vs branch 749 ms, ratio 1.010; control 1.000.

### Further pre-existing bugs filed

All verified present on `main` first: #2737 (nested zero-arg `def` not in scope in its
own body), #2738 (an earlier `def` shadows a later def's same-named *parameter* —
`def b: 8; def h(b): b; h(6)` is `8`, jq says `6`; the mirror of #2555 in
`install_def_calls`), #2739 (`label $x` wrongly shadows the *variable* `$x`), #2740 (an
undefined name inside a never-called `def` is compile-rejected; jq accepts the program).
#2727's `Expr::Error` non-recursion also now has an end-to-end repro:
`def f($a): def h(a): error("x\($a)"); try h(1) catch .; f(9)` is `"x9"` in jq,
`"undefined variable: $a"` here.

Issue #2744, opened 2026-09-10
jq: path()'s untracked-navigation table can't answer for nth/reverse/indices, whose element depends on an argument or length


newhoggy:
**Severity: Low**

## Summary

#2646 (PR pending) added `builtin_navigation` (`src/jq/eval.rs`), a static table saying
which element each jq-defined navigating builtin would index, so `path()` can raise
against an untracked value the way jq does. A third group needs a value the table is not
given — an evaluated argument, or the input's `length` — so they are absent and still
wrongly succeed.

## Repro

Confirmed live against jq 1.7.1, and still present with #2646's fix applied:

```console
$ echo '{}' | jq -c '[path(([1,2,3] | nth(2)) | empty)]'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 2 of [1,2,3]
$ echo '{}' | succinctly jq -c '[path(([1,2,3] | nth(2)) | empty)]'
[]

$ echo '{}' | jq -c '[path(([1,2,3] | reverse) | empty)]'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 2 of [1,2,3]
$ echo '{}' | succinctly jq -c '[path(([1,2,3] | reverse) | empty)]'
[]

$ echo '{}' | jq -c '[path(([1] | indices(1)) | empty)]'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element [1] of [1]
$ echo '{}' | succinctly jq -c '[path(([1] | indices(1)) | empty)]'
[]
```

Affected: `nth(n)`, `reverse`, `indices(i)`, `index(i)`, `rindex(i)`.

## Why each needs more than a constant

- **`nth(n)`** — the element *is* the argument, so it must be evaluated against the
  current input first (`nth("x")` reports `"x"`, not a type error).
- **`reverse`** — the element is `length - 1`, so it depends on the input. It also does
  **not raise at all** on an empty input: `[] | reverse` inside `path()` answers `[]`,
  because jq defines it as `[.[length - 1 - range(0;length)]]` and `range(0;0)` is empty,
  so no navigation happens. A naive `length-1` entry would wrongly raise there.
- **`indices(i)`/`index(i)`/`rindex(i)`** — the element is the argument wrapped in an
  array (`indices(1)` reports `[1]`), so it needs the evaluated argument too.

## Suggested fix direction

Widen `builtin_navigation` from a `fn(&Builtin, &OwnedValue) -> Option<BuiltinNavigation>`
[...]

Issue #2743, opened 2026-09-10
jq: path()'s untracked-navigation check can't name a derived container, so unique/map_values/with_entries/walk-on-object still don't raise


newhoggy:
**Severity: Low**

## Summary

#2646 (PR pending) made the jq-defined navigating builtins raise inside `path()` when
applied to an untracked value, via a static table in `builtin_navigation`
(`src/jq/eval.rs`). A second group raises in jq against a container that is *derived*
from the input rather than the input itself, which a static answer cannot produce — so
they are deliberately absent from that table and still wrongly succeed.

## Repro

Confirmed live against jq 1.7.1, and still present with #2646's fix applied:

```console
$ echo '{}' | jq -c '[path(([1] | unique) | empty)]'
jq: error (at <stdin>:1): Invalid path expression near attempt to iterate through [[1]]
$ echo '{}' | succinctly jq -c '[path(([1] | unique) | empty)]'
[]

$ echo '{}' | jq -c '[path(([1] | map_values(.)) | empty)]'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 0 of [[1],[]]
$ echo '{}' | succinctly jq -c '[path(([1] | map_values(.)) | empty)]'
[]

$ echo '{}' | jq -c '[path(({"k":1} | walk(.)) | empty)]'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 0 of [{"k":1},[]]
$ echo '{}' | succinctly jq -c '[path(({"k":1} | walk(.)) | empty)]'
[]
```

Affected: `unique`, `unique_by(f)`, `map_values(f)`, `with_entries(f)`, `fromstream(f)`,
and `walk(f)` **on an object only** (its array arm is fixed by #2646 — `walk`'s own
definition branches on input type, so only the object arm goes through `map_values`).

## Root cause

Each is defined in jq in terms of another builtin that materializes an intermediate:
`unique` is `group_by(.) | map(.[0])`, so the container jq names is `group_by`'s output
`[[1]]`, not the input `[1]`. `map_values(f)` is `.[] |= f`, reporting `[[1],[]]`.
`with_entries(f)` reports the `to_entries` output.

#2646's `builtin_navigation` answers with a `BuiltinNavigation::Access(element)` or
`Iterate` against the value it is handed, so it can only ever name the input. Reporting
these with the input as the container would produce a *wrong* message rather than a
missing one, so they were left out on purpose — see that function's own doc comment.

[...]

PR #2745, opened 2026-09-10
fix(jq): raise inside path() for the builtins jq defines as navigation

### LINKED ISSUE #2646: jq: `first`/`last`/`add` on a constructed value inside `path()` never raise, so an empty continuation succeeds where jq errors
**Severity: Medium** (accept-where-jq-refuses; the accepted answer is *empty*, so `del()`/`|=` return the document unchanged where jq errors — no wrong value is written, but exit 0 replaces exit 5)

## Summary

`first`, `last` and `add` navigate inside their own jq-level definition (`.[0]`, `.[-1]`,
`reduce .[] as $x ...`). Applied to a *constructed* value inside `path()`, jq raises the
"near attempt to access element" error at that internal navigation. succinctly's resolver
sees an opaque builtin, evaluates it as a computed value and carries on untracked — so when
the rest of the pipe produces **no output at all**, nothing ever checks the untracked branch
and the program succeeds with an empty result.

Found by `scripts/jq-bind-origin-fuzz.py` (#2042) and confirmed pre-existing: every shape
below reproduces byte-for-byte on `main` @ baa26d72e.

## Repro (jq 1.7.1)

```console
$ echo '{"a":1}' | jq -c '[path(([1] | first) | empty)]'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 0 of [1]
$ echo '{"a":1}' | succinctly jq -c '[path(([1] | first) | empty)]'
[]

$ echo '{"a":1}' | jq -c 'del(([1] | first) | empty)'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 0 of [1]
$ echo '{"a":1}' | succinctly jq -c 'del(([1] | first) | empty)'
{"a":1}                                   # exit 0

$ echo '{"a":1}' | jq -c '[path(([1] | last) | empty)]'          # refuses;  succinctly: []
$ echo '{"a":1}' | jq -c '[path(([[1]] | add) | empty)]'         # refuses;  succinctly: []
$ echo '{"a":1}' | jq -c '[path(([1] | first) | select(false))]' # refuses;  succinctly: []
```

Controls (both refuse): the explicit navigation `[path(([1] | .[0]) | empty)]`, and any shape
where the untracked value *is* consumed — `[path(([1] | first) | .b?)]` raises here too, only
with the later `.b` in the message instead of jq's `0`.

## Why #2041's allowlist does not cover it

`cannot_move_register` (#2041) already knows `first`/`add`/a `def` body can move jq's
register, and refuses to *carry a register across* them. That gate only fires when a later
stage tries to re-establish; it does not make the stage itself raise. jq raises at the
`INDEX` inside `first` regardless of what follows.

## Fix direction

Treat the jq-defined navigating builtins as navigation on an untracked input: when the
ambient branch is untracked (a constructed value), `first`/`last`/`add`/`nth`/... should
raise `invalid_path_expression_near_access` for the element they would index (`0`, `-1`,
`iterate`) before evaluating, the way `resolve_leaf` already does for a bare `Field`/`Index`
via `navigation_element`. Probably a small table next to `cannot_move_register` so the two
cannot drift.

## Related

- #2041 — `cannot_move_register`'s allowlist, where these builtins are already named.
- #2042 — the fuzz that found it; classified there as `fabricate-baseline` (pre-existing).

newhoggy:
## Summary

jq defines a family of builtins *in jq itself*, in terms of ordinary navigation — `first`
is `.[0]`, `last` is `.[-1]`, `add` is `reduce .[] as $x ...`. Inside `path()`, that
internal navigation obeys the same rule as one the user wrote: applied to a value the
resolver is not tracking, it raises. succinctly saw an opaque builtin, evaluated it as a
plain value, and let the branch go untracked — so a continuation that emits nothing left
nothing for the terminal check to object to.

```console
$ echo '{"a":1}' | jq -c '[path(([1] | first) | empty)]'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 0 of [1]
$ echo '{"a":1}' | succinctly jq -c '[path(([1] | first) | empty)]'      # before
[]

$ echo '{"a":1}' | jq -c 'del(([1] | first) | empty)'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 0 of [1]
$ echo '{"a":1}' | succinctly jq -c 'del(([1] | first) | empty)'          # before
{"a":1}                                                                   # exit 0
```

The `del()` shape is the one that matters: no wrong value is written, but a **refused
edit is reported as a successful no-op** (exit 0 replacing exit 5).

## How

`builtin_navigation` answers the `navigation_element` question for a builtin — "what does
this navigate on its input, and with which element?" — and feeds the same two error
constructors. `resolve_leaf`'s existing `!trackable` block consults it immediately after
the `Field`/`Index` check it mirrors.

**Membership is decided by jq's implementation, not by whether a builtin looks like it
reads a container**, and those are different sets. `sort`, `min`, `max`, `group_by`,
`to_entries`, `keys`, `has`, `getpath` and `tostream` all read their input and all stay
silent, because jq implements them in C and no jq-level navigation happens. `unique`
raises only because it is *defined* as `group_by(.) | map(.[0])`. Every entry, and every
negative, was captured live from jq 1.7.1.

### Two boundaries the oracle settled

Both would have been got wrong by a plausible-looking table:

- **Arity changes the answer.** `any`/`any(cond)` iterate their input and raise;
  `any(gen; cond)` iterates `gen` instead, so `[1] | any(1;.)` must stay silent. That
  `any(.[];.)` raises is its *argument* navigating — something the resolver already
  handled — not `any` itself. `AnyCond`/`AllCond` are therefore deliberately out.
- **`walk` is the only input-dependent entry**, because its own definition branches on
  `type`. Only the array arm reaches a plain `.[]`. A scalar navigates nothing, so
  `5 | walk(.)` must **not** raise (an unconditional entry would have introduced a fresh
  divergence), and the object arm goes through `map_values`, naming a derived container
  this table cannot produce.

### Deliberately deferred

Left out rather than reported with a *wrong container*:

- #2743 — `unique`, `unique_by`, `map_values`, `with_entries`, `fromstream`, and `walk`
  on an object: each raises against a value derived from the input
  (`[1] | map_values(.)` reports `element 0 of [[1],[]]`), which a static answer cannot
  name.
- #2744 — `nth(n)`, `reverse`, `indices(i)`, `index(i)`, `rindex(i)`: the element depends
  on an evaluated argument or on `length`, and `reverse` does not raise at all on an
  empty input.

## Also fixed

The "already refuses" control diverged in its *message*, and now matches:

```console
$ echo '{"a":1}' | jq -c '[path(([1] | first))]'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 0 of [1]
$ echo '{"a":1}' | succinctly jq -c '[path(([1] | first))]'    # before
jq: error (at <stdin>:1): Invalid path expression with result 1
```

Same cause: the raise now happens at the navigation instead of one stage later at the
terminal check. This also sharpened an existing pin —
`test_path_register_dropped_across_opaque_navigating_stage_1573` asserted only "some path
error", and four of its five rows had been getting the generic wording; they now match
jq exactly, so each row asserts its own message. Its `def f: .a` row keeps the generic
wording, in jq too (a user-defined body is not in the table).

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — full suite, 62 binaries, 0 failures
- [x] **296-case sweep** — 37 builtins × 8 input shapes (`[1]`, `[]`, `[1,2,3]`,
      `{"k":1}`, `{}`, `5`, `"s"`, `null`) diffed against `/usr/bin/jq` 1.7.1: **292
      match**, and the 4 that do not are exactly #2743's `walk`-on-object
- [x] Three new CLI tests: the fixed shapes, the negative boundary (C-implemented
      builtins, the 2-arg `any`/`all`, `walk` on a scalar), and input-type independence

Fixes #2646

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -737,7 +737,18 @@ is the revert that established what the other one costs.
    navigating inside their own jq-level definitions against a *constructed* value inside
    `path()` never raise, found by `scripts/jq-bind-origin-fuzz.py`'s differential fuzz and
    confirmed pre-existing (reproduces byte-for-byte on the commit before #2042), not caused by
-   this change.
+   this change. **#2646 is now fixed** — `builtin_navigation` (`src/jq/eval.rs`) answers what
+   each jq-defined navigating builtin indexes, and `resolve_leaf`'s `!trackable` guard raises
+   on it. Three groups remain, deliberately, and are divergences in their own right:
+
+   | Still diverging                                                                     | jq 1.7.1                                        | succinctly | Tracked                                                                  |
+   | ----------------------------------------------------------------------------------- | ----------------------------------------------- | ---------- | ------------------------------------------------------------------------ |
+   | `[path(([1] \| unique) \| empty)]`, and `unique_by`/`map_values`/`with_entries`/`fromstream`, `walk` on an *object*, `ascii_downcase`/`ascii_upcase`, the `match`/`sub`/`gsub` family | raises, naming a **derived** container (`iterate through [[1]]`) | `[]`       | [#2743](https://github.com/rust-works/succinctly/issues/2743) |
+   | `[path(([1,2,3] \| nth(2)) \| empty)]`, and `reverse`/`indices`/`index`/`rindex`; `INDEX(f)` and `transpose` | raises; element depends on an argument or `length` | `[]`       | [#2744](https://github.com/rust-works/succinctly/issues/2744) |
+   | `[path(([1] \| last(.[])) \| empty)]`, and `isempty(f)`/`any(gen;cond)`/`all(gen;cond)`/`INDEX(gen;f)` | raises — the *argument* navigates                | `[]`       | [#2746](https://github.com/rust-works/succinctly/issues/2746) |
+
+   The rule is **jq-mode only** (ADR-0018): `map`, `any`, `all` and `flatten` are real yq
+   builtins, and yq v4.53.3 raises for none of them, so `succinctly yq` does not either.
 
    [#2072](https://github.com/rust-works/succinctly/issues/2072) supplied the missing
    half of that but deliberately did not spend it here. `Expr::TrackedVar` now carries a
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -28565,6 +28565,86 @@ fn navigation_element(component: &Expr) -> Option<OwnedValue> {
     }
 }
 
+/// How a builtin navigates *its own input*, for the untracked-value check
+/// in [`resolve_leaf`] (#2646).
+enum BuiltinNavigation {
+    /// Indexes one element, like `.[k]` — carries the `k` jq names.
+    Access(OwnedValue),
+    /// Iterates, like `.[]`.
+    Iterate,
+}
+
+/// The navigation a builtin performs on its input before anything else, or
+/// `None` if it performs none (#2646).
+///
+/// jq defines a family of builtins *in jq itself*, in terms of ordinary
+/// navigation — `first` is `.[0]`, `last` is `.[-1]`, `add` is
+/// `reduce .[] as $x ...`. Inside `path()`, that internal navigation is
+/// subject to the same rule as one the user wrote: applied to a value the
+/// resolver is not tracking, it raises rather than producing a path. This
+/// is the [`navigation_element`] question for a builtin, and the answer
+/// feeds the same two error constructors.
+///
+/// **Membership is decided by jq's own implementation, not by whether the
+/// builtin "looks like" it reads a container**, and the two are not the
+/// same set: `sort`, `min`, `max`, `group_by`, `to_entries`, `keys`,
+/// `has`, `getpath` and `tostream` all read their input and all answer
+/// `[]` here, because jq implements them in C, where no jq-level
+/// navigation ever happens. `unique` raises only because it is *defined*
+/// as `group_by(.) | map(.[0])`. Every entry below, and every one of those
+/// negatives, was captured live from jq 1.7.1 via
+/// `[path(([1] | X) | empty)]` — see `tests/jq_cli_tests.rs`'
+/// `test_jq_defined_navigating_builtins_raise_inside_path_2646` and its two
+/// boundary siblings, `test_only_jq_defined_navigators_raise_inside_path_2646`
+/// and `test_navigating_builtins_raise_regardless_of_input_type_2646`.
+///
+/// Two boundaries worth keeping in view when adding to this:
+///
+/// - **Arity changes the answer.** `any`/`any(cond)` iterate their input
+///   and raise; `any(gen; cond)` does not — it iterates `gen`, so
+///   `[1] | any(1;.)` answers `[]` while `[1] | any(true)` raises. If
+///   `any(.[];.)` raises, that is its *argument* navigating, which the
+///   resolver already handles on its own.
+/// - **Some raise against a container this function cannot name.**
+///   `unique`, `unique_by`, `map_values`, `with_entries` and `fromstream`
+///   raise against a *derived* value (`[1] | map_values(.)` reports
+///   `element 0 of [[1],[]]`, the `group_by`/`to_entries` intermediate),
+///   which a static answer here cannot produce. They are deliberately
+///   absent rather than reported with the wrong container — #2743.
+///   `nth(n)`, `reverse` and `indices(i)` are absent for a related reason:
+///   their element depends on an argument or on `length` (and `reverse`
+///   does not raise at all on an empty input), so they need a value this
+///   function is not given — #2744.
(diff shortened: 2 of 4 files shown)


newhoggy:
## Adversarial review pass — three regressions found and fixed

Worth stating up front: **CI was 29/29 green with 100% patch coverage while all three of
these were live.** A ~5,000-program random fuzz also found none of them; the first only
appears once the argument alphabet includes an *erroring* or *negative* argument.

### 1. `join($x)` / `flatten($x)` pre-empted jq's own error

Both are `$`-parameter builtins. jq's `builtin.jq`:

```
def join($x): reduce .[] as $i (null; ...);
def flatten($x): if $x < 0 then error("flatten depth must not be negative") else _flatten($x) end;
```

`def f($a): B` desugars to `a as $a | B`, so the argument is evaluated — and for
`flatten`, range-checked — **before** any iteration. Answering `Iterate` replaced that
error with a path one:

| program | jq 1.7.1 | `main` | first draft | now |
|---|---|---|---|---|
| `[path(([1] \| join(error("boom"))) \| empty)]` | `boom` | `boom` | **path error** | `boom` |
| `[path(([1] \| flatten(error("e"))) \| empty)]` | `e` | `e` | **path error** | `e` |
| `[path(([1] \| flatten(-1)) \| empty)]` | depth error | depth error | **path error** | depth error |

Both are out of the table now, under #2744 — whose stated rationale ("element depends on
an argument") covers them verbatim. **Bare `Flatten` stays**: it is `_flatten(-1)`, whose
argument is jq's own literal and cannot raise.

### 2. The rule leaked into yq mode

`map`, `any`, `all` and `flatten` are real yq builtins with their own oracle, and yq
v4.53.3 raises for none of them:

```console
$ printf 'a: [1,2]\nb: 3\n' | yq 'del(([1,2] | map(.)) | select(false))'
a: [1, 2]
b: 3                                                          # exit 0
$ ... | succinctly yq '...'      # first draft
Error: Invalid path expression near attempt to iterate through [1,2]   # exit 1
```

ADR-0018: the mode decides, never the input format. Now gated on
`S::TAG == EvalTag::Jq`. The rule is derived from how *jq* defines these in its own
`builtin.jq`; yq's implementation shares none of it.

When writing that test I listed `add` and `first` from memory — the oracle says `add`
is not a yq builtin at all (its lexer rejects it) and `first` errors there, so the test
asserts only the four verified-silent ones, with the boundary documented.

### 3. A doc comment asserting the opposite of the truth

It claimed the resolver already catches a navigating *argument* like `any(.[];.)`. It
does not — `last(f)`, `isempty(f)`, `any(gen;cond)`, `all(gen;cond)` and `INDEX(gen;f)`
all stay silent where jq raises (pre-existing; `first(f)`/`limit(n;f)`/`nth(n;f)` do have
the arms). Now stated accurately, and filed as #2746.

### The coverage gap that let two of these ship

The table had assertions for **4 of its 13 entries**, and `Join`/`FlattenDepth` — the two
that turned out wrong — had none. Every entry is now asserted, plus the `$`-parameter
boundary and the yq-mode gate. 100% patch coverage did not catch this, because the
uncovered thing was a *table row*, not a line.

### Confirmed clean by the review

- **No snapshot problem.** `$x`-frozen / snapshot-marked branches were probed across
  `reduce`/`foreach`/`getpath([])` shapes and match jq exactly; omitting
  `!snapshot.is_marked()` is correct here, since unlike `recurse` these builtins' first
  act *is* navigation.
- **No performance concern** — the check runs only under `!trackable`, only after
  `navigation_element` returns `None`, and an untracked leaf raises immediately, so the
  path cannot be hot. Interleaved timing showed no difference.
- **All write routes** (`del`, `=`, `|=`, `+=`, `delpaths([path(…)])`) match jq
  byte-for-byte for every table entry, and a 130-case tracked-input sweep confirms the
  guard never fires where the input *is* tracked.

### Re-verified after the fixes

240-case sweep across 30 builtins × 8 input shapes: 214 match. All 26 remaining are
deliberate — `walk`-on-object (#2743), the now-deferred `join`/`flatten(N)` (#2744), and
`flatten(-1)`'s message (#2747, newly filed, pre-existing). Every one of those 26 is
byte-identical to `main` except `walk` on arrays, which is the intended fix.

Also updated `docs/compliance/jq/limitations.md`, which still listed this as an open
divergence and now records the fix plus the three groups still deferred.

### Further pre-existing gaps filed

#2746 (argument-side navigation in `last(f)`/`isempty(f)`/`any(gen;cond)`/`INDEX(gen;f)`)
and #2747 (`flatten`'s negative-depth message). The review also noted `INDEX(f)` and
`transpose` as absent table entries and `reduce`/`foreach` *sources* bypassing the check
entirely — both now named in `builtin_navigation`'s own prose so the derivation does not
read as exhaustive, and rolled into #2744/#2746.

Issue #2389, opened 2026-09-04
jq: ~40 more wrong-arity call sites in try_parse_builtin still raise raw parse errors, not jq's arity error


newhoggy:
**Severity: Low** (identical shape to #2237's own scope, error-message quality only)

## Summary

#2237 (PR #2387) extended #2110's fix to 11 single-required-arg builtins
(`has`/`select`/`env`/`min_by`/`max_by`/`in`/`ltrimstr`/`rtrimstr`/`startswith`/
`endswith`/`sort_by`) and the 7-member dedicated-parse-function family
(`range`/`limit`/`until`/`while`/`repeat`/`first`/`last`) -- covering every builtin
that issue itself named a live repro for. A further ~40 call sites inside
`try_parse_builtin` (`src/jq/parser.rs`) share the identical inline
`self.expect('(')?`/`self.parse_expr()?`/`self.expect(')')?` shape but are
unconverted, so a wrong-arity call to any of them still raises a raw parser
rejection instead of jq's own "X/N is not defined" compile error.

## Why this wasn't folded into #2237

Doing all ~40 sites safely is a materially larger undertaking than #2237's own
scope: two real regressions were caught during that PR's own testing/review from
sites that *looked* like simple copies of an already-verified pattern but had
site-specific subtleties (see #2387's own commit history) --

- `limit`'s deliberate restriction against a comma-valued `n` (real jq's `$n`
  per-output fanout convention isn't implemented here) was silently bypassed by an
  overly-broad rewind checkpoint.
- The generic `try_parse_builtin` fallback's own postfix-application fix broke yq
  mode's merge-flag scanning (`.a *=x .b`), since real yq's lexer unconditionally
  rejects non-flag content there in a way jq's grammar doesn't.

At least two of the remaining ~40 sites are flagged as likely carrying the *first*
kind of subtlety already: `skip(n; expr)` and `nth(n; ...)` both use
`self.parse_pipe_no_comma()` for their own `n` argument, with the identical
"comma-valued `n` is a deliberate restriction, not a raw arity issue" comment
`limit`'s own code carries. Each site converted needs the same
live-oracle-verified, comma-restriction-aware checkpoint `parse_required_single_arg`
(for the 11 already-converted single-arg builtins) established, not a blind
copy-paste.

## Suggested fix direction

Same helpers #2237 already introduced (`parse_required_single_arg` for
`try_parse_builtin`-routed single/fixed-arity builtins, a variant for any
[...]

Issue #2237, opened 2026-09-02
jq: wrong-arity calls to has/select/env/range/limit/... still report a raw parse error, not #2110's fix


newhoggy:
**Severity: Low** (identical to #2110's own scope: this only fires when a program is
already invalid, calling a builtin with an arity it never had; no valid program's
output is affected).

## Summary

Follow-up to #2110 (fixed in PR #2233). That fix reports a wrong-arity call to a
zero-arity builtin (`length(1)`, `not(1)`, `null(1)`, ...) as jq's own compile-time
"X/N is not defined" instead of a raw parser rejection of the `(`. It's implemented as
one shared helper (`zero_arity_or_wrong_arity_call`, `src/jq/parser.rs`) called from the
small number of sites that recognize a *bare* zero-arity keyword directly.

Two other categories of builtin share #2110's exact symptom (a wrong-arity call gives a
raw, confusing parser error instead of jq's own "X/N is not defined") but were found
during #2233's own code review to sit outside the one call site the fix covers, and
were deliberately left unaddressed as a scope decision, not an oversight:

## Category 1: builtins that parse their own required argument(s) inline

`has`, `select`, `env`, and by the same pattern `min_by`/`max_by`/`sort_by`/`ltrimstr`/
and many other `try_parse_builtin` entries call `self.expect('(')` directly and
propagate a raw `ParseError` via `?` *before* `zero_arity_or_wrong_arity_call` ever
runs -- they never reach the shared fix, even though they live inside the one function
(`try_parse_builtin`) #2233's own commit message credited with covering "all ~126
branches."

```console
$ echo null | succinctly jq 'has'
jq: compile error: parse error at position 3: expected '(', found end of input
$ /usr/bin/jq -n 'has'
jq: error: has/0 is not defined at <top-level>, line 1:
has
jq: 1 compile error

$ echo null | succinctly jq 'has(1;2)'
jq: compile error: parse error: expected ')', found ';'
$ /usr/bin/jq -n 'has(1;2)'
jq: error: has/2 is not defined at <top-level>, line 1:
has(1;2)
jq: 1 compile error
```

## Category 2: keywords with their own dedicated parse function

`range`, `limit`, `until`, `while`, `repeat` sit in the same `parse_primary_inner` match
arm as `null`/`true`/`false`/`not` (#2110's own fixed sites) but dispatch to dedicated
`parse_range_expr`/`parse_limit_expr`/etc. functions that unconditionally
[...]

PR #2750, opened 2026-09-10
fix(jq): report "not defined" for the last 13 wrong-arity builtin sites

### LINKED ISSUE #2573: jq: ~32 more wrong-arity call sites in try_parse_builtin remain after #2389's first batch
## Background

#2389 asked for a per-site survey (not a blind mechanical sweep) of the ~40 remaining `try_parse_builtin` (`src/jq/parser.rs`) call sites still using the old inline `self.expect('(')?`/`self.parse_expr()?`/`self.expect(')')?` shape instead of the shared `parse_required_single_arg` helper (#2237). That survey (done as part of implementing #2389) classified all 44 inline sites found at the time:

- **27 SIMPLE-COMMA-OK** — genuinely single-required-arg, no complications. #2389 converted the first 12 of these: `map`, `map_values`, `with_entries`, `walk`, `del`, `delpaths`, `getpath`, `group_by`, `unique_by`, `join`, `contains`, `isempty`.
- **16 MULTI-ARG** — fixed 2-arg or optional-second-arg shapes, out of scope for `parse_required_single_arg` as it stands (`IN`, `splits`, `split`, `match`, `capture`, `sub`, `gsub`, `scan`, `test`, `INDEX`, `setpath`, `pow`, `atan2`, `skip`, `nth`, `at_position`). Two of these (`skip`, `nth`) use `parse_pipe_no_comma_with_booleans` for their `n` argument — a deliberate comma restriction that must not be silently widened by any future conversion attempt.
- **1 COMPLICATED** — `strenv(VAR)` parses its argument with `self.parse_ident()`, not `self.parse_expr()`; `parse_required_single_arg` hard-codes `parse_expr()` internally, so this site needs either a `parse_ident`-based sibling helper or to stay hand-rolled (see `env(VAR)`'s own existing hand-rolled rewind logic just above it as the template).

## What's left

The remaining 15 SIMPLE-COMMA-OK sites #2389 didn't get to (each still needs its own live-oracle wrong-arity verification before converting, per #2237's own established discipline — two real regressions were caught there from sites that looked like simple copies but weren't):

`bsearch`, `pick`, `omit`, `strftime`, `strptime`, `tz`, `load`, `at_offset`, `fromstream`, `truncate_stream`, `indices`, `rindex`, `index`, `inside`, `isvalid`

(Line numbers will have shifted after #2389's own edits — re-locate each by `self.matches_keyword("...")` rather than trusting a cached line number.)

Plus the two structurally different cases the original survey flagged as needing their own approach rather than the plain helper:
- `strenv(VAR)` — needs a `parse_ident`-aware variant.
- The 16 MULTI-ARG sites — would need a genuinely different helper (2-required-arg, or required-plus-optional), not `parse_required_single_arg` itself.

## Suggested next step

Same discipline as #2389 and #2237: convert the SIMPLE-COMMA-OK sites first (lowest risk, mechanical once individually verified), each confirmed live against the pinned jq 1.7.1 oracle for both the no-args and too-many-args wrong-arity shapes, plus a shadowing (`def <name>(x): ...`) sanity check. Consider the multi-arg family and `strenv` as separate follow-up scope, since they need new helper shapes rather than reusing the existing one.

newhoggy:
## Summary

Finishes the batch #2389 started. Thirteen `try_parse_builtin` sites still carried
#2237's inline `self.expect('(')?`/`self.parse_expr()?`/`self.expect(')')?` shape, which
hard-errors on a wrong-arity spelling instead of rewinding — so succinctly reported its
own parse error where jq reports the arity:

```console
$ echo '[1]' | jq -c 'bsearch(1;2)'
jq: error: bsearch/2 is not defined at <top-level>, line 1:
$ echo '[1]' | succinctly jq -c 'bsearch(1;2)'      # before
jq: compile error: parse error at position 9
```

**All 26 wrong-arity spellings** (0-arg and 2-arg across the thirteen names) diverged
from jq 1.7.1 before this. All 26 match after.

The rewind is the substance, not just the wording — a wrong-arity spelling now falls back
to an ordinary call, so a user `def` at that arity resolves rather than the parse dying
first:

```console
$ echo '[1]' | jq -c 'def bsearch(a;b): "shadow"; bsearch(1;2)'
"shadow"
```

Converted: `bsearch`, `pick`, `omit`, `strftime`, `strptime`, `tz`, `load`, `at_offset`,
`fromstream`, `truncate_stream`, `indices`, `rindex`, `index`.

## On the per-site discipline

#2389 cautioned that this is a survey, not a sweep — two earlier regressions came from
sites that looked like simple copies but were not. Each of the thirteen was checked
individually against the oracle before converting.

They all turned out to be genuine copies. The only variation is an optional
`reject_unless_jq_extensions` line, which keeps its position ahead of `keyword_start` —
matching the already-converted `getpath`, which sits two lines below `truncate_stream`
and provided the exact gated template.

Confirmed unchanged, not just assumed:

- **Correct-arity behaviour** for all thirteen, diffed against a `main` build — 13/13
  byte-identical.
- **The `--jq-extensions` gate** in yq mode for the eight gated names (still rejected
  without the flag, still accepted with it), and that the five ungated ones
  (`pick`, `omit`, `tz`, `load`, `at_offset`) stay ungated.
- **Site ordering**, which matters here (`rindex` is checked before `index`) — untouched.

Five of these names (`omit`, `tz`, `load`, `at_offset`, and `bsearch` in yq mode) are
succinctly or yq extensions rather than jq builtins. jq still answers "not defined" for
*every* arity of a name it does not have, so the expectation is identical and the oracle
is still authoritative.

## Test plan

- [x] All 26 wrong-arity spellings diffed against `/usr/bin/jq` 1.7.1: 0/26 matched
      before, 26/26 after
- [x] All 13 correct-arity spellings diffed against a `main` build: unchanged
- [x] New `test_remaining_single_arg_builtins_wrong_arity_reports_not_defined_2573`
      (all 26 spellings) and
      `test_remaining_single_arg_builtins_allow_a_shadowing_def_2573` (the rewind, all 13)

Fixes #2573

### DIFF
--- a/src/jq/parser.rs
+++ b/src/jq/parser.rs
@@ -3807,14 +3807,11 @@ impl<'a> Parser<'a> {
         }
         if self.matches_keyword("inside") {
             self.reject_unless_jq_extensions("inside")?;
+            let keyword_start = self.pos;
             self.consume_keyword("inside");
-            self.skip_ws();
-            self.expect('(')?;
-            self.skip_ws();
-            let b = self.parse_expr()?;
-            self.skip_ws();
-            self.expect(')')?;
-            return Ok(Some(Builtin::Inside(Box::new(b))));
+            return Ok(self
+                .parse_required_single_arg(keyword_start)?
+                .map(|b| Builtin::Inside(Box::new(b))));
         }
 
         // Phase 5: Array Functions
@@ -4057,14 +4054,11 @@ impl<'a> Parser<'a> {
 
         // isvalid(expr)
         if self.matches_keyword("isvalid") {
+            let keyword_start = self.pos;
             self.consume_keyword("isvalid");
-            self.skip_ws();
-            self.expect('(')?;
-            self.skip_ws();
-            let expr = self.parse_expr()?;
-            self.skip_ws();
-            self.expect(')')?;
-            return Ok(Some(Builtin::IsValid(Box::new(expr))));
+            return Ok(self
+                .parse_required_single_arg(keyword_start)?
+                .map(|expr| Builtin::IsValid(Box::new(expr))));
         }
 
         // Phase 10: Path Expressions
--- a/src/jq/parser.rs
+++ b/src/jq/parser.rs
@@ -3936,37 +3936,28 @@ impl<'a> Parser<'a> {
         }
         if self.matches_keyword("indices") {
             self.reject_unless_jq_extensions("indices")?;
+            let keyword_start = self.pos;
             self.consume_keyword("indices");
-            self.skip_ws();
-            self.expect('(')?;
-            self.skip_ws();
-            let s = self.parse_expr()?;
-            self.skip_ws();
-            self.expect(')')?;
-            return Ok(Some(Builtin::Indices(Box::new(s))));
+            return Ok(self
+                .parse_required_single_arg(keyword_start)?
+                .map(|s| Builtin::Indices(Box::new(s))));
         }
         // Check index before rindex since rindex contains "index"
         if self.matches_keyword("rindex") {
             self.reject_unless_jq_extensions("rindex")?;
+            let keyword_start = self.pos;
             self.consume_keyword("rindex");
-            self.skip_ws();
-            self.expect('(')?;
-            self.skip_ws();
-            let s = self.parse_expr()?;
-            self.skip_ws();
-            self.expect(')')?;
-            return Ok(Some(Builtin::Rindex(Box::new(s))));
+            return Ok(self
+                .parse_required_single_arg(keyword_start)?
+                .map(|s| Builtin::Rindex(Box::new(s))));
         }
         if self.matches_keyword("index") {
             self.reject_unless_jq_extensions("index")?;
+            let keyword_start = self.pos;
             self.consume_keyword("index");
-            self.skip_ws();
-            self.expect('(')?;
-            self.skip_ws();
-            let s = self.parse_expr()?;
-            self.skip_ws();
-            self.expect(')')?;
-            return Ok(Some(Builtin::Index(Box::new(s))));
+            return Ok(self
+                .parse_required_single_arg(keyword_start)?
+                .map(|s| Builtin::Index(Box::new(s))));
         }
         // INDEX(idx_expr) - build an object keyed by idx_expr from `.[]`
         // INDEX(stream; idx_expr) - build an object keyed by idx_expr from stream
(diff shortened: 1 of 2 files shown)


newhoggy:
## Adversarial review pass — one claim of mine was false

The conversion itself came back clean: **all thirteen were genuine copies**, verified
line-by-line (`parse_required_single_arg` is byte-equivalent to the inline shape, since
`expect()` does its own `skip_ws()` first) and empirically across 14 names × 14 spellings
— whitespace variants, comments before the paren, nested/`as`-bound arguments, `X()`,
`X(1,2)`, in pipelines. **Exactly 39 rows changed = 13 names × {arity 0, 2, 3}**, all
matching jq byte-for-byte, and nothing else moved. Prefix collisions are safe because
`matches_keyword` rejects an alphanumeric follower. yq mode: 14 names × 5 spellings × 2
flag states, **zero** differences vs `main`.

Three things were wrong around it, now fixed.

### 1. My "the rewind is the substance" claim was false — corrected

I asserted, here and on the issue, that a wrong-arity spelling now lets a `def` at that
arity shadow the builtin. **It already did on `main`**, via `try_parse_builtin`'s
pre-existing `Err`-arm retry (`retry_shadow_candidate_as_generic_call`):

```console
$ echo null | main       jq -c 'def bsearch(a;b): "shadow"; bsearch(1;2)'   # "shadow"
$ echo null | HEAD       jq -c 'def bsearch(a;b): "shadow"; bsearch(1;2)'   # "shadow"
```

So `test_remaining_single_arg_builtins_allow_a_shadowing_def_2573` **passed unchanged on
`main`** — a regression test that discriminated nothing — and its doc comment asserted a
change that never happened. Deleted, and replaced with an arity-**3** test, which does
discriminate:

| | jq 1.7.1 | `main` | HEAD |
|---|---|---|---|
| `bsearch(1;2;3)` | `bsearch/3 is not defined` | **parse error** | `bsearch/3 is not defined` |

The real substance of this change is the arity *report*, which the sibling test covers —
and it holds at every wrong arity, not just the two the first draft happened to try.

### 2. "The last 13" was 15

`inside` carries the identical inline shape and sits immediately below the
already-converted `contains`. It is a real jq builtin, so the oracle applies directly.
`isvalid` is the same shape (a succinctly extension; jq names an arity for a name it does
not have, so the expectation is the same). Both converted and verified at arities 0/2/3.

A scan of every `self.expect('(')?` in the file now reports **zero** remaining inline
single-required-arg sites — the class is closed, not merely reduced. (`nth`, `skip`,
`setpath`, `pow`, `atan2`, `at_position` are multi-arity and correctly untouched, as is
`strenv`'s `parse_ident` argument.)

### 3. A pre-existing DoS shape that this widens — filed as #2751

The `Ok(None)` rewind re-parses the whole argument subtree, and
`retry_shadow_candidate_as_generic_call`'s `shadow_retry_budget` — added *precisely* to
bound this — does not cover that path. Nesting wrong-arity calls is therefore exponential:

```
has(has(...has(1);2...);2)   depth 18:  succinctly ~2.5s,  jq instant   (~4x per level)
```

**Pre-existing, and identical on `main`** for `has` (2.47 s), `contains` (2.87 s) and
`getpath` (3.03 s) from the #2237/#2389 batches — so this does not change the security
posture, since plain `has` already reaches it with no flags. What this PR does is add 15
names to the affected set, some of them (`index`, `indices`, `pick`) likelier to nest in
real programs than `getpath`.

**Not fixed here on purpose.** The reviewer's one-line suggestion — route `Ok(None)`
through the same budget — is wrong: past the budget that path propagates a parse error,
which would make an ordinary wrong-arity call regress to exactly the divergence
#2237/#2389/#2573 exist to remove. The real fix is probably to build the generic call from
the already-parsed argument (as `wrap_shadowable_call` does on the success path) so no
reparse happens at all. Written up in #2751.

### Also noted (pre-existing, out of scope)

Real yq accepts extra arguments on its own builtins (`pick(1;2)` → `{}`, `omit(1;2)` →
`a: 1`) where succinctly raises a parse error, and yq's zero-arg message is
`'pick' expects 1 arg but received none`. Identical on `main`. And the comment at
`parser.rs:3944` (`// Check index before rindex since rindex contains "index"`) is
inverted relative to what the code correctly does — carried through untouched.

### Re-verified

Full suite (62 binaries), both clippy legs, fmt: green. 30 wrong-arity spellings (15
names × arity 0/2) plus 15 arity-3 spellings all match jq; all 15 correct-arity spellings
unchanged vs `main`.

Issue #2752, opened 2026-09-10
jq: (gen)? keeps generating after an element raises; jq stops the whole generator


newhoggy:
**Severity: Low**

## Summary

Postfix `?` applied to a parenthesised generator suppresses the error *and continues to
the next element*. jq stops the generator entirely at the first raise, so elements after
the failing one are never produced.

## Repro

Confirmed live against jq 1.7.1 and on `main` (`56926e912`):

```console
$ echo '{"a":[5,{"b":{}}]}' | jq -c 'path((.a[] | .b)? | .c[0:1])'
                                             # nothing, exit 0
$ echo '{"a":[5,{"b":{}}]}' | succinctly jq -c 'path((.a[] | .b)? | .c[0:1])'
["a",1,"b","c",{"start":0,"end":1}]          # exit 0
```

Element 0 (`5 | .b`) raises. jq's `?` catches it and the generator ends there, so element
1 is never reached. succinctly skips the failing element and carries on.

The order matters, which is what makes this the *continuation* question rather than the
prefix one: with the failing element **last**, both tools agree, because there is nothing
after it to disagree about.

```console
$ echo '{"a":[{"b":{}},5]}' | jq -c 'path((.a[] | .b)? | .c[0:1])'
["a",0,"b","c",{"start":0,"end":1}]
$ echo '{"a":[{"b":{}},5]}' | succinctly jq -c 'path((.a[] | .b)? | .c[0:1])'
["a",0,"b","c",{"start":0,"end":1}]          # agrees
```

## Why this is separate from #2132

#2132 is the mirror image: whether values produced *before* the error survive the `?`.
This is about values produced *after* it, which jq does not produce at all. A fix for one
does not settle the other, and this reproduces with no `range`/`while` cap involved.

## Why this is separate from #2680

Found while fixing #2680 (PR pending), which is about a *non*-`?` stage-1 failure
dropping its already-reached prefix. That fix is in `walk_pipe`; this reproduces
identically before and after it (verified on both builds), and is about `?`'s scope over
a generator rather than about buffering.

## Suggested fix direction

Establish first whether this is `?`-specific or general: check `try (…) catch` and a bare
generator raising mid-stream in non-path contexts, since jq's rule here ("an error ends
the generator") may already be implemented correctly elsewhere in the evaluator and only
the path walker's `optional` handling diverges. `step_into`'s
`Err(EvalEscape::Error(_)) if optional => return Ok(())` (`src/jq/eval.rs`) turns a
[...]

PR #2753, opened 2026-09-10
fix(jq): keep the prefix a failing nested path stage already reached

### LINKED ISSUE #2680: jq: path()'s DOM walker drops an output prefix jq keeps when a nested pipe stage errors

newhoggy:
## Summary

`walk_pipe` (`src/jq/eval.rs` — the DOM route `path()` takes for anything not
cursor-navigable) `?`-propagated stage 1's error *before* walking the positions that stage
had already reached, so those positions never made it through the remaining stages:

```console
$ echo '{"a":[{"b":{}},5]}' | jq -c 'path((.a[] | .b) | .c[0:1])'
jq: error (at <stdin>:1): Cannot index number with string "b"
["a",0,"b","c",{"start":0,"end":1}]          # emitted, exit 5

$ echo '{"a":[{"b":{}},5]}' | succinctly jq -c 'path((.a[] | .b) | .c[0:1])'   # before
jq: error (at <stdin>:1): Cannot index number with string "b"
                                             # nothing, exit 5
```

Positions reached before a failure come earlier in jq's generator order than the failure
itself, so the stage's `Result` is now held back, `reached` is walked through `rest`, and
only then does the error surface.

This is verbatim the shape `eval_generic.rs`'s twin `path_walk_pipe_generic` already used
— including its comment giving the reason: *never un-emit an output already produced*. So
`eval.rs`'s walker was the outlier against its own sibling, which is what made the fix
four lines.

## Why the parentheses were the whole difference

`out` is the **caller's** buffer. The flat spelling pushes its earlier positions straight
there, where they survive the error propagation — which is why it was already correct:

```console
$ echo '{"a":[{"b":{}},5]}' | succinctly jq -c 'path(.a[] | .b | .c[0:1])'
jq: error (at <stdin>:1): Cannot index number with string "b"
["a",0,"b","c",{"start":0,"end":1}]          # correct, before this change
```

Only a *nested* stage-1 pipe buffers its positions in a local `reached`, and that local
buffer was what got dropped. Both spellings are now asserted, so a future change cannot
fix one and lose the other.

## Verification

A 70-case sweep — 10 filter shapes (`path`, `[path(...)]`, `del`, `=`, `?`, flat vs
nested, longer nested stage) × 7 documents (failing element first / last / absent, empty
array, object-valued, `null`) — against `/usr/bin/jq` 1.7.1:

- **69 of 70 match.**
- **Two rows changed from `main`, both to jq's answer**; nothing else moved.
- The single remaining difference is **pre-existing**, byte-identical on `main`, and about
  `?`'s scope rather than buffering: `(gen)?` keeps generating after an element raises
  where jq ends the generator. Filed as #2752 with its own repro, including the detail
  that the two tools *agree* when the failing element is last — which is what identifies
  it as the continuation question rather than this one.

The boundary is pinned too, so the fix can't degrade into "always emit something": a
stage-1 failure that reached nothing still emits nothing, and a nested pipe that never
fails is unaffected.

## Test plan

- [x] 70-case sweep vs jq 1.7.1 (above), plus a diff of every row against a `main` build
      to separate the fix from pre-existing behaviour
- [x] New `test_path_keeps_the_prefix_a_failing_nested_stage_reached_2680` (nested, longer
      nested, and the flat control) and
      `test_path_emits_nothing_when_the_failing_stage_reached_nothing_2680` (both boundary
      directions)

Fixes #2680

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -38993,6 +38993,18 @@ fn walk_pipe<'v, S: EvalSemantics>(
     // Exactly the shape `eval_generic.rs`'s twin `path_walk_pipe_generic`
     // already uses, and for the reason its own comment gives: never un-emit
     // an output already produced.
+    //
+    // Two consequences worth knowing before touching this. The inner `?`
+    // makes a failure while walking `rest` for an *earlier* position win
+    // over `stepped` -- correct, since it comes first in generator order,
+    // and it changes the reported *message*, not just whether a prefix
+    // appears (`path((.a[] | .b) | .c[0:1])` on `{"a":[{"b":5},7]}` reports
+    // the `.c` failure, where propagating `stepped` first reported the `.b`
+    // one). And this function has no depth guard of its own: it relies on
+    // the `assert_value_tree_depth` at the top of `walk_path`, which it
+    // calls once per stage before recursing. The twin carries its own at
+    // entry -- an early return added *above* the `walk_path` call here would
+    // bypass the guard entirely.
     let mut reached = Vec::new();
     let stepped = walk_path::<S>(first, value, current_path, &mut reached, optional);
     for (path, val) in reached {
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -38975,12 +38975,32 @@ fn walk_pipe<'v, S: EvalSemantics>(
         return walk_path::<S>(first, value, current_path, out, optional);
     }
 
+    // #2680: the stage's `Result` is held back rather than `?`-propagated
+    // here. Positions it reached *before* failing come earlier in jq's
+    // generator order than its own error, so they still have to be walked
+    // through `rest` -- only then does the error surface. Propagating first
+    // dropped `reached` on the floor, and with it every path those positions
+    // would have named: `path((.a[] | .b) | .c[0:1])` on
+    // `{"a":[{"b":{}},5]}` printed the error alone, where jq prints
+    // `["a",0,"b","c",{"start":0,"end":1}]` and *then* the error.
+    //
+    // `out` is the caller's buffer, which is why the same pipe written flat
+    // (`path(.a[] | .b | .c[0:1])`) was already correct -- its earlier
+    // positions were pushed straight to `out` and survived the propagation.
+    // Only this local `reached` was lost, so only a *nested* (parenthesised)
+    // stage-1 pipe showed the bug.
+    //
+    // Exactly the shape `eval_generic.rs`'s twin `path_walk_pipe_generic`
+    // already uses, and for the reason its own comment gives: never un-emit
+    // an output already produced.
     let mut reached = Vec::new();
-    walk_path::<S>(first, value, current_path, &mut reached, optional)?;
+    let stepped = walk_path::<S>(first, value, current_path, &mut reached, optional);
     for (path, val) in reached {
+        // A failure *here* is earlier in generator order than `stepped`'s
+        // own, so it wins -- same as the twin.
         walk_pipe::<S>(rest, val, &path, out, optional)?;
     }
-    Ok(())
+    stepped
 }
 
 /// Take one path step: `component` names it, and the value evaluator decides
--- a/tests/jq_cli_tests.rs
+++ b/tests/jq_cli_tests.rs
@@ -43194,6 +43194,103 @@ fn test_path_emits_nothing_when_the_failing_stage_reached_nothing_2680() -> Resu
     Ok(())
 }
 
+/// #2680, the subtlest consequence: when stage 1 fails *and* a later stage
+/// fails for an earlier position, the later error wins -- it comes first in
+/// jq's generator order. The reported *message* therefore changed, not just
+/// whether a prefix appears: `main` reported stage 1's `"b"` error, jq and
+/// this report `rest`'s `"c"` one.
+///
+/// This is what the inner loop's `?` buys, and nothing pinned it. Captured
+/// live from jq 1.7.1.
+#[test]
+fn test_path_reports_the_earlier_failure_when_both_stages_fail_2680() -> Result<()> {
+    // `.a[0].b` is 5, so stage 1 succeeds there and `.c` fails on it;
+    // `.a[1]` is 7, so stage 1 itself fails there. `rest`'s failure is
+    // earlier, so it is the one reported.
+    let (stdout, stderr, code) = run_jq_full(
+        &["-c", "path((.a[] | .b) | .c[0:1])"],
+        Some(r#"{"a":[{"b":5},7]}"#),
+    )?;
+    assert_ne!(code, 0, "stdout {stdout:?}");
+    assert!(
+        stderr.contains(r#"Cannot index number with string "c""#),
+        "the later stage's error must win, got: {stderr:?}"
+    );
+    assert!(
+        !stderr.contains(r#"Cannot index number with string "b""#),
+        "stage 1's own error must not be the one reported: {stderr:?}"
+    );
+
+    // Stage 1 reaching two positions, the *second* of which fails in `rest`:
+    // the first still emits, then that failure is reported.
+    let (stdout, stderr, code) = run_jq_full(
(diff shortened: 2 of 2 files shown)


newhoggy:
## Adversarial review pass — clean on the risk that mattered

I flagged one worry above all: this path machinery is shared with `del`/`=`/`|=`, so
emitting a previously-dropped path could cause a **write that did not happen before**.
The review went after it specifically and it comes back clean:

- The write machinery is **untouched** — `del`/`=`/`|=`/`+=`/`//=`/literal `setpath` go
  through `set_path`/`set_path_steps`/`update_path`/`delete_at_path`/
  `value_after_components`, none of which this change reaches. `walk_path`'s only
  non-test callers are `builtin_path_on_owned` and `eval_generic.rs:19100`.
- **36 write-form probes** (`= |= += //= del() del(…[0:1]) delpaths([path(…)]) getpath` ×
  3 partial-failure documents) are byte-identical to `main` *and* to jq.
- ~150 three-way probes overall (jq / HEAD / a `git archive main` build): **no
  newly-introduced wrong answer**, and every HEAD-vs-jq difference found is byte-identical
  on `main`.

It also confirmed the fix makes the *two* `path()` routes agree — the cursor route
(`path_walk_generic`) already emitted the prefix, so `eval.rs` was the sole outlier — and
that the depth guard still bounds recursion (500 / 5 000 / 40 000 / 200 000-stage pipes all
report `nesting depth exceeds limit of 256`, no abort, timings identical to `main`).

### Three consequences it found that I had neither advertised nor tested — now pinned

All verified independently, all moving toward jq:

| | jq 1.7.1 | `main` | now |
|---|---|---|---|
| `path((.a[] \| .b) \| .c[0:1])` on `{"a":[{"b":5},7]}` — **which error** | `…string "c"` | `…string "b"` | `…string "c"` |
| `setpath(path((.a[] \| .b) \| .c); 9)` on `{"a":[{"b":{}},5]}` | `{"a":[{"b":{"c":9}},5]}` | *(no output)* | `{"a":[{"b":{"c":9}},5]}` |
| `first(path((.a[] \| .b) \| .c[0:1]))` — **exit code** | `0` + output | `5`, no output | `0` + output |

The first is the subtlest: the inner loop's `?` means a failure walking `rest` for an
*earlier* position wins over stage 1's own, which is right (it comes first in generator
order) but changes the reported *message*, not just whether a prefix appears. The third is
the largest user-visible effect — a truncating consumer now succeeds outright.

Also recorded on the function: `walk_pipe` has no depth guard of its own and relies on the
`assert_value_tree_depth` inside `walk_path`, which it calls once per stage. The generic
twin carries its own at entry, so an early return added *above* that call here would
silently bypass the guard.

### Perf

The loop now always runs, so HEAD does work `main` skipped — bounded, and strictly below
HEAD's own success path. Debug build, 200 000-wide fan-out: all-success 1.54s → 1.60s;
stage-1-fails-last 0.46s → 1.03s (jq: 0.17s). No asymptotic change, and the success path
is untouched.

### Two pre-existing gaps filed rather than fixed

- **#2754** — `reduce` does not consume a source's partial prefix where `foreach` does.
  The messages are the tell: jq reports the UPDATE's `setpath` failure (it folded the
  prefix), succinctly reports the source's walk error (it did not). Reproduces with the
  *flat* spelling, so unrelated to this fix.
- **#2752 needs re-triaging from Low.** Its `(gen)?` scope error performs **extra writes**,
  not merely extra output — `del((.a[] | .b)? | .c)` deletes a key jq leaves alone, and
  `= 9` writes one. Identical on `main`, so pre-existing, but that is the
  silent-extra-write class rather than the fidelity class. Commented there with repros.

### Re-verified after the additions

Full suite (62 binaries), both clippy legs, fmt: green. Five `_2680` tests now — the two
original (nested/flat prefix, and both boundary directions) plus the three above.

Issue #2761, opened 2026-09-11
jq: `..` on an untracked input inside path() says "with result", jq says "near attempt to iterate through"


newhoggy:
**Severity: Low** (message-only; both refuse, same exit code)

## Repro

Confirmed live against jq 1.7.1 and on `main` (`f2e4a1e5e`):

```console
$ echo null | jq -c 'path(1 | .. | empty)'
jq: error (at <stdin>:1): Invalid path expression near attempt to iterate through 1
$ echo null | succinctly jq -c 'path(1 | .. | empty)'
jq: error (at <stdin>:1): Invalid path expression with result 1
```

Same for `path(1 | [..] | empty)` once #2689 (PR pending) makes the bracketed form refuse
at all — it inherits this wording from the bare-stage arm.

## Root cause

`..` is `recurse`, whose first act on its input is `.[]?`. The `RecursiveDescent`/`Recurse`
family's shared untracked guard (`recurse_untracked_error`, `src/jq/eval.rs`) raises the
generic `invalid_path_expression(value)` ("with result"), where a bare `.[]` on the same
untracked input raises `invalid_path_expression_near_iterate(value)` ("near attempt to
iterate through"). jq reports the iterate wording because that is the instruction that
actually failed.

## Suggested fix direction

Change `recurse_untracked_error` to build `invalid_path_expression_near_iterate` instead.
It is a single definition shared by all four spellings (its own doc comment says so), so
the change is one line — but check whether any of the four spellings should genuinely keep
the generic wording in jq (e.g. `recurse(f)` with an `f` that does not iterate), since
the shared helper currently cannot distinguish them.

Found via the investigation for #2689.

newhoggy:
🔧 Claimed by agent `Amanda` — branch `issue-2761-recurse-path-error`

Issue #2760, opened 2026-09-11
jq: and/or operands and unary minus on an untracked input inside path() are opaque, so navigation inside them never refuses


newhoggy:
**Severity: Medium** (the `and`/`or` half is accept-where-jq-refuses; unary minus is
wrong-wording only)

## Summary

#2689 (PR pending) fixed `[f]` on an untracked input inside `path()` by resolving the
inside of the brackets instead of evaluating it by value. Its investigation found that
`and`/`or` operands and unary minus have the identical gap: jq keeps path tracking live
inside them, succinctly treats the whole expression as one opaque value.

## Repro

Confirmed live against jq 1.7.1, on `main` (`f2e4a1e5e`), input `{"a":false}`:

```console
$ jq -c 'path(. as {a:$v0} | (.a and true) | empty)'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of {"a":false}
$ succinctly jq -c 'path(. as {a:$v0} | (.a and true) | empty)'
                                             # nothing, exit 0

$ jq -c 'path(. as {a:$v0} | (true and .a) | empty)'    # same error
$ jq -c 'path(. as {a:$v0} | (.a or true) | empty)'     # same error
# succinctly: nothing, exit 0, for both

$ jq -c 'path(. as {a:$v0} | -(.a) | empty)'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of {"a":false}
$ succinctly jq -c 'path(. as {a:$v0} | -(.a) | empty)'
jq: error (at <stdin>:1): boolean (false) cannot be negated
```

The `del()`/`=` forms of the `and`/`or` rows are the damaging version — a refused edit
reported as a successful no-op — same as #2689 and #2646.

## Root cause

Same class as #2689: `Expr::And`/`Expr::Or`/`Expr::Negate` have no arm in
`resolve_node_sink` (`src/jq/eval.rs`), so on an untracked input they fall to
`resolve_leaf`'s catch-all, which evaluates by value and never inspects the operands'
navigation. jq's `gen_and`/`gen_or`/negation run their operands with tracking live, the
way `gen_collect` does for `[f]`.

## Why not folded into #2689

Each needs its own oracle bracket rather than a copy of the `[f]` arm. In particular
`and`/`or` **short-circuit**: `(false and .a)` never evaluates `.a`, so only the operand
that actually runs is tracked, and the refusal has to follow evaluation order rather than
[...]

Issue #2759, opened 2026-09-11
jq: a TrackedVar inside [...] on an untracked input can't re-establish the register, so #2689's arm has to fall back to the catch-all


newhoggy:
**Severity: Low** (a deliberate scope limit of #2689's fix, not a wrong answer)

## Summary

#2689 (PR pending) made `path()` resolve the *inside* of an array constructor when its
input is untracked, so `[.a]` on a constructed value refuses the way jq does. The new
`Expr::Array` arm in `resolve_node_sink` (`src/jq/eval.rs`) deliberately **does not fire
when the brackets contain a `TrackedVar`**, falling back to the old evaluate-by-value
catch-all for that case. This issue records why, and what the complete fix needs.

## The gap

A `$var` whose frozen value is still the live register *re-establishes* tracking — it is
not navigation, it is the register (#1573). That comparison happens in `resolve_seq_stage`,
against the incoming branch's `carried_register`. `resolve_node_sink` — where the new arm
lives, and the only place `[inner]` can be intercepted for an arbitrary `inner` — is not
handed that register. Resolving `inner` from there with `trackable: false` therefore sees
`$v` as just another untracked value:

```console
$ echo '{"a":false}' | jq -c 'path(. as {a:$v0} | [$v0.b?] | empty)'
                                                     # nothing, exit 0
# with the arm applied unconditionally:
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "b" of false
```

That is an over-refusal, so the arm bails out on any `TrackedVar` inside the brackets. The
catch-all evaluates `[$v0.b?]` by value, which matches jq on every shape found — but for
the wrong reason (it never checks the navigation at all), so a marker that re-establishes
and *then* navigates illegitimately inside the brackets would still be accepted.

## What the complete fix needs

Either thread the register into `resolve_node_sink` (a signature change with many call
sites), or intercept `Expr::Array` at the **stage** level in `resolve_seq_stage`, where
`carried_register` is in scope, and resolve `inner` through `resolve_seq_from_seed` with a
seed branch carrying it — that also gives nested `[[...]]` the register for free. The
stage-level placement was considered for #2689 and deferred: `resolve_seq_stage`'s register
placement (`reestablishes_register`/`carry_register`) is the most intricate logic in the
resolver, and #2689's reported shapes needed none of it.

## Related

- #2689 — the fix this is scoped out of.
- #1573 / #2042 / #2649 — the register model, marker certification, and destructuring's own
  register moves.

[...]

Issue #2649, opened 2026-09-08
jq: destructuring moves jq's path register — path(. as {a:$q} | $q) is ["a"] and path(. as {a:$q} | .a) raises


newhoggy:
**Severity: Low** (refuse-only on one row; a differently-worded refusal on the other — no
accept-where-jq-refuses)

## Summary

jq's destructuring bind (`as {a:$q}`) moves jq's own path register onto the *matched field*,
unlike plain `as $x`, which leaves the register on the ambient input. Real jq 1.7.1 therefore
resolves `path(. as {a:$q} | $q)` to `["a"]` and *refuses* `path(. as {a:$q} | .a)` (since `.a`
navigated off the still-root-anchored register no longer identifies the matched register).
succinctly gets both wrong: it refuses the first row too (jq accepts it), and raises a
different-shaped error on the second.

## Repro (jq 1.7.1 vs succinctly jq, on `{"a":[1,2,3]}`)

```console
$ echo '{"a":[1,2,3]}' | jq -c 'path(. as {a:$q} | $q)'
["a"]
$ echo '{"a":[1,2,3]}' | succinctly jq -c 'path(. as {a:$q} | $q)'
jq: error (at <stdin>:1): Invalid path expression with result [1,2,3]

$ echo '{"a":[1,2,3]}' | jq -c 'path(. as {a:$q} | .a)'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of {"a":[1,2,3]}
$ echo '{"a":[1,2,3]}' | succinctly jq -c 'path(. as {a:$q} | .a)'
jq: error (at <stdin>:1): Invalid path expression with result [1,2,3]
```

Both exit 5 in both tools on the second row, but real jq's message names the register value
that made it invalid (`{"a":[1,2,3]}`, the root register — destructuring never moved it to
`.a`); succinctly's message instead names `[1,2,3]` — the value `.a` itself resolves to —
suggesting succinctly's register tracking doesn't move through the pattern bind the way jq's
does, landing on a different (also-wrong) answer for a different reason on each row.

## Root cause

`each_as_pattern`/`Expr::AsPattern` (`src/jq/eval.rs`) evaluates a destructuring bind's body,
but the path/register bookkeeping `resolve_node`'s `Expr::TrackedVar` arm relies on
(`register_identical`, value-equality-plus-snapshot) is built around plain `as $x`'s single
substitution; nothing moves the register onto the *matched sub-field* the way jq's own
[...]

PR #2762, opened 2026-09-11
fix(jq): resolve the inside of [f] on an untracked input inside path()

### LINKED ISSUE #2689: jq: an untracked-navigation + `?`-swallowed stage after `as` fabricates where jq refuses
**Severity: Medium** (fabricate direction — succinctly accepts and silently succeeds/no-ops where jq raises exit 5; confirmed via `del()`, not just `path()`)

## Summary

An untracked construction stage (`[.a]`) feeding a `?`-caught error, downstream of an `as`-bind, makes `path()`/`del()`/assignment silently succeed where real jq raises `Invalid path expression`. Reproduces with both a plain `as` chain used in path position and a destructuring `as {a:$v0}` bind, but only the destructuring form actually diverges from jq — the equivalent plain-`as` query agrees with jq (exit 0 on both sides).

## Repro (jq 1.7.1 vs succinctly jq, on `{"a":false}`)

```console
$ echo '{"a":false}' | jq -c 'path(. as {a:$v0} | [.a] | ($v0.b?) as $v1 | $v1)'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of {"a":false}
$ echo '{"a":false}' | succinctly jq -c 'path(. as {a:$v0} | [.a] | ($v0.b?) as $v1 | $v1)'
(nothing printed, exit 0)

$ echo '{"a":false}' | jq -c 'del(. as {a:$v0} | [.a] | ($v0.b?) as $v1 | $v1)'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of {"a":false}
$ echo '{"a":false}' | succinctly jq -c 'del(. as {a:$v0} | [.a] | ($v0.b?) as $v1 | $v1)'
{"a":false}
```

Confirmed against both the pinned oracle (`/usr/bin/jq` 1.7.1) and Homebrew `jq` 1.8.2.

For contrast, the plain-`as` equivalent (no destructuring) agrees with jq — jq itself does **not** error here:

```console
$ echo '{"a":false}' | jq -c '.a as $v0 | [.a] | ($v0.b?) as $v1 | $v1'
(nothing printed, exit 0)   # jq and succinctly both exit 0
```

So jq's error is specific to the destructuring bind reaching a later stage through an untracked construction + `?`-swallowed error — not a general "any `as`-bind reaching this shape" rule.

## Root cause (hypothesis)

Likely the same underlying gap as the pre-existing (non-destructuring) fabrication independently found during #2649's review: `resolve_terminal`'s eager per-branch check never fires when an untracked navigation feeds a `?`-filtered stage that produces **zero** branches (`[.a] | ($v0.b?)` — indexing a boolean with `.b` errors, `?` swallows it, and the pipeline produces nothing downstream). With no branch ever reaching `resolve_terminal`, the violation is never raised, so the whole `path()`/`del()` call falls through to a quiet empty/no-op result instead of surfacing the refusal jq computes eagerly at compile/walk time.

## Confirmed not a regression

Reproduces identically on `main` at the #2649 merge-base (`4e5975946`, built directly) — i.e. before any of #2649's destructuring-register work landed. #2649 neither introduces nor fixes this; it's an independent, pre-existing gap that happens to be reachable through destructuring syntax.

## Related

- #2649 — the destructuring path-register fix; its own `docs/compliance/jq/limitations.md` entry claims the arm's residue is "refuse-only," which this repro falsifies for this specific shape (via a mechanism outside the new arm itself, per the above).
- The `[1] | first`-shaped variant of this same "untracked navigation + `?`-filtered zero-result branch" root cause, independently found during #2649's code review (same `resolve_terminal` suspect).

ADR-0018: jq mode only, rule 3 fidelity gap (no rule-4 justification to refuse this shape; real jq's behaviour here is a readable, consistent error).

newhoggy:
## Summary

jq's `[f]` (`gen_collect`) runs `f` with path tracking **live** — unlike `{k:f}`, `if f`,
`select(f)`, `try f`, an `as` source, or `"\(f)"`, which all suspend it. So an INDEX/EACH
inside the brackets on a value that is not `value_at_path` raises exactly as it would as a
bare stage.

`Expr::Array` had no arm in `resolve_node_sink`. On an untracked input it fell to
`resolve_leaf`'s catch-all and was evaluated **by value**: the `.a` inside succeeded, the
array came back untracked, and a continuation that emitted nothing left nothing for
`resolve_terminal` to object to.

```console
$ echo null | jq -c 'path(1 | [.[]?] | empty)'
jq: error (at <stdin>:1): Invalid path expression near attempt to iterate through 1
$ echo null | succinctly jq -c 'path(1 | [.[]?] | empty)'     # before
                                                              # nothing, exit 0
```

### The issue's framing was wrong, usefully

#2689 reported this through a destructuring `. as {a:$v0}` and identified the destructuring
as load-bearing. It isn't — it only *supplies* an untracked ambient value, and the shape
above has no `as` at all. The destructuring's own register move (#2649) was already
correct. The damaging form is the write:

```console
$ echo '{"a":false}' | jq -c 'del(. as {a:$v0} | [.a] | ($v0.b?) as $v1 | $v1)'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of {"a":false}
$ echo '{"a":false}' | succinctly jq -c 'del(. as {a:$v0} | [.a] | ($v0.b?) as $v1 | $v1)'   # before
{"a":false}                                                    # exit 0 — a refused edit reported as success
```

## How

The new arm resolves `inner` instead of evaluating it, which is what surfaces the
navigation: the inner `Field`/`Index`/`Iterate` arms already raise jq's exact wording
against an untracked input, and a literal, `.`, or an empty constructor pass straight
through. jq drains the whole collect, so the inner bound is unlimited whatever the outer
`keep` — `keep` governs how many *arrays* the consumer wants, and there is one. Same
accept-where-jq-refuses shape as #2646, on a leaf its static table cannot express because
`inner` is arbitrary code.

### Two deliberate limits

- **Keyed on `!trackable`.** A trackable input (`path([.a] | empty)`) and the plain-`as`
  control from the issue (`path(.a as $v0 | [.a] | ...)`, which jq accepts) still take the
  catch-all, so the seq-level register handling (`cannot_move_register`'s own `Array` arm)
  is untouched.
- **Bails out when the brackets contain a `TrackedVar`.** A marker can *re-establish*
  tracking — `$v` whose frozen value is still the live register is not navigation
  (#1573) — but that comparison happens one level up in `resolve_seq_stage`, against a
  `carried_register` this function is not handed. My first cut without this guard resolved
  `[$v0.b?]` with `trackable: false`, saw `$v0` as just another untracked value, and
  refused `.b` where jq accepts. So the marker case keeps the catch-all exactly as before;
  #2759 records what the complete fix needs (thread the register, or intercept at the
  stage level — deferred because `resolve_seq_stage`'s register placement is the most
  intricate logic in the resolver and none of the reported shapes needed it).

jq mode only (ADR-0018): real yq's lexer rejects every shape that reaches here, so there
is no yq oracle and yq mode is untouched.

## Verification

**102-case sweep** — 17 filter shapes × 6 documents — against `/usr/bin/jq` 1.7.1, with
every row also diffed against a `main` build:

- **102 of 102 match jq.**
- **37 rows changed from `main`, every one to jq's answer** — including the `del()` and
  `= 1` write forms, and `path(. as {a:$v0} | [.a][0] | empty)`, which moved from a
  *wrong-wording* error to jq's exact message.
- Zero rows changed away from jq.

The boundary is pinned in both directions: a trackable input, the plain-`as` control,
literal/identity/empty constructors, an object constructor (tracking suspended in jq), and
a marker inside the brackets all stay accepted.

### Sibling gaps surfaced and filed, not fixed here

- **#2760** — `and`/`or` operands and unary minus have the identical opacity
  (`path(. as {a:$v0} | (.a and true) | empty)` is silently accepted; jq refuses). Needs
  its own oracle bracket because `and`/`or` short-circuit, so only the operand that runs
  is tracked.
- **#2761** — `..` on an untracked input says "with result" where jq says "near attempt
  to iterate through"; pre-existing on the bare-stage arm, now also visible inside `[..]`.

## Test plan

- [x] 102-case sweep vs jq 1.7.1 and vs `main` (above)
- [x] Existing #2649 destructuring matrix, #1573/#2042 register tests, and the #2646 tests
      all pass unchanged
- [x] New: `test_path_refuses_navigation_inside_array_on_untracked_input_2689` (7 shapes,
      including the no-`as` root cause and nested/comma/pipe/postfix insides),
      `test_write_refuses_navigation_inside_array_on_untracked_input_2689` (the `del`/`=`
      forms), and `test_array_on_untracked_input_boundary_2689` (8 must-stay-accepted rows)

Fixes #2689

### DIFF
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -27963,12 +27963,11 @@ fn resolve_node_sink<'a, S: EvalSemantics>(
         // static table cannot express because `inner` is arbitrary code.
         //
         // Resolving `inner` instead of evaluating it is what surfaces that
-        // navigation: the inner `Field`/`Index`/`Iterate`/`RecursiveDescent`
-        // arms already raise jq's exact wording against an untracked input,
-        // and a `TrackedVar` marker, a literal or `.` inside the brackets
-        // pass straight through, so `[5]`, `[$v]`, `[$v.b?]` and `[.]` stay
-        // accepted. jq drains the whole collect, so the inner bound is
-        // unlimited whatever the outer `keep` -- `keep` governs how many
+        // navigation: the inner `Field`/`Index`/`Slice`/`Iterate` arms raise
+        // jq's exact wording against an untracked input, and a literal, `.`
+        // or an empty constructor pass straight through, so `[5]`, `[.]` and
+        // `[]` stay accepted. jq drains the whole collect, so the inner bound
+        // is unlimited whatever the outer `keep` -- `keep` governs how many
         // *arrays* the consumer wants, and there is exactly one.
         //
         // Keyed on `!trackable`: a trackable input (`path([.a] | empty)`,
@@ -27978,21 +27977,54 @@ fn resolve_node_sink<'a, S: EvalSemantics>(
         // only (ADR-0018): real yq's lexer rejects every shape that reaches
         // here, so there is no yq oracle, and yq mode keeps `skip_untracked`.
         //
-        // **Also keyed on the brackets holding no `TrackedVar`.** A marker
-        // can *re-establish* tracking -- `$v` whose frozen value is still the
-        // live register is not navigation, it is the register (#1573) -- but
-        // that comparison happens one level up, in `resolve_seq_stage`,
-        // against a `carried_register` this function is not handed. Resolved
-        // from here with `trackable: false`, `[$v.b?]` saw `$v` as just
-        // another untracked value and refused `.b` on it, where jq (and the
-        // catch-all, by value) accepts. So the marker case keeps the
-        // catch-all, exactly as before; only pure navigation on the ambient
-        // input is intercepted. Re-establishment inside the brackets needs
-        // the register threaded down, which is #2759.
+        // **Also keyed on `inner` containing none of the shapes whose own
+        // untracked handling is wrong.** Evaluating by value had been
+        // *masking* those bugs inside brackets, and routing the brackets
+        // through the resolver unmasks them -- turning working programs into
+        // errors, the one direction this fix must never move. Each is a
+        // pre-existing bare-stage bug with its own oracle bracket, kept on
+        // the catch-all here rather than fixed in passing:
+        //
+        // - `TrackedVar`: a marker *re-establishes* tracking -- `$v` whose
+        //   frozen value is still the live register is not navigation, it is
+        //   the register (#1573) -- but that comparison happens one level up
+        //   in `resolve_seq_stage`, against a `carried_register` this
+        //   function is not handed. Resolved from here, `[$v.b?]` saw `$v` as
+        //   just another untracked value and refused `.b` where jq accepts.
+        //   #2759.
+        // - `GetPath`: `getpath([])` refuses *eagerly* on an untracked input
+        //   (its own `keys.is_empty()` arm), where jq treats it as a
+        //   passthrough and refuses only at the terminal. `path(1 |
+        //   [getpath([])] | empty)` is accepted by jq. #2764.
+        // - `Recurse`/`RecurseF`/`RecurseCond`/`RecursiveDescent`: the shared
+        //   `recurse_untracked_error` guard refuses eagerly whether or not
+        //   `f` navigates -- `recurse(empty)`, `recurse(.+1; .<3)` and a
+        //   caught `(..)?` are all accepted by jq. And its error is
+        //   uncatchable here, so `[try (..)]` cannot rescue it either. #2764.
+        // - `Optional`: the parser erases the parentheses in `(.a)?`, which
+        //   is jq's `try .a` (a *caught* path error), into `.a?`, which is
+        //   jq's `INDEX_OPT` (an uncaught one). Bare `[.a?]` happens to agree
+        //   with jq either way, but `[(.a)? | f]` does not, and this function
+        //   cannot tell the two spellings apart. #2764.
+        //
+        // The scan is O(|inner|), the same order as resolving it.
         Expr::Array(inner)
             if S::TAG == EvalTag::Jq
                 && !trackable
-                && !any_subexpr(inner, &mut |e| matches!(e, Expr::TrackedVar(_))) =>
+                && !any_subexpr(inner, &mut |e| {
+                    matches!(
+                        e,
+                        Expr::TrackedVar(_)
+                            | Expr::Optional(_)
+                            | Expr::RecursiveDescent
+                            | Expr::Builtin(
+                                Builtin::GetPath(_)
+                                    | Builtin::Recurse
+                                    | Builtin::RecurseF(_)
+                                    | Builtin::RecurseCond(_, _)
+                            )
+                    )
+                }) =>
         {
             let mut items = Vec::new();
             let flow = resolve_node_sink::<S>(
--- a/src/jq/eval.rs
+++ b/src/jq/eval.rs
@@ -27947,6 +27947,77 @@ fn resolve_node_sink<'a, S: EvalSemantics>(
             resolve_nth_sink::<S>(n, expr, value, trackable, snapshot, frame, keep, sink)
         }
 
+        // #2689: array construction on an *untracked* input. jq's `[f]`
+        // (`gen_collect`) runs `f` with path tracking live -- unlike `{k:f}`,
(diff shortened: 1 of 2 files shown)


newhoggy:
## Adversarial review pass — a real over-refusal, now fixed

I flagged over-refusal as the dangerous direction for this change, and the review found
it: **69 working programs that the first draft turned into errors**, including `del`/`=`
forms refusing edits jq performs. Three sweeps totalling ~5,500 rows against jq 1.7.1, each
row also run on `main` to separate regressions from pre-existing gaps.

### The bug in my premise

The arm assumed "the inner resolver arms already refuse correctly against an untracked
input". That is false for three of them — and evaluating the brackets **by value** had
been masking those bugs. Routing the brackets through the resolver unmasked them:

| class | shape | jq | first draft | `main` |
|---|---|---|---|---|
| **(a)** `getpath([])` refuses eagerly | `path(1 \| [getpath([])] \| empty)` | exit 0 | **refuse** | exit 0 |
| | `del(. as {a:$x} \| [getpath([])] \| select(false))` | doc unchanged | **refuse** | doc unchanged |
| **(b)** `recurse` family refuses on sight, uncatchably | `path(1 \| [recurse(empty)] \| empty)` | exit 0 | **refuse** | exit 0 |
| | `path(1 \| [(..)?] \| empty)` | exit 0 | **refuse** | exit 0 |
| | `(1 \| [recurse(empty)] \| select(false)) = 1` | `null` | **refuse** | `null` |
| **(c)** parser stores `(.a)?` (jq's caught `try .a`) as `.a?` (jq's uncaught `INDEX_OPT`) | `path(1 \| [(.a.b)? \| type] \| empty)` | exit 0 | **refuse** | exit 0 |
| | `del([.a] \| [(.a)? \| type] \| select(false))` | doc unchanged | **refuse** | doc unchanged |

Each is a pre-existing bare-stage bug (`path(1 | getpath([]) | empty)` refuses on `main`
too) with its own oracle bracket. Filed as **#2764** rather than fixed in passing.

### The fix

The `TrackedVar` bail-out is widened to also hand `GetPath`, the `recurse` family, and
any `Optional` back to the old catch-all. **Re-swept 140 rows vs jq and vs `main`: zero
regressions, 36 rows to jq's answer**, and every row that still differs from jq is
byte-identical on `main`.

### The honest cost

Two rows the first draft had fixed return to `main`'s accept: `path(1 | [.[]?] | empty)`
and `[.a?]` (jq refuses — `INDEX_OPT`). `Optional` cannot be admitted without also
admitting `(.a)?`, because the parser has already erased the distinction. That is the safe
direction, and #2764 records that the guard can readmit `Optional` once the parser keeps
the two apart. The issue's own reported shapes are unaffected — none contains an
`Optional` outside a marker.

### Also corrected

- The doc comment and test doc listed `try f` among constructs that *suspend* tracking.
  False — `path(try .a)` is `["a"]` in jq. `try` keeps tracking live and **catches** the
  path error, which is precisely the mechanism behind class (c).
- Tests now pin both directions: the deferred shapes staying accepted, and a write that jq
  performs as a no-op staying a no-op (the first draft refused every one).

### Confirmed clean by the review

- **Escape/demand**: `[., break $out]`, `[halt]`, `[error("x")]`, `[error("x")]?` all
  match jq and `main`; `Stopped` from the inner sink is unreachable; the collect is fully
  drained under `first`/`limit` (verified via `debug` counts). One improvement:
  `path(1 | try [.[]] catch .)` now reproduces jq's caught message.
- **`snapshot`/`frame` threading**: correct. `path(.a as $y | 5 | $y | [.b] | empty)` now
  refuses as jq does (`main` accepted); a live register (`.a as $y | .a | $y | [.b]`) stays
  accepted; `catch [.b]` payloads now give jq's exact wording.
- **Marker guard**: bails too eagerly (`[.a, $x]` still accepted where jq refuses), but
  identical on `main` and entirely within #2759's description.
- **yq mode**: zero change with or without `--jq-extensions`.
- **Perf**: ~+20% on a 200k-element debug-build `del(.[] | . as {a:$v} | [range(20)] |
  select(false))` — the resolve path is heavier than by-value eval for a generator inner;
  `[5]` is within noise. Niche shape, not measured in release.

Issue #2235, opened 2026-09-02
jq: resolve_fold_source's eager Keep::AtMost pull fires a foreach source's side effects before an early UPDATE failure would have stopped it


newhoggy:
**Severity: Low** (side-effect count only — final disposition, stdout, and exit code all
already agree with jq)

## Summary

`resolve_fold_source` (`src/jq/eval.rs`) resolves a navigating fold source with
`Keep::AtMost(usize::MAX)` (jq mode) so every one of its outputs is pulled and
materialized *before* `resolve_foreach`'s own per-step loop ever runs. Real jq's actual
generator model is demand-driven: it only pulls the source's next element once the
*previous* step has fully resolved, so a `foreach` whose first step's own UPDATE turns
out to be untrackable stops the whole computation — including any further source
pulls — right there.

## Repro (confirmed live against jq 1.7.1)

```console
$ echo '[1,2,3]' | jq -c '. as $x | path(foreach (.[] | stderr) as $i (0; $x))'
1jq: error (at <stdin>:1): Invalid path expression with result [1,2,3]
# stderr: "1" -- one write, then the error (source never pulled past its first element)

$ echo '[1,2,3]' | succinctly jq -c '. as $x | path(foreach (.[] | stderr) as $i (0; $x))'
123jq: error (at <stdin>:1): Invalid path expression with result [1,2,3]
# stderr: "123" -- all three writes fire before the error, since the source is pulled
# upfront regardless of what the fold loop will do with it
```

Exit code, stdout, and the error message itself already match jq exactly — this is
purely a side-effect-count divergence, the same class #1872 already fixed one layer of
(a source's first output no longer firing its side effect *twice*). This is the
remaining layer: pulling the *whole* source eagerly, rather than one element at a time
as the fold loop actually consumes it.

## Why this is separate from #2234

#2234 fixed `stderr`/`debug`'s own path-trackability (they used to raise an unrelated,
spurious \"Invalid path expression\" for *any* use inside `path()`, masking this
divergence entirely — before that fix, this exact query silently answered `[]` three
times at exit 0, matching neither jq's stderr count *nor* its disposition). Once
trackability was fixed, this query's disposition (exit code, stdout, the raised error)
now matches jq correctly; only the fold source's own pull-eagerness remains a
divergence, and that's a property of `resolve_fold_source`'s architecture, unrelated to
whether any particular builtin is path-trackable.

## Root cause

[...]

PR #2773, opened 2026-09-11
test(jq): make a new Expr variant fail to compile until path dispatch is decided

### LINKED ISSUE #2697: jq: resolve_node_sink/resolve_node_eager's dispatch wildcards give no compiler nudge for new Expr variants
**Severity: Low** (maintainability/process gap, not a live bug)

## Summary

`resolve_node_sink` and `resolve_node_eager` (`src/jq/eval.rs`) are the two dispatch
points #2235's demand-driven migration has been incrementally moving `Expr`/`Builtin`
arms out of (eager -> sink, one construct at a time: `Optional`, `Try`, `Alternative`,
`Reduce`, `Foreach`, `recurse(f)` so far — see #2693/#2694/#2696 for what's left). Both
matches end in an unchecked wildcard arm (`other => ...`), so adding a new `Expr`/
`Builtin` variant compiles cleanly and silently lands on the eager, non-streaming path.
Nothing forces a human to notice, from prose alone, that a new bounded-generator-shaped
construct needs a `_sink` arm to actually stream under `limit`/fold-source contexts.

Related: naming is also inconsistent about which functions are eager vs. sink-based —
migrated functions are conspicuously suffixed `_sink` (`resolve_recurse_sink`,
`resolve_catch_sink`, `resolve_optional_sink`, `resolve_alternative_sink`) or renamed to
`drive_*` (`drive_fold_source`), while pre-existing eager helpers (`resolve_recursive_descent`,
`push_recursive_branches`, `resolve_bind_source`, `resolve_leaf`) carry no `_eager`/`_vec`
marker a maintainer could grep for.

Found during review of #2685 (the fix for #2190).

## Fix direction

No compiler mechanism can force exhaustiveness on an already-open-ended wildcard match
without a larger restructuring (the match is hundreds of arms deep, and `Expr`/`Builtin`
are large enums with many non-path-relevant variants that legitimately belong on the
eager fallback forever). Cheaper options worth considering instead of a full rewrite:
- A repo-local lint/CI check (e.g. `omni-dev` or a custom `grep`-based CI step) that flags
  a new `Expr`/`Builtin` variant added without a corresponding decision recorded nearby.
- Consistently naming every function that collects eagerly (not just the ones already
  renamed) with a shared marker, so `git grep` for the marker is a reliable audit tool.
- A doc comment on both wildcard arms' `other` binding that explicitly asks new authors to
  update a checklist (partially already done, but not enforced).

This is a discussion/tracking issue, not a specific patch — the right shape of the fix
(lint vs. naming convention vs. accepted status quo) is a judgment call for whoever picks
it up.

newhoggy:
## Summary

`resolve_node_sink`/`resolve_node_eager` both end in an unchecked `other =>` arm, so a new
`Expr` variant compiles cleanly and lands silently on the eager, non-streaming path.
Nothing made a human notice that a new bounded-generator-shaped construct needs a `_sink`
arm to actually stream under `limit`/fold-source contexts.

## The issue's premise was too pessimistic

#2697 says no compiler mechanism can help "without a larger restructuring", and proposes a
repo-local lint or a naming convention instead. Neither is needed — **this repo already
uses the technique one module over.** `walk.rs`'s `builtin_kids` names every `Builtin`
variant with no wildcard, precisely so "adding a variant fails to compile here until its
sub-expressions are declared", and its doc comment argues the general case:

> Resist the temptation to shorten this with a catch-all: the cost of the long match is
> paid once, by whoever adds a variant; the cost of a wildcard is paid silently and
> repeatedly by everyone downstream.

What makes it apply here without restructuring is that **the witness does not have to be
the dispatch**. `classify` is a separate, wildcard-free match over all 56 `Expr` variants
whose only job is to stop compiling when the enum grows. `eval.rs`'s hundreds-of-arms match
keeps its catch-all; what it loses is the ability to absorb a new variant unnoticed.

## Why the classification is cross-checked, not just written down

A hand-maintained list of "which variants stream" is exactly the artefact that rots: it
would still compile after someone migrated a construct between the two paths, and would
then describe the code as it was. So it is re-derived from `eval.rs`'s own source
(`include_str!` — the shape `jq_member_validation_audit.rs` and
`jq_optional_suppression_audit.rs` already use).

The check is **set equality in both directions**: the variants classified `Sink` are
exactly the variants `resolve_node_sink` names. That held on the nose — **24 and 24, no
stragglers either way** — which is what makes the strong form worth asserting. I had
initially written a one-directional check with a hedge that the converse "isn't asserted
because the function also names variants it delegates elsewhere"; that turned out to be
false when I checked, so the hedge is gone and the strong invariant is in its place. It
matters: a one-directional check still passes after a construct is migrated *back* to
eager, and the reverse direction — a new `_sink` arm appearing — is the one #2235 is
actually travelling.

## Both halves are negative-tested

A guard that cannot fail is worthless, so I verified each can:

```console
# delete one arm from `classify`
error[E0004]: non-exhaustive patterns: `&succinctly::jq::Expr::Break(_)` not covered

# drop a variant from the asserted list
thread 'sink_classification_matches_the_dispatch' panicked at tests/jq_path_dispatch_audit.rs:273:5:
named in `resolve_node_sink` but not classified `Sink`: ["Repeat"] -- a construct gained a
native demand-driven arm; classify it `Sink` in `classify` and add it here
```

## What a future maintainer sees

Adding an `Expr` variant now stops the build with a message naming the variant, at a file
whose doc comment says what decision to make: give it a `_sink` arm and classify it `Sink`,
or classify it `Eager` and say why. Each `Eager` group carries its rationale — single-valued
navigation has nothing to truncate; value-producing constructs can't *be* a path;
`LastExpr`/`Until`/`While`/`Range` must run to completion anyway (`last(f)` has to drain `f`
to know its final output, which is why it has no arm beside `FirstExpr`'s); `FuncDef`/
`FuncCall` are resolved away before dispatch; assignment forms are consumers of `path()`,
not path expressions.

## Scope

The naming-consistency half of #2697 (eager helpers carrying no `_eager`/`_vec` marker a
maintainer could grep for) is **not** addressed here — that is a rename touching many call
sites, and the compiler nudge is what makes the grep unnecessary for the case that
matters. Say the word if you want it split into its own issue rather than left on #2697.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — full suite, 63 binaries, 0 failures
- [x] `cargo build --tests` with **default features** (the new file uses only
      `succinctly::jq::{Builtin, Expr}`, so it must not need `cli`)
- [x] Negative-tested both mechanisms (above), then restored and re-confirmed green
- [x] **No behaviour change** — adds a test file, touches nothing else

Fixes #2697

### DIFF
--- a/tests/jq_path_dispatch_audit.rs
+++ b/tests/jq_path_dispatch_audit.rs
@@ -1,0 +1,305 @@
+//! #2697's compiler nudge: a wildcard-free witness over `Expr`, so adding a
+//! variant cannot silently land on `path()`'s eager fallback.
+//!
+//! # The gap this closes
+//!
+//! `resolve_node_sink` and `resolve_node_eager` (`src/jq/eval.rs`) are the two
+//! dispatch points #2235's demand-driven migration moves constructs through,
+//! one at a time. Both end in an unchecked `other => ...` arm, so a new
+//! `Expr` variant compiles cleanly and lands on the eager, non-streaming
+//! path. Nothing makes a human notice that a new bounded-generator-shaped
+//! construct needs a `_sink` arm to stream under `limit`/fold-source
+//! contexts.
+//!
+//! #2697 assumed no compiler mechanism could help "without a larger
+//! restructuring", and proposed a lint or a naming convention instead. Both
+//! are weaker than what this repo already does one module over:
+//! `walk.rs`'s `builtin_kids` names every `Builtin` variant with no wildcard,
+//! precisely so "adding a variant fails to compile here until its
+//! sub-expressions are declared". Its doc comment is worth re-reading before
+//! touching this file:
+//!
+//! > Resist the temptation to shorten this with a catch-all: the cost of the
+//! > long match is paid once, by whoever adds a variant; the cost of a
+//! > wildcard is paid silently and repeatedly by everyone downstream.
+//!
+//! The same technique works here **without restructuring the real dispatch**,
+//! because the witness does not have to *be* the dispatch. `classify` below
+//! is a separate, wildcard-free match whose only job is to fail compilation
+//! when `Expr` grows. The hundreds-of-arms match in `eval.rs` keeps its
+//! catch-all; what it loses is the ability to absorb a new variant unnoticed.
+//!
+//! # Why the classification is cross-checked and not just written down
+//!
+//! A hand-maintained list of "which variants stream" is exactly the artefact
+//! that rots -- it would still compile after someone migrates a construct
+//! from the eager path to a `_sink` arm, and would then describe the code as
+//! it was, not as it is. So `sink_classification_matches_the_dispatch`
+//! re-derives the truth from `eval.rs`'s own source (`include_str!`, the same
+//! shape `jq_member_validation_audit.rs` and `jq_optional_suppression_audit.rs`
+//! use) and asserts the witness agrees.
+//!
+//! That makes the file self-correcting in both directions: adding an `Expr`
+//! variant breaks the build here, and migrating a construct between the two
+//! dispatch paths breaks this test until the witness is updated to match.
+
+use succinctly::jq::{Builtin, Expr};
+
+/// Which of `path()`'s two dispatch paths a variant takes in
+/// `resolve_node_sink` (`src/jq/eval.rs`).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum PathDispatch {
+    /// Has a native demand-driven arm: it can stop a generator early, so a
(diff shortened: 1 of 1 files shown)


newhoggy:
Closing unmerged. The premise this PR rests on is false — a new `Expr` variant already fails the library build with 5 `error[E0004]` sites, and the audit file asserted three wrong classifications (`Identity`, `RecursiveDescent`, `TrackedVar` are all eager, not sink). Full evidence in https://github.com/rust-works/succinctly/issues/2697#issuecomment-latest.

PR #2805, opened 2026-09-11
fix(jq): let a def parameter shadow a same-named builtin

### LINKED ISSUE #2723: jq: a bare def parameter named after a builtin loses to the builtin (def f(length): length; f(1) is 0, not 1)
**Severity: Medium**

## Summary

A `def` parameter whose name collides with a zero-arity builtin is never substituted at
the call site — the builtin wins instead, and the program silently produces a wrong
answer rather than an error.

## Repro

Confirmed live against jq 1.7.1 and on `main` (`7d84b2329`):

```console
$ echo null | jq -c 'def f(length): length; f(1)'
1
$ echo null | succinctly jq -c 'def f(length): length; f(1)'
0

$ echo null | jq -c 'def f(not): not; f(1)'
1
$ echo null | succinctly jq -c 'def f(not): not; f(1)'
true

$ echo null | jq -c 'def f(empty): [empty]; f(1)'
[1]
$ echo null | succinctly jq -c 'def f(empty): [empty]; f(1)'
[]

$ echo null | jq -c 'def f(recurse): recurse; f(1)'
1
$ echo null | succinctly jq -c 'def f(recurse): recurse; f(1)'
null
```

Note `empty` is the nastiest of these: the parameter reference is not merely misread, it
*deletes the output*.

## Root cause

`substitute_func_param_impl`'s bare-parameter arm (`src/jq/eval.rs`) matches
`Expr::FuncCall { name, args: [] }` where `name == param`. A builtin-named identifier
never reaches that arm as a `FuncCall` — the parser (or `builtin_fallback` resolution)
has already committed it to `Expr::Builtin`, so the substitution walks straight past it.
In jq, a parameter binding shadows a builtin of the same name for the whole body.

## Why this is separate from #2560

#2560 was `bind_def_call`'s *ordering* of substitutions across a duplicated parameter
name. This is about a single, non-duplicated parameter never being substituted at all
because of how its name resolves — the substitution machinery is not reached, so
#2560's fix neither causes nor addresses it. Verified identical on `main` and on
#2560's branch.

## Suggested fix direction

Decide the bare-parameter-vs-builtin precedence before the builtin is committed —
either resolve parameter names to `FuncCall` at parse/resolve time when a name is
in scope as a parameter, or teach the substitution's `Expr::Builtin` arm to rewrite a
zero-argument builtin whose name matches `param` back into the argument. The former is
closer to how jq scopes it; the latter is more localized. Whichever is chosen, the whole
family above should be pinned, `empty` included.

Found via adversarial review of #2560 (PR #2722).

newhoggy:
## Summary
- A `def` parameter named after a zero-arity builtin (`length`, `not`, `empty`, `recurse`, ...) was never substituted at the call site — the builtin won instead, silently producing a wrong answer (`def f(empty): [empty]; f(1)` was `[]`, not `[1]`).
- Root cause: the #2036 shadow-candidate prescan (`collect_def_names`) only ever collected a `def`'s own name, not its parameters, so a bare reference to a builtin-shadowing parameter committed straight to the builtin's dedicated `Expr` variant and never reached `resolve.rs`'s scope-aware check — even though that check already pushes every parameter into scope exactly like a `def`.
- Fix: extend the same prescan to also collect parameter names (stripping a leading `$`, since a `$`-style parameter binds the bare namespace too per jq's own desugaring). This routes every affected builtin through the existing, already-tested shadow-candidate machinery instead of adding a new mechanism.

## Test plan
- [x] Live-verified every repro and edge case against `/usr/bin/jq` 1.7.1 (oracle)
- [x] Local patch coverage: 100% (27/27 new lines)

Fixes #2723

### DIFF
--- a/src/jq/parser.rs
+++ b/src/jq/parser.rs
@@ -209,7 +209,8 @@ fn dollar_var_expr(name: String, line: usize) -> Expr {
 }
 
 /// #2036 Direction 3: a cheap, deliberately over-approximate scan for every
-/// identifier spelled after a `def` keyword anywhere in `input` -- see
+/// identifier spelled after a `def` keyword anywhere in `input`, plus (#2723)
+/// every identifier spelled in that `def`'s own parameter list -- see
 /// `Parser::shadowable_defs`'s own doc comment for why imprecision here is
 /// safe (a false positive costs one wasted double-parse at one call site; a
 /// false negative is impossible to turn into a wrong *answer* either, since
@@ -225,6 +226,23 @@ fn collect_def_names(input: &str) -> BTreeSet<String> {
     // only ever widens the (already-imprecise) set, never narrows it.
     fn is_ident_continue(c: char) -> bool {
         c.is_alphanumeric() || c == '_' || c == '-'
+    }
+
+    // #2036 review round 2's comment-skip loop, factored out so the
+    // parameter-list scan below (#2723) can reuse it verbatim instead of
+    // hand-duplicating it -- real jq's own `skip_ws` skips both whitespace
+    // and comments, repeatedly, wherever it appears in a `def` header.
+    fn skip_ws_and_comments(mut rest: &str) -> &str {
+        loop {
+            let trimmed = rest.trim_start();
+            let Some(after_hash) = trimmed.strip_prefix('#') else {
+                return trimmed;
+            };
+            rest = match after_hash.find('\n') {
+                Some(i) => &after_hash[i..],
+                None => "",
+            };
+        }
     }
 
     let mut names = BTreeSet::new();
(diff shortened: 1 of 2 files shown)


newhoggy:
Ran a 5-way code review round against this PR. Three real correctness bugs were found in the initial parameter-list scan (all confirmed live against jq 1.7.1) and fixed in a follow-up commit:

1. A `#`-comment inside a def's parameter list containing its own `)` or `;` fooled the naive `find(')')`/`split(';')`, silently reopening this exact issue's bug for a comment-containing header.
2. `,` (accepted by `Parser::parse_func_params` alongside `;`) was never handled by the scan, so a comma-separated parameter list silently dropped its shadow candidate.
3. The unbounded `find(')')` scan was O(n) per `def` occurrence even when unclosed, making a crafted file O(n²) overall — confirmed empirically (~2.7s at 128k reps pre-fix).

Replaced with a token-by-token walk that's comment-aware by construction and bails cleanly on the first non-conforming token, fixing all three. Added regression tests for each (including a timing-guard test for the O(n²) fix, following the existing `test_def_shadow_deep_nesting_does_not_blow_up_2036` pattern) and re-verified 100% patch coverage.

One lower-confidence finding (widening the shadow-candidate set with parameter names could plausibly exhaust the shared `SHADOW_RETRY_BUDGET` faster in a large program) had no live repro constructed during review — filed as #2807 for future investigation rather than blocking this PR.

Issue #2818, opened 2026-09-12
jq: bare flatten (no depth argument) flattens only 1 level; real jq flattens fully


newhoggy:
**Severity: Medium** (a correctness bug on a very commonly-used builtin with no argument)

## Summary

Found while implementing #2755's fix for `flatten(depth)`'s argument handling. `Builtin::Flatten` (the bare, no-argument form) is wired to `builtin_flatten::<W, S>(value, optional, OwnedValue::Int(1))` — i.e. bare `flatten` is treated as `flatten(1)`. Real jq's bare `flatten` instead flattens *fully*, all the way down, with no depth limit at all.

## Repro

Confirmed live against jq 1.7.1 and on `main`:

```console
$ echo null | jq -c '[[1,[2,[3,[4]]]],5] | flatten'
[1,2,3,4,5]
$ echo null | succinctly jq -c '[[1,[2,[3,[4]]]],5] | flatten'
[1,[2,[3,[4]]],5]

$ echo null | jq -c '[[1,[2,[3,[4]]]],5] | flatten(1)'
[1,[2,[3,[4]]],5]
```

Bare `flatten` and `flatten(1)` are genuinely different operations in jq; succinctly currently treats them as identical.

## Note

An existing unit test, `test_builtin_flatten` (`src/jq/eval.rs`), currently pins the wrong (1-level-only) behavior for bare `flatten` — e.g. asserting that `[[1, 2], [3, [4]]] | flatten` keeps `[4]` as a nested array rather than fully flattening it. That assertion will need updating alongside the fix.

## Suggested fix direction

Real jq's `builtin.jq` defines bare `flatten` as `def flatten: flatten(1/0);` — dividing by zero produces `nan`/infinity depending on context, but the practical effect is "no depth ever compares as reached." Whatever internal sentinel `builtin_flatten`'s `depth: OwnedValue` parameter (post-#2755) uses for "never stop," bare `flatten` should pass that instead of `OwnedValue::Int(1)`.

Filed separately from #2755 to keep that PR focused on the depth-*argument* handling; this is a distinct bug in the bare (zero-argument) form's wiring.

newhoggy:
Claimed by grinding-backlog agent `Blake` — branch `issue-2818-bare-flatten-depth`. Bare `flatten` is wired to depth=1 instead of jq's actual infinite-depth default; fixing the dispatch to use the same never-stop sentinel `flatten(depth)` uses post-#2755.

Issue #1198, opened 2026-08-17
yq: arith_sub has no null-left identity arm, unlike real yq


newhoggy:
**Severity: Low**

Found during code review of #1175/PR #1196 (`arith_mul`'s null-multiply fix). Same "null needs yq-specific gating" pattern that PR applied to `*`, left unaddressed for `-`.

## Root cause

`arith_sub` (`src/jq/eval.rs`, near line 1847) has no `Null`-handling arm at all — every null-involving subtraction falls straight through to the generic type-mismatch error arm, in both jq and yq mode.

Real yq treats a null *left* operand as an identity element for `-`, across every type on the right. Real jq has no such exception (matches succinctly's current jq-mode behavior, which is correct and should not change).

## Repro (live-verified against real yq v4.53.3)

```
$ echo null | yq 'null - 7' -o=json
7

$ echo null | yq 'null - "a"' -o=json
"a"

$ echo null | yq 'null - []' -o=json
[]

$ echo null | yq 'null - {}' -o=json
{}
```

succinctly (current `main`):

```
$ echo null | succinctly yq -o=json '. - 7'
Error: null (null) and number (7) cannot be subtracted   # should be 7

$ echo null | succinctly yq -o=json 'null - "a"'
Error: null (null) and string ("a") cannot be subtracted  # should be "a"

$ echo null | succinctly yq -o=json 'null - []'
Error: null (null) and array ([]) cannot be subtracted    # should be []
```

## Scope note

The fix likely means adding a yq-gated arm mirroring `arith_mul`'s own null-left-identity handling, e.g. `(OwnedValue::Null, other) if S::TAG == EvalTag::Yq => Ok(other)`, placed before the generic error arm. A right-null operand (`7 - null`) was not part of this investigation's repro set and should be live-verified against real yq before assuming it's unaffected.

## Severity

Low: a currently-erroring operation should succeed instead, in yq mode only.

newhoggy:
🔧 Claimed by agent `aiden` — branch `issue-1197-yq-arith-null-gating` (bundled with #1197: same file, same root-cause pattern — `arith_add`/`arith_sub` in src/jq/eval.rs both need yq-specific null-operand gating mirroring #1175's `arith_mul` fix).

PR #2819, opened 2026-09-12
fix(jq): thread flatten's depth argument lazily, matching jq

### LINKED ISSUE #2755: jq: flatten's non-negative-depth argument rejects float/string/array/object depths jq accepts (or errors differently)
**Severity: Medium** (a correctness bug, not just a message mismatch — several
depth-argument shapes real jq accepts are rejected outright, or vice versa)

## Summary

While fixing #2747 (flatten's negative-depth error message), live-verification against
jq 1.7.1 turned up a second, deeper divergence in the *positive* side of `flatten`'s
depth-argument handling. `builtin_flatten_depth` (`src/jq/eval.rs`) only accepts a
literal non-negative `Int`/`NumberLiteral(Int)` on the success path; anything else falls
to a blanket `type_error("number", "non-number")`. Real jq's actual definition is

```jq
def flatten($x): if $x < 0 then error("flatten depth must not be negative") else _flatten($x) end;
def _flatten($x):
  reduce .[] as $i
    ([]; if ($i|type) == "array" and $x != 0
         then . + ($i | _flatten($x - 1))
         else . + [$i]
         end);
```

`$x` is decremented (`$x - 1`) once per recursion level and compared with `!= 0` — not
converted to an integer count up front — so several depth shapes jq happily accepts (or
errors on differently) are currently misclassified.

## Repro (confirmed live against jq 1.7.1, `main` at the time of #2747)

```console
# Integral float depth: jq accepts (0.0 decrement path hits exactly 0), succinctly rejects
$ echo '{}' | jq -c '[[1,[2]],3] | flatten(2.0)'
[1,2,3]
$ echo '{}' | succinctly jq -c '[[1,[2]],3] | flatten(2.0)'
jq: error (at <stdin>:1): expected number, got non-number

# Non-numeric depth on genuinely nested input: jq only errors once _flatten actually
# tries `$x - 1` (a type-specific subtraction error), succinctly errors unconditionally
# and with a generic message, even for input with no nesting to trigger the real error
$ echo '{}' | jq -c '[[1,[2]],3] | flatten("a")'
jq: error (at <stdin>:1): string ("a") and number (1) cannot be subtracted
$ echo '{}' | succinctly jq -c '[[1,[2]],3] | flatten("a")'
jq: error (at <stdin>:1): expected number, got non-number
```

## Root cause

`builtin_flatten_depth`'s `depth_value` match only has a success arm for
`Int`/`NumberLiteral(Int) if d >= 0`, converting straight to a `usize` depth counter
consumed top-down by `builtin_flatten`. jq's own model instead threads `$x` itself
through each recursion level, decrementing and comparing with `!= 0` lazily — so a
non-integer numeric value, or a non-numeric value that a real nested nesting would
eventually try to subtract from, behaves differently depending on how deep the actual
input nesting goes, not on the depth argument's type alone.

## Why this is separate from #2747

#2747 is specifically about the *negative*-depth error message (and, per the same
investigation, correctly reaches that arm for `null`/`false`/`true`/`nan` via jq's own
total ordering). This issue is about the disjoint set of depth values that are neither
provably negative nor a plain non-negative literal integer — a materially larger change
(`builtin_flatten`'s recursion model would need to carry `$x` itself, or an equivalent
lazy comparison, rather than a pre-converted `usize`), not something to bundle into
#2747's narrowly-scoped fix.

## Suggested fix direction

Rework `builtin_flatten`/`builtin_flatten_depth` to thread the depth value's actual jq
semantics (`$x != 0` check per recursion level, `$x - 1` per descent, raising jq's own
"X and number cannot be subtracted" type error when `$x` isn't itself numeric and a
descent is actually attempted) rather than eagerly converting to a `usize` up front.
Acceptance test: the two repro rows above, plus a differential fuzz sweep over
depth-argument shapes (int/float/string/array/object/bool/null) x nesting shapes, since
the observable behavior is nesting-shape-dependent for non-numeric depths.

newhoggy:
## Summary
- `flatten(depth)`'s success path eagerly required a literal non-negative integer, rejecting everything else with a generic "expected number, got non-number" — but jq's own reference definition (`_flatten($x): reduce .[] as $i ([]; if ($i|type)=="array" and $x!=0 then . + ($i|_flatten($x-1)) else . + [$i] end)`) never converts `$x` up front. It only ever inspects `$x` via `!= 0` and `$x - 1`, both lazily, once per recursion level, only when a genuinely nested array item is reached.
- `flatten_owned`/`flatten_owned_at_depth` now thread `depth` as `OwnedValue` instead of `usize`, reusing the existing `compare_values`/`arith_sub` machinery for those two operations. An integral float depth now succeeds (`flatten(2.0)` decrements to exactly `0.0`, which compares equal to `0`), and a non-numeric depth only errors once real nesting is reached, with jq's own subtraction-type error wording.
- Gated to **jq mode only**: real yq's own grammar accepts only a bare non-negative integer literal for `flatten`'s argument, rejecting everything this widens for at parse time — so yq mode keeps its exact pre-existing eager rejection unchanged (confirmed during review, which caught a first draft accidentally letting yq's `null - x = x` identity, #1198, leak into `flatten(null)`).
- Filed #2818 separately for an unrelated pre-existing bug found along the way: bare `flatten` (no args) is wired to `flatten(1)` instead of unlimited depth.

## Test plan
- [x] Live-verified every repro against `/usr/bin/jq` 1.7.1 and yq v4.53.3
- [x] Local patch coverage: 100% (28/28 new lines)

Fixes #2755

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -417,17 +417,30 @@ have to be recorded here.
 |--------------------|---------|-------------------------------------|-----------------------------------|
 | `@uri`             | `[1,2]` | `"%5B1%2C2%5D"`                     | `expected string, got array`      |
 | `@base64`          | `5`     | `"NQ=="`                            | `expected string, got number`     |
-| `flatten("x")`     | `[1,2]` | `[1,2]` (ignores non-integer depth) | `expected number, got non-number` |
 
-[#929](https://github.com/rust-works/succinctly/issues/929) found these while auditing
-`EvalError::type_error` wording: real jq's `@uri`/`@base64` (and every other format string
-except `@csv`/`@tsv`/`@sh`) auto-`tostring`s a non-string argument before formatting rather
-than refusing it outright, and `flatten`'s depth argument is silently ignored if it isn't a
-number rather than validated. `@base64d` is a related but distinct case — jq *does* still
-error on `5 | @base64d` (`string ("5") trailing base64 byte found`), just from attempting to
-base64-decode the auto-stringified `"5"` rather than refusing the number up front, so it
-isn't purely a "jq doesn't error" gap; matching it needs the same underlying
-auto-`tostring`-first change `@uri`/`@base64` do, not just a different error message.
+[#929](https://github.com/rust-works/succinctly/issues/929) found these (a third row,
+`flatten("x")` on `[1,2]`, is below) while auditing `EvalError::type_error` wording: real
+jq's `@uri`/`@base64` (and every other format string except `@csv`/`@tsv`/`@sh`)
+auto-`tostring`s a non-string argument before formatting rather than refusing it outright.
+`@base64d` is a related but distinct case — jq *does* still error on `5 | @base64d`
+(`string ("5") trailing base64 byte found`), just from attempting to base64-decode the
+auto-stringified `"5"` rather than refusing the number up front, so it isn't purely a "jq
+doesn't error" gap; matching it needs the same underlying auto-`tostring`-first change
+`@uri`/`@base64` do, not just a different error message.
+
+`flatten("x")` on `[1,2]` (`[1,2]` in jq, `expected number, got non-number` here) was the
+third #929 row: jq's own `flatten` never validates its depth argument up front, only ever
+inspecting it lazily (`!= 0`, `- 1`) once per recursion level, so a non-numeric depth is
+silent whenever the input has no nesting deep enough to actually reach a real `- 1`
+attempt. [#2755](https://github.com/rust-works/succinctly/issues/2755) closed it:
+`flatten`'s depth argument now threads through as the raw value (`OwnedValue`) rather than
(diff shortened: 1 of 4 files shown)


newhoggy:
Ran a code review round against this PR (extensive live A/B testing against jq 1.7.1 and yq v4.53.3, ~25 edge cases — the implementation itself came back clean). Two documentation-fidelity gaps were found and fixed:

1. The "Where succinctly errors and jq does not" table in `docs/compliance/jq/limitations.md` still listed the `flatten("x")` divergence this PR fixes — removed the row and added prose noting #2755 closed it, matching the doc's own convention.
2. `flatten_owned_at_depth`'s doc comment still described the early `if depth == 0 return items` return this PR replaced with a per-item match guard — could have misled a future maintainer into reintroducing it ahead of the `assert_value_tree_depth` call, re-masking #1017's depth-limit assert.

Both fixed in the follow-up commit; full test suite + clippy re-verified clean.

Issue #2549, opened 2026-09-06
jq: group (c) Expr-dispatch catch-alls have grown since #2182's own count (option 3 follow-up)


newhoggy:
**Severity: Low** (no known live bug; a triage/documentation gap, same spirit as #2182 itself)

## Summary

#2182 classified this codebase's 16 catch-all `unreachable!()` arms into three groups and stated group (c) ("`Expr`-dispatch catch-alls") was "effectively 2, and ... essentially closed" — one `#[cfg(test)]` helper (`via_cursor`) and one production site (`path_context_step_generic`, gated by `path_context_is_cursor_walkable`).

Re-auditing the current codebase while implementing #2182's own group (a) fix found **6 more** `Expr`-dispatch-shaped catch-alls that weren't in that count, all following the identical pattern (match on `Expr` variants, `_ =>`/`other =>` fallback, guarded by a caller-side or sibling-function precondition that a *different* `Expr` shape never reaches this point):

```
src/jq/eval_generic.rs  yq_list_stage         "yq_list_stage selects only Compare/Arithmetic as a binary stage"
src/jq/eval_generic.rs  owned_identity_step   "owned_identity_nav_supported admits no other shape"
src/jq/eval_generic.rs  owned_identity_step   "not a computed navigation step: {other:?}"
src/jq/eval.rs          wrap_fresh            "wrap_fresh only ever walks a fresh Field/Index run"
src/jq/eval.rs          (setpath dispatch)    "matches! guard above restricts inner to IndexExpr|SliceExpr"
```//exact function names for the last two need re-deriving at pickup time; line numbers drift too fast to cite here (see #235's own convention against citing line numbers in issues)

Either the codebase has grown 3x as many of this class since #2182 was written (a few months of otherwise-unrelated work adding new dispatch sites with the same pattern), or #2182's own grep/count missed them the first time. Either way, the "essentially closed, nothing to do here" conclusion in #2182 no longer describes the current state.

## Why this is separate from #2182

#2182's own scope statement is explicit: "Group (a) is the only group worth real work... group (c): consider deleting the two [not "audit for more"]." Re-deriving group (c)'s current membership and applying its own recommended treatment (verify each is genuinely closed by its guard, then either spell out an exhaustive match the same way #2182's group (a) fix did, or leave a stronger comment citing the specific guard) is real, separate work #2182 didn't scope in, once its actual size turned out to be 3x the original estimate.

## Suggested next step

For each of the 6 sites above:
[...]

PR #2823, opened 2026-09-12
fix(jq): give owned_identity_step a Literal arm, stop a process abort

### LINKED ISSUE #2771: jq/yq: a literal inside a pipe aborts the process — owned_identity_step has no Expr::Literal arm
**Severity: High** (process abort, exit 101, reachable from the CLI in both modes — not a wrong answer, a crash)

## Summary

`owned_identity_step` (`src/jq/eval_generic.rs`) has two call sites gated by **different**
predicates:

- `owned_identity_pipe`/`owned_identity_stages` gate on `owned_identity_nav_supported`
- `owned_identity_operand` gates on `owned_identity_operand_supported`

They disagree on exactly one shape: **`Expr::Literal`**, which
`owned_identity_operand_supported` admits and `owned_identity_step` has no arm for. A
top-level literal is short-circuited by `owned_identity_operand` itself (after
`strip_parens`), but a literal *inside a pipe* reaches the dispatch and hits its
`unreachable!`.

## Repro

Live on `main` @ `c0a442806`:

```console
$ printf 'a: {b: 1}\n' | succinctly yq '.a | ((.b | 1) + 2) | key'
thread '<unnamed>' panicked at src/jq/eval_generic.rs:20627:14:
internal error: entered unreachable code: owned_identity_nav_supported admits no other shape

$ echo '{"a":{"b":1}}' | succinctly jq -c '.a | ((.b | 1) + 2) | path'
thread '<unnamed>' panicked at src/jq/eval_generic.rs:20627:14:
internal error: entered unreachable code: owned_identity_nav_supported admits no other shape
```

Real jq 1.7.1 answers the `path` spelling with its ordinary refusal, not a crash:

```console
$ echo '{"a":{"b":1}}' | jq -c '.a | ((.b | 1) + 2) | path'
jq: error (at <stdin>:1): Invalid path expression with result 3
```

The trigger is the owned-identity operand route, so it needs a `key`/`path`-style
metadata read downstream to be entered at all; `.a | ((.b | 1) + 2)` alone is fine.

## Root cause

`owned_identity_operand_supported` admits `Expr::Literal(_)` (it is a legitimate operand
of a ruled arithmetic/comparison stage — `.a + 1`). `owned_identity_nav_supported` does
not, because a literal is not *navigation*. `owned_identity_step`'s match is written
against the navigation set, so the operand route can hand it a shape it never enumerated.

`owned_identity_operand` short-circuits a *bare* literal before stepping, which is why the
common `.a + 1` spelling never hits this; a literal reached through
`Expr::Pipe`/`Expr::Paren` inside the operand is not short-circuited and does.

## Why this is separate from #2549

Found by review of #2549 (PR #2770), whose audit initially recorded this catch-all as
closed by `owned_identity_nav_supported`'s conservative `_ => false` default — true for the
pipe/stages call site, and the audit missed the second, differently-gated one. #2549 is a
triage/documentation ticket and says explicitly: "If NOT closed (a real gap), file it as
its own separate correctness issue — don't fold a live bug discovery into this triage
ticket." This is that issue. #2770 corrects its table to point here and pins the divergence
rather than claiming the site is closed.

## Suggested fix direction

Give `owned_identity_step` an `Expr::Literal` arm that pushes the literal with a detached
identity, matching what `owned_identity_operand`'s own top-level short-circuit already
does for the bare case — then the two guards' admitted sets agree by construction.

Worth checking while there, since the same asymmetry could hide more than one shape:
diff `owned_identity_operand_supported`'s admitted set against `owned_identity_step`'s arms
in full, not just for `Literal`. `#2770`'s test
`expr_dispatch_catchall_guards_default_conservatively_2549` is the place to turn that into
an assertion once the arm exists — the property to pin is "everything either guard admits
has an arm", not "these particular shapes are rejected".

## Related

- #2549 / #2770 — the audit that found it.
- #2182, #1401 — the catch-all exhaustiveness thread this sits under.

newhoggy:
## Summary
- `owned_identity_step` (`src/jq/eval_generic.rs`) is reached from two call sites gated by different predicates: `owned_identity_nav_supported` (pipe/stages route) and `owned_identity_operand_supported` (operand route). They disagree on exactly one shape, `Expr::Literal` — admitted by the operand guard (a legitimate operand of a ruled arithmetic/comparison stage, e.g. `.a + 1`) but not by the nav guard (a literal is not navigation). `owned_identity_step` has no arm for it, so a literal reached through a `Pipe`/`Paren` inside the operand aborts the process with exit 101 whenever a downstream metadata read (`key`/`path`/`parent`/`file_index`) enters the identity route: `.a | ((.b | 1) + 2) | key`.
- The bare-literal spelling (`.a | (1 + 2) | key`) never hit this — `owned_identity_operand`'s own top-level short-circuit handles it before stepping begins.
- Fixed by giving `owned_identity_step` the identical treatment for the pipe-reached case, so both spellings agree by construction.
- Found by review of #2549 (PR #2770), whose audit initially recorded this catch-all as closed — true for the pipe/stages call site alone, missing the second, differently-gated one. This PR flips the audit's own pinned test back to closed and adds the positive property + a process-level regression guard.

Implementation follows a fully-specified plan from the issue's own comment (root cause, exact fix, full guard-vs-arm diff, and expected outputs for every trigger shape, all independently re-verified live against jq 1.7.1 and yq v4.53.3 during this PR).

## Test plan
- [x] Live-verified every repro and must-not-change row against `/usr/bin/jq` 1.7.1 and yq v4.53.3
- [x] Local patch coverage: 95% (19/20) — the one uncovered line is the `unreachable!()` panic arm itself, unreachable by construction

Fixes #2771

### DIFF
--- a/src/jq/eval_generic.rs
+++ b/src/jq/eval_generic.rs
@@ -21398,14 +21398,25 @@ fn owned_identity_step<S: EvalSemantics, V: DocumentValue>(
             owned_identity_recurse_step::<S, V>(value, id, out);
             Ok(())
         }
-        // #2549 audit: `owned_identity_nav_supported`'s own `_ => false` keeps
-        // the *pipe/stages* route from reaching here. It is **not** the only
-        // gate: `owned_identity_operand` reaches this same dispatch under
-        // `owned_identity_operand_supported`, which admits `Expr::Literal` --
-        // a shape with no arm below. `.a | ((.b | 1) + 2) | key` aborts the
-        // process today; tracked as #2771, pinned by
+        // #2771: a literal has no position of its own. Reached only from
+        // `owned_identity_operand` (its top-level `strip_parens` short-circuit
+        // already handles the bare spelling; `((.b | 1) + 2)` arrives here
+        // through the `Pipe` arm above instead). `optional` is irrelevant: a
+        // literal cannot fail. Not added to `owned_identity_nav_supported` --
+        // a literal is not navigation, and admitting it there would route a
+        // pipe that no longer needs a node through the materializing path.
+        Expr::Literal(lit) => {
+            out.push((literal_to_owned(lit), OwnedIdentity::detached()));
+            Ok(())
+        }
+        // Closed by the union of `owned_identity_nav_supported` and
+        // `owned_identity_operand_supported`, both `_ => false` -- every
+        // shape either guard admits has an arm above, `Expr::Literal`
+        // (#2771) included. Pinned by
         // `expr_dispatch_catchall_guards_default_conservatively_2549`.
-        _ => unreachable!("owned_identity_nav_supported admits no other shape"),
+        _ => unreachable!(
+            "neither owned_identity_nav_supported nor owned_identity_operand_supported admits any other shape"
+        ), // omni-dev: coverage tolerate-line reason="unreachable by construction: every shape either guard admits now has an arm above (#2771), and expr_dispatch_catchall_guards_default_conservatively_2549 pins both guards' `_ => false` defaults directly"
     }
 }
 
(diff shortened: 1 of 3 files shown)


newhoggy:
Ran a 5-way code review round. All finder angles (removed-behavior audit, cross-file call-graph trace, cleanup/altitude, and a line-by-line diff scan) converged on the same conclusion: no correctness bug in the fix itself — the new `Expr::Literal` arm is a precedented lift of logic already present in `owned_identity_operand`, and every downstream consumer of a detached `OwnedIdentity` already handles the shape correctly.

Applied the two genuinely worthwhile findings:
1. Factored the duplicated `(literal_to_owned(lit), OwnedIdentity::detached())` pairing into one `owned_literal_identity` helper, used at both call sites — matching this codebase's own "duplicated predicates diverge silently" lesson (#106).
2. Broadened the yq-mode regression test to cover `parent`/`file_index`/the combined-array shape, matching the jq test's coverage (a review found the yq twin only checked `path`/`key`).

The remaining findings (a suggestion to structurally link the two gating predicates rather than patch the one gap found, and a note that the library-level unit test only asserts non-panic while the CLI tests own value-correctness) are architectural observations rather than defects — the latter is this codebase's own established split-testing convention (a panic in the library is a failed assertion, a panic in the binary is a crash), and the former is a larger refactor out of scope for a targeted bug fix.

Full test suite + clippy re-verified clean; patch coverage 95.83% (23/24), the one uncovered line is the `unreachable!()` panic arm itself, unreachable by construction.

Issue #2811, opened 2026-09-11
yaml: #798 standalone-comment attribution — compact/nested item heads go to the first key, and column-0 EOF/dedent comments go to PREV instead of the node at that column


newhoggy:
**Severity: Low**

## Summary

Two attribution gaps in #798's standalone-comment capture, found by differential fuzzing
of the `head_comment`/`foot_comment` getters against pinned yq v4.53.3 while doing
#1079's deferred half. Both are pre-existing (they reproduce with no `- # c` item in
the document) and neither is user-visible until #2795 prints head/foot comments — but
#2795's rendering will inherit whichever slot the parser chose, so they should land
before or with it.

### 1. A block above a compact-mapping / nested-sequence item attaches to the first key / inner item, not the item

```
$ printf -- '- 1\n# h\n- b: 1\n' | yq '.[1] | head_comment'          # h
$ printf -- '- 1\n# h\n- b: 1\n' | yq '.[1].b | key | head_comment'  # (empty)
$ printf -- '- 1\n# h\n- b: 1\n' | succinctly yq '.[1] | head_comment'          # (empty)
$ printf -- '- 1\n# h\n- b: 1\n' | succinctly yq '.[1].b | key | head_comment'  # h

$ printf -- '- 1\n# h\n- - 1\n' | yq '.[1] | head_comment'      # h
$ printf -- '- 1\n# h\n- - 1\n' | succinctly yq '.[1][0] | head_comment'  # h
```

Cause: `parse_sequence_item_inner` only calls `attach_head_foot_at` in its plain-scalar
arm (and, since #1079, for a bare `-` whose value is deferred); the compact-mapping arm
defers to `parse_compact_mapping_entry`'s key open and the `- - x` arm to the inner
item's own open. Real yq answers `.[1] | head_comment` directly for both. Note the
asymmetry that must be kept: an *anchored* item's block does go to the key
(`- 1\n# h\n- &x\n  b: 1\n` is `.[1].b | key | head_comment` = `h`, measured), so this is
the bare compact case only.

Also inherited by #1079's floated comments (`- # c\n- b: 1\n`), pinned as a residual in
`test_absent_item_comment_residuals_1079_float`.

### 2. A column-0 comment at end of document (or before a dedent) after nested content goes to the outermost node at that column, not `PREV`

```
$ printf -- 'a:\n  b:\n    - 1\n# c\n' | yq '.a | key | foot_comment'       # c
[...]

Issue #2795, opened 2026-09-11
yq: identity round-trip drops standalone comments — no emitter prints head/foot


newhoggy:
Spun out of #798, which is being closed as an umbrella. The parser captures
standalone comment lines (#2704, #2718) and #2758 added the getters to read them —
but **no emitter ever prints them**, so every YAML-to-YAML route silently discards
them. This is data loss on a round trip, and it is the largest remaining
user-visible gap in the cluster.

Recorded in `docs/compliance/yq/limitations.md` (#2758), untracked by any issue
until now.

## Measured

Input:

```yaml
# lead
a: 1
# mid
b: 2

# trail
```

Against pinned `yq` v4.53.3 and `succinctly` at `54658bca0`:

| route | yq | succinctly |
|---|---|---|
| `yq '.'` | all three kept | **all dropped** |
| `yq -P '.'` | kept | dropped |
| `yq '.a = 5'` | kept | dropped |
| `yq 'del(.b)'` | `# lead`, `# trail` kept; `# mid` goes with `b` | dropped |
| `yq 'select(true)'` | kept | dropped |
| `yq -o=json '.'` | dropped | dropped — **correct**, JSON has no comments |

`del(.b)` dropping only `# mid` is a useful confirmation of the ownership model:
that comment belongs to `b`'s key, so deleting the entry deletes it.

## Two independent halves

The two emitters do not share a comment source, so this splits cleanly into two PRs.

### PR A — the streaming emitter (this is what plain `yq '.'` uses)

`YamlCursor::stream_yaml_value_at` in `src/yaml/light.rs` reads comments straight
off the live cursor at ~9 sites (`write_line_comment(out, self_.line_comment_raw())`
and friends) and **never consults `NodeMeta`**. #2758 already added
`head_comment_raw`/`foot_comment_raw` there, so the reads exist; what is missing is
`write_head_comments`/`write_foot_comments` threaded through those sites at the
right indent. This half closes the headline gap.

### PR B — the DOM emitter (writes, `-P`)

`emit_yaml_value_at_depth` in `src/bin/succinctly/yq_runner.rs` already consults
`NodeMeta` for the trailing comment, style and anchor, and
`append_own_comment_line` is already multi-line aware — so this is smaller than it
looks, but it needs three things:

- **Populate `NodeMeta`**: `to_owned_with_comments_at_depth`
  (`src/jq/eval_generic.rs`) hardcodes `head_foot_comment: None`. The field is
[...]

PR #2824, opened 2026-09-12
feat(yaml): print standalone comments and the verbatim header on the streaming yq route (#2795)

newhoggy:
Closes the streaming half of #2795 (plain `yq '.'`, `select(true)`, every cursor-forwarded result). The DOM half (`-P`, writes) is a follow-up PR stacked on this one; #2811 (parser attribution) is landing separately from its own worktree and this branch will be rebased onto it — `src/yaml/parser.rs` is untouched here.

## What changes

- **Head/foot comments print on the streaming route.** `stream_yaml_as_document` prints a document's own leading block and foot; the block-mapping and block-sequence loops in `stream_yaml_value_at` print each key's/item's head above it at its own indent and its foot below it, with real yq's one blank line after a foot that a sibling follows and none after a foot that ends its collection. A blank line *inside* a head block is derived from the source bytes between its lines (`blank_line_between` on the cursor side), not stored. A scalar root prints its head and drops its foot, as yq does.
- **The verbatim header.** Not in the issue's probe table: yq's default `--header-preprocess` re-emits the file's prefix verbatim — every leading blank line, `#` line at any indent, `%YA..` directive or `---` marker up to the first content line — whenever the result is the first document itself. That is what keeps `# lead\n\na: 1`'s blank line and an explicit leading `---` (`---\na: 1` round-trips as `---\na: 1`; measured, and the old `--doc 0` test that pinned the marker being dropped now pins yq's output). `-N` drops the header's `---` lines, via `YamlIndex::suppress_header_doc_markers`. The header length is cached on the index (`YamlIndex::yq_header`) so a comment range the header already covered never prints again as a node's head — the fuzz caught `#c1\n---\n\n# c2` printing `# c2` twice, because the header runs past `---`.
- **Cost on comment-free input is gated.** `YamlIndex::has_standalone_comments` is computed at build; every head/foot site is one predictable branch when it is false. Before the gate the M4 Pro A/B read +2.1% median / +5.4% on sequences; after it, +1.35% median (range +0.9..+2.2%, 12/12 rows slower) against a `--control` run of −0.75% (range −1.4..+1.0%). The residual is small and consistent; 7950X numbers to follow in a comment.

## Measured, not stated

Every expected string in the tests and goldens was captured from pinned yq v4.53.3. The 11 new goldens under `tests/data/yq-golden/cases/standalone_comments_*_2795/` were written by `scripts/sync-yq-golden.sh`, never by hand; the three pre-existing tests whose expectations encoded the old dropping (`test_block_scalar_does_not_steal_following_comment_710`, its empty-scalar twin, `test_doc_select_first`) now match yq byte-for-byte.

`scripts/yq-comment-oracle-fuzz.py` is the differential gate. It generates small block documents with `#` lines at every position and column the parser can attribute differently, runs both binaries on `.`, `-P .`, `.a = 5`, `del(.b)`, `select(true)`, `-o=json .` and on the getters, and classifies each mismatch by direction. On the identity route, seed 7, 300 documents (baseline at the merge-base in parentheses):

| class | count | meaning |
|---|---:|---|
| `same` | 122 (16) | byte-identical |
| `we-emit-yq-doesnt` | **0** | a `#` line neither in yq's output nor in the input — the fabrication class, the gate |
| `yq-drops-we-keep` | 21 | a comment from the input yq's printer lost and we kept (recorded in `limitations.md`; ADR-0018 rule 4) |
| `yq-emits-we-dont` | 18 (246) | comments still dropped — all #2811 attribution shapes |
| `placement-differs` / `blank-differs` | 58 / 19 | same lines, different node or indent — #2811 attribution (column rule, compact-item head) |
| `values-differ` | 62 (38) | non-comment text differs: the pre-existing empty-document `null` rendering and yq's trailing comma on a commented flow node; the 24 new ones are the `- - x` compact-nested arm printing `- #c\n  - x` where #2811 moves the comment to the outer item |

`-o=json` stays identical on every generated document.

## Verification

- `cargo test --features cli`, `--no-default-features`, `--features simd,scalar-yaml`: green.
- `cargo clippy --all-targets --all-features -- -D warnings` and the second CI leg: clean. `cargo fmt --check`, `cargo doc` with `RUSTDOCFLAGS=-D warnings`: clean.
- `omni-dev coverage diff`: 94.4% patch (338/358); the uncovered lines are the `)?;` error arms of `fmt::Write` calls (never taken with a `String` sink, same as every existing multi-line `?` call in `light.rs`).
- Interleaved A/B on `johns-mac-mini` (M4 Pro), release profile, comment-free `users`/`nested`/`sequences` at 1 MB and 10 MB, `.` and `.[]` with `-o yaml -I=2`, 9 reps, output identity 12/12 — numbers above.

## Docs

`docs/compliance/yq/limitations.md` rewrites the "standalone comments are dropped" entry into what the streaming route now reproduces and the residual divergences (comments yq loses, the `--- # c` header line yq glues into unreadable output, attribution pending #2811, DOM route pending); `docs/reference/yq-language.md`'s stale "comments not stored" line is corrected.


newhoggy:
7950X (`terminus`, WSL2, `taskset -c 0`, same corpus/queries/reps, release profile, output identity 12/12), gated build vs merge-base:

| workload | before med | after med | med Δ | min Δ |
|---|---:|---:|---:|---:|
| nested-10mb `.` | 123.4m | 121.7m | −1.4% | −1.4% |
| nested-10mb `.[]` | 124.0m | 122.0m | −1.6% | −1.9% |
| nested-1mb `.` | 13.9m | 13.7m | −1.3% | −0.8% |
| nested-1mb `.[]` | 13.9m | 13.8m | −0.8% | −1.1% |
| sequences-10mb `.` | 171.0m | 182.0m | +6.4% | +1.6% |
| sequences-10mb `.[]` | 171.1m | 172.7m | +0.9% | +1.0% |
| sequences-1mb `.` | 18.1m | 18.5m | +2.1% | +0.6% |
| sequences-1mb `.[]` | 18.2m | 19.3m | +5.5% | +0.7% |
| users-10mb `.` | 198.0m | 198.8m | +0.4% | +0.2% |
| users-10mb `.[]` | 196.1m | 195.3m | −0.4% | −0.6% |
| users-1mb `.` | 21.0m | 21.0m | +0.3% | +0.6% |
| users-1mb `.[]` | 20.8m | 20.7m | −0.4% | −0.2% |

median of medians **−0.06%** (mean +0.81%, 6/12 rows slower); `--control` on the same box: −0.29% (range −1.0..+0.8%). The two large sequences medians (+6.4%, +5.5%) sit on mins of +1.6%/+0.7% — tail noise rather than a shift, unlike the pre-gate run on this box (+1.68% median, 9/12 slower). Before the gate, the ungated build read +1.68% here and +2.09% on the M4 Pro; after it, −0.06% here and +1.35% on the M4 Pro (control −0.75%). So the gate recovers the 7950X fully and most of the M4 Pro; the ~1% M4 Pro residue is the one branch per node plus the per-document header probe, and is the number to beat if anyone wants to chase it further.

newhoggy:
Rebased onto #2825 (`c0d73d068`) and retargeted the PR base to its branch so the diff is this commit alone; merge after #2825 lands and the base will move back to `main`. With the parser attribution in place, the identity-route fuzz on the same seed reads 156/300 byte-identical (was 122), zero fabrications, and every `- - x` compact-nested case now matches — the remaining classes are comments yq's own printer loses (kept, recorded), blank-line text inside getter strings (`getter-blank-text`, now classified separately), and one residual attribution shape reported to #2825.

Issue #2724, opened 2026-09-10
jq/yq: object-construction shorthand {$a} is a parse error — jq builds {"a":1}


newhoggy:
**Severity: Low**

## Summary

`{$a}` in *object construction* position (sugar for `{a: $a}`, key inferred from the
variable name) is rejected by the parser. #1139 fixed the same shorthand in
*destructuring pattern* position (`. as {$a}`); construction was never covered and is
still broken.

## Repro

Confirmed live against jq 1.7.1 and on `main` (`7d84b2329`):

```console
$ echo null | jq -c '1 as $a | {$a}'
{"a":1}
$ echo null | succinctly jq -c '1 as $a | {$a}'
jq: compile error: parse error at position 11: expected identifier, found '$'
```

Destructuring, for contrast, works today (that is #1139, closed):

```console
$ echo '{"a":1}' | succinctly jq -c '. as {$a} | $a'
1
```

yq mode diverges too, though not identically — real yq v4.53.3 *accepts* the program and
emits nothing at all (exit 0, empty output), where succinctly errors:

```console
$ printf 'x: 1\n' | yq '1 as $a | {$a}'          # exit 0, no output
$ printf 'x: 1\n' | succinctly yq '1 as $a | {$a}'
Error: parse error: parse error at position 11: expected identifier, found '$'
```

So the two modes need separately-captured expectations: jq's `{"a":1}` vs yq's silent
empty. Whoever takes this should re-probe yq's behaviour rather than assuming it is a
plain accept.

`$__loc__` has the same shape (`{$__loc__}` is a parse error here, `{"__loc__":{...}}`
in jq) but interacts with #2688, which is about `$__loc__`'s own contents.

## Root cause

The object-construction parser (`src/jq/parser.rs`) expects an identifier, a string, or
a parenthesized computed key at each entry, and has no arm for a bare `$var` entry.
This is the construction-side mirror of the `parse_pattern` gap #1139 closed.

## Why this is separate from #2560

Entirely unrelated mechanism (parser, not `bind_def_call`'s argument binding); found
incidentally while fuzzing duplicate-parameter programs for #2560 (PR #2722). Present
identically on `main`.

## Note

Not currently recorded in `docs/compliance/jq/limitations.md`.

newhoggy:
🔧 Claimed by agent `Agatha` — branch `issue-2724-object-shorthand-var`

PR #2826, opened 2026-09-12
fix(jq): accept a trailing comma in object construction

### LINKED ISSUE #2788: jq: trailing comma in object construction {a: 1,} is rejected; real jq accepts it
## Summary

Surfaced during #2724's code review (independently confirmed by two review
passes). Real jq allows a trailing comma before the closing `}` in object
*construction* -- `{a: 1,}` and `{a,}` both parse and evaluate normally.
succinctly's `parse_object_construction` loop always re-attempts a key
parse after consuming a comma, so a trailing one hits "expected
identifier, found '}'" instead.

```console
$ echo null | jq -c '{a:1,}'
{"a":1}
$ echo null | succinctly jq -c '{a:1,}'
jq: compile error: parse error at position 5: expected identifier, found '}'
```

This is narrower than "trailing comma everywhere" -- confirmed live that
real jq does **not** extend the same allowance to array construction
(`[1,2,]` is a compile error in real jq too, matching succinctly already)
or to destructuring patterns (`. as {$a,} | $a` is also a compile error in
real jq, matching succinctly already). It's specific to object
*construction*'s own grammar production.

## Why this is separate from #2724

Pre-existing on `main`, unrelated to #2724's `{$a}`/`{$a: expr}` shorthand
mechanism -- the same "expected identifier, found '}'" already fires for
`{a:1,}` before that PR's changes. #2724 only made it newly reachable
through a natural-looking new construct (`{$a,}`), with no test coverage
for that shape.

## Suggested next step

In `parse_object_construction`'s loop (`src/jq/parser.rs`), after consuming
a `,` separator, peek for an immediate `}` and break out of the loop
instead of re-entering key parsing.

newhoggy:
## Summary
- Real jq's object-construction production alone accepts a single trailing comma before the closing `}` — `{a:1,}` parses and evaluates identically to `{a:1}`. `parse_object_construction`'s loop always re-entered key parsing after consuming a comma, so a trailing one hit `expected identifier, found '}'` instead.
- jq mode only: real yq rejects the same shape too, just for an unrelated reason of its own (`',' expects 2 args but there is 1`), and array construction / destructuring patterns don't share this allowance in real jq either. A doubled comma still re-enters key parsing and hits the existing error unchanged.
- Surfaced during #2724's code review (independently confirmed by two review passes); pre-existing on `main`, unrelated to #2724's own mechanism.

## Test plan
- [x] Live-verified every repro and must-not-widen row against `/usr/bin/jq` 1.7.1 and yq v4.53.3
- [x] Local patch coverage: 100% (23/23 new lines)

Fixes #2788

### DIFF
--- a/docs/reference/jq-language.md
+++ b/docs/reference/jq-language.md
@@ -90,7 +90,7 @@ below and [ADR-0019](../adrs/adr-0019.md).
 - [x] `to_entries` / `from_entries` / `with_entries(f)`
 - [x] `pick(keys)` - select only specified keys (yq)
 - [x] `omit(keys)` - remove specified keys (inverse of pick, yq)
-- [x] Object construction: `{foo: .bar}`, `{(expr): value}`, shorthand `{foo}`
+- [x] Object construction: `{foo: .bar}`, `{(expr): value}`, shorthand `{foo}` -- a single trailing comma before `}` is accepted in jq mode only (`{foo: 1,}`, #2788), matching real jq; real yq rejects it
 
 ### Array Operations
 - [x] `length`
--- a/src/jq/parser.rs
+++ b/src/jq/parser.rs
@@ -1531,6 +1531,20 @@ impl<'a> Parser<'a> {
             match self.peek() {
                 Some(',') => {
                     self.next();
+                    self.skip_ws();
+                    // #2788: jq's object-construction production alone
+                    // accepts a trailing comma (`{a:1,}`) -- array
+                    // construction and destructuring patterns don't, and
+                    // neither does real yq (`',' expects 2 args but there
+                    // is 1`), so this is jq-mode only. A doubled comma
+                    // (`{a:1,,}`) still re-enters key parsing below and
+                    // hits its existing "expected identifier" error --
+                    // this only short-circuits the single-trailing-comma
+                    // shape.
+                    if self.mode == ParserMode::Jq && self.peek() == Some('}') {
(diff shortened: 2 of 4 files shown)


newhoggy:
Ran a 4-way code review round (line-by-line diff scan, removed-behavior audit, cross-file call-graph trace, and reuse/simplification/altitude/conventions). All four converged: no correctness bug found. Every edge case probed (multi-entry, `$`-shorthand, dynamic keys, doubled/leading comma, comment-before-close, array/pattern non-leakage, yq-mode gating) matches real jq/yq exactly.

The only substantive suggestion — a regression test pinning that object *patterns* (`. as {$a,} | $a`) still reject a trailing comma, so a future shared-refactor of the three near-identical comma loops (object construction / object pattern / array pattern) can't silently widen the wrong one — is already covered by this PR's own `test_object_construction_trailing_comma_does_not_widen_2788`.

Remaining findings (a suggestion to factor the three comma loops into one shared helper, and a note that `skip_ws()` runs twice in a row on the non-trailing-comma path) are architectural/micro-perf observations, not defects — the former is out of scope for this issue's intentionally narrow 5-line fix (per the implementation plan), and the latter's `skip_ws()` call is actually load-bearing for the `{a:1, }` (trailing whitespace before `}`) case, not truly redundant.

No changes made; full test suite + clippy already clean, patch coverage 100%.

Issue #2817, opened 2026-09-11
yq: `//` is per-left-output and keeps a falsy left value when the right side is empty — succinctly yq follows jq's collect-then-fallback


newhoggy:
**Severity: Medium** (real yq syntax, silent wrong answer in `succinctly yq`; pre-existing, surfaced by #2782's widening of `scripts/jq-path-context-oracle-sweep.sh`'s alphabet with `//` wrappers)

## Summary

`succinctly yq` evaluates `L // R` with jq's rule — collect `L`'s outputs, keep the truthy ones, and run `R` only if none survived. Real yq v4.53.3 does something different on two axes, captured live:

**1. Per left output, not collect-then-fallback.** Each falsy output of `L` is replaced by `R`'s outputs *in place*; truthy ones pass through in order.

| filter (`yq -n -o=json`) | yq v4.53.3 | succinctly yq (= jq 1.7.1) |
|---|---|---|
| `(1, null) // 2` | `1` `2` | `1` |
| `(null, 1) // 2` | `2` `1` | `1` |
| `(null, false) // 2` | `2` `2` | `2` |
| `(null, 1, null) // 2` | `2` `1` `2` | `1` |
| `(false, null) // (3, 4)` | `3` `4` `3` `4` | `3` `4` |
| `(1, 2) // 3` | `1` `2` | `1` `2` |

**2. An empty right side keeps the falsy left value.** When `R` yields nothing for a falsy left output, yq emits that left output itself; jq emits nothing.

| filter (`yq -n -o=json`) | yq v4.53.3 | succinctly yq (= jq 1.7.1) |
|---|---|---|
| `{} \| (null // .[])` | `null` | *(nothing)* |
| `{} \| (false // .[])` | `false` | *(nothing)* |
| `{"a":1} \| (null // key)` (root `key` is empty) | `null` | *(nothing)* |
| `. \| (null // parent)` at the root | `null` | *(nothing)* |
| `{"a":1} \| (null // .[])` | `1` | `1` |

## Where it shows

[...]

Issue #2665, opened 2026-09-09
jq: #2476's native and/not/alternative/any-all arms duplicate validation-gating and truthy-check logic


newhoggy:
Follow-up from the review of PR #2660 (#2476). Four cleanup items, none of them correctness bugs — filed together since they're all instances of the same underlying pattern: the PR added four independent native arms (`Expr::And`/`Expr::Or` via `eval_boolean_generic`, `Expr::Not`, `Expr::Alternative`, `Builtin::Any`/`Builtin::All` via `any_all_generic`) that each re-derive, rather than share, the same small pieces of logic.

## 1. The "charge the ambient validation walk" policy is hand-copied into 4 sites

`src/jq/eval_generic.rs`: `eval_boolean_generic`'s own gate, the `Expr::Not` arm, the `Expr::Alternative` arm, and `any_all_generic`. Each independently decides whether/how to call `ambient_validation_error`/`push_generic_document_validation_error`, justified by its own standalone doc comment, rather than one shared policy function or table.

This is more than tidiness: it's what let a real structural gap ship silently. `needs_path_context` and `path_context_single_native` gained an `Expr::And | Expr::Or` recursive arm back in #1405/#2473, but #2476's three new native constructs — `Expr::Alternative`, `Expr::Not`, `Builtin::Any`/`Builtin::All` — have no matching arm in either function; they fall to those functions' `_ => false` / `_ => !needs_path_context(expr)` defaults. Confirmed live and pre-existing but newly exposed by this asymmetry: `.a | (key // 1)`/`.a | to_entries | .[] | (key // 99)` are documented gaps in the merged PR's own test comments (`test_alternative_operands_read_the_real_position_2476`). A future 5th bridged construct (`while`/`until`/`isvalid`, named as follow-ups in that PR's CHANGELOG) has no single place to consult for getting this right the first time.

**Suggested fix**: give `path_context_single_native` and `needs_path_context` matching arms for `Expr::Alternative`, `Expr::Not`, and `Builtin::Any`/`Builtin::All` (mirroring the existing `Expr::And | Expr::Or` arms), and consider centralizing the "which constructs used to bridge and need the ambient walk" decision into one place both the routing functions and the eval arms consult.

## 2. Redundant double validation walk when the `//`/`and`/`or` operand is the ambient position

[...]

Issue #2435, opened 2026-09-04
yq: reading an absent path vivifies it in real yq -- .a.b | parent shows {b: null} and .a[5] | key on a mapping is 5; succinctly keeps the document as read


newhoggy:
Filed against spine 2416 (hygiene rule 1; number without a hash on purpose). Found while capturing absent-node shapes for the Phase 2 sink walk, which has to model an absent position rather than refuse it.

## Capture (yq v4.53.3, document `a: {}` / `z: [1]`)

```console
$ yq -o=json -I0 '.a.b | parent'
{"b":null}
$ succinctly yq -o=json '.a.b | parent'
{}

$ yq -o=json -I0 '.a.b | parent | parent'
{"a":{"b":null},"z":[1]}
$ succinctly yq -o=json '.a.b | parent | parent'
{"a":{},"z":[1]}

$ yq -o=json -I0 '.a[5] | key'
5
$ yq -o=json -I0 '.a[5] | path'
["a",5]
$ succinctly yq '.a[5] | key'
Error: Cannot index object with number
```

Real yq's traversal of a missing key *creates* it (`b: null`) in the in-memory document, so a later `parent` shows the vivified child, and indexing a mapping with a number is accepted (it is what yq's `.a[5] = x` would create). succinctly navigates without mutating and applies jq's "Cannot index object with number" rule in yq mode.

The rows that already agree: `.a.b` → `null`, `.a.b | key` → `"b"`, `.a.b | path` → `["a","b"]`, `.a.b.c | parent | key` → `"b"`, `.x.y | key` → `"y"`, `[.a.b | parent | key]` → `["a"]`.

## Why parked

The sink walk (Phase 2) will represent an absent position as `PathNode::Absent`-style state so it can emit `null` with its path and hop to the real ancestor — matching today's eager evaluator, i.e. the *non-vivifying* answer. Matching yq here means either mutating the read model (which succinctly never does) or synthesising the vivified view at `parent`; both are design decisions for the walk's `Parent` arm, not a standalone fix. The `Cannot index object with number` row is a separate yq-mode indexing rule (`.a[5]` on a mapping) that lives in `path_step_generic`/`eval_index` and is upstream of path context; capture more shapes (`.a[5]` alone, `.a[5] = 1`, `.a["x"]` on a sequence) before touching it.

ADR-0018: mode decides, so both are divergences to document in `docs/compliance/yq/limitations.md` if they outlive a release (hygiene rule 3).

PR #2827, opened 2026-09-12
fix(jq): make // a transparent construct so it agrees with the bare spelling

### LINKED ISSUE #2782: jq: path-context // loses parent/path/file_index and truncates a multi-output left operand — blocks #2665 item 1
**Severity: Medium** (latent today — reachable only once `//` is routed through the path-context walk, which is exactly what #2665 item 1 asks for)

## Summary

`needs_path_context` has no `Expr::Alternative` arm, so `//` never routes through the
path-context walk and a `key`/`parent`/`file_index` inside it silently resolves to nothing.
#2665 item 1 asks for that arm. **Adding it cannot be done safely today**: the walk's own
`//` handling mishandles two shapes, and the routing change is what exposes them.

Measured by adding the arm on a branch and diffing 61,536 (program, document) pairs against
`main`. 670 rows improve — the reported gap closes — but these regress:

| filter | `main` | with the arm |
|---|---|---|
| `.a \| (parent // 1)` on `null` | `{}` | `1` |
| `.a \| (parent // empty)` on `null` | `{}` | *(nothing)* |
| `.a \| (parent // file_index)` on `null` | `{}` | `0` |
| `.a \| ((key//9) \| parent)` | `{"a":{"b":1}}` | *(nothing)* |
| `.a \| ((key//9) \| path)` | `["a"]` | `[]` |
| `.a \| (file_index // .a\|key)` | `"a"` | *(nothing)* |
| `.a \| (.a\|key // .a\|key)` | `"a"` | *(nothing)* |
| `.a \| ((1,2) // parent)` on `null` | `1` `2` | `1` |

(`{"a":{"b":1}}` unless noted. The bare spellings are unaffected and correct throughout:
`.a | parent` is `{"a":{"b":1}}`, `.a | path` is `["a"]`, `.a | file_index` is `0`.)

Two distinct defects:

1. **A path reader in the left operand stops resolving** once the walk owns the expression —
   `parent`, `path` and `file_index` all come back empty or falsy, so `//` falls through to
   the right operand (or emits nothing). `key` is the one that *works*, which is why the
   gap looked like a simple missing arm.
2. **A multi-output left operand is truncated to its first output** — `(1,2) // parent`
   gives `1` where every other route gives `1 2`. Narrow: a `null` document at top level;
   `[(1,2) // parent]` and non-`null` documents are fine, which locates it in the streaming
   arm rather than in the operator's semantics.

## Reproducing

Apply to `needs_path_context` (`src/jq/eval.rs`), beside the existing `Expr::And | Expr::Or`
arm, and the table above follows:

```rust
Expr::Alternative(left, right) => needs_path_context(left) || needs_path_context(right),
```

Recursing into the left operand alone still produces rows 1–7; the right-operand half is
what produces row 8.

## Also blocked: `any`/`all`

#2665 item 1 names `Builtin::Any`/`All` alongside `//`. Adding routing arms for them changes
**nothing observable**, because their evaluation arm does not thread the position into the
operands either: `.a | any(key; .)` is `false` where `.a | key` is `"a"`, with and without
the arm. That is an evaluation-side fix of its own, not a routing one.

## Why this is separate from #2665

#2665 is a four-item cleanup ticket whose item 1 proposes the routing arms. This is the
evaluation-side work those arms depend on. Filing separately rather than widening #2665, and
#2665 is marked `Blocked` on this with the same evidence.

## Suggested fix direction

Start with defect 1, since it is what makes the arm unsafe at all: find why the walk's `//`
arm resolves `key` but not `parent`/`path`/`file_index` in the left operand — most likely
the operand is evaluated at a position the walk has already moved past, or through a route
that drops the frame those three read while `key` is recovered some other way. Defect 2 is
plausibly the same arm's streaming half not draining the left operand before deciding
truthiness.

The 61,536-pair differential above is the right gate for the fix: with both defects closed,
adding the routing arms should show **only** the 670 improving rows.

newhoggy:
## Summary

- `needs_path_context` had no `Expr::Alternative` arm, so a pipe whose only path-context read sat inside a `//` was never routed to path-context evaluation at all (`.a | to_entries | .[] | (key // 99)` answered `99` where bare `.[] | key` is `0`).
- Adding the arm alone traded that gap for several: the owned-identity route's own `//` arm rewrote a read inside either operand into a detached literal, stamped a multi-output left side with its first output's (re-evaluated) identity, and forced `optional = true` onto the left operand.
- `//` is now a fully transparent construct on that route (like `,` and `if`): `eval_owned_identity_alternative` runs the left operand as stages of the pipe, continuing `rest` from every truthy output at the position it was actually produced, splicing the right side only when nothing truthy came out.
- A `//` at a live cursor with nothing reading after it stays on the cursor-native route (`alternative_stays_cursor_native`) so yq's flow style/anchor/comments survive (`.a | (parent // 1)`).
- `owned_identity_pipe_applies` (the non-sink gate) learned that a stage with no placement rule can also be a leaving stage reached after a navigational head, so the yq CLI's DOM path and the jq CLI's sink pipeline now agree on `. | (null // .b) | key`.

Two divergences the fix's own oracle sweep surfaced are filed separately rather than chased here:
- #2817 — real yq's `//` maps per left output and keeps a falsy left value when the right side is empty (succinctly follows jq's collect-then-fallback rule in both modes).
- #2435 (pre-existing) — the null-document `parent` autovivification question, orthogonal to where a `//` output stands.

Fixes #2782 (#2665 item 1).

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 3688+ tests, all green)
- [x] `scripts/jq-path-context-oracle-sweep.sh --self-test` (alphabet coverage, extended with `alt_l`/`alt_r` wrappers)
- [x] `scripts/jq-path-context-oracle-sweep.sh --summary` against pinned jq 1.7.1 / yq v4.53.3: 6300 cases, 0 unexpected divergences, every known divergence still accounted for
- [x] New CLI tests pinning both defects and their fix in `tests/jq_cli_tests.rs` and `tests/yq_cli_tests.rs`
- [x] `docs/adrs/adr-0021.md` amendment and `docs/compliance/yq/limitations.md` entry for the #2817 divergence

### DIFF
--- a/docs/adrs/adr-0021.md
+++ b/docs/adrs/adr-0021.md
@@ -420,6 +420,48 @@ rejected: the only *safe* allowlist of "cannot read a node" was the one enumerat
 benchmarked queries. [#2789](https://github.com/rust-works/succinctly/issues/2789) carries
 the two remaining shapes a real fix could take.
 
+### Amendment (#2782, implemented 2026-09-12): `//` is a transparent construct
+
+`needs_path_context` (`src/jq/eval.rs`) had no `Expr::Alternative` arm, so a pipe whose
+only read sat inside a `//` was never routed anywhere: `.a | to_entries | .[] | (key //
+99)` answered `99` where the bare `.[] | key` is `0` (#2665 item 1). Adding the arm alone
+traded that gap for several (#2782), because the identity pass had given `//` an
+*operand-shaped* arm on the owned identity route: it answered one output per side,
+rewrote a read inside either side to a detached literal before evaluating it, stamped a
+multi-output left side with the identity of its first output (re-evaluated), and forced
+`optional = true` onto the left side — a rule its own comment attributed to jq, which jq
+does not have (`5 | (.a // 1)` is an error in 1.7.1).
+
+**`//` has no placement rule; it is run like `,` and `if`.** `owned_identity_rule` answers
+`None` for it, so a pipe that *starts* with one enters the identity route whole
+(`owned_identity_pipe_entry_stage`, the door #2563 opened for `as`), and
+`eval_owned_identity_alternative` runs the left side as stages of the pipe, continues
+`rest` from every truthy output standing exactly where it was produced — a `key` at its
+key node, a `file_index` where the node stood, each output of a comma at its own — and
+splices the right side only when nothing truthy came out. Errors, `break` and laziness
+follow `eval_single`'s cursor-native arm: a falsy output is dropped, a truthy one before an
+error survives it, the right side never runs after one, and a bounded consumer stops the
+left side early. The gate (`owned_identity_pipe_supported`) asks of each side what it asks
+of a comma branch.
+
+**A `//` at a live cursor that nothing reads after stays on the cursor route**
+(`alternative_stays_cursor_native`). There the identity route has nothing to answer that
+the cursor-native arm does not, and materializing would re-render the node `parent`
+landed on: `.a | (parent // 1)` keeps its flow style, anchor and trailing comment, as it
+did before the arm and as yq prints it. Only a read *after* the operator needs to know
+which operand produced the output, and only the identity route tracks that.
+
+The non-sink route's gate (`owned_identity_pipe_applies`) learned the same door for a
+stage with no rule reached *after* a navigational head, so the yq CLI's DOM path and the
+jq CLI's sink pipeline answer `. | (null // .b) | key` alike. `path_context_single_native`
+gained the matching `Expr::Alternative` entry.
+
+Two divergences this surfaced are not this amendment's: real yq's `//` maps *per left
+output* and passes a falsy left value through when the right side is empty (#2817, recorded
+in the sweep's manifest), and the vivified `parent` at an absent position (#2435) decides
+which side `.a | (parent // .)` takes on a `null` document. The sweep's alphabet gained
+`alt_l`/`alt_r` wrappers so a leaf on either side of `//` is part of the claim.
+
 ## Consequences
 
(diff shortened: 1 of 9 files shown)


newhoggy:
Ran an 8-way code review before merging. No correctness bugs found — three independent finder agents re-ran the 6300-case oracle sweep against live jq 1.7.1/yq v4.53.3 and confirmed 0 unexpected divergences. Addressed in a follow-up commit:

- `owned_identity_pipe_applies`'s new disjunct now delegates to the shared `owned_identity_pipe_enters_at_first` helper instead of hand-duplicating it (four reviewers flagged this as a fragile invariant).
- Fixed a stale doc-comment count on `mark_nonretryable_escape`.
- Added test coverage for two review-verified-correct-but-untested shapes: `//` nested inside an arithmetic/and/or operand, and a `rest`-side error after a truthy `//` output staying correctly attributed.

Filed as follow-ups rather than risked in this PR:
- #2829 — an unconditional `value`/`id` clone on every `//` evaluation (a real perf regression on the previously-zero-clone common case), needs an A/B benchmark before any fix.
- #2830 — the "stash rest's escape" idiom is now duplicated across four owned-identity drivers; pure cleanup, no correctness impact.
- #2831 — the oracle sweep's `alt_l`/`alt_r` wrappers don't compose with a multi-output `rest`, missing an already-known #2817-class divergence.

Rebased on latest `main` (picked up #2665's overlapping `cursor_is_truthy` refactor commits) and re-verified the full suite + both clippy configs + the full oracle sweep after the rebase — all green.

Issue #1982, opened 2026-08-31
yq: DOM output path diverges from the streaming path on !!str tags and \L/\P escapes


newhoggy:
**Severity: Medium** (part 1 is a silent type change; part 2 is output fidelity)

## Summary

Found while fixing #1700. `succinctly yq`'s DOM (`OwnedValue`) output path disagrees with
*both* real yq and succinctly's own M2 streaming path on two unrelated points. Every route
that forces materialization reaches it: `-P` *without* `-I0`, `--arg`/`--argjson`,
`-r`/`-j`/`-0`, `--eval-all`, `--split-exp`, and any filter shape `can_use_m2_streaming`
rejects. (Corrected from this issue's first revision, which listed a bare `-P` and
`--slurpfile`: `-P -I0` in fact *streams*, because `-I0`'s own `compact` already satisfies
the fast-path gate, and `--slurpfile` is a jq flag `succinctly yq` does not accept at all —
it exits 2 with "unexpected argument". Each route above re-verified live against the
binary by whether a duplicate mapping key survives.)

Pinned against Homebrew `yq` v4.53.3, the repo's own oracle.

### 1. An explicit `!!str` tag is dropped, changing the value's type

```console
$ printf 'a: !!str 5\n' | yq -o json -I0 '.'                    # real yq
{"a":"5"}
$ printf 'a: !!str 5\n' | syq -o json -I0 '.'                   # succinctly, streaming path
{"a":"5"}
$ printf 'a: !!str 5\n' | syq -P -o json '.'                    # succinctly, DOM path
{
  "a": 5
}
```

`!!str 5` is the *string* `"5"`, and both real yq and succinctly's streaming path say so
(that is #224's rule). On the DOM path the tag is lost and the scalar re-resolves as a
number, so a `-P` or `--arg` in the command line silently changes a value's type. This is
the half worth fixing first — it is data corruption, not a rendering difference, and it
survives a round-trip.

### 2. `\L` / `\P` lose their `\u2028` / `\u2029` escaping

```console
$ printf 'a: "x\\Ly\\Pz"\n' | yq -o json -I0 '.'             # real yq
{"a":"x\u2028y\u2029z"}
$ printf 'a: "x\\Ly\\Pz"\n' | syq -o json -I0 '.'            # succinctly, streaming
{"a":"x\u2028y\u2029z"}
[...]

PR #2828, opened 2026-09-12
fix(yq): match real yq's -I0 -> 4-space YAML width

### LINKED ISSUE #2606: yq: -I0 indents 2 spaces, real yq indents 4
**Severity: Low** (cosmetic output divergence, no data loss)

Found while checking an assumption during #1448's benchmarking work.

`succinctly yq -I0` renders 2-space block indentation; real yq's `-I0` renders
**4**-space. Every other width I checked agrees.

Captured live against the pinned oracle (Homebrew `yq` v4.53.3), input `a:\n  b: 1\n`:

| flag | `succinctly yq` | `yq` v4.53.3 |
|------|-----------------|--------------|
| `-I0` | `a:\|  b: 1` (2) | `a:\|    b: 1` (**4**) |
| `-I1` | 2 | 2 |
| `-I2` | 2 | 2 |
| `-I3` | 3 | 3 |
| `-I4` | 4 | 4 |

So succinctly treats `-I0` as "fall back to the 2-space default", while real yq
treats it as something that lands on 4. Per ADR-0018 the mode decides and
bug-for-bug fidelity is the default, so unless this qualifies for one of the
rule-4 exemptions (it does not appear to — the output is readable back, nothing
is corrupted, no write is discarded) `succinctly yq` should match.

## Why it went unnoticed

`scripts/ab-cli.py` passed `-I0` on every CLI A/B run, and `tests/data/yq-golden`
has no `-I0` case (`grep -rl 'I0' tests/data/yq-golden/cases` returns nothing).
So the one flag value that diverges is simultaneously the harness default and
absent from the goldens.

## Suggested approach

- Confirm the above against the pinned binary rather than taking this table on
  trust, and check `-I0` combined with `-P`, `-o json`, and flow-style input
  before deciding what the rule actually is (yq's own `-I0` may mean something
  other than "indent 0", given it lands on 4 rather than 0).
- Add a `-I0` golden case either way, so whichever behaviour is chosen is pinned.
- If it is a real divergence and gets fixed, record it; if it is deliberate,
  record it in `docs/compliance/yq/limitations.md`.

newhoggy:
## Summary
- `IndentSpec::for_yaml` clamped `-I0`/`-I1` to width 2, documented as deliberately not modeling real yq's own `-I0`-behaves-like-`-I4` quirk. That quirk turned out not to exist: real yq's YAML width table, pinned from its own go-yaml dependency source, is `0 -> 4`, `1 -> 2`, `2..=9 -> n` unchanged, `>= 10 -> 2` — verified live against yq v4.53.3 and swept across all 150 `tests/data/yq-golden` inputs with zero divergences. Only the `0 -> 4` mapping was ever missing; the "irregular per-level" residual this file's own doc comment described as unmodeled was go-yaml's ordinary indent-rounding rule, which succinctly's emitter already reproduces correctly at every width.
- Two more bugs surfaced and closed alongside the width table:
  - `can_yaml_fast_path`/`can_inplace_yaml_fast_path` OR'd in `output_config.compact`, a JSON concept, so `-I0`'s own (then-"compact") status satisfied the gate regardless of `-P` — `-P -I0` on flow-style input wrongly stayed flow instead of expanding to block style.
  - `YqCommand::indent`'s clap parser capped every value at 7, jq's own `--indent` limit wrongly applied to yq mode. Real yq accepts any non-negative width for JSON output (unbounded above; only YAML is clamped per the table above).
- A negative `-I` value still cannot reach real yq's own panic path here (exit 2, "cannot indent to a negative number of spaces") — clap's own unsigned-integer rejection remains the ADR-0018 rule-4(c) answer, not something this changes.

## Test plan
- [x] Live-verified every repro against pinned Homebrew `yq` v4.53.3
- [x] 3 new golden cases (`indent_eight_2606`, `indent_zero_pretty_expands_flow_2606`, `map_nested_indent_zero_mapping_2606`) captured from the pinned oracle
- [x] Local patch coverage: 100% (16/16 new lines)

Fixes #2606


newhoggy:
Ran a 5-way code review round. No correctness bugs found in the fix itself — three independent finder angles confirmed the width table matches real yq's own go-yaml dependency source and the two related bugs (`-P -I0` fast-path leak, the wrongly-jq-scoped clap range) are correctly closed.

Fixed the review findings:
1. Two doc comments in `yq_runner.rs` (`compute_indent_str` and the M2 fast-path indent setup) still described the pre-#2606 "-I0/-I1 both land on width 2" rule — independently found by two review passes, since the PR carefully updated the adjacent `can_yaml_fast_path` comments but missed these two.
2. A pre-existing `#1982` paragraph in `docs/compliance/yq/limitations.md` asserted "`-P -I0` streams" as the reason certain DOM-only divergences don't apply there — backwards after this PR's own fix, which makes `-P` force the DOM route unconditionally now regardless of `-I`.
3. Added a regression test for a large `-I` value (200) against JSON output — the actually-unbounded path this issue's clap-range removal introduces, which the original test suite didn't exercise past `-I10`.
4. Shortened the `yq-language.md` indent-flag table cell to fit the table's existing column width (flagged by the `markdown-tables` convention).

One suggestion (a redundant `1 => 2` match arm in `IndentSpec::for_yaml`, technically subsumed by the `_ => 2` catch-all) was considered and kept as-is — the explicit arm mirrors the documented 4-way rule table 1:1, which I judged clearer than the marginal DRY savings from merging it.

Full test suite + clippy re-verified clean; patch coverage 100% (16/16 new lines).

Issue #2777, opened 2026-09-11
yq: -p json flow-mapping grammar diverges from real yq in both directions (one silent wrong value)


newhoggy:
**Severity: Medium** — one row is a *silent wrong value* (exit 0, different data), the rest are wrong accept/reject.

## What's wrong

Found while fixing #2279 (which closes the flow-**sequence** comma family only). `succinctly yq --input-format json` diverges from real yq across the flow-**mapping** grammar, in both directions. Captured live against Homebrew `yq` v4.53.3, comparing `(exit code, stdout)`:

| input | real yq | succinctly yq | |
|---|---|---|---|
| `{"a":1 "b":2}` | `{"a":1,"b":2}` | `{"a":"1 \"b\":2"}` | **silent wrong value** |
| `{"a":1"b":2}` | `{"a":1,"b":2}` | `{"a":"1\"b\":2"}` | **silent wrong value** |
| `{,}` | `{}` | `{"":null}` | wrong value |
| `{,"a":1}` | `{"a":1}` | `{"":null,"a":1}` | wrong value |
| `{"a":1,,"b":2}` | `{"a":1,"b":2}` | `{"a":1,"":null,"b":2}` | wrong value |
| `{"a" 1}` | `{"a":1}` | `<ERROR>` | we **over**-reject |
| `{"a":}` | `<ERROR>` | `{"a":null}` | we under-reject |
| `{:1}` | `<ERROR>` | `{"":1}` | we under-reject |

## The model real yq follows

Consistent across every row above: **yq ignores punctuation inside `{}` entirely and pairs up tokens** — key, value, key, value. Stray commas vanish (`{,}` → `{}`), missing commas and colons are inferred (`{"a" 1}` → `{"a":1}`), and the *only* object error is an odd token count (`{"a":}`, `{:1}`).

succinctly instead reads these as YAML flow mappings, which is why `{"a":1 "b":2}` comes back as a single plain scalar `1 "b":2` — the correct *YAML* reading, and the wrong *yq* one. Root cause is the same as #2279's: #1398 routes `-p json` through `YamlIndex`.

## Why it wasn't fixed with #2279

#2279's fix is deliberately flow-sequences-only, because tightening mappings toward JSON would make us reject `{"a":1,}`, `{,}` and `{"a" 1}` — all of which real yq **accepts**. Matching yq here is not "validate more", it is "reproduce a token-pairing parser", which is a different and larger change.

## Open design question

[...]

PR #2832, opened 2026-09-12
fix(yq): validate JSON scalar grammar while parsing JSON-sourced input

### LINKED ISSUE #2279: jq/yq: plain stdout M2 fast path shares #2262's comma-validation gap, with no already-validating fallback
**Severity: Low** (non-destructive — wrong stdout output/exit code, not data loss like #2262's `--inplace` finding, which is already fixed)

## What's wrong

`succinctly yq`'s plain stdout fast path (`can_json_fast_path`/`can_yaml_fast_path`, i.e.
ordinary `succinctly yq --input-format json '<filter>'` with no `--slurp`/`--inplace`) has
the identical malformed-stray-comma validation gap that #2262/#2276 just fixed for
`--slurp`/`--inplace`'s own M2 fast paths — but was deliberately left unfixed there, because
its fallback (`evaluate_yaml_direct_filtered`, #1398's cursor-native evaluator, used for
*every* ordinary `--input-format json` filter regardless of M2 eligibility) shares the
identical `YamlIndex`/`mark_json_sourced` flow-grammar validation hole. Declining the fast
path here would cost M2's performance with zero correctness gain.

## Repro (confirmed live against Homebrew yq v4.53.3)

```console
$ echo '[1,]' | succinctly yq --input-format json -o json '.[0]'
1                                    # exit 0 -- wrong
$ echo '[1,]' | yq -p json '.[0]'
Error: bad file '-': json: null unexpected end of JSON input   # exit 1
```

## Why it's separate from #2262/#2276

#2276 (the code-review follow-up to #2262) fixed `--slurp`'s and `--inplace`'s own M2 fast
paths by declining them for JSON-sourced input and falling back to the DOM path (`parse_input`),
which #2262 already fixed. This plain stdout path's *fallback itself* is a different function
(`evaluate_yaml_direct_filtered`) that never received any of #2211/#2243/#2262's checks and
shares the same `YamlIndex` flow-grammar leniency the fast path itself has — so simply
declining the fast path here (the fix pattern #2276 used) would not help; there is no
"fall back to something that already validates" option for this specific route without a
real fix to `evaluate_yaml_direct_filtered` (or its own equivalent bridge) itself.

## Suggested fix direction

Either give `evaluate_yaml_direct_filtered`'s own JSON-sourced-input handling the same
`container_gap_ok`/`trailing_element_gap_ok`/`preceding_delimiter_ok`-family checks (via
`YamlCursor`'s own `DocumentCursor` trait implementation, gated on `canonicalize_numbers()`/
JSON-sourced detection, which is already exposed elsewhere in this codebase), or find some
other route back to a function that already validates. Needs its own investigation — not a
small follow-up.

## Found via

Code review on #2276 (itself a code-review follow-up to #2262), while confirming that
declining the `--slurp`/`--inplace` fast paths correctly routed to an already-validating
fallback — this plain stdout path's own fallback does not.

## Related

#2211, #2243, #2262, #2276

### LINKED ISSUE #2778: yq: -p json accepts YAML-only scalars and missing separators inside flow sequences
**Severity: Low** — wrong accept (exit 0 where real yq exits 1), non-destructive.

## What's wrong

Found while fixing #2279, which closes the flow-sequence **comma** family (`[1,]`, `[,1]`, `[1,,2]`, `[,]`). The same sequences diverge on everything that is *not* a comma: `succinctly yq --input-format json` accepts a range of array contents real yq rejects, because #1398 routes `-p json` through `YamlIndex` and these are all valid **YAML** scalars.

Captured live against Homebrew `yq` v4.53.3:

| input | real yq | succinctly yq |
|---|---|---|
| `[a]` | `<ERROR>` | `["a"]` |
| `['a']` | `<ERROR>` | `["a"]` |
| `[True]` | `<ERROR>` | `[true]` |
| `[a: 1]` | `<ERROR>` | `[{"a":1}]` |
| `[1 2]` | `<ERROR>` | `["1 2"]` |
| `[1"a"]` | `<ERROR>` | `["1\"a\""]` |
| `["a":1]` | `<ERROR>` | `[{"a":1}]` |
| `[.5]` | `<ERROR>` | `[0.5]` |
| `[+1]` | `<ERROR>` | `[1]` |
| `[1e]` | `<ERROR>` | `["1e"]` |

## The boundary, which is not simply "strict JSON"

yq's array path rejects every row above but **accepts** these, which RFC 8259 forbids:

| input | real yq |
|---|---|
| `[01]` | `[1]` |
| `[00]` | `[0]` |
| `[1.]` | `[1]` |

So this is not "run a JSON validator" — leading zeros and a bare trailing `.` survive, while `.5`, `+1` and `1e` do not. That looks like YAML-style scalar resolution applied to tokens a JSON-shaped tokenizer produced, and any fix needs to reproduce that boundary rather than assume strictness.

This is precisely why #2279 was scoped to delimiters only: the delimiter checks never look at scalar text, so they fixed the comma rows without disturbing `[01]`/`[00]`/`[1.]`. Closing this issue means taking on the scalar grammar, which is the harder half.

## Repro

```console
$ echo '[a]' | yq -p json -o json -I0 '.'
Error: bad file '-': json: null unexpected end of JSON input
$ echo '[a]' | succinctly yq --input-format json -o json -I0 '.'
["a"]
```

Related: #2279 (delimiters, fixed), #2777 (the mapping side), #1398.

newhoggy:
## Summary

- Closes #2279's remaining gap: `-p json`/`--input-format json` accepted every YAML-only scalar spelling real yq rejects (`True`, `'a'`, `.5`, `+1`, `~`, a bare `- 1`/`a: 1` at the top level, an implicit `key: value` pair inside a JSON array), because #2279's delimiter checks never looked at scalar text.
- The boundary is not "strict JSON" — it's `goccy/go-json` v0.10.6's own two-layer grammar (token scanner + `strconv.ParseFloat`), re-derived from the reference's source and pinned against ~60 live probes against Homebrew `yq` v4.53.3. Leading zeros and a bare trailing `.` still survive (`[01]` → `[1]`, `[1.]` → `[1]`), matching real yq's own leniency, while `.5`/`+1`/`1e`/`~` do not.
- Applied at every scalar *value* position the JSON-sourced parser reaches (flow-sequence item, flow-mapping value, bare block/top-level scalar) under the existing `json_strict` flag — genuine YAML input is completely unaffected.
- Deliberately out of scope, and documented as such: mapping *keys* (real yq's own JSON decoder panics on a non-string key, `{1:1}`/`{true:1}`, which this does not reproduce per ADR-0018 rule 4(c) — the mapping-key delimiter/quoting side is #2777's territory) and anchor/tag-prefixed content (no JSON spelling exists for it at all).

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, green)
- [x] New unit test (`test_json_strict_plain_scalar_ok_matrix`, `test_json_strict_plain_scalar_ok_rejects_overflow`) pins the predicate's accept/reject boundary
- [x] New CLI tests (`json_sourced_scalar_grammar_matches_yq_2778`, `genuine_yaml_scalars_are_unaffected_by_json_strict_2778`) pin the full triage-plan matrix against the live oracle and confirm genuine YAML is unaffected
- [x] Manually diffed output against the pinned `/opt/homebrew/bin/yq` v4.53.3 oracle across the full accept/reject matrix (50+ rows) plus the bare top-level and mapping-value cases — all match exactly
- [x] Must-not-change sweep: genuine YAML (no `-p json`), `succinctly jq`, and `succinctly json validate` all confirmed unaffected

Closes #2778

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -990,10 +990,30 @@ token through to `ParseFloat`, not because `ParseFloat` itself would reject them
 Applied at every scalar *value* position the JSON-sourced parser reaches — flow-sequence
 item, flow-mapping value, and bare block/top-level scalar — under the same `json_strict`
 flag #2279 introduced, so `a`, `True`, `.5`, `1e999` are rejected identically whether nested
-in `[...]`, in `{"k": ...}`, or standing alone as the whole document. A bare top-level
-`key: value` (no enclosing `{}`) and a bare `- item` (no enclosing `[]`) are rejected
-structurally at the same point — the reference's token scanner never reaches YAML's block
-grammar at all, since it has no spelling for it.
+in `[...]`, in `{"k": ...}`, or standing alone as the whole document.
+
+Alongside scalar-grammar validation, four YAML-only *structural* tokens are rejected the
+same way, because none of them have a JSON spelling at all regardless of what text follows
+them — real yq's token scanner refuses the byte itself, never reaching a "value" to
+validate:
+
+- A bare top-level `key: value` (no enclosing `{}`) or `- item` (no enclosing `[]`) —
+  rejected at `parse_block_node`'s single chokepoint for every illegal value-start byte,
+  which also covers `?`/`&`/`!`/`*`/`'` uniformly rather than each needing its own arm.
+- `?` (the explicit-key indicator), in *both* the block-context and flow-mapping (`{? "a":
+  1}`) spellings — a structural rejection of the marker itself, not a key-grammar check, so
+  it stays in scope even though mapping-key *text* grammar (#2777) does not.
+- `&anchor`/`!!tag` prefixing any value — real yq errors on the prefix byte before it would
+  ever reach a scalar, an anchor, or a nested container behind it, so `&x True`, `&x 'y'`,
+  and even this fix's own new `&x - 1` structural rejection are all refused uniformly.
+- `--- |`/`--- >` (a block-scalar indicator right after a document marker) — this shape
+  bypasses the ordinary block dispatcher via its own dedicated fast path
+  (`parse_inline_document_value`), so it needed its own separate gate alongside the
+  chokepoint above.
+
+An implicit single-pair mapping inside a JSON *array* (`[a: 1]`, `["a":1]`) is rejected the
+same way, in `parse_flow_sequence_inner` — a JSON array element is a value, never a `key:
+value` pair, so the whole construct errors before the key's own text is even examined.
 
 **Mapping *keys* are exempt — deliberately, not an oversight.** Real yq's own JSON decoder
 *panics* (does not cleanly error) on a non-string key — `{1:1}`, `{true:1}` both crash with
@@ -1002,13 +1022,8 @@ ADR-0018 rule 4(c), succinctly does not reproduce a reference panic; `{"1":1}` (
 existing behavior for a numeric-looking key) is unchanged. Key *delimiter* and *quoting*
 rules (e.g. real yq also rejects an unquoted or single-quoted key, `{a:1}`/`{'a':1}`, which
 succinctly still accepts) are the mapping-key half of #2777, not this issue — #2778 only
-ever touches what a scalar *value* token may spell.
-
-**Anchor/tag-prefixed content is untouched.** `&anchor`/`!!tag` have no JSON spelling at
-all, so real yq's front end would reject them before ever reaching a scalar to validate —
-but this is unobserved territory (not in the pinned matrix) and this fix does not attempt
-it; a document reaching a scalar through an anchor/tag prefix parses exactly as it did
-before #2778.
+ever touches what a scalar *value* token may spell, plus the four structural, key-agnostic
+rejections above.
 
 **A pre-existing divergence this widens by one case, not a new one.** Real yq's own
 `-p json`/`eval-all -p json --input-format json` path (confirmed live against v4.53.3) is far
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -1025,6 +1025,25 @@ succinctly still accepts) are the mapping-key half of #2777, not this issue — 
 ever touches what a scalar *value* token may spell, plus the four structural, key-agnostic
 rejections above.
 
+**Cost: `-p json`/`--input-format json` parsing is measurably slower, isolated entirely to
+that path.** Every plain (unquoted) scalar under `json_strict` now costs a byte-charset scan
+plus an `f64` parse it did not pay before -- CI's `perf-guard.py` (issue #1523) measured
+`users_yq_keys_unsorted` (a `keys_unsorted` query over a 2MB `-p json`-detected fixture,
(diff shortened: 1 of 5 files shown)


newhoggy:
Ran a code review after the initial push — found and fixed 4 real dispatch gaps where `json_strict` was bypassed entirely (not just a scalar-grammar miss): block/flow-mapping `?` explicit-key indicator, `&anchor`/`!tag`-prefixed content (which bypassed the new structural bare-`- item`/`key: value` rejections too), and `--- |`/`--- >`'s own dedicated dispatch path. Each confirmed live against `yq` v4.53.3 before and after the fix, consolidated the block-context checks into one chokepoint, reused an existing finite-float-parse primitive instead of a second copy, and added regression tests + updated `docs/compliance/yq/limitations.md` for all four. Rebased on latest `main`, full local suite + both clippy configs green.

For the deepest possible pass, `/code-review ultra` is available if you'd like one before merge — not required, just an option.

newhoggy:
**Perf Regression Guard note (x86_64/ARM64-Linux fail):** investigated, real, and isolated. `users_yq_keys_unsorted` (a `keys_unsorted` query over a 2MB `-p json`-detected fixture, mostly numeric fields) is +13.5%/+12.5% instructions — every other query in the suite, including `jq`-mode queries over the *identical* fixture (which never touch `YamlIndex` at all), measured within ±0.1%. The cost is 100% attributable to the new `json_strict` scalar-grammar validation this PR adds (a byte-charset scan + `f64` parse per plain scalar), and only fires for `-p json`/`--input-format json`-detected input — genuine YAML parsing is unaffected.

Added `#[inline]` on the hot-path validation functions (the one free win — removing the redundant `str::from_utf8` re-validation after the charset check would help further but needs `unsafe`, which this crate forbids crate-wide via `-D unsafe-code`). The remaining cost is inherent to actually doing the validation real yq's own JSON front end does. Per ADR-0018, fidelity to the reference is the default and a performance cost isn't among the rule's carve-outs for declining to match it — and `Perf Regression Guard` is deliberately outside the merge queue's required-checks subset for exactly this kind of narrowly-scoped, understood tradeoff. Full writeup in `docs/compliance/yq/limitations.md`.

Proceeding toward merge on that basis; flagging here in case a maintainer wants to weigh in before it lands.

newhoggy:
While chasing patch coverage on the structural-rejection checks, found a real bug in my own review-round fix: I had added a blanket "reject any bare \`key: value\`" rule under \`json_strict\`, but it turns out real yq's \`-p json\` (no \`--slurp\`) actually reads a *stream* of concatenated top-level JSON values (jq-style) — \`true: 1\` and \`"a": 1\` are both valid two-value streams in real yq (confirmed live, exit 0), not invalid mappings. My blanket rule over-rejected them.

Reverted that check and filed #2839 to track the real gap (succinctly has no multi-value streaming support at all — a separate, larger architectural item). Restored grammar validation on the two functions (\`parse_mapping_entry\`/\`parse_inline_value\`) that the bad check had made unreachable, since a streamed value's own grammar is still checked in real yq (\`true: 'v'\`/\`true: True\` still error). Added tests pinning both sides. Full local suite + clippy green again; pushing now.

Issue #2667, opened 2026-09-09
jq: a postfix applied directly to an array/object literal ([1,2][], {a:1}.a) is a parse error everywhere


newhoggy:
**Severity: Medium** (valid jq programs rejected at parse time; no wrong output)

Found while probing #2180. A postfix applied directly to an array or object *literal* is rejected by the parser everywhere, top level included. Real jq 1.7.1 accepts every one of these:

| filter | jq 1.7.1 | `succinctly jq -cn` |
|---|---|---|
| `[1,2][]` | `1` `2` | `parse error at position 5: unexpected character '['` |
| `{a:1}.a` | `1` | `parse error at position 5: unexpected character '.'` |
| `[[1,2][]]` | `[1,2]` | `parse error at position 6: expected ']', found '['` |
| `first([1,2][])` | `1` | `parse error at position 11: expected ';' or ')' in function arguments` |
| `[limit(1; [1,2][])]` | `[1]` | same |
| `def f(g): g; f([1,2][])` | `1` `2` | same |
| `[first({a:(1 as $x ?// $y \| 1)}.a)]` | `[1,1]` | same |

Piping instead works (`[first([1,2] \| .[])]` is `[1]` in both), so the parser handles the postfix suffix chain only after a term that is not a collection literal. Compare `src/jq/parser.rs`'s postfix loop against the terms it is applied to — array/object construction should be a primary term like `(...)`, `$var` and a function call, all of which do accept `[]`/`.a`/`[0]` suffixes today.

Unrelated to laziness or `?//`; listed here only because the #2180 probes are how it surfaced.

newhoggy:
🔧 Claimed by agent `Cadence` — branch `issue-2667-postfix-on-literal`

newhoggy:
Fix in #2736. Two follow-ups filed while working it:

- **#2741** — the same `Term`-rule gap on **string and number** literals (`"abc"[0:1]`
  is `"a"` in jq, still a parse error here). Same one-line shape of fix; left out of
  #2736 because this issue's title, table and repro are all specifically about
  collection literals. It carries two checks #2667 did not need: string interpolation,
  and whether `"abc".x` should parse at all.
- **#2742** — `$__loc__.file` is `"<stdin>"` where jq says `"<top-level>"`. Unrelated
  (it is a variable, and already took postfix before this fix); reproduces on `main`.

Two notes on this issue's own text, both confirmed against a build:

[...]

PR #2833, opened 2026-09-12
fix(jq): let a postfix chain apply to a string or number literal

### LINKED ISSUE #2741: jq: a postfix chain still cannot apply to a string or number literal (`"abc"[0:1]`)
**Severity: Low** (valid jq programs rejected at parse time; no wrong output)

## Summary

#2667 made an array/object literal a postfix target. jq's `Term` rule allows the same
suffix chain on **string and number** literals, and those are still parse errors here.

## Repro

Live against the pinned oracle (`/usr/bin/jq` 1.7.1) and a `--features cli` build of
the #2667 branch:

|---|---|---|
| `"abc"[0:1]` | `"a"` | `parse error` |
| `"abc"[1:]` | `"bc"` | `parse error` |
| `"abc"[0]` | `jq: error ... Cannot index string with number` | `parse error` |
| `1[0]` | `jq: error ... Cannot index number with number` | `parse error` |

The slice forms are the ones that matter: they are useful programs jq accepts and
succinctly refuses. The plain-index rows only differ in *when* the failure happens
(runtime vs parse), but they still show the grammar gap.

Controls, working in both before and after #2667:

```console
$ jq -cn '"abc" | .[0:1]'   # "a"
$ jq -cn '("abc")[0:1]'     # "a"
```

## Root cause

Same shape as #2667: `parse_primary_inner` (`src/jq/parser.rs`) routes only some
primary terms through `parse_postfix`. #2667 added the `'['` and `'{'` arms; the
string arm (`Some('"') => self.parse_string_or_interpolation()`) and the number arm
still return directly.

## Why this is separate from #2667

Found by review of #2667 and confirmed pre-existing (it reproduces on `main` before
that fix as well). #2667's issue text, repro table and title are all specifically
about *collection* literals, so this was left out rather than folded in as scope
creep — but it is the same `Term` rule and the same one-line shape of fix, so closing
it finishes the job.

## Suggested fix direction

Wrap the string and number arms in `parse_postfix` the way #2667 wrapped the
collection arms. Two things to check that #2667 did not have to:

1. **String interpolation.** `parse_string_or_interpolation` handles `"\(...)"`;
   confirm a postfix after an interpolated string behaves (`"\(1)"[0:1]`) and that
   the suffix is not mistaken for part of the interpolation.
2. **`.foo` after a string.** jq's `Term` allows `'.' String` as a *field* spelling
   (`.["a"]` / `."a"`), so check `"abc".x` against the oracle before assuming the
   postfix loop should accept it — the leading-dot form is a different production
   from a suffix on a string term.

Also worth adding a row per literal kind to
`test_postfix_applies_directly_to_a_collection_literal_2667`'s neighbours, so the
`Term` rule is pinned once, completely, rather than one literal kind at a time.

newhoggy:
## Summary
- #2667 made an array/object literal a postfix target; jq's `Term` rule allows the same suffix chain on string and number literals, and those were still parse errors: `"abc"[0:1]`, `"abc"[0]` and `1[0]` are all valid jq 1.7.1 programs that succinctly rejected before evaluation.
- `parse_primary_inner`'s string arm now routes through `parse_postfix` the same way #2667's array/object arms do (plain and interpolated strings alike), and the digit-literal arm does the same.
- Not extended to the unary-minus arm: real jq's `-1[0]` parses as `-(1[0])` (confirmed live, errors naming the *positive* operand), which would need that arm to stop eagerly folding `-`+digits into one negative-literal token whenever a postfix could follow — a materially larger restructure of the #1035 negative-literal-preservation logic there, out of this issue's scope.
- **jq mode only**: unlike #2667's own collection-literal fix (which real yq happens to accept for the same shapes), real yq rejects a postfix chain on a string or number literal at parse time (confirmed live against v4.53.3), so widening this arm unconditionally would introduce a new divergence rather than close one.

## Test plan
- [x] Live-verified every repro against `/usr/bin/jq` 1.7.1 and yq v4.53.3
- [x] Local patch coverage: 100% (6/6 new lines)

Fixes #2741

### DIFF
--- a/src/jq/parser.rs
+++ b/src/jq/parser.rs
@@ -1666,37 +1666,36 @@ impl<'a> Parser<'a> {
                 self.parse_postfix(object)
             }
 
-            // String literal or interpolation. #2741, jq mode only: same
-            // `Term` rule #2667 fixed for collection literals --
-            // `"abc"[0:1]`/`"abc"[0]`/`"abc"[]` are all valid jq 1.7.1 and
-            // were parse errors here, since only `(...)`/`[...]`/`{...}`
-            // routed through `parse_postfix`. This covers plain and
-            // interpolated strings alike (`parse_string_or_interpolation`
-            // returns one `Expr` either way); confirmed live that jq
-            // applies the same postfix chain to an interpolated string
-            // (`"\(1)"[0:1]` is `"1"`). yq mode is excluded: real yq
-            // rejects every one of these at parse time too (`bad
-            // expression, please check expression syntax`, confirmed live
-            // against v4.53.3 for both `"abc"[0:1]` and `"abc".x`) --
-            // unlike #2667's own array-literal case, where real yq
-            // happens to accept the same postfix shape (`[1,2][0]`
-            // succeeds there), there is no reachable yq behaviour here to
-            // match, so widening this arm unconditionally would introduce
-            // a brand new divergence rather than close one.
-            Some('"') if self.mode == ParserMode::Jq => {
+            // String literal or interpolation. #2741: same `Term` rule
+            // #2667 fixed for collection literals -- `"abc"[0:1]`/
+            // `"abc"[0]`/`"abc"[]` are all valid jq 1.7.1 and were parse
+            // errors here, since only `(...)`/`[...]`/`{...}` routed
+            // through `parse_postfix`. This covers plain and interpolated
+            // strings alike (`parse_string_or_interpolation` returns one
+            // `Expr` either way); confirmed live that jq applies the same
+            // postfix chain to an interpolated string (`"\(1)"[0:1]` is
+            // `"1"`). See `postfix_if_jq`'s own doc comment for the
+            // jq-mode gate.
+            Some('"') => {
                 let string = self.parse_string_or_interpolation()?;
-                self.parse_postfix(string)
+                self.postfix_if_jq(string)
             }
-            Some('"') => self.parse_string_or_interpolation(),
 
-            // Format strings: @text, @json, @uri, etc.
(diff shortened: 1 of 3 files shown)


newhoggy:
Ran a 4-way code review round. No correctness bugs found in the fix itself — extensive live oracle probing (path-tracking, `?`-suppression, destructuring interaction, whitespace handling, interpolation) all matched jq/yq exactly.

Applied the genuinely worthwhile findings:
1. Factored the duplicated jq-mode-gate/non-gate arm pairs into one `postfix_if_jq` helper.
2. Extended the same fix to the `@format` arm — `@base64[0]` is the identical `Term` rule, was still a parse error, and real yq rejects it the same way (confirmed live). Found independently by one of the reviewers.
3. Strengthened the yq-mode-only test to pin the specific error text per filter rather than just a nonzero exit code, so a future regression that let these start *parsing* in yq mode (even if they then failed at evaluation, mirroring jq mode's own runtime errors) would still be caught.

Filed #2834 separately for a real finding that's out of scope here: `key`/`path`/`parent` give bogus position output (not a crash — wrong output) for a navigational step on a detached literal. Confirmed this is pre-existing (reproduces via the explicit `.a | ("x" | .[0:1]) | key` pipe spelling that predates this PR entirely) — this PR just makes it much easier to trigger via the new, more natural `"x"[0:1] | key` spelling. It's a cross-cutting bug in `eval_generic.rs`'s identity-tracking (`OwnedIdentity::child`), a different subsystem than this PR's narrow parser fix.

Full test suite + clippy re-verified clean; patch coverage 100% (10/10 new lines).

Issue #1582, opened 2026-08-23
ci(bench): perf-guard's x86 leg false-positives on codegen shift — +5.2% on a query the change never executes


newhoggy:
**Severity: Medium** (a false positive that will recur on any PR touching a large hot file, and
the sanctioned remedy — `--update-baseline` — silently degrades the guard)

## Summary

`scripts/perf-guard.py`'s x86_64 leg reported `wide_keys_unsorted` at **+5.2%** on
[#1578](https://github.com/rust-works/succinctly/pull/1578) (`573,924,928` -> `603,974,111`).
The change that triggered it provably does not execute on that query. The ARM64-Linux leg of
the same run is **+0.0% on all six queries**.

The guard's premise — "a real 2-3x work increase shows up as a 2-3x instruction-count
increase" — holds. What it does not account for is the converse: an instruction count can move
without any change in the work performed, because the backend compiles unchanged code
differently once a large translation unit changes shape. `src/jq/eval.rs` is ~53k lines, and
#1578 added ~250 to it.

## Evidence that the work is unchanged

1. **The changed code never runs on this query.** `succinctly jq keys_unsorted` on the guard's
   own 2mb `wide` fixture, under `cargo llvm-cov run`: of the 67 *executable* lines the PR
   added to `src/jq/eval.rs`, **0 were executed**. (Instrumentation was live — 33 `eval.rs`
   lines executed overall.) `keys_unsorted` is a value-position filter; it never reaches
   `resolve_node`/`PathBranch`, which is all the PR touches.
2. **ARM64-Linux is flat.** All six queries `+0.0%` (drift of a few thousand instructions out
   of hundreds of millions). Same source, same logic, same fixtures.
3. **x86 drifted on unrelated queries too**, all sub-threshold so invisible in the pass/fail:
   `wide_identity` +0.8%, `users_identity` +1.0%, `arrays_identity` +0.5%. A targeted
   regression does not move four unrelated queries by correlated amounts; a global codegen
   shift does.
4. **No layout change to blame.** `size_of::<PathBranch>()` is **104 bytes before and after** —
   the added `bool` lands in existing padding. So this is inlining/vectorization budget, not
   struct growth, and no repacking of the type can avoid it.

## Why this matters

The script tells a failing PR to "re-run with `--update-baseline` and say why in the commit
message" if the drift is "real and understood". Neither branch of that instruction fits here:

- Updating the baseline records a number no source change explains, and ratchets the guard by
[...]

PR #2836, opened 2026-09-12
feat(bench): add a wide-escaped-keys JSON generator pattern

### LINKED ISSUE #2637: bench: no generator emits object keys containing escapes
**Severity: Low**

## Summary

While fixing #965 item 10 (PR #2636, `DocumentValue::decoded_key_str`), item 10's strongest
claim — that decoding an escaped object *key* twice costs a real, avoidable second allocation
via `decode_escapes` — was left unmeasured, because nothing in `src/bin/succinctly/generators.rs`
can produce an object with an escaped key. Per
[docs/guides/benchmarking.md § A/B Benchmarking Method, rule 7](docs/guides/benchmarking.md#7-a-benchmark-cannot-measure-a-shape-it-does-not-generate)
("a benchmark cannot measure a shape it does not generate"), that means half of #965's item 10
is currently unfalsifiable either way — not "shown to be negligible," just invisible to the
suite.

## What's missing, concretely

- `generate_wide_json` (`src/bin/succinctly/generators.rs:483`) — the pattern used by
  `wide_keys_unsorted`/`wide_identity` in `scripts/perf-guard.py`'s `QUERIES` matrix, i.e. the
  one shape that actually exercises the field-iteration/key-decode path — emits plain
  `"k{i}"` keys via `format!(r#""k{i}":{val}"#)`. It doesn't even take an `escape_density`
  parameter.
- `add_string_variations` (`generators.rs:132`) mixes escaped and plain strings based on
  `escape_density`, but only for array *elements* (string values), never object keys.
- `add_realistic_records` (`generators.rs:357`) applies `escape_density` only to the `name`
  *field's value* (`format!(r#"User \"{count}\" Smith"#)`); the object's own keys
  (`id`, `name`, `email`, `age`, ...) are always the same fixed, unescaped literals.

So across every generator pattern, an object key containing a JSON escape sequence (`\"`, `\n`,
`\t`, `\uXXXX`, etc.) never occurs in generated test data at all.

## Consequence

Any per-key decode/allocation cost that's specific to *escaped* keys — as opposed to plain
ASCII keys, which already take a fast unescaped path — is unmeasurable today. This blocked half
of #965 item 10's claim: PR #2636 fixed the doubled-`as_str()`-call shape (one JSON re-scan +
one YAML `contains()` triple-pass, always paid), but could not confirm or refute the additional
"escaped keys reallocate via `decode_escapes` twice" sub-claim with real numbers, since no
generator can produce that input shape.

## Suggested fix direction

- Either extend `generate_wide_json` to accept (and apply) an `escape_density`-style parameter
  to a fraction of its keys, or add a new pattern (e.g. `wide_escaped_keys`) that's `wide`-shaped
  but with escape sequences injected into a controlled fraction of the `k{i}` keys.
- Extend `scripts/perf-guard.py`'s `QUERIES` matrix with a query against that new shape (e.g.
  `wide_escaped_keys_unsorted` using `keys_unsorted`, mirroring the existing
  `wide_keys_unsorted` entry) so the escaped-key decode path has ongoing regression coverage,
  not just a one-off manual benchmark.

## References

- #965 (item 10)
- PR #2636 (fixes the always-paid half of item 10; this issue covers the unmeasured half)

newhoggy:
## Summary

- No generator could produce an object with an escaped *key*: `generate_wide_json` (the pattern exercising `keys_unsorted`'s field-iteration path) always writes plain `"k{i}"` keys and doesn't take an `escape_density` parameter; `add_string_variations`/`add_realistic_records` vary `escape_density` only for string *values*.
- This left half of #965 item 10's claim — that decoding an escaped key specifically costs a real, avoidable second allocation via `decode_escapes` — unmeasurable (no shape to run it against), per `docs/guides/benchmarking.md`'s rule 7 ("a benchmark cannot measure a shape it does not generate").
- Adds `Pattern::WideEscapedKeys` / `generate_wide_escaped_keys_json`, mirroring `generate_wide_json` but drawing a controllable fraction of keys from six escape-class patterns (`\"`, `\\`, `\n`, `\t`, `\/`, `é`). Every key keeps its index regardless of escape status, so keys stay distinct at any density.
- Wired through `PatternArg::WideEscapedKeys` (renders as `wide-escaped-keys` on the CLI).

**Not in this PR**: wiring a `wide_escaped_keys_unsorted` row into `scripts/perf-guard.py`'s `QUERIES` matrix, which the issue's suggested fix direction also proposes. CI's perf-guard check compares the PR-head binary against a freshly-built merge-base binary (`--baseline-binary`, #1582) for both `pull_request` and `push` events — a `QUERIES` row naming a pattern that doesn't exist yet on the merge-base would break `generate_fixture` against that binary the moment this PR lands, since the merge-base predates the new `PatternArg` variant. That wiring needs a follow-up PR once this generator support is on `main`. Filed as a follow-up issue.

Fixes #2637.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, all green)
- [x] Unit tests: valid JSON at every density (0.0–1.0), `succinctly jq`'s `keys_unsorted` count matches real `jq`'s, `--escape-density` actually controls the escaped fraction (checked against raw JSON key literals, not a decoded `Value`), keys stay distinct (no collisions)
- [x] CLI golden test (`cli_golden_tests.rs`) + updated `--help` snapshot
- [x] Manually verified: `succinctly json generate ... -p wide-escaped-keys --escape-density 0.5 | succinctly jq keys_unsorted` decodes correctly and agrees with real `jq`

### DIFF
--- a/src/bin/succinctly/generators.rs
+++ b/src/bin/succinctly/generators.rs
@@ -17,6 +17,7 @@ pub enum Pattern {
     Pathological,
     Pretty,
     Wide,
+    WideEscapedKeys,
 }
 
 /// Generate JSON of approximately target_size bytes
@@ -42,6 +43,9 @@ pub fn generate_json(
         Pattern::Pathological => generate_pathological_json(target_size, seed),
         Pattern::Pretty => generate_pretty_json(target_size, seed),
         Pattern::Wide => generate_wide_json(target_size, seed),
+        Pattern::WideEscapedKeys => {
+            generate_wide_escaped_keys_json(target_size, seed, escape_density)
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


newhoggy:
Follow-up filed: #2837 (wire `wide-escaped-keys` into `scripts/perf-guard.py`'s `QUERIES` matrix once this lands).

newhoggy:
Patch coverage: 97.48% (116/119 new lines). The 3 uncovered lines (`generators.rs:959,974,986`) are all inside `panic!`/`assert!` failure-message format arguments (mirroring the sibling `wide` pattern's own tests) — unreachable in a passing test suite by design, the same class already tolerated throughout this codebase's coverage exemptions.

Issue #2840, opened 2026-09-12
jq: label's own error re-raise is also gen_call(error) in real jq, so a shadowing def error: also intercepts it — not just break


newhoggy:
**Severity: Low**

## Summary

#2687 fixed `break $x`'s desugaring to a resolvable `error/0` call, so a shadowing `def error:` (arity 0) in scope at the break's own position intercepts it, matching real jq. Found during that investigation: real jq's `label`'s own **error re-raise** — what happens when a `label $x | ...` block's body raises an escape that is *not* a matching `Control::Break($x)` (an uncaught `error(...)`, a division-by-zero, any other propagating error) — is *also* `gen_call("error", gen_noop())` in jq's `parser.y`, resolved at the **label's own lexical position**. A `def error:` in scope there therefore also intercepts an error escaping through the label, exactly as it now does for `break`.

## Repro

Captured live against `jq-1.7.1-apple`:

```console
$ jq            -nc 'def error: "S"; label $out | (1, error("x"), 2)'
1
"S"
$ succinctly jq -nc 'def error: "S"; label $out | (1, error("x"), 2)'
1
jq: error (at <unknown>): x
```

```console
$ jq            -nc 'def error: "S"; label $out | (1, (1/0), 2)'
1
"S"
$ succinctly jq -nc 'def error: "S"; label $out | (1, (1/0), 2)'
1
jq: error (at <unknown>): number (1) and number (0) cannot be divided because the divisor is zero
```

Position-dependent, same as `break`'s own shadowing:

```console
$ jq -nc 'label $out | (def error: "S"; error("x"))'
jq: error (at <unknown>): x
```
(not shadowed — the `def` isn't in scope at the *label's* position, only inside its own body)

## Root cause

`src/jq/eval.rs`'s `eval_label`/`each_label`, `src/jq/eval_generic.rs`'s `each_label_generic`, and the owned-identity/path-context `Label` arms (`eval_generic.rs`) all catch a matching `Control::Break(name)` and let every other escape (including `Control::Error`) propagate unconditionally. None of these route a non-matching escape through a resolvable call the way #2687 now routes `break` itself.

## Why this is separate from #2687

[...]

PR #2842, opened 2026-09-12
fix(jq): let a def error: shadow break $label, matching jq's own desugaring

### LINKED ISSUE #2687: jq: a `def error:` does not intercept `break $label`, though jq desugars `break` into a call to `error/0`
## What's wrong

Found while reviewing #2395's branch, probing #2036's shadowing surface.

In real jq, `break $label` is **not** a primitive — the parser desugars it into a call to
`error/0`, resolved through ordinary lexical scope. So a user `def error:` intercepts every
`break` in the program: the label stops catching it, the def's own value is emitted as an
extra output, and evaluation continues past the `break` instead of unwinding.

succinctly keeps `break` primitive, so a shadowing `def error` has no effect on it.

Verified live against `jq-1.7.1-apple`:

```console
$ jq            -nc 'def error: "s-error"; label $out | 1, break $out'
1
"s-error"

$ succinctly jq -nc 'def error: "s-error"; label $out | 1, break $out'
1
```

Evaluation genuinely continues past the `break` in jq — a trailing term still runs:

```console
$ jq            -nc 'def error: "s-error"; label $out | (1, break $out), 2'
1
"s-error"
2

$ succinctly jq -nc 'def error: "s-error"; label $out | (1, break $out), 2'
1
```

## It is specifically `error/0`

An `error/1` def does not intercept, in either tool — which is exactly what "jq calls
`error` with no arguments" predicts:

```console
$ jq            -nc 'def error(m): "E"; label $out | 1, break $out'   # 1
$ succinctly jq -nc 'def error(m): "E"; label $out | 1, break $out'   # 1
```

And with no `def error` at all the two already agree (`1`).

## Root cause

jq's `parser.y` builds `break $x` as a load of the `*label-x` sentinel followed by
`gen_call("error", gen_noop())` — a *named* call, not an opcode. Name resolution then binds
that `error` to whatever is in scope, so a user def wins, gets `.` (the sentinel) as its
input, ignores it, and returns its own value as an ordinary output. The label's own handler
only ever fires for the sentinel *error*, which is never raised.

succinctly represents `break` as a dedicated `Expr` node evaluated directly, with no name
lookup, so nothing in scope can intercept it.

## Why fix it

ADR-0018 rule: `succinctly jq` follows jq bug-for-bug, and none of the three divergence
exceptions apply here — the output is readable, nothing is corrupted or discarded, and the
process stays up. This is the reference tool's own inconsistency, which is exactly the
category the rule says to reproduce.

Small and rare in practice (it needs a `def error:` at arity 0 *and* a `label`/`break` in
the same program), but it is a silently wrong *number of outputs*, not just a wrong value.

## Suggested next step

Lower `break $x` to the same shape an ordinary `error` call takes — a `FuncCall`
named `error` at arity 0 over the sentinel value — so `resolve.rs` binds it through normal
scope, rather than to a dedicated node the resolver never sees. The sentinel and the
label-catch machinery can stay exactly as they are; only the *dispatch* of the raise needs
to become a resolvable call.

Worth checking at the same time whether `error/0` shadowing affects any other construct jq
desugars through a named call, and whether `succinctly yq` needs a carve-out (real yq has no
`label`/`break` at all, so most likely the change is jq-mode-only).

newhoggy:
## Summary

Real jq does not treat `break` as a primitive — its parser desugars `break $x` into a named call to `error/0` (`gen_call("error", gen_noop())` in `parser.y`), resolved through ordinary lexical scope. A `def error:` (arity 0) in scope at the break's own position therefore shadows it: the label stops catching it, the def's own value becomes an ordinary extra output, and evaluation continues past the break instead of unwinding.

succinctly represented `break` as a dedicated `Expr` node evaluated directly with no name lookup, so a shadowing `def error:` had no effect — a silently wrong *number* of outputs, not just a wrong value. Per ADR-0018, this is the reference tool's own inconsistency, in the category the rule says to reproduce bug-for-bug.

`parse_break_expr` now inlines the same shadow-gate every other special form (`error`, `limit`, `until`, ...) already goes through: on `Parser::shadowable_defs.contains("error")`, the freshly-built `Expr::Break(name)` is wrapped via `wrap_shadowable_call("error", ..)` into a `FuncCall{name: "error", args: [], builtin_fallback: Some(Break(name))}` that `resolve.rs`'s existing three-stage mechanism already knows how to unwrap. Three small additions to `resolve.rs`'s existing match arms: arity 0 in `builtin_fallback_arity`, an empty `Vec` in `builtin_fallback_into_args`, and `Expr::Break` added to `wrap_shadowable_call`'s recognized-shape allowlist.

On a program with no `def error` anywhere — the overwhelming majority — `Parser::shadowable_defs`'s fast-reject means `break` parses to the exact same `Expr::Break` node it always has: confirmed by the full existing test suite (99 break-touching tests in `jq_cli_tests.rs` alone) passing completely unmodified.

Deliberately does not carry jq's own `{"__jq":N}` sentinel value as the wrapped call's input — a shadowing `def error: .;` observes the ambient `.` instead. Recorded in `docs/compliance/jq/limitations.md` rather than chased.

Also documents and defers a second, related gap found during investigation: `label`'s own error re-raise is *also* `gen_call(error)` in real jq, resolved at the label's own position — filed separately as #2840 (a materially larger change spanning both evaluators' label-catching arms).

Fixes #2687.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 1848 jq_cli_tests + all others, all green — every existing break/label test passes unmodified)
- [x] Verified against pinned jq 1.7.1 for every case in the issue's own repro table, plus position-dependent-scope, try/catch-doesn't-catch-the-value, and repeat/limit cases
- [x] New `resolve.rs` unit tests pinning both the shadowed and not-shadowed AST shapes directly
- [x] New CLI tests in `jq_cli_tests.rs` and `yq_cli_tests.rs` (yq mode picks up the fix for free — `label`/`break`/`def` are all ungated succinctly extensions there)

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -5431,6 +5431,64 @@ whole document that a streaming arm never read, and the M2 path was gated to kee
 from disagreeing. The gate is gone; the entry under "Real-time stdout/stderr interleaving"
 records the 19 rows that moved away from jq and the spelling rule it left in place.
 
+### `break $x` shadowed by a `def error:` observes the ambient input, not jq's own `{"__jq":N}` sentinel (#2687)
+
+[#2687](https://github.com/rust-works/succinctly/issues/2687): real jq does not treat `break`
+as a primitive — its parser desugars `break $x` into a named call to `error/0`
+(`gen_call("error", gen_noop())` in `parser.y`), applied to a synthetic sentinel value
+(`{"__jq":N}`, `N` the label's compile-time index) as `.`. Name resolution then runs as it
+would for any other call, so a `def error:` (arity 0) in scope at the break's own position
+shadows it: the label stops catching it, the def's own value becomes an ordinary output, and
+evaluation continues past the break instead of unwinding.
+
+succinctly reproduces this (`src/jq/parser.rs`'s `parse_break_expr` wraps `break $x` as a
+shadowable `error` call the same way `error`, `limit`, `until`, ... already are, per
+`Parser::shadowable_defs`/`wrap_shadowable_call`, #2036), but does **not** carry jq's
+`{"__jq":N}` sentinel as the wrapped call's input — the shadowing def instead sees the ambient
+`.` the break itself was reached with:
+
+```console
+$ jq            -nc 'def error: .; label $out | break $out'
+{"__jq":0}
+$ succinctly jq -nc 'def error: .; label $out | break $out'
+null
+```
+
+**Why not carried.** The sentinel is observable only through a `def error:` whose body reads
+`.` — `def error: "S";` (the shape #2687 was actually filed over, and every case in its own
+repro table) is unaffected, since a literal body never looks at its input. Carrying it would
+mean the break can no longer fall back to a bare `Expr::Break` node when nothing shadows it —
+the overwhelmingly common case, and the one this fix's whole design (`Parser::shadowable_defs`'s
+text-only fast-reject) keeps at zero cost and byte-identical AST. It would instead need
+`Expr::Pipe(vec![Expr::Literal(<sentinel>), <the wrapped call>])`, widening the shape every
+break-aware evaluator arm has to recognize (`needs_path_context`, `owned_identity_step`,
+`path_context_step_generic`, `step_can_yield_absent`, `path_context_stage_preserves_node`, …)
+for a construction with no purpose outside probing this one divergence. Per ADR-0018's decision
+order: the reference behavior here is readable and nothing is corrupted, discarded, or takes the
+process down (no rule-4 condition), so the order reaches step 2 (closer match) — but the cost
+this row would add to every break site, shadowed or not, makes the narrower fix (#2687 itself,
+closing the *shadowing* gap) the one taken, with this payload detail recorded rather than chased.
+
+A second, related gap is **not** covered by #2687 or this entry: `label`'s own error re-raise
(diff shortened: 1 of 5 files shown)


newhoggy:
Ran a high-effort code review (5 finder angles) before merging. No blocking correctness bugs — the core shadowing mechanism was independently verified against live jq 1.7.1 by multiple reviewers, and every existing break/label test passes unmodified.

Addressed in a follow-up commit:
- The sentinel-payload divergence entry in `limitations.md` understated its severity (only showed the "sees null" case). The substitute value is actually unbounded — whatever `.` happens to be at the break's textual position, however unrelated to the label. Rewrote the entry and added a CLI test pinning the exact shape.

Filed as follow-ups rather than fixed here:
- #2843 — a *pre-existing* divergence (reproduces on `main` before this PR, with plain `error` alone) that this PR's break-desugaring newly makes reachable via `break` too: a shadowed call inside `path(...)` under an enclosing `label` wrongly raises `Invalid path expression`.
- #2840 (already noted in the PR description) — `label`'s own error re-raise has the same desugaring-to-`error/0` shape in real jq, a materially larger fix spanning both evaluators' label-catching arms.

One minor reuse finding (a small duplicated fast-reject pattern between `parse_break_expr` and `parse_shadowable_special_form`) was reviewed and declined — the two are intentionally kept separate per `parse_break_expr`'s own doc comment, and extracting a helper for a 4-line pattern used twice wasn't judged worth the churn.

Issue #2403, opened 2026-09-04
jq: eval.rs's own to_owned_at_depth and yq_runner's to_owned_canonicalizing_numbers_at_depth likely have the same container_gap_ok gap #2358 closed


newhoggy:
Found while implementing #2358.

## Background

#2349's own divergence matrix named five hand-copied object/array traversals sharing this codebase's delimiter/gap-check family (#1677/#2211/#2243/#2262). #2358 (row 2 of that matrix) audited `eval_generic.rs`'s own cursor-less `to_owned_at_depth` and found that #2211's `container_gap_ok` check (a stray `,` with zero real children, `[,]`/`{,}`) — documented in `docs/compliance/jq/limitations.md` as expected to stay permanently open, since the function only ever receives a bare `value: &V`, never a cursor for the container itself — was actually closable: every recursive call site inside the function already resolves the child's own cursor for an unrelated check (the #1677 delimiter check, or `text_position()` for the array gap check) and simply discarded it. Threading it through as a new `cursor: Option<&V::Cursor>` parameter closed the gap for every nested container (all except the true top level, which by design has no cursor).

## What's still open

The same `docs/compliance/jq/limitations.md` entry names **two more** materializers with the identical original shape and the identical "expected to stay open" reasoning, neither audited by #2358 (whose own scope named only the `eval_generic.rs` function):

- `eval.rs`'s own separate `to_owned_at_depth` (a different function in a different file, behind this crate's documented public `eval()` API)
- `yq_runner.rs`'s `to_owned_canonicalizing_numbers_at_depth` (behind `succinctly yq --slurp`/`--eval-all`/`--inplace --input-format json`)

Given #2358's own investigation found the "no cursor available" framing was true of the *original* signature but not a genuine constraint of the call graph, it's plausible both of these have the same recoverable gap — but this needs the same audit #2358 did (`eval.rs`'s own `to_owned_at_depth` is a large, separately-written function; `yq_runner.rs`'s is a third, YAML-flavored copy with its own recursive structure — neither should be assumed to close the same way without checking each caller).

## Suggested next step

[...]

Issue #2358, opened 2026-09-04
jq: two non-live rows of #2349's delimiter matrix are untracked — lazy.rs missing trailing_element_gap_ok, and to_owned_at_depth's {,} residual


newhoggy:
#2349 built a divergence matrix across the five hand-copied object/array traversals #1803
names, and closed the one column whose gaps are **live**. Two rows in that matrix are
non-live but real, and neither is tracked anywhere as work. Filing them together so they are
not rediscovered a fourth time.

## Row 1 — `lazy.rs`'s `cursor_to_owned_at_depth` is missing `trailing_element_gap_ok`

Verified against `main` at `90ecf68f8`: the function implements
`preceding_delimiter_ok` (array arm), `container_gap_ok` (both arms), and
`key_delimiter_ok`/`value_delimiter_ok` (object arm) — it has every check its siblings gained
from #1677 and #2211 — but **not** #2243/#2262's `trailing_element_gap_ok`. A grep over the
function body finds no call to it in either arm.

**Currently masked, not live.** This function is reached via `-e`/`--exit-status`, which runs
`to_owned_at_depth`/`to_owned_cursor_at_depth` alongside it, and the stricter sibling raises
first — so `[1,]`/`{"a":1,}` is caught today regardless. The gap is latent.

**Why it should still be closed.** "Masked by a sibling that happens to run alongside" is not
an invariant; it is a property of the current call graph, and #1803's whole history is
follow-ups quietly changing which traversal a feature reaches. #1645's own site started life
as `select`'s private helper and now gates four features. The cost is one call per container,
matching what its siblings already pay.

## Row 2 — `eval_generic.rs`'s cursor-less `to_owned_at_depth` still cannot check `{,}`/`[,]`

#2262 closed the trailing-comma case for this function but left #2211's `container_gap_ok`
(a stray `,` with **no** real child) open, and documented why in
`docs/compliance/jq/limitations.md`: `container_gap_ok` needs the *container's own* cursor to
find its opening bracket, and this function only ever receives a bare `value: &V`. Once the
child walk is exhausted there is nothing left to reconstruct one from.

The limitations doc records this as expected to stay open. That is a reasonable reading of
the constraint as the signature stands — but it has not actually been checked whether the
callers can *supply* a container cursor, which is the question that would settle whether this
is a real limitation or just the shape the function was first written in. Its cursor-carrying
sibling `to_owned_cursor_at_depth` takes exactly such a parameter and closes the case with it;
[...]

PR #2845, opened 2026-09-12
fix(yq): pass the root cursor into the JSON DOM bridge, close #2781's top-level [,] leak

### LINKED ISSUE #2781: yq: top-level [,] still accepted on --slurp/--eval-all/--inplace after #2279 (inplace rewrites the file)
**Severity: Medium** — `--inplace` rewrites a file real yq refuses to read, so a malformed document is silently "repaired" on disk.

## What's wrong

#2279 made `succinctly yq --input-format json` reject malformed flow-sequence commas at parse time, which fixed every route that goes through `YamlIndex`. The `--slurp`/`--eval-all`/`--inplace` family materializes through `JsonIndex`'s DOM bridge instead, and one shape still gets through there: a **top-level** `[,]`.

```console
$ printf '[,]' > t.json

$ succinctly yq -p json -o=json -I=0 '.' t.json
Error: YAML parse error in t.json: unexpected character ',' at offset 1: expected value or ']' in flow sequence   # exit 1, correct

$ succinctly yq -p json -o=json -I=0 --slurp '.' t.json
[[]]                                  # exit 0 -- wrong
$ succinctly yq -p json -o=json -I=0 --eval-all '.' t.json
[]                                    # exit 0 -- wrong

$ succinctly yq -p json -i '.' t.json ; cat t.json
[]                                    # exit 0, FILE REWRITTEN

$ yq -p json -i '.' t.json ; cat t.json
Error: bad file 't.json': json: null unexpected end of JSON input
[,]                                   # real yq: exit 1, file untouched
```

Only the top-level case leaks. Nested ones are already refused on these routes:

```console
$ printf '[[,]]' | succinctly yq -p json -o=json --slurp '.'
Error: Invalid JSON text: expected JSON value, found ','
```

## Why it is top-level only

This is the `container_gap_ok` shape: the check needs a cursor for the container's *own* opening bracket to measure the gap to its closer, and the DOM bridge has no cursor for the outermost container — the same reasoning `child_tail_gap_ok`'s doc comment already gives for why it cannot cover `{,}`/`[,]` without one.

## Why this is worth fixing rather than leaving

The two route families now **disagree** on accept/reject for the same bytes, which they did not before #2279 (both previously accepted: stdout gave `[null]`, `-i` gave `[]`). A disagreement is worse than a shared gap, because the answer depends on which flag the user passed. And the `--inplace` case is the destructive one: exit 0 plus a rewritten file, for input the reference tool will not read at all.

## Note on #2403

[#2403](https://github.com/rust-works/succinctly/issues/2403) tracked the `[,]`/`{,}` zero-child variant on the DOM bridge and is **closed as completed**, but this top-level residue survives it — so this is not a duplicate, it is the part that closure did not reach.

## Found via

Adversarial review of [#2780](https://github.com/rust-works/succinctly/pull/2780), which also corrected that PR's own `limitations.md` claim that the fix covered "every route".

Related: #2279, #2403, #2358, #2262.

newhoggy:
## Summary

`--slurp`/`--eval-all`/`--inplace`'s `--input-format json` route materializes through `JsonIndex`'s DOM bridge (`yq_runner.rs`'s `to_owned_canonicalizing_numbers`), not the `YamlIndex` parser #2279 fixed. A stray comma with zero real children at the *true top level* (`[,]`, `{,}`) still leaked through: `parse_input`'s JSON arm called the cursor-less wrapper, so `tail_gap_ok` fell back to `child_tail_gap_ok`, which can only see a last child. Every nested level already got a real cursor this same way (#2403), so the "no cursor for the outermost container" premise the old doc comments stated was false — `parse_input` already holds the root cursor, it was simply never threaded down.

`--inplace` is the destructive case: it rewrote a file containing `[,]` to `[]` on disk, where real yq v4.53.3 refuses to read `[,]` at all and leaves the file untouched.

Fix: pass `Some(&cursor)` from `parse_input`'s JSON arm into `to_owned_canonicalizing_numbers_at_depth` instead of the depth-0 `None` wrapper, which is now removed (no real caller left) so the gap can't be silently reintroduced by a future call reaching for the wrong entry point.

Also makes top-level `{,}` consistent with the DOM bridge's own nested `{,}` rejection (already refused since #2403) rather than carving out a top-level exception to preserve an accidental agreement with real yq — recorded under #2777's existing flow-mapping delimiter tracking, not a new special case.

This issue already carried a detailed triage plan from the repo owner, posted the same day — followed closely, with every claim re-verified live against the pinned yq v4.53.3 binary before implementing.

Fixes #2781.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 1854 yq/jq CLI tests + all others, all green)
- [x] Verified against pinned yq v4.53.3 for every case in the issue's own repro table, plus the `{,}` consistency claim
- [x] Flipped the "known gap" unit test into one asserting the fix; added a well-formed-empty-container regression guard
- [x] New CLI tests covering `--slurp`/`--eval-all`/`--inplace`, including `--inplace`'s file-untouched-on-error assertion

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -1642,8 +1642,21 @@ practice #2243's own issue text cites for not folding into #2211):
   before the tail check -- and now thread it through the same way, via a shared
   `tail_gap_ok` helper (`document.rs`) extracted in review rather than a third and fourth
   hand-copy of the `container_tail_gap_ok`/`child_tail_gap_ok` dispatch. Only the true top
-  level (`to_owned`'s and `to_owned_canonicalizing_numbers`'s own depth-0 entry points, no
-  cursor to give) keeps the gap, matching `eval_generic.rs`'s own residual scope above.
+  level (`to_owned`'s own depth-0 entry point, no cursor to give) keeps the gap, matching
+  `eval_generic.rs`'s own residual scope above.
+
+  **Update (#2781, yq mode)**: `to_owned_canonicalizing_numbers`'s own depth-0 entry point
+  -- the true-top-level exception this note originally named alongside `to_owned`'s -- is
+  closed. `parse_input`'s JSON arm (`yq_runner.rs`, the only caller reaching this
+  materializer, for `--slurp`/`--eval-all`/`--inplace --input-format json`) already held the
+  root cursor and simply was not threading it through; `to_owned_canonicalizing_numbers`
+  is now folded into `to_owned_canonicalizing_numbers_at_depth`, whose `cursor` parameter is
+  a bare `&V::Cursor` (no longer `Option`) since every call site in that file has one.
+  `to_owned`'s own top-level gap (jq mode, `eval.rs`/`eval_generic.rs`) is untouched by
+  #2781 -- currently not independently reachable there (parser-level validation intercepts
+  a top-level `[,]`/`{,}` before `to_owned` runs on every route probed), but that is
+  incidental protection from a different layer, not a guarantee `to_owned_at_depth`'s own
+  signature enforces the way `to_owned_canonicalizing_numbers_at_depth`'s now does.
 - **#2263**: `jq_runner.rs` still carries its own independent, hand-copied
   `trailing_gap_ok`/`scalar_end_pos` pair rather than the new trait methods --
   a cleanup, not a behavior gap, but the same "duplicated predicates diverge
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -945,14 +945,25 @@ input), so `[1,]`, `[,1]`, `[1,,2]`, `[,]` and their nested forms error on every
 parses through `YamlIndex` — the plain stdout path this issue was filed about, for every
 filter.
 
-**One shape is still accepted, and it is on the other family of routes**: a *top-level*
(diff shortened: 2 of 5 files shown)


newhoggy:
Ran a code review before merging. No correctness bugs found — confirmed sound by two independent finder angles (removed-behavior audit, cross-file tracer). Addressed in a follow-up commit:

- Narrowed `to_owned_canonicalizing_numbers_at_depth`'s `cursor` parameter from `Option<&V::Cursor>` to a bare `&V::Cursor` — every call site in this file now always has one, so the gap this PR fixes can't be silently reintroduced by a future call reaching for a `None` fallback that no longer needs to exist.
- Fixed two stale doc references review found: an in-code comment still naming the now-deleted wrapper function, and `docs/compliance/jq/limitations.md`'s #2403 note, which grouped `to_owned`'s and `to_owned_canonicalizing_numbers`'s top-level gaps together — only the latter is closed by this PR.

Filed as a follow-up rather than expanded into this PR: #2846 — `eval.rs`'s and `eval_generic.rs`'s own `to_owned_at_depth` still carry the identical `Option<&C>`-with-a-`None`-top-level shape this is the third bug found in (#2358, #2403, this PR). Confirmed live it's not a bug today (parser-level validation protects jq mode's own routes upstream of `to_owned`), but it's a structural risk with no type-level guard the way this file's sibling now has.

Issue #2847, opened 2026-09-12
perf(jq): -s/--slurp still materializes a full DOM before validating/evaluating


newhoggy:
**Severity: Low** (no wrong data; a memory/architecture follow-up)

## Summary

#2662 moved `-S`/`-a`/`-C` onto the lazy route so they validate/materialize only what a filter actually reads, closing three of the five rows in #2103's "materializing flag routes still validate whatever the filter" divergence. `-s`/`--slurp` and the `-n`/`input` bridge were deliberately left out of that PR:

- `-n`/`input` is the *correct*, final answer, not deferred work: `input` must hand the evaluator an owned value it can return from a builtin, so materializing what it reads is the rule working as intended (`docs/compliance/jq/limitations.md`, "The materializing flag routes still validate whatever the filter").
- `-s`/`--slurp` is real remaining work. It still goes through `evaluate_input_streaming`, building a full `Vec<OwnedValue>` DOM before the filter runs at all, so e.g. `succinctly jq -sc '1+1'` on a 10MB malformed-or-otherwise document still exits 5 (validates everything) and pays the whole-document `OwnedValue` cost, where the default route now answers `2` at a fraction of the memory (#2662 measured `-S '1+1'` on a 7MB `users` fixture at 86.5MB → 11.3MB peak RSS for the routes it did move).

## Ask

Move `-s`/`--slurp` onto the lazy route too, so it materializes only the values a filter actually reads rather than the whole input up front.

## Suggested approach (from #2662's own triage)

`find_json_values(raw)` already yields each top-level document's byte span without materializing it. Slurp could synthesize one document — `[` + spans joined by `,` + `]` — into a single fresh `Vec<u8>`, index *that* once, and run the ordinary lazy branch over it with `at` fixed to the originating file, rather than building an `OwnedValue` array via the current DOM path.

The fiddly part: jq's own slurp parse-error messages name the line of the fault in the *original* stream (e.g. `{123:1} 2` → "line 1, column 5"), so the lazy branch's `line_at` would need to map the synthetic offset (inside the synthesized `[...]` buffer) back to the original stream's offset — a small monotone table subtracting the `[`/`,` bytes inserted before each span should do it, but the exact error text needs to be captured from the pinned jq binary before implementing, not assumed. `--slurpfile`/`--rawfile`/`-R -s` are a different, unrelated code path and are out of scope here — only bare `-s`/`--slurp` on ordinary JSON input.

[...]

Issue #1830, opened 2026-08-29
jq: --raw-output0 performs no NUL-content validation, unlike real jq (exit 0 where jq exits 5)


newhoggy:
**Severity: Low** (ambiguous output stream to a NUL-delimited consumer; exit 0 where jq exits 5)

Split out of #1709's post-mortem (scope gap 5), which found that jq mode has the identical
unfixed gap in a completely different mechanism -- `jq_runner.rs` has no `Terminator`/`ColorSink`
machinery to reuse, so the fix shape differs and it wants its own issue.

## The gap

Real jq refuses `--raw-output0` when a string's own content contains a NUL:

```console
$ printf '{"a":"b\x00c"}' | jq -r --raw-output0 '.a'
jq: error (at <stdin>:1): Cannot dump a string containing NUL with --raw-output0 option
# exit 5
```

`succinctly jq` emits the NUL inline and exits 0, making the record boundary ambiguous to any
NUL-delimited consumer (`xargs -0`, `read -d ''`).

## Where

`write_terminator` / `write_output` / `write_output_jq_value` in
`src/bin/succinctly/jq_runner.rs`. None of them scans content, and there is no shared
scalar-emission chokepoint they funnel through.

## The design constraint the yq side already established

#1709's own closed PR (#1767) proved that "buffer the whole rendered result, scan it, then write"
is the wrong shape -- it discards valid earlier results in a multi-result stream, and it forces
full in-memory materialisation (+65% peak RSS measured on a 100 MB input). Real jq, like real yq,
**flushes each result as it is produced and errors on first sighting a NUL**.

So this wants an **incremental NUL-scanning writer wrapper**: a thin `Write` adapter that scans
each chunk as it goes and errors on the first NUL byte, preserving O(1) memory for the common
NUL-free case and preserving already-emitted output.

Confirm jq's own flush-then-error ordering against the pinned binary before implementing.
yq's was verified live -- `yq -0 '.[]'` over a sequence whose third element contains a NUL emits
the first two records to stdout *then* errors on stderr, exit 1 -- but jq's `--raw-output0`
equivalent should be checked the same way rather than assumed symmetric.

## Acceptance

- `jq -r --raw-output0` on an embedded-NUL string errors with jq's own wording and exit 5.
- Earlier results in a multi-result stream are flushed before the error, matching jq.
- No measurable peak-RSS increase on a large NUL-free `--raw-output0` query.
- Regression tests for the single-result and multi-result shapes.

## Related

[...]

PR #2848, opened 2026-09-12
fix(jq): move -S/-a/-C onto the lazy route, validating only what's read

### LINKED ISSUE #2662: jq: -S/-a/-s/-C/-n input still validate the whole input where the default route validates only what it reads (#2103 option 1)
**Severity: Low** (no wrong data; the answer depends on a CLI flag)

## Summary

#2103 (PR #2652) made the default M2 route validate only what a filter reads: `1+1` on a malformed document answers `2`. The routes that go through `evaluate_input_streaming` did not move, because they materialize the whole input into an `OwnedValue` before evaluating anything:

```console
$ printf '%s' '{123:1,"b":2}' | succinctly jq -c  '1+1'        # 2, exit 0
$ printf '%s' '{123:1,"b":2}' | succinctly jq -Sc '1+1'        # Invalid JSON text, exit 5
$ printf '%s' '{123:1,"b":2}' | succinctly jq -ac '1+1'        # exit 5
$ printf '%s' '{123:1,"b":2}' | succinctly jq -sc '1+1'        # exit 5
$ printf '%s' '{123:1,"b":2}' | succinctly jq -Cc '1+1'        # exit 5
$ printf '%s' '{123:1,"b":2}' | succinctly jq -nc 'input | 1+1'  # exit 5
```

`-e` is a second, smaller shape: it streams the input but materializes each *output* to decide the exit status, so `-e '.,.'` on `{"\ud800":1,"\ud800":2}` raises the #1642 collision where `-c '.,.'` echoes.

Real jq 1.7.1 exits 5 on every row above, so the flag routes are the ones closer to the oracle; the default route is the recorded #2103 divergence. Both behaviours are the recorded rule ("a route that materializes validates what it materializes") — the problem is that which one applies is decided by a flag, by the same accident #2103 removed from the default route: those routes materialize the input because they always did, not because `-S '1+1'` needs it. Documented and pinned by `test_materializing_flag_routes_still_validate_2103` in PR #2652 (code review, option 2); this issue is option 1.

## Ask

Move `-S`, `-a`, `-s`, `-C` and the `-n`/`input` bridge onto cursor evaluation so they materialize only the *outputs* they need (sorting, ASCII escaping, slurping and colouring are all output-side transforms), and then validate only what the filter reads. `-e` should keep materializing outputs — that is reading them. Measure the memory side too: on those routes a 10 MB `-S '1+1'` pays the whole-document `OwnedValue` today.

## Related

- #2103 / PR #2652 (the decision and its option-2 record), #2644 (the evaluator-side gates), #2645 (yq mode), #1642 (the collision the `-e` row raises).

newhoggy:
## Summary

- Closes three of #2103's five "materializing flag routes still validate whatever the filter" rows: `-S` (sort_keys), `-a` (ascii_output), and `-C` (color_output) forced the whole document through `evaluate_input_streaming`'s `OwnedValue` materialization before a filter ever ran — so `1+1` on a malformed document exited 5 under any of them where the default lazy route already answered `2`, purely for historical reasons, not because sorting/escaping/coloring *output* needs the *input* validated.
- None of the three actually need it: `write_output_jq_value` already materializes just the one *output* value being printed when `sort_keys`/`color_output`/`ascii_output` is set, reusing `format_json`'s existing `ascii`/`sort_keys` options (already correct on the pre-existing DOM route). Dropped the three flags from `can_use_lazy_path`; excluded `color_output`/`ascii_output` from `can_json_fast_path` (the M2 fast path takes `sort_keys` directly but has no color/ascii handling of its own).
- **Two real bugs found while verifying against jq 1.7.1, fixed alongside:**
  - `-a` wins over `-r`/`-j` for a *string* value in real jq (`-acr '"café"'` prints `"café"`, quoted and escaped, not the unquoted raw string `-r` alone gives — a non-string value is unaffected). Both `write_output_jq_value` and `write_output` had the same latent gap; fixed identically on both.
  - `write_output_jq_value`'s own materialize arm converted a decode failure to a bare `anyhow::anyhow!` instead of `MalformedJsonError`, skipping jq's diagnostic channel entirely (exit 1 instead of exit 5) — unreachable before this change since `sort_keys`/`color_output` couldn't reach the lazy path at all; first exercised by moving `-S`/`-C` onto it.
- **Measured** (Apple M-series, `-S '1+1'` on a 7MB `users` fixture): peak RSS 86.5MB → 11.3MB, instructions retired 5.17B → 268M.
- `-s`/`--slurp` is real remaining work (needs an offset-mapping redesign to slurp without a DOM) — filed separately as #2847. `-n`/`input` stays materializing deliberately: `input` must hand the evaluator an owned value it can return from a builtin, so validating what it reads there is already the rule working as intended, not an accident.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite green, including the full `jq_cli_tests` battery and the `fuzz_select_and_materialize_agree_on_corruption_1965` property re-run 3× for confidence)
- [x] `test_materializing_flag_routes_still_validate_2103` rewritten row-by-row for the new behavior; `test_jq_missing_delimiter_raises_under_sort_keys_mid_stream_1643` updated (its old assertion was itself pinning a divergence from jq — verified live that real jq prints the first well-formed value before erroring on a later malformed one)
- [x] New tests: well-formed `-S`/`-a`/`-C` output unaffected, the `-a`-wins-over-`-r` fix on both routes, the `MalformedJsonError` routing fix
- [x] Manually diffed against the pinned `/usr/bin/jq` 1.7.1 oracle throughout
- [x] Must-not-change sweep: `-s`/`-n` still materialize (unchanged rows in the rewritten test)

Closes #2662

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -3731,7 +3731,11 @@ the three actually need it: `write_output_jq_value` already materializes just th
 *output* value being printed when `sort_keys`/`color_output`/`ascii_output` is set (reusing
 `format_json`'s existing `ascii`/`sort_keys` options, which already handled all three
 correctly for the pre-existing DOM route), so moving them onto the lazy route costs nothing
-in fidelity and validates only the value a filter actually reads, same as the default.
+in *this* fidelity concern and validates only the value a filter actually reads, same as the
+default. (A separate, pre-existing number-formatting gap in that same `format_json` call --
+`--preserve-input` silently reformatting numbers under any of `-S`/`-a`/`-C`/`-s` -- predates
+this change and is tracked independently as
+[#2852](https://github.com/rust-works/succinctly/issues/2852), not claimed fixed here.)
 `-s` (slurp) and the `-n`/`input` bridge stay materializing — `-s` needs a real
 offset-mapping redesign to slurp without building a DOM (tracked, not done), and `input`
 must hand the evaluator an owned value it can return from a builtin by construction, so
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -3720,20 +3720,47 @@ The context that makes the loss survivable is the one #2168's entry states: succ
 already diverges on this whole class of document through `.` itself, deliberately, and a
 user who wants jq's rejection has `--validate` and `succinctly json validate`.
 
-**The materializing flag routes still validate whatever the filter.** `-S`, `-a`, `-s`, `-C`
-and the `-n`/`input` bridge do not take the M2 route at all: `evaluate_input_streaming`
-materializes the whole input into an `OwnedValue` *before* evaluating, so on those routes
-`1+1` on `{123:1,"b":2}` still exits 5, exactly as it did before #2103, while the default
-route answers `2`. `-e` streams like the default but materializes each *output* to decide the
-exit status, so `-e '.,.'` on `{"\ud800":1,"\ud800":2}` still raises the collision where
-`-c '.,.'` echoes. Both are the rule applied to a route that materializes — the flag routes
-materialize the input, `-e` the outputs — not an exception to it, but they make the answer
-depend on a flag in the way this entry's point 2 objects to for filters, and by the same
-accident: those routes materialize the input because they always did, not because `-S '1+1'`
-needs it. Recorded here so the dependence is stated rather than discovered; moving those
-routes onto cursors so they too validate only what they read is
-[#2662](https://github.com/rust-works/succinctly/issues/2662).
-Pinned by `test_materializing_flag_routes_still_validate_2103`.
+**The materializing flag routes still validate whatever the filter — down to `-s` and
+`-n`/`input` now, closed by [#2662](https://github.com/rust-works/succinctly/issues/2662)
+for `-S`/`-a`/`-C`.** `-S` (sort keys), `-a` (ASCII output) and `-C` (color) used to force
+`evaluate_input_streaming`, which materializes the whole input into an `OwnedValue` *before*
+evaluating — so `1+1` on `{123:1,"b":2}` exited 5 under any of them where the default route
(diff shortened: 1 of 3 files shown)


newhoggy:
Ran an 8-agent code review after the initial push. Found and fixed two real correctness bugs beyond the ones I'd already caught myself:

1. **`--seq`'s RS-separator byte was keyed on the wrong side of the `-a`-wins-over-`-r` fix.** `--seq -acr` on a string emitted a spurious leading RS byte — confirmed live that real yq's RS-suppression decision depends on whether `-r` alone would have made the value raw-shaped (a string, full stop), independent of whether `-a` later overrode the actual formatting. Fixed in both `write_output_jq_value` and `write_output`, preserving the pre-existing #1830 NUL-check-before-any-write ordering.
2. Corrected two comments (flagged independently by 4 of 8 review passes) that claimed ascii output reuses `yq_runner.rs`'s `AsciiEscapeWriter` — that type isn't used anywhere in this file; the actual mechanism is materializing the one output value and reusing `format_json`'s existing options.

Also filed two follow-ups for real, pre-existing (not introduced by this PR) issues the review surfaced, both narrowly scoped and out of place to fix inline here:
- #2850 — `-S`/`-a`/`-C`/`-e` leak a raw panic backtrace to stderr past 256 levels of nesting (exit code and message stay correct; `-e` alone already had this on `main` before #2662, which widens the exposure to 3 more flags). A proper fix needs a hand-written checked sibling to `lazy.rs`'s own materializer — delegating to `eval_generic.rs`'s already-checked version isn't safe, since it reformats numbers differently (canonicalizes) than the raw-byte-preserving one this route needs.
- #2852 — `--preserve-input` silently reformats numbers under `-S`/`-a`/`-C`/`-s` (pre-existing gap in `format_json`, unrelated to this PR's own claims about sort/ascii/color fidelity).

Full local suite + both clippy configs green again; pushed and re-running CI now.

Issue #2858, opened 2026-09-12
jq: a `?//`-alternatives fold under `|=`/`=`/`del()` silently no-ops instead of raising jq's own path error


newhoggy:
**Severity: Low** (write-side no-op / silent data loss on a narrow, already-refuse-documented shape)

## Summary

`docs/compliance/jq/limitations.md` documents `reduce`/`foreach`'s `?//`-alternatives pattern
as a **refuse-only** divergence in path position: `resolve_node`'s dispatch arm only ever
admits a single pattern alternative (`patterns.len() == 1`), so a `?//` chain always falls
through to `resolve_leaf`'s catch-all, and `path(...)` on one is expected to raise.

That premise is true for `path(...)` read alone, but **not for the write forms** (`|=`, `=`,
`del()`) — on those, `resolve_leaf`'s catch-all evaluates the fold as an ordinary, untracked
*value* (via #1365's own `?//`-retry machinery, which only exists in the value evaluator), and
the write operator then silently **no-ops** instead of raising, leaving the document unchanged
at exit 0. Confirmed live against jq 1.7.1 and the pinned reference binary — jq raises exit 5
for every row below; `succinctly jq` exits 0 with the input echoed back unchanged.

## Repro (jq 1.7.1 vs succinctly jq)

```console
$ echo '{"a":false,"c":false,"d":[1],"x":{"a":false,"c":1}}' \
  | jq -c '(foreach (.a|.b?) as {a:$v0,d:$w0} ?// {c:$v0} (.a; (if true then $v0 else 1 end); $v0)) |= .'
jq: error (at <stdin>:0): Invalid path expression with result false

$ echo '{"a":false,"c":false,"d":[1],"x":{"a":false,"c":1}}' \
  | succinctly jq -c '(foreach (.a|.b?) as {a:$v0,d:$w0} ?// {c:$v0} (.a; (if true then $v0 else 1 end); $v0)) |= .'
{"a":false,"c":false,"d":[1],"x":{"a":false,"c":1}}
```

Three more, same shape (`?//`-alternatives fold under a write operator, all exit 0 with the
document unchanged here, exit 5 in jq):

```console
$ echo '{"a":"s","c":"s","d":1,"x":{"a":"s","c":1},"arr":["s","s",{"b":1}]}' \
  | jq -c 'del(foreach .c as {a:$v0} ?// $v0 (null; (.arr[-1]?|$v0); ($v0|first(.[]?))))'
[...]

Issue #2046, opened 2026-08-31
jq: FoldRegister doesn't expose its register to UPDATE/EXTRACT, so a $var can't re-establish tracking inside a fold


newhoggy:
**Severity: Low** (refuse-only — visible exit 5, no document written)

## Summary

`FoldRegister` knows where the path register is, but does not make it available to the
resolution of its own `UPDATE`/`EXTRACT`. So a `$var` frozen at the register cannot
re-establish trackability inside a fold the way it now can in a plain pipe (#2041), and
navigation off that variable refuses.

This is divergence shape #1 in
[`docs/compliance/jq/limitations.md`](https://github.com/rust-works/succinctly/blob/main/docs/compliance/jq/limitations.md),
reduced to what actually remains of it after #2041.

## Repro (verified against pinned jq 1.7.1)

```console
$ echo '{"a":{"b":1},"c":2}' | jq -c 'path(. as $x | foreach (1,2) as $i (0; $x.a; .b))'
["a","b"]
["a","b"]
$ echo '{"a":{"b":1},"c":2}' | succinctly jq -c 'path(. as $x | foreach (1,2) as $i (0; $x.a; .b))'
jq: error (at <stdin>:1): Invalid path expression ...
```

The equivalent plain pipe **works** since #2041, which is what isolates this to the fold:

```console
$ echo '{"a":1}' | succinctly jq -c 'path(. as $x | 5 | $x.a)'
["a"]
```

## Root cause

`FoldRegister::enter` correctly leaves the register at the caller's ambient position when
INIT is computed (`0` never moves it) — that part already models jq. But
`FoldRegister::resolve` then calls

```rust
resolve_against_cow::<S>(expr, Cow::Owned(input), tr, snapshot)
```

with `tr = self.trackable && at_register`, and nothing else. For a computed accumulator
`at_register` is `false`, so `UPDATE` is resolved with tracking off **and no register to
compare against**. `$x` inside it produces an untracked snapshot branch, and `.a` then
raises rather than re-establishing.

`resolve_seq` solves the same problem by carrying the register on `PathBranch` and
consulting it in `reestablishes_register` (#2041) — but that register is created *inside*
one `resolve_seq` call, at `resolve_leaf`'s computed-value leaf. A fold's register comes
from outside the resolution entirely, so there is no way to inject it today.

## Why this is separate from #1573 / #2042

- #2041 closed #1573's plain-pipe half.
[...]

PR #2859, opened 2026-09-12
fix(jq): track a fold's own destructuring pattern through the path register

### LINKED ISSUE #2676: jq: reduce/foreach destructuring pattern doesn't move the path register — path(foreach .b as {c:$x} (.; .; $x)) is ["b","c"] in jq, refused here
**Severity: Low** (refuse-only across every row below — no accept-where-jq-refuses)

## Summary

jq compiles a fold's own loop-variable pattern (`reduce SOURCE as PATTERN (INIT; UPDATE)`,
`foreach SOURCE as PATTERN (INIT; UPDATE; EXTRACT)`) with the same tracked `INDEX` matchers
`bind_alternation_matchers` gives a plain `. as PATTERN` bind, and the fold's SOURCE is itself
tracked whenever it resolves through the path register rather than being an untracked
subexpression. So when SOURCE is a register-derived path (`.b`, `.a`, ...), destructuring the
loop variable moves the register the same way `. as {..}` does for #2649's plain bind —
`path(foreach .b as {c:$x} (.; .; $x))` is `["b","c"]` in real jq, not a refusal.
`succinctly jq` refuses this and every row below where jq accepts (exit 5 both sides, but
succinctly's message never distinguishes "refused because a destructuring pattern is
structurally impossible to track" from "refused only because this particular resolver has no
tracked path for it").

## Repro (jq 1.7.1 vs succinctly jq, on `{"a":[1,2,3],"b":{"c":5}}`)

```console
$ echo '{"a":[1,2,3],"b":{"c":5}}' | jq -c 'path(foreach .b as {c:$x} (.; .; $x))'
["b","c"]
$ echo '{"a":[1,2,3],"b":{"c":5}}' | succinctly jq -c 'path(foreach .b as {c:$x} (.; .; $x))'
jq: error (at <stdin>:1): Invalid path expression with result 5

$ echo '{"a":[1,2,3],"b":{"c":5}}' | jq -c 'path(foreach .a as [$x] (.; .; $x))'
["a",0]
$ echo '{"a":[1,2,3],"b":{"c":5}}' | succinctly jq -c 'path(foreach .a as [$x] (.; .; $x))'
jq: error (at <stdin>:1): Invalid path expression with result 1

$ echo '{"a":[1,2,3],"b":{"c":5}}' | jq -c 'path(foreach .a as [$x,$y] (.; .; $x))'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 0 of [1,2,3]
$ echo '{"a":[1,2,3],"b":{"c":5}}' | succinctly jq -c 'path(foreach .a as [$x,$y] (.; .; $x))'
jq: error (at <stdin>:1): Invalid path expression with result 1

$ echo '{"a":[1,2,3],"b":{"c":5}}' | jq -c 'path(reduce .b as {c:$x} (.; .))'
[]
$ echo '{"a":[1,2,3],"b":{"c":5}}' | succinctly jq -c 'path(reduce .b as {c:$x} (.; .))'
jq: error (at <stdin>:1): Invalid path expression with result {"a":[1,2,3],"b":{"c":5}}

$ echo '{"a":[1,2,3],"b":{"c":5}}' | jq -c 'path(reduce .b as {c:$x} (.; .b))'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "b" of {"a":[1,2,3],"b":{"c":5}}
$ echo '{"a":[1,2,3],"b":{"c":5}}' | succinctly jq -c 'path(reduce .b as {c:$x} (.; .b))'
jq: error (at <stdin>:1): Invalid path expression with result {"c":5}

$ echo '{"a":[1,2,3],"b":{"c":5}}' | jq -c 'path(reduce (.a) as [$i] (.; $i))'
jq: error (at <stdin>:1): Invalid path expression with result 1
$ echo '{"a":[1,2,3],"b":{"c":5}}' | succinctly jq -c 'path(reduce (.a) as [$i] (.; $i))'
jq: error (at <stdin>:1): Invalid path expression with result 1
```

Both exit 5. The last pair already agrees (wording included) — `(.a)` parenthesizes the
source, and jq's own path tracking does not reach through a parenthesized subexpression here
either, so this row is not part of the divergence; it's included to show the boundary. All
other rows are the divergence: real jq accepts three of them outright (`["b","c"]`, `["a",0]`,
`[]`) and refuses two with a message naming the destructuring step (`near attempt to access
element ...`); succinctly refuses all five with its generic catch-all message naming the whole
fold's value instead.

The bare-`$var` spelling of the same fold already matches on both sides:

```console
$ echo '{"a":[1,2,3],"b":{"c":5}}' | jq -c 'path(foreach .b as $x (.; .; $x))'
["b"]
$ echo '{"a":[1,2,3],"b":{"c":5}}' | succinctly jq -c 'path(foreach .b as $x (.; .; $x))'
["b"]
```

Real yq is out of scope: its lexer rejects `as PATTERN` destructuring outright (`lexer:
invalid input text`), so none of these queries parse there at all.

## Root cause

`resolve_node_eager`'s `Expr::Reduce`/`Expr::Foreach` dispatch in `src/jq/eval.rs` gates on
`patterns.as_slice()`, admitting only `[Pattern::Var(_)]` — a single, bare-variable
alternative — into `resolve_reduce`/`resolve_foreach`; every other pattern shape (any
destructuring, any `?//` chain) falls through to `resolve_leaf`'s opaque catch-all, which
evaluates the whole fold with the ordinary value evaluator and always marks the result
untracked. The comment directly above that dispatch (and the parallel paragraph "The fold's
own loop-variable destructuring pattern ... is not a behavioural divergence" in
`docs/compliance/jq/limitations.md`) both justify the destructuring branch of that guard by
claiming real jq refuses *every* destructuring fold in path position "even when it matches
cleanly," citing `path(. as $x | reduce ([1]) as [$i] (0; $x))` as the confirming example.
That example's SOURCE is the literal `([1])`, not a register-derived path — and the rows
above show the claim doesn't generalize: once SOURCE is `.b`/`.a` (on the register), jq's
`bind_alternation_matchers` tracks the destructuring step exactly as it does for a plain `. as
{..}` bind, so several of these folds *are* representable as tracked paths and jq accepts
them. The guard (and both comments explaining it) were written against a test matrix that
happened to use only non-register sources, so the "unconditional refusal" premise held for
every case actually tried and was generalized past where it's true.

Sketch of a possible fix, given what's already landing on this branch for #2649: widen the
`Expr::Reduce`/`Expr::Foreach` guard to admit any single alternative (`patterns.len() == 1`,
keeping `?//` chains — `patterns.len() > 1` — on the refuse-only path, per the guard's other,
still-valid justification for that case); after `resolve_fold_source` produces each element's
`FoldSourceValue`, walk the pattern against it the same way a plain `. as PATTERN` bind would
(`register_path` standing in for whatever provenance the bind-time walk uses elsewhere in the
#2649 fix), and seed `FoldRegister` from the walk's resulting `(path, value)` instead of from
the raw source element. `reduce`'s persistent, INIT-seeded register additionally needs its own
check against every step (mirroring `resolve_reduce`'s existing re-entry-boundary check), since
the `[]` row above shows the register can end each iteration back on the ambient input rather
than staying parked on the matched field. `test_reduce_foreach_path_dispatch_falls_back_1440`
pins the current `[Pattern::Var(_)]`-only behaviour and will need updating alongside any
fix, and the `docs/compliance/jq/limitations.md` paragraph's blanket "refuses every such fold"
claim needs correcting to the register-source-dependent behaviour shown above.

## Related

- #2649 — parent issue; its own repro is the plain `. as {..}` bind, one register step. This
  issue is the same register-move story one level up, for a fold's own pattern. The fix
  landing there deliberately does not cover fold patterns — its `resolve_node`/`FoldRegister`
  work only reaches `Expr::TrackedVar` under `[Pattern::Var(_)]`, and the fold dispatch guard
  above is unchanged on this branch as of this issue being filed.
- `FoldRegister`, `resolve_fold_source`, `FoldSourceValue::register_path` (`src/jq/eval.rs`) —
  the existing per-element register-provenance machinery a fix here would extend.
- `test_reduce_foreach_path_dispatch_falls_back_1440` (`src/jq/eval.rs`) — currently pins the
  narrow guard as correct; needs revisiting once this is fixed.
- `docs/compliance/jq/limitations.md`, the "fold's own loop-variable destructuring pattern"
  paragraph — currently overclaims "real jq refuses every such fold in path position"; the
  rows above show that only holds when SOURCE isn't itself register-derived.

ADR-0018: jq mode only. Step 1 of the decision order doesn't apply — real jq's behaviour here
is readable and consistent (it's the same `bind_alternation_matchers` tracking #2649 already
establishes for a plain bind, just reached through a fold's SOURCE), not something rule 4
permits refusing. This is a rule-3 fidelity gap, one level deeper than #2649's.

newhoggy:
## Summary

`reduce`/`foreach`'s own loop-variable pattern (`as PATTERN`) is now tracked through the path
register the same way #2649's plain `. as PATTERN` bind is: when the fold's SOURCE resolves
through the register, destructuring its element moves the register too.

```console
$ echo '{"a":[1,2,3],"b":{"c":5}}' | jq -c 'path(foreach .b as {c:$x} (.; .; $x))'
["b","c"]
$ echo '{"a":[1,2,3],"b":{"c":5}}' | succinctly jq -c 'path(foreach .b as {c:$x} (.; .; $x))'
["b","c"]   # was: refused
```

- `may_bind_navigated` widens to recognise a destructuring `Reduce`/`Foreach` pattern (jq mode
  only), and `resolve_node_sink`'s dispatch guard now admits a single destructuring alternative
  alongside the existing bare-`$var` case.
- `resolve_foreach` reuses #2649's own `walk_pattern`, seeded at the source element's own
  register path (`elem.register_path`, #2031) instead of at `PathPrefix::root()`; the walk's
  final position becomes the active per-step register UPDATE/EXTRACT resolve against, replacing
  the naive whole-element `step_reg` the bare-`$var` arm still uses.
- `resolve_reduce` needed no analogous override: its own accumulator is always checked against
  the fold's *persistent*, INIT-seeded register, regardless of pattern shape (a pre-existing
  #2031 rule) — the walk still runs there, purely to reproduce jq's own refusal for the
  pattern's own step (`path(. as $x | reduce ([1]) as [$i] (0; $x))` refuses in jq even though
  `$i` goes unused, because the pattern's own step is checked against `value_at_path`
  regardless of whether any bound name is read downstream).

Two residual, refuse-only gaps remain (documented in `docs/compliance/jq/limitations.md`):
a destructuring pattern whose SOURCE is not register-derived, and a multi-element array
pattern's reverse-order refusal (both already refuse-only under the old, narrower guard too).

## Verification

- `cargo test`/`cargo clippy --all-targets --all-features -- -D warnings`/`cargo doc` clean.
- `cargo test --no-default-features` (no_std) clean.
- New unit tests in `src/jq/eval.rs`: `test_fold_pattern_destructuring_tracks_register_2676`,
  `test_frame_enter_gate_widens_only_for_destructuring_fold_patterns_2676`, and a tightened
  `test_reduce_foreach_path_dispatch_falls_back_1440`.
- New CLI test `test_fold_destructuring_pattern_moves_path_register_2676` in
  `tests/jq_cli_tests.rs`, every row captured live against jq 1.7.1.
- Extended `scripts/jq-bind-origin-oracle-sweep.sh` (14 new rows) and
  `scripts/jq-bind-origin-fuzz.py` (a `fold_program()` generator reusing the existing pattern
  pools) — 4300+ differential-fuzz cases across three seeds, zero fabricate/mismatch
  attributable to this change (`--baseline` against a pre-#2676 build isolated 11 pre-existing
  cases on `?//`-alternatives folds under a write operator, unrelated to destructuring — filed
  separately as #2858).

Fixes #2676

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -863,14 +863,13 @@ is the revert that established what the other one costs.
    (`path(def f: . as {a:$q} \| $q; f)`, jq `["a"]`) resolves as an opaque leaf — the same `def`
    limitation that table already records.
 
-   Two follow-ups are filed rather than closed here.
-   [#2676](https://github.com/rust-works/succinctly/issues/2676) is the fold's own loop
-   variable: `path(foreach .b as {c:$x} (.; .; $x))` is `["b","c"]` and
-   `path(reduce .b as {c:$x} (.; .))` is `[]` in jq, while `resolve_node`'s `[Pattern::Var(_)]`
-   guard still falls through (see the fold paragraph below).
+   One follow-up is filed rather than closed here.
    [#2678](https://github.com/rust-works/succinctly/issues/2678) is a parse gap that predates
    path mode entirely: a computed-key pattern (`. as {("a"): $q} \| $q`, jq `["a"]`) is
-   `parse error at position 11: expected identifier, found '('` here, exit 3.
+   `parse error at position 11: expected identifier, found '('` here, exit 3. The fold's own
+   loop variable (`reduce`/`foreach`'s own `as PATTERN`, one level up from a plain `as` bind) is
+   closed by [#2676](https://github.com/rust-works/succinctly/issues/2676) — see the fold
+   paragraph below.
 
 2. **`?//`-alternatives folds aren't path-tracked at all** (refuse-only) —
    `path(. as $x \| reduce (1) as $y ?// $z (0; $x))` on `{"a":1}` is `[]` in jq; succinctly
@@ -878,8 +877,11 @@ is the revert that established what the other one costs.
    support for `reduce`/`foreach`'s own `as` clause) landed after `resolve_reduce`/
    `resolve_foreach` were designed, adding a retry-with-rollback matrix
    (`try_reduce_step_alternatives`/`try_foreach_step_alternatives`) neither function threads
-   path-tracking through. `resolve_node`'s dispatch arm admits exactly `[Pattern::Var(_)]`
-   and falls to `resolve_leaf`'s catch-all otherwise.
+   path-tracking through. `resolve_node`'s dispatch arm admits a single pattern alternative
+   only (`patterns.len() == 1`) — a bare `$var` in any mode, or, since
+   [#2676](https://github.com/rust-works/succinctly/issues/2676), a destructuring
+   `Pattern::Object`/`Pattern::Array` in jq mode — and falls to `resolve_leaf`'s catch-all for
+   a `?//` chain of any pattern shape.
 3. **jq's pointer-identity artifacts on `*`/`+` with an empty operand** —
    `path(. as $x \| reduce (1) as $i (0; $x + {}))` on `{"a":1}` is `[]` in jq; succinctly
    refuses (likewise `$x * {}` and `$x + null`). This is not a rule jq implements but an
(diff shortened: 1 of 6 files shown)


newhoggy:
## Code review synthesis (8-agent, high effort)

All 8 finder agents completed. Summary:

**One real, high-severity finding** — a `foreach` EXTRACT fabrication bug (accept-where-jq-
refuses, corrupts a write). Independently re-verified: reproduces with a **bare `$var` pattern,
no destructuring**, on a pre-#2676 `main` build (`path(foreach .a as $v0 (.; $v0; (.zzz | $v0)))`
on `{"a":{"b":1}}` — jq refuses, `main` answers `["a"]`). The root cause lives entirely in
`FoldRegister`'s pre-existing UPDATE→EXTRACT chain / `resolve_seq`'s #2046 carried-register
mechanism, which this PR does not touch — #2676 only gives a destructured loop variable a
tracked marker for the first time, inheriting the same pre-existing defect the bare-`$var`
splice already had. Filed as **#2860** (tagged `bug`) with full root-cause analysis rather than
fixed here, since a correct fix needs its own dedicated trace through `resolve_seq_stage`'s
missing-key/null-absorption handling — documented in `docs/compliance/jq/limitations.md`.

**One real coverage gap**, fixed: no pinned test had a register-derived source where the walk
succeeds *and* UPDATE/EXTRACT reads the bound var directly. Added rows for both
`resolve_foreach` (accepts) and `resolve_reduce` (always refuses there, confirmed live — the
fold's accumulator is checked against the persistent register regardless of pattern shape).

**Four independent angles flagged the same two code-structure issues** (no correctness impact),
both fixed:
- The dispatch guard was spelled out twice (once per `Reduce`/`Foreach` arm) — extracted to
  `fold_pattern_admitted::<S>(patterns)`, now also directly unit-tested (the original
  "yq mode" test was vacuous: comparing two `eval()` results can't observe the gate, since
  `path(foreach ...)` collapses to `None` under `YqSemantics` for unrelated reasons regardless
  of pattern shape).
- The `PatternRegister` seed construction was copy-pasted between `resolve_reduce` and
  `resolve_foreach` — extracted to `fold_pattern_seed(elem, reg)`.

**One efficiency finding acted on**: `resolve_reduce` walked the pattern twice per step (once,
tracked, discarded; once via `extract_pattern_bindings` for the actual substitution). Now reuses
`walk_pattern`'s own bindings (origin cleared, since `reduce` never seeds a per-step register)
— eliminates a full pattern-tree traversal and value clone per step, behaviorally unchanged.

**No conventions violations, no removed-behavior regressions, no cross-file/cross-function
correctness bugs** found by the remaining angles.

Re-verified after applying fixes: full test suite, clippy (`-D warnings`), `cargo doc`,
`cargo test --no-default-features` (no_std), the extended oracle sweep (134 rows, 0
fabricate/mismatch/new-refuse), and a fresh 3000-case differential fuzz run with `--baseline`
attribution (0 fabricate/mismatch attributable to this branch).

Issue #2872, opened 2026-09-12
jq: computed-key destructuring pattern fan-out — path position and reduce/foreach still refuse cleanly (#2677 follow-up)


newhoggy:
## Context

#2677 adds support for a computed/interpolated key in an object destructuring
pattern (`{(EXPR): $var}`, `{"\(EXPR)": $var}`). That PR covers:

- Full parse-acceptance of the syntax everywhere a pattern can appear.
- Correct evaluation of a **single-output** computed key in both value
  position and path position (`path()`, `del()`, assignment targets `|=`
  etc.), including nested computed keys.
- Correct evaluation of a **multi-output** (generator) or **zero-output**
  computed key, but **only in value position** (`. as PATTERN | body`, with
  or without `?//` alternatives) — via `each_pattern_alternatives`
  (`src/jq/eval.rs`) and `each_pattern_alternatives_generic`
  (`src/jq/eval_generic.rs`).

## Remaining scope

The following call sites still refuse cleanly (a dedicated, correct error —
never a wrong answer) when a computed key yields other than exactly one
value, and would need the same generator/cartesian-fold treatment
`extract_pattern_bindings` already implements internally:

1. **Path position** — `walk_pattern` / `resolve_as_pattern` /
   `try_pattern_alternatives` (`src/jq/eval.rs`). A multi-output key inside
   `path(. as {("a","b"):$q} | $q)` or as an assignment target currently
   refuses rather than fanning out.
2. **`reduce`/`foreach`'s own pattern** — `eval_reduce_with_values` /
   `substitute_foreach_steps` (`src/jq/eval.rs`) build a
   `Vec<Result<Expr, EvalError>>` matrix via `extract_single_pattern_binding`
   *before* the fold loop runs, so a multi-output key in
   `reduce .[] as {("a","b"):$q} (...)` also refuses rather than fanning out
   across the fold.
3. **Residual walker gaps**, confirmed safe-direction-only (a missed static
   optimization, or a compile-time-shaped check surfacing one stage later as
   a runtime error — never a wrong answer): `any_subexpr`,
   `node_reads_ambient` (`src/jq/walk.rs`), and `resolve::check`
   (`src/jq/resolve.rs`) still treat a pattern as holding only destructuring
   names, not an arbitrary `Expr` inside `ObjectKey::Expr`. Worth a full audit
   pass once the above are closed, to confirm none of them can produce a
   wrong answer (as opposed to a missed optimization) for a computed key.

[...]

PR #2873, opened 2026-09-12
fix(jq): accept and evaluate a computed key in a destructuring pattern (#2677)

### LINKED ISSUE #2677: jq: object destructuring pattern with a computed key ({("a"):$q}, {(.k):$q}) is a parse error — jq accepts it
**Severity: Low** (exit-3 parse rejection only — no accept-where-jq-refuses, no wrong
document written; #2649's own path-register question doesn't even arise here since these
queries never reach evaluation)

## Summary

Real jq's object destructuring pattern accepts a parenthesized, computed key —
`{(EXPR): $var}` — the same key-expression shape a plain object *construction* already
allows, evaluated as an untracked subexpression before the (tracked) `INDEX` step it feeds. A
computed key is also a generator: `{("a","b"):$q}` produces one binding per key it yields.
`succinctly jq`'s pattern parser only knows `$name`, a bare identifier or string-literal key,
and the `{$name}`/`{$name: P}` shorthands — a parenthesized key is a hard parse error, so every
query below fails to compile at all (exit 3) rather than misbehaving at evaluation time.

## Repro (jq 1.7.1 vs succinctly jq)

```console
$ echo '{"a":[1,2,3]}' | jq -c '. as {("a"):$q} | $q'
[1,2,3]
$ echo '{"a":[1,2,3]}' | succinctly jq -c '. as {("a"):$q} | $q'
jq: compile error: parse error at position 6: expected identifier, found '('

$ echo '{"a":1,"b":2}' | jq -c '. as {("a","b"):$q} | $q'
1
2
$ echo '{"a":1,"b":2}' | succinctly jq -c '. as {("a","b"):$q} | $q'
jq: compile error: parse error at position 6: expected identifier, found '('

$ echo '{"a":1,"b":2}' | jq -c 'path(. as {("a","b"):$q} | $q)'
["a"]
["b"]
$ echo '{"a":1,"b":2}' | succinctly jq -c 'path(. as {("a","b"):$q} | $q)'
jq: compile error: parse error at position 11: expected identifier, found '('

$ echo '{"a":[1,2,3],"k":"a"}' | jq -c '. as {"\(.k)":$q} | $q'
[1,2,3]
$ echo '{"a":[1,2,3],"k":"a"}' | succinctly jq -c '. as {"\(.k)":$q} | $q'
jq: compile error: parse error at position 8: invalid escape sequence '\('
```

All four rows: jq exit 0, succinctly exit 3. A plain literal-string key already matches
(control case, both sides `[1,2,3]`):

```console
$ echo '{"a":[1,2,3]}' | jq -c '. as {"a":$q} | $q'
[1,2,3]
$ echo '{"a":[1,2,3]}' | succinctly jq -c '. as {"a":$q} | $q'
[1,2,3]
```

jq also accepts an interpolated string as a key (`{"\(.k)": $q}`, third block above) — that's
the same computed-key gap, just spelled with string interpolation instead of a bare
parenthesized expression; succinctly's parser rejects it earlier and differently (`invalid
escape sequence '\('`, since it never gets past treating `"..."` as a plain string literal to
notice the `\(` is interpolation at all). One piece of related trivia, *not* part of this
issue: succinctly accepts `. as {$__loc__} | $__loc__` (`{"file":"<stdin>","line":1}`) where
real jq gives a syntax error (`unexpected $__loc__`) — confirmed live, but its own, much
smaller shape; not worth a separate issue or folding into this one's fix.

Real yq is out of scope: its lexer rejects `as PATTERN` destructuring outright, so none of
these queries parse there either.

## Root cause

`parse_pattern_inner` (`src/jq/parser.rs`) branches on `self.peek()` inside an object pattern:
`$` for the `{$name}`/`{$name: P}` shorthands, else a hard choice between `"` (string-literal
key, via `parse_string_literal`) and everything else (`parse_ident`). There is no branch for
`(`, so `{(EXPR):...}` hits the `parse_ident` arm and fails immediately with "expected
identifier". `PatternEntry.key` is a plain `String`, which is itself too narrow to hold a
computed key — real jq's grammar (`gen_object_matcher(gen_const(name), ...)` for a literal key
vs. the general `Exp` production for a parenthesized one) treats the two as structurally
different productions, and `parse_string_literal`'s current use here doesn't handle `\(...)`
interpolation at all (it's called expecting a plain literal, not a `StringPart`-bearing
interpolated string).

A fix needs a new key shape, e.g. `PatternKey { Literal(String), Expr(Box<Expr>) }` in place
of `PatternEntry.key: String`, parsed as `(` EXPR `)` or an interpolated string literal
(reusing whatever produces `StringPart` elsewhere in the parser), and evaluated the way real
jq's `gen_object_matcher` does: as an *untracked* subexpression (a generator, for the
multi-key case) whose yielded key value then feeds the existing tracked `INDEX` step — the key
expression itself is never part of the path, only what it evaluates to. Every existing
`Pattern`/`PatternEntry` consumer would need a new arm for the computed-key case:
`extract_pattern_bindings`, `collect_pattern_var_names`, `pattern_binds_var` (all
`src/jq/eval.rs`), and the `walk_pattern` helper landing alongside #2649's fix. This is a
parser/AST change, not a path-tracking change — orthogonal to #2649 and to the fold-pattern
gap in the sibling issue, which is why it's filed separately.

## Related

- #2649 — parent issue; unaffected by this one and vice versa. #2649's fix works entirely
  within the existing `Pattern`/`PatternEntry` shape and does not touch the parser's key
  grammar.
- The sibling fold-destructuring issue spun out alongside this one — also blocked on
  `Pattern`/`PatternEntry`, but on the *evaluator's* register-tracking side, not the parser.
- `docs/reference/jq-language.md`, which documents supported pattern syntax, may need a line
  added once this lands.

ADR-0018: jq mode only. Step 1 of the decision order doesn't apply — real jq's grammar here is
a real, documented feature (`jq` manual's "Destructuring Alternative Operator" section covers
plain patterns; computed keys fall out of the same `Exp` production object construction
already uses), not something rule 4 permits refusing. This is a rule-3 fidelity gap: a parser
feature gap rather than a path-tracking one.

newhoggy:
## Summary

Adds support for a computed or interpolated key in an object destructuring
pattern — `{(EXPR): $var}`, `{"\(EXPR)": $var}` — which real jq accepts and
succinctly previously rejected as a parse error.

- Parsing: `parse_pattern_inner` now accepts `(EXPR)` and `"\(EXPR)"` object
  keys in a pattern, alongside the existing literal/`$name`-shorthand forms.
  `PatternEntry.key` becomes `ObjectKey` (already used by object
  construction) instead of a bare `String`.
- Evaluation, single-output key (the common case): correct in both value
  position and path position (`path()`, `del()`, assignment targets like
  `|=`), including nested computed keys, matching jq's exact `INDEX`
  bytecode semantics — a non-string key value on a valid object gives
  `Cannot index <type> with <type>`; any key on an invalid container gives
  `Cannot index <type> with string "..."`.
- Evaluation, multi-output (generator) or zero-output key, **value position
  only** (`. as PATTERN | body`, with or without `?//` alternatives): fans
  out into one full pattern-match per key output, matching jq's own
  left-to-right, outer-to-inner nested-fork order, including the two
  trickiest `?//`-interaction edge cases — a body error partway through a
  fan-out retries the next alternative while keeping already-emitted output,
  and a zero-output key still wins a `?//` chain without falling through.
- Fixed an independently-discovered bug while working on this: `map_subexprs`
  cloned a pattern's own computed key verbatim instead of descending into
  it, so def-call resolution and variable substitution never saw a
  `FuncCall`/`Var` reference embedded in a computed key (`def f: "a"; . as
  {(f):$q} | $q` wrongly said "undefined function: f/0"). Fixed with a new
  `map_pattern_subexprs` helper wired into `map_subexprs`'s `Reduce`/
  `Foreach`/`AsPattern` arms.
- `docs/compliance/jq/limitations.md` and `docs/reference/jq-language.md`
  updated to describe exactly what's supported and what's deliberately
  scoped out.

## Scope

Path-position fan-out for a multi-/zero-output computed key, and
`reduce`/`foreach`'s own pattern fan-out, still refuse cleanly (never a wrong
answer) rather than fanning out — filed as a follow-up, #2872.

Fixes #2677

## Test plan
- [x] `cargo test --features cli --lib jq::` (2295 passed)
- [x] `cargo test --features cli --test jq_cli_tests` (1680 passed)
- [x] `cargo test --features cli --test yq_cli_tests` (1869 passed)
- [x] Oracle sweep (`scripts/jq-bind-origin-oracle-sweep.sh`) and differential
      fuzz (`scripts/jq-bind-origin-fuzz.py`, multiple 2000-3000 case runs
      with `--baseline` attribution) — zero fabricate/mismatch/new-refuse

### DIFF
--- a/docs/compliance/jq/limitations.md
+++ b/docs/compliance/jq/limitations.md
@@ -976,23 +976,32 @@ a plain `String` to `ObjectKey` (the same type object construction already used)
 `extract_pattern_bindings` (value mode)/`walk_pattern` (path mode) resolve the key expression
 against the pattern's own current node, using the same "Cannot index `<type>` with `<type>`"
 wording jq's own `INDEX` bytecode gives for any key, computed or literal — confirmed live in
-both positions, including nested computed keys and through `|=`/`del()`. **Scoped to a key
-expression that yields exactly one string** — the common case (`{(.k): $q}`, a string
-interpolation, which by construction can only ever yield one value). A key expression that is
-itself a multi-output *generator* (`{("a","b"): $q}` — real jq fans out one full pattern-match,
-and one full run of the surrounding body, per key it yields: `{"a":1,"b":2} \| . as
-{("a","b"):$q} \| $q` is `1` then `2`) refuses clearly instead — `extract_pattern_bindings`
-itself models the full cartesian-fold generator semantics real jq has, but every one of its
-eight call sites (across both evaluators, plus `reduce`/`foreach`'s own pre-loop substitution-
-matrix builders) collapses back to a single result via `extract_single_pattern_binding`, since
-fanning out correctly needs each call site's own `?//`-alternative-retry/fold-matrix machinery
-threaded through, not just the core function — not yet done. The same refusal covers a key
-expression producing **zero** outputs (`{(empty): $q}`, where real jq legitimately binds
-nothing and the body never runs, `[. as {(empty):$q} \| $q]` is `[]`) and one interrupted by
-`break`/`halt`, rather than risking a silently wrong answer at any of the eight sites
-individually. `test_pattern_computed_key_evaluates_2677`/
-`test_pattern_computed_key_multi_output_refuses_cleanly_2677` (`tests/jq_cli_tests.rs`) pin both
-halves.
+both positions, including nested computed keys and through `|=`/`del()`. A key expression that
+is itself a multi-output *generator* (`{("a","b"): $q}` — real jq fans out one full pattern-
+match, and one full run of the surrounding body, per key it yields: `{"a":1,"b":2} \| . as
+{("a","b"):$q} \| $q` is `1` then `2`) or a **zero**-output one (`{(empty): $q}`, where real jq
+legitimately binds nothing and the body never runs, `[. as {(empty):$q} \| $q]` is `[]`) now
+**fans out correctly in value position**: `each_pattern_alternatives`/
+`each_pattern_alternatives_generic` (the `?//`-alternative loops both evaluators route a bare,
+non-`?//` pattern through too) run `body` once per binding-set `extract_pattern_bindings`
+yields, applying the pre-#2677 match-failure retry rule only to the key generator's own
+trailing error/break/halt once every binding-set's body has run — including the two hardest
+interaction cases: a body error partway through a fan-out abandons the *remaining* key outputs
+and retries the next `?//` alternative, keeping whatever already ran (`[. as {("a","b"):$q} ?//
+$z \| if $q==2 then error("boom") else $q end]` on `{"a":1,"b":2}` is `[1,null]`), and a
+zero-output key still *wins* inside a `?//` chain (no fallthrough) even though it contributes
+nothing. `test_pattern_computed_key_fans_out_in_value_position_2677` (`tests/jq_cli_tests.rs`)
+pins both.
+
+**Path position does not fan out yet** — `walk_pattern`/`resolve_as_pattern`/
+`try_pattern_alternatives` still collapse to a single result via
+`extract_single_pattern_binding` and refuse a multi- or zero-output computed key cleanly
+(`test_pattern_computed_key_multi_output_refuses_in_path_position_2677`), the same reasoning
+that motivated `extract_single_pattern_binding` in the first place: fanning out correctly needs
+each call site's own retry/register machinery threaded through, not just the core function.
+`reduce`/`foreach`'s own pattern (`eval_reduce_with_values`/`substitute_foreach_steps`'s
+pre-loop substitution-matrix builders) is unaffected either way — a multi-output computed key
+there still refuses in both value and path position.
 
 A related, narrower gap the fix surfaced and closed along the way: `map_subexprs`
 (`src/jq/walk.rs`) — the shared tree-rewrite primitive `install_def_calls`/`bind_def` and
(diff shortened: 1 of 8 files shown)


newhoggy:
Code review found four real bugs, all fixed and pinned with new tests:

1. A computed key resolving to a number against an array target wrongly
   required a string key ("Cannot index array with number" where real jq
   indexes the array). Fixed by routing through `index_one_owned` (value
   mode) and widening `PatternStep::Index` to `i64` (path mode, so a
   negative/out-of-range key resolves correctly, including matching jq's
   own raw-key path recording — `path(.arr as {(-1):$q}|$q)` on an array
   now gives `[-1]`, matching `path(.arr[-1])`).
2. `each_pattern_alternatives`/`_generic` dropped a pending `Halt`/uncatchable
   `Error` from the key generator's own trailing control whenever an earlier
   binding-set's body already triggered a `?//` retry. Fixed by checking
   that control first, with priority over any retry decision.
3. `substitute_func_param_impl` and `substitute_var_impl`'s shadow-guarded
   arms didn't descend into a pattern's computed key — the same
   `map_subexprs` bug class this PR itself fixed, found in two more
   separate hand-written traversals. Fixed with the same `map_pattern_subexprs`
   helper.
4. Added a fast path avoiding an O(entries) clone-per-entry in the common
   (no computed key) case, per the review's efficiency finding.

Follow-up #2872 updated to reflect the narrower remaining scope (path-position
and reduce/foreach fan-out unaffected; added a new item for `walk_pattern`'s
own Halt/Break-downgrades-to-catchable gap, out of scope for this pass).

Re-ran the full verification suite (unit + CLI test suites, clippy,
`cargo doc`, no_std, release build) plus the oracle sweep and a 2000-case
differential fuzz run with `--baseline` attribution against pre-#2677 `main`
— zero new fabricate/mismatch findings (a handful of pre-existing,
already-tracked `foreach`/`?//` fan-out bugs unrelated to computed keys
showed up identically on both, confirming no regression).

Issue #2807, opened 2026-09-11
jq: widening shadowable_defs with parameter names may increase SHADOW_RETRY_BUDGET exhaustion risk


newhoggy:
**Severity: Low (unconfirmed)**

## Summary

Filed from PR #2805's code review (Angle C / cross-file tracer), which flagged this as a plausible but unconfirmed risk — no live repro was constructed, so this is a "worth a second look" investigation, not a known bug.

`collect_def_names` (`src/jq/parser.rs`) feeds `Parser::shadowable_defs`, which gates a small number of parse choke points (`parse_shadowable_special_form`, `zero_arity_or_wrong_arity_call`, `try_parse_builtin`'s call site) into a "wrap as shadow candidate, let resolve.rs decide" path. When the shadow candidate's own dedicated parser fails outright (not merely wrong-arity), `retry_shadow_candidate_as_generic_call` consumes one unit from a single process-wide `SHADOW_RETRY_BUDGET` (64, `src/jq/parser.rs`) shared across the *entire program's* parse.

#2723 widened `collect_def_names` to also collect every `def`'s *parameter* names (previously only the `def`'s own name was collected). Parameter names are plausibly far more likely to incidentally collide with common builtin names (`first`, `last`, `range`, `has`, `error`, `contains`, `ltrimstr`, ...) than deliberately-chosen `def` names, since programmers commonly reuse short generic words as parameter names. If a large real-world program has many such incidental collisions, and enough of them hit a shadow candidate's dedicated parser failing outright (not just wrong-arity), the shared 64-unit budget could exhaust earlier than before #2723, causing a *later*, otherwise-legitimate shadowing call whose own dedicated parse also fails to get the raw `original_err` propagated instead of correctly falling back to generic-call + `resolve.rs` resolution — i.e. a different/wrong compile-error message for that later call, not a wrong answer or a crash.

## Why unconfirmed

The reviewing agent could not construct a minimal live repro in the time available — most "wrong shape" shadow-candidate cases route through the non-budget-consuming rewind paths (#2110/#2237/#2686 already rewind wrong-arity-but-well-formed cases without touching the budget), so triggering the *budget-consuming* path specifically, 64+ times in one program, needs a more deliberate construction.

## Suggested next step

[...]

Issue #2686, opened 2026-09-10
jq: nested wrong-arity `until`/`limit` parses in O(2^depth) — 18 s at depth 24 where jq takes 20 ms


newhoggy:
## What's wrong

Found while reviewing #2395's branch. A **2-arity special form called with 3 arguments**
(`until(a;b;c)`, `limit(a;b;c)`) costs `O(2^depth)` **to parse** when nested. Real jq
rejects the same text in constant time.

Measured on this machine (Apple M-series, release build, `succinctly` at `23337c17f`):

| depth | `until(...;1;2;3)` succinctly | jq 1.7.1 |
|-------|-------------------------------|----------|
| 8     | 20 ms                         | 19 ms    |
| 12    | 28 ms                         | 20 ms    |
| 16    | 89 ms                         | 18 ms    |
| 20    | 1 172 ms                      | 19 ms    |
| 24    | 18 358 ms                     | 20 ms    |
| 30    | > 25 s (killed)               | 19 ms    |

`limit` is identical to the milliseconds. Growth is ~1.9x per level — 2^depth.

Repro:

```console
$ python3 -c "print('until('*24 + '1;2;3' + ')'*24)" > /tmp/f.jq
$ time jq            -nc -f /tmp/f.jq   # 0.02s, "jq: error: until/1 is not defined"
$ time succinctly jq -nc -f /tmp/f.jq   # 18s,   same message
```

## It is the parser, not the resolver

Appending a hard lexical error so the program can never reach name resolution keeps the
cost:

```console
$ f="$(python3 -c "print('until('*22+'1;2;3'+')'*22)") | @@@"
$ time succinctly jq -nc "$f"
jq: compile error: parse error at position 163: unknown format '@'
# 2.3 s — a pure parse error, nothing resolved
```

## Scope: only the 2-arity forms, and only at wrong arity

Everything else I probed is flat to depth 24 (~20 ms, indistinguishable from jq):

| shape                            | depth 24 |
|----------------------------------|----------|
| `until(a;b;c)` (until/2 + 1 arg) | 18 358 ms |
| `limit(a;b;c)` (limit/2 + 1 arg) | 17 990 ms |
| `first(a;b)`   (first/1 + 1 arg) | 19 ms     |
| `range(a;b;c;d)`                 | 21 ms     |
| `nosuch(a;b;c)` (generic call)   | 21 ms     |
| `until(.;1)` (correct arity)     | 21 ms     |

[...]

PR #2879, opened 2026-09-12
perf(jq): stop parse_required_single_arg's O(2^depth) wrong-arity reparse

### LINKED ISSUE #2749: jq: parse_required_single_arg keeps #2686's O(2^depth) wrong-arity re-parse for ~30 builtins
**Severity: Medium** (parser-level denial of service from a tiny filter string)

## Summary

#2686 removed the `O(2^depth)` wrong-arity re-parse from the eight dedicated
special-form parsers (`until`, `while`, `limit`, `first`, `last`, `repeat`, `range`,
`error`). The identical shape survives behind `parse_required_single_arg`
(`src/jq/parser.rs`), so the ~30 single-argument builtins routed through
`try_parse_builtin` still have it.

## Repro

Measured on a `--features cli` build, both at #2686's merge base and on its branch —
**identical**, confirming it is untouched by that fix:

| filter | depth | merge base | #2686 branch |
|---|---:|---:|---:|
| `"has(" * d + "1" + ";2)" * d` | 14 | 152 ms | 157 ms |
| | 16 | 614 ms | 604 ms |
| | 18 | 2424 ms | 2420 ms |
| `"select(" * d + "1" + ";2)" * d` | 18 | 2509 ms | 2532 ms |
| `"until(" * d + "1" + ";2;3)" * d` (fixed by #2686, for contrast) | 18 | 3345 ms | **6 ms** |

~3.9x per two levels (~2x per level) from a filter string well under 200 bytes. A
review of #2686 measured it further out: 41.5 s at depth 22 and **158.7 s at depth 24**
from a 168-byte filter.

`env` is already flat — it has its own hand-rolled rewind logic (the same one
`strenv` is noted as needing in #2573), so it is not affected.

## Root cause

Same mechanism #2686 fixed, different protocol.

`parse_required_single_arg` parses its argument, and on a wrong-arity mismatch
`expect_or_none(')')` rewinds to `start_pos` and returns `Ok(None)` — "not a match".
The caller then re-runs `parse_func_call_or_error` over the same text, which descends
into the next nesting level, which does the same: `T(d) = 2·T(d-1)`.

#2686's fix does not apply directly. It works by handing the already-parsed arguments
to `wrong_arity_call_from_parsed`, which resolves the call **in place**;
`parse_required_single_arg` instead returns `Result<Option<_>>` and hands resolution
back to the caller's own shadow-candidate fallback. Closing this means either
threading the parsed argument out through that `Option` protocol, or giving the helper
the same in-place resolution the special forms now use.

## Why this is separate from #2686

Found by review of #2686 and confirmed pre-existing there (identical timings on both
binaries, table above). #2686's issue, repro and fix are all about the eight dedicated
special-form parsers; this is a different helper with a different return protocol and
~30 call sites behind it, so it was scoped out rather than folded in.
`test_wrong_arity_deep_nesting_does_not_blow_up_2686`'s doc comment records the
boundary explicitly so the wider surface is not assumed fixed.

## Suggested fix direction

Give `parse_required_single_arg` an in-place resolution like
`wrong_arity_call_from_parsed`, or thread the already-parsed argument out so the
caller's fallback does not re-parse it. Two things to check that #2686 did not have to:

1. **The shadow-candidate fallback.** `Ok(None)` currently means "let the caller try a
   shadowed `def`". Any in-place resolution must not bypass that, or a legitimate
   `def has(a;b): ...` stops shadowing.
2. **`SHADOW_RETRY_BUDGET`.** #2036 added it precisely because this fallback re-parses;
   removing the re-parse may make the budget dead code, or may change which programs
   exhaust it. Check both directions.

Verify with the same 1710-pair merge-base differential #2686 used (all builtins ×
argument counts × shadowed spellings × both modes), plus a depth-40 timing guard
beside `test_wrong_arity_deep_nesting_does_not_blow_up_2686`.

Note #2573 is converting call sites in this same helper's neighbourhood; whoever takes
this should check for overlap first.

newhoggy:
## Summary
- `parse_required_single_arg`'s trailing-`)` checkpoint rewound and discarded an already-parsed argument on a wrong-arity call (e.g. `has(1;2)`), so the caller's fallback re-parsed the same argument text from scratch. Nested (`has(has(...1...;2);2);2)`), each level re-parses everything below it: `O(2^depth)`. `"has(".repeat(18) + "1" + ";2)".repeat(18)` took ~315ms before this fix; depth 40 now takes ~12ms.
- This is the identical shape #2686 fixed for the eight dedicated special-form parsers (`until`/`while`/`limit`/...), surviving at this site only because its `Result<Option<Expr>, ParseError>` contract had no slot for an already-parsed argument.
- Fix: generalize #2807's `Builtin`-only `builtin_wrong_arity_or_expect` into `builtin_wrong_arity_or_expect<T>` (the body never touched `T`, it only ever parked the resolved call in the existing `wrong_arity_call` field), so this checkpoint can park an `Expr` there too instead of rewinding. Affects all ~30 single-argument builtins routed through `try_parse_builtin` (`has`, `select`, `map`, `getpath`, `join`, `bsearch`, `ltrimstr`, `startswith`, ...).
- Output is byte-identical to before at every arity/spelling — confirmed via a 666-way differential sweep (37 names × 9 shapes × both modes) against a binary built from the merge base — only the timing changes.

Fixes #2749

## Test plan
- [x] New timing-guard test (`test_wrong_arity_single_arg_deep_nesting_does_not_blow_up_2749`): 8 representative names at depth 40, must reject in well under 15s.
- [x] New parity tests (`test_wrong_arity_single_arg_resolution_is_unchanged_2749` jq-mode, `test_wrong_arity_single_arg_keeps_yq_parse_error_2749` yq-mode) pinning name/arity resolution, postfix tails, shadowing `def`s, and yq's own raw-error carve-out — every expectation captured live from `/usr/bin/jq` 1.7.1.
- [x] Existing `#2686`/`#2807` regression tests still pass unmodified.
- [x] `cargo test --features cli` (full suite, all targets): 0 failures.

### DIFF
--- a/src/jq/parser.rs
+++ b/src/jq/parser.rs
@@ -3281,13 +3281,20 @@ impl<'a> Parser<'a> {
         self.expect(expected).map(|()| unreachable!())
     }
 
-    /// [`Self::wrong_arity_or_expect`] for a [`Self::try_parse_builtin`]
-    /// arm, whose answer is a `Builtin` rather than an `Expr` (#2807).
+    /// [`Self::wrong_arity_or_expect`] for a checkpoint whose caller's own
+    /// return type has no slot for an already-resolved `Expr` -- originally
+    /// every [`Self::try_parse_builtin`] arm answering a `Builtin` (#2807),
+    /// now also [`Self::parse_required_single_arg`] answering an `Expr`
+    /// directly (#2749). Generic over that answer type `T` purely so both
+    /// callers can return their own `Ok(None)`; the body never touches `T`
+    /// itself -- every real value travels through [`Self::wrong_arity_call`]
+    /// instead, exactly as it always did.
     ///
-    /// The 27 multi-argument and optional-arity arms in that function
-    /// checked their argument boundaries with a bare `self.expect(';')?` /
-    /// `self.expect(')')?`, so a wrong-arity call under a shadowing `def`
-    /// reached `parse_primary_inner`'s `Err` arm and was recovered only by
+    /// #2807's original motivation: the 27 multi-argument and
+    /// optional-arity arms in `try_parse_builtin` checked their argument
+    /// boundaries with a bare `self.expect(';')?` / `self.expect(')')?`, so
+    /// a wrong-arity call under a shadowing `def` reached
+    /// `parse_primary_inner`'s `Err` arm and was recovered only by
     /// [`Self::retry_shadow_candidate_as_generic_call`]'s re-parse -- which
     /// re-scans the whole argument text from the keyword (`O(2^depth)` when
     /// nested, #2686's shape) and spends one unit of the process-wide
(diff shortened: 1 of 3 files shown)


newhoggy:
Ran the mandatory 4-angle code review. No correctness bugs — all four angles independently confirmed the output is byte-identical to the pre-fix binary across large differential sweeps (666+ programs, both modes).

Two stale doc-comment findings fixed (the `wrong_arity_call` field doc and `rewind_to_wrong_arity_call`'s doc both still described `parse_required_single_arg` in its pre-fix, rewind-only shape).

One internal-mechanism note judged **no change needed**: a malformed argument separator under a shadowing `def` (e.g. `has(1 2)`) now reaches `parse_primary_inner`'s `Err(e)` arm instead of `Ok(None)`, spending one `SHADOW_RETRY_BUDGET` unit where the old rewind never did. Confirmed this can never be externally observable: a malformed separator is a hard parse error that terminates the whole parse immediately (jq's parser has no fault-tolerant continuation), so there is no scenario where this extra spend combines with other calls' budget usage before the program fails anyway.

One finding was real but pre-existing and out of scope for this performance-only fix: `succinctly yq` mode rejects `;` inside a single-required-argument builtin call (`has(1;2)`) where real yq treats it as a BLOCK expression. Confirmed byte-identical before/after this PR. Filed as #2881.

Issue #2883, opened 2026-09-12
yq: JSON flow-mapping token-pairing grammar (#2777) needs route unification for --inplace/--slurp/--eval-all


newhoggy:
## Context

#2777 (fixed in this issue's own PR) adds real yq's token-pairing grammar for
JSON-sourced flow mappings (`-p json`'s `{...}`) on the "plain" route
(`YamlIndex`-backed parsing, used by ordinary evaluation without
`--inplace`/`--slurp`/`--eval-all`). Two things were deliberately scoped out
of that fix, both documented in `docs/compliance/yq/limitations.md`'s
"Closed by #2777" section:

## 1. Route unification (DOM bridge still uses the old strict grammar)

`--inplace`/`--slurp`/`--eval-all` materialize JSON input through a separate
code path — `yq_runner::parse_input`'s `InputFormat::Json` arm, which uses
`JsonIndex` + `to_owned_canonicalizing_numbers_at_depth` rather than
`YamlIndex::build_json_sourced`. That route still applies #1975's own
stricter-than-real-yq JSON grammar, so it does not get #2777's leniency:

```console
$ printf '{"a":1 "b":2}' > /tmp/f.json
$ yq -p json -o json -I0 --inplace '.' /tmp/f.json && cat /tmp/f.json
{"a":1,"b":2}
$ succinctly yq -p json -o json -I0 --inplace '.' /tmp/f.json
Error: Invalid JSON text: expected ',' or '}', found '"'
```

Fixing this means making `yq_runner::parse_input`'s `InputFormat::Json` arm
build through `YamlIndex::build_json_sourced` too (instead of `JsonIndex`),
which would also require carrying `JsonIndex`'s own number-canonicalization
pass over to the YAML-cursor-backed walk. This is a separate, larger change
that touches the destructive `--inplace` write path, so it needs its own
careful test coverage (proving nothing is written on a rejected file, and
that `--inplace` now rewrites what real yq rewrites) rather than being
folded into #2777's own fix.

## 2. "Any bracket ends the loop" is not reproduced

Real yq's token scanner treats *every* bracket character (`{`, `}`, `[`,
`]`) as a `Delim` that ends the *current* object's key-reading loop the
instant one is seen — not just its own matching closing brace:

```console
$ printf '{"a":1 ]}' | yq -p json -o json -I0 '.'
{"a":1}
$ printf '{[1]:2}' | yq -p json -o json -I0 '.'
{}
```

[...]

Issue #1975, opened 2026-08-30
yq --input-format json --slurp loses #1194/#1677 malformed-JSON detection entirely


newhoggy:
## Summary

`succinctly yq --input-format json --slurp` silently accepts structurally malformed JSON (missing `,`/`:` delimiters, #1677) that every other route into the evaluator correctly rejects. This looks unrelated to the `DistinctKeyCursors::is_malformed()` gaps #1956/PR #1971 just fixed -- those fixes make **no observable difference** to this specific route, which points at a different root cause upstream of `DistinctKeyCursors` entirely.

## Repro

```
$ printf '{"a" 1, "b": 2}' | succinctly jq -c 'keys_unsorted'
jq: error (at <stdin>:0): Invalid JSON text: expected ':', found '1'
(exit 5 -- correct)

$ printf '{"a" 1, "b": 2}' | succinctly jq --slurp -c '.[0] | keys_unsorted'
jq: error (at <stdin>:0): Invalid JSON text: expected ':', found '1'
(exit 5 -- correct; --slurp alone is fine)

$ printf '{"a" 1, "b": 2}' | succinctly yq --input-format json 'keys_unsorted'
Error: YAML parse error: unexpected character '1' at offset 5: expected ':', ',' or '}' after key in flow mapping
(exit 1 -- correct, but via the *YAML* flow-mapping parser: --input-format json without
--slurp appears to route the single document through the YAML parser rather than JSON's
own semi-index, which is strict enough to catch this upstream of DistinctKeyCursors)

$ printf '{"a" 1, "b": 2}' | succinctly yq --input-format json --slurp -o=json '.[0] | keys_unsorted'
[
  "a",
  "b"
]
(exit 0 -- WRONG: silently succeeds)

$ printf '{"a" 1, "b": 2}' | succinctly yq --input-format json --slurp '.[0] | keys_unsorted'
- a
- b
(exit 0 -- WRONG: same document, YAML output)
```

Only the `--input-format json --slurp` combination is affected. Every other combination tested (`jq` native, `jq --slurp`, `yq --input-format json` without `--slurp`) correctly rejects the same document.

## Investigation notes (from #1956/PR #1971's review)

[...]

PR #2884, opened 2026-09-12
fix(yq): reproduce real yq's token-pairing grammar for JSON flow mappings (#2777)

### LINKED ISSUE #2777: yq: -p json flow-mapping grammar diverges from real yq in both directions (one silent wrong value)

newhoggy:
## Summary

Real yq's own `-p json` object grammar isn't a stricter or looser version
of JSON's or YAML's flow-mapping grammar -- it isn't a mapping grammar at
all. Traced from `goccy/go-json`'s own source (`Stream.Token()`/
`PrepareForDecode`, the machinery `CandidateNode.UnmarshalJSON` actually
calls for an object body): the decoder reads a flat stream of tokens and
pairs them up (key, value, key, value, ...), where punctuation is just
noise to skip, not structure to validate.

- Before each key: whitespace and *any* run of `,`/`:` is skipped (`{,}`,
  `{,"a":1}`, `{"a":1,,"b":2}`, `{:1}` all collapse).
- A key must be a JSON string; a well-formed *non-string* JSON literal key
  (`{1:1}`, `{true:1}`, `{null:1}`) is the reference's own interface-
  conversion panic (exit 2) -- reproduced as a clean rejection per
  ADR-0018 rule 4(c), not the crash.
- After the key: whitespace and *at most one* `,`/`:` is skipped -- a
  stricter rule, so `{"a" 1}`/`{"a":1}`/`{"a",1}` reach the value but
  `{"a"::1}`/`{"a",,1}` do not.
- The value must start on a genuine JSON value byte, with its own token
  boundary (stopping at the first byte that can't continue it, not the
  next comma/brace) -- this is what actually fixes the issue's own
  headline row, `{"a":1 "b":2}` -> `{"a":1,"b":2}` (previously a silent
  wrong value). Nested arrays are unaffected, keeping their own strict
  `[...]` rules (#2279/#2778).

`Parser::parse_json_strict_flow_mapping_entries` (`src/yaml/parser.rs`) is
a wholly separate loop from the ordinary YAML flow-mapping entries
(`parse_yaml_flow_mapping_entries`), entered under the existing
`json_strict` flag.

## Scope

Applies to the "plain" route (`YamlIndex`-backed) only -- `--inplace`/
`--slurp`/`--eval-all` still materialize through a separate `JsonIndex`-
backed DOM bridge that keeps #1975's stricter grammar. Route unification is
a separate, larger change touching the destructive `--inplace` write path,
filed as #2883.

Also deliberately not reproduced (documented in
`docs/compliance/yq/limitations.md`): the reference's token scanner treats
*every* bracket character as ending the current object's key-reading loop,
even a mismatched one (`{"a":1 ]}` -> `{"a":1}` in real yq) -- essentially
never occurs in real-world JSON, tracked in the same follow-up.

Fixes #2777

## Test plan
- [x] `cargo test --features cli --lib` (3729 passed)
- [x] `cargo test --features cli --test yq_cli_tests` (1883 passed, +5 new)
- [x] `cargo test --features cli --test jq_cli_tests` (1684 passed)
- [x] `cargo test --features cli --test yq_golden_tests` / `yaml_test_suite`
      (conformance intact, unaffected)
- [x] Live oracle spot-checks against Homebrew yq v4.53.3 across the
      issue's own 8 repro rows plus nested/mixed-leniency/bool/null/float
      combinations -- all match exactly

### DIFF
--- a/docs/compliance/yq/limitations.md
+++ b/docs/compliance/yq/limitations.md
@@ -1009,13 +1009,15 @@ over preserving that accidental top-level agreement; the resulting divergence fr
 the DOM bridge's other flow-mapping delimiter gaps in #2777 below, not carved out as a
 special case here.
 
-Scope is flow **sequence delimiters** only, and never mapping delimiters, because that is
-where real yq (v4.53.3, captured live) actually draws the line — the asymmetry this section
-already describes above. Tightening mappings would refuse `{"a":1,}`/`{,}`, all of which
-real yq accepts. The remaining divergences on this route are tracked separately: the
-flow-mapping *delimiter* grammar in
-[#2777](https://github.com/rust-works/succinctly/issues/2777) (including one silent wrong
-value, `{"a":1 "b":2}`), and genuine-YAML `[,]`/`{,}` in
+Scope was originally flow **sequence delimiters** only, and never mapping delimiters,
+because that is where real yq (v4.53.3, captured live) actually draws the line — the
+asymmetry this section already describes above: tightening mappings toward JSON would
+refuse `{"a":1,}`/`{,}`, both of which real yq accepts. The flow-**mapping** side (real yq's
+own, quite different, token-*pairing* leniency — including the one silent wrong value this
+issue originally found, `{"a":1 "b":2}`) is closed separately below by
+[#2777](https://github.com/rust-works/succinctly/issues/2777), on the plain route only —
+see that section for what it does and does not cover. Genuine-YAML `[,]`/`{,}` (unrelated to
+JSON-sourced input at all) is tracked separately as
 [#2779](https://github.com/rust-works/succinctly/issues/2779). (Scalar grammar, `[01]` vs
 `.5`, was the other half of #2279's original scope note — closed separately below by
 [#2778](https://github.com/rust-works/succinctly/issues/2778).)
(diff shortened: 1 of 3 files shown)


newhoggy:
Code review (5 finder angles) found one real bug and several doc/dead-code
issues, all fixed:

1. **Real bug**: `parse_double_quoted()` returns the token's own length, not
   an absolute position — both new call sites here passed that length
   straight to `set_bp_text_end`, which expects an absolute offset, writing
   wrong end positions for every JSON-sourced quoted mapping key/value.
   Fixed by using `self.pos` (matching every one of the ~12 other call sites
   in the file), and extracted the duplicated open/parse/close idiom into
   `parse_json_strict_quoted_node` so both key and value positions share one
   correct implementation.
2. Removed a dead `json_strict` branch in `parse_yaml_flow_mapping_entries`
   (unreachable after the #2777 dispatch split) and fixed three stale doc
   comments that still described flow-mapping keys/delimiters as untouched.
3. Corrected an overclaim in the "Structural, key-agnostic rejections" doc
   comment — live-verified that `{"a": &x 5}`/`{&x "a":1}` were silently
   accepted on `main` before this fix, closed as a side effect of requiring
   a specific key/value shape, now documented and pinned with a new test.
4. Two proposed fixes (number-token `+`/`-` boundary, keyword trailing-byte
   boundary) were checked against the live oracle first and found to already
   match real yq's own greedy-tokenize-then-`ParseFloat`-reject behavior —
   not applied, since they'd have introduced a real divergence rather than
   fixing one.
5. Extracted a shared `is_json_strict_number_byte` predicate and collapsed
   two sequential match blocks into one, per the reuse/efficiency pass.

Re-ran the full suite (lib/yq_cli/jq_cli/yq_golden/yaml_test_suite, clippy,
doc, no_std, release) plus live oracle spot-checks — all green.

newhoggy:
`Perf Regression Guard (ARM64-Linux)`/`Perf Regression Guard (x86_64)` are
failing, but in the *improving* direction — `users_yq_keys_unsorted` measured
**-10.8%/-10.4% instructions** against this PR's merge-base (the guard fails
on drift in either direction by design, so an improvement fails it too).

Mechanism: `parse_json_strict_flow_mapping_entries` is a dedicated loop for
JSON-sourced flow mappings, replacing the generic YAML flow-mapping loop
genuine YAML still uses. That generic loop's per-entry overhead (anchor/tag/
alias lookahead, broader key dispatch) was paid on every key even though
almost none of it can fire under `json_strict` — the dedicated loop's key
check is just `self.peek() == Some(b'"')`. `keys_unsorted` is the shape most
exposed to that (all key parsing), which is why this recovers roughly what
#2778 previously cost that same benchmark, landing back within ~1% of the
pre-#2778 baseline. Documented in `docs/compliance/yq/limitations.md`.

Not a required check (confirmed via the branch protection ruleset), so this
doesn't block merge. I haven't touched `tests/data/perf-guard-baseline.json`
since that file already appears stale relative to current `main` (predating
even #2778, independent of this PR) and updating it correctly needs the
exact CI-matched x86_64-Linux/ARM64-Linux hardware — leaving that
reconciliation to whoever normally owns it rather than guessing from
non-matching hardware.

Issue #2886, opened 2026-09-13
yaml: succinctly yq accepts a YAML flow-mapping stray comma (`{,}`) that real yq rejects


newhoggy:
**Severity: Low** (succinctly is more lenient than real yq here, not silently wrong — but it's an unrecorded ADR-0018 divergence)

## Summary

Surfaced during #2731's investigation (a JSON/jq-mode fix for the same byte shape, `{,}`/`[,]` — this issue is the YAML/yq-mode side, deliberately out of scope there).

The same stray-comma-with-zero-members bytes are legal JSON-in-YAML flow-mapping syntax when read via `succinctly yq`, but real yq (v4.53.3) rejects them:

```
$ printf '{"a":{"b":{,}}}' | succinctly yq -o json .
{
  "a": {
    "b": {
      "": null
    }
  }
}
$ printf '{"a":{"b":{,}}}' | yq -o=j .
Error: bad file '-': yaml: while parsing a flow node at <unknown position>: did not find expected node content
```

succinctly's YAML flow-mapping parser accepts the bare `,` as if it introduced an
empty-string key with a `null` value (`{"": null}`), where real yq's underlying
`gopkg.in/yaml.v3` parser treats it as a flow-node-content error.

## Scope

This is YAML-grammar-scoped (unrelated to #2594/#2731's JSON semi-indexing fix, which
lives entirely in `src/jq/eval_generic.rs`'s value-domain walks over an already-parsed
document) — the divergence is in how succinctly's YAML flow-collection parser accepts
this byte sequence in the first place, not in how a later `paths`/`getpath`-style walk
handles an already-parsed container.

No open PR references this issue.

Issue #2262, opened 2026-09-03
jq/yq: #2211/#2243's malformed-comma checks haven't reached eval.rs's own to_owned_at_depth, eval_generic's cursor-less sibling, or yq_runner.rs's --slurp/--eval-all materializer


newhoggy:
## What's wrong

`eval_generic::to_owned_cursor_at_depth` -- the function #2211 and #2243 both
fixed -- is one of (at least) four independent materializers in this codebase
that each walk an object/array and build an `OwnedValue`. The other three
never received either fix:

1. **`src/jq/eval.rs`'s own `to_owned_at_depth`** -- the materializer behind
   this crate's *documented public API* (`use succinctly::jq::{parse, eval};`
   per this repo's own CLAUDE.md example). Has zero gap-checking of any kind:
   not #1677's delimiter check, not #2211's `container_gap_ok`, not #2243's
   `trailing_element_gap_ok`. Confirmed live by calling `succinctly::jq::eval`
   directly on a cursor built from raw `[1,]` bytes: `length` returns
   `Owned(Int(1))`, `.[]` returns `Many([Number(1)])`, `to_owned(&cursor.value())`
   returns `Ok(Array([1]))` -- no error, for all three.

   Currently **not independently reachable with raw untrusted document text**
   through the shipped CLI: the only non-test production callers
   (`eval_generic.rs`'s `eval_on_owned` reindex bridge, `yq_runner.rs`'s
   `evaluate_input`/`eval_owned_with_file_index`) always re-serialize an
   already-materialized `OwnedValue` via `to_json_for_reindex()` before
   building a fresh index and calling `eval::<...>` -- so this is a latent
   completeness gap for the library API, not a currently-exploitable CLI one.
   Still worth closing so a future direct caller of the documented public
   entry point doesn't silently inherit bugs #2211/#2243 already fixed
   elsewhere.

2. **`eval_generic.rs`'s own cursor-less sibling `to_owned_at_depth`**
   (distinct from the cursor-carrying `to_owned_cursor_at_depth` #2243
   patched, living ~167 lines above it in the same file) -- reached from
   `GenericResult::One(v) => to_owned(&v)` at dozens of call sites whenever a
   value materializes without a live cursor. Same zero-checks gap as (1),
   but inside the very file #2243 just edited.

3. **`src/bin/succinctly/yq_runner.rs`'s `to_owned_canonicalizing_numbers_at_depth`**
[...]

Issue #2261, opened 2026-09-03
jq: trailing comma after a real last child still passes cursor-transparent fast paths (.[], keys, .a, length, .[0])


newhoggy:
## What's wrong

#2243 closed the trailing-stray-comma-after-a-real-last-child gap (`[1,]`,
`{"a":1,}`) for `eval_generic::to_owned_cursor_at_depth` -- the materializer
behind the `evaluate_bytes_lazy`/non-cursor-transparent path (`if`/arithmetic/
function calls, anything `expr_is_cursor_transparent` answers `false` for).

That fix does not reach the *cursor-transparent* fast paths, which take a
completely different route through `eval_generic.rs` and never call
`to_owned_cursor_at_depth` (or its `container_gap_ok`/`trailing_element_gap_ok`
checks) at all:

- `.[]` array iteration -- `each_lazy_array_iterate_sink` only checks each
  element's own *leading* `,` (#1677's `element_gap_ok`), never the trailing
  one after the last real element.
- `keys`/`keys_unsorted`/`to_entries`/plain field materialization --
  `each_lazy_keys_iterate_sink` has the identical bypass for objects.
- Bare field/index access -- the native `Expr::Field` (`.a`) and `Expr::Index`
  (`.[0]`, `.["a"]`) arms resolve via `fields.find_cursor(name)` /
  `elements.get_cursor(idx)`, neither of which ever reaches
  `to_owned_cursor_at_depth`'s checks.
- `Builtin::Length` -- answers from `elements.len()`/`effective_len` (a
  structural count derived from the semi-index's own child walk), never from
  a materializing path, so a trailing stray comma is invisible to it too.

## Repro (confirmed live against `/usr/bin/jq` 1.7.1)

```console
$ echo '[1,]' | jq -c '.[]'          # parse error, exit 5
$ echo '[1,]' | sjq -c '.[]'         # 1, exit 0 -- wrong
$ echo '[1,]' | jq -c 'length'       # parse error, exit 5
$ echo '[1,]' | sjq -c 'length'      # 1, exit 0 -- wrong
$ echo '[1,]' | jq -c '.[0]'         # parse error, exit 5
$ echo '[1,]' | sjq -c '.[0]'        # 1, exit 0 -- wrong
$ echo '{"a":1,}' | jq -c 'keys'     # parse error, exit 5
$ echo '{"a":1,}' | sjq -c 'keys'    # ["a"], exit 0 -- wrong
[...]

PR #2888, opened 2026-09-13
fix(jq): close #2594's stray-comma gap for paths/leaf_paths/getpath's nested walk

### LINKED ISSUE #2594: jq: the zero-child stray comma ({,} / [,]) is still accepted by every filter but identity
**Severity: Medium** (wrong output + exit 0 on malformed input, silently -- the
same class as #2349, on documents real jq rejects outright)

## Summary

The zero-child stray comma (`{,}` and `[,]`, #2211's own shape) is still
accepted by every filter except identity. `.` was fixed; `.a`, `keys`,
`length`, `has(...)`, `to_entries` and `.[0]` all still return an answer for a
document real jq refuses to parse.

The **non**-zero-child case (`{"a":1,}`, #2243's shape) is correctly rejected on
all of them -- which is the tell for the mechanism below.

## Repro (jq 1.7.1, verified live 2026-09-07)

Real jq rejects all of these with `jq: parse error`. succinctly:

| document | filter | succinctly |
|---|---|---|
| `{,}` | `.` | rejects (correct, #2211) |
| `{,}` | `.a` | `null` |
| `{,}` | `.a?` | `null` |
| `{,}` | `keys` | `[]` |
| `{,}` | `length` | `0` |
| `{,}` | `has("a")` | `false` |
| `{,}` | `to_entries` | `[]` |
| `[,]` | `.` | rejects (correct) |
| `[,]` | `.[0]` | `null` |
| `[,]` | `keys` | `[]` |
| `[,]` | `length` | `0` |
| `[,]` | `to_entries` | `[]` |
| `[,]` | `.a?` | (empty) |

Control, all correctly rejected: `{"a":1,}` with `.`, `.a`, `keys`.

## Root cause

`child_tail_gap_ok` (`src/jq/document.rs`) is documented as deliberately
unable to see the zero-child case -- it takes `last_child: Option<&C>` and its
`None` arm is `Ok(())`, because a walk holding only children has nothing to
check `{,}` against. `container_tail_gap_ok` exists for exactly that gap: its
`None` arm calls `container.container_gap_ok(close)`.

The walks behind the diverging rows never reach either. They hand-roll
`child_tail_gap_ok`'s body -- the one-armed
`if let Some(last) = last_value_cursor { if !trailing_element_gap_ok(..) {..} }`
-- so they inherit its blind spot without inheriting the `container_*` fix:

- `JsonFields::find` / `find_cursor` (`src/json/light.rs`) -- the `.a`/`has`
  rows.
- `standard_json_to_jq_value` (`src/bin/succinctly/jq_runner.rs`) -- no tail
  check at all.
- `validate_json_delimiters` / `print_json` (same file) -- a merged
  `gap_start` shape that likewise never checks the empty case.

That `{"a":1,}` is caught everywhere and `{,}` nowhere is the signature: the
`Some(last)` arm they hand-copied works; the `None` arm they did not copy is
the one that catches `{,}`.

## How it was found

By the walk-shaped tail audit added for #2404
(`style_0013_tail_gap_is_routed_or_exempted`). It flags a walk that validates
its members but never reaches a tail helper; these five functions are five of
its six findings. The audit is a static rule, so it found them without needing
the corruption to be reachable from any particular query -- the property
#2404 argued for.

## Suggested fix direction

Route each through `container_tail_gap_ok`/`tail_gap_ok` where a container
cursor is in scope, which fixes the `None` arm by construction. Two cautions:

1. **The error values differ.** These sites raise
   `EvalError::malformed_json_text(..)` / `malformed_member_error()` /
   `malformed_element_error()`; the helper raises
   `last.malformed_delimiter_error()`. Routing verbatim changes the message
   on inputs that already error today, so check the goldens rather than
   assuming the swap is behaviour-preserving.
2. `print_json` and `validate_json_delimiters` carry `// STYLE-0013:`
   exemptions citing a measured perf rationale (#1643/#1676) for hand-rolling
   the member half. Whether that rationale extends to the tail half is a
   separate question -- it is one check per container, not per member.

Once fixed, drop the corresponding `// STYLE-0013-TAIL:` exemptions so the
audit locks the fix in.

## Related

- #2404 -- the audit that found this.
- #2211 / #2243 / #2262 / #2349 / #2358 / #2409 -- the same drift, earlier rounds.

### LINKED ISSUE #2731: jq: the zero-child stray comma survives in the two value-domain walks #2594 could not reach
**Severity: Low**

## Summary

#2594 closed the `{,}`/`[,]` gap at every evaluator arm that holds a container cursor.
Two walks remain, both because they carry **no** cursor by construction, so there is
nothing for `container_tail_gap_ok` to check against:

1. `to_owned_at_depth`'s cursor-less top level (`src/jq/eval_generic.rs`) — the
   long-standing #2262 residual, pinned by
   `test_to_owned_stray_comma_in_empty_container_remains_a_known_gap_2262`.
2. `collect_paths_generic`'s value-domain recursion — a `{,}`/`[,]` nested *inside* a
   `paths`/`leaf_paths` walk. #2594 guards the container `paths` is standing on; the
   recursion below it descends over values only.

The same shape also survives one lazy read path:

## Repro

`{,}`/`[,]` at the root is now rejected on every filter. Nested, reached through a
lazy read that never materializes the parent, it is not:

```console
$ echo '{"a":1,"b":{,}}' | jq -c '.a'
jq: parse error: Expected value before ',' at line 1, column 13
$ echo '{"a":1,"b":{,}}' | succinctly jq -c '.a'
1
```

That third case is **not** a bug in the same sense: it is the standing lazy-validation
stance, and the controls behave identically —

```console
$ echo '{"a":1,"b":xyz}'   | succinctly jq -c '.a'   # 1
$ echo '{"a":1,"b":[1 2]}' | succinctly jq -c '.a'   # 1
```

— so it is listed here only so the accounting is complete, not as something to fix
without deciding the wider question (`--validate` exists for callers who want it).

## Root cause

`JsonFields`/`JsonElements` (`src/json/light.rs`) store only
`Option<JsonCursor>` — "the first child's cursor, or `None`". A zero-child container
therefore leaves the walk with nothing at all, and `container_gap_ok` needs the
container's own opening bracket. #2594 worked around this by checking at the *callers*,
which do hold a cursor. These two walks have no such caller: `to_owned`'s public entry
point starts at `cursor: None` by design, and `collect_paths_generic` recurses over
values.

## Why this is separate from #2594

#2594's fix is complete for what it scoped — every filter in its repro table, plus the
path domain found in review. These two need a different mechanism, not another call
site, and that mechanism is a data-structure change worth its own decision.

## Suggested fix direction

The only thing that closes the class by construction is giving `JsonFields`/
`JsonElements` the container cursor they are built from — `from_object_cursor`/
`from_array_cursor` (`src/json/light.rs:1146`/`:1529`) already receive it and discard
it, and there are only five construction sites in total (those two plus three `rest`
sites in `uncons`/`uncons_cursor`). A `DocumentFields`/`DocumentElements` accessor on
top of that would let every shared walk in `document.rs` upgrade `child_tail_gap_ok`
to `container_tail_gap_ok` in place, covering `eval.rs`, `lazy.rs`, `yq_runner.rs` and
`jq_runner.rs` at once, and would let #2594's per-arm guards be deleted again.

**This is why it is filed rather than done.** Measured on `main`:

| type | now | with the container cursor added |
|---|---:|---:|
| `JsonCursor` | 32 B | 32 B |
| `JsonFields` / `JsonElements` | 32 B | 64 B naive, 40 B packed (`{container: JsonCursor, child_bp: usize}`, exhausted ⟺ `child_bp == container.bp_pos`) |
| `StandardJson` | 40 B | 48 B at best |

That is a 20% growth of the crate's hottest value type, so it needs a real A/B on both
architectures per `docs/guides/benchmarking.md`, and is Tier 3 (benchmark-machine
gated) rather than something to land opportunistically. A cheaper alternative worth
pricing first: rejecting in `JsonCursor::value_at` by returning `StandardJson::Error`
for a bad empty container — nearly free (the emptiness is already known there), but it
downgrades the error message, since `StandardJson::Error` carries a `&'static str`
while today's raise re-runs the strict validator to name the real syntax error
(`expected string key, found ','`).

## Related

- #2594 — the arms this could replace.
- #2262 — where residual (1) is already recorded.
- #2211 / #2243 / #2349 / #2358 / #2409 — earlier rounds of the same drift.

### LINKED ISSUE #2751: jq: a wrong-arity builtin call re-parses its whole argument subtree, so nesting is exponential (has(...) at depth 18 takes ~2.5s)
**Severity: Medium** (parse-time DoS on attacker-supplied filters; no wrong answer)

## Summary

When `try_parse_builtin` detects a wrong arity it rewinds and re-parses the call as an
ordinary generic call. That reparse walks the *entire* argument subtree again, so nesting
wrong-arity calls inside each other compounds: each level re-parses every level below it.

## Repro

`has` is a plain jq builtin needing no flags. Program is
`has(has(...has(1);2...);2)` at N levels, debug build, on `main` (`944892a1b`):

```console
$ depth=12   succinctly 0.02s    jq 0.00s
$ depth=16   succinctly 0.71s    jq 0.00s
$ depth=18   succinctly 2.47s    jq 0.00s
$ depth=20   succinctly ~11s     jq 0.00s
```

Roughly 4x per level. jq is instant at every depth. `contains` (2.87s) and `getpath`
(3.03s) behave identically at depth 18. Well under `MAX_EXPR_DEPTH`, so the existing
depth guard never fires.

## Root cause

Two paths reach the generic-call fallback in `try_parse_builtin`'s caller
(`src/jq/parser.rs`):

- the **`Err(...)` arm** goes through `retry_shadow_candidate_as_generic_call`, which
  holds a `shadow_retry_budget` added *precisely* to bound this — its doc comment
  describes the `O(2^depth)` blowup and calls it "a real, if far more contrived, second
  instance of the same denial-of-service shape";
- the **`Ok(None)` arm** (`parser.rs:1824`, the wrong-arity rewind that
  `parse_required_single_arg` produces) calls `parse_func_call_or_error()` **unguarded**.

So the budget that exists for one path does not cover the other.

## Why the obvious fix is not obviously right

Routing `Ok(None)` through the same budget is a one-line change, but the two paths want
different post-budget behaviour. Past the budget `retry_shadow_candidate_as_generic_call`
propagates the original parse error — acceptable for its adversarial-shadow case. Doing
that on the `Ok(None)` path would make an *ordinary* wrong-arity call start reporting a
parse error again past some threshold, which is exactly the divergence #2237/#2389/#2573
exist to remove (`bsearch(1;2)` must report `bsearch/2 is not defined`). A program with
many independent wrong-arity calls would regress.

Likely wants either memoisation of the reparse, or building the generic call from the
already-parsed argument the way `wrap_shadowable_call` does on the successful path, so no
reparse happens at all.

## Scope note

Pre-existing and **not** introduced by #2573 (PR #2750) — `has`/`contains`/`getpath`
reproduce identically on `main`, from the #2237/#2389 batches. #2573 does widen the set of
affected names by 15 (`bsearch`, `pick`, `omit`, `strftime`, `strptime`, `tz`, `load`,
`at_offset`, `fromstream`, `truncate_stream`, `indices`, `rindex`, `index`, `inside`,
`isvalid`), several of which — `index`, `indices`, `pick` — are likelier to appear nested
in real programs than `getpath`. It does not change the security posture, since `has`
already reaches the same shape with no flags required.

Found via adversarial review of #2573 (PR #2750).

newhoggy:
## Summary
- `collect_paths_generic` (`paths`/`leaf_paths`) and `getpath_walk_cursor` (`getpath`) are value-domain walks that carry no cursor to check a container's own trailing `{,}`/`[,]` gap against — #2594's per-arm pre-check only ever covered the *outermost* container these builtins were called on, so a `{,}`/`[,]` reached through recursion (`paths`) or a child-miss exit (`getpath`) silently answered `null`/skipped it instead of raising, e.g.:
  ```
  $ echo '{"a":1,"b":{,}}' | succinctly jq -c '[paths]'        # was [["a"],["b"]], now raises
  $ echo '{"a":1,"b":{,}}' | succinctly jq -c 'getpath(["b","x"])'  # was null, now raises
  ```
- `collect_paths_generic` now threads a `cursor: Option<&V::Cursor>` down through every recursive call, mirroring `to_owned_at_depth`'s identical shape (#2358). `getpath_walk_cursor` runs the same gap check at its two child-miss exits before answering `null`.
- The two `Builtin::Paths`/`Builtin::LeafPaths` dispatch sites' own top-level pre-check is now redundant (the walk covers every level itself) and is removed.
- Residual 1 from #2262 (`to_owned(&V)` at the true top level, `cursor: None`) is unaffected — no CLI spelling, out of scope per the issue's own triage.
- Two follow-ups filed for out-of-scope findings surfaced during investigation: #2886 (yq mode accepts YAML flow-mapping `{,}` where real yq rejects it — unrelated YAML-grammar-scoped divergence).
- Also closed #2751 as a duplicate of #2749 (a separate, already-merged PR from this session) — unrelated to this PR's own diff.

Fixes #2731

## Test plan
- [x] New positive-repro test (`test_jq_nested_value_domain_walks_empty_container_now_raises_2731`) covering the issue's own 6-row table plus array-nested, leaf_paths, and the outermost-container-redundancy case.
- [x] New negative-control test (`test_jq_nested_value_domain_walks_wellformed_unaffected_2731`) confirming genuinely empty nested containers are unaffected.
- [x] Existing `#2594`/`#2261`/`#2262` regression tests pass unmodified (17 tests).
- [x] Interleaved A/B timing on a 20MB document's `[paths] | length` — noisy on this non-benchmark machine (excluded from `nodes.yaml`'s bench fleet) but no clear regression; the added check only fires on the empty-container branch, so a well-formed document pays only the cursor-threading cost `to_owned_at_depth` already established as negligible. CI's perf-guard is authoritative here.


newhoggy:
Ran the mandatory 8-angle code review. Multiple independent angles (A, D, G) converged on the same real, confirmed regression:

**Confirmed and fixed:** the array arm of `collect_paths_generic` switched from `uncons()` to `uncons_cursor()` purely to get a cursor for recursion, but dropped the per-element/trailing gap checks the object arm gets for free from `effective_fields_checked`. A non-empty array with a stray comma nested inside a `paths`/`leaf_paths` walk (e.g. `{"a":[1,,2]}`, or a trailing `[1,2,]`) silently walked past it instead of raising, where `.a[]` on the same document already did. Verified live before and after the fix. Fixed by routing through `collect_cursors_checked` (already used elsewhere in this file) instead of a second hand-written copy of `to_owned_at_depth`'s inline checks — same-shape fix, no new mechanism. Added a dedicated regression test.

**Also fixed** (lower-severity, from angles E/F): the two new `getpath_walk_cursor` gap checks routed through `empty_container_gap_error(&v, ..)`, which re-derives `fields`/`elements` via a second `v.as_object()`/`as_array()` call — on a YAML alias this re-walks the chain a second time for nothing, exactly the cost this function's own doc comment already warns against. Both now call the specific helper directly against the container already in scope.

**Also fixed** (angles A/C/E): `empty_container_gap_error`'s doc comment was stale, still describing the nested-`{,}` gap as unchecked and naming `getpath_walk_cursor` as a caller (no longer true after the fix above).

All existing + new tests pass (48 total for this area), clippy/fmt/docs clean, full suite green.

Issue #2889, opened 2026-09-13
jq: TrackedVar rebuilt-root fix (#2642) refuse-only residual — owned embed map for pass-through constructions


newhoggy:
## Context

#2642 (fixed in this issue's own PR) adds `RootWitness`/`demote_rebuilt_markers`
(`src/jq/eval.rs`) so a root-bound `$x`'s `Origin::Snapshot` marker is demoted
to `Untracked` before crossing into the owned-value evaluator whenever it
isn't proven to name that call's own document node -- closing the
value-equality-only certification hole that let a rebuilt equal-valued copy
be wrongly admitted (`. as $x | {a:1} | ($x.a) = 9` used to write `{"a":9}`;
jq refuses).

The fix is refuse-only by construction (a marker is only ever demoted from
`Snapshot` to `Untracked`, never the reverse), which means it also demotes a
few shapes that real jq's own reference-counted `jv` keeps as the *same*
node without ever allocating a new container for it. These are documented,
accepted residuals, not new correctness gaps -- but they are real,
verified-live regressions from "accept" to "refuse" that #2642's own triage
plan explicitly scoped to a follow-up rather than this fix.

## 1. "Owned embed map" — a handful of constructions lose accepted status

```console
$ echo '{"a":1}' | jq -c '. as $x | [.] | .[0] | path($x)'
[]
$ echo '{"a":1}' | succinctly jq -c '. as $x | [.] | .[0] | path($x)'
jq: error (at <stdin>:0): Invalid path expression with result {"a":1}
```

Same for `{k:.} | .k | path($x)`, `. + {} | path($x)`, and the navigation
happening *inside* the same `path()` invocation the marker is referenced in
(`[.] | path(.[0] | $x)`). In each case jq's own `jv` for the embedded
position is literally the same allocation as `.` itself (array/object
construction increments a refcount rather than copying), so `jv_identical`
still recognizes it; succinctly's `OwnedValue`-cloning model has no
per-position "this embeds a document node" fact to check against.

Recovering this needs the generic evaluator to record, for a constructed
value, which of its positions embed a document node (and which document/id)
— an "owned embed map" alongside the existing per-binding `BindOrigin`
tracking (#2072). #2642's own triage plan named this "Out of scope, to file
at stage 4" and it's filed here per that plan.

## 2. `reduce empty as $i (.; .)` — zero iterations, but still demoted

```console
[...]

PR #2890, opened 2026-09-13
fix(jq): stop a root-bound $x from certifying against a rebuilt copy (#2642)

### LINKED ISSUE #2642: jq: a root-bound `$x` matches a rebuilt equal-valued copy — TrackedVar's value witness writes where jq refuses
**Severity: Medium** (accept-where-jq-refuses — `=`/`del()` write a document jq leaves untouched)

## Summary

A variable bound from `.` by the **value evaluator** (`eval_as`) is certified in the path
resolver by *value equality* against the ambient input. When the resolver's root is a
**rebuilt copy** of that value — a construction or a `tojson|fromjson` round-trip between the
binding and the write — the copy passes the check, and the write goes through. jq's
`jv_identical` compares pointers, so it refuses.

Found while planning #2042 (2026-09-08); it is #2042's bug class (equal value ≠ same node)
reached through a rebuild instead of a sibling, on the marker that *already* exists.

## Repro (`main` @ baa26d72e vs jq 1.7.1)

```console
$ echo '{"a":1}' | jq -c '. as $x | {a:1} | ($x.a) = 9'
jq: error (at <stdin>:0): Invalid path expression near attempt to access element "a" of {"a":1}
$ echo '{"a":1}' | succinctly jq -c '. as $x | {a:1} | ($x.a) = 9'
{"a":9}                                    # <- written

$ echo '{"a":1}' | jq -c '. as $x | [(tojson|fromjson)] | del(.[0] | $x)'
jq: error (at <stdin>:0): Invalid path expression with result {"a":1}
$ echo '{"a":1}' | succinctly jq -c '. as $x | [(tojson|fromjson)] | del(.[0] | $x)'
[]                                         # <- deleted

$ echo '{"a":1}' | jq -c '. as $x | {a:1} | path($x)'                 # refuses
$ echo '{"a":1}' | succinctly jq -c '. as $x | {a:1} | path($x)'      # []
$ echo '{"a":1}' | jq -c '. as $x | (tojson|fromjson) | path($x)'     # refuses
$ echo '{"a":1}' | succinctly jq -c '. as $x | (tojson|fromjson) | path($x)'   # []
$ echo '{"a":1}' | jq -c '. as $x | {k: (tojson|fromjson)} | path(.k | $x)'    # refuses
$ echo '{"a":1}' | succinctly jq -c '. as $x | {k: (tojson|fromjson)} | path(.k | $x)'  # ["k"]
```

Controls (both agree):

```console
$ echo '{"a":1}' | jq -c '. as $x | [.] | path(.[0] | $x)'                          # [0] — same node
$ echo '{"a":1}' | jq -c 'path(. as $x | [(tojson|fromjson)] | .[0] | $x)'          # refuses in both
```

The last control is the tell: with the binding **inside** `path(...)`, #2041's
`cannot_move_register` allowlist sees the construction stage and refuses. Across the
value-evaluator boundary nothing sees it.

## Root cause

`resolve_node`'s `Expr::TrackedVar` arm (`src/jq/eval.rs`) certifies the marker by
`marker_value == value`. That is sound only while a value-equal ambient is necessarily the
same node — true for a root snapshot against its own proper descendants (a finite tree cannot
contain itself), false once a stage between the binding and the write **rebuilds** an equal
value. The binding is made by `eval_as` (value mode), which has no path and no "was this
rebuilt" fact to pass on.

## Fix direction

Either of two witnesses closes it; both belong to the frame problem #2042's plan lays out:

- **Rebuilt mark on the value side**: `eval_as`'s body evaluation knows which stages sit
  between the binding and the `path()`/write call; a stage failing `cannot_move_register`
  demotes the marker to an untracked snapshot (refuse-only, so this can only *lose*
  acceptances jq also refuses).
- **Document-absolute identity**: a value-mode binding taken from a cursor carries its
  document path; a write/`path()` root that is `Cow::Owned` (rebuilt) has none, so the
  marker cannot match. This is the same mechanism the value-mode half of #2042 (row J4,
  `.a as $y | path(.a | $y)`, jq `["a"]`, succinctly refuses) needs in the *accepting*
  direction, so the two should be designed together.

## Related

- #2042 — the plan comment there records this as the "K-rows", out of its scope.
- #1466 — the accept-where-jq-refuses class.
- #2041 — `cannot_move_register`, which is why the inside-`path()` control refuses.

newhoggy:
## Summary

`resolve_node`'s `Expr::TrackedVar` arm certifies an `Origin::Snapshot`
marker (a root-bound `. as $x`) by value equality alone. That's sound only
while a value-equal ambient is necessarily the same node — true for a root
snapshot against its own proper descendants (a finite tree cannot contain
itself), false the moment a stage between the binding and a `path()`/write
call **rebuilds** an equal-valued copy:

```console
$ echo '{"a":1}' | jq -c '. as $x | {a:1} | ($x.a) = 9'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of {"a":1}
$ echo '{"a":1}' | succinctly jq -c '. as $x | {a:1} | ($x.a) = 9'
{"a":9}    # <- wrongly written
```

Adds `RootWitness` (a funnel call site's own document node, or `Owned` when
there is none) and `demote_rebuilt_markers`, which rewrites a `Snapshot`
marker not proven to name that root to `Untracked` before the expression
ever reaches `Frame::certifies` — refuse-only by construction, since the
only transition is `Snapshot -> Untracked` and `Untracked` never certifies.
Applied at every "funnel" call site in `eval_generic.rs` that hands an
expression from the generic (cursor-based) evaluator to `eval.rs`'s owned
evaluator (`eval_on_owned`, `eval_each_owned`, `bridge_to_full_evaluator(_flow)`,
`builtin_path_on_owned`, and the top-level `eval_with_cursor_using`/
`eval_each_with_cursor_using` entry points — ~29 call sites total).

## Scope

Documented, accepted residual: a handful of constructions jq's own
reference-counted `jv` passes an embedded node through without copying it
(`[.] | .[0]`, `{k:.} | .k`, `. + {}`, `reduce empty as $i (.; .)`) now
refuse rather than silently accept a copy, since succinctly's
`OwnedValue`-cloning model has no per-position "this embeds a document
node" fact to check — filed as a follow-up (#2889, "owned embed map"), not
attempted here.

Fixes #2642

## Test plan
- [x] `cargo test --features cli --lib jq::` (2302 passed)
- [x] `cargo test --features cli --test jq_cli_tests` (1694 passed, +3 new)
- [x] `cargo test --features cli --test yq_cli_tests` (1888 passed, unaffected)
- [x] Live oracle verification against jq 1.7.1 across the issue's own 5
      repro rows, its full "must stay accepted" control table, and the
      documented refuse-only residual rows


newhoggy:
Code review found two genuine bugs and one genuine regression, all fixed:

1. **Real gap**: `reduce`/`foreach`'s own UPDATE/EXTRACT positions were a
   missed funnel site — the exact #2642 bug survived through
   `. as $x | reduce (1) as $i ({a:1}; ($x.a) = 9)`. Fixed (both the eager
   `eval_single` arm and the streaming `each_foreach_generic` twin), using
   `RootWitness::Owned` since the fold's own accumulator is never
   cursor-backed.
2. **Real regression**: an over-eager `Owned` demotion at the catch-handler
   funnel sites broke `. as $x | try error($x) catch path($x)`, which
   previously worked correctly — `error($x)` raises a marker's own value
   verbatim, and jq's catch handler runs against that same value, so this
   needs to stay untouched. Reverted.
3. **Same over-eager pattern, same fix**: at the `OwnedIdentity`/#2072
   machinery (reached whenever a `key`/`parent`/`path`/`file_index` sibling
   read is present), blanket `Owned` demotion broke a previously-correct
   write (`.foo | . as $x | (parent, ($x.a = 9))`, a succinctly-only
   extension). Reverted to exactly pre-#2642 behavior at these three sites
   rather than threading `OwnedIdentity`'s own `id.base`/`id.ancestors`
   witness through under time pressure — added as a real, still-open item
   to the #2889 follow-up.

Live-verified every reverted site now matches `main`'s own pre-#2642
behavior exactly (neither newly broken nor newly fixed), and every one of
the fix's own repros/controls still holds. Re-ran the full suite
(lib/jq_cli/yq_cli, clippy, doc, no_std, release) — all green.

Thank you for the extremely thorough review passes on this one — the
`reduce`/`foreach` gap and the two over-demotion regressions were real,
and I would not have caught them without the live-comparison-against-main
verification this round did.

Issue #1800, opened 2026-08-29
jq: a comparison-key decode failure preempts a sibling generator's own error/side effects (in/IN/contains/inside)


newhoggy:
**Severity: Low** — a genuinely rare intersection (a document byte-level
decode failure co-occurring with an independent `error(...)`/side effect in
a sibling generator expression), with no real jq/yq oracle to consult since
the triggering condition (an already-materialized invalid-UTF-8 string
reaching this code at all) can't arise in either reference tool, which
validates upstream.

## Summary

Found during `/code-review` on #1799 (the "indirect comparison-key" slice
of #1755). `builtin_in`/`builtin_upper_in`/`builtin_contains`/
`builtin_inside` now materialize their primary input via `to_owned_checked`
*before* evaluating the builtin's own argument expression (`xs`/`s`/
`b_expr`). If the primary input fails to decode, the argument expression is
never evaluated at all -- including any of its own independent errors or
side effects.

## Repro (confirmed live via the public `succinctly::jq::eval` library API)

```
"<undecodable string>" | in({"a":1}, error("boom"))
```
raises the input's own decode-failure error ("invalid UTF-8 in string"),
never reaching `error("boom")` in the argument expression at all.

```
[<array containing an undecodable string>] | contains(error("boom"))
```
same shape: the array's own decode failure preempts `error("boom")`.

## Why this is filed rather than fixed inline

This is the identical "hoisted, pre-fanout materialization" structure that
already existed before #1755 touched these functions (`// Hoisted out of the
fan-out: the input does not vary with b` -- `builtin_contains`'s own
pre-existing comment). Before #1755, that materialization (`to_owned`) was
infallible, so the ordering question never arose; #1755 made it fallible,
which is the minimal, established-pattern fix (matching `to_owned_checked`'s
use everywhere else in this file) -- but it surfaces this pre-existing
structural choice as a newly-observable behavior.

A "fix" that preserves the argument expression's own evaluation/side effects
even when the primary input fails to decode would need to thread the
pending decode failure through the same kind of `Partial`-preserving
machinery this file already uses for `Break`/`Halt` escapes elsewhere (e.g.
`eval_index_expr`'s own `pending_halt`), rather than an unconditional early
`return`. That's a materially larger, riskier change than the narrow slice
each #1755 PR has scoped itself to, and needs its own design pass: should a
decode failure ever defer to let a sibling expression's own independent
error/side effect surface first, or does it always win regardless? There is
[...]

PR #2891, opened 2026-09-13
fix(jq): decide #2202's in/IN decode-order, pin the CLI/library route split

### LINKED ISSUE #2202: jq: in/IN(s) have the identical comparison-key decode-preemption shape as #1800's contains/inside, but with a structural blocker
**Severity: Low** — same rare intersection as #1800 (a document byte-level decode failure racing an independent `error(...)`/side effect in a sibling generator expression), still with no real jq/yq oracle to consult.

## Summary

#1800 fixed `builtin_contains`/`builtin_inside`: their primary input's decode failure (`to_owned`) was returned eagerly, before the builtin's own argument expression (`b_expr`) ever ran — so an argument-side escape occurring *before* the argument's first candidate (e.g. `contains(error("boom"), 1)`) never got a chance to surface, even though it should win per real jq's `. as $x | b as $y | ($x | contains($y))` desugar (`$x`'s content is only needed once `b`'s first candidate exists).

`builtin_in`/`builtin_upper_in` (`in(xs)`/`IN(s)`) have the exact same shape: `key_owned`/`current` (the primary input's decode) is forced *before* `xs`/`s` is evaluated at all, via `to_owned_or_suppress!`, so `in(error("boom"), {"a":1})` on an undecodable-string input raises the decode failure instead of `"boom"`, even though `error("boom")` is `xs`'s leading branch and should win the same way #1800's `contains(error("boom"), 1)` now does.

## Why this needs its own fix (not folded into #1800)

`builtin_contains`/`builtin_inside` evaluate their argument (`b_expr`) against the *raw, undecoded* cursor via `fanout_arg`/`eval_each` (`eval.rs`) — so deferring the primary input's decode into `fanout_arg`'s own `body` closure was a small, safe change: `b_expr`'s evaluation never depended on the primary input's decoded content at all.

`builtin_in`/`builtin_upper_in` instead evaluate their argument (`xs`/`s`) via `eval_owned_multi_keep_partial`/`eval_each_owned`, which take an already-decoded `&OwnedValue` as `.`'s context — there is no lazy-cursor variant of this owned-based machinery. Concretely: if the primary input's decode fails, there is currently no `OwnedValue` to even *start* evaluating `xs`/`s` with, so the #1800-style fix ("defer the decode `Result`, consult it lazily once the argument's own escape has had its chance") isn't directly mechanical here the way it was for `contains`/`inside`.

A sound fix requires converting `xs`'s/`s`'s evaluation to a lazy, cursor-based generator (mirroring how `fanout_arg`/`eval_each` already drive `contains`'s `b_expr` against the raw `value` cursor) instead of the current owned-only helpers — a materially larger, riskier rewrite of two functions with real prior-fix history (`#909`/`#917`/`#932`/`#2015`), not a narrow reorder. That's out of scope for a Low-severity, oracle-less edge case, per this repo's own narrow-PR-plus-tracking-issue precedent (`#1746`→`#1755`).

## Repro (confirmed live via the public `succinctly::jq::eval` library API)

```
"<undecodable string>" | in(error("boom"), {"a":1})
```
should raise `"boom"` (per the ordering rule above) but currently raises the input's own decode-failure error, since `key_owned` is force-decoded before `xs` is ever touched.

## Related

- #1800 — the sibling issue this was split out of, and its already-merged fix for `contains`/`inside` (same file, same `EvalError::is_decode_failure()`/`suppress_or_raise` primitives).
- #910/#932/#1553 — the file's own established generator-evaluation-order and side-effect-preservation rules that any fix here needs to stay consistent with.

newhoggy:
## Summary
- `in(xs)`/`IN(s)` evaluate their argument against the already-decoded input, so they keep the eager decode-failure return rather than `contains`/`inside`'s #1800 deferral. This agrees with `in(xs)`'s own desugar (`. as $x | xs | has($x)`, which already materializes at the `as` binding) rather than diverging from it, and there is no jq oracle for an undecodable input either way, since a reference document containing one is rejected at JSON-parse time — per ADR-0018 this falls through to step 3 (the eager order does strictly less work).
- Rewrites the two stale `#2202` marker comments in `builtin_in`/`builtin_upper_in` (`src/jq/eval.rs`) from "see #2202 before completing #1800 here" to the decision.
- Adds two library-level tests pinning the agreement (`in`/`IN`/the `as`-desugar all raise the same decode failure) and the asymmetry (`contains` still raises `boom`, only observable via the library route, not the CLI).
- Adds a CLI-route test pinning that `eval_generic.rs`'s bridge materializes the input a layer earlier, so `succinctly jq` itself raises the decode failure uniformly for all three builtins, `contains` included.
- Documents the rule in `docs/compliance/jq/limitations.md`.

Closes #2202.

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` (full suite, before and after rebase)
- [x] Verified live against `jq` 1.7.1 and the current binary before implementing


newhoggy:
Ran a 5-way multi-agent review. Summary of what was addressed:

**Fixed (confirmed by 2 independent review agents):** the comment rewrite in `builtin_in`/`builtin_upper_in` overclaimed "there is nothing 'argument first' to defer to" as a semantic necessity of `in(xs)`'s desugar. That's not accurate — `builtin_contains`'s own #1800 fix already shows the same poisoned-value shape (fan the argument out over the cursor, consult a stored `to_owned` `Result` lazily) would work for `in`/`IN` too; it just isn't built. Restored the original comment's honesty about this being an unbuilt option, decided against for now per ADR-0018 step 3 (no oracle either way, current eager `as` binding already agrees, not worth the larger rewrite) — see the latest commit.

**Addressed via added reasoning:** one reviewer raised whether this falls under ADR-0018's #2103/#2168/#2692 amendment (whole-document-rejection references get the *uniform divergence* treatment, not a per-spelling one). I don't think it does — that amendment governs whether *navigating to one value* triggers its own validation depending on filter spelling (one value, one answer either way). #2202 is a different question: two *independent* computations (the ambient input's own decode, and an unrelated argument-side `error(...)`) can each fail, and the question is only which failure is reported first when both are present — the same "earlier step wins" precedent `IN(s)`'s #910/#932 fix and `contains`'s own #1800 fix already establish. Added a paragraph to `limitations.md` making this distinction explicit so it doesn't get re-litigated as ambiguous.

**Fixed (cheap/mechanical):** reused the existing `query!` test macro instead of hand-rolled eval dispatch, hoisted a loop-invariant `JsonIndex::build`, trimmed the same argument being restated near-verbatim across 3-4 doc-comment sites, and reworded the first commit from `fix(jq)` to `docs(jq)` to match repo convention for a no-production-logic-change PR (precedent: `42ab3f38a`).

Full test suite + both clippy variants + rustdoc all green after the fixes and after two rebases onto a fast-moving `main`.

Issue #2632, opened 2026-09-08
jq: resolve_reduce carries the same path-comparison bug #2161 fixed in resolve_foreach (696 measured divergences)


newhoggy:
**Severity: Medium** (wrong error where jq succeeds — a refusal, not a wrong path)

## Summary

#2161 fixed `resolve_foreach`: it decided each fold step's `at_register` from the previous
step's `FoldRegister::branch_provenance` ("did the last UPDATE branch end *at* the
register's own path"), which is `false` from step 2 onward because UPDATE's whole job is to
navigate away from the register. jq's actual rule is
`jv_identical(current_input, value_at_path)` — a step is at the register when the
accumulator arriving at it is still the register's value.

`resolve_reduce` has the identical bug at its own carry-forward site and was deliberately
left untouched by that PR.

## Repro (confirmed live against jq 1.7.1)

```console
$ echo '{"a":1}' | jq -c 'path(reduce (1,2) as $k (.b; .a))'
["b"]
$ echo '{"a":1}' | succinctly jq -c 'path(reduce (1,2) as $k (.b; .a))'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of null
```

The `foreach` spelling of the same fold now succeeds, so the two constructs currently
disagree with each other.

## Root cause

`src/jq/eval.rs` — `resolve_reduce`'s carry-forward:

```rust
let last = branches.into_iter().last();
(acc_at_register, acc_snapshot) = reg.branch_provenance(last.as_ref());
acc = last.map(|b| b.value.into_owned());
```

Same path-comparison-instead-of-identity-check as `resolve_foreach` had.

## Measured, not guessed: the analogous fix works

I prototyped it while fixing #2161 and measured it on the same generated 12,048-case sweep
(4 source streams × 7 INITs × 11 UPDATEs × 12 documents, classified against jq 1.7.1
case by case):

| | count |
|---|---|
| `reduce` divergences before | 728 |
| fixed by the analogous change | **696** |
| broken by it | **0** |
| still diverging after | 32 |

The prototype was:

```rust
if let Some(a) = acc.as_ref() {
    acc_at_register = acc_at_register
        || (reg.trackable && register_identical(&reg.value, a, acc_snapshot));
}
```

## Two things whoever picks this up must not skip

1. **The `||` is load-bearing.** My first attempt at the `foreach` fix *replaced* the
   provenance check rather than OR-ing it, and broke step 1 for object-valued registers:
[...]

Issue #2388, opened 2026-09-04
jq: path(foreach ... as $k ((0,1); $k)) fabricates a second path where jq errors on the second INIT fork


newhoggy:
**Severity: Low**

## Summary

Found during review of #2386 (documenting #2163's INIT-fork null-ambient-input divergence
from real jq). succinctly's own three jq evaluators don't agree with each other on
`foreach`/`reduce`'s INIT-comma fork behavior when SOURCE reads the ambient document — a
pre-existing gap, unrelated to jq fidelity.

## Repro

The value evaluator (`eval.rs`/`eval_generic.rs`, shared core) answers a consistent value
for every INIT fork:

```console
$ echo '{"a":1,"c":2}' | succinctly jq -c '[foreach (.a) as $k ((0,.c); $k)]'
[1,1]
```

But the path-context evaluator (`resolve_reduce`/`resolve_foreach` in `src/jq/eval.rs`) hard
errors on the same filter shape in path/assignment position:

```console
$ echo '{"a":1,"c":2}' | succinctly jq -c 'path(foreach (.a) as $k ((0,.c); $k))'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of {"a":1,"c":2}
["a"]
$ echo '{"a":1,"c":2}' | succinctly jq -c 'path(reduce (.a) as $k ((0,.c); $k))'
jq: error (at <stdin>:1): Invalid path expression with result 1
```

(Note `foreach` prints a partial `["a"]` for its first fork before erroring on the second;
`reduce` prints nothing at all — the two also disagree with each other's partial-output
shape, not just with the value evaluator.)

## Root cause

`resolve_reduce`/`resolve_foreach` (`src/jq/eval.rs`) call `resolve_fold_source` fresh once
per INIT fork (added by #2031 for path-trackability), rather than computing SOURCE once
upfront the way the shared `eval_reduce_with_values`/`eval_foreach_with_values` core does.
#2031's own restriction on navigating INIT forks (`FoldRegister::branch_provenance`) appears
to be what turns a second, non-navigating-but-still-`.`-reading INIT fork into an "Invalid
path expression" error instead of a value.

## Why this is separate from #2163/#2386

#2163/#2386 are about jq-fidelity: real jq's own VM quirk (SOURCE re-entry against a
synthetic `null`) versus succinctly's consistent-value choice, scoped to the two evaluators
that share the once-upfront core. This issue is about succinctly's own *internal*
[...]

PR #2893, opened 2026-09-13
fix(jq): stop foreach EXTRACT fabricating a tracked path, write corruption

### LINKED ISSUE #2860: jq: foreach's EXTRACT fabricates a tracked path through a missing-key/null-absorbing step, corrupting writes (pre-existing, bare $var; newly reachable via #2676 destructuring)
**Severity: High** (wrong-accept / write corruption — `=`/`|=`/`del()` mutate or delete a
document real jq leaves untouched)

## Summary

`foreach SOURCE as PATTERN (INIT; UPDATE; EXTRACT)` fabricates a tracked path (and, on the
write side, actually corrupts the document) when EXTRACT references the loop variable again
after navigating through a **missing key** (or any other null-absorbing step) on it first. This
reproduces with a **bare `$var` pattern — no destructuring involved at all** — and is present
on `main` today (pre-#2676), so it predates and is independent of
[#2676](https://github.com/rust-works/succinctly/issues/2676)'s own destructuring-pattern work,
though #2676 does newly expose the same mechanism to destructured loop variables too (see
below).

## Repro (jq 1.7.1 vs succinctly jq, on `{"a":{"b":1}}`, bare `$var` — pre-#2676 baseline)

```console
$ echo '{"a":{"b":1}}' | jq -c 'path(foreach .a as $v0 (.; $v0; (.zzz | $v0)))'
jq: error (at <stdin>:1): Invalid path expression with result {"b":1}

$ echo '{"a":{"b":1}}' | succinctly jq -c 'path(foreach .a as $v0 (.; $v0; (.zzz | $v0)))'
["a"]
```

Write side actually corrupts the document (confirmed against a build of `main` at commit
`7bea9bfbc`, i.e. before #2676 touched anything in this area):

```console
$ echo '{"a":{"b":1}}' | jq -c '(foreach .a as $v0 (.; $v0; (.zzz | $v0))) = 999'
jq: error (at <stdin>:1): Invalid path expression with result {"b":1}

$ echo '{"a":{"b":1}}' | succinctly jq -c '(foreach .a as $v0 (.; $v0; (.zzz | $v0))) = 999'
{"a":999}
```

The same shape with a destructuring pattern (only reachable after #2676 widened
`resolve_node_sink`'s dispatch guard — refused unconditionally before that, so this specific
row is new *reachable* surface, not a new *bug*):

```console
$ echo '{"a":[{"b":1}],"x":{"a":[{"b":1}]}}' \
  | jq -c 'path(foreach .a as [$v0] (.; $v0; (.x | (.a | $v0))))'
jq: error (at <stdin>:1): Invalid path expression with result {"b":1}

$ echo '{"a":[{"b":1}],"x":{"a":[{"b":1}]}}' \
  | succinctly jq -c 'path(foreach .a as [$v0] (.; $v0; (.x | (.a | $v0))))'
["a",0]
```

## What's confirmed pre-existing vs. what #2676 changes

- The **bare-`$var` repro above needs no destructuring pattern** and reproduces identically on
  a `main` build from before #2676's first commit — confirmed with `git worktree`/build
  comparison. #2676 does not touch `resolve_foreach`'s EXTRACT-resolution machinery
  (`FoldRegister::advance`, `FoldRegister::resolve`, the `resolve_seq` pipe-carried-register
  path) at all.
- The plain, non-fold `#2042` bind mechanism (`. as $v0 | BODY`) does **not** have this bug —
  confirmed live: `path(.a as $v0 | ($v0.zzz | $v0))` on `{"a":{"b":1}}` correctly refuses on
  both jq and succinctly, wording included. So the defect is specific to `resolve_foreach`'s own
  UPDATE→EXTRACT chain (`active_reg.advance(update_branch, ...)` then a *second*, separate
  `extract_reg.resolve(...)` call for EXTRACT), not the general `Frame`/`Origin::At` machinery
  #2042/#2649 share.
- #2676 gives destructured loop variables (`as {c:$x}`, `as [$x]`) a tracked marker for the
  first time via `apply_pattern_bindings`, spliced into both UPDATE and EXTRACT the same way
  the pre-existing bare-`$var` splice (`substitute_var_tracked`) already does — so it inherits
  this exact defect for a class of programs that could not reach `resolve_foreach` before
  (destructuring patterns fell through to the untracked `resolve_leaf` catch-all
  unconditionally, pre-#2676, which cannot fabricate).

## Suspected root cause (not fully traced to a fix)

`FoldRegister::resolve`'s `Expr::Pipe` arm dispatches EXTRACT through `resolve_seq` with the
step's own value threaded in as an *externally carried register*
(`self.trackable.then_some(&self.value)`, #2046). A pipe stage that navigates through a
**missing object key** or **null-absorption** step (`.zzz` on an object without that key, or
`.foo` on `null`) appears to leave that stage's result *untracked* rather than correctly
tracked-at-a-new-position (jq's own model tracks a missing-key access as a genuine, if
null-valued, navigation step — `path({} | .foo)` is `["foo"]` in real jq, confirmed live) — and
then a later `$var` reference in the *same* pipe re-certifies against the **original** carried
register/position rather than the (correctly tracked, but different) position the intervening
stage actually reached, because nothing invalidates the carried register across that stage. This
is speculative; whoever picks this up should trace `resolve_seq`'s per-stage frame threading
and `resolve_seq_stage`'s handling of a missing-key/null-absorbing `Field`/`Index` step before
committing to a fix, and verify against real jq's own missing-key path-tracking behavior
(`path({} | .foo)` = `["foo"]`) as part of it.

## Related

- [#2676](https://github.com/rust-works/succinctly/issues/2676) — where this was found (its own
  code-review's differential fuzz + adversarial probing), and the PR that gives destructured
  loop variables new reachability into this pre-existing mechanism.
- `FoldRegister::advance`/`FoldRegister::resolve` (`src/jq/eval.rs`) — the two-call
  UPDATE→EXTRACT chain this lives in.
- `resolve_seq`'s carried-register mechanism, #2046 — the machinery whose interaction with a
  missing-key/null-absorbing stage is the likely culprit.

ADR-0018: jq mode only (real yq's lexer rejects `reduce`/`foreach`/`path` outright).

newhoggy:
## Summary
- **High severity, confirmed write corruption.** `FoldRegister::resolve`'s epilogue calls `relocate()` unconditionally on every UPDATE/EXTRACT branch, and `relocate`'s `identical()` fallback re-derives trackability from `resolve()`'s own *entry-time* register by nothing more than a value-equality check — a tautology whenever the branch's final value is the loop variable's own frozen value, which it always is when `$var` is referenced anywhere in the pipe. A missing-key/OOB step before a later `$var` reference silently re-certified even though `resolve_seq`'s own finer-grained tracking had already, correctly, marked that branch untracked:
  ```
  $ echo '{"a":{"b":1}}' | jq -c '(foreach .a as $v0 (.; $v0; (.zzz | $v0))) = 999'
  jq: error (at <stdin>:1): Invalid path expression with result {"b":1}
  $ echo '{"a":{"b":1}}' | succinctly jq -c '(foreach .a as $v0 (.; $v0; (.zzz | $v0))) = 999'
  {"a":999}   # write corruption -- real jq refuses this write entirely
  ```
- Fixed by gating `identical()` on `cannot_move_register(expr)` — the same allowlist `FoldRegister::advance` already uses for its own analogous carry-forward (#2046). `resolve_reduce`'s own final-emission call site (which builds its branch directly from the fold's accumulator, not from a `resolve()` call) keeps `identical()` unconditionally available, unaffected.
- `cannot_move_register` gained a `Reduce`/`Foreach` arm (recursing into `input`/`init`/`update`/`extract`, refusing a destructuring pattern) instead of treating every nested fold as "assumed moved" — needed because the existing test suite pins a nested-fold snapshot-passthrough case the narrower gate alone would have wrongly refused. This is a pure widening in the same direction #2042/#2649 already widened `Expr::As`/`Expr::AsPattern`, and it incidentally also lets `resolve_seq`'s own plain-pipe carrying (used far beyond fold UPDATE/EXTRACT) correctly accept a filter the existing test suite had pinned as refuse-only — moved to the accepting matrix, confirmed live against jq 1.7.1.

Fixes #2860

## Verification
- A background investigation agent traced the mechanism precisely (file:line citations) and corrected the issue's own speculative root-cause theory — the actual defect is in `FoldRegister::relocate`, one layer downstream of where the issue pointed.
- Interleaved differential sweep (pre-fix vs post-fix binary, 63 filter/input combinations across `reduce`/`foreach`/plain `as`-binding shapes): 4 mismatches, every one a confirmed correction (pre-fix wrongly accepted, post-fix matches jq's refusal exactly), 0 unintended regressions.
- Two additional regressions surfaced during that verification and fixed in the same commit (both existing test-suite pins, not new findings): a nested-fold snapshot-passthrough case, and one filter that moved from "refuse-only" to correctly "accepting" (both re-verified live against `/usr/bin/jq` 1.7.1).

## Test plan
- [x] New tests pinning the primary bug (`path()`, `=`, `|=`, `del()`, array-OOB variant, `if`/`try`-wrapped variants, present-key variant) — all previously wrongly accepted/corrupted, all now correctly refuse matching jq's exact message.
- [x] Must-not-regress controls: `null`-source (genuine reestablishment via jq's own `null`-identity rule), non-navigating-literal shape, destructuring-pattern shape.
- [x] All existing `#1466`/`#2042`/`#2046`/`#2031`/`#2161`/`#2632`/`#2388` register-tracking regression tests pass unmodified (except the one row moved between matrices as described above, an intentional correction per that test's own self-checking design).


newhoggy:
Ran the mandatory 4-angle code review. Given the depth of this fix (shared register/path-tracking machinery), took the findings very seriously.

**Confirmed and fixed, real regression**: `cannot_move_register`'s gate is syntactic — it requires every branch of an `if`/`try` to be navigation-free, not only the one actually taken. Gating `identical()` on it cost a refusal whenever a navigating branch sat *unreached* alongside a safe one: `path(. as $x | foreach (1,2) as $i (0; (if false then .a else $x end); .)))` is jq's `[]` twice, but post-fix succinctly refused it. Verified live before and after against `/usr/bin/jq` 1.7.1.

I traced this carefully (including whether `PathBranch`'s own `register` field could supply the missing runtime signal — it's populated only within `resolve_seq`'s internal per-stage loop and isn't preserved on the final returned branch, so it isn't usable at `relocate()`'s call site without deeper redesign). A properly general fix needs per-branch information about which control-flow path was actually taken, which isn't available where `relocate()` runs today. Given:
- the regression is strictly refuse-only (a write fails rather than corrupting data — the established safe direction throughout this exact subsystem's multi-year history: #1466, #2042, #2046, #2049, #1573, #2044, #2649),
- it requires a specifically contrived nested-fold-in-an-untaken-branch shape to trigger,
- and this is documented, precedented territory (#2046 already accepts the identical "scope limit, deliberately not closed" for a bare `$var` inside `if`/`try` with no register threading at all — this is that same limit reached through a different door),

I documented it as a known residual (matching the file's own established pattern for this exact class of trade-off) and pinned it with a regression test, rather than attempting a riskier architectural redesign under review-response time pressure.

**Also fixed** (reuse finding): the destructuring-pattern check was duplicated three times across `AsPattern`/`Reduce`/`Foreach` — factored into one `patterns_all_bare` helper.

**Also fixed** (doc nit): `cannot_move_register`'s own doc comment now names all three of its call sites (`resolve_seq`'s carrying, `advance`'s carry-forward, and the new `identical_eligible` gate) — a third, pre-existing caller wasn't mentioned before. Two independent review agents differentially fuzzed this third call site and the `advance` call site specifically for divergence from the widening; both came back clean.

**Efficiency finding acknowledged, not fixed here**: `cannot_move_register(expr)` is recomputed on every fold element inside `resolve()`, though its result is invariant across elements (substitution only ever produces `Literal`/`TrackedVar` leaves, both unconditionally `true`). Hoisting this would need threading a precomputed bool through `resolve()`'s signature and both `resolve_reduce`/`resolve_foreach` call sites — a real but non-blocking perf optimization, deferred rather than bundled into an already-large correctness fix.

All existing + new tests pass (21 for this area, full suite green), clippy/fmt/docs clean.

PR #2894, opened 2026-09-13
perf(jq): make bare ../recurse/recurse_down lazy under a bound (#2696)

### LINKED ISSUE #2696: jq: bare `..`/`recurse` predates the #2235 sink migration, still eager under a bound
**Severity: Low** (efficiency gap under a bound, not incorrect output — mirrors #2190's own
shape one level down)

## Summary

#2235 migrated `recurse(f)`/`recurse(f; cond)` to a demand-driven `resolve_recurse_sink`,
so `limit(1; recurse(f))` now visits only what it asks for instead of building the whole
visited set up front. Bare `..`/`recurse` — same construct, no arguments — was left out of
that migration: it still goes through the pre-#2235 `resolve_recursive_descent`/
`push_recursive_branches`, which builds a full `Vec<PathBranch>` for the entire document
tree regardless of any wrapping bound.

Found during review of #2685 (the fix for #2190). Not yet in
`docs/compliance/jq/limitations.md` since output is identical to jq — this is purely an
internal efficiency gap, not a behavioral divergence — but it undermines, one commit away,
the exact "a bounded consumer no longer pays for nodes past the ones it actually asked
for" guarantee #2235/#2685 shipped for the parameterised spelling.

## Repro shape

`path(limit(1; ..))` or `first(..)` on a large/deep document still does O(document size)
work — visiting and buffering every node — before `resolve_limit_sink` gets a chance to
stop it, where `path(limit(1; recurse(f)))` (the parameterised sibling) now stops
immediately after the first node.

## Fix direction

Give `resolve_recursive_descent`/`push_recursive_branches` the same sink treatment
`resolve_recurse_sink` already has — likely the smaller of the two remaining migration
gaps (#2693, #2694 are the other two), since `..`/`recurse`'s traversal has no `f`/`cond`
to evaluate per node, only the DFS-stack bookkeeping `resolve_recurse_sink` already
demonstrates the pattern for. Needs a benchmark isolating the fix the way #2190's own PR
(#2685) did, plus the usual live-oracle differential pass against jq 1.7.1 (ADR-0018).

newhoggy:
## Summary
- Bare `..`/bare `recurse`/`recurse_down` predates #2235's migration of the parameterised `recurse(f)`/`recurse(f;cond)` pair onto a sink, so a bounded consumer (`limit`, `first`, an enclosing fold's own early stop) still paid to collect the whole tree (`push_recursive_branches`) before ever seeing the first branch, unlike its already-lazy sibling.
- Adds `resolve_recursive_descent_sink`, a stack-based sink-shaped walk mirroring `resolve_recurse_sink`'s own shape (self before children, one explicit stack instead of native recursion) but with no `f`/`cond` to evaluate and **no `RECURSE_MAX_ITEMS` cap** — structural descent's node count is exactly the document's own tree size, so `[path(..)] | length` must answer jq's true count even past that constant.
- Folds the recurse-family untracked guard into one arm in `resolve_node_sink` covering all five spellings (CLAUDE.md's "one predicate" rule), removing the now-unreachable guard and handling arms from `resolve_node_eager` along with `push_recursive_branches`/`resolve_recursive_descent` themselves.
- Adds a benchmark group (`jq_write_path_recurse_family_bounded`) and CLI/unit tests.
- Updates `docs/compliance/jq/limitations.md` to close out the residual this issue tracked.

Closes #2696.

## Performance
Single-machine (Apple M5 Max) sanity measurement on a 10MB generated document — not the full two-architecture interleaved A/B, since the issue's own triage plan judged the mechanism unambiguous (buffer the tree vs. don't) rather than a micro-optimization needing it:

| query | wall (before → after) | peak RSS (before → after) |
|---|---:|---:|
| `path(limit(1; ..))` | 0.19s → 0.12s | 573 MB → 190 MB |
| `[path(..)] \| length` (unbounded holdout) | ~1.2s → ~1.2s (noise) | ~2.1 GB → ~2.1 GB (noise) |

190 MB matches `recurse(.[]?)`'s own floor. Output verified identical to jq 1.7.1 in both rows, including past the 10,000-node `RECURSE_MAX_ITEMS` constant (12,884-node test document, exact count agreement).

## Test plan
- [x] `cargo test --features cli,simd,regex,serde` (full suite, before and after rebase)
- [x] `cargo bench --bench jq_write_path_bench -- recurse_family_bounded --test` (smoke run)
- [x] Verified live against jq 1.7.1: order/output identity for `..`/`recurse`, bounded-consumer agreement (`limit`, `first`, `nth`, `del(first(..))`, `select`), and no truncation past `RECURSE_MAX_ITEMS`
- [x] Retargeted the #1021 depth-panic unit test at the new sink; added a demand-count laziness pin (`resolve_recursive_descent_sink_stops_after_first_demand_2696`)


newhoggy:
Ran a large multi-agent review (9 finder passes across correctness, reuse, simplification, efficiency, altitude, and conventions angles). Summary:

**No correctness bugs found** — order, depth-guard placement, snapshot/trackable threading, and the \`Cow::Borrowed\` invariant all checked out across multiple independent line-by-line/cross-file passes, corroborated by live oracle probing (deep mixed trees, \`del(.. | select(...))\`, in-place \`|=\` through recurse-derived paths, snapshot-through-\`as\`-binding).

**Fixed:**
- **Reuse** (confirmed by 3 independent agents): the root-seed construction was duplicated verbatim between \`resolve_recursive_descent_sink\` and \`resolve_recurse_sink\`. Extracted a shared \`recurse_family_root_seed\` helper.
- **Simplification** (confirmed by 2 agents): reordered per-node delivery to call \`sink()\` with a cheap \`current.clone()\` before unwrapping the \`Cow\`, matching the sibling's simpler idiom and deferring the \`unreachable!()\`-guarded unwrap to right before it's actually needed.
- Added the same \`debug_assert!\` the sibling carries for the trackable-or-snapshot invariant (the #843 "an invariant that only lives in a comment is one a later edit can quietly break" lesson).
- **Doc accuracy**: corrected the laziness claim — children of one popped node are still fully expanded onto the stack before the loop re-checks \`Demand::Stop\`, so laziness holds *between* levels, not *within* one node's fan-out. \`path(limit(1; ..))\` (what the benchmark measures) is unaffected; \`path(limit(2; ..))\` over a very wide single node still pays for that node's whole child set. Filed as #2895 for a proper per-child-cursor fix, rather than overclaiming full laziness.
- Fixed stale \`push_recursive_branches\` references in benches/docs/tests outside \`eval.rs\` (already renamed there) and a doc comment misattributing dispatch rationale to \`resolve_node\` instead of \`resolve_node_sink\`.

**Considered and declined:** a \`stack.reserve()\` hint at the fan-out sites — this repo's clippy config disallows infallible \`Vec::reserve\` (#1670, needs \`try_reserve\` + an escape path), and wiring a new fallible-allocation escape through \`ResolveFlow\` for an explicitly-optional "missed opportunity" finding was judged out of scope here.

**Left as-is (defensible, not a violation):** one reviewer raised whether \`resolve_recursive_descent_sink\`/\`resolve_recurse_sink\`'s remaining structural duplication (the full pop/deliver/push-children skeleton, not just the seed) should be unified into one generic walker. Given the two functions differ meaningfully in what varies per node (structural descent vs. an arbitrary \`f\`/\`cond\` evaluation) and in their invariants (depth-cap vs. item-cap), full unification is a larger design decision than this fix's scope — noted in #2895 as a secondary consideration for whoever picks that up, not attempted here.

Full test suite + both clippy variants + rustdoc all green after fixes and after two rebases onto a fast-moving \`main\` (one required real conflict resolution — both sides had appended new tests at the same point in \`tests/jq_cli_tests.rs\`; resolved by keeping both test sets intact).

Issue #2896, opened 2026-09-13
jq: foreach + getpath's pointer-aliased position isn't modeled, EXTRACT can't continue from it


newhoggy:
**Severity: Low** (message/refusal-only in the read direction confirmed so far; a write-side
repro was not attempted as part of this filing)

## Summary

Surfaced while triaging #2732 (Shape 3 there, explicitly out of that issue's own scope).

Real jq's `getpath` is backed by `_jq_path_append` (a C function), which silently returns its
argument value *unchanged* when the input is not itself intact — and the value it returns is
*pointer-identical* to `value_at_path` left by whatever navigated there before it. So inside a
`foreach`'s EXTRACT, a `getpath(...)` step transparently continues from SOURCE's own tracked
position, and a later step in the same EXTRACT sees jq's real path register exactly as if
`getpath` had never run:

```
$ echo '{"a":{"b":2}}' | jq -c 'path(foreach .[] as $k (.; getpath(["a"]); .b))'
["a","b"]
$ echo '{"a":{"b":2}}' | succinctly jq -c 'path(foreach .[] as $k (.; getpath(["a"]); .b))'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "b" of {"b":2}
```

The `reduce` spelling of the identical shape already agrees with jq (`with result 1`, the
persistent-register position restored by `reduce`'s own final re-entry check) — this is
specific to `foreach`'s per-step register plus `getpath` in particular.

## Root cause

`getpath` returns a *value*, not a tracked path in this evaluator's own model
(`PathBranch::trackable`/`snapshot`) — nothing keys `getpath`'s result back to the position it
was called from unless `getpath`'s own argument is itself literally a tracked navigation
component list. jq's C-level pointer identity has no analogue in an owned-value model; closing
this would mean `getpath` growing its own register-threading (recognizing when its own result
happens to alias the position it read from), unlike the ordinary `Field`/`Index` steps every
other fold-register divergence in `docs/compliance/jq/limitations.md` is about.

## Scope

Filed as its own issue per #2732's own triage plan ("file a separate `jq:`-prefixed issue...
and record it in `limitations.md`; do not chase it here"). No open PR references this issue.

newhoggy:
## Triage plan (2026-09-13)

Probed live against the pinned oracle (`/usr/bin/jq` 1.7.1) and `target/release/succinctly`
built at `db974b52e`. Every command below is reproducible as written.

[...]

PR #2898, opened 2026-09-13
fix(jq): reestablish a null/bool fold source against an equal register

### LINKED ISSUE #2732: jq: resolve_reduce's residual 32 path() divergences (of #2632's 728) — unclassified
**Severity: Low**

## Summary

#2632 (fixed in PR #2721) gave `resolve_reduce` the same register-identity widening
#2161 gave `resolve_foreach`, measured on a generated 12,048-case sweep (4 source
streams × 7 INITs × 11 UPDATEs × 12 documents), classified against jq 1.7.1:

| | count |
|---|---|
| `reduce` divergences before | 728 |
| fixed by the widening | 696 |
| broken by it | 0 |
| still diverging after | **32** |

#2632's own text flagged these 32 explicitly as not investigated: "they may be a
third, unrelated shape. Worth classifying before assuming this fix closes them." This
issue tracks that follow-up — it was out of scope for #2632 itself, which already had
a large, self-contained fix (root cause pinpointed, prototype pre-measured, its own
write-side yq-mode hazard).

## Root cause

Not yet investigated. Candidate starting point: re-run #2632's own sweep harness
(referenced in that issue and PR #2630/#2721) against current `main`, isolate the 32
residual cases, and classify whether they're a single shared mechanism or several
unrelated ones.

## Why this is separate from #2632

#2632's own scope was specifically the register-identity widening (jv_identical vs.
path-comparison) that #2161 already validated as the right mechanism for `foreach`.
The residual 32 may need a different mechanism entirely — bundling them into #2632
would have risked scope creep on an already well-defined fix.

## Suggested fix direction

Classify the 32 residual cases first (see #2632's sweep harness) before proposing a
fix — per this codebase's own benchmarking/verification discipline, a "surprising"
residual should be checked against the oracle and categorized, not assumed to share
one root cause.

## Related

- #2632 / PR #2721 — the register-identity widening and its sweep harness.
- #2161 / PR #2630 — the `foreach` counterpart.

newhoggy:
## Summary
- `fold_source_ambient`'s fork-0 arm decided SOURCE's own trackability with a pure path comparison (`branch_provenance`), never consulting the `null`/`bool` kind-equality rule (`register_identical`'s `null_bool_identical` clause) the *later*-fork arm of the same function already ORs in. A `null`/`bool` document is `jv_identical` to a `null`/`bool` register in real jq regardless of position:
  ```
  $ echo null | jq -c 'path(reduce .a as $k (.b; .))'
  ["b"]
  $ echo null | succinctly jq -c 'path(reduce .a as $k (.b; .))'
  jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of null
  ```
  `foreach` shares the identical gap since it calls the same function.
- Fixed by ORing `register_identical` into fork 0's `source_trackable`, gated jq-mode only (real yq's lexer rejects `reduce`/`foreach`/`path` outright, and a widened acceptance would be the wrong direction there per #2161/#2632's own precedent for this exact gate).
- Also documents (not fixed, per this issue's own triage) two residuals classified during the investigation:
  - `reduce`'s mid-step register position is message-only wrong (exit codes and values already agree with jq — Shape 2).
  - `foreach` + `getpath`'s pointer-aliased position isn't modeled (Shape 3, filed separately as #2896, out of this issue's scope).

Fixes #2732

## Verification
- All 5 rows from the issue's own repro table (`path()`, `=`, `.[]?`/`.[]` distinction) confirmed live against `/usr/bin/jq` 1.7.1, now matching exactly.
- Full differential sweep against a merge-base binary: 3,696 `reduce` programs (4 sources × 7 INITs × 11 UPDATEs × 12 documents) and 6,720 `foreach` programs (4 sources × 7 INITs × 6 UPDATEs × 4 EXTRACTs × 10 documents) — 140 changed rows total, every one now matching `/usr/bin/jq` 1.7.1 exactly, zero unintended divergences.
- Negative controls: an object/array document is never promoted (kind-equality only fires for `null`/`true`/`false`), a `true` document's own type error is unaffected, yq mode's identical write-side filter still refuses.

## Test plan
- [x] New tests pinning the primary fix (`path()`, `=`, `.[]?` acceptance; `.[]` unsuppressed iteration error kept distinct from the path-tracking fix).
- [x] Must-not-regress test (object/array document, `true` document).
- [x] Residual-pinning tests for Shape 2 (message-only) and the yq-mode gate.


newhoggy:
Ran the mandatory code review (6 angles). No correctness bugs found — the fix's own 10,000+-program differential sweep (3,696 `reduce` + 6,720 `foreach` combinations against a merge-base binary and `/usr/bin/jq` 1.7.1) already ruled that out, and every review angle's own live probes confirmed the same.

Several documentation-precision issues were confirmed and fixed:
- `relocate_base`'s own doc comment stated an invariant this fix's null/bool clause falsifies ("a trackable first-fork ambient implies `reg_path` is already root") — the function's actual condition already handles both origins of `trackable` correctly, only the comment was stale. Fixed.
- The new comment claimed only `register_identical`'s null/bool clause was consulted, when the call site correctly uses the full function (modeling jq's real `jv_identical`, not an over-widening). Tightened.
- The "message-only" residual claim (for `reduce`'s mid-step position) overclaimed "every value agrees" — under `try`/`catch`, the error message's own *embedded* value can differ, even though neither side ever produces a successful output. Verified live and tightened the wording in both the code comment and `limitations.md`.
- The "real yq's lexer rejects `reduce`/`foreach`/`path` outright" gate rationale was accurate for real yq but implied succinctly's own yq mode does too — it doesn't for `reduce`/`foreach` themselves (only `path(...)` call syntax is gated), confirmed live via an assignment-context repro. Fixed.
- Softened Shape 3's internal-mechanism explanation (`_jq_path_append`, pointer identity) to a hypothesis, since only the observable jq behavior was independently verified live, not jq's own C source.

One asymmetry was documented rather than changed: the pre-existing `fork != 0` arm applies the same reestablishment ungated on `S::TAG`, unlike fork 0's new jq-only gate. Left as-is (widening or narrowing it is outside this fix's own scope and touches pre-existing, unrelated code), with a comment explaining the asymmetry for a future audit.

Filed #2901 for the reuse finding (the `register_identical`-based reestablishment check is hand-inlined at three call sites with no shared helper) as its own follow-up rather than a refactor bundled into this fix, given how heavily-iterated-on and fragile this exact machinery already is (#1466, #2042, #2044, #2046, #2049, #1573, #2388, #2632, #2649).

All existing + new tests pass (20 for this area, full suite green), clippy/fmt/docs clean.

Issue #2899, opened 2026-09-13
jq: reduce's own INIT position still materializes a ?// bind, hiding a consumer's stop (#2668 residual)


newhoggy:
**Severity: Low** (surprising output cardinality; the same class as #2180/#2668, no data loss)

Residual left by #2668, recorded in `docs/compliance/jq/limitations.md`.

#2668 gave `foreach`'s INIT position a demand-forwarding drive (`ForeachInitDrive`, threaded through `foreach_forks`), closing the identical class of divergence #2180 WP3 already closed for `foreach`'s source, pattern, EXTRACT and UPDATE. `reduce` shares `foreach`'s UPDATE fix (`fold_step_each`) since it was reshaped as one shared dispatch, but `reduce`'s own **INIT** position was explicitly left open, and has the same bug:

| filter | jq 1.7.1 | succinctly jq |
|---|---|---|
| `[first(reduce (1) as $v ((1 as $x ?// $y \| 1); . + 1))]` | `[2,2]` | `[2]` |
| `[first(reduce (1) as $v (0; . + (1 as $x ?// $y \| 1)))]` (UPDATE, control — already matches) | `[1]` | `[1]` |

Confirmed live against jq 1.7.1 and pinned as a CLI row in `test_nested_short_circuit_consumer_hides_the_stop_2180` (`tests/jq_cli_tests.rs`).

**Why #2668 didn't close this too.** Unlike `foreach`, `reduce` has **no native, demand-forwarding dispatch arm at all**: neither `eval_each` (`src/jq/eval.rs`) nor `eval_each_generic` (`src/jq/eval_generic.rs`) has an `Expr::Reduce` case. Every `reduce` evaluation — including one wrapped in `first(...)`/`limit(...)`/etc. — reaches `eval_reduce`/`eval_reduce_with_values` through the fully-eager, collecting `eval_single`/`QueryResult` path, then whatever wrapper is watching drains the (already-complete) result afterward. Confirmed live: `first(reduce (1,2,("X"|stderr),4) as $x (0; .+$x))` writes `X` in *both* jq and succinctly, since neither tool ever gets to stop `reduce`'s own source early for a plain (non-`?//`) reduce — this is not itself a bug, it establishes that `reduce`'s output position is not, and was never meant to be, demand-forwarding by default.

[...]

Issue #2440, opened 2026-09-04
jq: foreach/reduce evaluate INIT only after the first SOURCE output, so a SOURCE that errors first hides INIT's own error/halt (jq evaluates INIT first)


newhoggy:
Surfaced by spine 2416's phase 3 while re-pinning `test_path_context_escape_routes_pinned_2216`; the divergence itself is in the value evaluator and predates the migration (the #2216 report noted the `error("init")` case).

## Capture (jq 1.7.1, input `{"a":1}`)

```console
$ jq -c 'foreach (1, error("in")) as $x (error("init"); .)'
jq: error (at <stdin>:1): init              # rc 5
$ succinctly jq -c 'foreach (1, error("in")) as $x (error("init"); .)'
jq: error (at <stdin>:1): in                # rc 5   <-- SOURCE's error, INIT never ran

$ jq -c 'reduce (1, error("in")) as $x (error("init"); .)'
jq: error (at <stdin>:1): init

$ jq -c 'foreach (1, error("in")) as $x (halt_error; .)'
{"a":1}                                     # rc 5: INIT's halt_error
$ succinctly jq -c 'foreach (1, error("in")) as $x (halt_error; .)'
{"a":1}                                     # rc 5: matches -- but only because the first item arrived before INIT ran
```

jq evaluates INIT before pulling SOURCE, so INIT's escape always wins. succinctly's `eval_foreach_with_values`/`eval_reduce_with_values` pull the first SOURCE item first and evaluate INIT after it, so:

- a SOURCE whose *first* output is an error reports that error and never evaluates INIT (`foreach (error("in")) as $x (halt_error; .)` reports `in`; jq halts);
- when SOURCE yields a value before erroring, INIT runs and its halt wins, matching jq by coincidence of ordering.

The path-context twin of this pin is `test_path_context_escape_routes_pinned_2216` in `src/jq/eval.rs`, where `foreach (key, error("in")) as $x (halt_error; .)` now reads `error:in` because `key` at the root emits nothing (#2421) and the source's first output is therefore its error.

## Ask

[...]

PR #2900, opened 2026-09-13
fix(jq): drive foreach's UPDATE and INIT through demand-forwarding sinks

### LINKED ISSUE #2668: jq: foreach's UPDATE and INIT positions still materialize a ?// bind, hiding a consumer's stop (#2180 residual)
**Severity: Low** (surprising output cardinality; the same class as #2180, no data loss)

Residual left by #2180 WP3, recorded in `docs/compliance/jq/limitations.md` under "A `?//`-alternatives bind sees a short-circuiting consumer's stop only when nothing materializes it first" and pinned in `test_nested_short_circuit_consumer_hides_the_stop_2180` (`tests/jq_cli_tests.rs`).

#2180 gave `foreach` a demand-forwarding arm for its **source**, its **pattern** `?//` and its **EXTRACT**, but UPDATE and INIT are still evaluated eagerly, so a `?//` bind sitting in either position still answers once where jq answers twice. Confirmed live against jq 1.7.1 (`-cn`):

|---|---|---|
| `[first(foreach (1) as $v (0; . + (1 as $x ?// $y \| 1)))]` (UPDATE) | `[1,1]` | `[1]` |
| `[first(foreach (1) as $v ((1 as $x ?// $y \| 1); . + 1))]` (INIT) | `[2,2]` | `[2]` |
| `first(foreach (1) as $x (0; (.+1, ("U"\|stderr)); .))` (UPDATE side effect) | `1`, no stderr | `1`, writes `U` |

**Why they were left.** UPDATE's outputs go through `fold_step_via_accumulator_or_fork` (`src/jq/eval.rs`), the helper `reduce`'s O(n) accumulator fix (#2157) shares, and whose outputs are simultaneously the fold's next state; making it demand-driven means reshaping it into a sink-shaped fold (`fold_step_each(expr, state, optional, &mut |update_val| ...)`) with `reduce` wrapping it in a collecting sink — a change with real regression surface on a perf-sensitive path, so it was split out. INIT is jq's outermost loop (#534) and must be evaluated before the source is pulled (#2440), so a lazy INIT needs the fork loop itself to become a sink over INIT's outputs.

**Fix criterion** is the same one #2180 used throughout: give the position a demand-forwarding arm and its rows in the characterization test flip to jq's answers; `scripts/jq-alt-retry-oracle-sweep.sh` deliberately excludes these two positions so it keeps its 0-unexpected/0-known contract — add them back to `W_ENTRIES` as part of the fix.

The `/code-review` pass on WP3 estimated the UPDATE reshaping at roughly thirty lines and noted it also removes the accumulator clone on the implicit-EXTRACT push that #2157's doc comment calls "structurally unavoidable".

newhoggy:
## Summary

- Reshapes `fold_step_via_accumulator_or_fork` into `fold_step_each`, a sink-shaped helper delivering each UPDATE output to a callback as it is produced instead of collecting into a `Vec` first. `try_foreach_step_alternatives` now runs EXTRACT (or the implicit identity push) from inside that callback, so a consumer's `Demand::Stop` reaches UPDATE's own generator — and any `?//` bind inside it — before it produces anything further. `try_reduce_step_alternatives` wraps the same helper in a collecting-last sink (reduce's own UPDATE has no per-step visible output for a consumer to stop after, confirmed unaffected).
- Reshapes `foreach_forks` to take a `ForeachInitDrive` closure (mirroring its existing `ForeachSourceDrive`) instead of a pre-collected `Vec<OwnedValue>` of INIT's outputs, so a consumer's stop reaches INIT's own generator too — `#2440`'s "zero INIT outputs ⇒ source never pulled" now falls out structurally instead of needing its own guard.
- All four `foreach_forks` call sites (`eval_foreach`/`each_foreach` in `eval.rs`, the eager `Expr::Foreach` arm and `each_foreach_generic` in `eval_generic.rs`) build `drive_init` from `eval_each`/`eval_each_generic`, mirroring how they already build `drive_source`; `eval_foreach` keeps an eager fallback for an unbounded INIT, gated identically to its existing source-side one.
- `reduce`'s own INIT has the identical bug but is **not** fixed here: `reduce` has no native, demand-forwarding dispatch arm at all (`first(reduce(...))` always reaches `eval_reduce` through the fully-eager, collecting `eval_single` path), so nothing ever drives it through a live sink regardless of this reshape. Documented as a residual and filed as a follow-up: #2899.
- Adds the `foreach-update`/`foreach-init` rows back to `jq-alt-retry-oracle-sweep.sh`'s `W_ENTRIES` (861 cases, 0 unexpected, 0 known) and flips the characterization rows in `tests/jq_cli_tests.rs` to jq's answers.

Fixes #2668

## Test plan
- [x] `cargo test --features cli` (full suite, 1699 jq CLI tests + 3746 lib tests) — all green
- [x] `cargo test --no-default-features --verbose` (no_std) — clean
- [x] `./scripts/jq-alt-retry-oracle-sweep.sh` — 861 cases, 0 unexpected, 0 known
- [x] Interleaved A/B (`scripts/ab-cli.py`) against pre-fix `main` on `reduce`/`foreach` accumulator and INIT-fan-out shapes over a 10MB corpus — 0 output differences, timing within noise (median of medians +0.71%, range -2.3%..+1.6%)
- [x] Live-verified every characterization row against the pinned oracle (`/usr/bin/jq` 1.7.1)


newhoggy:
## Code review addressed

8-angle review found 5 real findings, 4 fixed here, 1 deferred with a follow-up:

- **Fixed**: \`foreach_forks\`'s \`terminal\` match arm bypassed \`finish_fork_flow\`'s suppression check for a fork's own escape (asymmetric vs. the sibling \`init_flow\` arm; currently unreachable but a real contract violation) — now routes through it and reuses \`stop_with_downstream\` for both \`ended\`/\`terminal\`.
- **Fixed**: the INIT-driving closures I added (\`eval_foreach\`'s \`lazy_drive_init\`/\`eager_drive_init\`, \`each_foreach\`'s \`drive_init\`) were near-identical copies of the pre-existing source-driving ones — factored into two shared functions, \`drive_lazy\`/\`drive_eager\`, parameterized on the \`Expr\`.
- **Fixed**: the generic-evaluator twin (\`drive_foreach_source_generic\`/\`drive_foreach_init_generic\`) had the identical duplication — collapsed into one \`drive_foreach_expr_generic\`.
- **Fixed**: \`pattern_alternatives_var_names\` now computed lazily on the first successful INIT fork instead of unconditionally, restoring the short-circuit a zero-output INIT used to get for free.
- **Deferred, filed as #2903**: \`resolve_foreach\` (path-mode \`foreach\`, reached via \`path(foreach(...))\`) has its own separate fold loop that this PR didn't touch — an explicitly anticipated risk in #2668's own triage plan, which checked it doesn't call \`foreach_forks\` and left it alone. Before this PR both evaluators collected INIT eagerly and agreed with each other; this PR fixed only value-mode, so path-mode now disagrees with value-mode too (on top of its own unchanged divergence from jq). Documented in \`limitations.md\` and pinned as a new row in \`test_short_circuit_side_effect_leaks_820_932_987\`.

Full suite (3746 lib + 1699 jq_cli + 1902 yq_cli + doctests), clippy, fmt, no_std, rustdoc, and the oracle sweep (861/0/0) all re-verified clean after the fixes.

Issue #2666, opened 2026-09-09
map(f) | .[] loses jq's array-construction atomicity outside first/limit (#2103 regression)


newhoggy:
`map(f) | .[]` at the top level (no `first`/`limit`/`nth` wrapping) has lost jq's
array-construction atomicity as a side effect of #2103 (`fold_pipe_stages_sink` now
handles *every* top-level query, not just the M2-eligible ones the old
`expr_is_cursor_transparent` gate admitted).

```
$ echo '[1,"x",3]' | jq            -c 'map(.+1) | .[]'
jq: error (at <stdin>:1): string ("x") and number (1) cannot be added
$ echo '[1,"x",3]' | succinctly jq -c 'map(.+1) | .[]'
jq: error (at <stdin>:1): string ("x") and number (1) cannot be added
2
```

Real jq's `map(f)` is `[.[] | f]` — array construction is atomic, so a failing element
anywhere discards every element, even the ones before it. `succinctly jq` leaks the
already-computed prefix (`2`, from index 0) before reporting the error on index 1.

Confirmed this is new to #2103's branch: `main` at `ce691ca85` (the merge-base) gives
the correct, jq-matching output (error only, no leaked `2`) for the same command.

## Root cause

`fold_pipe_stages_sink` (`src/jq/eval_generic.rs`) unconditionally routes any
`Expr::Iterate` following a `map()`-produced `GenericResult::LazySeq` through
`each_lazy_seq_iterate_sink` — the demand-driven, non-atomic iteration #1565 built
*specifically* for `first`/`limit` early-exit (pulling one element at a time so a
truncating consumer never pays to evaluate elements past its stopping point). Its own
doc comment and `docs/compliance/jq/limitations.md` both state the resulting divergence
is "bounded to consumers that genuinely truncate" — anything that has to see the whole
array (bare `.[]`, `to_entries`, `last`, wrapping in `[...]`, etc.) was supposed to still
error atomically, and the doc gives worked examples showing exactly that.

That bound held only by accident: before #2103, an ordinary non-early-exit top-level
query like `map(.+1) | .[]` took the *eager* evaluator entirely
(`expr_is_cursor_transparent` gate), never reaching `fold_pipe_stages_sink`, so the
`#1565` fast path's scope was implicitly limited to `first(...)`/`limit(...)` bodies.
#2103 deleted that gate so every top-level query now goes through the generic
[...]

PR #2904, opened 2026-09-13
ci(perf-guard): add path-mode rows to the instruction-count matrix (#2655)

### LINKED ISSUE #2655: ci: perf-guard has no path-mode row, so a +21–33% resolver regression passed at −0.0% (#2042)
**Severity: Low** (a measurement gap, not a behaviour bug — but it is the gap that let #2042's first cut carry a +21–33% regression on a common `del`/`|=` shape past a perf-guard PASS at −0.0%)

## Summary

`scripts/perf-guard.py`'s fixed matrix is six queries: `.` identity on `wide`/`users`/`arrays`, `keys_unsorted` on `wide`/`users`, and one yq-mode `keys_unsorted`. None of them is a **path-mode** target — no `path()`, `del()`, `=`, `|=` — and none contains an `as` binding. The whole path resolver (`resolve_node` and everything under it in `src/jq/eval.rs`) is therefore outside what perf-guard can see.

Concretely, during #2042's A/B on the pinned boxes:

| what | perf-guard | interleaved A/B (11 reps) | cachegrind, 1 MB |
|---|---|---|---|
| `del(.users[] \| .score as $y \| select($y < 100))` | not in matrix | **+21%** 7950X / **+31%** M4 Pro | **+30.9%** Ir |
| all six matrix queries | PASS, −0.0% | — | −0.0% |

The regression was real, deterministic and constant-factor (identical at 1 MB and 10 MB); perf-guard was structurally incapable of noticing it. Its PASS was evidence for the untouched hot paths only.

## Proposal

Add path-mode rows to `QUERIES`, on the existing `users` 2 MB fixture so no new generator is needed:

- `del(.users[] | select(.score < 100))` — plain `del` target, no binding (the register/`Frame` machinery with `at = None`).
- `del(.users[] | .score as $y | select($y < 100))` — a target that **binds from a navigated position**, the shape #2042 widened.
- `(.users[] | .score) = 1` — an assignment target.
- `[path(.users[] | .age)] | length` — a bare `path()` walk.

Then regenerate the checked-in baselines on both pinned arches (`--update-baseline`, x86_64 on `terminus`, arm64 on the mini), since perf-guard fails on any drift — improvements included.

Cost: four more cachegrind runs per CI invocation, each on the order of a second.

## Context

Found while measuring #2042 (see the perf section of its PR). The regression itself is fixed there (the `select` condition's `Expr::TrackedVar` now takes the #2048 owned fast path instead of the reindex bridge); this issue is only about the guard.

newhoggy:
## Summary
- Every row in `scripts/perf-guard.py`'s `QUERIES` matrix read only — none exercised `resolve_node`/`resolve_node_sink` (`src/jq/eval.rs`) or anything beneath it, so a regression confined to that resolver (like #2042's own +21–33%) passed this guard at -0.0% no matter how large.
- Adds four jq-mode rows on the existing `users`/`2mb` fixture: `users_del_select` (a `del(select(...))` target), `users_del_bound_select` (a `del()` target bound through an `as` binding, matching #2042's own shape), `users_assign_scores` (an assignment target), `users_path_walk` (a bare `path()` walk).
- Adds one yq-mode row, `users_yq_del_select`: yq-mode writes take a different route (`evaluate_yaml_cursor`'s DOM path in `src/bin/succinctly/yq_runner.rs`) than jq-mode ones, and the matrix's only existing yq row (`users_yq_keys_unsorted`) is a read.
- No `QUERY_THRESHOLDS` overrides — the 5% default applies, matching #2666's own precedent for why a pre-emptive override is a permanent loosening of a row whose only job is to be watched.
- Regenerates both arch baselines in `tests/data/perf-guard-baseline.json`: x86_64 on `terminus` (7950X, valgrind 3.18.1), ARM64-Linux inside a native linux/arm64 Docker container on Apple Silicon (valgrind 3.22.0, matching `e97caf94d`'s established procedure and the `ubuntu-24.04-arm` image CI itself runs on). Both re-checked with `--check` immediately after: 0.0% drift on all 13 rows.
- Updates the module docstring and `--help` epilog to document the new row class.

Closes #2655.

## Verification
- All five new filters verified live against jq 1.7.1 / yq v4.53.3: exit 0, byte-identical (md5) output on the guard's own fixture (`succinctly json generate 2mb -p users -s 20260101`).
- **Negative test** (proves the new rows actually see the resolver, not just measure noise): a temporary 10,000-iteration `black_box` loop inserted at the top of `resolve_node_sink` reliably failed `users_del_select` (+30.4%), `users_del_bound_select` (+59.1%), and `users_yq_del_select` (+12.7%) while every read-only row (`users_identity`, `users_keys_unsorted`, etc.) stayed at -0.0% — reverted before the real baseline regen.
- The eight pre-existing rows moved within noise on both arches (< 0.01%) except `arrays_map_iterate` (+1.4%/+1.6%) and `users_yq_keys_unsorted` (+0.8%/+1.9%) — both well under the 5% threshold and consistent with ordinary staleness accumulation since the baseline's last refresh, not a regression this PR introduces.

## Test plan
- [x] `python3 -m py_compile scripts/perf-guard.py`
- [x] `scripts/perf-guard.py --check` on both x86_64 (terminus) and ARM64-Linux (container): 0.0% drift on all 13 rows
- [x] Negative test proving the new rows are load-bearing (see above)


newhoggy:
Ran an 8-way multi-agent review (line-by-line, removed-behavior, cross-file trace, reuse, simplification, efficiency, altitude, conventions). Summary:

**No output-correctness bugs found** — all 5 new filters independently re-verified byte-identical to jq 1.7.1/yq v4.53.3 by multiple agents, including edge cases (empty/absent `.users`, missing `score`). Baseline JSON structurally clean (no missing/duplicate keys, correctly sorted).

**Fixed — commit-scope violation** (confirmed via the project's own `omni-dev git commit message lint`): all three original commits used scope `perf-guard`, which isn't in `.omni-dev/scopes.yaml`'s enforced list. Reworded to `scripts`/`test` via history rewrite (solo branch, safe).

**Fixed — two stale row-count comments** this PR's own additions made worse: "one of the six tracked queries" (now 13) and "one yq-mode query" (this PR adds a second).

**Fixed — a real overclaim, caught by direct instrumentation**: one reviewer rebuilt with a temporary `eprintln!` in `resolve_node_sink` and found `users_assign_scores`/`users_path_walk` never call it at all — they're plain field/iterate targets, so `needs_path_prepass` routes them through the single-path walkers (`walk_path`/`set_path`) instead of `resolve_node`/`resolve_node_sink`. My original doc text claimed all four jq-mode rows exercised the resolver; that's only true for `users_del_select`/`users_del_bound_select` (their `select`/`as` triggers the multi-path route). This matches my own earlier negative-test observation (those two rows showed 0% drift under a `resolve_node_sink` slowdown, unlike the other three) — I'd noted it at the time but hadn't chased down why. Corrected the docstring/EPILOG/inline comment to describe what each row actually reaches; the single-path-walker coverage is still real and valuable, just not the coverage originally claimed.

**Fixed — dropped a false `|=` coverage claim**: no row anywhere exercises `|=`. Filed as #2905 rather than added here, since a `|=` row needs the same `select`/`as`-shaped target the `del_*` rows use to actually reach `resolve_node_sink` — a plain-target row would just re-cover the already-covered single-path walkers, not close the real gap.

All other angles (reuse, simplification, efficiency, altitude/root-cause, cross-file trace) came back clean — the hand-picked-rows pattern is this file's own pre-existing, explicitly-documented convention (and the file's own `#2666`-era comment literally anticipated this exact fix), no CI timeout/row-count risk, no redundant fixture generation, no duplicated seed/comparison logic worth extracting.

Rebuilt (`cargo build --features cli`) and Python-syntax-checked (`python3 -m py_compile`) clean after all fixes; full `cargo test` was stopped mid-run due to severe CPU starvation on this machine from concurrent sessions (confirmed via CPU-time-vs-wall-clock: near-zero progress over many minutes) — this PR touches zero Rust source, only the Python script and baseline JSON data, so CI's own suite on uncontended infrastructure is the authoritative gate here.

Issue #2906, opened 2026-09-13
jq: large-integer +/- arithmetic can round to a different f64 than real jq even when both operands are non-negative


newhoggy:
**Severity: Low**

## Summary

Found while broadly differential-fuzzing #2631's fix (large-`i64` `+`/`-`/`*` results past
`2^53`). Even with #2631's `a as f64 op b as f64` naive-cast approach correctly routing a
past-`2^53` exact integer result through `jq_bare_float_display`, a residual ~5-6% of
random large-integer `+`/`-` combinations still print a *different* double than real jq —
even with **both operands non-negative**, so this is unrelated to the separately-documented
"unary minus destroys number literal" behavior.

## Repro

```console
$ /usr/bin/jq -n '869389897822472004 + 944331'
869389897823416300
$ succinctly jq -n '869389897822472004 + 944331'
869389897823416400
```

Both sides agree the *exact* integer sum is `869389897823416335`; the divergence is in
*which* double each implementation's addition produces, not in a shortest-round-trip
tie-break once the double is known ([#2542](#) is unrelated; last digits are even on
both sides so its tie-break mechanism cannot fire here either).

Investigated with Python: neither `float(a) + float(b)` (each operand cast to `f64`
independently, then added — what succinctly does) nor `float(a + b)` (the *exact* integer
sum cast to `f64` once — what a naive "jq must use full-precision decNumber arithmetic and
round once" theory would predict) matches jq's actual answer in every case. Empirically,
across a 400-case-per-operator random sample:

| op | naive (`f64(a) op f64(b)`) only | exact (`f64(a op b)`) only | both agree | **neither matches jq** |
|----|----------------------------------|------------------------------|------------|--------------------------|
| `+` | 113 | 11 | 262 | 14 |
| `-` | 100 | 13 | 269 | 18 |
| `*` | 107 | 15 | 261 | 17 |

"naive" is the dominant model (~90% combined with the "both agree" cases), which is why
#2631's fix (adopting it) is a strict improvement over `main`, where *every* past-`2^53`
case was wrong (100% divergence, since the exact integer was kept verbatim). But the
remaining ~4% "neither" bucket shows real jq's actual arithmetic is doing something more
subtle than either simple model — almost certainly related to jq 1.7's decNumber-backed
literal preservation interacting with plain `f64` arithmetic in a mixed, not-fully-naive
way (e.g. converting a literal's *exact* decimal digits and the other side's already-computed
[...]

PR #2907, opened 2026-09-13
fix(jq): route past-2^53 exact int arithmetic through jq_bare_float_display

### LINKED ISSUE #2631: jq: computed large-integer-valued float ignores shortest-round-trip trailing-zero collapse (differs by literal spelling)
**Severity: Low**

## Summary

Found during #2542's PR review while fuzzing `1 * <literal>` against `/usr/bin/jq` 1.7.1
across random `f64` bit patterns. For a computed (forced, via `1 * ...`) float whose exact
value exceeds `2^53` (so multiple decimal integers collapse to the same double once the ULP
exceeds 1), `succinctly jq` sometimes prints the *literal's own parsed digit string* instead
of the double's true shortest round-tripping decimal — even though both strings parse to the
identical `f64` bit pattern.

## Repro

```console
$ echo null | jq -c '1 * -33201876582270392'
-33201876582270390
$ echo null | succinctly jq -c '1 * -33201876582270392'
-33201876582270392
```

Confirmed the two spellings parse to the *same* double (`-33201876582270390.0 ==
-33201876582270392.0` is `true` in Python, and both have bit pattern `c35d7d3b92e4016e`) —
at this magnitude the ULP is `4`, so several adjacent integers all round to one double, and
the *shortest* round-tripping spelling is the one needing fewer significant digits
(`...390`, trailing zero not counted as significant), not necessarily the one the source
literal happened to spell out.

Rust's own `f64::to_string()` gets this right in isolation:

```rust
fn main() {
    let f: f64 = -33201876582270392.0;
    println!("{}", f.to_string()); // -33201876582270390
}
```

...but `succinctly jq`'s actual `1 * <literal>` pipeline does not reproduce that — for the
*same* input literal spelled in scientific notation instead (`1 * -3.320187658227039e+16`,
which parses to the identical double), succinctly gives the correct `-33201876582270390`.
Only the plain-decimal-integer literal spelling triggers the wrong answer, which points at
an integer-shaped fast path (treating an operand that looks like a whole number, or the
`1 * x` identity case, as literal-preserving or integer arithmetic) bypassing the normal
`jq_bare_float_display`/`f.to_string()` shortest-round-trip formatting that the scientific
literal correctly goes through.

Reproduced this on unmodified `main` (`eef77e110`), confirming it predates #2542 and is
unrelated to that issue's own dtoa tie-break fix — 8 of 2000 random `f64` bit patterns hit
this exact shape in a differential fuzz run, always with the correct (jq) answer ending in
an even digit (i.e. never a case #2542's fix mechanism would touch, since that only ever
acts on an already-odd last digit).

## Root cause (not yet located)

Not yet traced to a specific function -- the discrepancy is keyed on the *literal's source
spelling* surviving through `1 * <literal>` in one case but not the other, which suggests
either:
- A literal-preservation path that keeps (or partially keeps) the original digit string for
  an integer-shaped decimal literal even after the `1 *` multiplication should force full
  recomputation, or
- An integer-domain fast path for `x * 1`-shaped arithmetic that formats via plain integer
  text rather than routing the result back through `jq_bare_float_display`.

## Suggested fix direction

Trace `eval.rs`'s/`eval_generic.rs`'s multiplication evaluator for the specific case where
one operand is the literal `1` and the other is a `NumberLiteral`/`NumberRepr::Int`-shaped
large integer text, to find where the result bypasses `jq_bare_float_display`.

## Why this is separate from #2542

#2542 is specifically about `jq_bare_float_display`'s (and the yq-mode equivalent's) own
dtoa *tie-break* disagreement with jq/yq's reference implementations, fixed by
`correct_shortest_decimal_tiebreak`. That function only ever acts when the *current* last
digit is odd; every case found here already has an even last digit on both sides, so
#2542's fix mechanism cannot fire on it either way -- this is a distinct bug in how (or
whether) the shortest-round-trip formatter is reached at all for this literal shape, not a
disagreement about which decimal that formatter should pick once reached.

newhoggy:
## Summary

- An exact, non-overflowing `i64` `+`/`-`/`*` result whose magnitude exceeds `2^53` was
  kept as `OwnedValue::Int` and printed via its own exact digits — bypassing
  `jq_bare_float_display`'s shortest-round-trip formatting entirely, even though jq's
  own numbers are always `f64` and multiple adjacent integers can share one double past
  that bound (`1 * -33201876582270392` printed `-33201876582270392` instead of jq's
  `-33201876582270390`).
- Adds `jq_int_result_within_exact_f64_range` and applies it to `arith_add`/`arith_sub`/
  `arith_mul`'s `(Int, Int)` arms (jq mode only — `S::OVERFLOW_WRAPS` keeps yq's wrapping
  semantics untouched).

## Scope note

Broad differential fuzzing against `/usr/bin/jq` 1.7.1 surfaced a separate, deeper,
pre-existing quirk: real jq's large-integer `+`/`-` arithmetic can round to a *different*
`f64` than a naive `a as f64 op b as f64` cast in ~5-6% of random cases (with **both**
operands non-negative, so unrelated to the documented unary-minus-destroys-literal
behavior) — likely tied to jq's decNumber-backed literal preservation interacting with
plain `f64` arithmetic in a way neither a pure-double nor pure-exact-integer model fully
predicts. That's out of scope for this issue (#2631 is specifically about the fast path
bypassing the formatter, not about which double the formatter receives) and doesn't
regress anything — `main` was wrong 100% of the time here, this fix is wrong ~5-6% of the
time (a strict improvement). Filed as follow-up #2906.

Fixes #2631

## Test plan

- [x] New regression tests: `test_int_arith_past_exact_f64_range_matches_jq_bare_float_display_2631`,
      `test_int_arith_within_exact_f64_range_or_already_overflowing_unaffected_2631`
      (`tests/jq_cli_tests.rs`), `test_yq_int_overflow_wraps_unaffected_by_2631`
      (`tests/yq_cli_tests.rs`)
- [x] Live-verified against `/usr/bin/jq` 1.7.1 and yq v4.53.3


newhoggy:
Addressed the code-review findings in 3710dd430:

- **Correctness (confirmed live, high-severity):** the past-`2^53` bound only checked the
  *result's* magnitude, not the *operands'* — `9007199254740995 - 9007199254740994` gave
  `1` instead of jq's `2`, since both operands individually exceed `2^53` (each already
  lossy in jq's number model) while the exact difference cancels back under it. Fixed by
  requiring both operands *and* the result to be within range before trusting the exact
  `i64` fast path (`jq_checked_int_arith`).
- **Simplification:** collapsed the triplicated add/sub/mul guard into that one shared
  helper.
- **Documentation:** added a `docs/compliance/jq/limitations.md` entry for the residual,
  disclosed decNumber-mixed-precision gap (filed separately as #2906) per ADR-0018.
- Added regression coverage for the two-large-operand shape.

Full test suite, fmt, clippy, and doc all green after the fix.

Issue #2692, opened 2026-09-10
jq/yq: `and`/`or`/`//`/`not` validate a document they never read, contradicting #2103's rule


newhoggy:
## Summary

#2103 (ADR-0018 amendment, adopted 2026-09-09) established that **a filter validates only what
it reads**: succinctly does not reproduce jq's whole-document parse-time rejection, and where
it cannot do so uniformly it takes the *uniform* divergence rather than one that depends on how
the filter is spelled.

Four expression arms still validate the entire ambient document despite reading none of it, via
`ambient_validation_error` (`src/jq/eval_generic.rs`, added by #2476 on 2026-09-08 — the day
*before* #2103's decision was recorded). The result is exactly the per-spelling split the
amendment exists to remove.

## Repro (fresh build of `main` @ `ab350e9f4`)

```console
$ printf '{"a": {"bad": "\x"}, "b": 5}' > bad.json   # jq 1.7.1: parse error, exit 5, every filter
```

| filter | succinctly | reads anything? |
|---|---|---|
| `1+1` | rc=0 `2` | no |
| `1==1` | rc=0 `true` | no |
| `"x"` | rc=0 `"x"` | no |
| `.b` | rc=0 `5` | yes — and never reaches `.a.bad` |
| `.` | rc=0 (echoes) | yes |
| **`1 and 1`** | **rc=5** | **no** |
| **`1 // 2`** | **rc=5** | **no** |
| **`not`** | **rc=5** | reads only the root's truthiness, which `is_falsy` answers in O(1) without decoding |

`1 and 1` and `1+1` are the same class of expression over the same input and disagree about
exit code. Same on the `{123: 1}` structural shape.

## Where

`ambient_validation_error` has three call sites, all of which run the walk on the
"reads nothing" shape:

- `eval_boolean_generic` — gated on `!needs_path_context(left) && !needs_path_context(right)`,
  covering `Expr::And` / `Expr::Or`
- `Expr::Alternative`'s arm — unconditional
- `Expr::Not`'s arm — inline via `push_generic_truthiness`'s `OneCursor` arm, which runs
  `push_generic_document_validation_error` before reading `is_falsy`

Each carries a comment explaining that it preserves the raise `to_owned_with_cursor` used to
produce for `try (1+1) catch "x"` (#1812). That justification was retired by #2103, which
changed `test_try_catch_contains_a_genuinely_catchable_malformed_key_error_1812` to assert
exit **0** for that very expression.

## Suggested fix

[...]

PR #2910, opened 2026-09-13
perf(jq): make [...] array construction lazy over a cursor-shaped source

### LINKED ISSUE #2575: perf(jq): `[.[]]` and `[.[] | select(f)]` materialize every element where `map(select(f))` stays lazy
**Severity: Medium** (perf only; no incorrect output). Found while investigating #2168.

## What

Array construction over a cursor-iterating body materializes every element as an
`OwnedValue` before the array is consumed, even when the consumer only wants its
`length`. The equivalent `map(select(f))` stays lazy and is ~4x faster.

Release build, 4.2 MB `users`-pattern JSON (41,943 user objects), this laptop,
interleaved medians of 5, peak RSS via `/usr/bin/time -l`:

| query | time | peak RSS |
|---|---|---|
| `.users \| map(select(.active)) \| length` | 40 ms | |
| `[.users[] \| .name] \| length` | 46 ms | |
| `[.users[] \| select(.active) \| path(.name)] \| length` | 114 ms | 28 MiB |
| `[.users[]] \| length` | **138 ms** | **79 MiB** |
| `[.users[] \| select(.active)] \| length` | **164 ms** | |

`[.users[] | .name]` is cheap because each element is a short scalar. `[.users[]]` and
`[.users[] | select(.active)]` pay a full `OwnedValue` copy per object. `map` over the
same elements does not, so the cost is the `[...]` constructor's collection step, not
`select`.

## Suggested direction

`LazySource::Cursors` already retains a cursor vector for the path-context walk (#2151),
and the sort family keeps elements as cursors (#1687). Array construction whose body
yields `OneCursor` items could collect cursors the same way and materialize only when a
consumer demands values — `length`, `.[i]`, `first`, `last` and re-iteration would not.

Whether that is worth its complexity depends on how often `[.[] | select(f)]` is followed
by something that does not need the values; `| length`, `| .[0]` and `| first` are common
enough shapes to measure before deciding.

Refs #2168, #1687, #2151.

newhoggy:
## Summary

- `Expr::Array`'s generic-evaluator arm used to materialize a cursor-shaped inner result (`GenericResult::OneCursor`/`ManyCursor`) into a fresh `OwnedValue::Array` up front, even when the consumer only wanted `length`/`first`/`.[n]`. `[.[]]` and `[.[] | select(f)]` now become `LazySeq::from_cursors` instead — the same lossless answer `map(f)` (already `LazySeq`-returning) and #1687's reordering builtins already use.
- **Phase 2**: fast paths for an instruction-free `LazySeq` (no `map` stage on top, so no element can error) — any integer index (negative included), `last`, and a literal-bounds slice answer directly from the cursor slice. A nested-pipe stage whose head is one of these (`.[0].name`) is peeled and folded through `fold_pipe_stages`, guarded on the head's own shape so an ordinary nested pipe (`.foo.bar`) is unaffected.
- **Phase 2b**: two mitigations for a real, measured tiny-array-construction regression (`[xs[] | [tiny]]`-shaped queries) — `LazySeq`'s `instructions` field is `Option<Rc<Vec<Instruction>>>` rather than an always-allocated `Rc<Vec::new()>`, and `materialize_atomic` gets its own fast path for an instruction-free `Cursors` source. Brought the regression from +11-16.5% to +3.4-4.3%, confirmed against the benchmark machine's own noise floor — accepted rather than chasing further (see the PR's own doc updates for the reasoning).
- The `#2103`/`#2692` "a filter validates only what it materializes" contract widens by one more construct: `[.[]] | length` no longer raises on a document with an undecodable member it never reads. Recorded in `docs/compliance/jq/limitations.md`.
- Closes one row of `#2642`'s "owned embed map" residual as a side effect: `[.] | .[0] | path($x)` now stays accepted, matching jq's own node-identity behavior. The sibling shape where navigation happens inside `path()`'s own argument is a different mechanism and stays refused.

Fixes #2575

## Test plan
- [x] `cargo test --features cli` (full suite: 3750 lib + 1709 jq_cli + 1903 yq_cli + doctests + nets) — all green
- [x] `cargo test --no-default-features --verbose` (no_std) — clean
- [x] `scripts/jq-alt-retry-oracle-sweep.sh` — 861/0/0
- [x] `scripts/jq-m2-streaming-sweep.sh` — 744 cases, golden regenerated and reviewed row-by-row (exactly the 2 predicted rows changed)
- [x] `scripts/jq-path-context-oracle-sweep.sh` — 6300 cases, 0 unexpected
- [x] `scripts/yq-alias-oracle-sweep.sh` — clean, no new divergence classes
- [x] Interleaved A/B (`scripts/ab-cli.py`) on the pinned ARM box (johns-mac-mini): target shapes 4.1-8.7x faster with proportional RSS reduction, matching the issue's own pre-measured spike numbers; holdouts inside the control range. x86_64/7950X was occupied by unrelated work during this session — CI's own dual-arch Perf Regression Guard is the independent confirmation there.
- [x] Live-verified every characterization row against the pinned oracle (`/usr/bin/jq` 1.7.1)


newhoggy:
## Code review addressed

8-angle review (two passes, several structural/efficiency findings converged across independent angles):

- **Fixed**: \`materialize_atomic\`'s new fast path now calls the existing \`to_owned_all_cursors\` helper instead of a hand-rolled duplicate loop.
- **Fixed**: the \`Expr::Pipe\` head-peeling arm no longer clones the remaining cursor slice — \`seq\` is already exactly the untouched source the peeled head needs, so it's moved into the recursive call instead of cloned. Restructured as if/else-if since the moving arm and the borrowing arms can't be sibling match arms over the same scrutinee.
- **Added**: yq-mode negative-index-out-of-range coverage through the new fast path specifically (\`[.a[]] | .[-N]\`), flagged by multiple reviewers as an untested shape.
- **Deferred, filed as #2913**: \`LazySeq::fold_one\`'s own per-element \`Vec\` allocation for an instruction-free source — a pre-existing cost (\`map(.)\`'s own #724/#725 landing already pays it) that this PR exposes to more callers (printing \`[...]\` directly), not something introduced by array-construction laziness specifically. A generally-applicable fix to \`LazySeq\`'s own \`Iterator\` impl, scoped as its own issue rather than folded into this one.
- **Not changed** (assessed, judged acceptable): the \`Expr::Slice\` arm's smaller sub-range clone, the \`Option<Rc<Vec<Instruction>>>\` design (an alternative shared-empty-\`Rc\` design was suggested; both give the same allocation-avoidance property, this one was already implemented/tested), and the "three places list which shapes get the fast path" structural observation (a genuine missed-optimization risk if a future fast-path shape forgets to update all three, not a correctness bug — noted for a future contributor rather than restructured now).

Full suite (3750 lib + 1709 jq_cli + 1904 yq_cli + doctests + nets), clippy, fmt, no_std, rustdoc, and all four oracle sweeps re-verified clean after the fixes.

Issue #2914, opened 2026-09-13
jq: computed-bracket .E[K] evaluates the target once instead of re-evaluating it per key output


newhoggy:
**Severity: Low**

## Summary

Found while implementing #2259's fix (`getpath`'s path-context argument generator was
eagerly drained instead of pulled lazily). `path_context_step_computed_index` (`.E[K] | key`
etc., `src/jq/eval_generic.rs`) evaluates the bracket's *target* `E` exactly once, before the
key stream `K` is pulled at all, and reuses that one value for every key output.

Real jq desugars `E[K]` as `K as $k | E | .[$k]` — confirmed live with unbuffered output
against `/usr/bin/jq` 1.7.1:

```console
$ echo '{"a":1,"b":2}' | jq --unbuffered -c '(debug("target")|.)[("a", (debug("key2")|"x"))]'
["DEBUG:","target"]
1
["DEBUG:","key2"]
["DEBUG:","target"]
null
```

`target` fires once *per key output* (interleaved with each key's own side effects), not
once up front. succinctly currently evaluates the key stream first (in full) and the target
second (once), so both the ordering *and* the "does `E` re-run per key" question diverge
from jq's real desugaring whenever `E` itself is a multi-output generator with observable
side effects, or is expected to see a different ambient value per key iteration.

## Why this is out of scope for #2259

#2259 was specifically about `getpath`'s own argument generator being pulled eagerly
instead of lazily. This issue's mechanism is different and, empirically, deeper: it isn't
just "stop pulling after a failure" (the shape #2259 fixed) but "the target must be
re-evaluated fresh for every key output, interleaved with that key's own evaluation" — a
structural change to `path_context_step_computed_index`, not a sink conversion of an
existing collector. Attempting it as a drive-by inside #2259's PR risked introducing a
different (also-wrong) ordering without full verification.

## Suggested fix direction

Restructure `path_context_step_computed_index` so the key stream drives a loop that
re-evaluates `target` fresh per key (via `path_context_step_target`), rather than computing
`targets` once outside the loop. Needs its own oracle-verified test matrix (single-output
target/key is unaffected; the multi-output×multi-output combination is where the current
implementation and jq's true semantics can diverge in both output values and side-effect
order).

newhoggy:
Re-verified against jq 1.7.1 and current `main` (`b0e648611`) — this no longer reproduces.

[...]

Issue #2429, opened 2026-09-04
jq: [.. | (getpath([paths]))?] drops the root position's own [] output


newhoggy:
Filed against spine #2416 (hygiene rule 1). Found by the Phase 0 sweep; pinned in the jq golden corpus as `path_ctx_optional_getpath_recurse` (known divergence).

## Capture (jq 1.7.1)

```console
$ echo '{"a":[1]}' | jq -c '[.. | (getpath([paths]))?]'
[[],1]
$ echo '{"a":[1]}' | succinctly jq -c '[.. | (getpath([paths]))?]'
[1]
```

At the root, `paths` yields `["a"]` and `["a",0]`, so `getpath([paths])` is `getpath([["a"],["a",0]])`, which errors in jq (a path element must be a string or number); `?` suppresses it. jq then continues with the *next* `..` output... but jq's first output is `[]` — capture more precisely which position produces `[]` (likely `getpath([])` on `1`, i.e. `paths` on a scalar yields nothing so the argument is `[]` and `getpath([])` is identity) before assuming the mechanism. The point is that succinctly is missing an output real jq produces, and the sweep says it is the `getpath` path-context arm under `?` (#2259 family).

## Why parked

`getpath`'s path-context arm is #2259, absorbed by #2416 (falls out of the sink protocol). Re-check this case after that migration; if it still diverges, it is a separate bug in how `?` interacts with a multi-output `getpath` argument.

newhoggy:
Re-verified post spine 2416's exit (`e4449b1f4`, `1db60e608`). Still diverges on the replacement route, same shape as before:

```console
$ echo '[1]' | /usr/bin/jq -c '[.. | (getpath([paths]))?]'
[[],1]
$ echo '[1]' | succinctly jq -c '[.. | (getpath([paths]))?]'
[1]
```

Root cause is the same function as #2259 (`path_context_step_getpath`/`path_context_component_values`, `src/jq/eval_generic.rs` ~line 13351/13675) -- this issue's own text already calls it the "#2259 family."

Moving to Ready: the original blocking rationale (wait for spine 2416 to land) is satisfied. Recommend picking this up together with #2259 given the shared root cause -- see that issue's own updated comment for the current scope note (likely needs a new lazy-sink primitive, not a quick patch).

PR #2915, opened 2026-09-13
fix(jq): pull getpath's path-context argument generator lazily

### LINKED ISSUE #2259: jq: getpath's path-context arm doesn't interleave a multi-output path argument with the walk (side-effect ordering)
Found during code review of #2258 (the #2253 fix) — deliberately scoped out of that PR
rather than rushed, since fixing it properly needs a new architectural primitive.

## What

Real jq's `getpath(EXPR)` desugars the same way every generator-argument builtin call
does (`EXPR as \$b | body`, per `fanout_arg`'s own extensive doc comment in
`src/jq/eval.rs`) — the argument generator is pulled lazily, one output at a time, and a
`body` (walk) failure on output N stops the generator before output N+1 is ever produced.
Side effects included.

`eval_getpath_with_path_context` (added by #2258, `src/jq/eval.rs`) evaluates the whole
path-argument generator up front via `eval_expr_needing_path_context` +
`drain_path_context_stream`, then walks each resolved path in a second pass. This means a
*later* output's side effects (`debug`, `stderr`, `halt_error`, ...) fire even when an
*earlier* output's walk has already failed — something real jq never lets happen, since it
never reaches the later output's evaluation at all.

Live-verified against `/usr/bin/jq` 1.7.1 and against succinctly's own plain (non-path-
context) `getpath` evaluator (`builtin_getpath`, which correctly uses `fanout_arg` +
`eval_each`'s lazy sink and has no version of this bug):

```bash
$ echo '{"a":5}' | succinctly jq -c 'getpath((["a","x"], (debug("side")|["a"]))) | key'
["DEBUG:","side"]
jq: error (at <stdin>:1): Cannot index number with string "x"
```

Real jq, and succinctly's own plain evaluator on the same filter without the trailing
`| key`, never print the debug line — `["a","x"]`'s walk fails first and the generator
never reaches its second alternative.

A `halt_error` variant is worse: it fires (printing its own stderr banner) and then its
halt is silently discarded in favor of an unrelated later error, violating this codebase's
own `#791` rule ("a halt always escapes").

## What #2258 already fixed, to be clear about the remaining scope

#2258 does correctly preserve *already-yielded* path-argument outputs across an
`Error`/`Break`/`Halt` from the generator itself (walks and folds them into `results`
before the escape fires) — that half of the correctness picture (no silently dropped
output) is solid. What's still wrong is strictly about *evaluation order*: a later
alternative's side effects/escapes shouldn't be observable at all once an earlier one's
walk has failed, and today they still run to completion before any walk is attempted.

## Why this needs a new primitive, not a quick patch

The lazy, per-output, early-stopping evaluation `fanout_arg` gives the plain evaluator for
free comes from `eval_each` (`src/jq/eval.rs`), a sink-based generator walker that only
exists for `StandardJson`-domain values. `eval_getpath_with_path_context` operates in the
`OwnedValue`/path-context domain (it needs to thread `current_path` through path-context-
requiring arguments too, via `eval_expr_needing_path_context`), and no `OwnedValue`-domain
equivalent of `eval_each`'s lazy sink exists yet. Building one (or otherwise achieving the
same interleaving guarantee) is comparable in scope to the whole of #2258's own fix, not a
follow-up patch to it.

## Scope note

Low real-world impact: this only manifests for a `getpath(...)` argument that is (a) a
multi-output generator, (b) has a downstream `key`/`parent`/`path`/`file_index` forcing
path-context routing, and (c) has an earlier alternative that fails while a later one has
an observable side effect or its own escape. The overwhelming majority of real `getpath`
usage passes a single literal array, where this gap cannot manifest at all (only one
output, nothing to interleave).

newhoggy:
## Summary

- `eval_getpath_with_path_context`'s path-argument generator (reached whenever a downstream
  `key`/`parent`/`path`/`file_index` forces path-context routing) was drained fully up front
  and walked in a second pass, so a *later* path's side effects (`debug`, `halt_error`, ...)
  fired even when an *earlier* path's walk had already failed — real jq never lets this
  happen, since `getpath(EXPR)` desugars to `EXPR as $p | body` and pulls the generator
  lazily, one output at a time (`fanout_arg`'s own doc comment).
- Adds `path_context_component_each`, a `Demand`-driven sink twin of
  `path_context_component_values`, built on the already-lazy `eval_each_generic`/
  `eval_each_owned`. Rewires `path_context_step_getpath` to walk each path as it's produced
  through the sink, stopping the generator the moment a walk fails.
  `path_context_component_values` is now implemented on top of the same sink, so there's one
  definition instead of two.
- `#2429` (`[.. | (getpath([paths]))?]` drops the root's own `[]` output) was re-checked on
  the fixed build and still diverges — confirmed a separate `?`-interaction bug, not fixed
  by this change, exactly as that issue's own "re-check after #2259" note anticipated.

## Scope note

The sibling `.E[K] | key` arm (`path_context_step_computed_index`) is deliberately left
alone. Converting its key stream to the same lazy-stop-on-failure shape interacts with a
separate, deeper, pre-existing divergence: the bracket's target `E` is evaluated once,
eagerly, before any key is pulled, while real jq's `K as $k | E | .[$k]` desugaring
re-evaluates `E` fresh for every key output, interleaved with that key's own side effects
(confirmed live with unbuffered output against `/usr/bin/jq` 1.7.1). Fixing that needs its
own oracle-verified test matrix and isn't a drive-by inside this PR. Filed as #2914.

Fixes #2259

## Test plan

- [x] New regression tests: `test_getpath_path_context_pulls_argument_generator_lazily_2259`
      (`tests/jq_cli_tests.rs`, covers the debug-suppression, halt-precedes/halt-follows, and
      success-then-fail-then-never-pulled shapes, plus plain-`getpath` oracle agreement) and
      `test_yq_getpath_path_context_pulls_argument_generator_lazily_2259`
      (`tests/yq_cli_tests.rs`, under `--jq-extensions`)
- [x] Live-verified against `/usr/bin/jq` 1.7.1


newhoggy:
Addressed the code-review findings in e79843ace:

- **Scope (confirmed live)**: three sibling call sites of `path_context_component_values`
  have the identical eager-drain-vs-side-effect-ordering bug —
  `path_context_step_computed_index`'s key stream, `path_context_step_computed_slice`'s
  bounds, and the `Expr::If`/`Expr::Limit` arms. Filed as #2916 rather than folded into this
  PR, since each needs its own careful oracle-verified conversion and test matrix. The doc
  comment on `path_context_step_getpath` now scopes its laziness claim to
  `eval_each_generic`'s own native-arm coverage (the same ceiling every other lazy consumer
  built on it already has, e.g. for `inputs`) instead of overclaiming universal laziness.
- **Documentation**: moved the "spine 2416 (walk residue)" values-before-escape invariant
  comment onto `path_context_component_each` (the primitive the new `getpath` consumer
  actually calls) rather than leaving it only on the eager wrapper.
- **Simplification**: collapsed the `PathNode::Absent`/`Owned` arms into one (they differed
  only in which value was passed to `eval_each_owned`).

On the `nonretryable_stop`/`stop_with_escape` finding: `stop_with_escape` is this codebase's
single established mechanism for exactly this purpose ("the one place that decides what such
a stop means to a `?//`", per its own doc comment) and is already used pervasively for decode
failures throughout `eval_generic.rs` (`stream_owned_outputs_generic`, `fanout_arg_each_generic`,
etc.). Using it for the new `walk_error` slot applies that same established, codebase-wide
convention to a call site that previously bypassed it — not a novel mechanism. The specific
"ordinary `try` doesn't clear `nonretryable_stop`" interaction the review raised (if real)
would already be a pre-existing, orthogonal property of every other existing
`stop_with_escape` call site near a `try`, not something this PR introduces or is positioned
to fix; no minimal repro was constructed by the review either. Left as-is.

Full test suite, fmt, clippy, and doc all green after the fix.

Issue #1003, opened 2026-08-15
jq: join/sub replacement-function error wording doesn't match jq's accumulator-state-derived messages


newhoggy:
## Summary

Split from #991 item 1. Real jq's error wording for a type mismatch inside `join`'s
separator handling or `sub`'s replacement-function evaluation comes from *internal
accumulator state* at the point jq's own `reduce`-based (`join`) or replacement-pipeline
(`sub`) definition fails — not simply "the array/string being processed and the bad
argument."

## Repro (verified against pinned jq oracle)

```
$ echo '[1,2]' | jq -c 'try join(1) catch .'
"string (\"1\") and number (1) cannot be added"
$ echo '[1,2]' | succinctly jq -c 'try join(1) catch .'
"expected string, got non-string"

$ echo '"a"' | jq -c 'try sub("a"; 5) catch .'
"string (\"\") and number (5) cannot be added"
$ echo '"a"' | succinctly jq -c 'try sub("a"; 5) catch .'
"expected number, got non-number"  # approximate -- re-verify exact current wording
```

## Why this isn't a simple wording swap

Both are `EvalError::binary_op`-shaped messages (`<left> and <right> cannot be added`), but
jq's operands are whatever its own `reduce`/pipeline-based definition's accumulator holds
at the failure point (e.g. `join`'s accumulator starts as `""`, then gets `+`'d against
the separator or next element) — not simply "the input array and the bad argument."
Reproducing this exactly likely needs restructuring `join`/`sub`'s error paths to fail at
the same *point* jq's own definition does, which is a real behavioral change to how these
builtins evaluate, not just swapping the error-message constructor.

## Suggested approach

Needs investigation into exactly where succinctly's `join`/`sub`-replacement
implementations diverge structurally from jq's own `reduce`/pipeline-based definitions
before a fix strategy can be chosen — not a direct port of #929's fix pattern
(existing-constructor-reuse or a new one-line wording constructor).

newhoggy:
🔧 Claimed by agent `cadence` — branch `issue-1003-join-sub-accumulator-wording`

PR #2917, opened 2026-09-13
fix(jq): bare flatten flattens fully, matching jq's unbounded default (#2818)

### LINKED ISSUE #2818: jq: bare flatten (no depth argument) flattens only 1 level; real jq flattens fully

newhoggy:
## Summary
- `Builtin::Flatten` was wired to `builtin_flatten(value, optional, OwnedValue::Int(1))` — i.e. bare `flatten` was treated as `flatten(1)`. Real jq's bare `flatten` flattens fully, all the way down, with no depth limit at all.
- The issue's own suggested fix (jq's `def flatten: flatten(1/0);`) doesn't hold against the pinned oracle — `1/0` genuinely raises in jq 1.7.1 (`divisor is zero`), and empirically bare `flatten` fully flattens a 200-level-deep array, so whatever jq's C implementation actually does, it is not literally evaluating that division.
- `depth` here is only ever compared against 0 and decremented by 1 per level (`flatten_owned_at_depth`), and every value tree this crate can hold is already capped at `MAX_VALUE_TREE_DEPTH` levels (`assert_value_tree_depth`, ~384) — so `i64::MAX` as the sentinel is indistinguishable from true infinity for any document this crate can represent, with no new "infinite" concept needed anywhere.
- Updates `test_builtin_flatten`'s two stale assertions (which pinned the buggy 1-level behavior), and adds a golden case (`flatten_bare_is_unbounded_depth_2818`) captured from the pinned jq 1.7.1 oracle via `sync-jq-golden.sh`.

Closes #2818.

## Verification
```
$ echo null | succinctly jq -c '[[1,[2,[3,[4]]]],5] | flatten'
[1,2,3,4,5]
$ echo null | jq -c '[[1,[2,[3,[4]]]],5] | flatten'
[1,2,3,4,5]

# flatten(1) with an explicit depth is unaffected:
$ echo null | succinctly jq -c '[[1,[2,[3,[4]]]],5] | flatten(1)'
[1,[2,[3,[4]]],5]
```

## Test plan
- [x] New golden case + `cargo test --test jq_golden_tests`
- [x] Verified live against jq 1.7.1, including the 200-level-deep unbounded-recursion case and that `flatten(1)`/`flatten(N)` are unaffected


newhoggy:
An independent multi-agent review pass traced the actual `arith_sub` behavior for the `i64::MAX` sentinel and found a real (non-observable) inaccuracy:

**Fixed — internal arithmetic mechanism didn't match the comment's claim:** jq-mode `arith_sub`'s exact-`Int` shortcut (`jq_checked_int_arith`, #2631) only keeps an `Int` result when *both* operands stay within `f64`'s exact-integer range (`<= 2^53`). `i64::MAX` fails that on the very first `- 1`, silently promoting `depth` to a `Float` whose magnitude never actually decrements (the ULP at that scale exceeds 1.0). This still worked — the float never becomes bit-identical to `0`/`0.0`, so `flatten` still recurses fully, and `MAX_VALUE_TREE_DEPTH`'s own guard independently bounds the recursion — but it contradicted my own comment/CHANGELOG claim of ordinary per-level integer decrementing.

Switched the sentinel to 1 billion — comfortably under `2^53` and comfortably over `MAX_VALUE_TREE_DEPTH` (~384) — so every decrement this can ever see stays exact `Int` arithmetic the whole way, making the "no new infinite concept needed" claim literally true rather than true-in-observable-output-only. Added a white-box unit test (`test_flatten_unbounded_depth_sentinel_stays_exact_int_arithmetic_2818`) pinning `arith_sub`'s own exact-`Int` result at the chosen sentinel, directly against the traced mechanism.

**Fixed — a second, independently confirmed finding:** `Builtin::Flatten`'s dispatch arm has no `S::TAG` gate, so the fix silently changed yq-mode behavior too, with zero yq-mode test coverage (only jq-mode `query!` cases existed) — the exact shared-builtin dual-mode gap this repo's own #1003/PR #1036 incident established a standing check for. Live-verified against yq v4.53.3 (bare `flatten` also fully flattens there; the pre-existing non-array rejection is unaffected) before adding a `yq_query!` unit test and a yq-golden fixture (`flatten_bare_is_unbounded_depth_2818`, captured via `sync-yq-golden.sh`) alongside the existing jq-golden one.

Full test suite + both golden suites + clippy all green after the fixes (the very last full-suite rerun was stopped mid-run due to severe, sustained CPU starvation on this machine from concurrent sessions — confirmed via static CPU-time-vs-wall-clock on the stalled test process; the incremental change since the prior full-green run is one isolated literal swap with its own direct unit-test coverage added on top).

PR #2919, opened 2026-09-13
feat(jq): echo already-canonical compact spans without re-rendering

### LINKED ISSUE #2608: perf(jq): compact (-c) M2 rendering is 18-24% behind the DOM path on large arrays

newhoggy:
## Summary

Extends the existing `--preserve-input` raw-echo fast path in `JsonCursor::stream_json` to the *default* compact (`-c`), unsorted-keys, jq-number-convention render path: when a node's raw source span is already byte-identical to what `stream_json_pretty(compact, unsorted, JqCompat)` would itself produce, echo it verbatim instead of re-rendering through the formatter.

Adds `is_canonical_compact_jq_span(bytes: &[u8]) -> bool` (`src/json/light.rs`) — a strict, single-pass, non-backtracking recursive-descent scanner over `value := object | array | string | number | true | false | null` with zero whitespace tolerance, checking:

- **Numbers**: delegates to the existing `is_jq_canonical_number` (no reimplementation).
- **Strings**: enforces jq's exact escape convention from `write_json_body_jq` — the 7 short escapes, `\u00xx` only for controls/DEL with no short form, rejects `\/`, uppercase hex, lone surrogates, and raw unescaped control/DEL bytes; validates UTF-8.
- **Duplicate object keys**: reuses the existing `KeyHashes`/`key_hash` machinery on raw (already-verified-canonical) key spans — sound without decoding, since jq's canonical string encoding is injective. Bails safely on any hash collision (real duplicate or not) rather than trying to disambiguate.

The new echo branch sits strictly after the existing `--preserve-input` branch, so that behavior is untouched. `-a` (ascii-output) cannot reach it: `JsonConvention::JqCompat` always renders through the non-ascii-escaping `write_json_body_jq`, and both call sites into `stream_json` already exclude ascii/color/raw-output modes before reaching it — pinned by a dedicated test.

Fixes #2608.

## Test plan

- [x] `cargo fmt --check`
- [x] no_std check (`cargo test --no-default-features --verbose`)
- [x] jq/yq/CLI golden suites
- [x] New differential/fuzz test (seeded, 2000 iterations, multi-byte UTF-8 + edge bytes) asserting the checker's `true` answers always round-trip byte-identically
- [x] Pinned CLI tests for malformed numbers/duplicate keys/escapes/`-a` interaction


newhoggy:
Closing without merging — see the full writeup on issue #2608: the implemented fast path measures as a consistent +6.8-7.1% regression on its own target workload (10MB canonical compact `-c .`) on the pinned idle ARM benchmark box, well outside the ±0.5% control noise floor. Correctness/review findings are all resolved (this branch is clean on that front), but the core performance premise doesn't hold, so this isn't mergeable as-is. Branch left in place pending a design decision on the issue.

Issue #2918, opened 2026-09-13
jq: recurse's f is not suspended between its own outputs, so a node's later f outputs still fire (#2693 residual)


newhoggy:
**Severity: Low** (side-effect count and stderr ordering; delivered values are correct)

## Summary

[#2693](https://github.com/rust-works/succinctly/issues/2693) stopped `recurse` from
evaluating `f` at a node the consumer was already satisfied by — `limit(1; recurse(f))` went
from running `f` at 10000 nodes to running it at none. What it did **not** change: at the node
the traversal *is* at, `f` runs to completion before any of its outputs is descended into.

jq's `def r: ., (f | r); r;` interleaves instead: the first output of `f` has its whole subtree
traversed before `f` is asked for its second. So jq stops *between* `f`'s outputs, and
succinctly cannot.

## Repro

Confirmed live against jq 1.7.1 and `succinctly` at `81209f1ab` (post-#2693). Values match in
every row; only stderr differs.

Under a bound — one node's fan-out too many (it was the whole remaining tree before #2693):

```console
$ echo '[1,[2,[3]]]' | jq -c '[limit(2; recurse(.[]?|debug))]'
["DEBUG:",1]                                  # succinctly: ["DEBUG:",1] ["DEBUG:",[2,[3]]]
```

Unbounded, it shows as stderr *ordering*, because a node's whole fan-out is evaluated before
its first child is descended into:

```console
$ echo '{"a":{"x":1},"b":{"y":2}}' | jq -c '[recurse(.[]?|debug)]'
["DEBUG:",{"x":1}] ["DEBUG:",1] ["DEBUG:",{"y":2}] ["DEBUG:",2]
# succinctly:
["DEBUG:",{"x":1}] ["DEBUG:",{"y":2}] ["DEBUG:",1] ["DEBUG:",2]
```

And once through `?`, an extra `f` run at the node that errors:

```console
$ echo '{"a":1}' | jq -c '[recurse(.[]?, ("x"|stderr|error("f")))?]'
x                                             # succinctly: xx
```

Both evaluators are affected identically (`each_recurse_walk` is shared), and
`resolve_recurse_sink` has the same residual on the path side —
`path(limit(2; recurse((.a|debug), (.b|debug))))` on `{"a":1,"b":2}` writes one DEBUG line in
jq and two here.

## Why it is not a small fix

The traversal is an explicit LIFO stack of materialized nodes, in all three walkers. To stop
[...]

Issue #2908, opened 2026-09-13
jq: path()'s own consumer side still collects, so a bound outside path() can't stop the fold (#2694 residual)


newhoggy:
**Severity: Low** (over-eager generator consumption / side-effect count under a bound; output is correct)

## Summary

[#2694](https://github.com/rust-works/succinctly/issues/2694) closed the four *source*-side
eager arms it listed (`resolve_leaf`'s general case, an `if`/`select` condition, an `as`-bind
source, a fold nested in a source). The same limitations bullet also noted a **consumer**-side
edge under the same issue number, which that work did not touch and which is a different
mechanism: `path()` hands its own results to its consumer only after resolving more than the
consumer asked for, so a bound sitting *outside* `path(...)` cannot stop the fold inside it.

## Repro

Confirmed live against jq 1.7.1 and `succinctly` at `a953112f2` (post-#2694):

```console
$ echo '{"a":1}' | jq -c '[limit(1; path(foreach ((1|stderr),(2|stderr),(3|stderr)) as $i (.; .a)))]'
1[["a"]]

$ echo '{"a":1}' | succinctly jq -c '[limit(1; path(foreach ((1|stderr),(2|stderr),(3|stderr)) as $i (.; .a)))]'
12[["a"]]
```

Same for `first(...)` and for a `label`/`break` bound in place of `limit(1)` — all three write
`12` here and `1` in jq. The output is right in every case; only the source consumption is one
element ahead of the bound.

The `?//` shape the limitations bullet already records is the same edge seen from the other
side, and diverges in the opposite direction because jq's own `limit` break is retried by the
source's `?//` and the retried output lands past the bound:

```console
$ echo '[true,false]' | jq -c '[limit(1; path(foreach (1 as $x ?// $y | (stderr|1)) as $v (.; .)))]'
[true,false][true,false][[],[]]      # succinctly: [true,false][[]]
```

## Fix direction

The source side is now demand-driven end to end (`drive_fold_source` ->
`resolve_node_sink` -> `resolve_leaf_sink`/`resolve_cond_fork_stream`/
`resolve_bind_source_sink`). What is left is the terminal: `path()`'s own entry point resolves
into a collection before its consumer (`limit`/`first`/`label`) sees the first path, so the
consumer's `Demand::Stop` never becomes the fold's. Giving that terminal a sink form is the
same migration shape #2235 used for the arms below it.

`docs/compliance/jq/limitations.md`'s "generator reached only through an eager arm" bullet
[...]

Issue #2180, opened 2026-09-02
jq: `?//` retry misses a short-circuit stop when `//`, a nested `first`, `foreach` or a pipe-stage boundary materializes the bind first


newhoggy:
**Severity: Low** (surprising output cardinality, no data loss, no stderr leak)

Follow-up from #1519, which made a `?//`-alternatives bind re-run once per alternative under a
short-circuiting consumer, matching real jq. That fix works by treating a consumer's
`Flow::Stopped` as jq's escaping `break` and retrying the next alternative
(`eval::is_retryable_stop`).

## The residual

The stop only reaches `each_pattern_alternatives` if **every construct between the consumer and
the bind forwards demand into it** rather than materializing it. Four do not, so these still
answer once where jq answers twice. All rows confirmed live against jq 1.7.1 and the current
build:

| filter | jq 1.7.1 | succinctly jq |
|---|---|---|
| `[first((1 as $x ?// $y \| 5)//9)]` | `[5,5]` | `[5]` |
| `[isempty((1 as $x ?// $y \| 5)//9)]` | `[false,false]` | `[false]` |
| `[first(first(1 as $x ?// $y \| 5))]` | `[5,5]` | `[5]` |
| `[label $o \| (1 as $x ?// $y \| 5) \| (., break $o)]` | `[5,5]` | `[5]` |
| `def m(g): label $o \| g \| ., break $o; [m(1 as $x ?// $y \| 5, 6)]` | `[5,5]` | `[5]` |
| `[first(foreach (1) as $x ?// $y (0;.+1;.), "z")]` | `[1,2]` | `[1]` |

## Root cause is not `?//`

It is the missing-lazy-arm class already tracked as items 9 and 10 of
`docs/plan/jq-lazy-generator-consumers.md` (#820's programme). Each of `//` (`Expr::Alternative`),
a nested `first(...)`, `foreach`, and a parenthesised bind used as a pipe stage whose break
arrives from a *downstream* stage, lacks a demand-forwarding `eval_each`/`eval_each_generic` arm,
so it evaluates the bind eagerly through `drain_result` and absorbs the stop.

The evidence that it is the intervening construct and not the bind: demand-forwarding spellings
of the very same filters already match, because the stop does reach the `?//`:

```
$ succinctly jq -cn '[first((1 as $x ?// $y | 5)|.)]'                      # [5,5]  = jq
[...]

Issue #1462, opened 2026-08-22
jq: If/Try/Label/As/FuncDef wrappers reopen #820's data loss — widen eval_each's lazy arms (Stage 5)


newhoggy:
**Severity: Medium** (silent data loss — the design predicted stderr-only; measurement says otherwise)

**Stage 5** of [`docs/plan/jq-lazy-generator-consumers.md`](https://github.com/rust-works/succinctly/blob/main/docs/plan/jq-lazy-generator-consumers.md).

## Summary

Stage 2 gave `eval_each` native lazy arms for `Comma`, `Pipe` and `Paren` only. Every other
`Expr` variant falls back to eager `eval_single` — so **any wrapper between a short-circuiting
consumer and its comma reopens #820's bug**, including its data loss.

The design listed these as a low-priority "widen the arm set" tail. Re-measured at `main` @
`1541b3508` against pinned jq 1.7.1, that under-rates them: they destroy documents exactly the
way #820 did.

## It destroys documents

```
$ printf '{"id":1} {"id":2} {"id":3} {"id":4}' | jq -c '[isempty(if true then (1, input) else 9 end), .id]'
[false,1] [false,2] [false,3] [false,4]
$ ... | succinctly jq -c '[isempty(if true then (1, input) else 9 end), .id]'
[false,1] [false,3]                          <- documents 2 and 4 consumed
```

Identical loss for all four wrappers tested:

| Shape | jq | succinctly |
|---|---|---|
| `[isempty(if true then (1, input) else 9 end), .id]` | 4 results | **2** |
| `[isempty(1 as $x \| (1, input)), .id]` | 4 results | **2** |
| `[isempty(try (1, input) catch 9), .id]` | 4 results | **2** |
| `[isempty(limit(3; 1, input)), .id]` | 4 results | **2** |

## The stderr half

| Query (`-cn`) | jq stderr | succinctly stderr |
|---|---|---|
| `first(if true then (1,("B"\|stderr)) else 9 end)` | (none) | `B` |
| `first(try (1,("B"\|stderr)) catch 9)` | (none) | `B` |
| `first((1,("B"\|stderr))?)` | (none) | `B` |
| `first(label $p \| (1,("B"\|stderr)))` | (none) | `B` |
| `first(1 as $x \| (1,("B"\|stderr)))` | (none) | `B` |
| `isempty(if true then (1,("B"\|stderr)) else 9 end)` | (none) | `B` |
[...]

PR #2921, opened 2026-09-13
fix(jq): run a recurse node's f only if the consumer still wants more

### LINKED ISSUE #2693: jq: recurse(f)'s own f/cond is still fully resolved per accepted node, not streamed
**Severity: Low** (correctness-adjacent efficiency gap, over-execution of side effects under a bound — not incorrect output)

## Summary

`resolve_recurse_sink` (#2235, `src/jq/eval.rs`) made `recurse(f)`/`recurse(f; cond)` pop
visited nodes by demand, so a bounded consumer (`limit`, `first`) stops visiting nodes it
never asked for. What it does *not* do: for a node it *has* accepted, `f`/`cond` are still
resolved in full via `resolve_against_cow` rather than streamed to the sink, so a
multi-output `f`'s later outputs still fire even when only the first is ever consumed.

This was found and self-disclosed during review of #2685 (the fix for #2190), and is
already documented in `docs/compliance/jq/limitations.md`'s "Five residual divergences"
section (the `recurse(f)` bullet), but has no tracked issue number the way its two
sibling bullets do (#2159, #2032). Filing this to close that gap.

## Repro

```console
$ echo '{"a":1,"b":2}' | jq -c 'path(limit(2; recurse((.a|debug), (.b|debug))))'
```

jq writes `debug` for `.a` only — the bound is satisfied by `.a`'s own self-emission
before `.b` is ever asked for. `succinctly jq` fires `debug` for both `.a` and `.b`.

## Why

`resolve_against_cow` (the helper `f`/`cond` are resolved through inside
`resolve_recurse_sink`) has no sink form of its own — same underlying cause as the
`drive_fold_source` eager-arms gap (a separate, related issue).

## Fix direction

Give `resolve_against_cow` a demand-driven sink twin, the same shape
`resolve_node_sink`'s own migration used elsewhere in this file, and thread it through
`resolve_recurse_sink`'s `f`/`cond` evaluation. Needs its own live-oracle differential
verification against jq 1.7.1 before landing, per this repo's fidelity rules (ADR-0018).

newhoggy:
Fixes #2693

## What the issue described, and what was actually there

#2693 reported that `recurse(f)`'s own `f` is resolved in full per accepted node rather than
streamed, with a path-mode repro that fires one extra `debug` (`path(limit(2; recurse((.a|debug),
(.b|debug))))`, jq 1 vs succinctly 2).

That repro is the small half. Probing the same shape in **value** position found a much larger,
untracked gap with the same cause: neither `eval_each` nor `eval_each_generic` had a lazy arm for
the parameterised `recurse` spellings, so both evaluators fell to the collecting route, walked to
`RECURSE_MAX_ITEMS`, and ran `f` at every one of those nodes before a bounded consumer could
truncate the finished list.

```console
$ echo '{"a":{"x":1},"b":{"y":2}}' | jq -c '[limit(1; recurse((.a|debug), (.b|debug)))]'
[{"a":{"x":1},"b":{"y":2}}]
```

jq writes no DEBUG line: `recurse` is `def r: ., (f | r); r;`, so the node is emitted *before* `f`
runs on it and a consumer the node itself satisfies never pays for its children. succinctly wrote
**20000** — 10000 visited nodes, two `f` outputs each — for a five-node document. `first(...)` was
the same. Delivered values were always correct; this is entirely `f`'s side effects and the work
behind them.

## The change

`each_recurse_walk` is the traversal, written once and shared by the two collecting builtins and
the two new lazy arms. It carries #490's "one output, one child", #635's DFS order, #636's
abort-on-`f`-error, #842/#854's deferred escapes and the `RECURSE_MAX_ITEMS` cap — a lazy copy
would have been a second place every one of those has to be maintained, so
`builtin_recurse_f`/`builtin_recurse_cond` are now an always-`Continue` sink over it rather than
two more loops. `..` is `Expr::RecursiveDescent`, a separate arm that already streamed, and bare
`recurse` over a live cursor keeps `each_recurse_cursor_generic`.

## A halt leak closed as a side effect

`test_short_circuit_side_effect_leaks_820_932_987` pinned two rows as known divergences, with a
written instruction to move them if a stage landed. Both now match jq byte for byte:

| filter | jq | before | now |
|---|---|---|---|
| `path((0 \| recurse(if . < 1 then .+1 else ("x"\|halt_error(3)) end)) == 1)` | path error, exit 5 | **exit 3, stderr `x`** | path error, exit 5 |
| `path(1 == (0 \| recurse(...)))` | path error, no stderr | stray `x` on stderr | path error, no stderr |

The fix is at the source rather than in the consumer: with `f` unevaluated the `halt_error` is
never reached, so there is no `pending` to keep or drop. Those rows moved to
`test_short_circuit_side_effect_shapes_already_match_jq_820`, and `while` — which has no lazy arm
— takes over pinning `binary_fanout_each`'s inner/outer `pending` asymmetry, the same hand-off
`if` made in #1462 and `foreach` in #2180 WP3.

## Verification

**Attributed differential sweep**, 2464 generated programs (22 `recurse` spellings x 16 wrappers x
7 inputs), each run against jq 1.7.1, this branch, and a build of the merge-base:

```
2464 programs | REGRESSIONS: 0 | fixed: 100 | still divergent: 285
```

Every one of the 285 is the #2918 residual below, byte-identical between base and this branch — no
row changed without landing on jq's answer. A handful of rows are excluded because **jq itself
hangs** on them (`[recurse(.a?)]` recurses forever; jq has no item cap, both succinctly builds
answer `10000` identically) — a pre-existing documented divergence, not this PR's.

Local gate: 8533 tests pass, both CI clippy configs clean, `cargo doc` with `-D warnings`,
`cargo test --no-default-features`, `cargo fmt --check`. **Patch coverage 100%** (92/92).
`cargo test --no-default-features`, `cargo fmt --check`.

## What this deliberately does not close — #2918

`f` still runs to completion at the node the traversal is *at*: jq descends into `f`'s first
output's whole subtree before asking `f` for its second, and an explicit stack of materialized
nodes cannot suspend a push-driven `f` mid-stream. So a bound overshoots by one node's fan-out
(`[limit(2; recurse(.[]?|debug))]` on `[1,[2,[3]]]` writes two DEBUG lines where jq writes one —
it wrote **five** before this PR), and unbounded it shows as stderr *ordering*. Values are
identical in every case, and all three walkers share it, path mode included.

`recurse_still_finishes_a_nodes_f_before_descending_2918` records each of those rows with both
jq's answer and ours and asserts they still differ, so closing #2918 trips the test rather than
passing quietly.

[#2918](https://github.com/rust-works/succinctly/issues/2918) sets out the fork, because it is a
real one: the natural fix is recursion inside `f`'s own sink callback, which makes native stack
depth equal traversal depth — and `RECURSE_MAX_ITEMS` allows 10000, where this repo's own
`MAX_EXPR_DEPTH` note measures the *parser*, a much smaller frame, aborting at around 96 levels in
debug on cargo's 2 MiB test thread. So it is bounded native recursion with a fallback (a second
traversal shape in each of three invariant-laden walkers) versus a pull-based generator protocol
(which would close this, #2908 and the path-side half at once, but is architecture-level).


newhoggy:
Addressed the `/code-review high` findings.

**One real regression, yq mode only** — the review's first finding was correct. `each_recurse_generic` read its root with the cursorless `to_owned`, where the eager wildcard it replaced went through `bridge_ambient_input` → `to_owned_with_cursor`. Only the cursor route applies an explicit YAML tag, so `!!str 123` came back as the number `123`, `!!int "9"` as `"9"`, `!!bool "yes"` as `"yes"` and `!!null ""` as `""`. Fixed by threading the cursor.

Why the suite missed it, which is the part worth recording: the arm is reached **only with a bound** (`first`/`limit`/`nth`). Bare `recurse` and `..` take the cursor-guarded `each_recurse_cursor_generic` above it, and an unbounded `[recurse(.[]?)]` never reaches the lazy arm at all — so every pre-existing yq recurse row exercised code this change does not touch.

`recurse_through_a_bound_still_resolves_explicit_tags_2693` closes that gap. Real yq's lexer rejects `recurse(...)` outright (confirmed live against v4.53.3), so there is no oracle for the filter itself; the test pins internal consistency instead — every bounded spelling must agree with `[.]`, whose answer is byte-identical to real yq's. I verified it **fails** with the cursorless read restored, so it genuinely catches the bug rather than just passing.

Both stale doc comments fixed too: `binary_fanout_each`'s worked example (the test-side copy was updated in the original push, this one was not) and `eval_each`'s "Native lazy arms" enumeration.

Re-verified after a rebase onto current main (which brought in #2267, touching the same generic evaluator): 8540 tests pass, both clippy configs, rustdoc, no_std and fmt clean, patch coverage still **100%** (116/116), and the oracle rows unchanged.

The rebase hit an append-conflict in `tests/yq_cli_tests.rs` that cut mid-function rather than being a clean append-vs-append; the file was rebuilt from upstream's version plus this branch's one test, and the diff against `origin/main` for that file is now exactly those 56 added lines.

Issue #2351, opened 2026-09-04
yq: eval_slice_bound doesn't gate a slice bound's own escaped Partial prefix on yq mode


newhoggy:
## What's wrong

Found during #2326's own code review. `#2226`'s target-side fix and `#2326`'s own
key-side fix both gate their "stream the already-produced prefix before the escaping
control" behavior on `S::TAG != EvalTag::Yq`, since real yq does not stream that prefix
(confirmed live for both). `eval_slice_bound` (`src/jq/eval.rs`) and its `eval_generic.rs`
twin -- which keep a **slice bound**'s own escaped generator prefix (`E[K1:K2]`'s `K1`/
`K2`, as opposed to `E[K]`'s single key) -- have no such gate at all: they unconditionally
keep the prefix in both modes.

Confirmed pre-existing: reproduces identically against `main` before #2326, so this
predates that PR and is unrelated to its own change -- a different function entirely
(slice bounds, not index keys).

## Repro (confirmed live against yq v4.53.3)

```console
$ echo '[1,2,3]' | yq -o json '.[(0,1,error("x")):3]'
Error: x
$ echo '[1,2,3]' | succinctly yq '.[(0,1,error("x")):3]' -o json
[1, 2, 3]
[2, 3]
Error: x
```

Real yq prints only the error; succinctly streams two extra, unwanted array outputs
before it.

## Suggested fix

Same shape as #2226/#2326's own gate: `if S::TAG == EvalTag::Yq { <discard the prefix,
old conservative behavior> }` ahead of `eval_slice_bound`'s (and its `eval_generic.rs`
twin's) own prefix-keeping logic. Needs live-oracle verification for both bounds
(`K1`/`K2` independently) and both evaluators, matching #2226/#2326's own verification
discipline.

newhoggy:
🔧 Claimed by Agatha — branch `issue-2351-yq-slice-bound-partial-prefix-gate`.

Issue #2267, opened 2026-09-03
jq: eager path-resolution vs jq's lazy generator can over/under-fire target side effects when a downstream write fails partway


newhoggy:
Follow-up from #2249's own live-oracle verification (fixed narrowly in PR TBD).

## What was found

`resolve_slice_expr`/`resolve_index_expr` (the `path()`/`=`/`|=`/`del()` write-path
resolvers) both batch *every* resolved path into a `Vec<PathBranch>` eagerly, all at
once, before any caller (`set_path`/`eval_update`/`delete_at_path`) applies a single
write. Real jq's own `S as $s | T as $t | E | .[$s:$t]` (and `K as $k | E | .[$k]`)
desugaring is a genuinely *lazy* generator pipeline: `reduce path(...) as $p (.;
setpath($p; ...))` pulls one path at a time and applies its write immediately, so a
write that fails validation on an early path (`setpath`'s own type check, e.g. "A slice
of an array can only be assigned another array") stops the whole `reduce` before the
path generator is ever asked for a *later* path — meaning `E`'s own side effects for
later `(s, t)`/key pairs never fire at all.

succinctly's eager batch-then-write model has no such short-circuit: every path (and
every side effect `E` produces while being resolved) is already computed *before* any
write validation runs, so a write failure on the first path can never retroactively
"undo" the side effects `E` already fired while resolving paths 2, 3, 4, ...

## Repro (verified live against jq 1.7.1)

```console
$ echo '[10,20,30]' | jq -c '(.|stderr)[(0,1):(2,3)] = 99' 2>&1 1>/dev/null
[10,20,30]jq: error (at <stdin>:1): A slice of an array can only be assigned another array
# stderr fires ONCE -- setpath fails on the very first path (0:2), the (s,t) generator
# never advances to (0:3)/(1:2)/(1:3)

$ echo '[10,20,30]' | succinctly jq -c '(.|stderr)[(0,1):(2,3)] = 99' 2>&1 1>/dev/null
[10,20,30][10,20,30][10,20,30][10,20,30]jq: error (at <stdin>:1): A slice of an array can only be assigned another array
# stderr fires all 4 times -- every (s,t) pair was resolved eagerly (and E's side
# effect fired for each) before setpath ever got a chance to reject the first one
```

Not new to #2249's own fix: this over-firing manifests specifically because #2249 made
`target` genuinely re-evaluate per pair (matching jq for `path()`/`del()`/`|=`, which
[...]

Issue #2248, opened 2026-09-02
jq: resolve_index_expr/resolve_slice_expr's untrackable-branch check discards already-successful target_branches output


newhoggy:
**Severity: Medium** (silent output loss on writes/`path()`, not data corruption — the whole computation raises with a truncated prefix instead of the correct partial one; found during #2245's own live-oracle verification, not a user report)

## Symptom

Both `resolve_index_expr` (the `path()`/`=`/`del()` write-path resolver for `E[K]`) and `resolve_slice_expr` (the same for `E[S:T]`) resolve `target` (`E`) once into `target_branches`, then check whether *any* branch is untrackable:

```rust
if let Some(PathBranch { value: first_value, .. }) =
    target_branches.iter().find(|b| !trackable || !b.trackable)
{
    return Err((Vec::new(), EvalError::invalid_path_expression_near_access(...).into()));
}
```

`Vec::new()` here discards every branch that came *before* the untrackable one in `target_branches`, even though those earlier branches are perfectly valid, trackable, already-resolved paths. Real jq processes each target branch's slice/index application individually and only raises once it actually reaches the untrackable one — so a comma-joined target where an untrackable branch isn't first still produces output for every branch before it.

## Repro (verified live against jq 1.7.1, `/usr/bin/jq`)

`resolve_slice_expr` (`E[S:T]`):
```console
$ echo null | jq -c 'path((.,5)[(0,1):(2,error("mid"))])'
[{"start":0,"end":2}]
jq: error (at <stdin>:1): Invalid path expression near attempt to access element {"start":0,... of 5

$ echo null | succinctly jq -c 'path((.,5)[(0,1):(2,error("mid"))])'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element {"start":0,... of 5
# (no stdout at all -- the `.` branch's own [{"start":0,"end":2}] output is silently dropped)
```

`resolve_index_expr` (`E[K]`) has the identical shape:
```console
$ echo null | jq -c 'path((.,5)[(0,error("mid"))])'
[0]
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 0 of 5

$ echo null | succinctly jq -c 'path((.,5)[(0,error("mid"))])'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 0 of 5
# (no stdout -- [0] dropped)
```

## Confirmed pre-existing, not a regression

[...]

PR #2922, opened 2026-09-13
fix(jq): interleave a computed index's key with its target, as jq does

newhoggy:
## Summary

The second PR of #2267's plan. Scope changed during review — see "What was
backed out" below.

`resolve_index_expr` (the `path()`/`=`/`|=`/`del()` resolver for `E[K]`) drained
`K` in full before reaching `E` for any key. jq compiles the same expression as
`K as $k | E | .[$k]`, so `E` runs before the next `k` is asked for. This is the
sibling of the slice-bound gap #2920 closed, and the one `limitations.md`
already attributed to this issue by name.

Captured live from `/usr/bin/jq` 1.7.1 (stderr only, `1>/dev/null`), on
`{"a":1,"b":2}`:

| query | jq 1.7.1 | before |
|---|---|---|
| `path((.\|debug("E"))[("a"\|debug("ka")),("b"\|debug("kb"))])` | `ka, E, kb, E` | `ka, kb, E, E` |
| `path((select((true,error("terr"))))[("a"\|debug("ka")),("b"\|debug("kb"))])` | `ka, terr` | `ka, kb, terr` |

The second row is the stopping half, and the reason the sink answers with
`Demand` rather than draining and discarding: once a key's `E` escapes, jq never
resumes the key generator, so `kb` is not skipped — it is never evaluated.

`K` drives through `drive_index_key`, `drive_slice_bound`'s sibling. Unlike the
slice bound this streams in **both** modes: #2351's discard-on-escape rule is
scoped to that call site by its own doc comment precisely because this one "must
keep its own prefix in yq mode too".

**Only ordering moves** — both shapes produced the same path outputs, in the
same order, before and after, so the tests pin stderr content and count
alongside stdout. #843's passthrough refusal moves onto the first delivered key
(it needs a key to name, and `out` is necessarily still empty there, so its
empty `Err` prefix is preserved exactly). #2139's per-key target re-resolution,
#896's key-and-target partial prefixes, #2248's output-loss rule and the
`target` > `key` escape priority are unchanged and re-verified.

## What was backed out, and why it matters

This PR originally also re-resolved an assignment's path once per right-hand-side
output (the issue's repro D), matching jq's `_assign(paths; $value)` binding
`$value` as the outer generator. It was implemented, reviewed, A/B-fuzzed against
the merge-base, and **removed**: it is not independently correct.

jq's RHS generator is lazy; `collect_rhs_outputs` is not. With every output's
document built eagerly, re-resolving per output fires the target's side effects
for outputs a downstream consumer never pulls:

```console
$ echo '{}' | jq -c 'first((.|debug("E"))["a"] = (1,2))'
  jq 1.7.1:  E once          with the change:  E twice
```

and with `input` in the path it stops being a side-effect-count question and
changes **stdout**. A 5,000-shape fuzz sweep put it at 87 regressions against 69
fixes on the fully-consumed shapes. The two halves are one mechanism and have to
land together.

`first((.|stderr)["a"] = (1,2))` is pinned as a holdout in this PR's tests, so a
future attempt that reintroduces this fails rather than passing quietly, and
`limitations.md` records the constraint so the next attempt starts from it.

## Still open on #2267

- **Repros A/B** — a failing write must stop the path producer.
  `Expr::IndexExpr`/`Expr::SliceExpr` are still `resolve_node_eager` arms, so
  every branch is produced before any reaches a write sink. Making them native
  `resolve_node_sink` arms would also change the `try_reserve_product_labeled`
  cross-product guard's contract (a streaming resolver has no accumulator to
  reserve against, so the guard becomes per-element growth rather than an
  up-front refusal) — and that guard exists under ADR-0018's "would take the
  host process down" exception, so it wants its own rationale.
- **Repro D**, as above.

Both recorded in `docs/compliance/jq/limitations.md` and the CHANGELOG.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — 8544 passed, 0 failed
- [x] `cargo fmt --check`
- [x] 30-row live differential sweep against `/usr/bin/jq` 1.7.1 (stdout, stderr
      content/count, exit code): the two rows above, the #2139/#896/#2248/#843
      rules, the `first`/`limit` shapes the backed-out half regressed, #2920's
      slice rows, and static `=`/`+=`/`|=`/`del`/`setpath`/`getpath`/`delpaths`/
      `paths`/`flatten` regressions. 29/30 match; the one that does not is repro
      D itself, deliberately left open and documented.
- [x] 3 new CLI regression tests, one of them a must-not-change holdout

Refs #2267


newhoggy:
Filed #2924 for the `/code-review` pass's fourth finding — the two escape-stashing sinks (this PR's key sink, and `resolve_slice_expr`'s pre-existing one on `main`) answer `Demand::Stop` without `mark_nonretryable_escape`, contrary to the contract `is_retryable_stop` states.

Not fixed here deliberately: it is latent (no repro across ~9,000 fuzzed shapes), one of the two sites predates this PR, and changing `?//` retry classification without an oracle capture to pin it is exactly the speculative change an ordering PR should not be making.

Issue #2691, opened 2026-09-10
jq: resolve_terminal's violation reset relies on an unverified re-invocation invariant


newhoggy:
**Severity: Low** (unconfirmed — a code-smell risk surfaced while reviewing #2649, not a reproduced bug)

## Summary

`resolve_terminal` (`src/jq/eval.rs`) resets its captured `violation` slot to `None` at the
top of every sink invocation:

```rust
&mut |branch| {
    violation = None;
    if !branch.trackable {
        ...
        violation = Some(...);
        return Demand::Stop;
    }
    kept.push(branch);
    Demand::Continue
},
```

This relies on an implicit invariant: the sink is only ever re-invoked *after* a prior
`Demand::Stop` in one documented case — a `?//` alternative retry (where the caller
deliberately continues past a stopped branch to try the next alternative). If any other
producer feeding into this sink across the many call sites of `resolve_node_sink` ever
re-invokes it after `Demand::Stop` outside that retry path, an already-recorded
untracked-navigation violation would be silently cleared by the next branch's reset —
turning a case jq refuses into a false success on `del`/`=`/`|=` (a data-corruption risk,
not just a wrong-output one).

## Status

Not independently reproduced — this is a structural concern raised during #2649's code
review, not a confirmed failing repro. `resolve_terminal` itself was **not** touched by
#2649's diff (predates it entirely), so it isn't a regression from that PR; it's flagged
here because the invariant is easy to violate silently and the function is reached from a
large number of call sites (`resolve_node_sink`'s ~58 producers).

## Ask

Either:
- Find a concrete repro where a second sink invocation after `Demand::Stop` reaches this
  reset outside the `?//`-retry path (would confirm a real bug), or
- Determine the invariant is actually enforced elsewhere (e.g. `resolve_node_sink` never
  calls a stopped sink again except for that one case) and tighten the comment / add a
  debug assertion so the invariant is checked rather than assumed.

newhoggy:
🔧 Claimed by agent `Cadence` — branch `issue-2691-resolve-terminal-violation-reset`

newhoggy:
Fix in #2768. The answer to the issue's fork is the second option — the invariant is enforced, not violated — but with one correction to how it was stated.

[...]

Issue #2061, opened 2026-08-31
perf(jq): path-context queries still materialize the whole document once (the residual half of #1909)


newhoggy:
**Severity: Low-Medium** (perf only; no incorrect output). The residual half of #1909,
split out because #1909's own title and framing ("dwarfing #1850's own fix") describe a
state that #2045 has now changed.

## What #2045 did and didn't fix

#2045 removed the *second* whole-document materialization on the path-context path: the
reindex bridge (`eval_generic.rs`'s `eval_on_owned`) used to serialize the tree to JSON,
rebuild a `JsonIndex`, and re-enter `eval.rs`, landing in `eval_pipe`/`builtin_path`,
whose own `needs_path_context` gate materialized the same document again.

The **first** materialization is untouched and is what this issue is about.
`eval_generic.rs`'s `Expr::Pipe` and `Builtin::Path` arms still call
`to_owned_with_cursor` over the whole input before any path expression runs, because
`resolve_node`/`resolve_seq`/`PathBranch` operate over `&'a OwnedValue` and need stable
references to build zero-copy `Cow::Borrowed` branches. The lazy cursor evaluator has no
equivalent path-tracking machinery.

## Measured, on `main` *after* #2045

Release build, interleaved, medians of 5, 1,000,000-element array of `{"k":i,"v":"x"}`
(21 MB):

| query | time | peak RSS |
|---|---|---|
| `length` (parse only, no path context) | 0.104s | 34.0 MiB |
| `.[0]` (plain read) | 0.104s | 34.0 MiB |
| `path(.[0])` | **1.062s** | **519 MiB** |
| `path(.)` | **1.133s** | **1003 MiB** |

`path(.)` outputs the single value `[]` and `path(.[0])` outputs `[0]`. Against the
parse-only baseline that is still ~10x time and 15-30x peak RSS, all of it the one
remaining `to_owned_with_cursor`. The ratios hold at 30k and 300k too, so it is the O(n)
term, not a constant.

## Scope

**In:** `path(...)`, `key`, `parent`, `parent(n)`, `file_index` — output is small and
bounded, so the O(document) materialization buys nothing.

**Out, deliberately:**

- The write family (`.[0] = 9`, `del(.[0])`, `.[0] |= f`) is inherently O(document) — its
  *output* is a whole modified document. Avoiding that needs streaming-with-splice off the
  semi-index, a much larger and separate idea.
- `getpath` is tracked in #2053, and **not** because it is a smaller version of this: its
[...]

PR #2923, opened 2026-09-13
fix(jq): stream path()'s results so a consumer outside it can stop the generator

### LINKED ISSUE #2908: jq: path()'s own consumer side still collects, so a bound outside path() can't stop the fold (#2694 residual)

newhoggy:
Fixes #2908

## The gap

`path(f)` is a generator like any other in jq: a consumer satisfied by the first path never
asks for the second, and `f` is never resumed to produce it. Both evaluators collected every
path before their consumer saw one, so a bound *outside* `path(...)` could only truncate a
finished list — and every output of `f` had already run.

```console
$ echo '{"a":1,"b":2,"c":3}' | jq -c '[limit(1; path((.a|stderr), (.b|stderr), (.c|stderr)))]'
1
[["a"]]
```

succinctly wrote `1`, `2` and `3`.

**Two corrections to the issue as I filed it**, both found by probing before implementing:

- It is **not fold-specific**. I wrote it up from a `foreach` repro, but a plain comma shows it
  just as well — any multi-output `f` inside `path()` ran to exhaustion. Pure navigation
  fan-out (`path(.[])`, `path(.a[])`) always agreed, because nothing in it is observable per
  element.
- My first probe suggested a fixed one-element overshoot. That was an artefact: the second fold
  step happened to error and abort the fold. With a non-erroring UPDATE the overshoot is the
  whole source — `n=8` ran all eight elements for one requested path.

## The change

`each_path_on_owned` interleaves resolution and walking: each resolved branch is walked and
emitted before the next is asked for. That is jq's own order, and it is what carries the
consumer's `Demand::Stop` back into the generator. It is reached through new `eval_each` /
`eval_each_generic` arms.

The first commit is pure restructuring with no behaviour change: `resolve_terminal` and
`resolve_dynamic_indexes` each become an always-`Continue` sink over a streaming twin, so the
untracked-branch refusal, #2691's retry rule and the trailing-iterate deferral keep one
definition rather than two. The write-side callers (`=`, `|=`, `del()`) keep the collecting
entry points — they want every branch and have nothing to stop for.

The generic arm's guard **mirrors the eager arm's own two gates** rather than restating the
decision, so a cursor-navigable expression keeps #2061/#2168's no-materialization walk and a
document the reindex bridge would not round-trip identically keeps the bridge.

## Two rules this had to preserve, and both were nearly lost

- **#2680's walk prefix.** `walk_path` fills its buffer as it goes, so a failing walk still has
  to emit what it reached. My first draft returned on the error without draining that buffer,
  which lost `path((.a[] | .b) | .c[0:1])`'s `["a",0,"b","c",{"start":0,"end":1}]` entirely and
  made `first(...)` of it fail where jq succeeds. Caught by the suite (4 failures).
- **A stop is not exhaustion.** Folding the two made `isempty(path(.a,.b))` answer *twice* —
  `false` for the item it took, then `true` for the exhaustion it was wrongly told about. Ten
  shapes hit it. Caught by the differential sweep, **not** by the suite.

## Verification

Attributed differential sweep — 2639 generated programs (29 inner expressions x 13 wrappers x 7
inputs), each run against jq 1.7.1, this branch, and a build of the merge-base:

```
2639 programs | REGRESSIONS: 0 | fixed: 52 | still divergent: 0
```

Every shape in the sweep now matches jq exactly, including `del(...)` and `|=` rows that
exercise the shared collecting entry points the refactor touched.

Local gate: 8542 tests pass, both CI clippy configs clean, `cargo doc` with `-D warnings`,
`cargo test --no-default-features`, `cargo fmt --check`. **Patch coverage 100%** (95/95).

## Note for #2918

#2918's text listed this issue as needing the same pull-based generator protocol it describes.
That turned out to be wrong and the limitations bullet now says so: `path()`'s branches were
already produced lazily by `resolve_node_sink`, and only the terminal was collecting them.
#2918's own residual — suspending `f` *between* its outputs — is unaffected and still needs
that design decision.


newhoggy:
Addressed the `/code-review high` findings. All four were correct; the two substantive ones are **not regressions** — both shapes behave exactly as they did on the merge-base — but they do mean my "still divergent: 0" and "closed" claims were too broad, so they are now filed, documented and pinned rather than left implied.

**Filed as #2925**, with both repros:
- the **reindex-bridge route** — a document the bridge won't round-trip identically takes the collecting path, so the *document* selects the route, not the filter: the same `[limit(1; path((.a|stderr),(.b|stderr)))]` writes `1` on `{"a":1,"b":2}` and `12` once a 300-digit literal is added;
- a **generator in index position** — `[limit(1; path(.[("a"|stderr),("b"|stderr)]))]` writes `ab` where jq writes `a`, which is `resolve_index_expr`'s eager key evaluation, the same function #2032 sits in.

My sweep never generated either shape, which is exactly why it reported zero. Both now sit in `path_results_stream_to_their_consumer_2908`'s residual block, carrying jq's answer and ours and asserting they still differ, so closing either trips the test instead of passing quietly. `limitations.md` no longer says the consumer side is closed outright.

**Double materialization** fixed: the bridge fallback re-entered `eval_single`, re-running `RootWitness::of` + `demote_rebuilt_markers` + `to_owned_with_cursor` over the whole document to rebuild what the arm had just built. It now hands `eval_on_owned` the `owned` it already has, as the eager arm does.

**Doc comment placement** fixed — the new arm had been inserted under #2693's comment, leaving that rationale reading as the preamble to `path(f)` and the `Recurse*` arms uncommented.

Re-verified: 8542 tests, both clippy configs, rustdoc, no_std and fmt clean; sweep unchanged at 2639 programs / 0 regressions / 52 fixed; patch coverage 100%.

Issue #2546, opened 2026-09-06
jq: eval_slice_expr / path-context IndexExpr siblings still eagerly materialize their computed-bound generators (#2138 follow-up)


newhoggy:
**Severity: Low**

## Summary

#2138 fixed `eval_index_expr`'s eager key-generator materialization in both `eval_generic.rs` (the ordinary CLI streaming path) and `eval.rs` (the public library API / `yq -n`/`-R`/`--slurp`/`--inplace`/`--split-exp` bridge): real jq's `K as $k | E | .[$k]` compilation stops pulling the key generator `K` the moment an earlier key's own indexing step raises, and both evaluators used to keep evaluating `K`'s later elements (including their side effects) regardless.

At least four structurally identical siblings share the same "eagerly collect a computed-bound generator via `eval_single`/`.materialize_cursor()`/`drain_path_context_stream_routed`, then loop" shape and were **not** touched by #2138:

- `eval.rs::eval_slice_expr` (`E[S:T]`'s `starts`/`ends` bound generators)
- `eval_generic.rs::eval_slice_expr` (its own twin)
- `eval.rs::eval_index_expr_with_path_context`
- `eval.rs::eval_slice_expr_with_path_context`

## Root cause

Each of these evaluates its own generator-shaped argument (slice bounds, or a path-context key/bound) fully up front via the eager (`eval_single`-family) route before any per-value indexing/navigation begins, the exact pre-#2138 architecture `eval_index_expr`'s own two copies were rewritten away from (see #2138's PRs for the `eval_each`/`eval_each_generic` sink-pull pattern and the `Item`/`Demand`/`Flow` (`GenericItem`/`Demand`/`Flow`) protocol reused for it).

## Why this is separate from #2138

#2138 was scoped to `eval_index_expr` specifically (the reported repro's exact shape, `.[K]`). These four siblings need the same kind of rewrite but are structurally distinct functions (different accumulator shapes -- `eval_slice_expr`'s `S`-outer/`T`-middle/`E`-inner triple nesting vs. `eval_index_expr`'s single key-outer/target-inner pair; the path-context siblings' own `drain_path_context_stream_routed`/cursor-walk machinery) that warrant their own dedicated fix and review rather than folding into an already-large PR.

## Suggested fix direction

[...]

Issue #2283, opened 2026-09-03
jq: nested def's bare-vs-$-style parameter can't be told apart post-parse, so its outer-$var shadow check can only be correct one way


newhoggy:
## What's wrong

`parse_func_def_parts` discards a function parameter's leading `$` at
parse time -- `def f(g): ...` and `def f($g): ...` both produce the
identical `params: ["g"]`. This is fine for a parameter's own
call-site substitution (the argument value is the same either way), but
it makes two *different* shadow-check questions structurally
indistinguishable, in two places:

1. **`substitute_var_impl`'s `Expr::FuncDef` arm** (`src/jq/eval.rs`) --
   deciding whether an *outer* `$var`-substitution reaches into a nested
   def's own body.
2. **`substitute_func_param_impl`'s `Expr::FuncDef` arm** (`src/jq/eval.rs`,
   the `body_shadowed` check) -- the same question, for an *outer bare
   parameter*'s own substitution.

For a **bare** nested parameter (`def h(g): ...`), the outer `$g`/`g`
should still reach into `h`'s body (no shadowing) -- `h`'s own `g` has no
`$`-binding of its own. For a **`$`-style** nested parameter
(`def h($g): ...`), the outer value must *not* reach in -- `$g` inside
`h`'s body is `h`'s own argument, and letting the outer substitution reach
it first permanently bakes the outer value into `h`'s body before `h` is
ever called, leaving nothing for `h`'s own later argument substitution to
replace.

Both functions currently pick the "always shadow" answer (correct for the
`$`-style case, wrong for the bare case) -- confirmed live against jq
1.7.1:

```console
$ jq -n '3 as $g | def f(g): $g; f(1)'
3
$ succinctly jq -n '3 as $g | def f(g): $g; f(1)'
1
```

## Why this isn't fixed as part of #2141

#2141 fixed the *unambiguous* half of this same conflation (a bare
top-level parameter being wrongly shadowed by an unrelated local `$`-var
binding, `As`/`Reduce`/`Foreach`/`AsPattern`). This nested-`def` case is
genuinely ambiguous with the current AST -- a first draft of #2141's own
fix tried "always substitute" here and broke the `$`-style shape instead
(`3 as $item | def double($item): $item * 2; double(10)` regressed from
the correct `20` to `6`), caught in that PR's own review. Reverted rather
than shipped, since the `$`-style shape is the more common one and
already covered by existing tests.

## Suggested fix direction

[...]

PR #2926, opened 2026-09-13
fix(jq): resolve a non-numeric slice bound over a null target, refuse at the write

### LINKED ISSUE #2853: jq: path mode raises on a non-numeric slice bound over a null target where jq resolves the path (#2546 residual)
**Severity: Low** (`path()` output and a `del()` no-op on a `null` target only; every write jq refuses errors identically)

## Summary

In path mode a computed, non-numeric slice bound over a **null** target is raised at resolution time, where jq resolves the path and leaves it to the write to refuse. jq's `INDEX` opcode answers `null` for a null target without parsing the descriptor's bounds, so the path exists (`?` or not) and only `setpath` later rejects it. succinctly's `Expr::Slice` path component holds integer bounds only (`start: Option<i64>`, `end: Option<i64>`, plus a `NumberKey` for float spelling), so there is no representation for `{"start":"x","end":null}` to carry through a `PathBranch`.

## Repro

All captured live against `/usr/bin/jq` 1.7.1, input `null`:

```
path(.["x":])                jq: [{"start":"x","end":null}]      succinctly: Array/string slice indices must be integers
path(.["x":]?)               jq: [{"start":"x","end":null}]      succinctly: same error
[path(.[(0,"x",1):]?)]       jq: [[{"start":0,"end":null}],[{"start":"x","end":null}],[{"start":1,"end":null}]]   succinctly: same error
del(.["x":])                 jq: null                            succinctly: same error
del(.["x":]?)                jq: null                            succinctly: same error
del(.[(0,"x",1):]?)          jq: null                            succinctly: same error
.["x":] = 5                  jq: Array/string slice indices must be integers   succinctly: same (agrees)
.[("x"):2]? = 5              jq: same error                      succinctly: same (agrees)
.[("x"):2]? |= 5             jq: same error                      succinctly: same (agrees)
```

Controls: `null | path(.[1:2])` is `[{"start":1,"end":2}]` in both; `[1,2] | [path(.[(0,"x",1):]?)]` is the `0`/`1` descriptors in both (fixed by #2546).

## Root cause

`resolve_slice_expr`'s per-branch bound ruling (`src/jq/eval.rs`, the `SliceTargetKind::Null` arm added by #2546) raises a failed `ComputedSliceBound` unsuppressed for a null target, because the `Expr::Slice` component it would otherwise build cannot hold a non-numeric bound. `PathSliceBound::raw` already keeps the value for the error message, so the value is on hand — only the component representation is missing. Real jq keeps the descriptor as an ordinary `jv` object in the path, so `path()` prints it and `delpaths` on `null` no-ops.

## Why this is separate from #2546

#2546 moved the bound ruling to the slice step in jq's order for both value-mode evaluators and the path resolver; this is the one path-mode target kind where "jq's order" needs a *path component* that does not exist yet, i.e. a representation change (`Expr::Slice`'s bound fields or `NumberKey`) rather than an ordering change — and `Expr`'s size is pinned. Recorded in `docs/compliance/jq/limitations.md` ("A computed bound is ruled on at the slice step, after the target's kind").

## Suggested fix direction

Either let `Expr::Slice`'s `start_key`/`end_key` (or a sibling field) carry an arbitrary `OwnedValue` bound for the path-only null case, so `slice_component_value` renders it and `through_slice`/`set_path` raise "Array/string slice indices must be integers" when a write reaches it while `del()` no-ops on `null`; or add a path-only component variant for a non-numeric slice descriptor. Check `Expr`'s size pin (`test_expr_size`) and the `#1326` float-spelling tests either way.

newhoggy:
## Summary

jq's `INDEX` opcode answers `null` for a null target **without reading the
slice descriptor's bounds**, so a bound jq never parsed resolves into the path
verbatim — `?` or not — and only the eventual write refuses it. succinctly
raised `Array/string slice indices must be integers` at *resolution*, which
matched jq for every write it refuses but was wrong for `path()` (jq reports
the descriptor) and `del()` (jq answers `null`).

All rows captured live from `/usr/bin/jq` 1.7.1, input `null`:

| filter | jq 1.7.1 | before |
|---|---|---|
| `path(.["x":])` / `path(.["x":]?)` | `[{"start":"x","end":null}]` | the integers error |
| `path(.[1.5:"y"])` | `[{"start":1.5,"end":"y"}]` | the integers error |
| `path(.[{}:])` / `.[[]:]` / `.[true:]` | the value verbatim | the integers error |
| `[path(.[(0,"x"):("y",1)])]` | all four cross-product descriptors | the integers error |
| `del(.["x":])` | `null` | the integers error |
| `.["x":] \|= empty` | `null` | the integers error |
| `.["x":] -= 5` | `null (null) and number (5) cannot be subtracted` | the integers error |
| `.["x":] \|= (debug\|5)` | `["DEBUG:",null]`, then the integers error | the error, no DEBUG |
| `.["x":] = 5`, `\|= 5`, `+= 5`, `.["x":][0] = 5`, `.a["x":] = 5` | the integers error | agreed already ✅ |

## How

**Representation.** `SliceBoundKey` widens `Expr::Slice`'s two key fields:
`Number(NumberKey)` is #1326's float spelling, `Raw(Box<OwnedValue>)` is a
bound jq never parsed, whose `i64` side is always `None`. `Expr::Index`'s key
keeps `NumberKey` directly, so `NumberKey::value` stays total and
`index_component_value` keeps one definition of #1088's rule.

`size_of::<Expr>()` is **unchanged at its pinned 96 bytes** — measured in-tree
*before* any semantics were written, along with the #1156 depth pin, because
#2283's lesson is that growing `Expr` overflows the CLI's real stack at a depth
that was previously safe.

**A field that disappears.** `PathSliceBound`'s `key`/`raw` pair collapses into
one `Option<SliceBoundKey>`: they were provably exclusive (a key exists only
for a `Float`/`NumberLiteral`, the raw value only for a *non*-number), so the
`unreachable!()` its rendering needed for "failed, but nothing kept" — a
coverage-tolerated line — goes with them. The descriptor built there and the
one rebuilt from the resolved component now render through one
`slice_bound_component_value`, so they cannot disagree.

**The refusal moves to the write**, as a `through_slice` arm gated by a
`non_integer_bound` flag that all six call sites compute through one
`slice_has_non_integer_bound` predicate — narrow on purpose, since a `Number`
key is a perfectly good integer bound and widening it would break
`.[1.5:] = [9]`.

`edit` runs **first**, against a throwaway `Null`, and its outcome propagates
before either disposition is chosen — the #1876/#1883 rule the string-slice arm
already states. That is what makes `-= 5` raise the arithmetic error from
inside the filter, `|= (debug|5)` print its DEBUG line, and `|= empty` no-op the
way jq's `_modify` fallback to `delpaths` does. The throwaway is `Null`, not
`[]`, which the DEBUG payload pins. Not gated on `terminal_write`: mid-chain,
`edit` is the inner walker whose own `wrote` answers the same question.

**`del()` needed no code**, which the tests pin rather than assume: the deletion
trie turns the component into `ArrayStep::Slice(None, None)`, and
`null | del(.[0:1])` already no-ops.

**yq mode is untouched and unreachable by construction**: `drive_slice_bound`
raises a failed classification at the pull site when `S::TAG == Yq`, so the
null-target-plus-`Err` combination never arises. Verified live that
`del(.a["x":])` and `.a["x":] = 5` still raise there (real yq's own
`strconv.ParseInt` wording divergence is pre-existing and out of scope).

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — 8550 passed, 0 failed
- [x] `cargo fmt --check`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` (new public type) and
      `cargo check --no-default-features` (`no_std`)
- [x] `test_expr_size_is_pinned_1401` (96, unchanged) and
      `test_expr_depth_limit_returns_parse_error_not_overflow_1156`
- [x] **54-row live differential sweep** against `/usr/bin/jq` 1.7.1 over `null`,
      array, object, string and nested-null inputs — every row above, plus the
      full control list (`path(.[nan:])`, `path(.[1e400:])`, `getpath`/`setpath`
      of a raw descriptor, ordinary slice writes/deletes, string-slice rows, and
      `[1,2] | [.[(0,"x",1):]?]` from #2546). All 54 match.
- [x] 4 new CLI regression tests, one of which is the negative control for the
      predicate's width

Fixes #2853


newhoggy:
Both `/code-review` findings addressed.

**Finding 1 was a real bug**, and a good catch: the new arm answered `Ok(true)` where the sibling `Null` arm answers `Ok(false)`. That `bool` is "did a real value reach this slot", not "did this call succeed" — so claiming a write that never happened kept a vivified `{"a":null}` where jq answers `null`, and in a fan-out silently **undid an already-completed delete** (`{"a":null,"b":1} | (.a, .a["x":]) |= empty` → `{"b":1,"a":null}` vs jq's `{"b":1}`). Eight shapes, all fixed by the one-word change, each now pinned by its own test.

The reviewer's diagnosis of *why* my tests missed it is exactly right and worth recording: they exercised `|= empty` only with the slice as the whole path — the one shape with no enclosing consumer of that `bool`.

**Finding 2 is documented, not fixed**, deliberately. It is not a regression (the resolver raised the same sentence for the same program before this PR), and the suggested narrowing — gating on `matches!(root, OwnedValue::Null)` — would be actively unsafe: a `SliceBoundKey::Raw` bound carries `start: None, end: None`, which every arm below reads as *the whole slice*, so falling through would silently splice an array target rather than refuse it. A real fix has to answer the kind-dependent error explicitly. Filed as #2927 with the oracle matrix it needs, and the reasoning recorded on the arm itself.

Re-verified after both: full suite 8551 passed, clippy clean on both feature sets, and a 69-row live differential sweep against `/usr/bin/jq` 1.7.1 — 68 match, the one that does not being #2927 itself.

PR #2928, opened 2026-09-13
fix(jq): resolve a path-mode fold's INIT by demand

### LINKED ISSUE #2903: jq: path(foreach(...)) INIT still eager, now inconsistent with value-mode foreach (#2668 review finding)
**Severity: Low** (surprising output cardinality/side-effect timing in `path(foreach(...))`, no data loss)

Review finding from #2668/PR #2900 (cross-file tracer angle), confirmed live.

#2668 made value-mode `foreach`'s INIT position demand-forwarding (`ForeachInitDrive`, threaded through `foreach_forks`). `resolve_foreach` (`src/jq/eval.rs`) — the **path-mode** `path(foreach(...))`/assignment-target evaluator, reached via `resolve_node_sink` — has its own, entirely separate fold loop (`FoldRegister`, `drive_fold_source`, its own path-tracking machinery) and was not touched by #2668. It still resolves INIT eagerly via `resolve_node` (the collecting wrapper around `resolve_node_sink`) at line ~32180, with its own `init_branches.is_empty()` short-circuit.

Before #2668, both value-mode and path-mode `foreach` collected INIT eagerly, so the two evaluators *agreed* with each other (both diverging from jq the same way). #2668 fixed value-mode only, so the two evaluators now **disagree with each other** — a new inconsistency #2668 introduces, on top of path-mode's own pre-existing (unchanged) divergence from jq.

Confirmed live (`/usr/bin/jq` 1.7.1 vs. the branch built for PR #2900):

```
$ echo '{"a":1}' | jq -c 'first(path(foreach (1) as $i ((.a, ("I"|stderr)); .; .)))'
["a"]                      # no stderr write -- jq never reaches INIT's 2nd fork
$ echo '{"a":1}' | succinctly jq -c 'first(path(foreach (1) as $i ((.a, ("I"|stderr)); .; .)))'
I["a"]                     # writes "I": INIT's 2nd fork evaluated anyway (path-mode, unfixed)
$ echo '{"a":1}' | succinctly jq -c 'first(foreach (1) as $i ((.a, ("I"|stderr)); .; .))'
1                          # value-mode: correct, matches jq (fixed by #2668)
```

**Why #2668 didn't touch this.** This was an explicitly anticipated risk in #2668's own triage plan (comment dated 2026-09-12, risk item (d)): *"INIT-as-driver changes `foreach_forks`'s signature for the resolver-side `resolve_foreach`? — check: it has its own fold loop and `drive_fold_source`; leave it alone unless it calls `foreach_forks` (today it does not)."* That check held — `resolve_foreach` does not call `foreach_forks` and was correctly left alone by #2668's own scope. This issue tracks giving `resolve_foreach`'s own INIT resolution the equivalent demand-forwarding treatment as a follow-up, now that the gap between the two evaluators is visible.

**Fix criterion**: `resolve_foreach`'s own INIT resolution needs the same "drive one output at a time, thread a consumer's stop back into INIT's own generator" treatment #2668 gave `foreach_forks`, adapted to `resolve_node`'s path-tracking machinery (`FoldRegister`, `fold_source_ambient`, `drive_fold_source`) rather than `eval_each`'s owned-value one. The example above should flip to jq's answer (no stderr write), and value-mode/path-mode should agree again.

**Risks**: this is a separate, independently-maintained fold implementation (path-context tracking: `FoldRegister`, snapshot/trackable propagation, register relocation) — not a simple reuse of #2668's own `ForeachInitDrive`/`foreach_forks` machinery. Expect this to need its own design pass, similar in shape to #2668 itself but working against `resolve_node_sink`'s own sink contract instead of `eval_each`'s.

**Class:** Opus — mirrors #2668's own reshape, but against a second, independent fold implementation with path-tracking invariants of its own (#1590/#1591/#2031/#2388/#2235 all cited in `resolve_foreach`'s own doc comment).

**Dependencies / related:** #2668 (closed the analogous value-mode gap), #2899 (the sibling `reduce`-INIT gap #2668 also left open), #2235 (the source-side demand-forwarding `resolve_foreach` already has, which this would extend one level further to INIT).

newhoggy:
Fixes #2903

## The gap

`resolve_foreach` collected every INIT fork before running any of them, so a consumer satisfied
by an earlier fork still paid for every later one:

```console
$ echo '{"a":1}' | jq -c 'first(path(foreach (1) as $i ((.a, ("I"|stderr)); .; .)))'
["a"]
```

jq writes nothing — it never reaches INIT's second fork. succinctly wrote `I`.

#2668 gave *value*-mode `foreach`'s INIT a demand-forwarding drive but could not reach this one:
it has its own fold loop (`FoldRegister`, `drive_fold_source`, its own path tracking) and does not
call `foreach_forks`, which #2668's triage plan checked and left alone on that basis. Before #2668
both evaluators collected INIT eagerly and so agreed with each other; afterwards path-mode
disagreed with value-mode as well as with jq. It agrees again now.

## The change

INIT is resolved through `resolve_node_sink`, with each fork's whole fold run inside that sink
before the next INIT output is asked for. **The per-fork body is unchanged** — it already lived in
a closure driven by `drive_fold_source` — so this moves the *outer* loop, not the fold logic under
it. The two fork-level exits stash their verdict and stop INIT rather than returning through it,
and that stashed verdict outranks INIT's own trailing escape, which jq never reaches once a
consumer has stopped. An INIT escape with no stop still surfaces after every fork it preceded,
exactly as the collecting `flow_result(init_escape)` tail did.

**`resolve_reduce` is fixed too, though #2903 names only `foreach`.** It is the same loop in the
adjacent function and diverged identically — `first(path(reduce (1) as $i ((.a, ("I"|stderr));
.)))` wrote `I` as well. Fixing one and not the other would have left the two path-mode folds
inconsistent with each other, which is the exact complaint #2903 makes about path versus value
mode.

## A note on the third commit

Writing the closure inline in `resolve_node_sink`'s argument list pushed the whole ~250-line fork
body in by four spaces. Nothing about those lines changed, but patch coverage counted every one of
them as **new**, which surfaced five pre-existing `stop_with_escape` arms (fold-budget exhaustion,
decode failure) as new uncovered lines. They are not cheaply coverable — `REDUCE_FOREACH_MAX_STEPS`
is 100,000 and `range` hits its own lower cap first — so rather than tolerate-marking five lines
that this PR does not touch, the closure is now bound before the call, which keeps the body at its
original indentation. Diff: **56 new lines instead of 248, and patch coverage 100%**.

## Verification

Attributed differential sweep — 5775 generated programs (15 INIT shapes x 7 fold spellings x 11
wrappers x 5 inputs), each run against jq 1.7.1, this branch, and a build of the merge-base:

```
5775 programs | REGRESSIONS: 0 | fixed: 288 | still divergent: 2
```

Both remaining divergences are `[reduce (1) as $i ((.a, ("I"|stderr)); .a)]` in **value** mode,
byte-identical between base and this branch — that is #2899, which needs a native `Expr::Reduce`
dispatch arm before a stop has anywhere to land. It is pinned by a trailing assertion in the new
test so closing it trips rather than passing quietly.

Local gate: 8546 tests pass, both CI clippy configs clean, `cargo doc` with `-D warnings`,
`cargo test --no-default-features`, `cargo fmt --check`, and commit subjects checked against the
`<type>(<scope>): …` format the CI lint enforces.

The known-leaks test's row for this moved into `test_short_circuit_side_effect_shapes_already_match_jq_820`
per that test's own written instruction, with the `reduce` spelling added beside it.


newhoggy:
Addressed the `/code-review high` findings — both were correct, and both were **a real regression this branch introduced**, not a pre-existing one.

Converting the fork-level exits into `Demand::Stop` dropped the marking that tells a `?//` the stop is not an ordinary satisfied-consumer one. A `?//` *in INIT* therefore took it for one, tried its next alternative, and ran the fold again past an escape jq reaches once:

```console
$ echo '{"a":1}' | jq -c '[path(foreach (1) as $v ((1 as $x ?// $y | .a); ("h"|halt_error(3)); .))]'
h                    # this branch wrote `hh`
```

Both sinks made the identical conversion, so `reduce` had it too. Fixed with `mark_nonretryable_escape(&control)` before each stop — the same call `stop_with_escape` already makes for the fold's own inner escapes; this is just the one escape that leaves through INIT's sink instead.

**Why my sweep missed it, since that's the more useful part:** it generated `(1 as $x ?// $y | .a)` as an INIT shape, and `halt_error` nowhere. The bug needs both at once. The generator now pairs every INIT shape with `halt_error`/`error`/`stderr` fold bodies — **12155 programs (up from 5775), 0 regressions, 1036 fixed**, and all 10 residuals byte-identical to the merge-base.

Four rows pin it: both fold spellings with `?//`-in-INIT plus `halt_error`, and two controls showing a *retryable* escape in the same position still retries, so the marking is on the escape's kind rather than on every fork stop. I verified the rows fail with the marking removed.

Re-verified: 8546 tests, both clippy configs, rustdoc, no_std, fmt; patch coverage was 100% before this commit and the added lines are covered by the new rows.

Issue #2929, opened 2026-09-13
jq: del() through a slice piped into repeated .. deletes the wrong element


newhoggy:
**Severity: Low** (found by a differential sweep while fixing #2909; pre-existing and unrelated to it)

## Summary

`del()` over a path that pipes a slice into `..` twice deletes more, or different, elements
than jq. Present on `main` and on the #2909 merge-base alike, and reproduces with **no `?`
anywhere**, which is what rules out #2909's `?`-group scope as the cause.

## Repro

Captured live against `/usr/bin/jq` 1.7.1:

```console
$ echo '[1,2,3]' | jq -c 'del((.[0:1] | .. | ..)?)'
[2,3]
$ echo '[1,2,3]' | succinctly jq -c 'del((.[0:1] | .. | ..)?)'
[3]

$ echo '[1,2,3]' | jq -c 'del(.[0:1] | .. | ..)'     # no `?` at all
[2,3]
$ echo '[1,2,3]' | succinctly jq -c 'del(.[0:1] | .. | ..)'
[3]
```

More rows, same shape:

| input | jq 1.7.1 | succinctly |
|---|---|---|
| `[true,false]` | `[false]` | `[]` |
| `[[1,2],[3]]` | `[[3]]` | `[]` |
| `[{"a":1},5,{"a":3}]` | `[5,{"a":3}]` | `[{"a":3}]` |

A single `..` agrees; it takes two.

## Root cause

Not investigated. The shape is a slice component followed by two recursive descents, which
produces overlapping and nested resolved paths — so the likely area is how the deletion trie
interns paths that are prefixes of one another after a slice step, rather than the descent
itself.

## Why this is separate from #2909

#2909 is about a `?` over a group being distributed onto the group's components instead of
scoping the whole group. This reproduces without a `?`, and diverges identically on the
merge-base — the #2909 sweep (2,400 combinations) found these 8 rows and no others, and they
were unchanged by that fix.

## Suggested fix direction

Start from `del(.[0:1] | ..)` (agrees) versus `del(.[0:1] | .. | ..)` (diverges) and diff the
resolved path lists the two produce, before looking at the trie. Capture the oracle for the
one-`..`, two-`..` and three-`..` forms over array, object and scalar-element inputs first —
the table will say whether this is about path *count*, path *order*, or prefix interning.

Refs #2909.

newhoggy:
## Triage plan (2026-09-17)

[...]

PR #2930, opened 2026-09-13
fix(jq): give a `?` over a fan-out group jq's single abort scope

### LINKED ISSUE #2909: jq: a `?`-wrapped generator piped from `..` in path mode resumes after the error, so its side effects over-fire
**Severity: Low** (side-effect count; output matches)

## Summary

In path mode only, and only when the `?`-wrapped generator is reached *through* `..`, an error
inside the generator does not backtrack out of the enclosing `.[]` — evaluation resumes with
the next element, so a side effect ahead of the error fires once per element instead of once.

## Repro

Confirmed live against jq 1.7.1 and `succinctly` at `a953112f2`; pre-existing (the same
divergence reproduces on `1b12aa528`, before #2694's streaming work, so it is not caused by it):

```console
$ echo '[true,false]' | jq -c '[path(.. | (.[]|stderr|.+0)?)]'
true[]

$ echo '[true,false]' | succinctly jq -c '[path(.. | (.[]|stderr|.+0)?)]'
truefalse[]
```

`.+0` on a boolean raises; jq's `try` aborts the whole body, so `.[]` is never resumed and
`stderr` writes once. Here `.[]` continues to the second element and writes twice.

Three neighbouring spellings all agree with jq, which is what localises it to the
`..`-piped path-mode route rather than to `?` or to `.[]`:

```console
$ echo '[true,false]' | jq -c '[(.[]|stderr|.+0)?]'            # true[]   — both agree
$ echo '[true,false]' | jq -c '[.. | (.[]|stderr|.+0)?]'       # true[]   — both agree (value mode)
$ echo '[true,false]' | jq -c '[path((.[]|stderr|.+0)?)]'      # true[]   — both agree
```

So the difference appears only when the `?` sits downstream of `..` *and* inside `path()`.
Likely the node `..` delivers is not the document's own tracked value, routing the `?` through
a different arm than the standalone `path((...)?)` spelling takes
(`resolve_optional_sink` vs. whatever the piped stage reaches).

## Fix direction

Find which arm the `.. |` stage routes the `Optional` through and make its error handling stop
the enclosing generator, as `resolve_optional_sink` already does for the standalone spelling.
Needs a live-oracle differential sweep against jq 1.7.1 per ADR-0018 before landing; the
neighbouring rows above are the regression guards.

newhoggy:
## Summary

`(A | B)?` is `try (A | B)` — **one** abort scope over the whole group: a
failure anywhere inside prunes the group's branch and stops the group's own
generators. Two sites rewrote it into `A? | B?` instead —
`push_path_components` (#1311) for the resolver, `splice_optional_group`
(#1294) for the write walkers — turning one abort scope into N independent
per-step ones.

Both rewrites exist to stop the group's `?` leaking onto what *follows* the
group, and both are correct for the pure-navigation groups they were written
for: a single-valued chain has one branch, so pruning at step *k* and aborting
the group are indistinguishable. Put a **generator** inside the group and they
part company — it resumes past an error a later component raised.

### The issue understated this

#2909 filed it as a side-effect count with output matching. The output
diverges, and silently. All captured live from `/usr/bin/jq` 1.7.1:

```console
$ echo '{"a":{"b":1},"c":[1,2]}' | jq -c 'del((.[] | .[0])?)'
{"a":{"b":1},"c":[1,2]}          # was: {"a":{"b":1},"c":[2]}   <- deletes an element jq keeps

$ echo '[{"a":1},5,{"a":3}]' | jq -c '((.[] | .a)?) |= 99'
[{"a":99},5,{"a":3}]             # was: [{"a":99},5,{"a":99}]   <- writes a slot jq never reaches

$ echo '{"a":{"b":1},"c":[1,2]}' | jq -c '[path(.. | ((.[], .a) | .b)?)]'
[["a","b"]]                      # was: [["a","b"],["a","b"],["a","a","b"]]
```

Two further corrections to the issue's own diagnosis: **`..` is not required**
(`[path(. | (.[]|stderr|.+0)?)]` diverges identically), and it is **not**
`resolve_optional_sink` vs another `?`-handling arm — in the diverging shapes
the `?` never reaches a `?`-handling arm at all, because it is dissolved before
resolution.

One row of the issue's own list does **not** diverge here: `((.[] | .a)?) = 99`
already matched, so it is in the must-not-change set rather than the fixed set.

## How

`optional_group_is_scope_safe` — one predicate, both rewrites gated on it:
distribute only when the group has fewer than two components, or every one of
them is single-valued (`Identity`/`Field`/`Index`/`Slice`, through
`Paren`/`Optional`). Conservative on purpose: a group like `(.[] | .a?)` that
happens not to diverge today is also called unsafe, which only changes *which*
correct arm handles it.

An unsafe group then has to **reach** that arm, so `needs_path_prepass` and
`needs_fanout_pass` both answer `true` for it. Both are needed, for different
reasons: the former is what routes `del`/`=`/`|=`/`path` through the resolver
at all; the latter peels `Optional` before testing, so
`Optional(Pipe([Iterate, Field]))` would otherwise be mistaken for a
single-valued static tail. With both set, the group reaches
`resolve_optional_sink` — the arm that was already right, and the reason the
issue's three "neighbouring spellings agree" rows agree.

`wrap_optional_branch`'s `depth() != 1` arm was annotated "unreached today" and
coverage-tolerated, on reasoning that held only while a group was always
distributed before reaching a resolver. **It is live now** — verified by
instrumenting it — and needed no new logic: the old comment had already stated
jq's answer for exactly this shape.

## Verification beyond the suite

- **2,400-combination differential sweep** against jq 1.7.1 (30 group bodies,
  14 of them randomised from a component alphabet rather than hand-picked, × 10
  outer spellings × 8 inputs): **8 divergences, 0 regressions**. All 8 are one
  pre-existing shape — `del((.[0:1] | .. | ..)?)` — which diverges identically
  on the merge-base *and* with no `?` anywhere, so it is not this class. Filed
  as #2929.
- **756-combination leak scan** for an un-distributed group reaching a native
  walker (the failure mode #1311 was filed for): **none**. That is the audit of
  `push_path_components`' 11 callers the plan asked for, done empirically.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — 8554 passed, 0 failed
      (including #1294's and #1311's own tests, which are the gate's lower edge)
- [x] `cargo fmt --check`
- [x] The two sweeps above
- [x] 3 new CLI regression tests: the fixed rows, the neighbours that must not
      move, and the scope-safe class that must keep being distributed

Fixes #2909


newhoggy:
Both findings addressed; the crash one was serious and the diagnosis exactly right.

**The crash was self-inflicted duplication.** I wrote the scope-safety rule twice — the expression form skipped `Identity` and recursed into `Pipe`/`Paren`, the slice form counted `Identity` and treated a `Paren(Pipe(..))` element as multi-valued. That let a group be *safe* to the routing gate (kept on the native walker) and *unsafe* to `splice_optional_group`, which then returned the very list the caller had just decomposed. This repo's own "duplicated predicates diverge silently" lesson (#106), in a commit whose message claimed "one predicate". There is one walk and one entry point now.

**The better half of the fix was dropping the splice gate entirely.** Gating that site is what makes it a fixed point, and it turns out to be unnecessary — `needs_path_prepass` already routes a fan-out group through `resolve_optional_sink` before any walker sees it. I re-ran the sweep with that site ungated, widened to **3,840 combinations** with your crash shapes and `.`/paren-heavy randomised bodies added: **0 output differences, 0 crashes**. Six of those shapes are now CLI tests (they fail with exit 134 rather than subtly if either half regresses).

Your point about the 756-combination sweep is the one I want to keep: it varied neither a bare `.` nor a parenthesised sub-pipe inside the group, which is precisely why it could not see this. The alphabet a fuzz sweep draws from is part of the claim it supports.

The three documentation findings are fixed too — including `is_atomic_path_component`'s comment, which stated the false invariant as the safety argument for the whole change, and `push_path_components`' doc comment, which I had accidentally absorbed into the new predicate's rustdoc (the same defect `b6abf5d76` fixed for `resolve_leaf`).

PR #2931, opened 2026-09-13
fix(jq): give reduce a demand-forwarding fork core and dispatch arms

### LINKED ISSUE #2899: jq: reduce's own INIT position still materializes a ?// bind, hiding a consumer's stop (#2668 residual)

newhoggy:
Fixes #2899

## The gap

`reduce` had no demand-forwarding dispatch arm in either evaluator, so every evaluation —
including one wrapped in `first`/`limit`/`isempty` — reached the collecting `eval_single` path
and the wrapper drained an already-finished result. A consumer's stop had nowhere live to land:

```console
$ jq -nc '[first(reduce (1) as $v ((1 as $x ?// $y | 1); . + 1))]'
[2,2]                # succinctly answered [2]
```

#2899's own diagnosis of this was right and is worth keeping: reshaping `reduce`'s core alone
could not have fixed it, because nothing would ever have called the reshaped core through a live
sink. Both halves were needed.

## The change

`reduce_forks` replaces `eval_reduce_with_values` and is deliberately `foreach_forks`' structural
copy — same `drive_init`/`drive_source` closures, same out-of-band `ended`/`terminal` slots, same
`stop_with_downstream` classification, same terminal adjudication. The two constructs differ in
exactly one place, which is the one place the function differs: `foreach` emits per UPDATE step,
`reduce` emits once per INIT fork. Replacing rather than sitting beside the old function is the
point — #2899 warned a second entry point would drift the way `foreach`'s two did before #2180 WP3
unified them.

`each_reduce` / `each_reduce_generic` are the dispatch arms, gated on `!streams_unbounded` exactly
as `foreach`'s are, so an unbounded source or INIT still reaches the bounded fallback.

## A second divergence the reshape fixed, which #2899 did not report

The collecting version pulled SOURCE **once** and reused its values across every INIT fork. jq
re-drives it per fork — which #534's "each INIT output forks the whole reduce into a fully
independent execution over the source" already stated; only the *values* were being shared.

```console
$ jq -nc '[reduce ("s"|stderr) as $x ((0,1); .)]'
ss                   # succinctly wrote `s`
```

That made `docs/compliance/jq/limitations.md`'s #2163 INIT-fork entry stale — it described
`reduce` as computing SOURCE once and reusing it, in contrast to `foreach`. Both folds now
re-drive per fork against the real ambient document, so only the synthetic-`null` half of #2163
remains open, and for both folds rather than one.

## Verification

Attributed differential sweep — 4860 generated programs (180 sampled INIT × SOURCE × UPDATE
combinations × 9 wrappers × 3 inputs), each run against jq 1.7.1, this branch, and a build of the
merge-base:

```
4860 programs | REGRESSIONS: 0 | fixed: 778 | still divergent: 46
```

Of the 46: 20 are byte-identical to the merge-base. The other 26 changed but still differ from
jq, and **24 of those are #2163** — jq errors with `Cannot iterate over null` where both our
builds succeed, because jq feeds SOURCE a synthetic `null` on every fork after the first. That is
not `?//`-specific and predates this PR: plain `[reduce .[] as $x ((1,2); .)]` on `[1,2]` already
showed it. This change makes more rows *reach* that fork, so more rows display the pre-existing
divergence. The remaining 2 have identical stdout to the merge-base with stderr ordering moving
*toward* jq's.

Local gate: 8552 tests pass, both CI clippy configs clean, `cargo doc` with `-D warnings`,
`cargo test --no-default-features`, `cargo fmt --check`, and commit subjects checked against the
format the CI lint enforces. Patch coverage 99.32% — the one uncovered line is `reduce_forks`'
`unreachable!` terminal arm, carrying a marker; `foreach_forks`' identical arm is 0-hit too and
goes unflagged only because it predates this diff.

#2903's residual pin for this fired exactly as designed and its row moved into the matching table.

## Stale references cleaned up

Removing `eval_reduce_with_values` left five behind: two parity-test doc comments (one describing
`reduce` as still passing a `source` closure), a CLI-test comment, and two limitations passages.


newhoggy:
Addressed the `/code-review high` findings. All three were correct.

**The third one is the important one, and it means my "REGRESSIONS: 0" claim was too strong.** `reduce`'s old value-caching *accidentally* matched jq wherever the only visible difference was between "re-drive against the real document" and "re-drive against jq's synthetic `null`" — because caching re-drove against nothing at all. Driving per fork is exactly what the `?//` fix requires, and on those shapes it exposes #2163:

```console
$ echo '{"a":1}' | jq -c 'reduce ((.[]|stderr)?) as $x ((0,1); .)'
1            # succinctly writes 11; it wrote 1 before this PR
```

jq's second fork iterates its synthetic `null` and the `?` swallows that error before `stderr` runs. Stdout and exit codes are unaffected in every such shape, and it is the divergence `limitations.md` already records — but reached by more spellings than before, which my sweep's shapes didn't generate. That is now stated explicitly in the #2163 entry and pinned as a test row carrying *our* answer, so a future #2163 fix trips the test rather than passing quietly.

**Both doc comments fixed.** Anchoring the new functions at `fn each_foreach`/`fn each_foreach_generic` reassigned their whole doc comments — foreach's per-step/EXTRACT semantics, the stop-as-break table, the oracle rows — to `each_reduce`/`each_reduce_generic`, leaving both `foreach` arms undocumented and breaking the "see that function's doc comment" cross-reference. Moved back. My own doc's "the same unbounded-stream fallback" was wrong too and is corrected: both drives are lazy unconditionally, and the gate is the dispatch site's `!streams_unbounded` guard.

Re-verified: 8552 tests, both clippy configs, rustdoc, no_std, fmt, subject-format check.

Issue #2964, opened 2026-09-14
jq: break $x has no compile-time label-scope check, unlike real jq's compiler


newhoggy:
**Severity: Low**

## Summary

Found while implementing #2840's fix (`label`'s own error re-raise made shadowable by a
`def error:` in scope). succinctly has no compile-time check for `break $x`'s label scope —
it resolves purely by name at runtime — where real jq's compiler rejects `break $x` at compile
time unless it is lexically inside a `label $x | ...` body.

**Edit:** corrected the repro below — the original example
(`def error: "S"; def f: break $x; label $x | f`, `def error` declared *before* `def f`) is
fully explained by the *pre-existing* #2687 shadowing mechanism alone (`error` is in scope at
`f`'s own lexical position too, so `break $x` inside `f` is shadowed directly there, before it
ever becomes a `Control::Break`) — it reproduces identically on `main` before #2840 and isn't
specific to this issue's own missing check. The corrected repro below (`def error` declared
*after* `def f`) isolates the actual missing-scope-check gap this issue is about.

## Repro

```console
$ jq -nc 'def f: break $x; label $x | f'
jq: error: $*label-x is not defined at <top-level>, line 1:
def f: break $x; label $x | f
jq: 1 compile error
$ succinctly jq -nc 'def f: break $x; label $x | f'
(exit 0, no output)
```

The `def f: break $x;` is declared *before* `label $x | ...` even starts, so `$x` is not
lexically in scope when `f`'s body is compiled — jq's own label-scope check catches this
regardless of whether `f` is ever called. succinctly accepts it at parse/compile time and
only resolves `break $x` by label name at runtime, silently doing nothing when the label
that would have caught it is never actually on the call stack.

This also interacts with #2840's newly-added shadowable re-raise: since succinctly has no
compile-time rejection, the break above *is* constructible and *does* reach a label's
`Control::Break` handling at runtime as a genuine, self-matching break (`$x` matches the
enclosing label) — which `eval_label`'s own primitive check consumes silently on its own.
But #2840's fix makes the label's body a `Try` whenever a `def error:` is in scope *at the
label's own position*, and `Expr::Try`'s catch clause catches *any* `Control::Break`
regardless of which label it targets (#562) — so when `error` is declared *after* `f` (not
in scope at `f`'s own lexical position, so #2687 doesn't shadow the break there), the
self-matching break reaches the label as a genuine escape and is diverted into the shadowing
`error()` call instead of being silently consumed:

```console
[...]

Issue #562, opened 2026-08-02
jq: `try`/`catch` does not catch `break`, but jq 1.7.1's `catch` does


newhoggy:
**Severity: Low** — reproduced against the built CLI (debug), on branch `issue-400-jq-queryresult-models-error-break-per-stream`.

Found while verifying #400/#494's fix against the pinned jq oracle.

```
$ echo null | jq -n -c 'label $out | try break $out catch "c"'
"c"

$ echo null | succinctly jq -n -c 'label $out | try break $out catch "c"'
# (nothing — exit 0)
```

`eval_try` (`src/jq/eval.rs`) only matches `QueryResult::Error` to invoke the catch handler; a `Break` (or, after #400/#494, a `Partial` ending in `Control::Break`) falls through to the `other => other` arm and propagates untouched to the enclosing `label`. Real jq's `catch` construct evidently catches a `break` the same way it catches a raised error, running the handler — `label $out | try break $out catch "c"` yields `"c"`, not empty output.

Pinned as a known failure in the jq golden corpus: `tests/data/jq-golden/cases/try_break_not_caught/`.

**Distinct from #400/#494**: those are about a stream losing the outputs it already produced before an error/break; this is about `try`/`catch` not catching `break` at all, regardless of any prefix.

PR #2965, opened 2026-09-14
fix(jq): label's error re-raise is also gen_call(error) in real jq

### LINKED ISSUE #2840: jq: label's own error re-raise is also gen_call(error) in real jq, so a shadowing def error: also intercepts it — not just break

newhoggy:
## Summary

- `label \$x | BODY` intercepts every escape leaving `BODY` that isn't its own `{"__jq":N}`
  break sentinel and re-raises it through a synthetic call to `error/0`
  (`gen_call("error", gen_noop())` in `parser.y`), resolved by ordinary lexical scope at the
  label's own position — so a `def error:` in scope there shadows the re-raise, including a
  non-matching `break` of an *outer* label, not just `error(...)`. succinctly didn't
  reproduce this at all.
- Adds the fix at `resolve.rs`'s `check`/`build_call_graph` time: when
  `in_scope(scope, "error", 0)` holds at a `label`, its body is rewritten to
  `try BODY catch <call to error/0>` — the same runtime semantics every evaluator arm already
  gives `try`/`catch`, so **no evaluator arm changes** (the issue's own premise budgeted for
  rewriting six evaluator arms; investigation found the existing `try`/`catch` machinery
  already does everything needed). `build_call_graph` also needs its own copy of this gate,
  since it runs on the pristine pre-`check` tree, mirroring #2687's own `break`-desugaring
  call-graph edge so an otherwise-unreferenced `def error:` reachable only through a label's
  re-raise is still compiled/reported.
- 18/18 oracle rows verified live against `/usr/bin/jq` 1.7.1 before writing any code (value
  position, path position, assignment LHS, `first`/`limit` demand, generators, nested labels,
  multi-output/raising def, `halt_error`, `as`-binding — see the linked issue's triage plan).

## Docs

- `docs/compliance/jq/limitations.md`: closes the "second, related gap ... not covered by
  #2687" block this issue's own filing left there, and documents that the existing
  break-payload (`null` vs `{"__jq":N}`) divergence extends identically to a break
  intercepted through this new re-raise path — same #562 mechanism, not new.
- **New gap found while verifying this fix**, documented and filed separately as #2964: a
  shadowing `def error:` can catch a `break \$x` that real jq's compiler would reject
  outright at compile time (`def f: break \$x;` declared outside the label's lexical scope
  entirely) — succinctly has no compile-time label-scope check at all, a pre-existing gap
  this fix makes newly *observable* through the re-raise route rather than causes.

Fixes #2840

## Test plan

- [x] New tests: `test_label_error_reraise_is_a_shadowable_error_call_2840`,
      `test_label_reraise_call_site_makes_def_error_reachable_2840`
      (`tests/jq_cli_tests.rs`), `test_yq_label_error_reraise_is_a_shadowable_error_call_2840`
      (`tests/yq_cli_tests.rs`)
- [x] `cargo fmt --check`, both required `cargo clippy` feature combinations, 
- [x] `omni-dev coverage diff`: 100% patch coverage (21/21 new lines)
- [x] Live-verified against `/usr/bin/jq` 1.7.1; existing break/label/#2687 test suites
      (79 tests) pass unmodified


newhoggy:
Addressed the code-review findings in 0a03ea3ae, and investigated the two other real
findings:

**Doc precision fix (committed):** one reviewer traced my own #2964 repro
(`def error: "S"; def f: break \$x; label \$x | f`) and found it's fully explained by
*pre-existing* #2687 shadowing alone — `error` is in scope at `f`'s own lexical position
too, so the break inside `f` is shadowed there, before it ever becomes a `Control::Break`.
Verified live: it reproduces identically on `main` *before* #2840. The precise repro that
actually isolates #2840's own contribution declares the shadowing `def error:` *after*
`def f`, so the break stays genuine at `f`'s own position and only #2840's `Try`-wrap
(which catches *any* `Control::Break` regardless of target, #562) diverts it — confirmed via
a pre-#2840 build (empty output) vs. this branch (`"S"`). Both directions trace to the same
pre-existing root (#2964 — no compile-time label-scope check); real jq rejects the whole
program either way, so there's no reference answer to diverge *from* here, only which wrong
answer succinctly gives for this one adversarial, jq-can't-even-compile-it shape changed.
Corrected both `docs/compliance/jq/limitations.md` and #2964's own issue body to the precise
repro.

**Lost fast-path classification (filed as #2967, not fixed here):** `#2840`'s `Try`-wrap
makes `path_context_absent_keeps_position` classify a shadowed label's body as "position
moved" even when the un-rewritten body would have classified as `true`, since that
classifier has no `Expr::FuncCall` arm. Verified functionally correct (falls back to the
slower general path, not wrong output) — purely a lost optimization, only when `def error:`
happens to be lexically in scope at a `label`. No magnitude measured; filed for someone to
benchmark before attempting a fix.

**Unconditional `in_scope` scan per label (not filed):** the new `Expr::Label` arms run
`in_scope`/`reach_in_scope` for every label in every compiled program, unlike #2687's own
zero-cost parser-level `shadowable_defs` gate. Multiple reviewers independently assessed
this as a real but negligible compile-time-only cost (O(scope-depth) per label, labels are
rare) — not filed separately; noting here for the record.

Full test suite, both clippy feature combinations, fmt, and doc all green after the fix.
100% patch coverage confirmed via `omni-dev coverage diff` before this round of fixes (only
doc changes since).

newhoggy:
CI caught a real gap in my local verification: `Test (ARM64)`/`Test (macOS ARM64)` (which build the lib target outside `cfg(test)`, i.e. `no_std`) failed with `E0433: cannot find type Box in this scope` — `resolve.rs` had no `alloc::boxed::Box` import, and my new `Box::new(...)` calls in the `Expr::Label` rewrite relied on it being in the prelude, which only holds under `std`. Fixed in 654f6909b (now rebased onto latest main, force-pushed). Reproduced locally with `cargo build --lib --no-default-features --features ""` before pushing again — clean.

PR #2970, opened 2026-09-14
test(jq): pin an unaudited to_owned_lossy call site as decode-safe (#1908)

### LINKED ISSUE #1908: jq: audit the ~47 remaining boilerplate to_owned(&value) sites in eval.rs (#1820 category 2)
#1820's audit split into two categories. Category 1 (16 named sites sharing #1755's exact "silent \`\"\"\` substitution" shape) is done: 3 were already fixed by #1829, the remaining 13 were fixed by #1905 (which also caught and fixed a real regression from an initial naive fix — see that PR's commit history and the follow-ups filed as #1907).

Category 2 was explicitly deferred by #1820 itself as unsized:

> These live in generic `QueryResult`-collector/fan-out plumbing (e.g. `stream_outputs`, `prepend`, and dozens of `QueryResult::One(v) => ... to_owned(&v)` sites scattered through control-flow combinators). `to_owned_checked`'s own doc comment brackets this whole class as "boilerplate" not needing the fix, but the five already-merged #1755 slices (plus category 1) repeatedly found that bracketing isn't reliably safe -- some of these plumbing sites genuinely do feed corrupted data into real results. `prepend` in particular backs `try`/`catch`'s output-prefix splicing and looks like a plausible next place to check.

Filing this as its own issue since #1820 is now closed and this scope would otherwise have no tracking home.

## Scope

1. Re-grep `to_owned(&e)` / `to_owned(&v)` / `to_owned(&elem)` / `to_owned(&item)` / `to_owned(&r)` / `to_owned(&c.value())` / `to_owned(&field.value())` (and similar variable-name variants) against current `main` -- line numbers from #1820's original audit are stale.
2. For each hit, check what it's actually used for downstream, same as #1905 had to (see [[#1907]] and #1905's commit `79ca66585`/`06a6d67bc`): a genuine #1755-shaped bug only exists where the corrupted content could plausibly reach real output or change a decision. Some plumbing sites are content-independent (see the "uniform swap can regress content-independent ops" lesson from #1905's own regression) and should NOT get a blanket `to_owned_checked` swap without checking first.
3. `prepend` (`try`/`catch`'s output-prefix splicing) is the one site #1820 flagged by name as a plausible next check -- start there.
4. Consider whether a shared helper (see #1907's "Simplification" section -- a `scalar_fallback`-style function, or a `QueryResult`-flavored analog of `eval_generic.rs`'s `owned_or_err!`/`decode_failure_or`) is worth introducing given the site count, rather than continuing the one-off hand-rolled pattern.

## Root cause

Same as every #1755/#1806/#1829/#1820 slice: `to_owned`/`to_owned_at_depth` silently substitute `OwnedValue::String(String::new())` for a string whose bytes don't decode, instead of raising `EvalError::decode_failure`. `to_owned_checked`/`to_owned_checked_at_depth` are the fallible twins.

newhoggy:
## Summary

Re-derived #1908's site inventory against current `main` and closed the gap it left open.

**Site count re-derivation:** the issue's own comment counted ~61 non-definition grep hits
for `to_owned_lossy(`; a precise word-boundary search (excluding `into_owned_lossy(`/
`push_owned_values_lossy(`-style substring false-positives, and test-module usages) finds
**53** bare `to_owned_lossy(&x)` production call sites in `eval.rs`. These fall into the same
three buckets `to_owned_lossy`'s own doc comment already names:

1. **~36 error-message rendering** — the decoded value only ever feeds a `type_name()`-style
   preview inside an already-erroring message. No action.
2. **~7 reindex-bridge / structurally-owned-input plumbing** (`result_to_owned_full`,
   `eval_owned_expr_full`, `detach_from_temp_document`, `into_owned_lossy`'s own definition,
   and — newly reclassified by this PR — `eval_owned_multi_first`'s `collect_owned`, reached
   only via `update_path`'s `Expr::Identity` arm with `root: &mut OwnedValue` already fully
   materialized, so no borrowed cursor is ever reachable there) — structurally safe, no
   undecodable borrowed string can exist at these call sites regardless of document content.
3. **~10 individually-justified value-reading sites**, each already carrying its own safety
   doc comment from prior audits: `item_to_owned` (#2023), `finish_result` (#1907),
   `builtin_mktime`/`builtin_strftime` (`scalar_decode_failure`-guarded first),
   `collect_join_parts`/`builtin_truncate_stream` (likewise guarded), the `stream_expr`
   site inside `builtin_limit` (`#2188 audit`, provably content-independent).

**Newly reclassified this PR (was bucket 3, now bucket 2):** `eval_owned_multi_first`'s
`collect_owned` call, reached via `update_path`'s `Expr::Identity` arm (`|=`'s update-filter
evaluation). Traced the mechanism directly: `root` here is always a pre-materialized
`OwnedValue` (`eval_update`'s own top-level entry already went through a separately-audited
checked conversion, #1902/#1934) — `collect_owned`'s `to_owned_lossy` arms are exercised by
every `|=`, but structurally can never meet undecodable *bytes* at this call site. Pinned as
`test_update_path_identity_arm_root_is_always_owned_1908`, which also verifies the overall
observable behavior stays correct (`decode_failure` propagates from the eager document-level
bind, matching jq's "no partial writes on undecodable data" expectation).

**Also fixed a stale doc comment found during this investigation:** `push_owned_values_lossy`
still credited `eval_binary_fanout` (#768) as its caller; #1972 moved `eval_binary_fanout` onto
`binary_fanout_each`'s checked `Item::into_owned` instead, leaving `eval_owned_multi_keep_partial`
(`in(xs)`, `limit`, pattern alternatives, `if`/`select` conditions, `resolve_leaf`'s general
case) as the real caller today.

**Known residual, not chased further** (given this issue's own "Low severity, classification"
scope): `eval_owned_multi_keep_partial`'s other callers (`builtin_in`, `limit`, pattern
alternatives, `if`/`select`) were not individually re-probed with a live repro — constructing
one for `in(xs)` specifically proved awkward, since its key and candidate-generator expressions
share the same ambient `.`. `try_pattern_alternatives`'s `carried` relies on the original
triage comment's own finding (guarded by the bridge, unreachable in practice).

Fixes #1908

## Test plan

- [x] New test: `test_update_path_identity_arm_root_is_always_owned_1908` (`src/jq/eval.rs`)
      `cargo build --lib --no-default-features --features ""` (no_std),
- [x] No behavior change (test/doc-only PR); no `docs/compliance/jq/limitations.md` entry
      needed (not a jq-fidelity divergence)


newhoggy:
Addressed a significant finding from code review in e71e767b6:

**The original test's doc comment misattributed the safety it observed.** The reviewer
constructed an isolated probe (`. as \$doc | 1`, never referencing \`\$doc\` at all) against the
same corrupt document and found it raises the *identical* `decode_failure` — proving the
error comes from the eager `. as \$doc` bind materializing the whole document up front, not
from `eval_owned_multi_first`'s own `collect_owned` call as claimed.

I verified this myself and traced the actual mechanism: `update_path`'s `Expr::Identity` arm
calls `eval_owned_multi_first(filter_expr, root)` with `root: &mut OwnedValue` — by the time
`update_path` runs, the whole document is *already* a fully-materialized `OwnedValue`
(`eval_update`'s own top-level entry, via a separately-audited checked conversion, #1902/#1934).
So `collect_owned`'s `to_owned_lossy` arms are exercised by every `|=` here, but structurally
can never meet undecodable *bytes* at this call site — there's no borrowed cursor left by the
time this code runs. Reclassified this site from "individually-justified value-reading risk"
to "structurally safe by construction," the same reasoning `to_owned_lossy`'s own doc comment
already gives the reindex-bridge sites. Renamed and rewrote the test's doc comment
accordingly; kept the regression coverage since it still correctly pins the overall observable
`|=` behavior.

**Also fixed a stale doc comment found during the same investigation**: `push_owned_values_lossy`
still credited `eval_binary_fanout` (#768) as its caller, but #1972 moved `eval_binary_fanout`
onto `binary_fanout_each`'s checked `Item::into_owned` instead — its real caller today is
`eval_owned_multi_keep_partial` (`in(xs)`, `limit`, pattern alternatives, `if`/`select`,
`resolve_leaf`'s general case). None of *those* callers were individually re-probed with a
live repro in this PR (constructing one for `in(xs)` specifically proved awkward — its key and
candidate-generator expressions share the same ambient `.`) — noted as a known residual gap in
the closing summary rather than chased further, given this issue's own "Low severity,
classification" scope.

Full test suite, clippy, fmt, doc, and no_std build all green after the fix.

Issue #2981, opened 2026-09-15
yq: an unbound $variable silently produces no output in real yq; succinctly yq raises a runtime error


newhoggy:
**Severity: Low**

## Summary

Real yq treats a reference to an unbound `$variable` as a zero-output generator (like
jq's `empty`) rather than an error of any kind — confirmed live against yq v4.53.3, not
taken from docs. `succinctly yq` currently raises a runtime error instead (`Error:
undefined variable: $name`, exit 1).

## Repro

Confirmed live against yq v4.53.3:

```console
$ echo null | yq -o=json '$nope'
                                    # nothing at all -- exit 0, zero output
$ echo null | yq -o=json '$nope, 1'
1                                   # only the resolvable branch's output
$ echo null | yq -o=json '[$nope, 1, $nope]'
[
  1
]                                   # the array constructor collects only the real output
$ echo null | yq '.foo'             # contrast: an ordinary null *value* still prints
null
```

But it is not simply "treated as `null`" either — `+` (and only `+`, not tested
exhaustively beyond this) appears to special-case it as an additive identity rather than
propagating the empty-generator semantics real jq's arithmetic operators would apply:

```console
$ echo null | yq -o=json '1 + $nope'
1
$ echo null | yq -o=json '$nope + 1'
1
```

(Real jq has no equivalent to probe against, since jq rejects an unbound `$var` at
compile time before evaluation — see #2734, which fixes that side for `succinctly jq`
only.)

succinctly yq currently diverges the other direction — stricter than real yq, not more
permissive:

```console
$ echo null | succinctly yq '$nope'
Error: undefined variable: $nope
$ echo null | succinctly yq '$nope, 1'
Error: undefined variable: $nope        # the `1` never prints either
```

## Root cause

`src/jq/eval.rs`'s `Expr::Var` runtime arm returns a hard error for any name not found
in the evaluator's binding environment. Real yq's Go implementation apparently resolves
an unbound variable reference to some form of "no result" (context needed: is this a nil
`CandidateNode` filtered out during output collection, with `+`'s own arm separately
treating a nil operand as identity? not root-caused here, only the black-box behavior is
pinned).

## Why this is separate from #2734

#2734 adds compile-time rejection of an unbound `$var` to `succinctly jq` (jq itself
rejects at compile time, matching that is the fix). Applying the same compile-time
rejection to yq mode would make succinctly's *already-wrong-direction* divergence here
[...]

Issue #2971, opened 2026-09-14
jq: a def inside a builtin argument is never compile-checked (reachability keyed on an address check clones away)


newhoggy:
**Severity: Medium** — a compile error is downgraded to a runtime error (exit 3 → 5), and on some inputs replaced by an unrelated one.

## Summary

A `def` nested inside a **builtin argument** is never compile-checked, even when it is
called. jq rejects the program at compile time; succinctly compiles it and fails later at
runtime, with a different exit code and sometimes a completely different message.

#2740's reachability pass identifies a `def` by its body's **heap address**, recorded by
`build_call_graph` before `check` runs. But `check`'s own `Expr::Builtin` arm *clones* each
sub-expression before checking it, so every `def` inside a builtin argument is checked at a
**new address** that is not in the reachable set — and is therefore skipped as if it were
never called.

## Repro

Captured live against `/usr/bin/jq` 1.7.1:

```console
$ jq -nc 'def g: nosuchfn; g'                 # jq: error: nosuchfn/0 is not defined   (exit 3)
$ succinctly jq -nc 'def g: nosuchfn; g'      # same, exit 3                            <- correct

$ jq -nc '[1]|map(def g: nosuchfn; g)'        # jq: error: nosuchfn/0 is not defined   (exit 3)
$ succinctly jq -nc '[1]|map(def g: nosuchfn; g)'
                                              # jq: error (at <unknown>): undefined
                                              #     function: nosuchfn/0              (exit 5)
```

Exit code 3 (compile error) vs 5 (runtime error) is user-visible: scripts branch on it, and
jq guarantees a program that fails to compile produces *no output at all*.

On an input that makes the builtin fail first, the diagnostic is not merely differently
worded but unrelated:

```console
$ jq -nc 'map(def g: nosuchfn; g)'            # jq: error: nosuchfn/0 is not defined   (exit 3)
$ succinctly jq -nc 'map(def g: nosuchfn; g)' # Cannot iterate over null (null)        (exit 5)
```

Reproduces for every builtin whose children go through `map_builtin_subexprs` — `map`,
`select`, `with_entries`, `any(f)`, `all(f)` all confirmed. It does **not** reproduce for
`first(...)`, which parses to `Expr::FirstExpr` rather than `Expr::Builtin` and so is walked
in place.

## Root cause

`src/jq/resolve.rs`.

- `build_call_graph` records a def's identity as `body.as_ref() as *const Expr as usize`,
  read-only, before `check` runs.
[...]

Issue #2885, opened 2026-09-13
jq: resolve.rs's build_call_graph is a third hand-written full-tree Expr match


newhoggy:
### Background

#2740 (PR #2882) added a reachability pass (`build_call_graph`, `src/jq/resolve.rs`) that computes which `def` bodies are ever textually referenced, so `resolve_func_calls_all`'s existing `check` walk only checks bodies that are actually reachable. `build_call_graph` needs the same scope-tracking, `#2036` builtin-fallback-shadowing-aware structure `check` already has, so its own review found it mirrors `check`'s ~230-line exhaustive `Expr` match arm for arm — differing only in the `FuncDef`/`FuncCall` handling.

This makes **three** independent, hand-written exhaustive matches over every `Expr` variant in this crate: `check` (`resolve.rs`), `build_call_graph` (`resolve.rs`, new), and `walk::any_subexpr`/`map_subexprs` (`walk.rs`). `walk.rs`'s own module doc already names the exact failure mode this pattern risks: *"Before this module there were two hand-written answers to that question and they disagreed... Resist the temptation to shorten this with a catch-all: the cost of the long match is paid once... the cost of a wildcard is paid silently and repeatedly."*

The three matches are all exhaustive (no wildcard arm), so the compiler forces a total update whenever a new `Expr` variant is added — but nothing ties their *arm bodies* together beyond convention and comments, so a future edit to how one of the shared, non-special constructs (`Try`, `If`, `Reduce`, `Object`, ...) is walked can silently drift between the three.

### What's already confirmed

- `build_call_graph`'s own review (#2740) found and fixed one concrete instance of this drift risk before merge: its `FuncCall` arm's "callee doesn't resolve" branch didn't originally mirror `check`'s own `#2037` rule (real jq never compiles an unresolved call's arguments), and silently over-approximated reachability until caught by an oracle-verified test (`an_unresolvable_callees_arguments_do_not_mark_anything_reachable`).
[...]

PR #2982, opened 2026-09-15
fix(jq): reject an unbound $variable at compile time, not evaluation

### LINKED ISSUE #2734: jq: an unbound variable is rejected at evaluation (exit 5), not at compile time (exit 3) like jq

newhoggy:
## Summary

jq resolves a `$variable` reference at compile time, exactly like a function call — a program mentioning an unbound `$name` never runs at all (exit 3, zero output), unconditional and uncatchable by `try`/`?`. `succinctly jq` instead let `Expr::Var` survive to evaluation and errored there (exit 5), so any output produced before the bad reference had already been written.

- Extends `resolve.rs`'s existing `check()` walk (already used for #1473's function-call resolution) with a parallel variable-scope stack, rather than adding a second full `Expr` match — per #2885's concern about duplicating that exhaustive walk (`check` and `build_call_graph` stay the only two).
- Tracks every real binding site: `as`/`as`-pattern, `reduce`/`foreach` patterns (bound in `update`/`extract`, not `init`), `$`-style `def` params (bound in `body` only). Confirms `label $name` does *not* bind a `$name` variable (separate namespace, verified live).
- Closes a pre-existing gap as a side effect: a pattern's computed key (`{(EXPR): P}`) was never checked for undefined calls *or* variables before this — the same walk now visits it in the pre-binding scope.
- Diagnostics interleave call and variable errors in source order, matching jq's own reporting (`$bar, foo, $baz` → three errors, left to right, not grouped by kind).

**yq mode is untouched.** Real yq (v4.53.3) treats an unbound `$variable` as a zero-output generator rather than any kind of error — the opposite direction of divergence. Filed separately as #2981; `yq_runner.rs` keeps using the function-only `resolve_func_calls`/`resolve_func_calls_all` view, unaffected by this change.

Also unblocks #2885 → #2971's design-space note is unrelated to this fix; this issue was picked from `Ready` independently.

Fixes #2734

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 1792/1792 + 3813 lib tests, 0 failures)
- [x] Every new behavior diffed byte-for-byte against the pinned oracle (`/usr/bin/jq` 1.7.1) for exit code, stdout, and stderr — including the multi-error interleaving case and the module-include location-omission fallback
- [x] Two pre-existing tests (`test_bare_param_does_not_create_a_dollar_binding_2726`, renamed `test_bare_param_dollar_rejection_...2734`) updated: they explicitly pinned the pre-fix "catchable via try/?" divergence this PR closes
- [x] `omni-dev coverage diff`: 99.36% patch coverage (309/311); the 2 uncovered lines are a defensive `panic!()` in a test helper, unreachable in a passing suite by design
- [x] yq-mode regression test added confirming `succinctly yq`'s existing (separately tracked) behavior is unaffected


newhoggy:
Ran \`/code-review high\` (4-agent fan-out). Two real, confirmed findings — both fixed:

1. **Position misattribution** — the unbound-variable diagnostic's blind \`${name}\` text search could land on an unrelated occurrence of the same spelling: a string literal (\`"look at $nope", $nope\`), or an earlier, genuinely *bound* reference of the same name. Fixed by adding a parser-tracked \`var_sites\` table (\`jq::VarSite\`/\`collect_var_sites\`), mirroring #2085's \`call_sites\` mechanism for functions. Closes the string-literal case completely (verified byte-for-byte against the oracle); the lexical-shadowing case is now at parity with — not worse than — the pre-existing, documented #2635 limitation the call-site mechanism already has (confirmed live: \`(def f: 1; f) | f\` mis-locates identically today, unrelated to this PR).

2. **Unintended yq-mode reach** — closing the "pattern computed key never checked" gap accidentally let a *function-call* error inside a computed key (\`{(EXPR): P}\`) reach yq mode too, contradicting this PR's "yq mode untouched" claim. Verified this is actually the correct extension of yq mode's own **pre-existing** undefined-function-call policy (\`succinctly yq 'nosuchfn'\` already fails today, unrelated to this PR) reaching a position it previously missed — not a new policy. Documented explicitly and pinned with a regression test.

Also applied 3 smaller reuse/simplification fixes: folded 3 copies of the pattern-binding union loop into one \`bind_patterns\` helper (reusing \`eval.rs\`'s existing \`pattern_alternatives_var_names\`, closing the exact duplication class #2180 already fixed once), replaced a hand-rolled \`$\`-string builder with the codebase's \`format!\` idiom, and combined \`FuncDef\`'s two passes over \`params\` into one.

Full suite re-verified after each round (build, full test suite, both clippy feature combos, fmt) — all green. Patch coverage: 97.46% (345/354); the 9 remaining uncovered lines are two defensive branches (a test-helper's unreachable-by-design panic arm, and a symmetry-only fallback branch for variables that — unlike its call-site twin — has no known live repro, documented inline).

PR #2983, opened 2026-09-15
fix(jq): stop peek_str panicking on a multi-byte character

### LINKED ISSUE #2975: jq: the parser panics on a multi-byte character after a field, instead of reporting a compile error
**Severity: Medium** (a panic, not a wrong answer — exit 101 with a Rust backtrace where
jq reports a compile error; reachable from any untrusted filter string)

## Summary

A non-ASCII character immediately following a field name makes `src/jq/parser.rs` slice a
`&str` on a non-character boundary and panic. Pre-existing on `main`; not introduced by
any current branch.

## Repro

```console
$ echo '{}' | succinctly jq '.a（'
thread '<unnamed>' panicked at src/jq/parser.rs:589:20:
end byte index 4 is not a char boundary; it is inside '（' (bytes 2..5 of string)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
$ echo $?
101

$ echo '{}' | jq '.a（'
jq: error: syntax error, unexpected INVALID_CHARACTER, expecting end of file (Unix shell quoting issues?) at <top-level>, line 1:
.a（
jq: 1 compile error
$ echo $?
3
```

`（` is U+FF08 FULLWIDTH LEFT PARENTHESIS (3 bytes in UTF-8); any multi-byte character in
that position should do.

Note the character has to follow something — a **bare** `（` is handled correctly and
reports a clean compile error, which is why this is easy to miss:

```console
$ echo '{}' | succinctly jq '（'
jq: compile error: parse error at position 0: unexpected character '（', expected expression
```

So the failing path is specifically the one that computes an end offset by adding an
ASCII-assuming length to a position, rather than the general "unexpected character" path.

## Root cause

`src/jq/parser.rs:589` slices with a byte index derived as if the character at the
position were one byte wide. Confirmed pre-existing: reproduces identically on a binary
built from `2891884b0`.

## Why this is separate

Found incidentally while reviewing PR #2972 (#2267's fix), which does not touch the
parser, and reproduces on its merge-base. Wrong layer, wrong change.

## Suggested fix direction

Use a char-boundary-safe end (`char_indices`, `floor_char_boundary`, or advancing by
`ch.len_utf8()`) at that site, and audit the file for the same `pos + n` pattern. jq's own
answer is a compile error naming the invalid character, which the bare-`（` path already
produces — routing this case there would match both jq and the parser's own existing
behaviour.

Worth a fuzz pass over filter strings with multi-byte characters in every token position
once fixed, since the surviving sites are likely the same shape.

newhoggy:
## Summary

`Parser::peek_str` computed its end offset as `self.pos + n`, assuming every one of the next `n` bytes belongs to a single-byte character — true for every ASCII operator spelling it's ever compared against (`..`, `::`, `|=`, `?//`, ...), but not for arbitrary filter text. A multi-byte character (e.g. U+FF08 FULLWIDTH LEFT PARENTHESIS) immediately after a token — a field name, a number, a bracket, a quoted string — landed that byte arithmetic mid-character and panicked on the slice (exit 101, a Rust backtrace, reachable from any untrusted filter string), where real jq reports a clean compile error for the identical input.

```console
$ echo '{}' | succinctly jq '.a（'
thread '<unnamed>' panicked at src/jq/parser.rs:640:20:
end byte index 4 is not a char boundary; it is inside '（' (bytes 2..5 of string)
$ echo '{}' | jq '.a（'
jq: error: syntax error, unexpected INVALID_CHARACTER, ... at <top-level>, line 1:
jq: 1 compile error
```

Fixed by rounding the end offset down to the nearest char boundary instead of assuming one lands there — safe because no call site's comparison target contains a multi-byte character, so a shortened slice can only ever compare as not-equal, the same outcome a slice that already ran off the end of the input produces via the existing `.min(self.input.len())`.

Confirmed the panic is shared by both `jq` and `yq` mode (same `Parser`/`peek_str`), and confirmed no other slicing site in the parser has the same byte-arithmetic pattern.

Fixes #2975

## Test plan

- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` (clean)
- [x] Fixed output diffed against the pinned oracle (`/usr/bin/jq` 1.7.1): matching exit code (3) on all rows; message wording routes into the parser's own existing "unexpected character" style (the same style the analogous bare-character case already uses), per the issue's own suggested fix direction
- [x] New unit tests exercise `peek_str` directly (boundary rounding) and `parse()` end-to-end across 11 token-position variants (field, number, bracket, string, operator contexts) — all previously panicked, now clean `Err`
- [x] New CLI tests for both `jq` and `yq` mode confirming exit code (not a panic) on the same repro
- [x] `omni-dev coverage diff`: 100% patch coverage (31/31)


newhoggy:
Ran \`/code-review\` (8-agent fan-out). No correctness bugs found — the fix's safety argument (every \`peek_str\` call site either equality-compares against a fixed ASCII literal, or does a single-character extraction, neither of which a shortened slice can spuriously satisfy) held up across independent line-by-line, cross-file, and removed-behavior audits.

One genuinely useful discovery: a **second, independent pre-existing panic site** at the unary-minus digit lookahead (\`.chars().nth(1)\`, not an equality check) — confirmed panicking on \`main\` too (\`-（\` → exit 101) and fixed by this same change. Added a test row for it (swapped out a redundant \`"5（"\` entry that exercised the same code path as an existing \`"1（"\` row).

Also fixed two doc-comment issues: added the "confirmed live against \`/usr/bin/jq\` 1.7.1" citation this repo's convention requires for a jq-behavior claim (re-verified all 6 CLI-test rows live — all exit 3, matching), and corrected \`peek_str\`'s doc comment, which claimed "every call site" compares against an ASCII literal when one call site actually does a \`.chars().nth(1)\` extraction instead (behaviorally fine, but the blanket claim was imprecise).

Three review angles flagged the same char-boundary round-down loop now existing 3x in \`src/jq/\` (this fix's new copy, plus two pre-existing ones in \`error.rs\`'s \`PreviewSink\`), and suggested routing \`peek_str\`'s callers through the already-boundary-safe \`matches_keyword\` idiom instead. Real observations, but out of proportion for a scoped crash fix — filing as a follow-up rather than expanding this PR's diff.

Full suite re-verified after the fix (build, full test suite, both clippy feature combos, fmt, cargo doc) — all green. Patch coverage: 100% (31/31).

newhoggy:
Filed #2984 for the char-boundary loop duplication finding from the review above.

Issue #2962, opened 2026-09-14
jq: a wrapped module dependency sits inside the including def's scope, so that def's own name or parameters can capture names free in it


newhoggy:
**Severity: Low**

## Summary

Found by code review on PR #2954 (#2865), and recorded there in
`docs/compliance/jq/limitations.md`. #2865 makes a module's own `include`s work
by wrapping each exported def's body in the dependencies it reaches. The wrap
nests **inside** that def's own scope, so a name left free in a dependency can
be captured by the def's own name or by one of its parameters.

Not a regression: every shape below needs a transitive `include`, which did not
work at all before #2865 (`main` answers `g/0 is not defined` / `k/0 is not
defined` for each). These are residual gaps in the new functionality.

## Repro

All rows confirmed live against jq 1.7.1 and PR #2954's head.

### 1. The self-(name, arity) exclusion strands a dependency another dependency calls

```console
$ cat inner.jq
def c: 7;
def g: c;
$ cat mid.jq
include "inner";
def c: if . == 0 then g else (. - 1 | c) end;

$ jq            -L . -nc 'include "mid"; 0 | c'   # 7
$ succinctly jq -L . -nc 'include "mid"; 0 | c'   # error: g/0 exceeded maximum recursion depth (exit 5)
```

`mid`'s body calls `c` directly, so the dependency `c` is excluded to leave the
self-recursive call binding to itself — but the kept dependency `g` still needs
it, and its `c` then resolves outward to `mid`'s own `c`.

### 2. A parameter satisfies a name a dependency left undefined, swallowing a compile error

```console
$ cat gb.jq
def g: b;
$ cat hb.jq
include "gb";
def h(b): g;
def q: h(99);

$ jq            -L . -nc 'include "hb"; q'   # error: b/0 is not defined at gb.jq (exit 3)
$ succinctly jq -L . -nc 'include "hb"; q'   # 99 (exit 0)
```

The worse half of the family: a program jq rejects produces an answer.

### 3. A parameter captures a dependency reached only indirectly

```console
$ cat inner3.jq
def g: 42;
def k: [g];
$ cat h3.jq
include "inner3";
def h($g): [g, k];
def q: h(7);

$ jq            -L . -nc 'include "h3"; q'   # [7,[42]]
$ succinctly jq -L . -nc 'include "h3"; q'   # [7,[7]]
```

`h`'s own `g` is correctly the parameter; `k`'s `g` should still be the
dependency, and is not.

## Root cause

[...]

Issue #2991, opened 2026-09-16
jq: a module-body compile error names its file but not jq's line number or source echo (#2951 residual)


newhoggy:
**Severity: Low** (diagnostic detail; the name, arity, count, order, exit code and file all match jq already)

## Summary

A compile error raised inside a module body now names the module's canonical
file, byte-identical to jq's — but not jq's trailing `, line N:` or its echo
of the offending source line.

## Repro

```console
$ cat lib/mymod.jq
def helper: nosuchfn_in_module;

$ jq -n -L lib 'include "mymod"; helper'
jq: error: nosuchfn_in_module/0 is not defined at /abs/lib/mymod.jq, line 1:
def helper: nosuchfn_in_module;
jq: 1 compile error

$ succinctly jq -n -L lib 'include "mymod"; helper'
jq: error: nosuchfn_in_module/0 is not defined at /abs/lib/mymod.jq
jq: 1 compile error
```

Exit code (3), the `name/arity is not defined` text, the file path, the error
count and the ordering across multiple errors all agree. Only the `, line N:`
suffix and the echoed line are missing.

## Root cause

`report_compile_errors` (`src/bin/succinctly/jq_runner.rs`) has exactly one
source text to work with: the main filter. It gets real positions from
`jq::collect_call_sites(filter, ...)` and falls back to a text search over
the same string.

#2951 gave `UnresolvedCall` an `origin` — the run id the failing call was
written in — and `ModuleLoader::run_origin` maps that to a canonical path,
which is what the message now prints. What is not available is the module's
*text*: the loader reads each module with `std::fs::read_to_string` inside
`load_and_bind_module` and drops the contents once parsed.

Before #2951 this was worse than merely missing: with nothing marking a call
as module-originated, the text-search fallback could match a coincidental
occurrence of the same name in the main filter and cite *that* line for the
module's error. That misfire is already fixed — the reporter now stops at the
file name rather than guessing.

## Suggested fix direction

1. Keep each module's source alongside its run id in `ModuleLoader`
   (`run_origins` already holds `id -> canonical path`; make it
   `id -> (path, source)`), or re-read the file on the error path only — this
   is a cold path and a second read costs nothing, which avoids holding every
   module's text for the lifetime of every run.
2. Generalise `report_compile_errors` from one source to one per origin:
   `jq::collect_call_sites(module_source, ParserMode::Jq, true)` with the same
   nth-occurrence rule already used for the main filter, keyed per origin so
[...]

PR #2992, opened 2026-09-16
fix(jq): seal module bodies from names they should not see

### LINKED ISSUE #2951: jq: a module body sees names it should not — ~/.jq's defs, and sibling included modules' defs in declaration order
**Severity: Low**

## Summary

Filed from the planning notes on #2865 (follow-up **F1** there), re-confirmed
live. It is the *opposite* direction from #2865: not a module failing to see
its own dependencies, but a module body seeing names from outside it that real
jq keeps out.

succinctly inlines every module's defs into one flat `Expr::FuncDef` chain
wrapped around the main filter, so each module body physically sits nested
inside every other unqualified source — `~/.jq` and every sibling `include`.
Ordinary lexical scoping then resolves outward into them. Real jq gives each
module its own sealed scope.

## Repro

Both confirmed live against jq 1.7.1 and `succinctly` at #2865's fix.

### `~/.jq` leaks into a module body

```console
$ cat ~/.jq
def hj: 1234;
$ cat usehj.jq
def uh: hj;

$ jq            -L . -nc 'include "usehj"; uh'
jq: error: hj/0 is not defined at /abs/path/usehj.jq, line 1:
def uh: hj;
jq: 1 compile error

$ succinctly jq -L . -nc 'include "usehj"; uh'   # 1234
```

### A sibling module's defs leak in, order-dependently

```console
$ cat sibA.jq
def sa: sb;
$ cat sibB.jq
def sb: 77;

$ jq            -L . -nc 'include "sibB"; include "sibA"; sa'   # error: sb/0 is not defined at .../sibA.jq
$ succinctly jq -L . -nc 'include "sibB"; include "sibA"; sa'   # 77      <- leak

$ jq            -L . -nc 'include "sibA"; include "sibB"; sa'   # error: sb/0 is not defined at .../sibA.jq
$ succinctly jq -L . -nc 'include "sibA"; include "sibB"; sa'   # error: sb/0 is not defined at <top-level>
```

Reversing the two `include`s makes succinctly agree with jq **by accident** —
`sibB` is no longer wrapped outside `sibA`'s body — which is the tell that
this is scope leakage through the chain, not a scoping rule.

## Root cause

`ModuleLoader::process_program` (`src/bin/succinctly/jq_runner.rs`) wraps each
unqualified source around the *whole* accumulated expression, so a module body
already placed in the chain ends up inside the scope of everything wrapped
around it afterwards. AST inlining has no way to express "this subtree may not
see its enclosing defs".

Note #2865 deliberately did **not** extend the leak: it wraps a module's own
dependencies into that module's def bodies (a sealed, per-body scope) and
leaves `~/.jq` out of that wrapping entirely. The leaks above are the residue
of the older outer-chain mechanism.

## Suggested fix direction

Needs a real module boundary, which AST inlining alone cannot express. Two
shapes worth costing:

- **Name mangling**: rewrite each module's own defs and their internal call
  sites to a per-module unique prefix before inlining, so an unprefixed
  sibling/`~/.jq` name simply does not match inside a module body. Cheap-ish,
  but every diagnostic and `$__loc__`-adjacent surface has to unmangle.
- **A resolver-side boundary**: give `src/jq/resolve.rs` a scope-barrier node
  so lookup from inside a module stops at its own frame plus its explicit
  dependencies. Cleaner semantically; touches the resolver's scope model.

Either is a design change rather than a localised fix — likely `Fable`-class,
and worth an ADR note on which boundary mechanism the crate adopts, since
#2865's per-body wrapping is a third (narrower) answer to the same question.

## Why this is separate from #2865

#2865 is "a module cannot see its own dependencies"; this is "a module can see
things that are not its dependencies". Opposite directions, different
mechanisms (#2865 is the loader, this is the top-level chain shape), and
#2865's fix is forward-compatible with either boundary above.

### LINKED ISSUE #2989: jq: a def in an imported module cannot call its own sibling by bare name
**Severity: Medium**

## Summary

A def inside an `import`ed module cannot call its own sibling by the bare
name it is written with. Every such module fails to compile the moment one of
its defs calls another, so `import` is only usable for modules whose defs are
all self-contained.

`include` of the same module works — the bug is specific to the namespaced
path.

## Repro

Confirmed live against `/usr/bin/jq` 1.7.1 and `succinctly` at `c8a4e22ba`.

```console
$ cat lib/x.jq
def f: 1; def g: f;
```

|---|---|---|
| `import "x" as a; a::g` | `1` | `f/0 is not defined at <top-level>`, exit 3 |
| `import "x" as a; import "x" as c; [a::g, c::g]` | `[1,1]` | the same error **twice**, `2 compile errors`, exit 3 |
| `import "x" as a; a::f` | `1` | `1` (agrees — `f` is not called from another def) |
| `include "x"; g` | `1` | `1` (agrees — bare names) |

## Root cause

`ModuleLoader::process_program` (`src/bin/succinctly/jq_runner.rs`) namespaces
an imported module's defs by rewriting only their *names*:

```rust
let defs = defs
    .into_iter()
    .map(|(name, params, body)| (format!("{namespace}::{name}"), params, body))
    .collect();
```

The bodies are left untouched. So the def chain ends up holding `a::f` and
`a::g`, while `g`'s body still says bare `f` — a name that now exists nowhere
in the chain. `resolve.rs`'s `check` reports it as an unresolved call, once per
import of the module (hence the doubled error on a twice-imported module).

`include` is unaffected because it leaves the names bare, so a sibling call
still matches.

Note the arity-0 case `a::f` agrees only because nothing calls it from another
def; it is not evidence that single-def modules are handled specially.

## Why this is separate from #2951

Found while verifying #2951's premise against the oracle — it is the same
"module bodies and the flat inlined chain disagree about what a name means"
family, and #2951's planned resolver-floor design fixes it as a by-product
(its import rule retries a bare miss as `alias::name` while the innermost open
run carries that alias). But it is a distinct, independently-reproducible bug
with its own severity: #2951 is about names leaking *in* that should not be
visible, this is about a module's own siblings not being visible *at all*.

It could equally be fixed on its own by rewriting call sites inside an
imported module's bodies at namespacing time — so it should not silently
depend on #2951 landing.

## Suggested fix direction

Either:

1. **As part of #2951's resolver floor** (planned): a bare `name/arity` that
   misses in the sealed window, while the innermost run carries alias `a`, is
   retried as `a::name` and the `FuncCall` renamed in place. Covers this
   automatically and keeps one mechanism.
2. **Standalone**: when namespacing an imported module's defs, also rewrite
   every call inside those bodies that resolves to a sibling of the same
   module. Needs care not to rewrite calls to builtins or to names the module
   legitimately gets from elsewhere — which is the resolver's job, and is why
   (1) is the cleaner home for it.

Whichever lands, pin: `import "x" as a; a::g` → `1`; the twice-imported
`[a::g, c::g]` → `[1,1]`; and `include "x"; g` → `1` unchanged.

Refs #2951, #2865.

newhoggy:
## Summary

Fixes #2951. Closes #2989.

succinctly inlines every module's defs into one flat `Expr::FuncDef` chain wrapped
around the main filter, so each module body physically sits nested inside every other
unqualified source. Ordinary lexical scoping then resolves outward into them, and
resolves names real jq keeps out: `~/.jq`'s defs, and sibling `include`d/`import`ed
modules' defs. Real jq gives each module its own sealed scope.

Every row below was captured live from `/usr/bin/jq` 1.7.1 against the same fixtures,
and every row that discriminates was also run against the pre-fix binary, so none of
the new tests is a vacuous pin.

| filter | jq 1.7.1 | before | after |
|---|---|---|---|
| `include "usehj"; uh` (`~/.jq` defines `hj`) | `hj/0 is not defined at …/usehj.jq` | `1234` | ✅ error, same file |
| `include "sibB"; include "sibA"; sa` | `sb/0 is not defined at …/sibA.jq` | `77` | ✅ error, same file |
| `include "sibA"; include "sibB"; sa` | same error | error **by accident** | ✅ error, same reason |
| `import "sibB" as b; include "sibA3"; sa3` | `b::sb/0 is not defined` | `77` | ✅ error |
| `include "sibB"; include "two"; include "useboth"; useboth` | 2 errors | `79` | ✅ 2 errors, same order |
| `import "x" as a; a::g` (#2989) | `1` | `f/0 is not defined` | ✅ `1` |

The third row is the tell that this was leakage rather than a scoping rule: reversing
the two `include`s made succinctly agree with jq *because* `sibB` was no longer wrapped
outside `sibA`'s body — the right answer for the wrong reason.

## Mechanism

The loader brackets every run of defs it wraps (each `include`, the `~/.jq` block, each
`import`) between two marker defs whose names begin with a **NUL byte**, which cannot be
lexed and so can never be called. `resolve.rs` then treats an unmatched begin marker as
a scope floor.

The pleasant part: because markers are ordinary `FuncDef`s, **both walks already push
them onto their scope stack and truncate them at the right moment** — no walk state is
threaded through either recursive walk. Only lookup changes, once, in `scan_scope`:
scanning innermost-first, an end marker opens a *closed* run (wrapped around this point,
so visible) and an unmatched begin marker is the floor. That is why this is ~200 lines
rather than the ~400 the issue's plan estimated.

`ModuleRun` owns the marker spelling, shared by the loader that writes them and the
resolver that reads them, so the two cannot drift.

## Why the alternatives don't work

Recorded in **ADR-0023**, because the third one is likely to be re-proposed:

- **A field on `Expr::FuncDef`** costs 8 bytes on *every* `Expr` (the discriminant stops
  riding a niche, measured for #2283) and eats the `MAX_EXPR_DEPTH` stack margin.
- **A side table keyed by node address** cannot survive `substitute_vars`, which rebuilds
  the tree between the loader and this pass.
- **Name mangling is unsound, not merely expensive.** It does not close the leak at all:
  an exported name must stay callable *bare* from the main filter, so its bare wrapper
  stays outside the sibling's body and `def sa: sb;` still finds it. The only way
  mangling closes the leak is by rewriting the *callers* inside the sibling — which
  needs to know `sb` is free there, i.e. the resolver's job. It is a worse spelling of
  the decision taken, not an alternative to it.

## Why the evaluator is untouched

The sealed window is a *suffix* of the very stack the evaluator's own innermost-first
lookups walk. The nearest match within a suffix is the nearest match overall whenever one
exists; when none exists the program is a compile error and evaluation never starts.

That argument needs one more fact, which I verified rather than assumed: a module's source
is parsed with **no shadow-candidate seeding**, so a builtin call inside a module body is
never an `Expr::FuncCall` and never consults the def chain. Confirmed live — with a sibling
defining `length`/`map`, both include orders answer `2`/`[2,3]` in both tools.

**Therefore a leaked name could only ever resolve where jq reports a compile error; it
could never out-shadow a name jq resolves.** This change can turn a wrong answer into an
error. It cannot alter a correct one. `test_module_scope_floor_does_not_hide_builtins_2951`
pins that property, because without it the whole approach would be unsound rather than
merely incomplete.

## Closes #2989 as a by-product

A def inside an `import`ed module could not call its own sibling: the loader namespaced
the defs (`a::f`, `a::g`) but not the call sites in their bodies, so `g`'s body said bare
`f` and nothing matched. The import alias now rides the begin marker, so a bare miss
inside that run — and only inside it — is retried as `alias::name` and renamed in place.
The retry lives in **both** walks, or reachability and checking would disagree about which
bodies get compiled at all.

## Diagnostics

`UnresolvedCall` gains `origin`, the run a failing call was written in, so a module-body
error names the module's own canonical file exactly as jq does instead of `<top-level>`.

That also fixes a real misfire: with nothing marking a call as module-originated, the
text-search fallback could match a coincidental occurrence of the same name in the *main
filter* and cite that line for the *module's* error. It now stops at the file name rather
than guessing. jq's trailing `, line N:` and source echo still need the module's text
threaded down — #2991.

## Deliberately out of scope

The same marker would also seal the per-body dependency runs `load_and_bind_module`
wraps, closing #2962's capture route. Per #2951's own plan that flips two of #2962's
three pinned rows from a wrong *answer* to a compile *error* — still not jq's answer.
Trading one divergence for another belongs in #2962's own PR, with its own decision about
whether an error beats a wrong value. Recorded in `limitations.md` and ADR-0023, and the
change there is now small.

## Two pre-existing tests re-pinned

Both still assert their original intent; only the literal text changed, and both comments
now record jq's own behaviour on the row.

- `..._omits_the_location_1473` documented "jq names the module file and line; succinctly
  says `<top-level>`" as a known limitation. The file half is now closed. The `, line `
  assertion stays.
- `..._keeps_repeat_arity_and_fallback_2085` pinned that a module error must not borrow
  the filter call's line marker — still true (`line 3:` appears once), but the module
  error now names its own file. Its fixture uses `nosuch` in *both* places, which is
  exactly the coincidence the old fallback could trip on.

Also made `resolve_all_strs` assert on each diagnostic's `Display` behind its kind rather
than `{:?}` of the whole struct: the Debug form names every field, so adding `origin` broke
a test that is about ordering and kind and nothing else.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — 63 legs, 0 failed
- [x] `cargo fmt --all`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` (caught a `pub`-item
      link to a private one)
- [x] `cargo test --no-default-features` (no_std)
- [x] Behaviour diffed directly against `/usr/bin/jq` 1.7.1 on every row above, not just
      against goldens
- [x] New coverage: 3 CLI tests (leaks, controls, #2989) + 6 unit tests on hand-built
      scope stacks
- [ ] `omni-dev coverage diff` — before merge


newhoggy:
## `/code-review` (max effort) — three findings, all fixed in 2d398597c

The review built both binaries and diffed live against jq 1.7.1 rather than reasoning from the diff. Two of the three were **my regressions**, and I reproduced both against the pre-PR binary before touching anything.

### 1. The alias retry escaped into dependency bodies — a real regression

```
A.jq = include "B"; def g: 1; def f: b1;
B.jq = def b1: g;
jq -L . 'import "A" as ns; ns::f'
```

| | result |
|---|---|
| jq 1.7.1 | `g/0 is not defined at …/B.jq` |
| pre-PR binary | `g/0 is not defined` |
| this branch (before the fix) | **`1`** |

A correct compile error became a wrong answer — exactly the direction the PR's own soundness argument says is impossible. I had left dependency runs unmarked on the grounds that sealing them is #2962's business, but the alias retry keys on the innermost *open* run, so an unmarked dependency body inherited the importing module's alias.

### 2. `origin` named the enclosing module, not the file that wrote the call

Same root cause. `outer.jq` includes `dep.jq`; the undefined name is written in `dep.jq`. jq says `dep.jq`; base said the honest `<top-level>`; this branch said `outer.jq` — a confident citation of a file that does not contain the call. Worse than what it replaced, and the new branch exists precisely to stop guessing.

### 3. `scan_scope` never short-circuited — a new quadratic term

Both callers read `run` only on a miss, so the scan can stop at its match. That is now the documented contract on `ScanResult::run`.

## The fix

The needle to thread: a dependency body must stop inheriting the alias, but a module's **own** bodies must keep inheriting it or #2989 breaks — and at load time a module does not know whether it will be included or imported, so the alias cannot just move onto the module. Hence a third marker kind, a **barrier** run around each dependency group, keyed by the module the group came from. A barrier claims authorship (innermost open run, so it owns both the alias question and error attribution) but does **not** floor visibility.

Not flooring is deliberate. Sealing dependency bodies is #2962's capture route, and closing it moves two of that issue's pinned rows from a wrong answer to a compile error — still not jq's answer. `test_import_alias_does_not_reach_a_dependency_body_2951` pins both halves in one test on purpose: the `import` form now errors like jq, the `include` form still answers `1` as it did before this branch, with a message telling whoever closes #2962 to update it.

## On finding 3, a note on measurement

The review's numbers were a real signal but my first check of the fix was not. A naive sequential run read N=3000 as 0.45s → 1.03s, which looked like the quadratic term surviving. Interleaved, 5 reps, median — per `docs/guides/benchmarking.md` — it is parity:

| N | base median | branch median |
|---|---|---|
| 3000 | 0.45s | 0.44s |
| 6000 | 1.77s | 1.75s |

The 1.03s was a cold-start artifact. Worth stating because it is the exact failure mode that guide exists to prevent.

## Re-verified after the fix

All nine oracle rows still match jq 1.7.1 on exit code and error name; #2989's rows unchanged; full suite 63 legs green; both clippy matrices, rustdoc `-D warnings`, `no_std`; patch coverage **100% (280/280)**.

PR #2993, opened 2026-09-16
perf(jq): thread Cow through the owned-identity pipe

### LINKED ISSUE #2829: perf(jq): // clones value/id unconditionally on every evaluation, a previously zero-clone hot path
**Severity: Medium**

## Summary

`eval_owned_identity_alternative` (`src/jq/eval_generic.rs`, added by #2782) unconditionally clones `value` and `id` before evaluating `//`'s left operand, so the original, untouched pair survives for a potential right-side splice if the left produces no truthy output:

```rust
let left_flow = eval_owned_identity_stages::<S, V>(
    owned_identity_body_stages(left),
    value.clone(),
    id.clone(),
    ...
```

`OwnedIdentity<V>::clone` is cheap (an `Rc` bump plus O(depth) small components), but `OwnedValue::clone` is a plain `#[derive(Clone)]` deep copy of whatever subtree `value` currently holds (`Array(Vec<Self>)`/`Object(IndexMap<String, Self>)`).

## Root cause

`.foo // default` is one of the most common jq/yq idioms. Before #2782, the operand-shaped `//` rule ran through `owned_identity_operand` on `&OwnedValue`/`&OwnedIdentity` (references, zero clones) and only touched the right side when actually reached. #2782's transparent-construct rewrite needs the same starting `value`/`id` potentially available to *both* branches (left and right), and since evaluation consumes them by value, it clones upfront to keep a spare copy for the right side -- paid even in the overwhelmingly common case where the left operand is present and truthy and the right side never runs.

For `.big_array // []`, this is now an O(document size) copy on every evaluation of an operator that used to cost nothing extra.

## Why not fixed in #2782

A real fix needs either: (a) restructuring so the clone is deferred until the left operand is actually known to be exhausted-without-truthy (non-trivial: by the time that's known, the original has already been consumed by the left branch's evaluation, so recovering it requires either cloning earlier anyway or a different ownership shape for this route), or (b) making `OwnedValue` cheaply shareable (e.g. `Rc`-wrapped) for this purpose, a larger representational change. Either direction needs a real A/B benchmark before landing (per this repo's benchmarking discipline), which is out of scope for #2782's correctness fix.

## Suggested fix direction

Investigate whether the left operand can run against a borrowed `&OwnedValue`/`&OwnedIdentity` for navigation-only reads (mirroring the pre-#2782 `owned_identity_operand` shape) with an owned clone taken only at the point a truthy output is actually placed, or measure whether wrapping `OwnedValue::Array`/`Object` payload with `Rc` for cheap structural sharing is worth the wider blast radius. Needs a real A/B benchmark (interleaved, ≥1MB inputs, both architectures) comparing `.foo // default` throughput before/after any fix, per `docs/guides/benchmarking.md`.

## Why this is separate from #2782

#2782 is a correctness fix (`//` disagreeing with its own bare spelling); this is a performance question about the mechanism #2782 introduced to fix it, requiring hardware A/B benchmarking gated on machine availability -- Tier 3 per `.claude/skills/grinding-backlog/DEFINITIONS.md`, not appropriate to bundle into #2782's merge.

newhoggy:
## Summary

Fixes #2829.

`eval_owned_identity_alternative` (added by #2782) cloned `value` and `id` before
evaluating `//`'s left operand, so the original pair survived for a right-side splice that
usually never happens. `OwnedValue::clone` is a plain derive — a deep copy of whatever
subtree the value holds.

The clone is not really a property of `//`. `eval_owned_identity_stages` takes `value` **by
value** but *reads* it almost everywhere (navigation already borrows), so the by-value
signature is what forces every construct that runs a body more than once — or that may need
the original afterwards — to clone upfront. `//` is one member; `Expr::If`, `Expr::Comma`,
`Expr::Limit`, `Expr::AsPattern`, `eval_owned_identity_as`, `owned_identity_bind_values`,
`owned_identity_prefetch` and `eval_path_context_pipe_detached` are the rest. A `//`-only
fix would leave the family in place.

So this changes the signature, not a predicate: `value: Cow<'_, OwnedValue>` across the
seven core functions plus `continue_owned_identity_items`, and `OwnedIdentityTail::emit`
takes a `Cow` whose `Sink` arm calls `into_owned()` — the one place a value genuinely
becomes an output. A `Pairs` tail forwards the `Cow` untouched, so an enclosing stage that
only reads it never pays for a copy. **No new behavioural predicate**: a mistake is a
compile error, not a silently different answer. (That mattered — see below.)

## Scope correction, verified rather than inherited

The issue says `.foo // default` is "one of the most common jq/yq idioms" costing "an
O(document size) copy on every evaluation". It doesn't. `eval_owned_identity_alternative`
is reachable only when the pipe also reads path context. Re-verified on this branch with an
instrumented build (reverted afterwards):

| query | arm reached |
|---|---|
| `.big // []` | no |
| `.big // [] \| length` | no |
| `.a \| key // 99` | no |
| `.a.b // 7` | no |
| `(.big // []) \| path` | **yes** |
| `.a \| (.b // 9) , key` | **yes** |
| `.a \| (.big // []) \| key` | **yes** |
| `.[] \| (. // 9) \| key` | **yes, once per member** |
| `to_entries \| .[] \| (key // 9)` | **yes, once per member** |

Plain `//` never enters this route — the eager `eval::eval_alternative` answers it,
unchanged. The affected class is `//` **inside a pipe that reads `key`/`path`/`parent`**.
The win is real but it is not on the idiom's hot path, and the PR should not claim it is.

## Two things worth recording that the diff doesn't show

**The HRTB was a non-event.** The plan flagged
`&mut dyn FnMut(Cow<'_, OwnedValue>, ..)` as "the one place this could stall". rustc
inferred the higher-ranked bound immediately; all ~40 resulting errors were plain `E0308`
type mismatches, not a single lifetime error.

**The real trap was the one the plan predicted in a footnote.** After converting the
signatures, the upfront sites still read `value.clone()` — which now clones a **`Cow`**, and
cloning a `Cow::Owned` still deep-copies. It compiles, it passes the entire suite, and it
does absolutely nothing. Only the benchmark would have caught it. They are now
`Cow::Borrowed(&value)`, a genuine borrow whichever variant `value` holds.

Also: the new import first landed under a `#[cfg(not(test))]` belonging to the next line, so
`cargo check --features cli` was clean while every test target failed to build.
`--all-targets` is what catches that.

## Measurements

Interleaved A/B (A and B alternate *within* each repetition), 7 reps, median, output
identity gated by hash before any timing is believed. 8.4 MB input.

**Target shapes** — this machine (M5 Max, indicative; the pinned-box numbers are in a
follow-up comment):

| query | base | branch | delta |
|---|---:|---:|---:|
| `(. // 9) \| path` | 196.6 ms | 159.8 ms | **−18.7%** |
| `.arrays \| (.[0] // 9) \| path` | 113.8 ms | 97.7 ms | **−14.1%** |
| `.simple_values \| (. // {}) \| key` | 30.8 ms | 29.5 ms | −4.2% |

The triage plan's own scoping probe predicted −17.8% / −14.8% / −5.3%, which is close
enough to be a genuine confirmation that the clone is what was removed.

**Holdouts** — no `//` anywhere, so the `Cow` signature cannot help them; this is the
precheck rule (measure where the guarded code is already fastest):

| query | base | branch | delta |
|---|---:|---:|---:|
| `.` | 78.9 ms | 78.6 ms | −0.4% |
| `.arrays` | 39.7 ms | 39.3 ms | −1.0% |
| `.arrays \| .[0] \| path` | 18.6 ms | 18.7 ms | +0.5% |
| `. \| path` | 19.0 ms | 18.5 ms | −2.6% |
| `.simple_values \| key` | 18.5 ms | 18.6 ms | +0.5% |
| `[.arrays[] \| length]` | 21.1 ms | 20.9 ms | −0.9% |

All within noise: the signature change costs nothing on the pipes it cannot help.

Every row above gated `identical: YES` on output.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` — 63 legs, 0 failed, including #2782's
      own acceptance tests (`(.a, .b) // 9 | key`, `(key // 9) | parent`,
      `(1, error("x")) // 2`, `break` escaping the operator)
- [x] Interleaved A/B with output-identity gating, target shapes + holdouts
- [ ] 2-architecture A/B on the pinned boxes (ARM M4 Pro + 7950X) — running; this is the
      merge gate, since a memory-traffic change does not port across architectures
- [ ] `cargo clippy` both matrices, `cargo fmt`, rustdoc `-D warnings`, `no_std`
- [ ] `omni-dev coverage diff`


newhoggy:
## 2-architecture acceptance A/B — the merge gate

Both pinned boxes, both binaries built from source on the box itself (base `16bc13e11`, branch `ee7007506`), interleaved within each repetition, 7 reps, median, output identity gated by hash before any timing is believed. 8.4 MB generated input. Both boxes confirmed idle before the run (top process 0.0% on x86; the only `cargo|rustc|criterion` matches there were the check command's own cmdline).

### Target shapes — `//` inside a path-context-reading pipe

| query | M4 Pro base | M4 Pro branch | Δ | 7950X base | 7950X branch | Δ |
|---|---:|---:|---:|---:|---:|---:|
| `(. // 9) \| path` | 180.2 ms | 150.3 ms | **−16.6%** | 523.2 ms | 394.6 ms | **−24.6%** |
| `.arrays \| (.[0] // 9) \| path` | 104.5 ms | 89.3 ms | **−14.6%** | 298.4 ms | 231.5 ms | **−22.4%** |
| `.simple_values \| (. // {}) \| key` | 28.4 ms | 27.2 ms | −4.4% | 56.5 ms | 51.1 ms | **−9.7%** |

### Holdouts — no `//` anywhere, so the `Cow` signature cannot help them

| query | M4 Pro Δ | 7950X Δ |
|---|---:|---:|
| `.` | −0.5% | −0.9% |
| `.arrays` | +0.7% | −1.9% |
| `. \| path` | +0.2% | +0.2% |
| `[.arrays[] \| length]` | +0.5% | +2.2% |

Every row on both machines gated `identical: YES`.

### Reading

The win **ports across architectures, and is larger on x86** — worth stating explicitly because this repo's standing lesson is that memory-bound effects often do *not* port (#106 was 6.1x on one box and 16.4x on the other). Here both agree on direction and the 7950X sees more of it, consistent with a deep copy being removed rather than a code-layout artifact.

The third row is the one that moved most between machines (−4.4% vs −9.7%). It is the shallowest subtree of the three, so the copy it avoids is smallest; that it still shows on x86 is consistent with the same mechanism at smaller scale.

Holdouts are flat on both, which is the precheck rule satisfied: the signature change is charged to every owned-identity pipe, including the ones it cannot help, and it costs them nothing measurable.

### What this does not claim

Per the scope correction in the PR body, these are `//` **inside a pipe that reads `key`/`path`/`parent`**. Plain `.foo // default` never enters this route at all — it is answered by the eager `eval::eval_alternative`, unchanged by this PR. The issue's framing of the idiom's frequency overstates the reach; the mechanism and the measurement are real, the blast radius is narrower than filed.

newhoggy:
## `/code-review` (max effort) — one finding, fixed in b3ea74d69

The review independently validated the refactor: ~14,000 differential cases base-vs-PR with **0 diffs** (171 hand-built path-context rows, all 6300 cases of `scripts/jq-path-context-oracle-sweep.sh --list-cases`, a 6400-case `//`×prefix×leaf fuzz, 1104 try/catch rows), and it validated its own harness had power by mutating `if !v.is_truthy()` and getting 33 diffs. It also confirmed the clone accounting is conserved-or-better at every site, that no `Cow`-cloning trap was left behind, and that no deref-coercion or pointer-identity hazard exists.

### The finding: my test had coverage but no authority

`test_try_handler_reads_path_context_from_the_failing_stage_2829` reached 100% patch coverage of the line it was written for and would not have noticed if that line changed. I confirmed it by mutation rather than taking the claim — replacing the call's arguments with `Cow::Owned(OwnedValue::Null)` / `OwnedIdentity::detached()` left **all six assertions green**. Five rows never entered the pipe at all; the sixth executed the line but read neither argument.

Fixed by adding rows that read the raised payload *and* the path in one output (`try error("x") catch [., path]`). Re-checked both directions: the mutant now fails (`[null,[]]` against `["x",[]]`), the real code passes.

Worth stating plainly: **coverage says a line ran; only a mutant says a test would notice if it changed.** That is the rule this repo already applies to differential harnesses, and it applies just as well to a unit test written specifically to close a coverage gap — arguably more so, since such a test is written against a line rather than against a behaviour.

### On the non-blocking observation — mechanism right, reach wrong

The review flagged `owned_identity_step`'s `Rc::new(value.clone())` as a larger remaining clone, "on the ordinary `.a.b.c` path rather than only path-context pipes". The site and mechanism are right and worth pursuing; the reach is not. Instrumented counts (build reverted afterwards):

| filter | clones |
|---|---:|
| `.a`, `.a.b`, `.a.b.c` | **0** |
| `.a.b \| path`, `.a \| key` | **0** |
| `(. // 9) \| .a \| path` | 1 |
| `(. // 9) \| .a.b.c \| path` | 3 |

Every door into the owned-identity pipe is gated on `needs_path_context`, so plain navigation never reaches that arm. It is one deep copy *per navigation stage* inside the same path-context route this PR already lives in — and the `Slice` arm is worse, cloning *per child* because its clone sits inside the `map`.

Filed as **#2994** with the instrumented counts, the per-stage cost (~20–58 ms per added stage on 8.4 MB), and the corrected reach — `Cow` does not reach it, since it is a materialized parent handle rather than a value being passed along.

### State

Full suite 63 legs green, both clippy matrices, rustdoc `-D warnings`, `no_std`, patch coverage 100%. The 2-architecture acceptance A/B is in the comment above.

PR #2997, opened 2026-09-16
fix(jq): derive --seq -s EOF locations from the reader, not a record heuristic

### LINKED ISSUE #2947: jq: --seq -s EOF error locations lose recovery state around final RS and newline
**Severity: Low**

## Summary

`jq --seq -s` runtime error locations diverge from jq 1.7.1 around a final record separator and parser errors followed by newline. These are pre-existing differences found during #2672's before/after oracle matrix: 31 final-RS cases and 26 malformed/depth-overlimit plus newline cases.

## Repro

Write the following bytes to `/tmp/seq.json`, then run `jq --seq -s -c 'error("x")' /tmp/seq.json` and `succinctly jq --seq -s -c 'error("x")' /tmp/seq.json`. `RS` means byte 0x1e and `LF` means byte 0x0a. All cases exit 5 with empty stdout; the table describes the final runtime error location.

| Bytes | jq 1.7.1 | succinctly |
| --- | --- | --- |
| `RS 0 RS` | `<unknown>` | `/tmp/seq.json:0` |
| `RS true RS` | `<unknown>` | `/tmp/seq.json:0` |
| `RS [0,] LF` | `/tmp/seq.json:1` | `<unknown>` |
| `RS "\q" LF` | `/tmp/seq.json:1` | `<unknown>` |

Controls: `RS [] RS` reports file:0; `RS 0 LF RS` reports file:1. A space or tab after malformed JSON leaves `<unknown>`, while `RS [0,] RS LF` reports file:1. Neither treating every final RS as unknown nor treating every whitespace character as recovery matches jq.

## Root cause

`src/bin/succinctly/jq_runner.rs`, `seq_trailing_record_is_dropped`: selecting the suffix after the final RS and returning false for an empty suffix discards the preceding unresolved state.

The same file's `seq_record_ends_unresolved` returns immediately on an invalid token, without observing a subsequent newline that restores a reportable input position. The location heuristic separately approximates state already modeled by the sequence reader.

## Why this is separate from #2672

#2672 fixes valid JSON above the standalone validator's 128-level recursion limit. These cases reproduce at depth zero and remain unchanged by that fix. This also differs from #2653, whose scope is emitted values before malformed suffixes, rather than runtime diagnostic locations.

## Suggested fix direction

Track EOF location recovery across reader events, including unresolved tokens, parser errors, RS and newline, rather than inferring it only from the final trimmed record. Keep the control cases above and oracle-check both filename/line diagnostics and emitted values.

newhoggy:
## Summary

`--seq -s` decides whether a runtime error can name a file and line, or must answer
`(at <unknown>)`. That was answered by a local heuristic over the text after the last RS
byte — `seq_trailing_record_is_dropped` plus four helpers — which re-derived, approximately,
state `jq_seq_reader` already models exactly.

Reading jq 1.7.1's own `src/util.c` settles what the real rule is:

- `jq_util_input_get_position` renders `<unknown>` when `current_filename` is not a string,
  and the only thing that clears that field is `jq_util_input_read_more` **closing a stream
  that is already at `feof`** — and it re-sets the field immediately if another file follows.
- jq calls `read_more` one extra time exactly when its parser hands back an **error**, because
  an error is the one outcome that `return`s early out of the read/parse loop and makes `main`
  call back in. A value, or a quiet end-of-buffer, leaves via `while (!is_last || has_more)`
  with the filename intact.
- Which errors are fatal is then decided by jq's `fgets` chunking. `read_more` calls
  `fgets(buf, sizeof(buf), f)` on a `char buf[4096]`, so a chunk ends at a newline **or after
  4095 bytes**, whichever comes first; an error detected on a chunk that still has a refill (or
  a next file) behind it survives, and only one detected while jq is working on the stream's
  **final buffer** leaves nothing to refill from.

So the whole check collapses to one comparison — where did jq last fail, and is that inside
the final buffer — against a reader that already tracks the *moment of detection* rather than
what was wrong with a record. ~200 lines of heuristic deleted.

### What it fixes

The four shapes #2947 tabulates, plus families the same heuristic got wrong for the same
reason. Note the record text is identical within each pair — only the bytes *after* the
failure differ, which is precisely what a record-shaped heuristic cannot see:

| stream | jq 1.7.1 | before | after |
|---|---|---|---|
| `\x1e0\x1e` | `<unknown>` | `file:0` | `<unknown>` |
| `\x1e0\x1e\n` | `file:1` | `file:1` | `file:1` |
| `\x1e[0,]` | `<unknown>` | `<unknown>` | `<unknown>` |
| `\x1e[0,]\n` | `file:1` | `<unknown>` | `file:1` |
| `\x1e"unterm\n` | `<unknown>` | `<unknown>` | `<unknown>` |
| `""` (empty, no RS) | `<unknown>` | `file:0` | `<unknown>` |
| `a=\x1e[0,]`, `b=""` | `b:0` | `<unknown>` | `b:0` |
| `\x1e[0,]` + spaces to 4094 B | `<unknown>` | `<unknown>` | `<unknown>` |
| `\x1e[0,]` + spaces to 4095 B | `file:0` | `<unknown>` | `file:0` |

The `\x1e"unterm\n` row is the discriminator that rules out reading the newline as plain
"recovery": a newline swallowed by an unterminated string is just more string, so the failure
still lands at EOF, in the final buffer.

The `b=""` row is the file-boundary half — jq closes a file and clears `current_filename` on
the way to *opening* the next, so opening an empty trailing file is what puts a position back.
Measuring the final buffer from the **last file's** start, not the stream's, is what gets this
right.

The last two rows are the buffer size, isolated: no newline in either stream, and one byte
apart. At 4095 `fgets` stops on the size limit rather than on EOF, so a further (empty) buffer
follows.

Values are untouched: `parse_json_seq_with_ends` already ran through the same reader, and this
path only ever answered a location question.

### Performance

The rule reads the same walk jq's stderr diagnostics already make, so `for_each_warning` now
returns the offset and the call site runs **one** pass for both. A separate pass measured
**+15.0%** on a 12 MB `--seq -s -c length`; as landed it is **-1.1%** median against a
base-vs-base control noise floor of **-1.2%** — i.e. neutral. (Interleaved A/B, release build,
9 reps, output-identity gated, Apple M-series; the box was not dedicated, so treat it as a
regression check rather than a benchmark claim.)

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, green; re-run after every step)
- [x] All three CI clippy variants (`--all-features`, the CI feature set, and `broadword-yaml`), `-D warnings`
- [x] `cargo test --no-default-features` (no_std) and `RUSTDOCFLAGS=-D warnings cargo doc --no-deps --all-features`
- [x] Patch coverage **100% (63/63 new lines)** via `cargo llvm-cov` + `omni-dev coverage diff`

### Oracle differential sweeps

Every claim above was captured from the pinned `/usr/bin/jq` 1.7.1, not from memory. A
differential harness compares exit code, full stdout and full stderr over a token alphabet of
the shapes that can move jq's seq reader state — RS, newline, other whitespace, resolving
values, tokens *pending* at a boundary, hard parse errors, unterminated tokens, a bare literal,
and (after review) a lone invalid byte, a whole BOM, a partial BOM and valid multi-byte
characters — across `--seq -s -c 'error("x")'`, `--seq -s -c .` and `--seq -c .`.

**The harness was validated against the parent binary first** — "0 differences" is worthless
until a known-bad build separates. Sets of failing cases are compared, not just counts, so a
fix that trades one divergence for another cannot hide in the total:

| sweep | cases | parent diffs | this branch | new |
|---|---:|---:|---:|---:|
| single-file grid, token strings of length 1–3 | 13,104 | 64 | 12 | **0** |
| multi-file, every split of those strings | 12,816 | 278 | **0** | **0** |
| randomised, seeded, 1–7 tokens over 1–3 files | 4,000 | 164 | 110 | **0** |
| invalid UTF-8 / BOM grid, single- and two-file | 4,815 | 135 | 76 | **0** |

The residual differences are **pre-existing** families, each a strict subset of the parent's
own failing set — no case that was clean on the parent fails here:

- `--seq` **non-slurp** (which this change cannot reach — the location is computed under
  `if args.slurp`) drops a record introduced by an empty record when the stream has no trailing
  newline: `printf '\x1e\x1etrue' | jq --seq -c .` prints nothing while succinctly prints
  `true`. Filed as **#2998**.
- A **partial BOM** alone in the first source makes succinctly print a `jq: ignoring parse
  error:` line jq does not print at all (on `main` too). Filed as **#3002**.

One honest caveat on the BOM family. Because the location is now derived from the same walk
that produces the stderr diagnostics, a spurious *warning* now costs the *location* too.
Isolating just the final `jq: error (at ...)` line over 2,405 invalid-UTF-8/BOM cases: **69
locations moved onto jq's answer, 7 moved off it**. All 7 are #3002's family (`EF BF` alone in
the first source), and every one of them already diverged on the parent via that stderr line.

### Regression tests

Three tables in `tests/jq_cli_tests.rs`, 43 cases, all captured live from jq 1.7.1. **16 fail
on the parent commit**; the rest are controls that must not move, including the two groups that
pin regressions an intermediate version of this PR introduced and review caught (see the test's
own doc comment, which says which is which).

## Review

`/code-review 2997 high` found five items; all five are addressed:

1. **Regression — decoded text instead of raw bytes.** Real, reproduced, fixed in `21c29e9d3`,
   with a test. This was the important one.
2. **jq's buffer is 4 KiB, not a line.** Confirmed against `src/util.c` and the oracle (the
   crossover is exactly 4095 bytes). Now modelled properly, which also fixes a pre-existing
   divergence.
3. **PR body overclaimed** that all residual sweep diffs were unreachable. Corrected above, and
   the `\x1e1{` family the reviewer named is filed as **#3003**.
4. **Stale "unreachable" comment** on `finish`'s `WaitingForRs` arm. Updated — the new caller
   depends on it.
5. **Doubled reader pass.** Removed; see Performance above.

Fixes #2947


newhoggy:
Follow-up filed from this PR's differential sweep: **#2998**. It accounts for every one of the 17 residual differences reported in the description (the `--seq -c .` non-slurp rows) — pre-existing, and out of this PR's reach since the location is computed under `if args.slurp`.

newhoggy:
Review round addressed in `21c29e9d3`. Both regressions the review found were real and are reproduced in the new tests; the PR body is updated with corrected claims and precise numbers.

- **Raw bytes, not decoded text** — fixed, with a test. This was the important one: `\x80\n` answered `input_line_number` 1 against jq's 0.
- **jq's 4 KiB buffer** — confirmed against `src/util.c` and the oracle; the crossover is exactly 4095 bytes. Now modelled, which also fixes a divergence the parent had.
- **Doubled reader pass** — removed. `for_each_warning` returns the offset and the call site runs one walk for both: the separate pass measured **+15.0%** on a 12 MB `--seq -s -c length`; as landed it is **-1.1%** median against a **-1.2%** base-vs-base control floor.
- **Stale "unreachable" comment** — updated; the new caller depends on that arm.
- **PR body overclaim** — corrected, and the `\x1e1{` family is filed as **#3003**.

A third regression, not in the review, turned up while re-sweeping: a *complete* BOM split across two files leaves the reader scanning no bytes, so its running offset could not stand in for "at EOF". Fixed by tagging EOF-detected diagnostics explicitly, with a test.

Follow-ups filed: **#2998** (non-slurp value drop), **#3002** (partial-BOM spurious warning), **#3003** (`\x1e1{`).

PR #3008, opened 2026-09-16
fix(jq): name the line and echo the source for a module-body compile error

### LINKED ISSUE #2991: jq: a module-body compile error names its file but not jq's line number or source echo (#2951 residual)

newhoggy:
## Summary

- A module-body compile error already named the module's own canonical file (#2951), but not jq's trailing `, line N:` suffix or its echo of the offending source line.
- `report_compile_errors` (`src/bin/succinctly/jq_runner.rs`) now re-reads the module's file on this cold error path (never at load time — most runs never take it), builds the same call-site table it already builds for the main filter, and applies the same `nth`-occurrence rule to avoid re-citing an already-used position for a repeated name.
- Also fixed a gap found while implementing this: a namespaced call inside a module body (`def f: ns::g;`) fell through to the file-name-only report even with a real, findable occurrence in the module's own text, because `collect_call_sites` never records a namespaced call at all (`parse_namespaced_call` has no `call_sites.push`, unlike the plain-call path). Added the same text-search fallback the main filter's own diagnostics already use for this exact reason.
- Confirmed live against `/usr/bin/jq` 1.7.1, byte-for-byte, across: single-line and multi-line module errors, two unresolved calls in the same module, two unresolved calls in different modules, an `import` (namespaced) form, a namespaced call inside a module body, and `~/.jq`'s own auto-loaded defs.
- Updated `docs/compliance/jq/limitations.md` — the "no location at all" gap this closes, and the "diagnostic detail" note in the module-scope section.

Fixes #2991

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, including 3 new regression tests + updated pre-existing tests whose old assertions pinned the now-fixed behavior)
- [x] `cargo fmt -- --check`
- [x] Diffed output against pinned real `jq` 1.7.1 oracle for every scenario listed above
- [x] `cargo llvm-cov` — every new/changed line has nonzero hit count, except the final graceful-degradation fallback (file unreadable or call genuinely not found in its own source), which mirrors an already-unhit precedent elsewhere in the same reporter (the main filter's own analogous `None` arm)


newhoggy:
Ran \`/code-review\` on this PR and addressed its findings:

1. **Fixed**: the module branch's table-hit arm never updated its resume counter, unlike the main filter's analogous arm (a latent inconsistency, not currently triggerable but real). Refactored both call sites into a single shared \`report_unresolved_call\` helper so the two paths structurally cannot drift apart again — also closes the review's third finding (duplication between the two near-identical blocks).
2. **Filed as #3010, not fixed here**: the review found that \`locate_identifier_from\`'s text-search fallback has no comment/string awareness, so a namespaced call whose spelling also appears in an earlier comment cites the wrong line. Verified this is a **pre-existing bug on unpatched \`main\`**, already present for the top-level filter's own identical fallback (independent of this PR) — this PR just makes it reachable a second way (the module branch now shares the same fallback). Scoped out as its own follow-up rather than attempting a general comment/string-aware scanner here, which is a materially bigger change than #2991 itself asks for.

PR #3016, opened 2026-09-16
fix(jq): give a slice write's non-integer-bound refusal a kind-aware message

### LINKED ISSUE #2927: jq: a slice write re-reads a target that stopped being null, but reports the wrong refusal
**Severity: Low** (message wording on a write that fails either way; found by `/code-review` on PR #2926)

## Summary

jq's `INDEX` re-reads the target's *kind* at write time, so a slice component whose bound
it never parsed reports a kind-dependent error. succinctly's `through_slice` refuses such a
component before looking at `root`, so it always reports the bound error.

Only reachable when a slot stops being `null` between path resolution and the write — i.e.
an earlier path in the same fan-out replaced it.

## Repro

Verified live against `/usr/bin/jq` 1.7.1 on `main` @ PR #2926's branch:

```console
$ echo '{"a":null}' | jq -c '(.a, .a["x":]) = 5'

$ echo '{"a":null}' | succinctly jq -c '(.a, .a["x":]) = 5'
jq: error (at <stdin>:1): Array/string slice indices must be integers
```

Same for `|= 5`. Both tools error and both exit 5; only the sentence differs.

## Not a regression

Before #2926 the resolver raised `Array/string slice indices must be integers` for this
same program, at resolution rather than at the write — so the observable behaviour is
unchanged. #2926 moved *where* the refusal happens without changing *what* it says.

## Root cause

`through_slice` (`src/jq/eval.rs`) checks `SliceEditFlags::non_integer_bound` ahead of its
`match root`, because that flag is a property of the component rather than of the target.
The component can only be *produced* over a `null` target (#2853), but by the time the
write runs the slot may hold something else.

## Why the obvious narrowing is wrong

Gating the arm on `matches!(root, OwnedValue::Null)` and letting everything else fall
through would be worse than a wrong sentence. A `SliceBoundKey::Raw` bound carries
`start: None, end: None`, which every arm below reads as *the whole slice* — so an array
target would be silently spliced rather than refused. Any fix has to answer the
kind-dependent error explicitly, not delegate to the existing arms.

## Suggested fix direction

Give the arm a kind-aware refusal, mirroring `SliceTargetKind`'s existing ordering (#2546):
array/string target → `Array/string slice indices must be integers`; anything else →
`Cannot index <kind> with object`. Capture the full oracle matrix first (number, string,
array, object, boolean, and the `?`/`|=`/`del` spellings of each) — the `del()` row
especially, since `delpaths` never parses the descriptor and may well no-op regardless of
kind.

## Why this is separate from #2853

#2853 is about a path that *resolves* where succinctly used to raise. This is about a
target whose kind changed underneath an already-resolved path, which is a different
mechanism, has its own oracle matrix, and was not a divergence #2853 introduced.

Refs #2853, #2546, #2926.

newhoggy:
## Summary

- `through_slice`'s `non_integer_bound` arm (`src/jq/eval.rs`) always raised `Array/string slice indices must be integers`, even when the target stopped being `null` between path resolution and write — e.g. `(.a, .a["x":]) = 5` on `{"a":null}`: the first path in the fan-out replaces `.a` with `5` before the second path's write runs, and jq's own `setpath` re-reads `root`'s *current* kind at that point.
- Classified via `SliceTargetKind::of_type_name` — the same three-way split slice resolution already uses: `Sliceable` (array/string) and `Null` both still complain about the non-integer bound; anything else (number, object, boolean, ...) answers `Cannot index <kind> with object` without looking at the bound at all.
- `del()`/`|= empty` are unaffected — `delpaths` never parses the descriptor at all, so that no-op path is untouched.
- Confirmed live against `/usr/bin/jq` 1.7.1, byte-for-byte, across every target kind (number/string/array/object/boolean), the root-stays-null case (unchanged), a `|=` real-update variant, and the `del()`/`|= empty` no-op.

Fixes #2927

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, including a new regression test covering the full oracle matrix)
- [x] `cargo fmt -- --check`
- [x] Diffed output against pinned real `jq` 1.7.1 oracle for every row above
- [x] `cargo llvm-cov` — both arms of the new match have nonzero hit counts


newhoggy:
Re: the \`/code-review high\` finding about yq-mode behavior being asserted without a live capture —

Verified this doesn't apply: the new kind-aware match is **provably unreachable in yq mode**, not just untested there.

\`drive_slice_bound\` (\`src/jq/eval.rs\`, the function that produces the \`SliceBoundKey\`s \`slice_has_non_integer_bound\`/\`through_slice\`'s \`non_integer_bound\` flag consume) has this in its \`S::TAG == EvalTag::Yq\` arm:

\`\`\`rust
let bound = PathSliceBound::classify(raw, round);
if let Err(e) = &bound.bound {
    return Err(e.clone().into());
}
\`\`\`

A non-integer bound's classification fails there and raises **immediately**, before the bound ever reaches \`through_slice\` at all — matching the doc comment a few lines up: "yq mode cannot stream... the classification failure... yq raises before the caller sees any bound at all." So \`non_integer_bound\` can never be \`true\` for a yq-mode call; the new code this PR adds is dead code in that mode, not an unverified behavior change.

Confirmed empirically too, both pre- and post-fix, across every target kind (number/string/object) — identical \`Error: Array/string slice indices must be integers\` in yq mode either way, no observable change:

\`\`\`console
$ printf 'a: null\n' | succinctly yq -o=json -I=0 '(.a, .a["x":]) = 5'    # same before and after this PR
Error: Array/string slice indices must be integers
\`\`\`

No fix or limitations.md entry needed — there's no yq divergence to record, since yq's own behavior here is completely untouched by this change.

newhoggy:
Ran \`/code-review high\` on this PR (touches \`through_slice\`, a 6-call-site shared choke point in \`eval.rs\`). Findings and responses:

**Fixed — the real bug**: the original fix called \`edit\` unconditionally and only chose the kind-aware message afterward. jq's own \`getpath\`/\`setpath\` only ever run the update filter when the target is genuinely \`null\`; for any other kind they raise *before* the filter gets a chance to run at all. Confirmed live: \`(.a,.a["x":]) |= (debug|5)\` on \`{"a":null}\` prints \`DEBUG\` once in real jq, twice in the old fix (an extra execution of the update filter, with its side effects, once the target stops being null) — and a chained path's own inner error (\`.a["x":][0:1]\`, \`.a["x":]["y":]\`) masked the outer kind-aware refusal entirely, since \`edit\` is what recurses into the rest of the chain.

Rewrote to classify \`root\`'s kind **before** calling \`edit\`, short-circuiting immediately for \`Sliceable\`/\`Other\`; only \`Null\` still runs \`edit\` first (unchanged, matching the existing #1876/#1883/#2853 behavior). Verified against every repro the review gave, plus the full original oracle matrix — all byte-for-byte against jq 1.7.1 now, including the chained-path and doubled-debug cases. New regression test: \`test_slice_write_refusal_skips_edit_entirely_for_a_non_null_target_2927\`.

(While first writing this fix I briefly let the classification run unconditionally instead of inside the \`non_integer_bound\` gate — regressed the unrelated float-bound case, \`.[1.5:] = [9]\`. Caught immediately by the pre-existing \`test_slice_bound_controls_unmoved_2853\` test before it ever left the worktree; fixed in the same commit.)

**Verified as a non-issue, not fixed**: the yq-mode concern. \`drive_slice_bound\`'s yq arm (\`src/jq/eval.rs\`) raises a failed bound classification immediately, before it ever reaches \`through_slice\` — confirmed by 3 independent review passes (code tracing + live differential testing, ~1,000+ configurations) that \`non_integer_bound\` can never be \`true\` for a yq-mode call. No yq behavior change, verified or otherwise.

**Not fixed — pre-existing gap, unrelated to this PR**: the finding that nested-slice masking existed even *before* this PR for the two-non-integer-bounds-in-a-row case is real but pre-existing (confirmed: same wrong answer on unpatched \`main\`). Not attempting a fix here since it's out of scope for #2927's own repro.

Thanks to all 6 review passes — this caught a real regression before merge.

Issue #3047, opened 2026-09-16
yq: a nested operand's as-body reads absent keys outside the enclosing operand's read-only scope


newhoggy:
**Severity: Low**

## Summary

In `succinctly yq`, an arithmetic (or `and`/`or`) operand's read-only scope (#2470) is supposed to cover the operand's *whole* subexpression, nested pipes included. It stops covering an `as` body inside a nested operand: an `x as $v | BODY` whose source is itself an arithmetic operand runs `BODY` with the scope switched **off**, so an absent-key read there (`.zzz | key`) is answered as if it were outside any operand.

## Repro

`n: 0` as input, default flags (no `--jq-extensions`), yq v4.53.3 vs `succinctly` at `cc5bcdf52`:

| filter | yq | succinctly |
|---|---|---|
| `((.n + 1) as $m \| [.zzz \| key] \| length) + 0` | `0` | `1` |
| `[(.n + 1) as $m \| [.zzz \| key] \| length] + []` | `[0]` | `[1]` |
| `[first((.n + 1) as $m \| [.zzz \| key] \| length)] + []` | `[0]` | `[1]` |
| `[(.n + 1) \| [.zzz \| key] \| length] + []` (plain pipe, control) | `[0]` | `[0]` |

## Root cause

`read_only_operand_strategy` (`src/jq/eval.rs`) and its twin in `src/jq/eval_generic.rs` wrap an operand's sink in `yq_read_only_context::suspend()`, which sets the scope to `false` unconditionally. That is right for a *top-level* operand's consumer (`.zzz + 1 | .yyy`), but when the operand is itself nested inside an enclosing operand's scope, its consumer — here the `as` body, which runs on the stack inside the source's sink — should get the *enclosing* state back (`true`), not `false`. The eager route restores the previous state on return, which is why the plain-pipe control and the eager paths agree with yq.

## Why this is separate from #2918

Found while fixing #2918, which made `recurse`'s `f` run through the lazy evaluator. #2918's PR restores the walk's starting scope at each native `recurse` level, so `recurse` itself stays consistent with its old queued order, but the `as`-body case above reproduces on `main` before that change, with no `recurse` involved.

## Suggested fix direction

[...]

Issue #3031, opened 2026-09-16
jq: a pipe stage after a parenthesized `?//` bind keeps a stop the retry absorbed, dropping later outputs


newhoggy:
**Severity: Low.** The output is silently wrong: later outputs are dropped, or an error jq absorbs is raised instead. It only happens in one shape: a **parenthesized** `?//` bind whose last alternative yields nothing, followed by a pipe stage, under a short-circuiting consumer (`first`/`limit`/`nth`/`isempty`/`label`+`break`) or with an error raised downstream.

Found while triaging #2968. It is why that issue's desugared row #7 disagreed with jq.

## Repro

`B` stands for `[1] as [$x] ?// $x | if $x == 1 then 9 else empty end`. Its first alternative yields `9`. The second alternative, retried after a stop, yields nothing. Captured against jq 1.7.1 and `main` `b509a8019`; `succinctly jq -n` and the document route (`echo 0 | succinctly jq`) give the same answer on every row:

| filter                                        | jq 1.7.1        | succinctly |
|-----------------------------------------------|-----------------|------------|
| `[first(((B) \| 3), 5)]`                      | `[3,5]`         | `[3]`      |
| `[limit(1; ((B) \| 3), 5)]`                   | `[3,5]`         | `[3]`      |
| `[label $o \| (((B) \| 3), 5) \| ., break $o]` | `[3,5]`         | `[3]`      |
| `[isempty(((B) \| 3), 5)]`                    | `[false,false]` | `[false]`  |
| `[first(((B) \| 3), 5), 6]`                   | `[3,5,6]`       | `[3,6]`    |
| `[first(((B) \| 3 \| 4), 5)]`                 | `[4,5]`         | `[4]`      |
| `[try (((B) \| error("e")), 5) catch "c"]`    | `[5]`           | `["c"]`    |

Controls that already match jq:

| filter                                  | both    | why it differs                                 |
|-----------------------------------------|---------|------------------------------------------------|
| `[first((B \| 3), 5)]` (no parentheses) | `[3,5]` | `\| 3` is inside the bind's body, not a pipe stage |
| `[first(((B)), 5)]`                     | `[9,5]` | no pipe stage after the bind                   |
| `[((B) \| 3), 5]`                       | `[3,5]` | no consumer stop, no downstream escape         |

[...]

Issue #636, opened 2026-08-04
jq: recurse(f; cond) silently prunes on a cond/f error instead of propagating it like jq does


newhoggy:
## Summary

`builtin_recurse_cond` and `resolve_recurse` (`src/jq/eval.rs`, the same
functions #627 fixed) both swallow an error raised while evaluating `cond`
(or `f`) and treat it as "prune this child" — the same behavior as a falsy
`cond`, not a query error. Real jq's `recurse(f; cond)` is `def r: .,
(f | select(cond) | r); r;`; an error raised inside `cond` is not caught by
anything in that definition, so it propagates as a fatal error for the
whole pipeline.

This predates #627: the pre-fix code already had `Err(_) => false` for
`cond` errors; #627's fix moved the check to per-child granularity but kept
the same swallow-on-error behavior (see the doc comment added at
`builtin_recurse_cond`, which explicitly documents this as a deliberate,
unchanged choice — filing this so the divergence itself is tracked, not
disputing that #627 was right to leave it alone).

## Repro (verified against jq 1.7.1)

```
$ echo '1' | jq -c '[limit(20; recurse(.+1; if . > 3 then error("boom") else . < 10 end))]'
jq: error (at <stdin>:1): boom

$ echo '1' | ./target/release/succinctly jq -c '[limit(20; recurse(.+1; if . > 3 then error("boom") else . < 10 end))]'
[1,2,3]
```

jq surfaces the error and aborts the whole query. succinctly silently stops
recursing into the branch that errored and returns the partial result as if
nothing happened.

## Related

Found during review of #627 (`fix(jq): recurse(f; cond) evaluate cond
per-child instead of collapsing it`). See also #635 (same functions'
breadth-first vs jq's depth-first traversal order), also found during that
review.

PR #3050, opened 2026-09-16
fix(jq): stop recurse's f and cond between their own outputs

### LINKED ISSUE #2918: jq: recurse's f is not suspended between its own outputs, so a node's later f outputs still fire (#2693 residual)

newhoggy:
## Summary

Fixes #2918. `recurse(f)` / `recurse(f; cond)` now stop *between* `f`'s (and `cond`'s) own outputs, as jq's `def r: ., (f | select(cond) | r); r;` does: each output's whole subtree is traversed before `f` is asked for the next. Values never differed; side effects did.

| input | filter | jq 1.7.1 stderr | before | after |
|---|---|---|---|---|
| `[1,[2,[3]]]` | `[limit(2; recurse(.[]?\|debug))]` | 1 DEBUG | 2 DEBUG | 1 DEBUG |
| `{"a":1}` | `[recurse(.[]?, ("x"\|stderr\|error("f")))?]` | `x` | `xx` | `x` |
| `{"a":{"x":1},"b":{"y":2}}` | `[recurse(.[]?\|debug)]` | interleaved | grouped per node | interleaved |
| `{"a":1,"b":2,"c":3}` | `[limit(2; recurse(.[]?; (true, ("c"\|stderr\|true))))]` | *(empty)* | `ccc` | *(empty)* |
| `[0,1]` | `[limit(3; recurse(.[]?\|debug; (.\|debug\|true)))]` | `0,0,1,1` | `0,1,0,1` | `0,0,1,1` |
| `0` | `[limit(3; recurse(.+1, ("s"\|stderr\|.+100)))]` | *(empty)* | `ss` | *(empty)* |
| `{"a":1,"b":2}` | `path(limit(2; recurse((.a\|debug), (.b\|debug))))` | 1 DEBUG | 2 DEBUG | 1 DEBUG |

### How

Following the issue's plan (option 1), with two changes forced by measurement:

1. **Refactor (byte-identical):** both walkers — `each_recurse_walk` (shared by `eval.rs` and `eval_generic.rs`) and `resolve_recurse_sink` (path mode) — become "visit the root, then expand", over one `RecurseAbort` outcome, so the old explicit stack can take over from an already-delivered node.
2. **Native order:** `expand` drives `f` through its sink form (`eval_each_owned` / `resolve_against_cow_sink`) and visits each child from inside the callback; `cond` the same way. Nothing is suspended — the child's subtree just runs before the sink returns to `f`. An abort from a subtree answers `Demand::Stop`, so `f` never reaches its later outputs; #636/#842/#854/#856's rules fall out of the recursion's shape.
3. **Budget — measured, not a constant.** Per-level stack cost depends heavily on `f`'s shape (release: `.+1` 1.4 KB, `if … end` 11 KB, 8 nested `if`s 25 KB, +8.5 KB per `(.+0)` pipe stage; a 30-stage pipe with a per-level constant overflowed the CLI's 256 MB thread). The walk records a stack address at start and checks the real distance before each level: **at most 512 KiB**, so it is safe for a library caller on an ordinary 2 MiB thread (a CLI-sized budget aborted `jq_computed_key_tests`' 10000-level `[recurse(.a?)]`). The spend is also charged on the ambient frame depth `MAX_EVAL_FRAMES` uses (queues if a walk starts past a quarter of it). Past the budget the old queued order resumes for the remaining levels — 46 levels of a reindex-route `f` in release, 9 in debug.
4. **Ambient scope:** each native level restores the yq read-only operand scope the walk started under, since a lazy operand inside `f` suspends it in its sink (the new yq test showed `[0,1]` vs the queued `[0,0]` without it).

### Measurements (Apple M5 Max, busy shared laptop — indicative only; release `cgu1+fat`, interleaved, outputs identical)

| input | query | time | peak RSS |
|---|---|---|---|
| users 10 MB | `[recurse(.[]?)] \| length` | neutral | 380 → 383 MB |
| users 10 MB | `[path(recurse(.[]?))] \| length` | neutral | 125 → 85 MB |
| nested 10 MB | `[recurse(.[]?)] \| length` | neutral | 513 → 569 MB |
| 200-deep chain, 1 MB bulk at leaf | `[recurse(.[]?)] \| length` | neutral | ~1.0 → ~1.3 GB |

Memory rises where `f` takes the reindex bridge (`.[]?` does): each native level's temporary document stays alive while its children run. Bounded by the 512 KiB budget's depth; documented in the walker and `limitations.md`. Not run on the pinned M4 Pro / 7950X boxes; the perf-guard CI leg covers instruction counts.

### Review round 1 (`/code-review high`), all fixed in `2bc4d99bb`

1. **Halt through a `?//` inside `f` retried.** A deeper level's escape reached `f`'s generator as a bare `Demand::Stop`. `recurse(. as [$a] ?// $a | if type == "number" then ("h\n"|halt_error(3)) else .[] end)` on `[1]` halted twice (jq: once). `stop_on_abort` now marks `nonretryable_stop` like `stop_with_escape` (halt, uncatchable error, item cap). The reviewer's ordinary-error variant already matches jq on this branch; the remaining plain-pipe shape is pre-existing and belongs to #3031 (repro added there).
2. **The consumer ran outside the yq read-only scope.** `[recurse((.n + 1) as $m | select($m < 3) | {"n": $m}) | [.zzz | key] | length] + []` gave `[0,1,1]` (queued: `[0,0,0]`). The scope is now re-established around every node's delivery and expansion.
3. **The 512 KiB budget was per walk.** `recurse(recurse(.[0]?))` nests a walk per level. Nested walks now measure from the outermost live level (`recurse_native_origin`).

Each fix has a test that fails without it (mutation-checked).

### Follow-up

- #3047 — pre-existing (on `main`, no `recurse` involved): a nested operand's `as` body reads absent keys outside the enclosing yq read-only scope (`((.n + 1) as $m | [.zzz | key] | length) + 0` → `1`, yq `0`).

## Test plan

- [x] `recurse_runs_f_only_on_demand_2693` gains the 7 rows above plus a stays-the-same #854 row, every one captured whole (stdout, stderr, exit) from jq 1.7.1; `-n` route row added. `recurse_still_finishes_a_nodes_f_before_descending_2918` removed (it was written to trip when this closed)
- [x] `recurse_native_and_queued_orders_agree_2918` (unit): 34 filters × 3 documents run fully queued vs native (value, path, cond, errors, break, limit/first/isempty, reduce/foreach, nested walks, `del`/`|=`, 10000-level chain across the budget handoff) — identical values and ending; mutation-checked (a cond gate ignoring falsy verdicts, and a null-node gate that descends, both fail it); runs on the 2 MiB harness thread
- [x] `recurse_native_budget_shares_the_frame_guard_2918` (unit): byte cap, frame charge, starting-depth ceiling, test level override
- [x] `recurse_queues_past_its_native_stack_budget_2918` (CLI): depth 3 matches jq, depth 1000 keeps the residual
- [x] `recurse_levels_keep_the_read_only_operand_scope_2918` (yq CLI): next level's `f` and the consumer
- [x] `nested_recurse_walks_share_one_native_stack_budget_2918` (unit, 2 MiB harness thread)
- [x] `?//` halt and error rows in `recurse_runs_f_only_on_demand_2693`, captured from jq 1.7.1
- [x] yq v4.53.3 rejects `recurse(f)`/`recurse(f; cond)` at the lexer, so no yq oracle rows (ADR-0018)


newhoggy:
Automated review ran at `/code-review high` (3 findings, all fixed with mutation-checked tests; a 4th scope gap found by a second session is fixed too). This change reorders side effects in a shared walker and adds a stack-address budget, so a human may want the deeper `/code-review ultra 3050` pass. It isn't blocking the merge.

Issue #3051, opened 2026-09-16
jq: a bad --slurpfile/--rawfile/--argjson exits 1 through anyhow; jq exits 2


newhoggy:
**Severity: Low**

## Summary

A bad `--slurpfile`, `--rawfile` or `--argjson` argument exits **1** through `anyhow` (`Error: …` / `Caused by:`). jq exits **2** with its own one-line diagnostic. The failure itself is right; the exit code and error channel aren't.

## Repro

`t1.json` = `{"ok":1}\n[1,2,`. Captured against `/usr/bin/jq` 1.7.1 and `succinctly` on the #2961 branch (these paths are untouched there):

| invocation | jq 1.7.1 | succinctly |
|---|---|---|
| `-nc 1 --slurpfile x t1.json` | exit 2, `jq: Bad JSON in --slurpfile x t1.json: Unfinished JSON term at EOF at line 2, column 5` | exit 1, `Error: Invalid JSON in stream` / `Caused by: EOF while parsing a value at line 2 column 5` |
| `-nc 1 --slurpfile x /nonexistent` | exit 2, `jq: Bad JSON in --slurpfile x /nonexistent: Could not open /nonexistent: No such file or directory` | exit 1, `Error: Failed to read file for --slurpfile x` / `Caused by: …` |
| `-nc 1 --rawfile x /nonexistent` | exit 2, `jq: Bad JSON in --rawfile x /nonexistent: Could not open …` | exit 1, `Error: Failed to read file for --rawfile x` / `Caused by: …` |
| `-nc 1 --argjson x '[1,'` | exit 2, `jq: invalid JSON text passed to --argjson` + usage hint | exit 1, `Error: Invalid JSON for --argjson x` / `Caused by: …` |

## Root cause

`build_context` (`src/bin/succinctly/jq_runner.rs`) propagates each argument's failure with `?`/`with_context` into `anyhow`, which the binary's `main` turns into exit 1 with the `Error:`/`Caused by:` rendering. jq treats these as usage errors (exit 2).

## Why this is separate from #2961

#2961 fixed the *document input* stream (prefix kept, parse error at exit 5). `--slurpfile` shares `parse_json_stream` with that code but is a command-line argument, not input. The triage plan for #2961 said to keep it all-or-nothing and not move its exit code silently, so it was left as-is.

## Suggested fix direction

[...]

Issue #2878, opened 2026-09-12
jq: a raw control character in a string is accepted on --slurpfile and primary input; jq rejects it


newhoggy:
**Severity: Low** (accept-where-jq-refuses, on malformed JSON only — no wrong value for well-formed input)

## Summary

A raw, unescaped control character inside a JSON string is rejected by jq on every path and accepted by `succinctly` on `--slurpfile` **and on the primary document input**. RFC 8259 requires `U+0000`–`U+001F` to be escaped; `json::validate::Validator` enforces that (`ValidationErrorKind::ControlCharacter`), but neither of these paths runs it — they delimit values with `find_json_values`/`serde_json::Deserializer::byte_offset()` and then materialize each span with this crate's own decoder, which tolerates the raw byte.

Found while resolving #2052 (which moved `--argjson`/`--jsonargs`/`--seq` onto the crate's own validator; those three reject it correctly).

## Repro

`printf '["tab\tin"]'` — a literal TAB inside the string — captured live against `/usr/bin/jq` 1.7.1:

```console
$ jq -c . f.json
jq: parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line 1, column 9
$ succinctly jq -c . f.json
["tab	in"]                          # accepted, raw TAB round-tripped

$ jq -nc --slurpfile x f.json '$x'
jq: Bad JSON in --slurpfile x f.json: Invalid string: control characters ...
$ succinctly jq -nc --slurpfile x f.json '$x'
[["tab\tin"]]                        # accepted, and re-escaped on output

$ succinctly jq -nc --argjson x '["tab<TAB>in"]' '$x'   # correctly rejected (#2052)
```

The `--slurpfile` output is the more surprising half: the raw byte is silently *normalized* to `\t` on the way out, so a malformed document becomes a well-formed one with no diagnostic.

## Root cause

`src/bin/succinctly/jq_runner.rs` — `parse_json_stream` (backing `--slurpfile` and, per its own doc comment, the primary input path) validates the *stream shape* with `serde_json::Deserializer` only to find each value's byte span, then hands the span to `json_bytes_to_owned_value_checked`. `serde_json`'s span walk is not consulted for string content at that granularity, and the decoder has no control-character rule. `find_json_values`, the non-serde fallback, is explicitly "boundary-only" and has none either.

## Why this is separate from #2052

[...]

PR #3052, opened 2026-09-16
fix(jq): keep a malformed input stream's clean prefix on both input routes

### LINKED ISSUE #2961: jq: a rejected multi-value document prints no partial output; jq streams the clean prefix first
**Severity: Low**

## Summary

When a multi-value document's *later* value is malformed, jq prints every clean value
that precedes it and *then* exits 5. `succinctly` prints **nothing** and exits 5. The
exit code agrees; the streamed prefix does not.

This is pre-existing and not caused by #2878 — but #2878 made it reachable for a new
input class (a raw control character), which is how it was noticed, and it is now
recorded in `docs/compliance/jq/limitations.md`.

## Repro

Captured live against `/usr/bin/jq` 1.7.1:

```console
$ printf '{"ok":1}\n[1,2,' > t.json        # truncated *second* value
$ jq -c . t.json;          echo $?          # {"ok":1}  then parse error   -> 5
$ succinctly jq -c . t.json; echo $?        # (nothing)                    -> 5
```

The same shape with a raw control character in the second value (`{"ok":1}\n["a<TAB>b"]`)
behaves identically as of #2878 — before it, succinctly accepted that document outright.

## Root cause

`src/bin/succinctly/jq_runner.rs` — `find_json_values` resolves **every** value's span
for the whole input before any value is emitted, so a failure anywhere discards the
values already scanned. jq's parser is incremental: it emits each value as it completes
one, so a later failure cannot retract an earlier success.

## Why this is separate from #2878

#2878 changed *which inputs* are rejected (accept-vs-reject fidelity) and deliberately
did not change *how* a rejected document is reported. The all-or-nothing shape predates
it and applies equally to the truncated-value case above, which has behaved this way for
as long as the splitter has rejected truncated input. Fixing it means making the splitter
incremental, which is a different change with its own performance question, and would
have been scope creep on a correctness fix.

Pinned (as current behaviour, not as desired behaviour) by
`test_control_character_in_a_later_value_rejects_the_whole_document_2878` in
`tests/jq_cli_tests.rs`, which asserts stdout rather than only the exit code.

## Suggested fix direction

Emit each value as its span resolves rather than collecting all spans first, so a
failure at value *n* still leaves values `0..n` printed. Worth checking first whether
`--seq`'s own reader (`jq_seq_reader.rs`) already has the incremental shape this needs,
since it reasons per-record by design — if so, the question is whether the primary path
can share it rather than grow a second one. Measure: the primary input path is the
crate's hottest route, and any change that emits before the whole input is scanned
interacts with the M2 streaming path.

newhoggy:
## Summary

Fixes #2961. When a multi-value JSON input stream turns malformed partway, jq's incremental parser has already processed the values before the bad one. It prints them, then reports the parse error. succinctly printed nothing. On the materializing route (`-s`, `-n`, `input`/`inputs`/`input_line_number`) it also exited **1** through `anyhow` with an `Error:`/`Caused by:` prefix jq never prints.

Fixtures: `t1` = `{"ok":1}\n[1,2,`, `t3` = `1 2 {invalid} 3\n`, `t4` = `1 2 }`, `t5` = `1 2 "abc`, `good` = `1\n2\n`. jq column captured from `/usr/bin/jq` 1.7.1.

| invocation | jq 1.7.1 | before | after |
|---|---|---|---|
| `-c . t1` | `{"ok":1}` / 5 | *(nothing)* / 5 | `{"ok":1}` / 5 |
| `-c . t4`, `-c . t5` | `1 2` / 5 | *(nothing)* / 5 | `1 2` / 5 |
| `-c . good t1` | `1 2 {"ok":1}` / 5 | `1 2` / 5 | `1 2 {"ok":1}` / 5 |
| `-c -s . t1` | *(nothing)* / 5 | `Error: Invalid JSON in stream` / **1** | *(nothing)* / 5 |
| `-nc input t1` | `{"ok":1}` / 0 | `Error: …` / **1** | `{"ok":1}` / 0 |
| `-nc 'input, input, input' t3` | `1 2` / 5 | *(nothing)* / 5 | `1 2` / 5 |
| `-nc '[limit(2; inputs)]' t3` | `[1,2]` / 0 | *(nothing)* / 5 | `[1,2]` / 0 |
| `-c input_line_number t4` | `0 0` / 5 | `Error: …` / **1** | `0 0` / 5 |
| `-c '[., (try input catch "c")]' t1` | `[{"ok":1},"c"]` / 0 | `Error: …` / **1** | `[{"ok":1},"c"]` / 0 |
| `-nc '(try input catch null) ×3, (try input catch .)' t3` | `1 2 null "break"` / 0 | *(nothing)* / 5 | `1 2 null "break"` / 0 |

### How

- **Lazy per-file route (`jq_runner.rs`):** `split_json_values` is `find_json_values` without discarding the spans found before a failure. The per-value loop processes them, then reports `Invalid JSON text` (the same report as before, just after the prefix). Halts, `-e`, M2, identity raw-bytes and the general path are all per-span, so they're unaffected. **The scan is byte-for-byte the work it did before**, just returning what it already collected. That is why I didn't take the triage plan's per-value iterator: it would have put the hottest input route up for a two-architecture A/B for no observable gain.
- **Materializing route:** `parse_json_stream_prefix` has the same accept-set as `parse_json_stream` value by value (strict pass, then the splitter plus checked-materializer fallback), but keeps the prefix and the first error. `get_inputs` stops reading there, since jq never reads past a parse error, even in later files. `--slurp` stays all-or-nothing, now reported in jq's channel at exit 5.
- **The `input` queue (`src/jq/eval.rs`):** `remaining_inputs` gains a one-shot trailing error. `pop_input()` → `InputPop::{Document, ParseError, Exhausted}`.
  - `input` raises the error as an ordinary catchable error. It's queued as a plain `EvalError`, because the materializer's own error is an uncatchable decode failure and jq's `try input catch .` catches it.
  - `inputs` re-raises it; `first(inputs)` never reaches it, as in jq.
  - The driver loop reports it uncaught (exit 5) and stops.
  - Later reads are `break`. The `(at file:line)` marker moves to where the parser stopped, and stays there.
  - Public API is additive: `seed_remaining_inputs_with_error`, `pop_input`, `InputPop`. `seed_remaining_inputs`/`pop_remaining_input` are unchanged (the latter reads a trailing error as `None`).

### Remaining, recorded (pre-existing, not changed)

- Wording: `Invalid JSON text` vs jq's parser text (`jq: parse error: Unfinished JSON term at EOF …`).
- Lazy route: a splitter error moves on to the next file, where jq stops the whole stream. This is the #355 continue-past-error rule; the `input`/`inputs` route does stop.
- `-c . t3`: a member the splitter accepts but the materializer rejects is a per-document error there (`1 2 3`, jq `1 2`), also #355.
- `--slurpfile` on a malformed file still exits 1 (jq: 2, `Bad JSON in --slurpfile …`). It is untouched here and filed as #3051, with `--rawfile`/`--argjson`.

`docs/compliance/jq/limitations.md`'s "a rejected document produces no partial output" paragraph is replaced accordingly.

### Review round 1 (`/code-review high`), fixed in `2952efbab`

1. **Marker line after a parse error** counted every newline in the file. jq counts through the end of the malformed value's own line: `1\n2\n}\n3\n4\n5\n6\n` reports 3 (was 7), and `1\n2 }\n\n\n` reports 2 (was 4). For a multi-line malformed value jq names its parser's internal position, and for one cut off at EOF `<unknown>`; both are recorded in `limitations.md`, not modelled.
2. **`input_line_number`** didn't move to the error's line when the error was read. `[input,input,(try input catch 0),input_line_number]` now gives `[1,2,0,2]`, as jq does.
3. **Collecting `inputs`** (`.a = inputs`, `{a: inputs}`, `inputs + 1`) dropped the documents before the error. `builtin_inputs` now returns them as a `Partial`.

Each fix is pinned by a row that fails without it (mutation-checked).

## Test plan

- [x] `malformed_later_value_keeps_the_clean_prefix_2961` (CLI): 27 rows across both routes (truncated / unterminated / unmatched / member-malformed × identity, M2, general, `-e`, `-S`, multi-file, `halt`, `-s`, `input`, `inputs`, `limit`/`first`, `input_line_number`, caught errors, `break` after), plus clean-stream negative controls. stdout and exit code match jq 1.7.1 on every row, and stderr is always jq's channel (`jq: error`, never `Error:`/`Caused by`). The marker after a caught error is pinned (`…:1): x`).
- [x] `test_control_character_in_a_later_value_keeps_the_clean_prefix_2878_2961` (inverted from #2878's pin of the old empty output)
- [x] Unit: `split_json_values_keeps_the_prefix_before_a_failure_2961`, `parse_json_stream_prefix_stops_at_the_first_malformed_value_2961`
- [x] `tests/jq_input_queue_trailing_error_test.rs`: once-only delivery, `Exhausted` after, marker placement, `pop_remaining_input` compatibility
- [x] `--slurpfile` behaviour unchanged (exit 1, same message), checked by hand


newhoggy:
Automated review ran at `/code-review high` (3 findings, all fixed and mutation-checked). This changes the public `input` queue API (additive) and the primary input route's error handling, so a human may want the deeper `/code-review ultra 3052` pass. It isn't blocking the merge.

Issue #3030, opened 2026-09-16
jq: a multi-path `|=` deletes each `empty` result immediately, shifting later indices; jq 1.7.1 deletes them all at the end


newhoggy:
**Severity: High.** It silently deletes the *wrong* elements at exit 0, on an ordinary `select`-shaped update.

## Summary

When a jq-mode `|=` resolves to more than one path, and the update filter produces no output for some of them (`empty`, `select(false)`), succinctly deletes each such path **immediately**, inside the per-path write loop. Every later array index is then resolved against a shifted array.

jq 1.7.1 **collects** those paths and deletes them all at once, in a single `delpaths` after the loop. Every row below fits that model, and no row fits immediate deletion.

Found while triaging #2974 ([triage comment](https://github.com/rust-works/succinctly/issues/2974#issuecomment-5699254923)).

## Repro

Oracle `/usr/bin/jq` 1.7.1 vs `succinctly jq` on `main` @ `b509a8019`:

| input               | filter                                             | jq 1.7.1    | succinctly  |
|---------------------|----------------------------------------------------|-------------|-------------|
| `[1,2,3,4]`         | `(.[] \| select(. % 2 == 1)) \|= empty`            | `[2,4]`     | `[2,3]`     |
| `[1,2,3,4]`         | `(.[] \| select(. <= 2)) \|= empty`                | `[3,4]`     | `[2,4]`     |
| `[1,2,3,4]`         | `(.[0], .[1]) \|= empty`                           | `[3,4]`     | `[2,4]`     |
| `[1,2,3,4]`         | `(.[0], .[1]) \|= select(false)`                   | `[3,4]`     | `[2,4]`     |
| `[1,2,3,4]`         | `.[(0,1)] \|= empty`                               | `[3,4]`     | `[2,4]`     |
| `[1,2,3]`           | `(.[-1], .[-2]) \|= empty`                         | `[1]`       | `[2]`       |
| `[1,2,3]`           | `(.[0], .[0]) \|= empty`                           | `[2,3]`     | `[3]`       |
| `[1,2,3]`           | `.[(0,0)] \|= if . == 1 then empty else 9 end`     | `[2,3]`     | `[9,3]`     |
| `[1,2,3]`           | `(.[0], .[1]) \|= if . == 1 then empty else 9 end` | `[9,3]`     | `[2,9]`     |
[...]

Issue #2999, opened 2026-09-16
design: structural sharing in OwnedValue (Rc containers, copy-on-write) — Phase 2 of #2976


newhoggy:
Refs #2976 (Phase 1 landed as #2990; the Phase 2 plan and class breakdown are in https://github.com/rust-works/succinctly/issues/2976#issuecomment-5692714710). This issue is the design record; the deliverable is ADR-0024, filed **Proposed** and decided by measurement.

Experiment B (the box-only mechanism holdout) is tracked separately as #3000; Experiment C and the ADR itself are this issue's work and are **Fable**-class.

`docs/compliance/jq/limitations.md` (the `.[(0,1)] = 0` residual paragraph) currently points at #2976; the first Phase 2 PR retargets it here.

## Context

Every write route that must keep an unmutated document alongside a written one pays a whole-tree
`OwnedValue` clone: `eval_assign_streaming` (`result = pristine.clone()`), `fork_rhs_over_paths`
(one clone per RHS output), `set_path`'s `Iterate` arm (`new_value.clone()` per slot), the
streaming write's per-path `value.clone()`, and `eval_update` if #2974 takes the streaming route.
jq pays nothing for the same separation because `jv` is refcounted and `jv_setpath` copies on write.

## Facts measured against the current tree (2026-09-16)

- `size_of::<OwnedValue>()` = 72 (`Object(IndexMap)` = 72 is the widest arm; `Vec` = 24,
  `String` = 24, `NumberRepr` = 16, `Box<str>` = 16).
- **Any** container-boxing layout — `Object(Box<IndexMap>)` alone, `Rc<Vec>`+`Rc<IndexMap>`, or
  `Rc<str>`+`Rc<[Self]>`+`Rc<IndexMap>` — measures **32 bytes**, not the ~40 #2976 estimated:
  `NumberLiteral(NumberRepr, Box<str>)` becomes the widest arm at exactly 32 and the discriminant
  packs into `NumberRepr`'s niche. That is −55% per array element on every route, eager included.
  For #2976's 200,000-element repro one copy drops from 14.4 MB to 6.4 MB before any sharing.
- `Rc` is admissible: `main` runs everything on one spawned big-stack thread and only
  `anyhow::Result<()>` crosses the join; `src/jq` has no `Send`/`Sync` bounds; `jq_runner` never
  converts an `EvalError` (which carries an `OwnedValue` payload) into `anyhow::Error`. `Arc` buys
[...]

Issue #2905, opened 2026-09-13
ci: perf-guard's path-mode matrix has no row for |= (compound update assignment)


newhoggy:
**Severity: Low**

## Summary

#2655/PR #2904 added path-mode rows to `scripts/perf-guard.py`'s instruction-count matrix, closing the "no row exercises `resolve_node`/`resolve_node_sink` at all" gap for `del()`, `=`, and `path()`. `|=` (compound update assignment) is not covered by any row, in either mode.

## Why this needs its own row, not just adding `|=` to an existing filter

A naive `|=` row on a plain field target (e.g. `.users[] | .score |= . + 1`) would **not** actually close this gap: per `needs_path_prepass` (`src/jq/eval.rs`), a plain `Identity`/`Field`/`Index`/`Slice`/`Iterate` chain takes the single-path walkers (`walk_path`/`set_path`/`set_path_steps`/`update_path`/`delete_at_path`), not `resolve_node`/`resolve_node_sink` — this is exactly what PR #2904's own review discovered for `users_assign_scores`/`users_path_walk` (verified by direct instrumentation of `resolve_node_sink`: zero calls for either row). A `|=` row needs the same `select`/`as`-bound shape `users_del_bound_select` uses (matching #2042's own repro) to actually reach `resolve_node_sink` rather than re-covering the already-covered single-path walkers.

## Suggested fix direction

Add a row along the lines of `.users[] | select(.score < 100) | .score |= . + 1` (or an `as`-bound equivalent) to `QUERIES` in `scripts/perf-guard.py`, verified live against jq 1.7.1 for output identity first, then regenerate both arch baselines (`terminus` for x86_64, a native linux/arm64 Docker container on Apple Silicon for ARM64-Linux, matching #2655/#2904's own procedure) and confirm with a negative test (a temporary slowdown in `resolve_node_sink`) that the new row actually fails while a plain-target row does not — the same discipline #2904 used for its own rows.

## Why this is separate from #2655/#2904

#2655's own scope was "no row exercises the resolver at all" — that's closed. This is a narrower residual within the fix: `|=` specifically remains unexercised, discovered during #2904's own code review rather than being part of the original issue's premise.

## Dependencies / related

#2655 (closed by PR #2904), #2042 (the incident whose shape `users_del_bound_select`/this issue's suggested row both reproduce).

PR #3053, opened 2026-09-16
fix(jq): stream |= and op= writes so a failing update stops its path generator

### LINKED ISSUE #2974: jq: `|=` does not stop its path generator when the update's write fails partway (#2267's scope note is wrong for this shape)
**Severity: Low**

## Summary

#2267 fixed `=` so an assignment applies each write as its path resolves and stops the
path generator on the first write that fails validation. Its own **Scope** section says
the gap is specific to `=`:

> Confirmed live that `del()`/`|=` (which have no equivalent "does the write even
> type-check" short-circuit) match jq's own per-pair firing exactly [...] so this gap is
> specific to plain `=`'s [...] interaction with a downstream write-validation failure

That is **not true for `|=`**. A slice update whose write is rejected has exactly the
equivalent short-circuit, and `|=` does not reproduce it.

## Repro

Oracle `/usr/bin/jq` 1.7.1, stderr only, `1>/dev/null`:

```console
$ echo '[10,20,30]' | jq            -c '(.|stderr)[(0,1):(2,3)] |= 99'
[10,20,30]

$ echo '[10,20,30]' | succinctly jq -c '(.|stderr)[(0,1):(2,3)] |= 99'
[10,20,30][10,20,30][10,20,30][10,20,30]
```

The target fires **4** times (once per `(s, t)` pair) where jq fires it **once**. Same
error message and exit code 5 either way — only the side-effect count diverges, which is
why it needs a `stderr`/`debug` target to see at all.

`del()` was re-checked at the same time and does match jq, so this is `|=` alone.

## Root cause

`eval_update`/`eval_update_impl` (`src/jq/eval.rs`) resolves through
`resolve_dynamic_indexes` — the collecting form — and then applies `update_path` per path,
which is the same eager batch-then-write shape `eval_assign` had before #2267. jq's
`_modify(paths; update)` is a `reduce`/`foreach` over `path(paths)` just as `_assign` is,
so the same short-circuit applies.

## Why this is separate from #2267

#2267's remaining work was scoped to `eval_assign` by its own triage plan, explicitly on
the premise quoted above, and the fix that landed
(https://github.com/rust-works/succinctly/pull/2972) is gated to jq-mode `=` with a single
RHS output. `eval_update` is a different entry point with its own filter-application
machinery (`update_path`, the yq `scalar_slice_noop` rules, `eval_update_multi`'s RHS
fork), so routing it through the streaming write is a separate change with its own oracle
matrix — not a widening of that PR.

Verified pre-existing: a binary built from that PR's merge-base (`2891884b0`) fires 4
times too, so it is not a regression from #2267's fix.

## Suggested fix direction

`eval_assign_streaming` is the shape to mirror, including its three gating conditions
(jq mode; a single RHS output; `assignment_path_needs_streaming`). The extra piece `|=`
has and `=` does not is that its write is a *filter application* rather than a value
splice, so the per-path sink has to carry `update_path`'s filter — which
`fork_rhs_over_paths`'s `prepare`/`build_filter` split already models for the eager route.

Capture `+=`/`-=`/`//=` and the yq-mode spellings against the pinned binaries before
deciding how wide the fix goes; the compound forms route through `eval_update_multi` and
may not share the short-circuit.

## Related

- #2267 — the `=` half, and the issue whose scope note this corrects
- PR #2972 — where this was found, and which records it as a known-unchanged neighbour

newhoggy:
## Summary

Fixes #2974. jq lowers `|=` to `_modify(paths; f)` and `op=`/`//=` to `$v as $tmp | _modify(paths; . op $tmp)`: a `reduce` over `path(paths)` that updates one path before asking for the next. succinctly resolved every path first and then wrote them all. #2267 already fixed this for `=`; this PR applies the same fix to the rest of the family. Captured against `/usr/bin/jq` 1.7.1:

| input | filter | jq 1.7.1 | before | after |
|---|---|---|---|---|
| `[10,20,30]` | `(.\|stderr)[(0,1):(2,3)] \|= 99` (issue) | target fires 1× | 4× | 1× |
| `[1,2]` | `(.\|stderr)[(0,1)] \|= error("x")` | 1× | 2× | 1× |
| `[1,2]` | `(.\|stderr)[(0,1)] += "x"` (also `-=` `/=` `%=` `//=`) | 1× | 2× | 1× |
| `[1,2]` | `try ((.\|stderr)[(0,1)] \|= error("x")) catch "c"`, `…?`, `halt_error`, `break`, nested `\|=` | 1× | 2× | 1× |
| `[0,0]` | `.[(0,1)\|debug("p")] \|= debug("f")` | `p f p f` | `p p f f` | `p f p f` |
| `[0,0]` + `0 1 7 8` | `.[(input,input)] \|= input` | `[1,0,null,null,null,null,null,8]` | `[7,8]` | jq's |
| `[{"a":1},"s"]` | `(.[] \| select(.a > 0)) \|= error("x")` | error `x` | `Cannot index string with string "a"` | `x` |
| `[10,20,30]` | `(. as $x ?// $y \| (.\|stderr)[(0,1)]) \|= ("h"\|halt_error(3))` | 1× (a halt never retries) | 2× | 1× |

The last two rows need no `stderr`, `debug` or `input` to observe: an ordinary `select`-shaped update raised the wrong error.

### How

- **One loop, not three.** #2267's streaming `=` loop becomes `stream_path_writes`: resolve against a pristine copy, write each path into `result` as it resolves, and stop at the first write that raises. `=`, `|=` and single-RHS `op=`/`//=` all go through it. `eval_assign_streaming` keeps the `=`/`op=` result shaping; `|=` shapes its own.
- **The parked escape is classified** (new, and not needed while only `=` used the loop): it goes through `stop_with_escape`/`resume_from_escape`, so a `?//` inside the path retries past an error or `break` but never past a halt. `|=`'s filter can raise all three.
- **Gate:** `=`'s own gate (jq mode, `needs_path_prepass`, `assignment_path_needs_streaming`, and a single RHS output for `op=`). `|=` adds two exclusions: the yq-only `sort_keys` skip-absent pre-filter, and a filter that reads `parent`. `parent` is a succinctly extension with no jq ordering, and it needs every target vivified before any filter runs. yq mode is untouched.
- **Re-entrancy:** the filter now runs inside the resolver's sink. The only resolver-held state an evaluated filter can reach is the `?//` retry side channels, and those are consulted only at a retry decision, after the sink returns (documented on `stream_path_writes`). The `?//`-inside-the-filter row is pinned.

### Cost (the plan asked for this to be measured, not estimated)

Release `cgu1+fat`, 7 interleaved repetitions, outputs identical, baseline = merge-base `06ba91257`:

| machine | 10 MB `(.users[] \| select(.age > 30)).score \|= . + 1` | 10 MB `.users[(0,1)].score \|= 0` | static control `.users[].score \|= . + 1` |
|---|---|---|---|
| Apple M4 Pro | 0.31 s / 254 MB → 0.32 s / 316 MB | 0.37 s / 255 MB → 0.38 s / 311 MB | unchanged (0.27 s / 199 MB) |
| AMD Ryzen 9 7950X | 0.51 s / 237 MB → 0.58 s / 306 MB | 0.59 s / 254 MB → 0.66 s / 322 MB | unchanged (0.43 s / 169 MB) |

That's +22–29% peak RSS and +3–7% (M4 Pro) / +12–14% (7950X) wall time on the streaming shapes. It's the second document, the same cost #2267 accepted for `=`. Under ADR-0018's decision order, performance only breaks a tie between faithful options, and the eager route wasn't faithful here. #2999's structural sharing retires both documents' copies. The full table (1 MB rows and `+=`) is in `limitations.md`. #2905 (perf-guard has no `|=` row) hasn't landed, so perf-guard won't see this shift.

### Holdouts and scope

- Multi-output RHS `op=` keeps the eager fork. That's #2267's still-open re-resolution half: `(.|stderr)[(0,1)] += (1,"x")` fires 2× where jq fires 3×. Pinned.
- `|= empty` on several paths still deletes immediately (#3030), and it's unchanged on both routes.
- The path-mode `?//` refusal (`[path(. as [$x] ?// $x | .[0])]`) is the documented artefact guard. Unchanged.
- jq's `Paths must be specified as an array` accumulator corruption after a `?//` retry is not reproduced, and is recorded in `limitations.md`.

### Review round 1 (`/code-review high`)

One finding, fixed with mutation-checked tests. `stream_path_writes` never cleared a write escape that an earlier `?//` alternative had parked, so a retry whose writes all succeeded still raised it. This was pre-existing in `=`'s own loop, and this PR shares that loop with `|=`/`op=`. jq 1.7.1 differs by operator, so the fix is an explicit `RetryResumes`:

- `=`: `_assign`'s `reduce` carries on, so `(. as $x ?// $y | .[if $x == null then 0 else -5 end]) = 1` on `[1]` is now `[1]`, as in jq. It used to raise `Out of bounds negative array index`.
- `|=` / `op=`: jq fails on any retry through `_modify` (`Paths must be specified as an array`). succinctly keeps the earlier write's error, so the exit code matches while the message differs. Recorded in `limitations.md`.

## Test plan

- [x] `update_streams_its_path_generator_2974`: 30 rows, each with stdout, stderr and exit code captured whole from jq 1.7.1. Covers failure rows F1–F13, success-path interleave, the `select` error rows, the `?//` retry rows (error, `.[] as`, halt, break, filter-internal `?//`) and the must-not-change rows.
- [x] `update_streaming_interleaves_input_reads_2974`: `input` order, and no over-consumption after a caught failure
- [x] `update_streaming_retry_after_a_failed_write_2974`: `=` retry succeeds (jq's `[1]`), and `|=`/`+=` keep exit 5
- [x] `update_streaming_holdouts_2974`: multi-RHS `op=` (2 fires, not jq's 3); `key`/`parent` extension outputs unchanged
- [x] Oracle matrix of 42 rows run against jq and the baseline binary. The only differences are the three scoped holdouts above, all identical to baseline.
- [x] A/B memory/time on M4 Pro and 7950X (above)
- [x] `cargo test --features cli,simd,regex,serde` (includes the yq CLI/golden suites, unchanged)


newhoggy:
Automated review ran at `/code-review high` (1 finding, fixed with tests that fail in both directions). This moves every jq-mode `|=`/`op=` with a computed or multi-path target onto a streaming route, at a measured +22–29% peak RSS. A human may want the deeper `/code-review ultra 3053` pass. It isn't blocking the merge.

Issue #3055, opened 2026-09-16
jq: strftime leaves %c %x %X %r %k %l literal and prints %z/%Z as UTC where jq uses the local zone


newhoggy:
**Severity: Low**

## Summary

`strftime`/`strflocaltime`'s formatter (`format_strftime`, `src/jq/eval.rs`) leaves several specifiers jq supports as literal text, and prints `%z`/`%Z` as `+0000`/`UTC` for `strftime` where jq names the local zone.

## Repro

`TZ=UTC`, `1700000000 | strftime("%X")`-style probes, jq 1.7.1 vs `succinctly` (#3046 branch):

| specifier | jq 1.7.1 | succinctly |
|---|---|---|
| `%c` | `Tue Nov 14 22:13:20 2023` | `%c` |
| `%x` | `11/14/23` | `%x` |
| `%X` | `22:13:20` | `%X` |
| `%r` | `10:13:20 PM` | `%r` |
| `%k` | `22` | `%k` |
| `%l` | `10` | `%l` |
| `%P` | `P` | `pm` |
| `%E` | `E` | `%E` |
| `%O` | `O` | `%O` |

And under `TZ=EST5EDT`, `0 | strftime("%z %Z")` is `"-0500 EST"` in jq and `"+0000 UTC"` here (jq's `strftime` formats the broken-down UTC time but takes `%z`/`%Z` from the local zone).

## Root cause

`format_strftime`'s specifier match has no arms for these. Its fallback pushes the unrecognised `%X` through literally, and `strftime` passes a fixed UTC zone to it. Note that `%P`/`%E`/`%O` are macOS libc behaviour (`P`, `E`, `O` printed bare), so a Linux jq may answer differently; the pinned oracle is macOS.

## Why this is separate from #3046

Found while adding `strflocaltime` (#3046), which shares the formatter. None of these specifiers is specific to the new builtin.

## Suggested fix direction

Add the missing arms (`%c` = `%a %b %e %H:%M:%S %Y`, `%x` = `%m/%d/%y`, `%X` = `%H:%M:%S`, `%r` = `%I:%M:%S %p`, `%k` space-padded hour, `%l` space-padded 12-hour), and decide `%P`/`%E`/`%O` against the pinned macOS oracle. Take `%z`/`%Z` for `strftime` from the local zone (see the IANA-zone follow-up for where that comes from). Pin every specifier against jq 1.7.1 in a table-driven test.

newhoggy:
[...]

Issue #3054, opened 2026-09-16
jq: localtime, mktime and strflocaltime ignore IANA zone names and the system zone, using UTC


newhoggy:
**Severity: Medium** (wrong output for any user whose zone is not a POSIX `TZ` offset string, which includes the default: an unset `TZ`)

## Summary

`localtime`, and `strflocaltime` since #3046 because it reuses `localtime`'s zone, only understand a POSIX offset string in `TZ` (`EST5EDT`, `UTC-9`). With an IANA zone name (`TZ=Asia/Tokyo`, the usual form) or no `TZ` at all (the system zone), they silently use UTC. `mktime` of a local time is off by the same amount.

## Repro

jq 1.7.1 vs `succinctly` (#3046 branch), filter `0 | [localtime, strflocaltime("%H %Z"), (localtime|mktime)]`:

| environment | jq 1.7.1 | succinctly |
|---|---|---|
| `TZ=Asia/Tokyo` | `[[1970,0,1,9,0,0,4,0],"09 JST",32400]` | `[[1970,0,1,0,0,0,4,0],"00 UTC",0]` |
| `TZ=America/New_York` | `[[1969,11,31,19,0,0,3,364],"19 EST",-18000]` | `[[1970,0,1,0,0,0,4,0],"00 UTC",0]` |
| `TZ` unset (this machine: AEST) | `localtime[3]` = `10`, `%Z` = `"AEST"` | `0`, `"UTC"` |

## Root cause

`src/jq/eval.rs`: `builtin_localtime` gets its offset from `estimate_local_offset`, which only parses `TZ` through `parse_simple_tz_offset` and returns `0` otherwise. Its own comments say "a proper implementation would use platform-specific APIs or chrono crate". It also takes the offset *now* rather than at the timestamp, so DST is wrong for any date whose offset differs from today's. `local_zone` (added by #3046 for `strflocaltime`) deliberately shares that model, so the two stay consistent.

## Why this is separate from #3046

#3046 added `strflocaltime` on top of the existing local-zone model rather than change it. Resolving IANA zones needs the system zone database. The crate denies `unsafe_code`, which rules out `localtime_r` FFI, so this is either a new dependency (e.g. a tz-database crate) or a `libc`/`unsafe` policy decision. Either needs its own review under the dependency policy.

## Suggested fix direction

[...]

PR #3056, opened 2026-09-16
feat(jq): implement the nine missing non-math jq builtins

### LINKED ISSUE #3046: jq: 9 non-math jq builtins are unimplemented (JOIN/2..4, format/1, input_filename, get_search_list, get_jq_origin, get_prog_origin, strflocaltime)
**Severity: Low**

Found while triaging #2937 / #3042. Calling every name in `tests/data/jq-builtin-names.txt` (the pinned jq 1.7.1 roster) on main @ `b509a8019` finds 45 builtins succinctly does not implement: they compile (via `JQ_BUILTIN_ROSTER`) and fail when reached with `undefined function: <name>/<arity>`, exit 5. #3042 covers the 36 libm functions. This issue covers the other **9**:

`JOIN/2`, `JOIN/3`, `JOIN/4`, `format/1`, `input_filename/0`, `get_search_list/0`, `get_jq_origin/0`, `get_prog_origin/0`, `strflocaltime/1`

yq v4.53.3's lexer rejects all of them (`lexer: invalid input text "input_filename"`, …), so in yq mode they belong behind `--jq-extensions`.

### Oracle capture (`/usr/bin/jq` 1.7.1-apple)

**`JOIN`** (jq-defined in `builtin.jq`: `def JOIN($idx; idx_expr): [.[] | [., $idx[idx_expr]]];`, `def JOIN($idx; stream; idx_expr): stream | [., $idx[idx_expr]];`, `def JOIN($idx; stream; idx_expr; join_expr): stream | [., $idx[idx_expr]] | join_expr;`):

```console
[{"id":"1","n":"a"},{"id":"2","n":"b"}] | INDEX(.id) as $idx | [["1","x"],["3","y"]] | JOIN($idx; .[0])
  => [[["1","x"],{"id":"1","n":"a"}],[["3","y"],null]]
… | [JOIN($idx; (["1","x"],["3","y"]); .[0])]         => same
… | [JOIN($idx; (["1","x"],["3","y"]); .[0]; add)]    => error: array (["1","x"]) and object ({"id":"1","...) cannot be added
{"1":{}} as $idx | [[1]] | JOIN($idx; .[0])           => error: Cannot index object with number
```

**`format/1`** (a dynamic `@name`):

```console
[1,"a b",null] | [format("csv"), format("tsv"), format("text"), format("json"), format("sh"), format("base64"), format("uri"), format("html")]
  => ["1,\"a b\",","1\ta b\t","[1,\"a b\",null]","[1,\"a b\",null]","1 'a b' null","WzEsImEgYiIsbnVsbF0=","%5B1%2C%22a%20b%22%2Cnull%5D","[1,&quot;a b&quot;,null]"]
"aGk=" | format("base64d")   => "hi"
[1] | [format("csv","json")] => ["1","[1]"]
[1] | format("@csv")         => error: @csv is not a valid format
"x" | format("nope")         => error: nope is not a valid format
"x" | format(1)              => error: number (1) is not a valid format
```

**`input_filename`:**

```console
jq -n 'input_filename'                                  => null
jq 'input_filename' in.json                             => "in.json"
jq 'input_filename' < in.json                           => "<stdin>"
jq 'input_filename' in.json b.json                      => "in.json", "b.json"
jq -n '[input_filename, (input|input_filename), input_filename]' in.json => [null,"in.json","in.json"]
```

**Path introspection:**

```console
jq -n 'get_search_list'            => ["~/.jq","$ORIGIN/../lib/jq","$ORIGIN/../lib"]   (unexpanded)
jq -L /tmp/x -n 'get_search_list'  => ["/tmp/x"]
jq -n 'get_jq_origin'              => "/usr/bin"     (directory of the jq binary)
(cd / && jq -n 'get_prog_origin')  => "/"
cd / && jq -n -f /some/dir/p.jq    => "/some/dir"    (directory of the -f program)
```

**`strflocaltime`** (`strftime` formatted in the local zone):

```console
TZ=Asia/Tokyo: 0 | strflocaltime("%Y-%m-%dT%H:%M:%S %Z")  => "1970-01-01T09:00:00 JST"
TZ=Asia/Tokyo: 0 | gmtime | strflocaltime("%H %Z")      => "00 JST"
TZ=Asia/Tokyo: 1.9 | strflocaltime("%S")                => "01"
TZ=UTC:        [0 | strflocaltime("%Y","%H")]           => ["1970","00"]
"x" | strflocaltime("%H")                               => error: strflocaltime/1 requires parsed datetime inputs
0 | strflocaltime(1)                                    => jq ABORTS: Assertion failed: (JVP_HAS_KIND(j, JV_KIND_STRING)), function jv_string_value, file jv.c, line 1370. (exit 134)
```

## Triage plan (2026-09-17)

The 9 names fall into four independent groups; each can land as its own PR.

1. **`JOIN/2..4`.** They are jq-defined, so express them as their `builtin.jq` definitions: desugar, or a jq-source prelude if one exists. Check how `INDEX`/`IN` are implemented (parser-level `Builtin::Index`/`UpperIndex`) and follow that precedent. They take `$idx` as a value parameter and use generator arguments, so pin fan-out order with a generator `stream`.
2. **`format/1`.** Evaluate the argument (a generator, one output per value), require a string naming a format *without* `@`, and dispatch to the existing `format_owned::<S>` (`src/jq/eval.rs`). Other values raise jq's two error wordings above. Check that `format("dsv")` and any succinctly-only format names are refused, since they are not jq formats. Watch the yq gate: `@yaml`/`@props` are yq formats, but `format/1` itself is lexer-rejected by yq.
3. **`input_filename`, `get_search_list`, `get_jq_origin`, `get_prog_origin`.** These need CLI context the evaluator does not have today: the current input's file name (`null` before any input is read with `-n`, `"<stdin>"` for stdin), the `-L` list or jq's unexpanded default, the directory of the running binary, and the directory of the `-f` file or else the current directory. Find how `input`/`inputs`/`input_line_number` already reach the evaluator (their shared queue is seeded in `src/bin/succinctly/jq_runner.rs`) and extend that channel rather than adding a global.
   - **`get_jq_origin`:** decide whether it is the directory of the invoked path or of the resolved symlink target (`sjq` → `succinctly`), capturing jq's behaviour through a symlink to decide.
   - **`get_search_list`:** echo jq's literal defaults, even though succinctly's `ModuleLoader` also reads `JQ_LIBRARY_PATH`; that difference is recorded, not changed here.
4. **`strflocaltime/1`.** Reuse `builtin_strftime`'s formatting with `builtin_localtime`'s zone conversion; the input is a number or a broken-down time array. **Divergence required:** `0 | strflocaltime(1)` aborts the pinned jq (exit 134, an assertion failure), so ADR-0018's "would take the host process down" condition applies. Raise a catchable error instead, and record it in `docs/compliance/jq/limitations.md`. Capture jq's wording for a non-string format *in a context that does not abort* (e.g. `strftime(1)`) and reuse it.

**Tests.** Add jq golden cases for every capture row that is not environment-dependent. Use CLI tests with a controlled `TZ`, cwd and temp files for `input_filename`, the `get_*` builtins and `strflocaltime`. Update `test_unimplemented_jq_builtins_still_compile_when_unreached_1473` and `…_still_errors_when_reached_1473` (`tests/jq_cli_tests.rs`), which sample `JOIN`, `format`, `input_filename`, `get_search_list` and `strflocaltime` as unimplemented; once #3042 lands too, the "unimplemented" set may be empty, so the tests may need a different premise. Add yq tests: rejected without `--jq-extensions`.

**Docs.** Shrink the "~45 jq builtins succinctly does not implement" sentence in `docs/compliance/jq/limitations.md`, update `JQ_BUILTIN_ROSTER`'s doc comment (`src/jq/resolve.rs`), and add the names to `docs/reference/yq-language.md`'s gated list.

**Noticed, not in scope:** succinctly's `strftime("%Z")` prints `UTC` under `TZ=Asia/Tokyo` (`0 | strftime("%H %Z")` → `"00 UTC"`), where jq prints the local abbreviation (`"00 JST"`). It is not recorded in `limitations.md`, and `strflocaltime` would share the `%Z` code.

**Related:** #3042 (the libm half of the same roster gap), #1887 (`--jq-extensions` allowlist).

### Class: **Opus**

Groups 1–2 alone would be Sonnet work. Group 3 needs a new CLI-to-evaluator context channel, with decisions about symlinks and stdin naming, and group 4 needs an ADR-0018 divergence record for a reference crash. If you would rather split them, JOIN + `format/1` can go out as a separate Sonnet issue.

newhoggy:
## Summary

Fixes #3046. Implements the nine non-math jq 1.7.1 builtins succinctly didn't have. They used to compile (via the roster) and then fail with `undefined function` when reached. All nine are gated behind `--jq-extensions` in `succinctly yq`, since yq v4.53.3's lexer rejects every one of them.

| builtin | how |
|---|---|
| `JOIN/2`, `JOIN/3`, `JOIN/4` | **Desugared at parse time** to jq's own `builtin.jq` definitions: `$idx` becomes an `as` binding under a variable name containing a space, so no program can capture it. No evaluator arm is needed. `resolve.rs` can read the desugaring back into arguments, so a user `def JOIN(..)` still shadows it at every arity. |
| `format/1` | New `Builtin::FormatNamed`. It fans out over the argument and dispatches to the existing `@format` code for jq's nine names (`text json csv tsv html uri sh base64 base64d`). It refuses succinctly-only formats (`urid`, `dsv`, `yaml`, `props`) and non-strings with jq's two wordings (`nope is not a valid format`, `object ({"aaaaaaaaa...) is not a valid format`). |
| `strflocaltime/1` | `strftime`'s body, shared (`strftime_in_zone`), with a number converted into the local zone and `%z`/`%Z` naming it. |
| `input_filename`, `get_search_list`, `get_jq_origin`, `get_prog_origin` | New zero-argument builtins that read `jq::cli_context`, which the CLI seeds alongside the input queue it already seeds. |

### Behaviour pinned from jq 1.7.1 (not guessed)

- `input_filename`:
  - `"<stdin>"` for stdin, and the path as typed (`./in.json`).
  - The last file under `-s`; `-R` works too.
  - `null` under `-n` until `input` reads. It follows the input queue's own position: `-n '[input_filename, (input|input_filename)]'` is `[null,"in.json"]`, and after `[inputs]` it is the last file.
- `get_jq_origin` is the directory of `argv[0]` **as typed, not resolved**. A bare name found on `PATH` gives `"."`, and a symlink gives the symlink's own directory. (Verified via symlinks and `PATH` lookups for both binaries.)
- `get_prog_origin` is the resolved directory of the `-f` file (`./sub/../sub/p.jq` → `…/sub`), else the working directory.
- `get_search_list` is the `-L` list, with each existing directory resolved to its real path and a missing one kept as given. Without `-L` it is jq's literal `["~/.jq","$ORIGIN/../lib/jq","$ORIGIN/../lib"]`. `JQ_LIBRARY_PATH` is ignored, as in jq.

### Divergences and gaps (recorded in `limitations.md`)

- **ADR-0018 rule 4:** `0 | strflocaltime(1)` (and the existing `strftime(1)`) aborts jq 1.7.1 on an assertion (exit 134). succinctly raises a catchable error (exit 5).
- **Found, not fixed here (follow-ups filed):**
  - #3054 (Medium): `localtime`, `mktime` and now `strflocaltime` only understand POSIX `TZ` offset strings. An IANA name (`TZ=Asia/Tokyo`) or an unset `TZ` silently uses UTC. `strflocaltime` shares `localtime`'s zone model deliberately, so the two agree; fixing it needs the zone database, meaning a dependency or `unsafe` policy decision.
  - #3055: the shared `strftime` formatter leaves `%c %x %X %r %k %l` literal, and `strftime`'s own `%z`/`%Z` are fixed to UTC.

### Review round 1 (`/code-review high`), fixed in the last commit

Four findings, each now pinned by a test row:

1. `input_filename` was `null` for stdin on the materializing routes (`[input, input_filename]`, `-R`, `-n input`). It is now `"<stdin>"`.
2. `strflocaltime("%s")` was off by the zone offset. It is now the epoch of the local fields: `TZ=EST5EDT`, `0|strflocaltime("%s")` is `"0"`, and `[1970,0,1,0,0,0,4,0]|…` is `"18000"`.
3. `get_search_list` didn't resolve existing `-L` directories (`-L lib` → `["/abs/lib"]` in jq).
4. `input_filename` was `null` on `--input-dsv`'s streaming route.

Two pre-existing findings are routed to follow-ups rather than fixed here:
- DST under a POSIX `TZ` (`EST5EDT` in summer) → added to #3054.
- `strftime("%s")` treating its fields as UTC where jq treats them as local → added to #3055.

Postfix after a special form (`JOIN(...)[0]`, like the existing `limit(...)[0]`) is a pre-existing parser class.

## Test plan

- [x] `join_and_format_match_jq_3046`: 18 rows, each with stdout, stderr and exit code captured whole from jq 1.7.1. Covers all three `JOIN` arities, generator `stream`, errors, shadowing by a user `def JOIN` at 2 and 4 args, every format name, the error wordings, a generator argument, and shadowing by `def format`.
- [x] `strflocaltime_matches_jq_3046`: 27 rows under `TZ` = `UTC`, `EST5EDT`, `UTC-9` and `PST8`, captured from jq (including `%z %Z`, gmtime-array input, fractional and negative seconds, and errors), plus the non-string-format divergence.
- [x] `cli_context_builtins_match_jq_3046`: 14 `input_filename`/`get_*` invocations in a temp directory, plus `get_jq_origin`
- [x] `jq_only_builtins_3046_are_gated_in_yq_mode` (yq CLI)
- [x] Wider oracle matrix run by hand: 25 `JOIN`/`format` rows, 40 `strflocaltime` rows and 21 CLI-context rows, all matching jq. The only differences are #3055's specifiers and `strftime`'s own `%z`/`%Z`.
- [x] `test_unimplemented_jq_builtins_still_compile_when_unreached_1473`'s sample moved to still-unimplemented libm names


newhoggy:
Automated review ran at `/code-review high` (4 findings, all fixed and pinned by test rows; 2 pre-existing gaps routed to #3054/#3055). This adds a public `jq::cli_context` seam and four new `Builtin` variants, so a human may want the deeper `/code-review ultra 3056` pass. It isn't blocking the merge.

Issue #2937, opened 2026-09-14
jq: math builtins widen large int literals with a bare cast (`as_f64`, `get_float_value`), and floor/ceil/round/trunc print exact ints past 2^53


newhoggy:
**Severity: Low**

Two related gaps #2906 left open (it fixed `+ - * / %` and comparisons only):

1. The math builtins reach their operand through `OwnedValue::as_f64` /
   `get_float_value` (`src/jq/eval.rs`), which widen an `Int` with `n as f64` (or parse the
   raw document text) instead of `jq_literal_int_to_f64`:

   ```console
   $ jq  -n '869389897822472004 | sqrt'   # 932410798.8555645
   $ sjq -n '869389897822472004 | sqrt'   # 932410798.8555647
   ```

2. `floor`/`ceil`/`round`/`trunc` return `OwnedValue::Int(floor_f64(n) as i64)`, so past
   `2^53` the result prints its exact integer digits where jq prints the double:

   ```console
   $ jq  -n '869389897822472004 | floor'  # 869389897822472000
   $ sjq -n '869389897822472004 | floor'  # 869389897822472064
   ```

   The value is fine once (1) is fixed (an `Int` that came from a double is exactly
   representable, and #2906's rounding is provably the identity on such values); the
   display/type is the gap — `jq_f64_backed_int` (`src/jq/eval.rs`, added by #2906) is the
   constructor to return here.

`as_f64` has ~24 call sites across both evaluators and is mode-less, so the fix needs
either an `S`-generic sibling or a per-site audit of which callers are jq-only.

newhoggy:
## Triage plan (2026-09-17)

### Premise check (main @ `b509a8019`, `/usr/bin/jq` 1.7.1-apple, yq v4.53.3)

Both claims reproduce. The gap is also wider than the body says, and two separate precision bugs sit in the same functions (`sqrt_f64`, `round_f64`). A fix limited to the operand widening would not make the issue's own `sqrt` repro pass. All rows below come from `jq -nc` vs `succinctly jq -nc` at this commit:

| filter                                               | jq 1.7.1                  | sjq (main)                 | cause                         |
|------------------------------------------------------|---------------------------|----------------------------|-------------------------------|
| `869389897822472004 \| sqrt`                         | `932410798.8555645`       | `932410798.8555647`        | widening **and** Newton sqrt  |
[...]

PR #3063, opened 2026-09-16
fix(jq): close #2642's fabrication on the routes that never cross a funnel

### LINKED ISSUE #2642: jq: a root-bound `$x` matches a rebuilt equal-valued copy — TrackedVar's value witness writes where jq refuses

### LINKED ISSUE #3036: jq: #2642 fabrication still live when the bind and the rebuild both run inside eval.rs (input builtins, |= RHS, fold UPDATE)
**Severity: High** (silent data modification: exit 0 with a written result where jq refuses)

## Summary

#2642 closed a root-bound `$x` certifying against a rebuilt equal-valued copy, but only where an
expression crosses a generic→`eval.rs` funnel. When the binding **and** the rebuild both happen
inside `eval.rs`'s owned-value evaluator, no funnel runs, and the original bug is fully live.
Four routes reach it from ordinary jq programs, and one more from the `parent`/`key` extensions.

#2642's plan listed this as the "in-bridge residual" / "owned-value identity" item "to file at
stage 4". It was never filed. It was split out of #2889's triage
(https://github.com/rust-works/succinctly/issues/2889#issuecomment-5699237783), which has the
full analysis.

## Repro

`/usr/bin/jq` 1.7.1 vs `main` `b509a8019`, input `{"a":1}`. Every succinctly row exits **0**;
every jq row exits **5** with `Invalid path expression`:

| flags | filter                                                          | succinctly          |
|-------|-----------------------------------------------------------------|---------------------|
| `-n`  | `input \| . as $x \| {a:1} \| ($x.a = 9)`                       | `{"a":9}`           |
| `-n`  | `[inputs] \| .[0] \| . as $x \| {a:1} \| ($x.a = 9)`            | `{"a":9}`           |
|       | `. as $x \| {a:1} \| (input_line_number, ($x.a = 9))`           | `0` `{"a":9}`       |
|       | `. as $x \| {a:1} \| (input_line_number, del($x.a))`            | `0` `{}`            |
|       | `. as $x \| {a:1} \| (input_line_number, path($x))`             | `0` `[]`            |
|       | `. as $x \| {a:1} \| input_line_number as $n \| ($x.a = 9)`     | `{"a":9}`           |
|       | `reduce (1) as $i (.; . as $x \| {a:1} \| ($x.a = 9))`          | `{"a":9}`           |
|       | `[.] \| .[] \|= (. as $x \| {a:1} \| ($x.a = 9))`               | `[{"a":9}]`         |
|       | `with_entries(.value \|= (. as $x \| 1 \| path($x)))`           | `{"a":[]}`          |
|       | `. as $x \| {a:1} \| (parent, ($x.a = 9))` (extension; jq exit 3) | `{"a":9}`         |

Controls that already refuse correctly: the bare `. as $x | {a:1} | ($x.a = 9)`, and the same body
inside `[...]`, `first(...)`, `limit`, `label`, `def`, `try`, `{r: ...}` or `map(...)`.

Must stay accepted (they match jq today): `-n 'input | . as $x | ($x.a) = 9'` → `{"a":9}`,
`… | $x = 5` → `5`, `… | del($x.a)` → `{}`, `… | first(.) | path($x)` → `[]`.

## Root cause

- `eval.rs`'s `eval_as`/`each_as`, and the fold sites' `substitute_var_tracked`, mint an
  `Origin::Snapshot` marker with `node: None`. `Frame::certifies` admits `Snapshot`
  unconditionally, and `resolve_node` then compares by value. `demote_rebuilt_markers` only runs
  at the generic evaluator's funnels (`eval_generic.rs`), so a marker minted *after* the last
  funnel is never checked.
- Routes into that position:
  - the whole-program input-queue bridge (`takes_input_queue_bridge`, #1504): any program that
    mentions `input`/`inputs`/`input_line_number` runs entirely in `eval.rs`;
  - a bind inside a fold's UPDATE (the funnel demotes `update` *before* its own `as` runs);
  - a bind inside `|=`'s RHS;
  - the owned-identity pipe (`eval_owned_identity_stages`, `owned_identity_values`,
    `try_path_context_absent_sink`), which never calls `demote_rebuilt_markers`.
    `test_tracked_var_owned_identity_sibling_gap_unaffected_2642` pins that route as unfixed.

## Why this is separate from #2889

#2889 is the *accepting* direction (pass-through constructions jq keeps as the same node now
refuse). It needs a representation decision for owned-value identity (Fable, tied to ADR-0024 /
#2999). This issue is the *fabricating* direction. It can be closed refuse-only by extending
#2642's proven witness rule, without that decision.

## Plan (from #2889's triage §3, Phases 0-1)

1. **Sweep/fuzz first.** Add an `in-evaluator-*` block to `scripts/jq-bind-origin-oracle-sweep.sh`
   (every row above plus controls). Add a ROUTE wrapper family to `scripts/jq-bind-origin-fuzz.py`
   (`-n input |`, an `input_line_number,` sibling, `input_line_number as $n |`, `.[] |= (…)`,
   `reduce`/`foreach` UPDATE, `with_entries(.value |= …)`, a `parent`/`key` sibling).
   Gate: reproduce exactly these rows on today's binary. Validate the harness's power against a
   build with `demote_rebuilt_markers` stubbed out.
2. **Witness `eval.rs`-minted markers.** When the bind source is an identity passthrough, record
   the source's own `StandardJson` identity in `Tracked::node` (`BindOrigin::Node`):
   - containers: index token + BP position;
   - `String`/`Number`: backing text + `start`;
   - `Null`/`Bool`: exempt, as today.
3. **Demote at `eval.rs`'s own resolver entries** (`eval_assign`, `eval_update`,
   `eval_compound_assign`, `eval_alternative_assign`, `builtin_del`, `builtin_path_on_owned`,
   every other `resolve_node` entry), with `RootWitness` taken from their own input. A rebuilt
   value is re-indexed, so it gets a fresh token.
4. **Re-stamp across the bridge (the hazard).** A generic-bound marker that passed its funnel
   arrives in `eval.rs` with the old document's token, so step 3 would demote it and break
   today's accepted writes. The funnel should hand its proven-root markers to the bridge, which
   re-stamps them to the new index's root after `JsonIndex::build`. Rejected: "a foreign token is
   already certified", which reopens the residual for a generic-bound marker followed by an
   in-`eval.rs` rebuild.
5. **Owned-identity route.** Add an `exact` bit to `OwnedIdentity` (set by `kept(cursor)` from a
   true passthrough, cleared by every other rule). The witness is `RootWitness::of(id.base)` iff
   `exact && ancestors.is_empty()`, else `Owned`. **Do not** key on `ancestors.is_empty()` alone,
   as #2890's review suggested: `OwnedIdentity` models yq *position*, and its `Keeps` rule covers
   rebuilding builtins (`sort`, `tostring`, `to_entries`, `[...]`, writes), so it would certify
   `sort`'s new array. `.foo | . as $x | (parent, ($x.a = 9))` must stay `{"a":9}`.
6. **Expected agree→refuse flips:** pass-through embeds on the `eval.rs` route
   (`input | . as $x | {k:.} | .k | path($x)`, `… | . + {} | …`, today accepted by value). List
   them in the PR as refuse-only, owned by #2889.

**Tests:** add `test_tracked_var_rebuilt_root_in_evaluator_routes_refuse_*` (every repro row:
exit 5 + message class) and a must-stay-accepted twin for the rows above. The existing #2642 tests
stay unchanged. Run the yq golden suite and `test_owned_identity_rules_match_yq_2416`.

**Docs:** `docs/compliance/jq/limitations.md` (the #2642 paragraph calls the fabrication closed),
the `Origin::Snapshot` doc comment in `src/jq/expr.rs`, and `docs/plan/jq-bind-origin-frame.md`
"What remains".

**Perf:** a `demote_rebuilt_markers` precheck per `eval.rs` resolver entry, plus witness recording
at binds. Run an interleaved A/B on M4 Pro and 7950X: the no-`as` holdout (`.[] |= f`,
`del(.[] | select(..))`) must read 0.00%, plus #2657's `. as $r` write shapes and the
input-queue route at ≥1 MB. perf-guard cannot see marker shapes (#2655).

**Risks:** step 4 is where a mistake regresses every bridged write (the must-stay-accepted rows
are the tripwire). Unmasking: sweep shapes expected *not* to change against both jq and a `main`
build. A scalar witness through `(text, start)` must survive the owned round trip.

**Class: Opus.** It mirrors #2642's refuse-only witness rule onto `eval.rs`'s own entries and the
owned-identity route, touching multiple sites across both evaluators with a two-architecture A/B.
The one judgment call (the re-stamp) is settled above, and a wrong choice fails toward refusing,
not fabricating.

**Related:** #2642, #2889, #2072, #1504, #2416 (step 3), #2655, #2657.

newhoggy:
## Summary

Fixes #3036.

#2642 checked a root-bound `$x` against a rebuilt copy only where an expression crosses from the generic evaluator into `eval.rs`. When the bind **and** the rebuild both run inside `eval.rs` — the input-queue bridge (any program mentioning `input`/`inputs`/`input_line_number`), a fold's UPDATE, a `|=` right-hand side, `with_entries`, a `catch` handler, the owned-surface consumers' arguments — no funnel ran, and every one of the issue's ten rows exited 0 with a written result where jq exits 5.

**Mechanism — the re-entry, not node identity.** The triage plan proposed witnessing `eval.rs`-minted markers with a node id and demoting at each resolver entry. That cannot work from a `StandardJson`: an empty container (`JsonFields`/`JsonElements` hold only the *first child's* cursor) and a scalar (`JsonString`/`JsonNumber` hold text + offset, no index) have no node id to recover, so `-n 'input | . as $x | ($x.a) = 9'` on `{}` would have had to either regress or stay a hole. Instead, every place `eval.rs` re-indexes an owned value into a throwaway document (`eval_each_owned`, `eval_owned_input_reindexed`, `eval_owned_expr_full`, `eval_path_context_pipe_owned`, `builtin_with_entries`) demotes the expression's `Snapshot` markers first: a freshly rebuilt document cannot be a node any earlier binding was frozen from — the one fact that holds for every value shape. The generic funnels, which already prove their markers against a live cursor, take a non-demoting `eval_each_owned_bridged`, so their proof is kept. Refuse-only by construction (the only transition is `Snapshot → Untracked`).

The two classes the #2642 review deliberately left open close with it:
- **Owned identity route** (`eval_owned_identity_stages`/`owned_identity_values`): checked against `OwnedIdentity::root_witness` — the base node only while the value *is* that node's own unrebuilt value (new `exact` bit, cleared by every stage rule that places a new value; never `ancestors.is_empty()` alone, which would certify `sort`'s new array). `.foo | . as $x | (parent, ($x.a = 9))` still writes; `.foo | . as $x | sort | (parent, ($x[0] = 9))` now refuses.
- **`catch` handler**: checked against the payload's own node (`try_payload_root`). `error($x)`/`$x | error` raise the marker's node verbatim, as jq's `catch` then sees the same `jv`, so `. as $x | try error($x) catch path($x)` stays `[]`; a value-equal `try error({a:1}) catch ($x.a = 9)` no longer certifies.

**Refuse-only flips** (recorded in `docs/compliance/jq/limitations.md`, pinned as residual): on the input-queue route, `input | . as $x | S | ($x.a = 9)` for S in {`[.] | .[0]`, `{k:.} | .k`, `[., .] | .[0]`, `reduce empty as $i (.; .)`, `reduce 1 as $i (.; .)`, `foreach 1 as $i (.; .; .)`, `getpath([])`, `nth(0; .)`, `until(true; .)`, `recurse(empty)`, `ltrimstr("x")`, `debug`, `stderr`} — each returns its input as an owned copy on that route — plus `input | reduce (.) as $x (.; ($x.a = 9))` and `input | . as $x | try error($x) catch path($x)`. `first(.)`, `select`, `if`, `//`, `try .`, `label`, `limit`, `. as $y | .` keep their cursor and still answer. Also closed beyond the filed rows: `. as $x | (.b, .) |= (if type == "object" then ($x.a = 9) else . end)` wrote `{"a":9,"b":2}` on `main` where jq refuses (a second path's root is a `setpath` copy), and `. as $x | def f: ($x.a = 9); {a:1} | f` wrote on every route (`map_subexprs` does not walk a resolved def body; the demotion now does).

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite; new `*_3036` CLI tests and two unit tests)
- [x] `cargo fmt --check`
- [x] Oracle diff against `/usr/bin/jq` 1.7.1 for every row above (all 15 fabrication rows now agree; all must-stay rows agree)
- [x] `scripts/jq-bind-origin-oracle-sweep.sh`: 0 fabricate / 0 mismatch / 0 new refuse-only on this branch; against a baseline binary built from the parent commit the 33 new `in-evaluator-*` rows fabricate 15/15 (harness has power); no pre-existing row changed class
- [x] `scripts/jq-bind-origin-fuzz.py` with the new ROUTES family, 3 seeds × 2000, `--baseline` = parent-commit binary: 0 fabricate / 0 mismatch here; 79 fabrications on seed 1 on the baseline, 59 of them from the new family
- [x] Interleaved A/B (`scripts/ab-cli.py`, 7 reps, 2.2 MB inputs, output identity gated) on **Apple M4 Pro** and **AMD 7950X**, plus a never-triggering holdout build on the 7950X (`demote_for_owned_reentry` compiled to a no-op) to bound codegen bias. Final head vs current `main` on the 7950X: array write/fold/marker shapes **−0.80% median, 0/7 rows slower** (control floor −0.21%); input-queue (`-n inputs`) shapes +0.28% median. Earlier heads were pinned and corrected twice from this loop: the first cut cost +5% on `[inputs | . as $x | {s:.score} | $x.id]` (a per-record demotion rebuild no resolver could observe — now skipped), and the per-re-entry walk read +1% against the holdout / +2.9% on a tight `reduce` (now one walk, decided once per fold). M4 Pro on the pre-final head: array −0.49%, input-queue +0.86%, holdout-range noise.
- [x] Three `/code-review high` rounds; every finding reproduced against jq 1.7.1 and a merge-base binary before fixing (accept-direction regressions on `any`/`all`/`IN`/`until`/`recurse`/`debug`/`|=`-root/resolver-internal/`catch`/identity-route shapes; a `DefCall`-body gap that fabricated on every route; the `no_std` build)
- [x] `omni-dev coverage diff`: 97.5% patch coverage on the first cut; CI's Coverage legs green on every subsequent head
- [x] Rebased onto `main` past #2937/#2968/#3079 (one silent compile break from `collect_owned`'s new `S` parameter, fixed); full local gate re-run after the rebase


newhoggy:
Applied a code-review follow-up pass before merge (see #3122, #3123 for the two items deliberately deferred as follow-ups rather than fixed here):

- Fixed a correctness gap: `resolve_as_pattern` (destructuring `EXPR as {pattern}`) and `walk_pattern`'s computed object key never threaded `trackable` into their bind-source fork, unlike `resolve_bind_source_sink` (plain `EXPR as $var`), which this PR already fixed for the same class of bug.
- Fixed a no_std/32-bit build-rule violation: `fresh_owned_root` used `AtomicU64` where this crate's own rule (#2042 review) requires `AtomicUsize`.
- Hoisted `demote_for_owned_reentry` out of `ValueRecurseWalk`'s native descent and `until_step`/`while_step`'s per-iteration loop (both re-walked a static AST on every visited node/step).
- `owned_child_token` now hashes `Int`/`String` path components directly instead of round-tripping through `to_json()` + a hand-rolled FNV loop.
- Deduped `is_root_path` into `is_effectively_identity`, and the assignment-variant list shared by `may_enter_resolver_node`/`contains_assign_scoped`.
- Added oracle citations to two doc comments asserting jq behaviour, and documented the type-erasure boundary between `OwnedIdentity` and `BindOrigin::Owned`.

Full test suite + clippy + live jq-oracle spot checks all green.

Issue #2357, opened 2026-09-04
jq: unary minus in filter text preserves an exact literal where real jq collapses it to a double, masking divergences in the large-integer class


newhoggy:
**Severity: Low** (extreme magnitudes only — beyond `2^53`, where a `double` can no longer
hold the literal exactly), but load-bearing for #2131/#2219's own reasoning, which is why it
is worth recording rather than leaving as folklore.

## The divergence

Real jq's unary minus **destroys literal preservation**: the operand stops being a preserved
`decNumber` literal and becomes a `double`. succinctly keeps the exact value. Verified live
against `/usr/bin/jq` 1.7.1 and `main` at `90ecf68f8`:

| filter | jq 1.7.1 | succinctly |
|---|---|---|
| `-9223372036854775758` | `-9223372036854776000` | **`-9223372036854775758`** |
| `-9007199254740993` | `-9007199254740992` | **`-9007199254740993`** |
| `-1.10` | `-1.1` | `-1.1` (agrees — magnitude-specific) |

```bash
jq  -nc -- '-9223372036854775758'   # -9223372036854776000
sjq -nc -- '-9223372036854775758'   # -9223372036854775758
```

It is specifically the *filter-text unary minus*. The same value arriving as **data** keeps
its exact spelling in both tools, so this is not the general large-integer class:

```bash
echo '[-9223372036854775758]' | jq  -c '.[0]'   # -9223372036854775758
echo '[-9223372036854775758]' | sjq -c '.[0]'   # -9223372036854775758
```

## Why this is worth a ticket rather than folklore

It silently changes the answer of anything downstream of a negative literal at that
magnitude, and it has already caused a real analysis error in this repo.

#2131's headline repro is `range(-9223372036854775758; -9223372036854775808; -100)`, where
succinctly answers `[]` and so does jq — and that agreement was read as evidence that
succinctly's overflow handling was correct there. It is not evidence of anything of the sort.
jq answers `[]` **only because** its unary minus has already collapsed `from` to `-2^63`,
making it equal to `to`. Feed the identical numbers in as data, where jq keeps its literals,
and jq answers `[-9223372036854775758]` while succinctly still answers `[]`:

```bash
echo '[-9223372036854775758,-9223372036854775808,-100]' | jq  -c '[range(.[0];.[1];.[2])]'
# [-9223372036854775758]
echo '[-9223372036854775758,-9223372036854775808,-100]' | sjq -c '[range(.[0];.[1];.[2])]'
# []
[...]

PR #3072, opened 2026-09-17
fix(jq): round a literal-adjacent unary minus past 2^53, matching jq (#3044)

### LINKED ISSUE #3044: jq: unary minus keeps an int literal past 2^53 exact where jq rounds it; #2357 accepted this on a premise #2906 made false
**Severity: Low**

Found while triaging #2937 / #3041. `docs/compliance/jq/limitations.md` accepts the unary-minus divergence ("Unary minus in filter text destroys literal preservation — inherits the `range` divergence's own rule 4(c) grant", #2357): succinctly keeps an integer literal past 2^53 exact under `-`, where jq produces a double. The only justification recorded is that matching jq would single out unary minus:

> …would reintroduce precisely that failure mode for this operator alone — while `9223372036854775758` negated via `0 - 9223372036854775758`, or the identical magnitude arriving as data with its sign already attached, both keep the exact value today.

**That premise is no longer true.** Since #2631/#2906, jq-mode `-`, `+` and `*` already produce jq's double past 2^53. Binary subtraction is exactly the rounding the entry says succinctly avoids, so unary minus is now the odd one out, not the one that would be singled out:

| filter (main @ `b509a8019`)                             | jq 1.7.1                  | sjq                          |
|---------------------------------------------------------|---------------------------|------------------------------|
| `0 - 869389897822472004`                                | `-869389897822472000`     | `-869389897822472000` ✓      |
| `0 - 9223372036854775758`                               | `-9223372036854776000`    | `-9223372036854776000` ✓     |
| `-869389897822472004`, `-(869389897822472004)`          | `-869389897822472000`     | `-869389897822472004`        |
| `869389897822472004 \| -.`                              | `-869389897822472000`     | `-869389897822472004`        |
| data: `echo 869389897822472004 \| … -- '-.'`            | `-869389897822472000`     | `-869389897822472004`        |
| `-(-869389897822472004)`                                | `869389897822472000`      | `869389897822472004`         |
| `[-(1,869389897822472004)]`                             | `[-1,-869389897822472000]`| `[-1,-869389897822472004]`   |
| `-869389897822472004 == -869389897822472005`            | `true`                    | `false`                      |
| `-869389897822472004 \| tostring`                       | `"-869389897822472000"`   | `"-869389897822472004"`      |
| `-100000000000000001`                                   | `-1e+17`                  | `-100000000000000001`        |
| `-9007199254740993`                                     | `-9007199254740992`       | `-9007199254740993`          |
| `-9223372036854775807`, `-9223372036854775808`          | `-9223372036854776000`    | exact digits                 |
| `[range(-9223372036854775758; -9223372036854775808; -100)]` | `[]`                  | `[-9223372036854775758]`     |

These already agree and must keep agreeing: `-9007199254740992` and `-9007199254740994` (within 2^53, exact), `-1.10` → `-1.1`, `-(-9223372036854775808)` → `9223372036854776000`, `(-869389897822472004) == (0 - 869389897822472004)` → `true`, the data-sourced `echo '[-9223372036854775758]' | … '.[0]'` → exact digits (no negation, so the literal is kept), and data-sourced `range` → `[-9223372036854775758]` in both.

jq 1.7.1 compiles `-e` to `_negate`, which is `jv_number(-jv_number_value(x))`: the operand goes through the literal's 17-digit rounding and the result is a computed double. That is the same model #2906 applies to `0 - x`.

**yq mode.** Real yq has no unary minus (`-.a` → `'-' expects 2 args but there is 1`). succinctly's yq unary minus is an extension whose exact `int64` negation matches yq's own arithmetic, and it must not change.

## Triage plan (2026-09-17)

1. **`arith_negate`** (`src/jq/eval.rs`), jq arm only (`S::DECNUMBER_LITERALS`). For `Int(n)` (including a `NumberLiteral` int via `into_plain_number`): if `n.unsigned_abs() <= 2^53`, keep today's exact negation, including the existing `Int(0)` → `Float(-0.0)` rule. Otherwise return `Float(-jq_literal_int_to_f64(n))`. Reuse `jq_int_within_exact_f64_range`. Better, route through the helper `0 - n` already uses (`jq_checked_int_arith`), so the two spellings cannot drift apart; check that `-0` still comes out right if you do. The yq arm is unchanged.
2. **Every path that negates.** `each_negate`, `negate_fanout_core`, `eval_negate` and the generic evaluator's negate, if it has its own, must all reach `arith_negate`. Verify with the generator row and the data-sourced row above. Also check whether the parser folds `-<literal>` into a negative literal token anywhere; if it does, that fold must produce the same value.
3. **Docs.**
   - Delete the "Unary minus in filter text destroys literal preservation" entry from `docs/compliance/jq/limitations.md`.
   - Revise the `range` entry's paragraph that leans on it: the filter-text `range(-9223372036854775758; …)` row now matches jq's `[]`, and the data-sourced row still shows the `range` divergence.
   - Fix `jq_literal_int_to_f64`'s soundness paragraph in `src/jq/value.rs` ("unary minus and `length` produce the exact negation or magnitude of a literal"). #2937's plan edits the `length` half of the same sentence; whichever PR lands second rebases onto the first.
4. **Tests.**
   - Invert `test_unary_minus_destroys_literal_preservation_2357` (`tests/jq_cli_tests.rs`) to pin jq's answers.
   - Add jq golden cases (captured with `scripts/sync-jq-golden.sh`) for every table row, including the data-sourced `-.`, the generator row and the `==` row.
   - A yq CLI test pinning `--jq-extensions`-free `-.a` on `a: 869389897822472004` → exact digits.
   - Search the tests for other assertions built on the old exact negation (for example the `range` tests around #2131/#2219) and re-derive each from the oracle rather than editing the expected values.

**Related:** #2357 (accepted the divergence), #2906 / PR #2939, #2937, and #3041 (`(-869389897822472004) | abs` depends on this).

### Class: **Sonnet**

The change is to one function's jq arm, reusing the helper binary `-` already uses, plus removing a limitations entry whose premise the table disproves. Every row is pinned to the oracle, the yq side is explicitly unchanged, and the only non-mechanical part (finding tests that inherited the old rule) is spelled out above.

newhoggy:
## Summary

- `-9223372036854775758` (and any negative integer literal past `2^53`, in every spelling — bare literal, `-(expr)`, `expr | -.`) now prints `-9223372036854776000`, matching jq 1.7.1, instead of keeping the exact `i64` value.
- Closes the `#2357` divergence recorded in `docs/compliance/jq/limitations.md`: it was accepted as a necessary consequence of the `range` divergence's own ADR-0018 rule-4(c) grant, but `#2631`/`#2906` already made binary `-`/`+`/`*` round past `2^53`, so unary minus had become the one remaining holdout, not a consistent exception.
- `arith_negate` (`src/jq/eval.rs`) gained the same `jq_checked_int_arith`-style rounding rule binary ops already use. The bare-literal spelling needed a second fix in the parser: a literal-adjacent `-` now folds into `-1 * <positive literal>` (the same desugaring `#1035` already used for a negative float/exponent literal) whenever the magnitude is outside `jq_int_within_exact_f64_range` (made `pub(crate)` for this reuse), not only when the literal contains `.`/`e`/`E`. A magnitude within range — the common case, e.g. every `.[-1]`-style index — keeps its old AST shape untouched, so `fold_index_key`'s literal-index fast path is unaffected.
- This cascades into `range`'s own accepted exact-`i64` divergence: a bare negative-literal bound past `2^53` no longer reaches its fast path directly (it's now a computed `Float`, like every other rounded operand), so `range(-9223372036854775758; -9223372036854775808; -100)` written literally now agrees with jq's own empty answer. The fast path itself is unchanged and still reachable with these exact magnitudes via data (`.[0]`, `--argjson`, ...) — retargeted the two existing tests that used to reach it through a bare literal to source from data instead.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite — 0 failures)
- [x] Diffed against the pinned `/usr/bin/jq` 1.7.1 oracle directly for every row in the issue's repro table, plus `.[-1]`/`path(.[-1])` to confirm the common small-index fast path is untouched
- [x] `cargo llvm-cov` / `omni-dev coverage diff`: 92.86% patch coverage; the 4 uncovered new lines are `let-else` panic bodies that can never fire in a passing suite (same `// omni-dev: coverage tolerate-line` pattern used ~100 times elsewhere in this codebase for the identical situation)

Fixes #3044


newhoggy:
Filed #3075 for a code-review finding that's out of scope here: the out-of-range negative-literal split builds a runtime `Arithmetic::Mul` node that recomputes jq's rounding on every evaluation instead of folding the (fully known at parse time) result once. Pure performance follow-up, no behavioral change.

Issue #3040, opened 2026-09-16
yq: a YAML float with >17 significant digits loses its source text, so unrelated writes rewrite it and == compares the wrong spelling


newhoggy:
**Severity: Medium.** A write that doesn't touch a value still rewrites it on disk.

Found while triaging #2936. Checked against Homebrew `yq` v4.53.3 and `main` @ `b509a8019`.

## Repro

When a plain or `!!float` YAML scalar has more than 17 significant mantissa digits, real yq keeps its source text byte-for-byte. `succinctly yq` replaces the text with the shortest spelling of the parsed `f64` on every route except top-level streaming of the scalar itself:

```console
$ printf 'z: 2.7293109604053567083\n' > w.yaml
$ yq -i '.n = 1' w.yaml            ; cat w.yaml   # z: 2.7293109604053567083  n: 1
$ succinctly yq -i '.n = 1' w.yaml ; cat w.yaml   # z: 2.729310960405357      n: 1   <- untouched key rewritten
```

On `p.yaml` (`c: 2.7293109604053567083`, `t: !!float 2.7293109604053567083`, `e: 2.7293109604053567083e5`, `z: 0.000000000000000000027293109604053567083`), yq first, then succinctly yq:

```console
.c                      # 2.7293109604053567083 | 2.7293109604053567083   (streaming path: already fine)
[.c,.t,.e,.z]           # 2.7293109604053567083, !!float 2.7293109604053567083, 2.7293109604053567083e5, 0.000000000000000000027293109604053567083
                        # | 2.729310960405357, 2.729310960405357, 272931.09604053566, 0.000000000000000000027293109604053566
-o json .               # {"c":2.7293109604053567083,...}   | {"c":2.729310960405357,...}
.c = .c                 # every long float kept             | every long float rewritten (the !!float tag is dropped too)
.c | tostring           # 2.7293109604053567083             | 2.729310960405357
{"v": .c}, .c as $v | $v, .c | @json, [.c] | sort   # source text | shortest f64
```

**`==` is also affected.** yq compares these scalars by their text, and succinctly yq compares against the rewritten text. On `c: 2.7293109604053567083`, yq first, then succinctly yq:

```console
.c == 2.7293109604053567083     # true  | false
.c == "2.7293109604053567083"   # true  | false
.c == 2.729310960405357         # false | true
.c == (.c + 0)                  # false | true
select(.c == 2.7293109604053567083)   # hit | no output
```

Controls that already match:
[...]

Issue #3032, opened 2026-09-16
jq: fromjson rejects 007, .5 and 1.e5 and clamps 1e400; its hand-written parser has none of the jq number leniencies


newhoggy:
**Severity: Medium** (refuse-where-jq-accepts on ordinary documents: any `fromjson` over text containing a leading-zero, leading-dot or trailing-dot number errors, and an overflowing literal loses its spelling)

## Summary

`fromjson` parses with its own hand-written JSON parser (`jq::eval::parse_json_value` and its `parse_json_number`). It predates, and never received, the number leniencies every other jq-mode input path has gained: #1149 (leading zero), #1171 (leading dot), #2220 (trailing dot before an exponent), and #1083/#2052 (preserving an overflowing literal). The primary input, `--argjson`, `--slurpfile` and `--seq` all accept these spellings; `fromjson` rejects them or loses them.

Found while re-triaging #2877 (comment 5699257118, §2(f)).

## Repro

Captured live, `/usr/bin/jq` 1.7.1 vs `succinctly` at `b509a8019`, each command run on its own:

| `$s \| fromjson` with `s =` | jq 1.7.1 | succinctly |
|---|---|---|
| `007` | `7` | exit 5 `Invalid numeric literal at EOF at line 1, column 3 (while parsing '007')` |
| `.5` | `0.5` | exit 5 `Invalid numeric literal …` |
| `1.e5` | `1E+5` | exit 5 `Invalid numeric literal …` |
| `007.500` | `7.500` | exit 5 `Invalid numeric literal …` |
| `[007,.5]` | `[7,0.5]` | exit 5 `Invalid numeric literal …` |
| `1e400` | `1E+400` | `1.7976931348623157e+308` (exit 0) |

```console
$ jq -nc --arg s 007 '$s|fromjson'                  # 7
$ succinctly jq -nc --arg s 007 '$s|fromjson'       # jq: error (at <unknown>): Invalid numeric literal ...  [exit 5]
$ echo '[007]' | succinctly jq -c .                 # [7]   <- the document path already accepts it
```

#2877's spellings (`+1`, `nan`, `sNaN`, `Infinity`) also fail here; #2877 tracks them on the other paths.

## Root cause

[...]

Issue #3068, opened 2026-09-17
jq: todate/todateiso8601 on a non-number say `math function requires number`; jq says `strftime/1 requires parsed datetime inputs`


newhoggy:
**Severity: Low**

## Summary

`todate` and `todateiso8601` are `def todate: strftime("%Y-%m-%dT%H:%M:%SZ")` in jq, so a non-number input raises `strftime`'s own message. succinctly's `builtin_todate` reads its input through the math family's `get_float_value_with` and reports a message neither jq nor (since #3042) the math builtins use.

## Repro

```console
$ jq  -n '"x" | todate'          # jq: error (at <unknown>): strftime/1 requires parsed datetime inputs
$ sjq -n '"x" | todate'          # jq: error (at <unknown>): math function requires number
$ jq  -n '"x" | todateiso8601'   # strftime/1 requires parsed datetime inputs
$ sjq -n '"x" | todateiso8601'   # math function requires number
$ sjq -n '"x" | strftime("%Y")'  # strftime/1 requires parsed datetime inputs   <- already right
```

jq prints the same `strftime/1 ...` message for `null`, arrays and strings (captured live from `/usr/bin/jq` 1.7.1).

## Root cause

`src/jq/eval.rs`, `builtin_todate` (and `builtin_tz`, a yq-mode extension with no jq oracle): `get_float_value_with(&value, optional, || EvalError::new("math function requires number"))`. #3042 moved the math family's `get_float_value` onto jq's `<type> (<value>) number required` and deliberately left these two on their old wording, since neither the old nor the new one is jq's for `todate`.

## Why this is separate from #3042

#3042 is the libm family and its `LIBM_DD` wording; `todate` is a datetime builtin whose reference wording comes from `strftime`. Fixing it means routing `todate`'s non-number case through whatever `builtin_strftime` already raises (it already matches), plus an error-corpus probe.

## Suggested fix direction

Make `builtin_todate`/`todateiso8601` raise the same error `strftime` does on a non-number/non-array input (`strftime/1 requires parsed datetime inputs`), add `todate_string`/`todate_null` probes to `tests/data/jq-error-probes.tsv`, and check `tz`'s message against real yq (`.a | tz("UTC")` on a string) rather than inheriting a jq-flavoured one.

newhoggy:
🔧 Claimed by agent `Brian` — branch `issue-3068-todate-error-message`

PR #3080, opened 2026-09-17
fix(jq): round every number literal to 17 significant digits before the double conversion

### LINKED ISSUE #2936: jq: >17-significant-digit float and over-i64 integer literals must round to 17 digits before f64 (same mechanism as #2906)
**Severity: Low**

#2906 established (from jq 1.7.1's `jvp_literal_number_to_double`, `src/jv.c`) that jq
converts every *literal* to a double by first rounding its decimal value to 17 significant
digits (half-even, `decNumberReduce` under a DECIMAL64 context with `digits = 17`), then
`strtod`. That fix covered `i64` literals only. The same rule applies to any literal with
more than 17 significant digits, which succinctly still parses with one correct rounding:

```console
$ jq  -n '2.7293109604053567083 + 0'    # 2.7293109604053565
$ sjq -n '2.7293109604053567083 + 0'    # 2.729310960405357
```

Verified on 40/40 random 18–30-digit fractions against `/usr/bin/jq` 1.7.1 (the 17-digit
model matched every one; the plain parse matched none). Integer literals beyond `i64`
(20+ digits, parsed as `f64`) are the same case.

Why it was left out of #2906: `parse_i64_or_f64` (`src/jq/value.rs`) is mode-less and is
reached from the program parser (`parser.rs`, which knows `ParserMode`), the JSON document
funnels (`OwnedValue::from_number_bytes`/`from_number_literal`, reached in yq mode too via
`--input-format json`), `JsonNumber::as_f64`, `JqValue::as_f64`, `tonumber`, `fromjson`
and `--argjson`. Real yq parses with Go's correctly-rounded `ParseFloat` (`.c + 0` on
`c: 2.7293109604053567083` is `2.729310960405357` in yq v4.53.3, as here), so the rounding
must be jq-mode only and the mode has to be threaded through those funnels.

Suggested shape: a text-level sibling of `jq_literal_int_to_f64` that rounds the digit
string to 17 significant digits half-even (carry-aware) and parses the result; apply it at
literal-materialisation time in jq mode so the stored `Float` is jq's double and every
consumer inherits it. Display is unaffected (`NumberLiteral` keeps the source text).

newhoggy:
## Summary

Fixes #2936.

jq 1.7.1 converts a literal to a double by first rounding its decimal value to 17 significant digits (`jvp_literal_number_to_double`: `decNumberReduce` under a DECIMAL64 context with `digits` = 17, half-even, then `strtod`). #2906 modelled that for `i64` literals only; every fraction, exponent form and integer past `i64` still took Rust's single correctly rounded parse, which lands on a different double on 2-6% of 18-30-digit literals:

```console
$ jq  -n '2.7293109604053567083 + 0'         # 2.7293109604053565
$ sjq -n '2.7293109604053567083 + 0'         # 2.729310960405357   (before)
$ jq  -n '2.4703282292062327209e-324 + 0'    # 0        (the subnormal half-way case; was 5e-324)
$ jq  -n '1.797693134862315808e308 | isinfinite'   # false  (was true)
```

Follows the 2026-09-17 triage plan: store jq's double when the value is created, and thread the mode in explicitly. Reviewed at high effort (three findings, all fixed in `f9bb2fb62`'s predecessor commit and pinned); rebased over #3044/#2929.

**What changed**

- **Helper**: `jq_literal_text_to_f64` (`value.rs`) is the sibling of #2906's `jq_literal_int_to_f64` for non-`i64` literals. A text of ≤17 bytes cannot hold 18 significant digits and takes the plain parse with no extra work at all (nearly every document number); a longer text is digit-counted, and only a genuine 18+-digit literal takes the slow path, which walks the mantissa once into a stack buffer of 18 digits plus a "nonzero beyond" flag, rounds half-even in place with carry, and parses a stack-spelled `d.ddd…e±x`. No allocation on either path. (The first A/B caught both a per-float digit scan and a `Vec`+`String` slow path; see the benchmark comment.) Pinned against `/usr/bin/jq` 1.7.1 on a 2,015-row table (`tests/data/jq-literal-17-digit-oracle-2936.tsv`: 18-30-digit fractions, over-i64 integers, exponent forms, a near-tie generator with `5`/`50`/`500001`/`49999`/`5000000000000001` tails, the lenient spellings, carry and overflow boundaries); the test also asserts the plain parse disagrees on enough rows to have power.
- **Mode-aware funnels**: `parse_i64_or_f64_in::<S>`; `OwnedValue::from_number_literal`, `from_number_literal_boxed` and `from_number_bytes` take `S: EvalSemantics` (jq's double under `DECNUMBER_LITERALS`, the plain parse for yq). **There is no mode-less form left**, so a funnel cannot default silently. The compiler then listed every site: the mode is threaded through both evaluators' materializers (`to_owned`, `to_owned_lossy`, `to_owned_checked`, `to_owned_cursor`, `collect_owned`, `into_owned`, `materialize_atomic`, `stream_json`/`stream_yaml`, …) and the ~60 W-only builtins that reach them. `JqValue` (`lazy.rs`) is `succinctly jq`'s own representation and reads under `JqSemantics`; a YAML scalar (`yaml/scalar.rs`) is yq's number and reads under `YqSemantics`; `LazySeq::fold_one` dispatches on the per-stage `Instruction::tag` it already carries (`Iterator::next` cannot be generic); each CLI runner pins its mode at its entry points. `from_document_float` stays mode-less on purpose (it bakes a double's own exact expansion, which the 17-digit rounding parses back to the same double). `JsonNumber::as_f64` stays the plain JSON-library answer; its fallback decodes the bridge tokens via the new mode-free `OwnedValue::bridge_nonfinite_from_bytes`.
- **Parser**: the number token stores the mode's double (`fold_index_key` and the unary-minus fold read that repr; negation is exact and the rounding sign-symmetric). `From<Literal>`'s debug assert accepts either mode's parse.
- **`tonumber` / `fromjson`** follow the rule in jq mode only (`fromjson`'s hand-written parser still yields a bare `Float`; moving it onto the funnels is #3032).
- **`jq_numeric_cmp` (plan step 7)**: `cmp_literals` orders two literals by their doubles before the exact decimal tie-break, which is sound only under one monotone rounding. An `Int` literal's double now comes from `jq_literal_int_to_f64`, never `as f64`. Without this, `869389897822472001 < 869389897822472004.9` flipped to `false` (the plain `…472064` vs jq's `…471936`); with it, both tie at `…471936` and the digit comparison decides. Guard rows pinned.
- **Cursor-level readers (plan step 5)**: `length`, `isnan`/`isinfinite`/`isnormal`, `normals`/`finites`, `strftime`, `implode` and every libm builtin through `get_float_value` never build an `OwnedValue`; they now read through one accessor, `document_number_f64::<S>`, which rounds a literal spelling in jq mode and hands everything else (bridge tokens, decNumber words, lenient spans) to the plain `as_f64`. #2937's `Int` widening can add its arm there. A side effect already visible: `869389897822472004 | sqrt` is jq's `932410798.8555645` now (the plan's #2906 characterization row moved; the floor-family *display* gap stays #2937's).

**yq is unmoved**: Go's `ParseFloat` is correctly rounded, so `YqSemantics` keeps the plain parse in every funnel, including the reindex bridge under a write and the `-p json` DOM route. Pinned against yq v4.53.3 in `test_yq_float_literal_stays_correctly_rounded_unaffected_by_2936`. Every yq mismatch a wider probe found pre-exists on `main` (the `.c == <long literal>` row is the side observation the triage plan already noted for its own issue; `.c | tostring` is #3040's 17-digit YAML cap).

**Docs**: new `docs/compliance/jq/limitations.md` section; the #2906 entry's "still open" paragraph re-scoped to #2937's display gap; CHANGELOG (the `pub` signature changes are called out).

## Test plan

- [x] `cargo test --no-default-features`
- [x] Oracle: 2,015-row literal table vs jq 1.7.1 (0 mismatches; plain parse separates on 40+ rows); the plan's full row set (program literals, 16 document filters, `--slurp`, `input`/`inputs`, `--argjson`, `--jsonargs`, `--slurpfile`, `--seq`, `tonumber`, `fromjson`) all match jq
- [x] New tests: `tests/jq_cli_tests.rs` entry-point × operation matrix with four separating witnesses, every-reader rows, `jq_numeric_cmp` guard rows, must-not-move display rows; `tests/jq_evaluator_parity_tests.rs` both-evaluator rows; `tests/yq_cli_tests.rs` must-not-change rows; `value.rs` unit tests for the helper and its oracle table
- [x] Patch coverage 96.8% before the last commit (2205/2278 new lines); most of the uncovered lines are pre-existing code rustfmt reflowed when a turbofish lengthened its line, in arms that were already uncovered on `main`. Every genuinely new uncovered line in `value.rs` is now either exercised (`+`-signed slow-path row) or annotated `tolerate-line` with the reason it is unreachable by construction
- [x] Interleaved A/B on the M4 Pro and the 7950X (two rounds; the first caught two real costs, fixed in `9e5e84711`, tables and reading in the comments below): `numbers`/`users`/`mixed` fixtures plus a control input where most numbers have 18-22 significant digits (the slow path), `-c .` / `[..|numbers]|add` / `[..|numbers|.+0]|length` / `[..|numbers]|sort|length` and pretty identity, output-identity gated, `--control` per box


newhoggy:
## Review (code-review, high effort) and interleaved A/B on both architectures

The review filtered the ~3,300-line diff down to the 141 non-mechanical hunks, checked every hard-coded `JqSemantics`/`YqSemantics` in non-test code, and ran ~180 differential probes plus a 1,186-literal seeded fuzz against jq 1.7.1 and yq v4.53.3 (0 rounding differences; 56 fuzz rows separate the plain parse). Three findings, all fixed in the follow-up commit with regression rows:

- **Real bug**: the slow path did not require the whole text to be a number, so a malformed 18+-digit span (`[123456789012345678901e]`, `[…901.2.3]`) had a valid prefix read out of it and printed a fabricated value on the materializing route where `main` (and the short spellings) answer #966's `null`. The exponent must now be sign + digits and nothing may follow the mantissa; the funnels fall back to `null` as before.
- **Doc overclaim**: the CHANGELOG said the math builtins' bare-cast `Int` widening had closed; only the unary `get_float_value` family had. `math_operand` (pow/fma/fmin/ldexp/…) now widens through `int_to_f64::<S>` too, so the claim is true (`pow(869389897822472004; 1)` prints jq's `869389897822472000`).
- A misplaced rustdoc block (`document_number_f64_generic` had been inserted mid-way through `to_owned_checked`'s doc comment).

### Interleaved A/B (`scripts/ab-cli.py`, 9 reps, release `codegen-units=1, lto=fat`, base `00e907ab3`, output-identity gated: 20 configurations, 0 differences on each box)

Fixtures: `numbers` 3.9 MB (8-digit floats), `users` 2.8 MB, `mixed` 1.5 MB, and a **control input `longnums` 4.0 MB whose 181,787 numbers all have 18-22 significant digits** — the slow path on every number, which no real document does. Queries: `-c .`, `[..|numbers]|add`, `[..|numbers|.+0]|length`, `[..|numbers]|sort|length`, plus pretty identity.

**The first run caught two real costs**, at head `eefcfd379`: `numbers-4mb` read +5-8% on the M4 Pro (a per-float significant-digit scan) and `longnums` read +37-60% on both boxes (a `Vec`+`String` slow path). Commit `9e5e84711` gates the count behind a byte-length check and makes the slow path allocation-free. Second run, at that head:

```
# Apple M4 Pro (johns-mac-mini)
# control (base vs copy of base):
median of medians: -0.28%   mean: +0.03%   range: -1.6%..+4.5%   (5/16 rows slower)
tool=jq -o json extra-args=['-c']
longnums-4mb.json .                                26.3m       26.5m   +0.9%       26.8m       27.0m   +0.5%
longnums-4mb.json [..|numbers]|add                271.6m      325.8m  +20.0%      276.3m      333.4m  +20.7%
longnums-4mb.json [..|numbers|.+0]|length         337.4m      385.5m  +14.3%      342.7m      392.7m  +14.6%
longnums-4mb.json [..|numbers]|sort|length        309.6m      372.9m  +20.5%      320.3m      377.4m  +17.8%
mixed-12mb.json .                                  14.5m       14.5m   +0.4%       14.7m       14.6m   -0.7%
mixed-12mb.json [..|numbers]|add                  196.9m      193.5m   -1.7%      203.6m      205.1m   +0.7%
mixed-12mb.json [..|numbers|.+0]|length           214.1m      217.1m   +1.4%      228.5m      224.7m   -1.7%
mixed-12mb.json [..|numbers]|sort|length          198.9m      200.9m   +1.0%      205.1m      211.5m   +3.1%
numbers-4mb.json .                                 23.5m       24.0m   +1.9%       24.0m       24.1m   +0.5%
numbers-4mb.json [..|numbers]|add                 309.4m      307.9m   -0.5%      319.6m      314.4m   -1.6%
numbers-4mb.json [..|numbers|.+0]|length          405.4m      407.7m   +0.6%      418.1m      423.5m   +1.3%
numbers-4mb.json [..|numbers]|sort|length         327.5m      326.7m   -0.2%      339.4m      338.9m   -0.2%
users-4mb.json .                                   29.7m       30.1m   +1.4%       30.3m       30.2m   -0.2%
users-4mb.json [..|numbers]|add                   298.9m      303.8m   +1.6%      316.7m      315.4m   -0.4%
users-4mb.json [..|numbers|.+0]|length            329.2m      329.6m   +0.1%      334.9m      336.6m   +0.5%
users-4mb.json [..|numbers]|sort|length           306.7m      307.2m   +0.2%      328.1m      322.8m   -1.6%
median of medians: +0.49%   mean: +3.33%   range: -1.7%..+20.7%   (9/16 rows slower)
# pretty identity:
median of medians: +0.29%   mean: +0.15%   range: -0.7%..+0.7%   (3/4 rows slower)

# AMD Ryzen 9 7950X (terminus)
# control (base vs copy of base):
median of medians: -0.62%   mean: -0.70%   range: -3.8%..+0.6%   (3/16 rows slower)
tool=jq -o json extra-args=['-c']
longnums-4mb.json .                                31.5m       31.0m   -1.7%       33.1m       31.1m   -5.8%
longnums-4mb.json [..|numbers]|add                320.4m      366.5m  +14.4%      328.1m      368.2m  +12.2%
longnums-4mb.json [..|numbers|.+0]|length         394.6m      431.0m   +9.2%      399.8m      434.2m   +8.6%
longnums-4mb.json [..|numbers]|sort|length        375.5m      429.0m  +14.3%      380.6m      431.4m  +13.3%
mixed-12mb.json .                                  17.4m       16.9m   -2.8%       17.7m       17.2m   -3.1%
mixed-12mb.json [..|numbers]|add                  239.7m      236.0m   -1.5%      240.7m      237.0m   -1.5%
mixed-12mb.json [..|numbers|.+0]|length           268.3m      265.2m   -1.1%      272.0m      265.8m   -2.3%
mixed-12mb.json [..|numbers]|sort|length          245.1m      240.6m   -1.8%      248.1m      242.4m   -2.3%
numbers-4mb.json .                                 29.8m       29.1m   -2.1%       30.2m       29.5m   -2.4%
numbers-4mb.json [..|numbers]|add                 394.9m      385.8m   -2.3%      400.1m      390.7m   -2.4%
numbers-4mb.json [..|numbers|.+0]|length          493.2m      487.7m   -1.1%      502.5m      491.3m   -2.2%
numbers-4mb.json [..|numbers]|sort|length         417.1m      408.0m   -2.2%      419.3m      412.8m   -1.6%
users-4mb.json .                                   41.3m       39.9m   -3.4%       41.8m       40.5m   -3.1%
users-4mb.json [..|numbers]|add                   372.9m      368.6m   -1.2%      379.3m      373.1m   -1.6%
users-4mb.json [..|numbers|.+0]|length            396.3m      390.6m   -1.4%      397.7m      394.5m   -0.8%
users-4mb.json [..|numbers]|sort|length           382.8m      376.5m   -1.7%      386.7m      381.9m   -1.2%
median of medians: -1.92%   mean: +0.24%   range: -5.8%..+13.3%   (3/16 rows slower)
# pretty identity:
median of medians: -3.82%   mean: -3.76%   range: -5.8%..-1.6%   (0/4 rows slower)
```

**Reading.** Every ordinary fixture (`numbers`, `users`, `mixed`, all three queries and pretty identity) is inside the control band on both boxes — the byte-length gate means a document float of at most 17 bytes does exactly the work it did before. The `longnums` control input still pays +15-21% (M4 Pro) / +7-13% (7950X) on the three materializing queries: that is the genuine cost of the rounding itself on a literal that needs it (a second pass to spell and parse the 17-digit intermediate), on an input where *every* number needs it, and it is the price of the correct answer on those literals. The 7950X control run reads -0.6% median and the identity rows -3.8%, so the x86 numbers carry a small favourable layout bias this time; the two boxes agree on the shape.

newhoggy:
## Perf Regression Guard (cachegrind instruction counts vs merge-base, same run): both legs pass at 0.0%

The deterministic check confirms what the second A/B read: on every query in the matrix, x86_64 and ARM64-Linux alike, the instruction count is unchanged to the rounding shown (largest move `arrays_map_iterate` -0.1% / +0.2%). A document number of at most 17 bytes does exactly the work it did before.

```
# x86_64
query                            baseline        current    drift
wide_keys_unsorted            350,244,354    350,244,611    +0.0%
wide_escaped_keys_unsorted    376,855,524    376,855,570    +0.0%
wide_identity                 586,946,191    586,946,546    +0.0%
users_keys_unsorted            67,904,493     67,904,320    -0.0%
users_identity                358,829,201    358,829,527    +0.0%
arrays_identity               676,041,409    676,041,718    +0.0%
users_yq_keys_unsorted         82,848,720     82,848,507    -0.0%
arrays_first_map_iterate      121,186,789    121,186,818    +0.0%
arrays_map_iterate          1,171,795,519  1,170,915,356    -0.1%
users_del_select            1,381,511,775  1,381,688,766    +0.0%
users_del_bound_select      1,415,571,545  1,415,928,596    +0.0%
users_assign_scores         1,218,118,818  1,218,118,404    -0.0%
users_path_walk               313,361,739    313,376,754    +0.0%
users_yq_del_select         3,266,163,935  3,266,292,714    +0.0%

# ARM64-Linux
query                            baseline        current    drift
wide_keys_unsorted            286,507,602    286,503,442    -0.0%
wide_escaped_keys_unsorted    313,956,302    313,951,947    -0.0%
wide_identity                 464,891,600    464,887,562    -0.0%
users_keys_unsorted            41,341,640     41,337,008    -0.0%
users_identity                287,633,120    287,629,019    -0.0%
arrays_identity               600,857,035    600,853,058    -0.0%
users_yq_keys_unsorted         74,071,587     74,066,858    -0.0%
arrays_first_map_iterate      106,995,841    106,991,481    -0.0%
arrays_map_iterate          1,127,738,557  1,129,591,478    +0.2%
users_del_select            1,290,756,050  1,291,177,177    +0.0%
users_del_bound_select      1,327,792,336  1,328,484,683    +0.1%
users_assign_scores         1,116,069,424  1,116,143,463    +0.0%
users_path_walk               264,833,005    264,827,801    -0.0%
users_yq_del_select         2,671,386,398  2,671,639,250    +0.0%

```

CI coverage comment: patch **96.89%** (2214/2285); total 93.73%, +0.02 pp vs `main`. The uncovered remainder is almost entirely pre-existing lines rustfmt reflowed when a turbofish lengthened them, in arms already uncovered on `main`; the genuinely new lines are exercised or annotated (commit `2f73b489e`).

Rebasing over the 17 commits `main` gained meanwhile (#2999's structural sharing among them) and re-running the local gates before enqueueing.

newhoggy:
## Rebased over #2999 (structural sharing) and re-reviewed

The rebase hit 60 conflict hunks in `eval.rs`/`eval_generic.rs`/`lazy.rs`/`value.rs`, all of the same shape: #2999 rewrote `OwnedValue::Array(v)` as `OwnedValue::array_from(v)` on lines where this branch had added a `::<S, _>` turbofish. Each hunk was resolved to `main`'s side and the mode threading re-applied on top with the same compiler-driven pass that produced it (so the sharing semantics are `main`'s, unchanged), then the whole local gate set re-run. Since that was real conflict resolution, the review was repeated against the rebased diff at high effort; it fuzzed 600 program and 600 document literals against jq 1.7.1 (0 rounding differences) and found two incompleteness gaps, both fixed and pinned:

- **The generic evaluator's lenient-span fallback.** A span `number_literal()` declines because the funnels do not *preserve* it (rule 4c's bare trailing dot, `[14455058590201385605.]`) is still a decimal jq reads through its 17-digit rounding, but the CLI's route reached it via the plain `as_f64()` and printed `…385000` while the eager route and jq print `…387000`. `DocumentValue` gains `number_text()` (the raw span; `number_literal()` minus the preservable gate), `document_number_f64` now relies on the helper's own whole-text validation rather than that gate, and both materializer fallbacks plus `eval_builtin`'s `tonumber` arm read through it. Parity rows for the two spellings.
- **`tonumber`'s jq-mode fallback** for the same lenient strings (`"14455058590201385605."`, `"014…605"`, `"+…605.e0"`) was Rust's bare parse; it is `parse_i64_or_f64_in::<JqSemantics>` now. CLI rows.

With these, "no mode-less funnel left" holds for every path the two reviews and the fuzzers reached. Gates re-running on the new head; CI will follow.

newhoggy:
Merged as e0339fd98 via the merge queue (rebase, head carried verbatim).

Two rebases landed between the last review round and the merge, both textually clean or resolved to main's side, each followed by the full local gate set and a fresh 65/65 CI run:

- **over #2968** (native isempty/any/all/IN/skip arms): its new `generic_item_into_owned` / `generic_item_truthiness` / `to_owned` call sites were mode-less on main and failed to compile here (E0283), so they were pinned to the enclosing `S` like every other consumer on this branch (df75d46's last commit).
- **over #3068** (`todate` accepts `strftime`'s inputs): #3068 extracted the shared `broken_down_time_fields` helper, whose raw-timestamp arm read the number with a plain `as_f64`. This branch had already routed that arm in `strftime` through `json_number_f64::<S>`, so the helper takes `S` and reads through the number model too. Probe (jq 1.7.1 agrees on all three routes): `1700000000.99999999999999999 | todate` → `"2023-11-14T22:13:21Z"` (17-digit rounding carries the second, plain parse would give `:20Z`). yq's `from_unix` was kept exactly as #3068 left it (numeric-only, its own wording).

Issue #3042, opened 2026-09-16
jq: 36 libm builtins (cbrt, frexp, ldexp, significand, gamma, fma, ...) are unimplemented, and the libm crate differs from the pinned jq in the last bit


newhoggy:
**Severity: Low**

Found while triaging #2937. The pinned jq (`/usr/bin/jq`, jq-1.7.1-apple) exposes 36 libm builtins that succinctly does not implement. `JQ_BUILTIN_ROSTER` (`src/jq/resolve.rs`) accepts them at compile time, but a call that is actually reached fails with `undefined function: <name>/<arity>` (exit 5). The list comes from calling every name in `tests/data/jq-builtin-names.txt` on main @ `b509a8019`:

- **Arity 0 (`.` is the operand):** `cbrt`, `erf`, `erfc`, `expm1`, `frexp`, `gamma`, `j0`, `j1`, `lgamma`, `lgamma_r`, `log1p`, `logb`, `modf`, `nearbyint`, `pow10`, `rint`, `significand`, `tgamma`, `y0`, `y1`
- **Arity 2 (both operands are arguments; `.` is ignored):** `copysign`, `drem`, `fdim`, `fmax`, `fmin`, `fmod`, `hypot`, `jn`, `ldexp`, `nextafter`, `nexttoward`, `remainder`, `scalb`, `scalbln`, `yn`
- **Arity 3:** `fma`

The other 9 missing roster names (`JOIN/2..4`, `format/1`, `input_filename/0`, `get_search_list/0`, `get_jq_origin/0`, `get_prog_origin/0`, `strflocaltime/1`) are not math and are out of scope here.

yq v4.53.3's lexer rejects all of these (`cbrt` → `lexer: invalid input text`), so in yq mode they belong behind `--jq-extensions` (#1887's gate), with no reference to match.

### Oracle capture (jq-1.7.1-apple)

```console
[8,27,-8,2]|map(cbrt)               => [2,3,-2,1.2599210498948732]
[5,0.5]|map(gamma,lgamma,tgamma)    => [24,3.1780538303479453,24,1.772453850905516,0.5723649429247,1.772453850905516]
[5,-0.5]|map(lgamma_r)              => [[3.1780538303479453,1],[1.2655121234846454,-1]]
[8,0.75,0,-3]|map(frexp)            => [[0.5,4],[0.75,0],[0,0],[-0.75,2]]
[...]

PR #3082, opened 2026-09-17
fix(jq): todate/todateiso8601 raise strftime's own error, accept its array input too (#3068)

### LINKED ISSUE #3068: jq: todate/todateiso8601 on a non-number say `math function requires number`; jq says `strftime/1 requires parsed datetime inputs`

newhoggy:
## Summary

- `todate`/`todateiso8601` are literally `def todate: strftime("%Y-%m-%dT%H:%M:%SZ")` in jq, so a non-number input raises `strftime`'s own `strftime/1 requires parsed datetime inputs` — succinctly's `builtin_todate` instead read its input through the math family's `get_float_value_with`, reporting `math function requires number` (a wording `#3042` explicitly left behind at the time).
- That same narrow numeric-only path also rejected the *other* input shape real jq's `todate` accepts: an 8-element broken-down-time array (since `strftime` itself takes either a raw Unix timestamp or one — confirmed live: `[1970,0,1,0,0,0,4,0] | todate` succeeds in jq 1.7.1, `"1970-01-01T00:00:00Z"`).
- Extracted the number-or-array parsing `strftime_in_zone` already implements into a shared `broken_down_time_fields` helper (`src/jq/eval.rs`), so `builtin_todate` reuses the exact same input handling and error wording instead of duplicating a narrower copy of it.
- `builtin_tz` (a yq-mode extension with no jq oracle) is deliberately left untouched — its own pinned test already documents that decision, and investigating it live surfaced a much deeper divergence (real yq's `tz` parses its input as a date *string* via a Go time layout, not a numeric timestamp at all) that needs its own issue.
- Added `todate_string`/`todate_null`/`todateiso8601_string` to the error-probe corpus and re-synced `tests/data/jq-error-messages.tsv` against the pinned jq 1.7.1, updating `limitations.md`'s probe-count summary to match.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 65 binaries — 0 failures)
- [x] Diffed against the pinned `/usr/bin/jq` 1.7.1 oracle directly: string/null/array/short-array/NaN/infinity inputs, exit codes
- [x] `./scripts/sync-jq-error-messages.sh --check` passes; `jq_error_message_tests.rs`'s three tests (parity, known-divergences, doc-numbers) all pass
- [x] `cargo llvm-cov` / `omni-dev coverage diff`: 92.86% patch coverage; every uncovered line carries a `// omni-dev: coverage tolerate-line` marker with a reason verified live (an `eprintln!` probe confirmed `optional` is never `true` through this call path, and a computed array element still decodes as `NumberLiteral` rather than the bare `Int`/`Float` `get_int` also handles) — not guessed, and matching the repo's established pattern for genuinely-dead defensive arms

Fixes #3068


newhoggy:
Code review found a real regression, fixed in a follow-up commit: `from_unix` (a real yq builtin, previously delegating straight to `builtin_todate`) inherited `todate`'s new array-acceptance and jq-flavored error wording, diverging from real yq's own `from_unix` (confirmed live: it rejects both strings and arrays with its own message). Decoupled `from_unix` back to its exact pre-PR numeric-only behavior.

Also filed #3083 for a separate, pre-existing (not introduced by this PR) gap the review surfaced: `checked_month_index` only guards `i64::MAX` overflow, not extreme negative values, producing garbage output for adversarial broken-down-time arrays -- already reachable via `strftime` on `main`, this PR only widens where it's reachable from.

Issue #2483, opened 2026-09-05
yq: ordering comparisons against null are always false in real yq (null < 1 is false), succinctly says true


newhoggy:
Found while implementing #2470 (2026-09-06); pre-existing. Captured from yq v4.53.3 (`-o=json -I0`) on `a: 1`:

```console
$ yq 'null < 1'      # false
$ yq 'null <= 1'     # false
$ yq '.zzz < 1'      # false        (a real null node here: comparison operands are not read-only)
$ yq 'null > 1'      # false
$ succinctly yq 'null < 1'   # true   (jq's total order: null sorts before everything)
```

jq 1.7.1 says `null < 1` is `true`. Per ADR-0018 the mode decides; #2460's matrix already reproduces yq's ordering rule for an *empty* operand (always false) and this is the same answer for a *null* operand. Capture the whole null row (`==`, `!=`, `<`, `<=`, `>`, `>=` with null on each side, null vs null, null vs string/array/object) and add it to the yq-mode comparison rule as one definition next to `yq_empty_operand_output`.

newhoggy:
🔧 Claimed by agent `Johns-M5-Pro-Max-82509354` — branch `-` (executing #2416's plan)

PR #3088, opened 2026-09-17
fix(jq): give abs jq's own semantics instead of aliasing fabs (#3041)

### LINKED ISSUE #3041: jq: abs is aliased to fabs, but the pinned jq defines it as `if . < 0 then -. else . end` (passes non-numbers through, keeps literals)
**Severity: Low**

Found while triaging #2937. In succinctly, `abs` is an alias for `fabs` (`Builtin::Abs => builtin_fabs` in `src/jq/eval.rs`). In the pinned jq (`/usr/bin/jq`, jq-1.7.1-apple), `abs` is not a libm call. On every probe below it behaves exactly like the jq-defined filter

```jq
def abs: if . < 0 then -. else . end;
```

Because it compares with `<`, it passes every value that doesn't sort below `0` through unchanged: strings, arrays, objects, non-negative numbers (keeping their literal spelling), and `-0`. `null` and `false` do sort below `0`, so they fail with jq's unary-minus error.

(This is **not** jq 1.8's `if type == "number" and . < 0 …`: under that definition `null | abs` would give `null`, but the pinned binary raises `null (null) cannot be negated`.)

### Reproduction (main @ `b509a8019`)

| filter                                               | jq 1.7.1                                  | sjq                                        |
|------------------------------------------------------|-------------------------------------------|--------------------------------------------|
| `"a" \| abs`                                         | `"a"`                                     | `math function requires number`            |
| `[1] \| abs`, `{} \| abs`                            | `[1]`, `{}`                               | `math function requires number`            |
| `null \| abs`                                        | `null (null) cannot be negated`           | `math function requires number`            |
| `false \| abs`                                       | `boolean (false) cannot be negated`       | `math function requires number`            |
| `1.50 \| abs` (also from data: `echo 1.50 \| … abs`) | `1.50`                                    | `1.5`                                      |
| `1e500 \| abs`                                       | `1E+500`                                  | `1.7976931348623157e+308`                  |
| `(-0) \| abs`, `(-0.0) \| abs`                       | `-0`                                      | `0`                                        |
| `869389897822472004 \| abs`                          | `869389897822472004`                      | `869389897822472100`                       |
| `5 \| path(abs)`                                     | `[]`                                      | `Invalid path expression with result 5`    |
| `[-1.50,"x",[1]] \| map(abs)`                        | `[1.5,"x",[1]]`                           | `math function requires number`            |

These already agree and must keep agreeing: `(-5)|abs` → `5`, `(-1.5)|abs` → `1.5`, `-1e500|abs` → `1.7976931348623157e+308`, `nan|abs` → `null`, `(-5)|path(abs)` → `Invalid path expression`, `{"a":-2}|.a|=abs` → `{"a":2}`, `def abs: 42; -1|abs` → `42`.

Spelling the definition out by hand already matches jq on **every** row above except one, which comes from a separate, documented divergence (see below):

```console
$ sjq -nc '[-5, -1.5, 0, -0, -0.0, 5, 1e500, -1e500, nan] | map(if . < 0 then -. else . end)'
[5,1.5,0,-0,-0,5,1E+500,1.7976931348623157e+308,null]      # identical to jq's abs
```

**yq mode.** yq v4.53.3's lexer rejects `abs` (`lexer: invalid input text "abs"`). succinctly gates it behind `--jq-extensions`, so there is no reference to match there.

## Triage plan (2026-09-17)

1. In jq mode, make `abs` behave exactly like `if . < 0 then -. else . end`. Either desugar it where the parser currently returns `Builtin::Abs` (the parser knows the mode), or evaluate that expression shape in place of `builtin_fabs`. Rewriting it into an existing expression shape inherits the `path()` behaviour, literal preservation, generator fan-out and error wording already verified above. Before adding logic to any evaluator branch, check the precedent: see if another jq-defined builtin is already desugared this way.
2. Keep user shadowing working: `def abs: 42; -1 | abs` must still print `42`. If the desugar happens in the parser, confirm a user `def abs` still wins. The resolver's `JQ_BUILTIN_ROSTER` already lists `abs/0`.
3. yq mode (`--jq-extensions`): keep today's `fabs` behaviour. It is an extension with no oracle, and there is no reason to change a config-value type there. If `Builtin::Abs` survives only for yq, update its dispatcher comment; otherwise remove it from `walk.rs` (three match lists) and from `eval_generic.rs`'s path-rule table.
4. Tests:
   - jq golden cases (`tests/data/jq-golden/cases/`, captured with `scripts/sync-jq-golden.sh`) for every table row, **including the data-sourced `1.50` spelling**.
   - A CLI test that `def abs` shadowing still works.
   - A yq CLI test pinning `--jq-extensions '.a | abs'` on `a: -5` → `5`.
   - Update the unit tests in `src/jq/eval.rs` that assert `abs` returns an `OwnedValue::Float` (the `query!(b"5", "abs", …Float…)` group), since `5 | abs` stays the literal `5`.

**Depends on nothing. Related:** #2937 (the same large-literal rows for `fabs`), #2357 and #3044 (below).

### Remaining row (not in scope, filed as #3044): `(-869389897822472004) | abs`

jq gives `869389897822472000`; sjq (hand-written definition) gives `869389897822472004`. The cause is the filter-text unary minus, a divergence accepted under ADR-0018 rule 4(c) in `docs/compliance/jq/limitations.md` ("Unary minus in filter text destroys literal preservation", #2357), not `abs` itself. That entry's justification may be **out of date**. It argues that forcing the double would be a unary-minus-only special case "while `9223372036854775758` negated via `0 - 9223372036854775758` … keep[s] the exact value today". Since #2631/#2906 that is no longer true: `0 - 9223372036854775758` → `-9223372036854776000` and `0 - 869389897822472004` → `-869389897822472000` in both tools. `-869389897822472004` is now the odd one out. Filed as #3044; this issue should not change it.

### Class: **Sonnet**

The fix is one rewrite into an expression shape that already exists and was shown to match the oracle on every row, plus goldens and cleanup of one enum variant's match arms. No new semantics, and the only mode decision is made above.

newhoggy:
## Summary

- Real jq defines `abs` as `if . < 0 then -. else . end`, not a libm `fabs` call. succinctly's `builtin_fabs`-backed `abs` diverged in three ways, all confirmed live against jq 1.7.1:
  - **Non-numbers**: jq's total ordering (`null < false < true < numbers < strings < arrays < objects`) puts strings/arrays/objects *above* every number, so they pass through unchanged (`"a" | abs` → `"a"`, `[1] | abs` → `[1]`). The old `fabs` alias rejected them with `math function requires number`.
  - **null/false/true**: these sort *below* every number, so they take the `-.` branch and raise unary minus's own `<type> (<value>) cannot be negated` — a different message from the old alias's generic type error.
  - **Non-negative numbers**: jq's `else . end` is a pure identity, so a literal keeps its own spelling (`1.50 | abs` stays `"1.50"`, `(-0.0) | abs` stays `"-0"`). `fabs` always recomputes through `f64` (`1.50 | abs` gave `1.5`, `-0.0` lost its sign).
- `builtin_abs` classifies the value directly by jq's ordering (null/bool sort below zero; numbers compare numerically; everything else sorts above) and routes the negative case through `arith_negate` — the same function unary minus itself uses, so the "cannot be negated" wording and #3044's large-int-literal rounding both apply for free.
- Added `abs_null`/`abs_false`/`abs_true` to the error-probe corpus, re-synced `tests/data/jq-error-messages.tsv`, and fixed two stale doc references calling `abs` an alias for `fabs`.

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, 65 binaries — 0 failures)
- [x] Diffed against the pinned `/usr/bin/jq` 1.7.1 oracle directly: strings/arrays/objects, null/false/true, positive/negative/zero/NaN/infinity numbers, data-sourced literal preservation, exit codes
- [x] `./scripts/sync-jq-error-messages.sh --check` passes; `jq_error_message_tests.rs`'s three tests all pass
- [x] `cargo llvm-cov` / `omni-dev coverage diff`: 94.87% patch coverage; the 2 uncovered lines carry `// omni-dev: coverage tolerate-line` markers with reasons verified live via an `eprintln!` probe (a computed `Int` reaching `abs` arrives re-baked as a `NumberLiteral` by the reindex bridge before this function ever sees a bare one; `optional` never reaches `true` inside `builtin_abs` regardless of whether the filter writes `abs?`, the suppression happens entirely outside builtin dispatch)

Fixes #3041


newhoggy:
Code review found a real, live self-contradiction, fixed in a follow-up commit: `builtin_abs`'s original hand-rolled `. < 0` classification unconditionally treated `null`/bool as sorting below zero regardless of mode, but yq mode's own `<` treats a real `null` operand as unconditionally `false` (`yq_null_ordering_is_false`, #2483) -- since `abs` is reachable in yq mode behind `--jq-extensions` (same gate as `todate`/`strftime`), this meant `null < 0` said `false` while `null | abs` errored "cannot be negated" in the same binary/mode. Fixed by routing through `apply_compare_op::<S>`, the same shared comparator `eval_compare` already uses, which also fixes NaN's classification for free (unobservable today, but no longer relying on a hand-rolled copy that got it backwards).

Also filed #3089 for a separate, unconfirmed defense-in-depth gap two review passes surfaced but could not reproduce live: `abs`'s pass-through branch returns `to_owned_lossy`'s value directly, which silently corrupts undecodable content instead of raising `decode_failure` the way sibling builtins (`length`, `keys`, ...) do -- not reachable through the CLI's normal evaluation paths (upstream validation already catches this before `abs` is reached), but worth tracking.

Issue #2968, opened 2026-09-14
jq/yq: isempty/any/all/IN bridge unconditionally, so a positional read in their argument stubs (#2803 residue)


newhoggy:
**Severity: Low** — wrong answers, but only for a positional builtin (`key`/`parent`/`line`/`path`/`file_index`) read inside an argument of `isempty`/`any(gen;cond)`/`all(gen;cond)`/`IN`.

## Summary

`isempty`, `any(gen; cond)`, `all(gen; cond)`, `IN` and `IN(src; ...)` bridge to the eager
`eval.rs` evaluator **unconditionally**, dropping the cursor. Any positional read inside
their argument therefore resolves to its no-cursor default, even when the enclosing pipe
is on a position-aware route.

This is the residue of #2803. That issue fixed the routes gated by
`path_context_resolvable` / `path_context_resolve_constants` (assignment RHS, string
interpolation, `map_values`, the absent-key walk); this family does not go through those
gates at all, so it was out of reach of that fix.

## Repro

`succinctly yq --jq-extensions -o=json -I=0` on `a: {b: 1, c: 2}` — `.a | key` is `"a"`,
so `(key|length)` is 1 and `range(0; 1)` yields the single value `0`:

```console
$ .a | [range(0;(key|length))]              # [0]      <- correct (#2698/#2803)
$ .a | isempty(range(0;(key|length)))       # true     <- wrong, should be false
$ .a | any(range(0;(key|length)); . >= 0)   # false    <- wrong, should be true
$ .a | all(range(0;(key|length)); . >= 0)   # true     <- right, but only vacuously
```

`all` agrees by accident: an empty generator makes it vacuously true, which is also the
correct answer here. That coincidence is worth noting, because it means `all` is **not** a
safe row to regression-test this on.

## Root cause

`src/jq/eval_generic.rs`, in `eval_each_generic`:

```rust
Expr::Builtin(
    Builtin::IsEmpty(_) | Builtin::AnyCond(..) | Builtin::AllCond(..)
    | Builtin::UpperIn(_) | Builtin::UpperInSrc(..),
) => bridge_to_each_owned_flow::<S, V>(expr, value, cursor, optional, sink),
```

`bridge_to_each_owned_flow` uses the `cursor` only for `RootWitness::of` and
`bridge_ambient_input`, then calls `eval_each_owned` — the eager evaluator, which has no
cursor parameter, so the argument is evaluated position-free.

[...]

Issue #2476, opened 2026-09-05
yaml/jq: and/or/not/`//`/any remain O(2^N) over an alias fan-out; #1804's fix does not reach them


newhoggy:
**Severity: Medium**

## Summary

#1804's fix (branch `issue-1804-truthiness-alias-skip`) adds `DocumentCursor::is_alias()`
and short-circuits `push_generic_truthiness_cursor_error`'s subtree walk at an alias node,
fixing the `O(2^N)` blowup that walk caused for `select(.)` and `if`/`then`/`else` over a
document shaped as an alias fan-out (`aN: &aN [*a(N-1), *a(N-1)]`).

The 2026-09-04 triage on #1804 assumed `and`, `or`, `not`, `//` (alternative), and `any` all
funnel through the same walk and would be fixed alongside `select`/`if`. That assumption
does not hold against the current codebase: confirmed live that these five remain fully
`O(2^N)` after the #1804 fix lands.

## Repro (against the #1804-fixed binary)

    $ N=26; python3 -c "
    print('a0: &a0 leaf')
    for i in range(1, $N+1): print(f'a{i}: &a{i} [*a{i-1}, *a{i-1}]')
    " > chain.yaml
    $ time syq '.a26 and true' chain.yaml
    true
    real  109.35s   user  84.82s   sys  19.09s

    # select/if, same file, same N -- both flat:
    $ time syq '.a26 | select(.) | length'   # real 0.00s
    $ time syq 'if .a26 then 1 else 2 end'   # real 0.00s

`.a20 and true`, `(.a20 // 1) | length`, `[.a20]|any`, `.a20|not` all measured in the
1-2s range at N=20 (vs 0.00s for select/if) -- clearly on the same exponential curve, just
not yet timed out at that depth.

## Root cause

`eval_single` (`src/jq/eval_generic.rs`) has no native arm for `Expr::And`/`Expr::Or` (nor
does `eval_builtin` have one for `not`/`any` -- neither symbol appears in a real evaluation
arm anywhere in the file). Both fall through to the wildcard fallback, which calls
`full_eval::<Vec<u64>, S>(expr, cursor)` -- the eager/materializing evaluator -- unlike
`Expr::If`, which was given a native, cursor-threaded arm (`collect_each_generic` /
`each_if_generic`) as part of #2416 phase 3. `push_generic_truthiness_cursor_error` (and
therefore this fix's `is_alias()` check) is never consulted from that bridge at all.

## Why this is separate from #1804

[...]

PR #3092, opened 2026-09-17
perf(jq): native cursor-threaded arms for any(cond)/all(cond), isvalid, until and while (#2658)

### LINKED ISSUE #2658: jq/yq: any(cond)/any(gen;cond)/isvalid/while/until remain O(2^N) over an alias fan-out (#2476 residue)
**Severity: Low** (the plain spellings are fixed by #2476; these are the remaining spellings of the same class)

## Summary

#2476 gave `and`/`or`/`not`/`//`/`any`/`all` native arms in `eval_single`/`eval_builtin`, so
they no longer reach the wildcard bridge whose ambient materialization is `O(2^N)` on an alias
fan-out (`aN: &aN [*a(N-1), *a(N-1)]`). The step-4 audit on that branch timed the sibling
spellings and found each still bridged, and each carrying semantics that make it more than the
one-arm mirror the fixed constructs were:

| construct                                | route today       | N=22 (base) | N=22 (#2476 branch) |
|------------------------------------------|-------------------|-------------|---------------------|
| `Builtin::AnyF`/`AllF` (`any(cond)`)     | wildcard bridge   | 3.47s       | 3.64s               |
| `Builtin::AnyCond`/`AllCond` (`any(gen; cond)`) | wildcard bridge | 8.04s / 7.56s | 7.90s / 7.39s   |
| `Builtin::IsValid` (`isvalid(f)`)        | wildcard bridge   | 5.54s       | 5.63s               |
| `Expr::While`                            | wildcard bridge   | 3.98s       | 3.95s               |
| `Expr::Until`                            | wildcard bridge   | 3.98s       | 3.98s               |

(Apple M-series, release build, interleaved, 3 reps median; the box was not idle, so treat as
indicative. Outputs identical on both binaries.)

## Why each was left out of #2476

- `any_all_gen_cond` (`src/jq/eval.rs`) documents its owned bridge as deliberate: switching to
  the cursor path "would silently change duplicate-key and number-spelling fidelity". Moving it
  is a fidelity decision, not a refactor.
- `any_all_f_over` must `to_owned` each element it *reaches* before `cond` runs (#1755's
  short-circuit-timing rule), and `cond` runs through `eval_owned_expr_fork`, which has no
  generic twin. A native arm needs a cursor-native `cond` probe with the #932/#1519 escape rules.
- `builtin_isvalid` carries the #791 halt rule, the #867 break rule and the
  `Partial(_, Control::Error)` shape; reproducing them over `GenericResult` is new work.
- `while`/`until` conditions: same bridge, lower priority; note `.aN | until(. == 1; .)` is a
  genuinely non-terminating query on either binary (the body never changes the value), so
  benchmark with a body that terminates.

## Suggested fix direction

Same mechanism as #2476: `ambient_validation_error` once over the input (alias-safe since
#1804, so `O(N)` on the fan-out), then evaluate with the cursor. The `retain_truthy_generic` /
`push_generic_truthiness` helpers from that branch are the truthiness primitives to reuse.
Each construct needs its own oracle table (jq 1.7.1 / yq v4.53.3) before pinning, and the
`any(gen; cond)` fidelity note above needs an explicit decision recorded in the PR.

Also worth noting while here: `[.aN] | any` is *not* in scope of any of this. Array
construction materializes its outputs by design (the output is the expanded tree), so it is
inherently exponential on this document; same for `.aN as $x`, `.aN == .aN` and `-o json`.

Refs #2476, #1804, #2416.

newhoggy:
Fixes #2658

## Summary

The last five spellings of #2476/#2968's class -- `any(cond)`/`all(cond)`, `isvalid(f)`, `until` and `while` -- each still reached the generic evaluator's wildcard bridge, whose first act is a `to_owned` of the whole ambient input: `O(2^N)` on an alias fan-out, and a whole-document validation for filters that read almost none of it. All five now have native, cursor-threaded arms in `src/jq/eval_generic.rs`:

- **`any_all_f_generic`**: the container's own elements (array items, or an object's values under the last-occurrence duplicate-key rule, as `.[]` walks them) each probed with `cond` at their own position through #2968's `any_all_probe_item_generic`, stopping at the first decisive one. *Not* desugared to `any(.[]; cond)` although jq defines it so: in yq mode `.[]` on a scalar yields nothing, where the eager arm raises `Cannot iterate over …` -- and the eager arm is what both routes have always agreed on (captured table in the arm's doc comment).
- **`isvalid_generic`**: `f` driven to exhaustion (a later error flips the verdict, #881) with each output run through `discard_generic_item` for the errors it still owes and otherwise left undecoded; `halt`/`break` pass through (#791/#867).
- **`each_loop_generic`/`loop_step_generic`**: `until_step`/`while_step` over a `LoopState` that keeps a document node as its cursor for as long as `update` stays in the document -- jq's backtracking definitions state for state (multi-output `cond` forks, the in-place single-output fast path, one shared `WHILE_UNTIL_MAX_STEPS` budget), streaming outputs to the sink as each branch decides them, so a downstream `first`/`limit` now stops an infinite loop on the cursor route (the eager arm ran to the step cap first).

Collecting twins in `eval_single`/`eval_builtin`, gated on a cursor like `Any`/`All`, with the #1309 live-input-queue deferral. The five are registered in the path-context gates (`needs_path_context`, `path_context_single_native`, `path_context_resolvable`, the absent/preserves/keeps-position gates) exactly as #2968 registered its consumers, so a `key`/`path` read inside `cond`/`update`/`f` answers the same whichever route the pipe takes -- the first review round found that leaving them out made `[.[] | any(key == "c")]` and `[.[] | any(key == "c"), key]` disagree. The positional *walk* (`path_context_step_generic`) is not taught the loops, the same residue #2968 left.

## Two eager-arm fixes the new arms exposed

- **`builtin_isvalid`**: `isvalid(.[])` on `[]`/`{}` answered `true` (an empty `.[]` comes back as an empty `Many`, not `None`) beside `isvalid(empty)`'s `false`; both are `false` now. And an error `try` cannot catch -- #1620's decode failure and its siblings -- passes through rather than being digested into `false`, the rule `try`/`?`/`//` already apply: `isvalid(.a | length)` on a document with an undecodable `.a` raises, `isvalid(.a)` (navigates, never decodes) is `true`.
- **`any_all_f`/`builtin_any`/`builtin_all`**: walk the object's effective values (`effective_fields`, as `eval.rs`'s own `Expr::Iterate` arm does), not every raw member -- `{"a":true,"a":false} | any(.)` answered `true` through the library's eager entry point where jq 1.7.1 says `false`. Unreachable from the CLI until now (the bridge collapsed the keys on the way in); the new parity test caught it.

## Validation contract (#2103)

What the five constructs validate follows the rule -- a filter validates exactly what it materializes -- because their input is a cursor rather than a copy. On `{"a":"\ud800","d":5}`: `any(true)`, `all(false)`, `isvalid(.d)`, `isvalid(.a)`, `until(true; .) | .d` and `[while(false; .)] | length` answer where the bridge raised; `[5, "\ud800"] | any(. == 5)` regains #1755's short-circuit; everything that reads the bad string (`any(. == "x")`, `isvalid(.a | length)`, a write in `update`) still raises. Recorded in `docs/compliance/jq/limitations.md` under the #2103 entry with the three-reason justification that entry already carries, and pinned in `test_any_all_cond_isvalid_loops_validate_only_what_they_read_2658` plus `test_lazy_validation_boundary_2168`'s answering column.

## Fidelity

- jq mode, both routes vs jq 1.7.1: a 1236-row differential (103 filters x 12 inputs; loop forks, `first`/`limit`, `label`/`break`, `?`/`try`, duplicate keys, number spellings, scalar errors) -- zero differences beyond jq's own OOM on two infinite-fork shapes and the documented step cap. Curated into `test_any_all_cond_while_until_native_arms_match_jq_2658` and `test_isvalid_native_arm_matches_eager_rules_2658`.
- yq mode (`--jq-extensions`; real yq rejects every spelling): 1545 rows new vs main on both routes -- identical except the `isvalid` fix above, the loops' new demand-driven stop (`first(while(true; .+1))` on an array, where the eager arm appended to the step cap and timed out), and `{a: true, a: false} | all(length > 0)`, which moved from the owned route's `boolean (false) has no length` to the cursor route's `true` -- real yq's `false | length` is `5`, so the cursor route is the faithful one; the owned-route `length`-on-scalar divergence is pre-existing and filed separately.
- Both evaluators: `test_any_all_cond_isvalid_loops_agree_across_evaluators_2658`, and `test_position_reads_inside_any_cond_isvalid_loops_agree_2658` for position reads inside the arguments in both modes. `scripts/jq-m2-streaming-sweep.sh` (744 cases) matches its expected table; `scripts/yq-alias-oracle-sweep.sh` shows the same 29 pre-existing divergence rows on both binaries.

## Numbers

Alias fan-out (`a0: &a0 [1]`, `aI: &aI [*a(I-1), *a(I-1)]`), `succinctly yq --jq-extensions`, Apple M-series release build, `/usr/bin/time -l`, indicative:

| query (`.aN \| …`)          | N=20 before       | N=22 before        | after, N=18..24 |
|-----------------------------|-------------------|--------------------|-----------------|
| `any(length == 2)`          | 0.85 s / 585 MB   | 3.46 s / 2361 MB   | 0.00 s / 9 MB   |
| `all(length == 2)`          | 0.72 s / 577 MB   | 4.35 s / 2361 MB   | 0.00 s / 9 MB   |
| `isvalid(.[0])`             | 0.61 s / 393 MB   | 2.50 s / 1581 MB   | 0.00 s / 8 MB   |
| `until(true; .) \| length`  | 1.60 s / 966 MB   | 4.43 s / 3060 MB   | 0.00 s / 8 MB   |
| `while(false; .)`           | 0.84 s / 762 MB   | 3.57 s / 3059 MB   | 0.00 s / 8 MB   |
| `until(length == 1; .[0])`  | --                | 5.60 s / 3198 MB   | 0.00 s / 9 MB   |
| `any(. == 1)` (compare term)| 0.36 s / 185 MB   | 1.47 s / 716 MB    | unchanged       |

`any(. == 1)` is the second exponential term the triage identified (`apply_compare_op` materializing the aliased element); it is step 7 of the plan and its own follow-up, as the acceptance bar here is "no root copy". Pinned as a completion test at N=24 in `test_any_all_cond_isvalid_loops_over_alias_fanout_complete_2658` (no wall-clock bound, per the #2476/#2173 precedent).

Interleaved A/B on ordinary JSON and on the fan-out (M4 Pro and 7950X, `scripts/ab-cli.py`, identity-gated, with never-triggering holdouts) is in the comment below; perf-guard's cachegrind counts are the deterministic arbiter.


newhoggy:
## Interleaved A/B, both architectures

`scripts/ab-cli.py` (binaries alternate within each repetition, output identity gated before any timing is read, idle/mains checks on), release profile `codegen-units=1, lto=fat` on both sides. **before** = merge-base `e0339fd98`, **after** = `64b7451a4` (the branch before the two style-only clippy commits; no code the compiler could treat differently). Boxes: `johns-mac-mini` (Apple M4 Pro, ARM64) and `terminus` (AMD Ryzen 9 7950X, x86_64). Every configuration passed the identity gate (18 + 14 per box, 0 differences).

### Control (before vs a copy of itself) — the noise floor

| box | median of medians | range |
|---|---:|---:|
| M4 Pro | -0.50% | -2.8% .. +0.7% |
| 7950X | -0.62% | -2.4% .. +0.8% |

### Ordinary JSON, jq mode (`users` generator, 1 MB and 10 MB, `-c`, 9 reps)

The first three rows never reach the new arms (holdouts); the rest do.

| input | query | M4 Pro before → after (median) | Δ | 7950X before → after (median) | Δ |
|---|---|---:|---:|---:|---:|
| `users-10mb.json` | `.` | 68.3 → 68.1 ms | -0.3% | 99.5 → 102.7 ms | +3.2% |
| `users-10mb.json` | `[.users[] \| .name] \| length` | 36.7 → 37.0 ms | +0.7% | 51.9 → 52.7 ms | +1.5% |
| `users-10mb.json` | `[.users[] \| select(.active)] \| length` | 39.3 → 39.5 ms | +0.5% | 54.2 → 54.7 ms | +0.9% |
| `users-10mb.json` | `[.users[] \| any(. == true)] \| length` | 198.9 → 99.9 ms | -49.8% | 236.8 → 134.9 ms | -43.0% |
| `users-10mb.json` | `[.users[] \| all(. != null)] \| length` | 210.0 → 107.3 ms | -48.9% | 245.0 → 144.2 ms | -41.1% |
| `users-10mb.json` | `[.users[] \| isvalid(.email)] \| length` | 188.5 → 39.8 ms | -78.9% | 215.6 → 60.0 ms | -72.2% |
| `users-10mb.json` | `[.users[] \| until(type != "object"; .name)] \| length` | 241.7 → 62.9 ms | -74.0% | 286.8 → 77.5 ms | -73.0% |
| `users-10mb.json` | `[.users[] \| first(while(true; .name))] \| length` | 306.0 → 26.1 ms | -91.5% | 437.1 → 36.0 ms | -91.8% |
| `users-1mb.json` | `.` | 9.5 → 9.5 ms | -0.4% | 11.2 → 11.5 ms | +2.6% |
| `users-1mb.json` | `[.users[] \| .name] \| length` | 6.5 → 6.5 ms | +0.3% | 6.3 → 6.5 ms | +3.5% |
| `users-1mb.json` | `[.users[] \| select(.active)] \| length` | 6.7 → 6.7 ms | -0.3% | 6.7 → 6.6 ms | -1.3% |
| `users-1mb.json` | `[.users[] \| any(. == true)] \| length` | 23.3 → 12.9 ms | -44.6% | 24.7 → 14.5 ms | -41.2% |
| `users-1mb.json` | `[.users[] \| all(. != null)] \| length` | 23.8 → 13.8 ms | -41.9% | 25.5 → 15.5 ms | -39.2% |
| `users-1mb.json` | `[.users[] \| isvalid(.email)] \| length` | 22.0 → 6.8 ms | -69.1% | 22.9 → 7.1 ms | -68.9% |
| `users-1mb.json` | `[.users[] \| until(type != "object"; .name)] \| length` | 27.3 → 9.2 ms | -66.3% | 29.5 → 9.2 ms | -68.9% |
| `users-1mb.json` | `[.users[] \| first(while(true; .name))] \| length` | 33.2 → 5.5 ms | -83.3% | 41.7 → 5.0 ms | -88.0% |

The new-arm rows are the bridge's per-call cost on ordinary input: the wildcard re-serialized the ambient value, re-indexed it and evaluated in `eval.rs` once per user object. `first(while(true; .name))` also shows the loops becoming demand-driven (the eager arm ran every loop to `WHILE_UNTIL_MAX_STEPS` before `first` saw anything).

**Holdouts.** M4 Pro: all inside the control range. 7950X: `.` reads +3.2% / +2.6% and `[.users[] | .name] | length` +1.5% / +3.5%, above that box's +0.8% control ceiling, while `select(.active)` sits inside it. Nothing in this diff is on the identity or `.name` path (the arms are reached only through `Builtin::AnyF/AllF/IsValid` and `Expr::Until/While`), so this is the x86 code-layout sensitivity the perf-guard notes already record for the identity queries (+5% from codegen-unit partitioning alone, #2720). Perf-guard's cachegrind instruction counts on this PR are the deterministic arbiter for that reading; a genuine identity-path regression would show there as a non-zero drift on `wide_identity`/`arrays_identity`, and a layout artefact as 0.0%.

A ninth row, `[.users[] | any(.score > 100)] | length`, was dropped from the table: `.score` on a scalar errors on the first user, so the row timed the error path on both sides (identical error, identity gate satisfied) — the "benchmark rows can time the error path" trap, caught by the row being 10x faster than its siblings.

### Alias fan-out, yq mode (`--jq-extensions -I0`, `z: *aN`, 3 reps)

Inputs are below 1 MB by construction: the term under test grows with N, not with bytes. `.z | any` is #2476's already-native arm, the control row.

| input | query | M4 Pro before → after (median) | Δ | 7950X before → after (median) | Δ |
|---|---|---:|---:|---:|---:|
| `fan18.yaml` | `.z \| any` | 3.0 → 3.0 ms | +0.1% | 1.3 → 1.3 ms | +1.8% |
| `fan18.yaml` | `.z \| any(length == 2)` | 197.0 → 4.2 ms | -97.9% | 396.3 → 2.0 ms | -99.5% |
| `fan18.yaml` | `.z \| all(length == 2)` | 248.0 → 4.1 ms | -98.4% | 505.6 → 2.1 ms | -99.6% |
| `fan18.yaml` | `.z \| isvalid(.[0])` | 152.0 → 4.1 ms | -97.3% | 292.5 → 2.0 ms | -99.3% |
| `fan18.yaml` | `.z \| until(true; .) \| length` | 258.0 → 4.1 ms | -98.4% | 517.0 → 1.9 ms | -99.6% |
| `fan18.yaml` | `.z \| [while(false; .)] \| length` | 200.8 → 4.1 ms | -98.0% | 434.2 → 2.0 ms | -99.5% |
| `fan18.yaml` | `.z \| until(length == 1; .[0]) \| .[0]` | 290.2 → 4.2 ms | -98.5% | 574.6 → 1.9 ms | -99.7% |
| `fan20.yaml` | `.z \| any` | 3.1 → 3.1 ms | -0.1% | 1.3 → 1.3 ms | +0.1% |
| `fan20.yaml` | `.z \| any(length == 2)` | 787.4 → 4.0 ms | -99.5% | 1601.4 → 1.9 ms | -99.9% |
| `fan20.yaml` | `.z \| all(length == 2)` | 1014.8 → 4.2 ms | -99.6% | 2135.1 → 2.0 ms | -99.9% |
| `fan20.yaml` | `.z \| isvalid(.[0])` | 589.4 → 4.1 ms | -99.3% | 1188.2 → 2.0 ms | -99.8% |
| `fan20.yaml` | `.z \| until(true; .) \| length` | 967.2 → 4.3 ms | -99.6% | 2052.2 → 2.0 ms | -99.9% |
| `fan20.yaml` | `.z \| [while(false; .)] \| length` | 818.3 → 4.3 ms | -99.5% | 1808.3 → 1.9 ms | -99.9% |
| `fan20.yaml` | `.z \| until(length == 1; .[0]) \| .[0]` | 1192.5 → 4.2 ms | -99.6% | 2359.2 → 1.9 ms | -99.9% |

Before: every row quadruples from N=18 to N=20 (`O(2^N)`). After: every row sits at the `.z | any` control's 3–4 ms (M4 Pro) / 2 ms (7950X) at both N, i.e. flat in N — the "linear" bar the triage plan's risk (a) asked to be named, for these six shapes. `any(. == 1)` is not in this table on purpose: its remaining term is the compare's own materialization of the aliased element (plan step 7), unchanged here and tracked as its own follow-up.

A first attempt at this section timed 2.7 ms on *both* sides with the identity gate passing: my `--extra-args` repeated the `-o json` the harness adds itself, so both binaries failed argument parsing identically. The run above is with that fixed, after timing the base binary directly on each box first (`.z | any(length == 2)` at N=20: 0.76 s / 614 MB on the M4 Pro) to confirm the query reaches the exponential path.

newhoggy:
## Review round 1 — both findings reproduced and fixed

**[medium] position reads inside `cond`/`update`/`f` were route-dependent.** Reproduced exactly as described (`[.[] | any(key == "b")]` vs the same stage beside `, key`, and the two library entry points disagreeing). The first round registered the arms but not the path-context gates, so the plain streaming route evaluated `cond` at the element's cursor while the positioned route fell back to the bridge and read `key` as `null`; the PR body's claim that leaving the gates alone "keeps whatever route it had" was wrong, because the arms' existence is what moved the plain route. Fixed the way #2968 registered its consumers: `needs_path_context`, `path_context_single_native`, `path_context_resolvable` (`isvalid`'s `f` rewritable at the stage's input; the element-positioned `any(cond)` `cond` and the state-positioned loop `cond`/`update` refused, so the pipe keeps the native streaming route), `step_can_yield_absent`, `path_context_stage_preserves_node` and `path_context_absent_keeps_position`.

Rebasing over #3079 (which landed meanwhile and gave `any(gen; cond)` a native owned-identity-pipe arm for a reading `cond`) exposed the same class on a third route: `.aa | map_values(any(key == "c"))` answered `false,false` beside `map_values(any(.[]; key == "c"))`'s `true,false`. `any(cond)` now takes #3079's three-part closure with `Expr::Iterate` as the generator (gate through the prefetch hook, rewriter prefetch, `eval_owned_identity_any_all` over `.[]`); a scalar member is answered in that dispatch with `any_all_f`'s own `Cannot iterate over …` rather than through `.[]` (yq-mode empty) or the ruled arm (whose prefetch recursed straight back — the first attempt overflowed the stack). The loops with a reading `cond`/`update` are resolved on the streaming and positioned routes but not on the owned identity pipe; that residue is recorded under #3079's limitations entry with the fix named, on the same footing as #3079's own `|=` gap.

Pinned in `test_position_reads_inside_any_cond_isvalid_loops_agree_2658` (both evaluators, both modes) and new rows in `test_any_all_cond_while_until_native_arms_match_jq_2658` (streaming, positioned and `map_values` spellings of the same read, all agreeing).

**[low] yq-mode duplicate-key rule differed between the twins.** Fixed: the eager `any`/`all`/`any(cond)`/`all(cond)` arms collapse with `effective_fields(.., true)` in both modes, as `any_all_f_generic`, the generic `.[]` arm and real yq's `.[]` (`{a: true, a: false} | [.[]]` is `[false]`) do; `S::COLLAPSE_DUPLICATE_KEYS` governs the *keys* builtins. The parity test asserts the yq-mode rows too.

Also since the review: rebased over 20 commits on `main` (#3041, #3079, #2937, #3068), a both-sides-append conflict in `tests/yq_cli_tests.rs` repaired (`76c792000`), and #3093 filed for the pre-existing owned-route `length`-on-scalar divergence the yq differential surfaced.

newhoggy:
## Perf Regression Guard (cachegrind instruction counts vs merge-base, same run): both legs pass at 0.0%

Every query in the matrix on x86_64 and ARM64-Linux is unchanged to the rounding shown (largest move `arrays_map_iterate` -0.0% / +0.0%, `users_yq_del_select` +0.0% on both). That is the deterministic reading for the A/B's one open question: the 7950X `.` and `[.users[] | .name] | length` rows at +1.5..+3.5% were code layout, not work -- `wide_identity`/`users_identity`/`arrays_identity` execute the identical instruction count on both binaries. Head `2f912f1f` (rebased on `71569fc87`, 65/65 checks green, patch coverage 93.1%).

newhoggy:
Merged as `bea1422a5` via the merge queue (rebase; all nine commits carried, #3051 and #3084 merged ahead in the same batch). Two commit subjects were reworded before the final push to fit the 100-character lint, content untouched. Follow-ups: #3093 (yq owned-route `length` on scalars, filed here) and the triage plan's step 7 (kind-first compare fast path for `any(. == 1)`'s remaining term), which stays with #2658's plan comment as the next piece.

Issue #2979, opened 2026-09-15
jq: a fold whose loop pattern is a `?//` alternation fabricates a path where jq refuses (15/6000 fuzzed, pre-existing)


newhoggy:
**Severity: High** (fabrication — succinctly answers where jq refuses; this is the direction `del()`/`=` write through)

## Summary

Surfaced while verifying #2896 with `scripts/jq-bind-origin-fuzz.py`. Independently of that
issue: a `reduce`/`foreach` whose loop pattern is a **`?//` alternation** fabricates a path
where jq refuses. Reproduced on `main` (`104405d2b`) as well as on #2896's branch — the two
binaries' fabricate sets are byte-identical, so this is pre-existing and unrelated to
`getpath`.

15-18 hits per 6000 generated programs, on every seed tried.

## Repro

No `getpath` anywhere — pure `?//` plus a nested fold:

```console
$ echo '{"a":2,"c":2,"d":true,"x":{"a":2,"c":{"b":1}},"arr":[2,2,2]}' \
  | jq -c 'path(foreach ([.a] | .[0]) as {a:$v0} ?// $v0 (.c; foreach (1) as $i (0; $v0; .); $v0.b?))'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "a" of {"a":2,...}
$ ... | succinctly jq -c '<same>'
        # no output, exit 0
```

The write direction is the hazard, and it is reachable:

```console
$ echo '{"a":{"b":1},"c":false,"d":{"b":1},"x":{"a":{"b":1},"c":"s"}}' \
  | jq 'del(foreach . as {a:$v0} ?// [$v0] (.c; ($v0 | .b?); ($v0 | getpath([]) | .b?)))'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 0 of {...}
```
(succinctly raises a *differently-worded* refusal here rather than writing, but other rows in
the same family exit 0 — the classifier counts 15 exit-0-where-jq-exits-5 cases per 6000.)

## Reproducing the full set

```console
$ cargo build --release --features cli
$ ./scripts/jq-bind-origin-fuzz.py -n 6000 --seed 20260915 --show 100
seed=20260915 n=6000 agree=5916 fabricate=15 mismatch=0 refuse-only=67 refuse-early=2
```

Every one of the 15 is a `?//` alternation in the fold's loop pattern. Seeds 777 and 31337
give 18 and 15 respectively, same family.

## Root cause

[...]

PR #3115, opened 2026-09-18
fix(jq): match destructuring patterns with jq's backtracking matcher

### LINKED ISSUE #2872: jq: computed-key destructuring pattern fan-out — path position and reduce/foreach still refuse cleanly (#2677 follow-up)

newhoggy:
## Summary

Fixes #2872.

One demand-driven, continuation-passing matcher (`walk_pattern_each`, generic over a `PatternMode`) now serves every consumer of a destructuring pattern: `. as PATTERN` in both evaluators, the owned-identity pipe, `reduce`/`foreach`'s own pattern, and every path-position form (`path()`, `del()`, `=`, `|=`, the path-mode folds). It follows jq's `gen_object_matcher`/`gen_array_matcher` compilation exactly: object entries in source order, array elements right to left, a computed key's outputs pulled one at a time (through `eval_each_owned`) only when something backtracks for them, a later entry's key re-run per earlier binding, and the body (or fold step) run at the end of every completed path.

What that closes, every row captured live against jq 1.7.1:

- **The wrong answers the triage found** (the issue's "refuses cleanly, never a wrong value" premise was false): `(. as {("a","c"):$q} ?// {a:$q} | $q) |= "X"` silently dropped a write (`{"a":"X","c":2}` → `{"a":"X","c":"X"}`); `reduce . as {("a","b"):$q} ?// $r (0; .+$q)` answered `0` (→ `3`); `[.[] | try (reduce ... as {("a","b"):$q} ...) catch "caught"]` answered `["caught","caught"]` (→ `[3,3]`); `null | . as {(error("E")):$q} | $q` answered `null` (→ raises `E`); `first(. as {("a", (input|"b")):$q} | $q)` consumed an input; `stderr` markers came out in the wrong order (`KB1B2` → `B1KB2`, `K1K2` → `K2K1`); and `. as {("a", halt_error(7)):$q} ?// $z | if $q==1 then error("boom") else ($q // $z) end` halted with exit 7 where jq (lazy) never reaches the halt and exits 0 — the existing #2873 test pinned the eager design's exit 7 and is corrected.
- **The fan-out itself** in path position and both folds, which refused: `[path(. as {("a","b"):$q} | $q)]` = `[["a"],["b"]]`, `[path(foreach .[] as {("c","d"):$q} (.; .; $q))]` walks every `(element, key)` pair with its own register, `[foreach (.,.) as {("a","b"):$q} (0; .+$q)]` = `[1,3,4,6]`, a zero-output key is zero steps and still wins a `?//` chain, a key `break` after k steps retries the next alternative from the state those steps left (`[label $o | reduce . as {("a", break $o):$q} ?// $r (0; .+1)]` = `[2]`), `halt` never retries.
- **The duplicate-name rule** is one pass over visit order (keep-first bare, keep-last under `?//`) at every nesting depth — `{"a":[1,2],"b":3} | . as {a:[$x,$x],b:$x} | $x` is `2`, `3` under `?//` — which is what jq's `bind_matcher`/preamble-slot compilation does; #1366's per-container spelling was this rule seen from index order.
- A float-spelled computed key keeps its spelling as a path component (`def f: 1.7; path(. as {(f):$q} | $q)` = `[1.7]`, was `[1]`), as `.[EXPR]` does since #1088.
- `reads_ambient_value`'s `Reduce`/`Foreach` arms now consult the pattern's computed keys (they run against the source element, so only side channels reach the document); the `walk.rs` comment claiming keys run against the *document* is corrected. The `AsPattern` arm's `any_subexpr` blind spot stays with its own issue (3017).
- `try_pattern_alternatives` is a collecting adapter over `each_pattern_alternatives` rather than a second copy of the `?//` loop; `extract_pattern_bindings`/`fold_object_pattern_entry`/`extract_single_pattern_binding`/`substitute_foreach_steps`/`ForeachStepAlternative`/`bind_fold_alternative`/`dedup_object_bindings`/`dedup_array_bindings`/`dedup_pattern_bindings` are gone.

**Deviation from the triage plan's Phase 0 ("containment, ship first"):** the full fix landed in this one PR, so the containment kind (an uncatchable "not yet supported" error that Phases 1–3 would then delete) was never needed. Phases 1–3 are the first commit, Phase 4's `reads_ambient_value` item the second.

**Perf** (the plan's flagged risk, the hot `.[] as {a:$x, b:$y}` path): interleaved A/B, 9 reps, main's release binary vs this branch, 4 MB input of 90k objects, output-identity gated, Apple M5 Max (not a pinned box — perf-guard runs on the PR for both archs):

| query | main med | branch med | Δ med |
|---|---|---|---|
| `[.[] \| .a + .b] \| add` (control) | 68.1 ms | 69.4 ms | +1.8% (noise) |
| `[.[] as {a:$x, b:$y} \| $x+$y] \| add` | 154.3 ms | 123.5 ms | **−20.0%** |
| `[.[] as {a:$x, c:{d:$z}} \| $x+$z] \| add` | 171.7 ms | 126.9 ms | **−26.1%** |
| `[.[] \| [.a,.b] as [$x,$y] \| $x+$y] \| add` | 153.8 ms | 115.5 ms | **−24.9%** |
| `reduce .[] as {a:$x, b:$y} (0; . + $x + $y)` | 185.0 ms | 149.8 ms | **−19.0%** |
| `[foreach .[] as {a:$x} (0; . + $x)] \| length` | 122.6 ms | 98.5 ms | **−19.7%** |
| `[path(.[] \| . as {a:$x} \| $x)] \| length` | 215.7 ms | 208.0 ms | −3.6% |
| `del(.[] \| . as {a:$x} \| $x) \| length` | 383.2 ms | 368.9 ms | −3.7% |

The bindings accumulate in one `Vec` (push/truncate on backtrack) instead of the per-entry `partial.clone()` of the cartesian fold, and the dedup pass allocates only when a name actually repeats.

**Pre-existing, out of scope (same output on `main` with a literal key; none involve a computed key):** #3125 (`null | path(null)` refuses where jq answers `[[]]`, and a `?//` then retries onto the wrong alternative), #3126 (a pattern-bound `$q` after a `stderr` stage in path position refuses; `try` answers the handler), #3127 (`path((BIND)?)` with a multi-output source and a `?//` chain swallows the terminal refusal into `[]`); `path(reduce . as {a:$q} …)`/`path(foreach . as {b:$q} …)` refusing on a source of `.` is the known #2979/#3124-class shape.

## Review round (`/code-review 3115 high`, 4 findings, all fixed as new commits)

1. **`null` + object key regressed vs `main`** — `null | def k: {}; [. as {(k):$a} | $a]` raised "Cannot index null with object" (jq/main: `[null]`). jq's `jv_get(null, k)` is `null` for a string, number *or object* key; `index_one_owned` lacked the object arm and the removed short-circuit had masked it. Fixed in `index_one_owned` (pattern path; `null | .[{}]` through the cursor-side `.[EXPR]` path is a separate pre-existing gap and unchanged), pinned.
2. **Recursion scaled with pattern *width*** — one nested continuation per entry: a 500k-entry literal pattern overflowed the CLI's 256 MB eval stack where `main`'s loop matcher handled 1M. Now a computed-key-free subtree is walked by a plain loop (`walk_pattern_once`) and the CPS recursion nests only at an entry whose subtree holds a computed key: 1M literal entries walk again. Nesting per *computed* key is inherent to the push model (each key's callback runs the rest of the walk) and still overflowed at 100k computed keys in one pattern, so the parser caps computed keys per pattern at `MAX_PATTERN_COMPUTED_KEYS = 4096` (one `?//` alternative, all nesting levels; ~2.5 KB/frame in release, so >25× margin) with a clean parse error beside `MAX_PATTERN_DEPTH` — the process-abort class of divergence ADR-0018 permits, recorded in `docs/compliance/jq/limitations.md`.
3. **O(n²) repeated-name check** per completed match — a set past 16 bindings now (a 20k-entry pattern over 200 elements had doubled its time).
4. **Doc claim inverted** — jq runs `path_intact` *before* `jv_get` for a computed key too: `{"a":[1]} | def k: {}; path(.a as {(k):$q} | $q)` is jq's "near attempt to access element {} of [1]" (was the kind error "Cannot index array with object" first, same on `main`). `PathPatternMode::step` now checks the register's identity first; pinned.

Re-verified after the fixes: full suite + `no_std` + both clippy configs + rustdoc + fmt + commit lint; fuzz seeds 5–6 (5,000 programs, `--baseline` main) — only the pre-existing classes above; A/B re-run unchanged (−18% to −27% on value/fold pattern queries, path mode −2 to −3%, control +1.4%).

## Test plan

- [x] `cargo test --features cli,simd,regex,serde --no-fail-fast` (65 binaries, all green; the jq/yq golden suites included)
- [x] `cargo test --no-default-features` (the `no_std` leg)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` and `cargo clippy --all-targets --features std,simd,serde,cli,regex,bench-runner,large-tests,mmap-tests -- -D warnings`
- [x] `RUSTDOCFLAGS='-D warnings' cargo doc --no-deps --all-features`
- [x] `cargo fmt --all -- --check`; `omni-dev git commit message lint --strict` on every commit
- [x] Oracle: 95-row hand table and `scripts/jq-pattern-fanout-fuzz.py` (new; stdout + exit + side-effect stream vs `/usr/bin/jq` 1.7.1), 7,500 programs over seeds 2–4 with `--baseline` = main's binary: 0 fabricate, 0 mismatch not reproduced by main; validated against main as `--bin` (flags 2 fabricate / 26 mismatch at n=300)
- [x] New pins: `test_pattern_computed_key_{fans_out_in_path_position,matcher_is_lazy,fans_out_in_folds}_2872` (`tests/jq_cli_tests.rs`, exact stderr for side-effect order), `test_pattern_matcher_{visits_in_block_order,pulls_keys_on_demand,dedups_in_visit_order}_2872` (`eval.rs`), closed/reading rows for computed keys in folds (`tests/jq_closed_term_tests.rs`, negative-tested against the un-fixed `walk.rs`)
- [x] Stack depth: patterns 256 deep (the parser's cap) and 100,000 entries wide in value, path and fold position, debug build
- [x] Interleaved A/B above, output-identity gated
- [x] Patch coverage 98.09% (1077/1098; CI sticky comment, two rounds of rows added for the walker's mixed shapes, the path-mode folds' `break`/`halt` arms and the owned-identity route). Left uncovered on purpose: the `tolerate-line` `unreachable!` arms, `fold_walk_refusal_retries`' uncatchable guard (needs a resource-limit raise inside a key) and the `foreach` EXTRACT budget charge inside the per-branch sink (needs ~33k fold steps; pre-existing logic relocated)


newhoggy:
Two open readability issues touch code this PR rewrote, noted rather than folded in: issue 3113 (`bind_fold_alternative`'s struct/params vs its two callers) — `bind_fold_alternative` is now `each_fold_alternative`, same `FoldAlternative` shape (`walked` still unused by `resolve_reduce`), so that issue's points stand as filed; issue 3114 (`resolve_foreach`'s per-alternative nesting) — the per-branch sink closure adds one more level there, which makes that extraction more worthwhile, not less.

Review was `/code-review 3115 high` (findings and fixes in the PR body). If a deeper pass is wanted, `/code-review ultra 3115` needs a human to trigger it; not blocking merge on it.

newhoggy:
Re-triggering CI: the synchronize event for 60f8ed0 never produced a workflow run (other PRs' pushes did, minutes later). Reopening immediately.

PR #3118, opened 2026-09-18
fix(jq): give an identity-passthrough bind a position for getpath to compose from

### LINKED ISSUE #2978: jq: an identity-passthrough bind (`. as $x`) carries no position, so getpath can't compose one from it (#2896 residue)
**Severity: Low** (refuse-only — succinctly refuses where jq answers; never the write-side hazard direction)

## Summary

Residue of #2896, which gave `getpath` its register transparency. `getpath`'s result now
carries its own absolute position (`Snapshot::At`), composed from the *input's* proven
position plus the keys it navigated. That composition needs the input's position to be
provable, and it is — for a navigated bind (`.a as $y`), for a fold accumulator seeded from a
trackable INIT/UPDATE branch, and for another `getpath`'s own result.

It is **not** provable when the input is a variable bound by an *identity-passthrough* source
(`. as $x`), which is the most idiomatic spelling of the shape.

## Repro

```console
$ echo '{"a":{"b":2}}' | jq            -c 'path(. as $x | .a | $x | getpath(["a"]) | .b)'
["a","b"]
$ echo '{"a":{"b":2}}' | succinctly jq -c 'path(. as $x | .a | $x | getpath(["a"]) | .b)'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "b" of {"b":2}
```

The same shape headed by a *navigated* bind already works after #2896, which is what isolates
the cause to the bind's origin and not to `getpath`:

```console
$ echo '{"a":1,"b":2}' | succinctly jq -c 'path(.a as $y | .a | { b: 9 } | getpath(["b"]) | $y)'
["a"]
```

## Root cause

`substitute_bound_var_at` (`src/jq/eval.rs`) deliberately forces `Origin::Snapshot` for an
identity-passthrough bind source — the #844 value-equality witness — regardless of any
`Origin::At` the resolver computed:

```rust
let origin = if is_identity_passthrough(bind_expr) {
    Origin::Snapshot
} else if let Some(origin) = origin {
    origin
} ...
```

`Origin::Snapshot` carries no position (`Frame::certifies` admits it unconditionally, by value
equality), so `Snapshot::position()` answers `None` and `getpath_result_position` mints nothing.

Treating `Origin::Snapshot` as "position = the frame root" would be **wrong**, not merely
approximate: an identity-passthrough bind can sit at a non-root ambient *within one resolver
invocation* (`path(.a | (. as $x | ...))`), where the bound node is at `["a"]`, not `[]`. A
rule that assumed the root there can compose a position that certifies against a register it
is not actually on — the accept-where-jq-refuses direction, i.e. what `del()`/`=` then write
through. That is why #2896 declined it rather than guessing.

## Why this is separate from #2896

#2896 closed the mechanism (both arms of `_jq_path_append`) and the filed repro, read and
write, with zero fabrications added over 18,000 fuzzed programs. This is a *provenance*
question about one bind form, one layer below `getpath`: it is the same gap for any future
consumer of positional provenance, not a `getpath` gap.

## Suggested fix direction

Give an identity-passthrough bind a position **without narrowing** `Origin::Snapshot`'s own
acceptance — the two are not the same rule and swapping one for the other would regress #844
shapes (`Origin::Snapshot` certifies anywhere by value equality; `Origin::At` certifies only
at one position). Options worth costing:

- Carry both witnesses on the marker (a position *and* the value-equality flag), so
  `Frame::certifies` keeps today's answer while `Snapshot::position()` gains one. Note #2896's
  triage explicitly flags "forcing `Snapshot` into a two-axis provenance type threaded as a new
  ambient through `resolve_node`'s ~20 recursive call sites" as the point at which this stops
  being an Opus-tier change.
- Or mint `Origin::At` only when the bind is provably at the invocation root, leaving the
  nested case on `Origin::Snapshot`. Cheaper, and covers the idiomatic spelling; needs a
  reliable "am I at the invocation root" test at the bind site.

Either way the acceptance test is #2896's: `scripts/jq-bind-origin-fuzz.py` (whose alphabet
already has the `getpath`-stage pool) at **zero fabricate / zero mismatch**, attributed
against a baseline binary, plus `scripts/jq-bind-origin-oracle-sweep.sh`.

Recorded in `docs/compliance/jq/limitations.md` under the #2896 entry.

Refs #2896, #2042, #844.

newhoggy:
## Summary

Fixes #2978. An identity-passthrough bind (`. as $x`) made inside a `path()`/`del()`/assignment resolution was forced to `Origin::Snapshot` — the #844 value-equality witness, which carries no position — so #2896's `getpath` composition had nothing to start from and refused where jq answers (read and write):

```console
$ echo '{"a":{"b":2}}' | jq -c 'path(. as $x | .a | $x | getpath(["a"]) | .b)'
["a","b"]
$ echo '{"a":{"b":2}}' | succinctly jq -c 'path(. as $x | .a | $x | getpath(["a"]) | .b)'   # main
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "b" of {"b":2}
```

**Mechanism.** A fourth `Origin` variant, `SnapshotAt { invocation, path }`, certified *exactly* like `Snapshot` (value equality, unconditionally — nothing #844 accepts is narrowed) but also carrying the register's absolute path at the bind, read only by the positional composers. `identity_bind_position` mints it (jq mode only) at the resolver's two bind sites, recursing over exactly `is_identity_passthrough`'s grammar:

| source | position |
|---|---|
| `.` | `Frame::at`, while the branch is trackable (the #2042 invariant) — so a bind below the root is at its real position, never assumed `[]` |
| `$p` (marker) | the marker's own position, never the frame's |
| `if c then A else B end` | only when both arms agree |
| `try A catch B`, `A // B` | `A`'s |

`Snapshot::position` now returns `(invocation, path)` for `At` and `SnapshotAt`; `getpath_preserves_register` compares positions via the new `Frame::names` (not `certifies`, whose value rule admits a `SnapshotAt` from anywhere); `marker_needs_demotion` (#2642) matches both value-certified origins so a positioned marker cannot skip the rebuilt-root demotion.

**Oracle matrix** (all captured live from `/usr/bin/jq` 1.7.1; all 16 accept rows refused on `main`, every trap row still refuses):
- accept: the repro read/`del`/`=`, below-root bind (`path(.x | . as $v | .a | $v | getpath(["a"]) | .b)` → `["x","a","b"]`), equal-valued sibling subtree, marker source, nested marker source, `//`/`try`/`if` sources, `?//` bare alternative, two `getpath`s composing, `reduce`/`foreach` seeded from the marker.
- must-refuse traps: sibling `getpath(["c"])` (equal and unequal), "identity bind is at the root" (`path(.a | . as $x | .c | $x | getpath(["a","c"]) | .b)` + its `del`), "marker source binds at the frame's position", a constructed `.`, `tojson|fromjson`, `(. // {..})` on a `null` register.

**Fuzz.** `scripts/jq-bind-origin-fuzz.py`'s alphabet was widened first (a navigation prefix before the first bind — the below-root hazard was unreachable before — and a parenthesised nested bind of the marker). 3 seeds × 6000 programs: **0 fabricate / 0 mismatch**, attributed against a binary built from `b0e648611`. The same seed on the baseline shows the target shapes as refuse-only, so the zero has discriminating power. Two `fabricate-baseline` rows surfaced (pre-existing, byte-identical on the baseline, neither involves `getpath` or an identity bind) — filed separately as a follow-up.

**Review (`/code-review`, high, run twice — before and after rebasing over #3036):** two findings of one class, each fixed as its own commit — the `try` arm minted `.`'s position for a source whose `if` *condition* raises, so the catch body's value (`$p`, frozen elsewhere) carried it and `del()` wrote through (`path(. as $p | .a | (try (if error("x") then . else . end) catch $p) as $x | .b | $x | getpath(["b"]) | .c)` → `["a","b","c"]`). The second review found the `//` twin in the *empty* direction (`((if empty then . else . end) // $p) as $x` — the left side yields nothing, so the right side runs on a live register). Both the `try` and `//` arms are now gated on `yields_exactly_once` (`.`, a marker, or a `try` of one); both traps pinned in the unit negative controls, the sweep (4 rows) and the fuzz pool; sweep/fuzz re-run clean. Follow-up #3124 filed for the two pre-existing `fabricate-baseline` rows.

**Still refusing** (safe direction, both pinned in `test_identity_bind_position_residuals_still_refuse_2978` and `limitations.md`): an `if` source whose arms sit at different positions; a marker used inside a nested `path()`; a navigated source headed by the marker used off its position (`$x.a as $w`).

## Test plan

- [x] `cargo test --features cli,simd,regex,serde` (full suite, under the stall guard) — green
- [x] `cargo fmt --check`
- [x] `scripts/jq-bind-origin-oracle-sweep.sh` on the release build: agree=181 fabricate=0 mismatch=0 refuse-only=27 refuse-only-NEW=0 (27 new `identity-bind-*` rows, 2 registered residuals)
- [x] `scripts/jq-bind-origin-fuzz.py` seeds 1–3 × 6000 with `--baseline`: 0 fabricate / 0 mismatch (re-run after the review fix)
- [x] Full gate re-run after rebasing onto `bfee1fb1f` (#2964 and #3036 — the latter touches the same `Origin`/demotion machinery; three both-sides-append conflicts in the scripts and CLI tests resolved by keeping both). Sweep after rebase: agree=222 fabricate=0 mismatch=0 refuse-only-NEW=0. Fuzz 3×6000 against a baseline rebuilt at the new merge-base: 0 fabricate / 0 mismatch (refuse-only ≈300 is `main`'s own — 298 on the merge-base binary for seed 1, 285 here)
- [x] Every expectation in the new unit/CLI tests diffed directly against `/usr/bin/jq` 1.7.1 (stdout, stderr, exit)
- [x] `cargo llvm-cov --features cli,simd,regex,serde --workspace`: 321 instrumented added lines in `eval.rs`/`expr.rs`, 0 uncovered
- [ ] CI perf-guard: one `Rc::clone` per identity bind inside a resolver invocation is the only new hot-path cost (`(.users[] | . as $r | .score) |= 1` passes through it); A/B only if the guard flags


newhoggy:
Superseded by #3121 (agent `Oscar` claimed #2978 three seconds ahead of this branch under a different slug, so the branch lock never collided; #3121 merged while this PR was in CI). Same design (`Origin::SnapshotAt`, minted at the two resolver bind sites from `Frame::at`, `matches!` in `marker_needs_demotion`, `Snapshot::position` as `(invocation, path)`), same oracle matrix, same fuzz widening.

One thing this branch closed that `main` still has open, and which turns out to predate #2978 entirely: `A // B` counts as an identity passthrough on the strength of `A` alone, but `//` runs `B` whenever `A` yields *nothing*, not only on null/false — an `if` whose condition is `empty` does that. `main` (`40e86d32f`) and the pre-#2978 baseline both write through it: `del(.a | ((if empty then . else . end) // {"b":1}) as $x | .k | $x | .b)` on `{"a":{"k":{"b":1}}}` → `{"a":{"k":{}}}` where jq 1.7.1 refuses. Filed separately (see the issue linked from #2978).

Issue #3117, opened 2026-09-18
jq: path-context Skip's count arm has the same eager-drain gap #2916 fixed for If/Limit


newhoggy:
**Severity: Low**

## Summary

`path_context_step_generic`'s `Expr::Builtin(Builtin::Skip(n, expr))` arm (`src/jq/eval_generic.rs`) drains its count generator `n` eagerly via `path_context_component_values`, the same bug shape #2916 fixed for the sibling `Expr::If`/`Expr::Limit` arms (and `path_context_step_computed_index`/`path_context_step_computed_slice`). This arm postdates #2916 (added by #2968, `skip(n; body)`'s own path-context support) so it was never in scope there.

## Repro

Pinned jq 1.7.1 has no native `skip/2` (per #2952's own note); checked the same way #2952
does, by prepending the def jq's own docs give for it:

```console
$ DEF='def skip($n; expr): if $n > 0 then foreach expr as $item ($n; .-1; if . < 0 then $item else empty end) elif $n == 0 then expr else error("skip doesn'"'"'t support negative count") end;'
$ echo '{}' | jq -c "$DEF"' skip((1,(debug("late")|2)); error("boom"))'
jq: error (at <stdin>:1): boom
$ echo '{}' | succinctly jq -c '(skip((1,(debug("late")|2)); error("boom"))) | key'
["DEBUG:","late"]
jq: error (at <stdin>:1): boom
```

`succinctly` prints the `DEBUG` line even though the first count value (`1`) already fails inside `error("boom")`'s body, which the def-polyfilled count generator never reaches.

## Root cause

`src/jq/eval_generic.rs`, the `Expr::Builtin(Builtin::Skip(n, expr))` arm of `path_context_step_generic` (see the arm's own comment, added when #2916 landed, cross-referencing this gap):

```rust
Expr::Builtin(Builtin::Skip(n, expr)) => {
    let produced_from = out.len();
    let (counts, control) = path_context_component_values::<S, V>(n, pos);
    for n_value in counts {
        ...
    }
    ...
}
```

## Suggested fix direction

[...]

Issue #2470, opened 2026-09-05
yq: an assignment RHS reading key through an absent path inside with_entries leaves the value unchanged in real yq, succinctly writes the key


newhoggy:
Found during spine 2416's door-closing step (2026-09-05); pre-existing, pinned as-is there. Captured from yq v4.53.3 (`-o=json -I0`) on `a:\n  b: 1\n  e: 2\nc: [10, 20]\n`:

```console
$ yq '.a | with_entries(.value = (.a.b | key))'      # {"b":1,"e":2}   values untouched
$ yq '.a | with_entries(.value = "z")'               # {"b":"z","e":"z"}
$ yq '.a | with_entries(.value = (.key | length))'   # {"b":1,"e":1}
$ succinctly yq '.a | with_entries(.value = (.a.b | key))'   # {"b":"b","e":"b"}
```

Inside `with_entries`, each entry is `{key, value}`, so `.a.b` is an absent path under it. succinctly answers `key` there as `"b"` (the absent position's last component, ADR-0021 decision 6) and writes it; real yq leaves the value untouched, which suggests its `=` sees an absent RHS as no write, or evaluates the RHS against a context where the absent read yields nothing. Capture the mechanism before deciding: `.a | with_entries(.value = (.zzz | key))`, `.value = (.a.b | key)` outside `with_entries` on the same entry shape, `.a.x = (.a.y | key)` at the top level, and the `|=` form. Decide under ADR-0018's order: no rule-4 condition is obvious (no data loss on the reference side), so fidelity would say match, once the rule is understood.

newhoggy:
## Capture: `key`/`path` on an absent read inside an assignment RHS (#2470)

Oracle: Homebrew `yq` v4.53.3 (confirmed via `yq --version`). Comparison: `succinctly yq` (release
binary, spine 2416 worktree). Both run with `-o=json -I0`. Document `d.yaml`:

```yaml
a:
  b: 1
  e: 2
c: [10, 20]
```

Second document `e.yaml` (entry-shaped, for the "outside `with_entries`" control):

```yaml
key: b
value: 1
```

All runs below produced no stderr and exit 0 on both sides unless noted.

### Required matrix

| # | Filter | yq output | succinctly output | Verdict |
|---|---|---|---|---|
| 1 | `.a \| with_entries(.value = (.a.b \| key))` | `{"b":1,"e":2}` | `{"b":"b","e":"b"}` | DIFF |
[...]

PR #3128, opened 2026-09-18
fix(jq): pull four path-context sites' generators lazily

### LINKED ISSUE #2916: jq: three more path-context sites (.E[K], .E[S:T], if, limit) drain their argument generator eagerly, same bug class as #2259
**Severity: Low**

## Summary

Found while implementing #2259's fix (`getpath`'s path-context argument generator was
eagerly drained instead of pulled lazily, so a later path's side effects fired even when an
earlier path had already failed to navigate). #2259 converted exactly one call site
(`path_context_step_getpath`) to pull its argument generator lazily through the new
`path_context_component_each` sink. Three sibling call sites of
`path_context_component_values` (the pre-existing *eager* collector, `src/jq/eval_generic.rs`)
have the identical bug shape and were deliberately left unconverted, to keep #2259's fix
scoped to the literal issue reported rather than growing into a larger refactor.

**Edit:** the repro commands below were corrected -- the original version used `| keys`
(plural, a real jq builtin needing no path-context routing at all) instead of `| key`
(singular, succinctly's own extension, which is what actually forces `path_context_step_*`
to run). With `| keys`, none of these repro; with the correct `| key`, all four still do,
confirmed against the current `main` tip (which has since picked up unrelated `path_context`
laziness fixes for #2694's `reduce`/`foreach` SOURCE handling -- a different mechanism that
does not touch these four sites).

All four confirmed live against `/usr/bin/jq` 1.7.1 (using the bare filter, since jq has no
`key/0` of its own) plus succinctly's own `| key`-forced path-context arm, using the same
`debug("late")`-in-a-later-comma-alternative repro shape #2259's own tests use:

## 1. `path_context_step_computed_index` (`.E[K] | key`) -- key stream

```console
$ echo '{"a":5}' | jq -c '.a[(1,(debug("late")|2))]'
$ echo '{"a":5}' | succinctly jq -c '.a[(1,(debug("late")|2))] | key'
["DEBUG:","late"]
```

Note: this function is *already* named in #2259's own follow-up (#2914), but for a
different, deeper mechanism (the bracket's target is evaluated once instead of re-evaluated
per key output). This eager-drain angle on the key stream itself is a separate, narrower
issue from #2914's, not covered by it.

## 2. `path_context_step_computed_slice` (`.E[S:T] | key`) -- start/end bound streams

```console
$ echo '{"a":5}' | jq -c '.a[(1,(debug("late")|2)):3]'
$ echo '{"a":5}' | succinctly jq -c '.a[(1,(debug("late")|2)):3] | key'
["DEBUG:","late"]
```

## 3. `Expr::If`'s condition arm (`path_context_step_generic`)

```console
$ echo '{}' | jq -c 'if (true, (debug("late")|false)) then error("boom") else empty end'
$ echo '{}' | succinctly jq -c '(if (true, (debug("late")|false)) then error("boom") else empty end) | key'
["DEBUG:","late"]
```

## 4. `Expr::Limit`'s count argument arm (`path_context_step_generic`)

```console
$ echo '{}' | jq -c 'limit((1,(debug("late")|"bad")); error("boom"))'
$ echo '{}' | succinctly jq -c '(limit((1,(debug("late")|"bad")); error("boom"))) | key'
["DEBUG:","late"]
```

## Root cause / mechanism

All four sites call `path_context_component_values` (the eager collector) to get every
value of a generator-shaped sub-expression before running the per-value work that can fail
(`.[K]`'s navigation, a slice bound's use, an `if`'s branch, `limit`'s body). Real jq pulls
each such generator lazily, one output at a time, and stops asking for the next output once
a downstream failure occurs. #2259 already built the right primitive for this
(`path_context_component_each`, a `Demand`-driven sink over the same underlying generator);
what's left is rewiring each of these four call sites to walk their per-value work inside
the sink and answer `Demand::Stop` on failure, the same shape #2259's own
`path_context_getpath_walk_one`/`path_context_step_getpath` split demonstrates.

## Why this needs its own issue rather than folding into #2259

Each site has its own per-value work shape (`Expr::If`'s two-branch dispatch,
`Expr::Limit`'s truncation-and-early-stop bookkeeping already present in
`path_context_step_bounded`, the slice's two-bound interaction, `.E[K]`'s existing
key-outer/target-inner ordering rule) needing its own careful oracle-verified conversion and
test matrix -- doing all four as a drive-by inside #2259 would have risked an unverified
regression in one of them. `path_context_step_computed_index` additionally interacts with
#2914's separate target-re-evaluation-order finding, so its full fix likely wants to land
together with (or after) that one.

## Suggested fix direction

For each site, follow #2259's own pattern: split the "one value's worth of work" into its
own function (as `path_context_getpath_walk_one` did for `getpath`), then drive it from
`path_context_component_each`'s sink instead of the eager
`path_context_component_values`+`for` loop, answering `Demand::Stop` on the first failure.

newhoggy:
## Summary

Fixes #2916. Four call sites in `path_context_step_generic`/`eval_generic.rs` drained their argument generator eagerly via `path_context_component_values` before doing any per-value work, so a later value's own side effects (`debug`, `halt_error`, ...) fired even when an earlier value had already failed: `.E[K]`'s key stream, `.E[S:T]`'s start/end bound streams, `if`'s condition arm, and `limit`'s count arm. Real jq pulls each of these lazily, one value at a time, stopping the generator as soon as a downstream failure occurs. #2259 already built the right primitive for this (`path_context_component_each`, a `Demand`-driven sink) for `getpath`; this rewires the four sibling sites onto it, following that same pattern.

`path_context_step_computed_slice` needed one more step beyond the other three: confirmed live against jq 1.7.1 that the end bound is re-evaluated fresh *per start value*, not once up front (`.a[(0,1):(debug("end")|3)]` prints its debug line twice) -- matching `S as $s | T as $t | E | .[$s:$t]`'s real nesting. Added `path_context_component_each_bound` to give a possibly-absent bound the same one-value-generator treatment.

## A second bug this surfaced

Rewiring `path_context_step_computed_index`'s key stream onto the lazy sink surfaced a second, independent bug, confirmed live: yq's read-only-operand scope (`(.zzz | key) + 1` reading a miss as "nothing" rather than `null`, #2470) is ambient on a thread-local, and `read_only_operand_strategy_generic`'s own sink-suspension rule (entered around every demand-driven producer's `sink.push`, on the theory that a consumer is downstream of the operand and must not inherit its read-only scope) fired around this sink too -- but the navigation this sink performs *is* the operand's own remaining work, not a downstream consumer of `key`'s output, so it must not run suspended.

`(.a[("z"+"z")] | key) + "!"` on `{"a":{"b":1}}` answered `"zz!"` instead of yq's own `"!"` until fixed by capturing the ambient read-only state before pulling the key stream and re-entering it inside the sink, around the navigation. This exact scenario is already pinned by the pre-existing `test_computed_bracket_head_is_walkable_2471` suite -- it failed on the first draft of this fix, caught it, and now passes. Applied the same defensive re-entry to the other three sites too, since the mechanism is identical wherever a lazily-pulled generator feeds navigation-sensitive work back into the same sink (no live repro found for those three, but the risk is structurally identical).

## Follow-up filed

`Expr::Builtin(Builtin::Skip)`'s count arm has the identical eager-drain shape but postdates this issue (added by #2968) so was out of scope here -- filed as #3117.

## Test plan

- [x] All four sites' laziness confirmed live against `/usr/bin/jq` 1.7.1 (no `DEBUG` output on an unreachable later value)
- [x] The slice site's fresh-per-start end-bound re-evaluation confirmed live via `path(...)`
- [x] New test `test_path_context_sibling_sites_pull_their_generators_lazily_2916` pins all four sites' laziness plus successful multi-output ordering
- [x] Pre-existing `test_computed_bracket_head_is_walkable_2471` (70 assertions) still passes, including the row that caught the read-only regression
- [x] `cargo test --features cli,simd,regex,serde` -- full suite, 0 failures
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` -- clean
- [x] `cargo build --no-default-features` -- clean (no_std)


newhoggy:
Addressed the code-review findings on the first draft (two real correctness bugs, plus a test-quality gap where the "successful shape" cases never actually exercised the changed code):

1. **Target-before-key/bound ordering regression**: `path_context_step_computed_index`/`path_context_step_computed_slice` stepped `target` unconditionally before pulling the key/bound generator at all, reversing jq's own `K as $k | E | .[$k]` nesting (`E`'s side effects fired before `K`'s). Fixed by memoizing `target`'s step on the *first* key/bound value the sink actually receives -- confirmed live against jq 1.7.1, which also never touches `target` at all when the key/bound generator produces zero values (`(.a|debug("E"))[(empty)]` never prints `E`'s debug).

2. **yq-mode truncation bypass**: the slice's end-bound generator's own escape was folded into the same slot as a real navigation/target failure and returned raw, bypassing `path_context_component_escape`'s yq-mode truncation every sibling site applies. Confirmed live: `[try (.a[(0,1):(1,error("boom"))] | key) catch "c"]` on yq mode answered `["a","c"]` instead of yq's own `["c"]`. Fixed by giving it its own escape slot, checked and truncated separately.

3. **Test quality**: the new test's multi-output success cases used plain value filters that never routed through `path_context_step_generic` at all -- confirmed by a reviewer instrumenting the function with a temporary probe. Rewrote all four to wrap in `path(...)`, verified against jq 1.7.1.

Re-verified all original repros plus these three fixes live; full suite and clippy clean.

Issue #3124, opened 2026-09-18
jq: a `?//` fold alternative's bare variable is certified by value inside a nested `path()`, so the alternative never errors and the retry never runs (exit 0, nothing emitted, where jq raises)


newhoggy:
**Severity: Low** (both refuse-side twins already refuse; the read-side symptom is exit 0 with nothing emitted where jq raises)

## Summary

A shape where succinctly exits 0 and prints nothing while jq 1.7.1 raises, found by the widened
`scripts/jq-bind-origin-fuzz.py` alphabet in #2978's run and attributed to the baseline binary
(`b0e648611`, pre-#2978) byte-for-byte — not #2978's. "Silent empty vs error", never a written
path, but it hides a resolver decision jq makes loudly. (A second shape originally filed here —
a refused pattern step on an untracked stage swallowed by the body's `?` — is #3120, filed
independently from the same fuzz run by the #3121 branch; dropped from this issue.)

## Repro

`/usr/bin/jq` 1.7.1 vs `succinctly jq -c` at `b0e648611` (identical on the #2978 branch).

**B. A `?//` fold pattern's first alternative certifies a navigated bind by value inside a
nested `path()`.**

```console
$ echo '{"a":"s","c":"s"}' | jq -c 'path(foreach .c as $v0 ?// {a:$v0} (.; path(.a | $v0); empty))'
jq: error (at <stdin>:0): Cannot index string with string "a"
$ echo '{"a":"s","c":"s"}' | succinctly jq -c 'path(foreach .c as $v0 ?// {a:$v0} (.; path(.a | $v0); empty))'
$ echo $?
0
```

jq: alternative 1 binds `$v0` to `.c`'s node, the inner `path(.a | $v0)` refuses (equal value,
different node), `?//` retries alternative 2 (`{a:$v0}` on `"s"`), which raises `Cannot index
string`. succinctly: the inner `path(.a | $v0)` *answers* `["a"]` — swap `empty` for `.` and the
outer `path()` reports `Invalid path expression with result ["a"]` — so alternative 1 never
errors and alternative 2 never runs. The plain fold (`foreach .c as $v0 (.; path(.a | $v0); .)`)
and the plain bind (`.c as $v0 | path(.a | $v0)`) both refuse correctly; only the `?//` seeding
path (`resolve_foreach`'s destructuring case → `apply_pattern_bindings`) hands `$v0` a marker the
[...]

Issue #3119, opened 2026-09-18
jq: a wrapped identity source (`(try . catch 1) as {a:$v} ?// $v`) fails the pattern's first step and retries onto the wrong alternative


newhoggy:
**Severity: High** (a `del()` through a `?//` chain deletes the whole document where jq deletes one key — silent wrong write; the other rows are refuse-only or wrong-answer)

## Summary

`resolve_as_pattern`'s first-step identity test (jq's `path_intact` at the pattern's first
`INDEX`) recognises only a *bare* `.` head as "the source value is the register's own node".
Every other spelling `is_identity_passthrough` recognises — `(try . catch 1)`,
`(if true then . else . end)`, `(. // 1)` — evaluates to the very same `jv` in jq, so jq's
first step succeeds; succinctly falls to `null_bool_identical`, decides the step is *not*
intact, and either refuses (single pattern) or **retries onto the next `?//` alternative**,
binding a different alternative than jq does — which `del`/`=` then write through.

Found by `scripts/jq-bind-origin-fuzz.py` once #2978 added the identity spellings to its
`SOURCES` pool; every row below reproduces byte-for-byte on `main` (`b0e648611`, the #2978
merge-base), so it is not #2978's.

## Repro

`/usr/bin/jq` 1.7.1 vs succinctly `main`, input `{"a":"s","c":"s"}`:

| filter                                                  | jq 1.7.1     | succinctly            |
|---------------------------------------------------------|--------------|-----------------------|
| `path((try . catch 1) as {a:$v} \| $v)`                 | `["a"]`      | refuses (exit 5)      |
| `path((. // 1) as {a:$v} ?// $v \| $v)`                 | `["a"]`      | `[]`                  |
| `del((. // 1) as {a:$v} ?// $v \| $v)`                  | `{"c":"s"}`  | `null` (**wrong write**) |
| `path((try . catch 1) as {a:$v} ?// $v \| ($v \| .[]?))` | (nothing)   | `["a"]`, `["c"]`      |
| `path(. as {a:$v} ?// $v \| ($v \| .[]?))` (control)     | (nothing)   | (nothing)             |
| `del(. as {a:$v} ?// $v \| $v)` (control)                | `{"c":"s"}` | `{"c":"s"}`           |

The two controls show the bare-`.` head is fine; only the wrapped spellings diverge.

## Root cause

`src/jq/eval.rs`, `resolve_as_pattern`:

```rust
let identical = trackable
    && match head {
[...]

PR #3130, opened 2026-09-18
fix(jq): check a destructuring bind on an untracked stage against the carried register (#3043, #3120)

### LINKED ISSUE #3043: jq: a destructuring `as` on an untracked `null` stage fabricates where jq refuses (`path(null | . as [$v] | empty)`)
**Severity: High.** This is a fabrication: succinctly answers where jq refuses, and `del()`/`|=`/`=` then succeed as silent no-ops where jq exits 5.

## Summary

Split out of #2979's triage (family C there). It showed up in `scripts/jq-bind-origin-fuzz.py` at about 1 hit per 6000 programs on seeds 20260915, 777 and 31337. A **single** destructuring `as` (no `?//`, no fold) whose stage input is a `null` that the resolver no longer tracks passes its pattern steps where jq refuses at the first one. The binding then yields nothing that the terminal `path()` check could refuse.

## Repro (main `b509a8019`, jq 1.7.1)

```console
$ echo '{"a":1}' | jq 'path(null | . as [$v] | empty)'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 0 of null   # exit 5
$ echo '{"a":1}' | succinctly jq 'path(null | . as [$v] | empty)'
                                                                                             # exit 0, no output
$ echo '{"a":1}' | succinctly jq 'del(null | . as [$v] | empty)'
{"a":1}                                                                                      # exit 0; jq exits 5
```

| Filter                                   | Input        | jq 1.7.1                                         | succinctly                                     |
|------------------------------------------|--------------|--------------------------------------------------|------------------------------------------------|
| `path(null \| . as [$v] \| empty)`         | `{"a":1}`    | exit 5, near attempt to access element 0 of null | **exit 0**                                     |
| `path(null \| . as {a:{b:$v}} \| empty)`   | `{"a":1}`    | exit 5, near … element "a" of null               | **exit 0**                                     |
| `path(.a \| null \| . as [$v] \| empty)`    | `{"a":1}`    | exit 5                                           | **exit 0**                                     |
| `del(null \| . as [$v] \| empty)`          | `{"a":1}`    | exit 5                                           | **exit 0**, `{"a":1}`                          |
| `(null \| . as [$v] \| empty) \|= 5`        | `{"a":1}`    | exit 5                                           | **exit 0**, `{"a":1}`                          |
| `path(null \| . as [$v] \| $v)`            | `{"a":1}`    | exit 5, near … element 0 of null                 | exit 5, "with result null" (wording only)      |
| `path(true \| . as [$v] \| empty)`         | `{"a":1}`    | exit 5, near … element 0 of true                 | exit 5, "Cannot index boolean" (wording only)  |
| `path(null \| . as [$v] \| empty)`         | `null`       | exit 0                                           | exit 0 (correct: the register is `null`)       |
| `path(.a \| null \| . as [$v] \| empty)`    | `{"a":null}` | exit 0                                           | exit 0 (correct)                               |
| `path(5 \| . as [$v] \| empty)`            | `{"a":1}`    | exit 5                                           | exit 5, same wording (not affected)            |
| `path(1 \| null as [$v] \| empty)`         | `{"a":1}`    | exit 5                                           | exit 5, same wording (not affected)            |
| `path(foreach null as [$v] (0; .; empty))` | `{"a":1}`    | exit 5                                           | exit 5, same wording (the fold walk is fine)   |

The affected rows need both a **`null` stage input** and a `null` bound value. `5 | . as [$v]` and `1 | null as [$v]` refuse correctly. A `true` stage passes the same identity check, but the walk then hits a genuine "Cannot index boolean" error, so it refuses with different wording and nothing is written.

## Root cause

In `resolve_as_pattern` (`src/jq/eval.rs`), each alternative's walk is seeded with

```rust
let seed = PatternRegister { path: PathPrefix::root(), value: value.clone(), is_input: identical };
```

`value` is the **stage's** input, not jq's register. On a trackable stage the two are the same node. On an untracked stage (after `null`, `5`, or any computed value) jq's register sits wherever the last navigation left it: the document root, or `.a`. `walk_pattern_step` admits a step when `reg.is_input || null_bool_identical(input, &reg.value)`, so a `null` bound value compares equal to the seed's copy of the `null` stage value. That makes it "identical to the register", when jq compares it against the real register (`{"a":1}`) and refuses.

The `Expr::AsPattern` dispatch guard (`trackable || patterns.len() == 1`) is what lets a single pattern reach `resolve_as_pattern` on an untracked stage at all. #2649's doc comment accepts that for the "refuses at the first step" outcome, which holds for every value except `null`/`bool`.

## Suggested fix

- When `!trackable`, seed the walk with **the register if it is known**, never with the stage's own value. `resolve_seq` carries one (`PathBranch::with_register`, #2046) in some shapes. Where none is known, the seed must not admit a `null`/`bool` coincidence, so the first step refuses.
- Refusing everywhere the register is unknown costs refuse-only rows where jq's register really is `null` (`path(.a | null | . as [$v] | empty)` on `{"a":null}`, jq exit 0). Close those through the carried register where it exists, and record the rest in `docs/compliance/jq/limitations.md` beside #2649's table.
- `Snapshot`/`TrackedVar` markers are a separate route. A marker-headed source (`$x as {a:$q}`) already takes the `frame.certifies` branch, so leave it unchanged.

## Tests

- Unit rows in `test_path_destructure_matrix_refuses_2649` (or a sibling matrix) for every table row, as jq literals, covering object and array patterns and nested ones.
- CLI rows in `tests/jq_cli_tests.rs` for `del`/`|=`: exit 5, document untouched.
- Fuzz: `./scripts/jq-bind-origin-fuzz.py -n 6000 --seed {20260915,777,31337} --baseline <merge-base binary>`. None of this family's hits may remain. (#2979's own hits will show as `fabricate-baseline` until it lands.)

## Class: Sonnet

The fix is local to one seed in `resolve_as_pattern`, every row is pinned against the oracle, and the fix only moves in the refusal direction. No resolver redesign is needed. The carried-register refinement is optional precision, not correctness.

Refs #2979, #2649, #2046.

### LINKED ISSUE #3120: jq: a refused pattern step on an untracked stage is swallowed by the body's `?` (`path(null | .a as {arr:[$v]} | .arr[]?)` exits 0)
**Severity: Medium** (accept-where-jq-refuses: exits 0 with nothing, or leaves a `|=` document untouched, where jq exits 5 — the write forms produce output jq never does, though never a *changed* document in the rows found)

## Summary

A destructuring bind on an *untracked* stage (after a literal) whose first pattern step jq
refuses is, by #2979's rule, not refused on the spot in succinctly — "off the register the
walk's verdict is a guess" — but deferred to the body. When the body's own navigation is
`?`-suppressed (`.arr[]?`, `first(.[]?)`), the deferred refusal is swallowed along with it:
succinctly emits nothing and exits 0 (or, under `|=`/`=`, prints the input document) where jq
exits 5 at the pattern step.

Found by `scripts/jq-bind-origin-fuzz.py` during #2978 (`fabricate-baseline` rows);
reproduces byte-for-byte on `main` (`b0e648611`).

## Repro

`/usr/bin/jq` 1.7.1 vs succinctly `main`, input `{"a":[1]}`:

| filter                                        | jq 1.7.1                                                                  | succinctly                    |
|-----------------------------------------------|---------------------------------------------------------------------------|-------------------------------|
| `path(null \| .a as {arr:[$v]} \| .arr[]?)`   | exit 5: `Invalid path expression near attempt to access element "arr" of null` | (nothing), **exit 0**    |
| `path(null \| .a as {arr:[$v]} \| .arr?)`     | exit 5, same message                                                      | exit 5, `... with result null` |
| `path(null \| .a as {arr:[$v]} \| .)`         | exit 5, same message                                                      | exit 5, `... with result null` |
| `path(null \| .a as {arr:[$v]} \| 5?)`        | exit 5, same message                                                      | exit 5, `... with result 5`   |

jq refuses *at the pattern step* in every row (the source value `null` is not the register,
which is still the document root after the literal — `jv_identical` on different kinds is
false, no guess involved). succinctly defers, and only the row whose body navigation is
`?`-suppressed loses the refusal entirely. Fuzz-found originals, all on `main`:

```
del(.x | .a as $v0 | null | (.a | ..) as {a:{b:$v1}} | .arr[]? | . as $q | . | path(.a | $v1) | null)
path(.arr[0:1]? as $v0 | .x.a | null | ([.a] | .[0]) as {arr:[$v1,$w1]} | (.b? as $z | $w1) | if true then . else 1 end | ($w1 | foreach .[]? as $i (.; getpath(["a"]); .b?)))
((try . catch 1) as $v0 | null | (if true then . else $v0 end) as {arr:[$v1,$w1]} | ($w1 | foreach .[]? as $i (.; getpath(["a"]); .b?))) = 9
```

## Root cause

`src/jq/eval.rs`, `resolve_as_pattern`: on `!trackable` the seed's `is_input` is `false` and a
walk error is propagated only when `is_resolver_refusal(&e)` — otherwise the body is resolved
by value from `resolve_seq_from_seed` with `PathBranch::passthrough(.., false, ..)`. The body's
`.arr[]?` then raises the deferred untracked-navigation error inside an `Optional`, whose
blanket arm prunes the branch (jq's `INDEX_OPT`), so nothing is left to refuse on.

The #2979 guess model is right that the walk cannot see the carried register on an untracked
stage — but when the source value's *kind* differs from every possible register value (a
`null`/number source against a document whose root is an object, as here), the refusal is
jq's verdict, not a guess, and could be raised eagerly. More generally, a deferred refusal
should not be prunable by a `?` that belongs to the body.

## Why this is separate from #2978

#2978 is the identity bind's *position* (`Origin::SnapshotAt`), consulted only by positional
readers; this is the #2979 untracked-stage pattern walk's refusal being swallowed by the
body's optional navigation. Different arm, predates #2978, and reproduces on the merge-base.

## Suggested fix direction

Either raise the walk's refusal eagerly when the source value cannot be the register at all
(kind mismatch with the ambient, or a literal stage whose register is provably the ambient
root), or carry the deferred refusal past the body's own `?`/`[]?` suppression (the
`Optional` arm prunes *navigation* failures, not a "this stage was never a path expression"
verdict — `invalid_path_expression` already has that special status in the `Try` arm and
should get it here). Add the rows above to `scripts/jq-bind-origin-oracle-sweep.sh` and rerun
the fuzz at zero fabricate.

Refs #2979, #2649, #2978.

newhoggy:
Fixes #3043, Fixes #3120
Refs #3124 (part A; part B stays open there), #2979, #2649, #2046

## Summary

A destructuring bind on an **untracked** stage seeded its pattern walk's register from the stage's *own* value, which is not jq's register at all once the stage has stepped off it. A `null` literal stage therefore passed `walk_pattern_step`'s null-identity clause against itself where jq refuses against the document root:

```console
$ echo '{"a":1}' | jq 'del(null | . as [$v] | empty)'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element 0 of null   # exit 5
$ echo '{"a":1}' | succinctly jq 'del(null | . as [$v] | empty)'     # main
{"a":1}                                                                                      # exit 0
```

(#3043's table; #3120/#3124-A are the same defect with a `?`-suppressed body — `path(null | .a as {arr:[$v]} | .arr[]?)` exited 0 with nothing.)

**Fix — hand the arm the register.** `resolve_seq_stage` is the one frame that holds the *carried* register (#2046), so it now dispatches an untracked `AsPattern` stage to `resolve_as_pattern` directly with `carried_register`, and the first step is checked against that (`PatternRegister` gains a `known` flag; with no register in hand — a nested pipe under `if`/`try`, a `catch` handler — the walk refuses unconditionally rather than guess). The outer placement trusts that route's trackability verdict (`place_step(step, step_certified)`) instead of applying "untracked is absorbing" to a step the arm has already certified. Every #3043 row now refuses with jq's exact message, and the same shapes accept exactly where jq does (`path(.a | null | . as {b:$v} | $v)` on `{"a":null}` → `["a","b"]`).

**Two families that were refuse-only for want of the register now answer** (this is the #2649 residue 2 / #2979 family B closure, five `REFUSE_ONLY` sweep entries removed):
- a marker that *is* the register at the head of the pattern: `path(. as $x | 5 | $x as {a:$q} ?// $z | $q)` → `["a"]`, `del(...)` → `{}` (the `alt-untracked-stage-marker-head` "MUST stay a refusal" pin was only because the arm couldn't see the register);
- a `?//` first-step refusal that is provably jq's own verdict (the source is not even value-equal to the register) retries as jq's fork does: `path(5 | 5 as {a:$v} ?// $v | empty)` → nothing, exit 0; `path(5 | . as {a:$v} ?// $v | .b?)` now refuses with jq's own `element "b" of 5`. An equal-valued non-null/bool source is still a guess (jq compares pointers) and keeps the no-retry rule.

**Residuals recorded** in `limitations.md` and the sweep: a `catch` handler (no register can be known — the try body may have navigated before raising; the old `["a"]` on a `null` document was a coincidence that wrote `del` → `{}` on `{"a":1}`), a nested pipe under `if`/`try` with a null/bool register behind a non-matching literal (refused before too), and — from the review — an opaque stage (`reduce`, a def call, `first`) dropping the carried register so the walk after it refuses without retrying. One pre-existing *discarded write* the review surfaced is filed as #3133: the same no-register refusal raised in a pipe nested under `try` is caught by it (`del(.a as $x | .a | 5 | try ($x as {b:$q} | $q))` echoes the document where jq writes `{"a":{}}`) — the nested-pipe register limitation, not this arm.

### Acceptance oracle
- `scripts/jq-bind-origin-oracle-sweep.sh`: 22 new `untracked-*` rows; `agree=247 fabricate=0 mismatch=0 refuse-only=31 refuse-only-NEW=0`. The script's two heredocs now use `read -r -d ''` instead of `$(cat <<EOF)` — bash scans a command substitution's body for quotes/parens even inside a quoted heredoc, and pruning an entry with an apostrophe broke the whole script (the #2978 round hit the same trap).
- `scripts/jq-bind-origin-fuzz.py`, 3 seeds × 6000 against a merge-base build: `fabricate=0 mismatch=0`, and the merge-base's own `fabricate-baseline` rows (this family) are gone; the remaining `mismatch-baseline` rows are #3119's.

### Review round (`/code-review 3130`, standard effort)

Three findings, all refuse-only/documentation (no fabrication or mismatch in the reviewer's own ~5000-row differential). Fixed: a bare-`$z` alternative reached by a `?//` retry resolved its body without the register, so `path(. as $x | 5 | $x as [$v] ?// $z | $z)` refused where jq answers `[]` (and `del(... | $z.a)` where jq writes `{}`) — its body is now seeded like the stepped arm's. Recorded: the opaque-stage register loss (refuse-only) and the `try`-swallowed discarded write (#3133, pre-existing). A second pass after rebasing onto #2872 (whose backtracking matcher restructured `resolve_as_pattern`, so the change was re-derived on it rather than spliced) found two more, both recorded: the body's nested `try`/`?` pipe has no register, so once the walk answers, `del(. as $x | 5 | $x as {a:$w} | try ($w | .b))` reaches the same silent discard the trackable twin already has on `main` — four rows pinned as a #3133 characterization test so that fix flips them visibly; and a later-step refusal after a marker-certified first step does not retry `?//` (refuse-only, sweep row). Suggest a human run `/code-review ultra` for the deepest pass if wanted; not a merge gate here.

### Tests
- `test_destructure_on_untracked_stage_checks_the_carried_register_3120` (eval.rs): #3043's table, #3120/#3124-A, the accept twins, the flipped residue rows, the guess/no-retry control, the nested-`if` pair.
- `test_as_pattern_in_catch_handler_refuses_without_a_register_3120` replaces `test_as_pattern_binding_without_marker_still_seeds_body_2649`, which pinned the `null`-document coincidence.
- CLI: `test_destructure_on_null_stage_refuses_like_jq_3120` (write forms, exit 5, nothing emitted, jq's message) and `test_destructure_on_untracked_stage_answers_with_the_register_3120`; two existing pins updated to jq's now-matching outputs.

## Test plan
- [x] Every row diffed against `/usr/bin/jq` 1.7.1 (unit matrix, CLI rows, sweep)
- [x] Sweep clean; fuzz 3 × 6000 at zero fabricate / zero mismatch with `--baseline`
- [x] `cargo llvm-cov --features cli,simd,regex,serde --workspace` — every instrumented added line in `eval.rs` covered except a test's own unreachable `panic!` arm
- [x] `/code-review 3130` twice (standard effort; the second after the #2872 rebase): one fix, residuals recorded, one follow-up filed (#3133); full verification, sweep and fuzz (3 × 6000 against a merge-base build) rerun after each


newhoggy:
Heads-up from `Miles` (#3115, merged at `492cc4bde`): the destructuring walk this PR edits was restructured on `main` — `walk_pattern`/`walk_pattern_step`/`ResolvedPatternKey`/`PatternStep` are gone. The path-mode pieces now live in:

- `PathPatternMode::step` (`src/jq/eval.rs`) — the per-step `path_intact` identity check (now ahead of the key's kind check) and the register move; `PatternRegister` is unchanged (`path`/`value`/`is_input`), so a `known` field slots in there and its check belongs in `step`.
- `each_pattern_walk` — the entry point `resolve_as_pattern` and `each_fold_alternative` (was `bind_fold_alternative`) call; it delivers one `(PatternRegister, &[PatternBinding])` per branch to a sink (a computed key forks), so a seed change is a change to what those two callers pass as `seed`.
- `fold_pattern_seed` is untouched; `bind_pattern_body` now takes a deduplicated `&[PatternBinding]` plus #2978's `identity_at`.
- The generic core (`walk_pattern_each`/`walk_pattern_once`/`PatternMode`) is mode-agnostic; nothing register-specific should need to go there.

The #2872 CLI pins (`test_pattern_computed_key_*_2872`) and `scripts/jq-pattern-fanout-fuzz.py --baseline` are the quickest way to check a rebased seed still fans out per branch.

PR #3142, opened 2026-09-18
fix(jq): require A // B's left to be raise-and-empty-free, not merely raise-free

### LINKED ISSUE #3129: jq: `A // B` counts as an identity passthrough when `A` yields nothing — a fresh right-side value is certified by value and written through
**Severity: High** (a fabricated path in the write direction: `del()`/`=` modify a node jq refuses to touch)

## Summary

`is_identity_passthrough` (`src/jq/eval.rs`) admits `A // B` on the strength of `A` alone, reasoning
that `B` only runs when `A` is null/false (where the value rule is harmless) or when `A` raises
(which `//` propagates). It misses the third case: `//` also runs `B` when `A` yields **nothing**.
An `if` inside `A` passes the passthrough grammar (both arms `.`) but its condition is
unconstrained, so `if empty then . else . end` yields zero outputs, `B` runs on a live register,
and `B`'s value — a fresh literal, or anything at all — is bound as an `Origin::Snapshot` marker.
`Frame::certifies` then admits it by value equality wherever an equal-valued node sits.

Found by the second `/code-review` pass on PR #3118 (the superseded twin of #3121), first as the
positional variant (`SnapshotAt` composing through `getpath`), then traced back: it reproduces
byte-for-byte on the pre-#2978 baseline (`b0e648611`), so it is #844's original `//` arm, not
#2978's.

## Repro

`/usr/bin/jq` 1.7.1 vs `main` `40e86d32f` (identical on `b0e648611`), input `{"a":{"k":{"b":1}}}`:

```console
$ jq -c         'del(.a | ((if empty then . else . end) // {"b":1}) as $x | .k | $x | .b)'
jq: error (at <stdin>:0): Invalid path expression near attempt to access element "b" of {"b":1}
$ succinctly jq -c 'del(.a | ((if empty then . else . end) // {"b":1}) as $x | .k | $x | .b)'
{"a":{"k":{}}}

$ jq -c         '.a | ((if empty then . else . end) // {"b":1}) as $x | .k | ($x.b) = 9'
jq: error (at <stdin>:0): Invalid path expression near attempt to access element "b" of {"b":1}
$ succinctly jq -c '.a | ((if empty then . else . end) // {"b":1}) as $x | .k | ($x.b) = 9'
{"b":9}

$ jq -c         'path(.a | ((if empty then . else . end) // {"b":1}) as $x | .k | $x | .b)'
jq: error ...
$ succinctly jq -c 'path(.a | ((if empty then . else . end) // {"b":1}) as $x | .k | $x | .b)'
["a","k","b"]
```

The positional twin on `main` (post-#3121), input `{"a":{"b":{"c":1}},"b":{"c":1}}`:

```console
$ jq -c         'del(. as $p | .a | ((if empty then . else . end) // $p) as $x | .b | $x | getpath(["b"]) | .c)'
jq: error (at <stdin>:1): Invalid path expression near attempt to access element "c" of {"c":1}
$ succinctly jq -c 'del(. as $p | .a | ((if empty then . else . end) // $p) as $x | .b | $x | getpath(["b"]) | .c)'
{"a":{"b":{}},"b":{"c":1}}
```

Same via `if (error("x")?) then . else . end` on the left. Controls that must keep agreeing:
`(. // 1) as $x` (P13 in #2978's matrix, `["a","b"]`), `((if true then . else . end) // 1)` is
currently accepted and would become refuse-only under the fix below — jq answers it, and that
is the safe direction.

## Root cause

`src/jq/eval.rs`, `is_identity_passthrough`'s `Expr::Alternative(left, _)` arm delegates to
`is_identity_passthrough(left)`, which admits an `If`. The doc comment on the arm states the raise
case is covered because `//` propagates errors (true) and concludes "an `if` on the left is fine
here" — the empty case is not considered. `identity_bind_position` (#2978/#3121) inherits the
same gap for the position it mints, since it recurses over the same grammar.

## Why this is separate from #2978 / #3121

#2978 was about giving an identity bind a *position*; this is the value-rule certification
(#844's `Origin::Snapshot`) admitting a value that is not `.` at all. It reproduces on the
pre-#2978 binary. #3121's review fix (`is_raise_free_identity_passthrough`) closed the `try`
direction and the `try`-then-`//` chain, but left a bare `if` on `//`'s left admitted.

## Suggested fix direction

- Make the `Alternative` arm require a left side that **yields `.` exactly once** — the same
  set `is_raise_free_identity_passthrough` already describes (`.`, a marker, a `try` of one),
  since that grammar excludes `if` (the only member that can yield zero outputs or raise).
  Concretely: `Expr::Alternative(left, _) => is_raise_free_identity_passthrough(left)`, and
  rename/document the helper as "yields exactly once" rather than "raise-free", since both
  properties are what `try` and `//` each need.
- `identity_bind_position`'s `Alternative` arm follows automatically if it keeps recursing over
  `is_identity_passthrough`'s grammar; check it does.
- Pin the three repros above (read, `del`, `=`) plus the positional twin as must-refuse rows in
  `scripts/jq-bind-origin-oracle-sweep.sh` and in the `#844`/`#2978` unit negative controls, and
  add `(((if empty then . else . end) // $v) as $w | $w | getpath(["a"]) | .b?)` to the fuzz
  `USES` pool (PR #3118's branch `issue-2978-identity-passthrough-bind-position` has all of these
  written, commit `468799f82`).

Refs #844, #2978, #3121, #3118.

newhoggy:
## Summary

`is_identity_passthrough`'s `Expr::Alternative(left, _)` arm admitted any `A` in
`A // B` that itself passed the wider identity-passthrough grammar, including an
`if`. But `//` runs `B` whenever `A` yields **nothing**, not only when `A` raises,
and an `if`'s condition is arbitrary enough to yield nothing on its own
(`if empty then . else . end`). `B`'s value was then certified as
`Origin::Snapshot` and written through wherever an equal-valued node existed —
`del()`/`=` fabricated a path jq refuses to touch.

Fix: require `A` to satisfy `is_raise_free_identity_passthrough` (the same
raise-and-empty-free grammar a `try` body already needs), not the wider
`is_identity_passthrough` grammar. `identity_bind_position`'s own `Alternative`
arm needs no change — it's only ever consulted once `is_identity_passthrough`
already holds for the bind source, so it's inert on shapes the narrower gate now
excludes.

Accepted, deliberate cost: `((if true then . else . end) // 1)` becomes
refuse-only under the fix even though jq itself would answer it — the safe
direction per this codebase's divergence rule, matching how a `try` body with an
`if` inside it already behaves (#2978).

## Test plan

- [x] `cargo fmt --check`
- [x] New regression test (`test_alternative_left_empty_if_is_not_an_identity_passthrough_3129`) pins all four repro forms from the issue (`path`/`del`/`=`, plus the positional twin) against `/usr/bin/jq` 1.7.1's own refusal message, and a control (`(. // 1) as $x` on a truthy `.`) confirming the base `A // B` rule is unchanged
- [x] Updated the pre-existing predicate table (`(if true then . else . end) // 1` flips `true` → `false`), verified live against jq 1.7.1 that jq still answers there while succinctly is now refuse-only (accepted per the issue)
- [x] `cargo llvm-cov` + `omni-dev coverage diff`: 100% patch coverage (24/24 new lines)
- [x] End-to-end verified against pinned oracle: `del(.a | ((if empty then . else . end) // {"b":1}) as $x | .k | $x | .b)` on `{"a":{"k":{"b":1}}}` now matches jq's refusal exactly, where it previously silently wrote `{"a":{"k":{}}}`

Fixes #3129.


newhoggy:
Ran \`/code-review\` (standard effort) on the initial commit. Fixed all 4 confirmed findings in a follow-up commit:

- **Structural risk**: \`identity_bind_position\`'s own \`position()\` closure had drifted wider than \`is_identity_passthrough\`'s grammar (its \`If\`/\`Alternative\` arms carry no raise/empty gate), safe only via a comment-documented invariant at a call site (\`substitute_bound_var_at\`) three functions away — a future edit trusting \`identity_at\` directly could silently reopen this exact hole. Gated \`identity_bind_position\` itself by \`is_identity_passthrough(source)\` at its own single choke point, matching its own doc's "mirrors `getpath_result_position`'s single choke point" framing.
- **Test duplication** (2 findings): the new regression test duplicated the doc/message fixture and \`query!\`-loop shape of two existing, table-driven tests instead of extending them. Folded its rows into \`test_raising_try_body_is_not_an_identity_passthrough_2978\` (main repro rows + control) and \`test_identity_bind_position_traps_keep_refusing_2978\` (positional twin), dropped the now-redundant standalone test.
- **Redundant comment**: trimmed an inline match-arm comment restating the same mechanism a third time.

One review pass (a subagent finding that arrived after the main synthesis) additionally raised a plausible-sounding deeper concern: that \`is_raise_free_identity_passthrough\`'s own \`Alternative\` arm (\`Expr::Alternative(left, _) => is_raise_free_identity_passthrough(left)\`) ignores its right operand, so a \`try (A // B) catch C\`-shaped body could theoretically let an unconstrained \`B\`/\`C\` through as "raise-free" whenever \`A\` is falsy. I spent real effort trying to construct a live divergence from \`/usr/bin/jq\` 1.7.1 for this (several structural variants, including one needing a \`$root\`-rebind to route around jq's own body-continues-from-pre-`as`-register semantics) and could not reproduce one — every construction I tried refuses identically in both jq and succinctly, jq losing its own path-tracking at the same structural point. Filing this as a separate follow-up for further investigation rather than blocking this PR on an unconfirmed claim; will link it here once filed.

All local checks green: full test suite, clippy (both feature sets), fmt, 100% patch coverage (\`omni-dev coverage diff\`), and the original repro re-verified end-to-end against the pinned oracle after the refactor.

newhoggy:
Filed #3143 for the deeper, unconfirmed `is_raise_free_identity_passthrough` concern mentioned above.

PR #3147, opened 2026-09-18
fix(jq): recognize every identity-passthrough spelling in a pattern's first-step identity test

### LINKED ISSUE #3119: jq: a wrapped identity source (`(try . catch 1) as {a:$v} ?// $v`) fails the pattern's first step and retries onto the wrong alternative

newhoggy:
## Summary

`resolve_as_pattern`'s first-step identity test (jq's `path_intact`) only recognized a
bare `.` head as "the source value is the register's own node." Every other
`is_identity_passthrough` spelling (`try . catch 1`, `if true then . else . end`,
`. // 1`) fell to the null/bool catch-all, wrongly decided the step was *not* intact, and
either refused (a single pattern) or **retried onto the next `?//` alternative** —
binding a different alternative than jq does, which `del`/`=` then write through. The
load-bearing case is a silent wrong delete: `del((. // 1) as {a:$v} ?// $v | $v)` on
`{"a":"s","c":"s"}` wrote `null` (deleting the whole document) where jq deletes one key,
leaving `{"c":"s"}`.

Fix: add dedicated `If`/`Try`/`Alternative` match arms to the `identical` computation,
mirroring `is_identity_passthrough`'s own grammar exactly. The `Alternative` (`A // B`)
arm additionally requires the register be truthy — `A // B` only equals `.` exactly when
`A` actually produced the value; `B` runs, and is completely unconstrained, whenever the
register is null/false. This is *stricter* than the issue's own suggested `bound == value`
gate (which alone could re-admit a #3129-style value-coincidence fabrication off-register,
per #3120's carried-register mechanism) and stays sound in that case, at the cost of one
confirmed refuse-only divergence from jq (documented and pinned in the new test).

## Test plan

- [x] `cargo fmt --check`
- [x] New regression test (`test_alternative_identity_passthrough_pattern_head_is_recognized_3119`) pins all 6 rows from the issue's own repro table against `/usr/bin/jq` 1.7.1 (including both controls), plus an `if` twin not in the issue but the same gap, plus a confirmed refuse-only off-register divergence from the `Alternative` arm's extra runtime gate
- [x] `cargo llvm-cov` + `omni-dev coverage diff`: 98.65% patch coverage (73/74), the one remaining line a defensive panic arm annotated `omni-dev: coverage tolerate-line`, matching the identical pre-existing pattern elsewhere in this file
- [x] End-to-end verified against the pinned oracle: `del((. // 1) as {a:$v} ?// $v | $v)` on `{"a":"s","c":"s"}` now matches jq's `{"c":"s"}` exactly, where it previously silently deleted the whole document

Fixes #3119.


newhoggy:
Ran \`/code-review\` (standard effort) three times on this PR — the first pass caught a fundamental soundness gap in the initial fix (reusing \`is_identity_passthrough\`'s weaker, deferred-certification guarantee for an immediate verdict), the second caught a residual empty-condition hole one level deeper in the corrected \`Alternative\` arm, and the third (after fixing both) found no further correctness issues across two independent fresh passes.

**Fixed across all three rounds:**
- Initial fix bolted three ad-hoc \`If\`/\`Try\`/\`Alternative\` arms onto \`resolve_as_pattern\`'s \`identical\` check, reusing \`is_identity_passthrough\`/\`is_raise_free_identity_passthrough\` directly — unsound, since those predicates only certify a shape as a *candidate* re-verified later at a separate use site; there is no later check here. Confirmed live: a nested \`A // B\`, or a \`TrackedVar\` frozen at an earlier position than the current register, both got wrongly certified and \`del\`/\`=\` wrote through nodes jq itself refuses to reach.
- Replaced with a single recursive helper, \`resolves_to_register\`, that re-runs the bare-\`TrackedVar\` arm's own \`marker.value == *reg && frame.certifies(...)\` check at every nesting level instead of trusting \`trackable\` as a stand-in.
- Its \`Alternative\` arm's own \`reg.is_truthy()\` gate turned out to prove "never filtered by \`//\`" but not "produces a value at all" — an \`If\` on \`left\` with an empty-yielding condition (\`if empty then . else . end\`) slipped through and got \`B\`'s unrelated literal certified as the register. Closed by additionally requiring \`is_raise_free_identity_passthrough(left)\`, matching the guard \`Try\` already needed and \`is_identity_passthrough\`'s own (already-#3129-fixed) \`Alternative\` arm.
- Removed the resulting duplication (Identity/TrackedVar checks existed both inline and in the new helper), hoisted the now-single \`resolves_to_register\` call out of the per-\`bound\` loop (it's loop-invariant), and dropped a redundant variant-list match at the call site.
- Recorded both confirmed refuse-only residuals in \`docs/compliance/jq/limitations.md\` per ADR-0018.

18 regression rows now cover the full history: the issue's original repro, the first fix's own gap, the second's, plus two rows proving \`resolves_to_register\`'s null/bool fallback and its \`If\` arm's \`&&\` (not \`||\`) are both load-bearing. Every row confirmed live against \`/usr/bin/jq\` 1.7.1. Full local suite green: tests, clippy (both feature sets), fmt, 99.32% patch coverage (the one uncovered line a pre-existing-pattern defensive panic arm, \`tolerate\`-annotated).

Issue #3149, opened 2026-09-18
jq: a $param bound to a generator argument is substituted, not bound per value -- def g($x; f): [$x, f]; g(1,2; 3,4) gives [[1,2,3,4]]


newhoggy:
**Severity: Low**

## Summary

Found while pinning #2955's rows. jq desugars `def g($x; f): body` to `def g(x; f): x as $x | body`, so a generator argument for `$x` runs `body` once per value. succinctly substitutes the argument *expression* for `$x`, so the generator runs inline where `$x` is read.

## Repro (jq 1.7.1 vs `main` @ `444c459c9`, no modules)

```console
$ jq -nc 'def g($x; f): [$x, f]; [g(1,2; 3,4)]'
[[1,3,4],[2,3,4]]
$ succinctly jq -nc 'def g($x; f): [$x, f]; [g(1,2; 3,4)]'
[[1,2,3,4]]
```

Same through a forwarding def (`def s(x; f): g(x; f);`), through a `$`-spelled forwarder, and through a module boundary before and after #2955 -- it is the evaluator's own model, not the loader's. With a single-valued argument (`g(1; 3,4)`) both agree (`[[1,3,4]]`).

Two `$x` reads in the body presumably also differ (jq reads one bound value twice; succinctly re-runs the generator -- or memoises it via `eval_shared_chain_link`, #3012; worth capturing).

## Where

`bind_def_call_params` (`src/jq/eval.rs`) builds a `DOLLAR_NAMESPACE_ONLY` substitution from `dollar_safe_shared(arg)`; `resolve.rs`'s module header table records jq's `x as $x` desugaring. Not recorded in `docs/compliance/jq/limitations.md` as far as a grep for the shape finds.

## Class: Opus

Changing `$param` from substitution to a per-value binding touches the `Shared` hygiene model (#2077) and the memoisation in #3012.

Issue #3148, opened 2026-09-18
perf(jq): the evaluator retains one bound body per DefCall node across the whole call tree; fib(21) overflows the stack


newhoggy:
**Severity: Medium**

## Summary

Found while closing #2955. `bind_def_call` caches each `DefCall` node's bound body in that node's `BoundBody`, and the bound body is a fresh copy (`install_def_calls` over `def.body`) whose own `DefCall` nodes are fresh too. So a call *tree* of N calls leaves N cached body copies behind, all reachable from the root call's cache, for the program's lifetime. Peak RSS grows with the number of calls, not the depth, and a plain binary recursion overflows the stack at depth 21. No module is involved.

## Repro (release build of `main` @ `444c459c9`, Apple M-series, `/usr/bin/time -l`)

```
def fib(n): if n < 2 then n else fib(n-1) + fib(n-2) end; fib(N)
```

| N  | calls   | succinctly peak RSS | jq 1.7.1 |
|----|---------|---------------------|----------|
| 16 | 3,193   | 41 MB               | 2 MB     |
| 18 | 8,361   | 95 MB               | 2 MB     |
| 20 | 21,891  | 234 MB              | 2 MB     |
| 21 | 35,421  | **stack overflow, exit 134** | 2 MB |

`def r(n): if n == 0 then 1 else r(n-1) + r(n-1) end; r(13)` (8,192 calls) is 177 MB. The `$n` spelling is the same. The same shape with distinct defs instead of recursion -- 56 defs in one file, each calling two of the level below, resolved at the top (#2955's generator written flat) -- is 58 MB for 16,384 calls, so it is the call tree, not recursion, that is retained.

## Mechanism (as read)

`bind_def_call` (`src/jq/eval.rs`) → `bound.get_or_try_init(|| install_def_calls(&def.body, def, frames + 1, true))`. Each `DefCall` inside that rebuilt body is a new node with an empty `BoundBody`; when it is evaluated it caches *its* copy of *its* callee's body, and so on down. `BoundBody` is write-once and the copies are retained by the parent copy's `Rc<Expr>`, so nothing is freed until the root call's node drops. The stack overflow is presumably `Drop` of, or the walk over, that nested structure at depth ~21 with 2^21 nodes -- to be confirmed.

[...]

PR #3150, opened 2026-09-18
fix(jq): link each dependency module once and forward to it through stubs (#2955, #3058)

### LINKED ISSUE #2955: perf(jq): transitively included modules compound in memory when a module's defs each call more than one def below
**Severity: Low**

## Summary

Found by code review on PR #2954 (#2865). #2865 makes a module's own `include`s
load transitively by wrapping each exported def's body in the defs it reaches.
Because that wrap **copies** the AST, a chain of modules compounds: each level's
bodies already carry the level below.

#2865 mitigates this with jq's own `block_bind_referenced` rule (wrap only what
the body transitively calls), which flattens the common shape completely. It does
**not** cure the general case: as soon as each def calls more than one def from
the level below, the referenced closure is itself exponential in chain depth.

Output is correct throughout — this is a memory/time scalability gap, not a
wrong answer, and not a regression (before #2865 these programs did not compile
at all).

## Repro

Synthetic chains of modules, each def calling `F` defs from the level below,
resolving one def at the top. Apple M-series, release build, measured with
`/usr/bin/time -l` at PR #2954's head:

| chain | fan-out | levels x defs | succinctly peak RSS | jq peak RSS |
|-------|---------|---------------|---------------------|-------------|
| flat  | 1       | 6 x 40        | 10 MB               | 2.5 MB      |
| wide  | 3       | 8 x 6         | 91 MB               | 2.6 MB      |
| wide  | 2       | 14 x 4        | **361 MB**          | 2.6 MB      |

Generator:

```python
def gen(L, D, F, d):
    with open(f"{d}/m0.jq", "w") as f:
        for i in range(D):
            f.write(f"def f0_{i}: {i};\n")
    for lvl in range(1, L):
        with open(f"{d}/m{lvl}.jq", "w") as f:
            f.write(f"include \"m{lvl-1}\";\n")
            for i in range(D):
                calls = " + ".join(f"f{lvl-1}_{(i+k)%D}" for k in range(F))
                f.write(f"def f{lvl}_{i}: {calls};\n")
```

Then `succinctly jq -L <dir> -nc 'include "m13"; f13_0'`.

## Root cause

`ModuleLoader::load_and_bind_module` / `visible_defs_for`
(`src/bin/succinctly/jq_runner.rs`): binding is done by **cloning** each
reachable def's `Expr` into the wrapping body. Real jq binds symbolically — its
`block` values are shared, so a def reachable by many paths exists once.
succinctly's module support is AST inlining, where "reachable by K paths" means
K copies.

## Suggested fix direction

Three shapes, roughly in increasing order of cost:

1. **Share the bound bodies.** `Expr` already has an `Rc`-shaded variant for
   exactly this purpose (the opaque already-substituted sub-expression #1371
   introduced to stop recursive calls growing with depth). If a bound def body
   could be spliced by handle rather than by clone, the copies collapse to
   pointers. Needs a check that every later pass over the tree treats it
   correctly.
2. **Bind a module once into a single shared scope** rather than per exported
   def — closer to jq's own block model, but it is the per-body wrapping that
   buys #2865 its scoping rows (a dependency visible inside the module, not
   re-exported, outranking a same-name sibling), so this needs the scope
   boundary from #2951 to exist first.
3. **A depth/size guard** that refuses with a clear error rather than
   exhausting memory, in the spirit of the other resource caps in
   `docs/compliance/jq/limitations.md`. A stopgap, not a fix.

Option 1 is the one worth costing first; #2951's sealed-scope work would
subsume option 2.

## Why this is separate from #2865, #2951

#2865 is the functional gap (transitive includes not processed at all) and
ships the mitigation that handles the common shape; this is the residual
scalability limit of the mechanism it uses. #2951 is a scoping-correctness gap
whose fix would likely also change this mechanism, but it is not itself about
memory.

### LINKED ISSUE #3058: jq: a compile error in a module dependency reached from two defs is reported once per copy
**Severity: Low**

Found while reviewing #3057 (#2962). A module dependency is bound by copying it into the body of every def that reaches it (#2865). When one of its bodies has a compile error and two reachable defs both reach it, succinctly reports the error once per copy. jq reports it once.

### Repro (captured against `/usr/bin/jq` 1.7.1)

`dep.jq`:
```jq
def g: nosuch;
```
`mid.jq`:
```jq
include "dep"; def a: g; def b: g;
```

```console
$ jq -L D -nc 'include "mid"; a, b'
jq: error: nosuch/0 is not defined at D/dep.jq, line 1:
def g: nosuch;       
jq: 1 compile error
```

`succinctly jq` on `main` (ce8f68e50):
```console
jq: error: nosuch/0 is not defined at D/dep.jq, line 1:
def g: nosuch;       
jq: error: nosuch/0 is not defined at D/dep.jq
jq: 2 compile errors
```

Both exit 3. The difference is in stderr only: the error count, plus a second report without a line number. The second copy's nth-occurrence lookup finds no second site in `dep.jq`, so it falls back to the file-only form. With #3057 the same applies to an unbound `$variable` in a dependency. Before #3057 that case was reported twice at `<top-level>`.

### Likely fix

Deduplicate the diagnostics for one module body, not one textual occurrence. For example, `report_compile_errors` could collapse identical `(origin, name, arity)` entries that come from separate copies of the same dependency. Keep two genuinely distinct occurrences of one name in the same file (`test_second_unresolved_call_in_the_same_module_finds_its_own_line_2991`). One option is to record which copy a diagnostic came from, or to resolve a module's body once at load time instead of once per copy. That second option is the "symbolic binding" direction ADR-0023 describes.

### Class: **Sonnet**

The fix is contained in the reporter, as long as it can tell a copy apart from a second occurrence, which is the part to design. If that needs load-time resolution, reclassify as Opus.

newhoggy:
Closes #2955. Closes #3058.

## What

Binding a module's `include`/`import` dependencies by **copying** their bodies into every def that reached them compounded down a module chain (`F^L` copies of the bottom level). A dependency module is now emitted **once**, outermost in `process_program`, as an ordinary run whose begin marker carries a hidden alias and whose defs are named `<NUL>link:<id>::<name>`; each consuming def's body is wrapped in one forwarding stub per dependency (name, arity) it calls directly. Inside the run a bare sibling call is retried under the alias by the code #2989 added for imported modules, so the evaluator is untouched. The one resolver change is that `scan_scope` steps over an open floor for a link-name lookup. `#2962`'s rename machinery and the transitive closure are gone. ADR-0023 gains the amendment with the soundness argument.

This is the triage plan on #2955, simplified after an adversarial re-derivation against the current code (no new marker kinds, no alias defs, no name-hiding rule) and with the hoisted run filtered to the defs reached by name from the main filter, which a review of the closure found must retry an imported module's bare sibling calls under its alias (pinned).

## Measured (release, Apple M-series, `/usr/bin/time -l`, the issue's generator; output identical to jq on every row)

| chain | before | after | jq 1.7.1 |
|---|---|---|---|
| 6 x 40 x 1 | 12 MB | 9 MB | 2 MB |
| 8 x 6 x 3 | 142 MB | 34 MB | 2 MB |
| 12 x 4 x 2 | 162 MB | 36 MB | 2 MB |
| 14 x 4 x 2 | 681 MB, 0.38 s | 112 MB, 0.03 s | 2 MB |
| 1000-def utility module, one def used through two modules | 9 MB | 9 MB | 2 MB |

Processed-program node count at 6 / 8 / 10 levels: 1253 / 5093 / 20453 before, 169 / 241 / 313 after (`link_size_guard_2955`, which fails against the old loader). Recursion ceiling through a linked dependency: 19494 before and after.

**M4 Pro, idle, interleaved, 25 reps, medians:** `[range(1e5) | f]` with `f` calling one dependency through a stub +1.6%, against +3.3% drift on the same defs inline (neutral within noise). A 20-module x 50-def chain with one def used: 20 ms → 9 ms. **The shape that pays:** the same chain with all 50 top-level defs used links all 950 dependency defs into the top-level chain: 32 ms → 136 ms, because every chain def is installed over the whole program below it at bind time — the evaluator's pre-existing `O(M x N)` (a single 3000-def `include` costs 1 GB / 0.5 s today). Recorded in limitations.md and filed with #3148.

What remains above jq's 2 MB is the evaluator's per-call bound-body retention (#3148, filed from this: the same 56 defs in one file cost 58 MB, `fib(20)` 234 MB, `fib(21)` overflows the stack). #3149 records a `$param`-generator divergence found while pinning rows; it predates this change and is not touched.

## Behaviour changes, all captured from `/usr/bin/jq` 1.7.1

- A compile error inside a dependency reached from two defs is reported **once** (#3058). Residual: a module that is both a top-level include and a dependency, reached both ways, still reports twice (documented).
- Cross-module compile errors now come out in jq's order: a dependency's before its includer's, last-declared dependency first, deepest first (was includer-first).
- `import "inner3" as i; def h: i::k;` inside a module, where `k` calls its sibling bare, now answers `[42]` (was `g/0 is not defined`).

All 30 rows of `test_dependencies_are_bound_in_their_own_scope_2962` and the #2865/#2951/#2989/#2774/#2857/#2991/#2740/#2971/#2395/#2682 module families are unchanged.

## Verification

`cargo fmt --check`, both ci.yml clippy sets, `cargo doc --all-features`, `cargo test --features cli`, `cargo test`, `cargo test --no-default-features` — all green on the rebased tree.


newhoggy:
Rebased onto `5f65aef02` and added one commit from review: `dep_stubs_for` compared an `import ... as ns` dependency's *bare* name against the consuming def's own (name, arity) and parameters, so `import "dep" as ns; def g: ns::g;` (and a parameter or a recursive def named like the dependency) dropped the stub and failed with `ns::g/0 is not defined` where jq 1.7.1 answers 99. Pinned as R34-R36 in the #2962 matrix, all re-captured live. Same commit: `hoist_order` reuses the run ids `process_program` already interned, the per-module closure indexes siblings by name, and both unresolved-call reporters go through `ModuleRun::display_name`. All ci.yml gates re-run green locally on the rebased tree.

PR #3152, opened 2026-09-18
fix(jq): admit a null/bool navigated-bind marker at an equal-valued sibling

### LINKED ISSUE #3136: jq: a null/bool marker bound from a navigated position refuses at an equal-valued sibling where jv_identical admits it
**Severity: Low** (refuse-only)

## Summary

jq's `jv_identical` treats `null`/`true`/`false` as identical by value regardless of node (`register_identical`'s own `null_bool_identical` carve-out models this). The resolver's `TrackedVar` arm consults the marker's origin *before* that carve-out, so a `null`/`bool` marker bound from a navigated position refuses at an equal-valued sibling where jq answers. Pre-existing (reproduces on `bfee1fb1f`); surfaced by #3037's sweep.

## Repro

`/usr/bin/jq` 1.7.1 vs `main`, input `{"a":true,"c":true}`:

| filter                          | jq   | succinctly |
|---------------------------------|------|------------|
| `.a as $y \| .c \| $y \|= 5`    | `5`  | exit 5     |
| `.a as $y \| .c \| path($y)`    | `[]` | exit 5     |
| `.a as $y \| .c \| ($y \| .) = 5` | `5` | `5` (the `.` stage re-establishes by value) |

## Root cause

`resolve_node_sink`'s `Expr::TrackedVar` arm (`src/jq/eval.rs`): `trackable && marker.value == *value && frame.certifies(&marker.origin)` — `certifies(Untracked)` is `false`, and the `null_bool_identical` admission lives only in `register_identical`/`reestablishes_register`, which a bare marker at a trackable position never reaches.

## Suggested fix direction

Admit `trackable && null_bool_identical(&marker.value, value)` in the `TrackedVar` arm regardless of origin (it is exactly jq's rule for those three values, and cannot fabricate: a `null` is identical to every `null`). Sweep row `navigated-bind-bool-sibling` is registered refuse-only and should flip to agree.

Refs #3037, #1466, #2044.

newhoggy:
## Summary

jq's own `jv_identical` treats `null`/`true`/`false` as identical by value regardless of
node (`register_identical`'s own `null_bool_identical` carve-out already models this for the
register itself). The resolver's `Expr::TrackedVar` arm in `resolve_node_eager` consulted the
marker's origin/node-identity check first, so a `null`/`bool` marker bound from a navigated
position refused at an equal-valued *sibling* where jq answers:

```
$ echo '{"a":true,"c":true}' | jq -c '.a as $y | .c | $y |= 5'          # 5
$ echo '{"a":true,"c":true}' | succinctly jq -c '.a as $y | .c | $y |= 5'  # exit 5
```

Fix ORs in `null_bool_identical(&marker.value, value)` alongside the existing node-identity
check, gated the same way by `trackable`. A non-null/bool equal-valued sibling is unaffected
and still refuses, since jq's own `jv_identical` requires actual pointer identity for those
(verified live: `.a as $y | .c | path($y)` on `{"a":{"b":1},"c":{"b":1}}` still refuses on
both).

Flipped the pre-existing oracle sweep row (`navigated-bind-bool-sibling`,
`scripts/jq-bind-origin-oracle-sweep.sh`) from `refuse-only` to `agree`, and the CLI test that
pinned the old refusal (`test_navigated_bind_residuals_refuse_cleanly_3037`).

## Test plan

- [x] `cargo fmt --check`
- [x] New CLI test (`test_navigated_bind_null_bool_sibling_answers_3136`) covers all three of
      `null`/`true`/`false`, both the read (`path()`) and write (`|=`, `del`) directions, plus
      a control confirming a non-null/bool equal-valued sibling is unaffected — every row
      confirmed live against `/usr/bin/jq` 1.7.1
- [x] Ran `scripts/jq-bind-origin-oracle-sweep.sh` against the release binary: `agree=302
      fabricate=0 mismatch=0 refuse-only=35 refuse-only-NEW=0` — no regressions across the
      whole `$var`-node-identity oracle matrix
- [x] `cargo llvm-cov` + `omni-dev coverage diff`: 100% patch coverage (3/3 new lines)

Fixes #3136.


newhoggy:
Ran \`/code-review\` (standard effort). Two independent review passes found:

1. **A structurally identical sibling gap** in \`resolves_to_register\` (backs \`resolve_as_pattern\`'s first-step identity test, added by #3119) — its own \`TrackedVar\` arm had the exact same node-identity-first ordering this PR just fixed in \`resolve_node_eager\`. Verified live: \`.a as \$y | .c | \$y as \$z | path(\$z)\` on \`{"a":true,"c":true}\` refused where jq answers \`[]\`. Fixed the same way (OR in \`null_bool_identical\`).
2. **Boolean-shape inconsistency** — the new condition read \`A && (B && C || D)\` where \`register_identical\` (the pre-existing sibling carve-out) reads \`A && B && (C || D)\`; logically identical (D already implies B) but structurally different for no reason. Restructured both new sites to match.
3. **Two stale doc comments** — \`Origin::Untracked\`'s own doc in \`expr.rs\`, and two \`docs/compliance/jq/limitations.md\` passages, flatly claimed a sibling "always refuses" without the null/bool exception this fix (and the original #3136 fix) carve out. Updated both.

Extended the regression test with the pattern-context case (\`\$y as \$z | ...\`) covering the sibling fix, plus its own non-null/bool control. Full local suite green: tests, clippy (both feature sets), fmt, 100% patch coverage, and \`scripts/jq-bind-origin-oracle-sweep.sh\` re-run clean against the release binary (\`agree=302 fabricate=0 mismatch=0 refuse-only-NEW=0\`).

newhoggy:
Second \`/code-review\` round found and fixed:
- **Extracted \`marker_identical\`**: both new \`TrackedVar\` arms hand-inlined the identical formula, matching \`register_identical\`'s own precedent and its documented reason for existing as one function (to prevent exactly this kind of drift).
- **Fixed three stale doc comments** left behind by the previous round's edits: two flatly claimed a marker "never certifies"/nothing touches the certification rule without the null/bool exception, one quoted the pre-refactor formula verbatim.
- **Fixed two wrong example outputs** I introduced in the previous round's own doc edits — \`path(.a as \$y | .c | \$y)\` on \`{"a":true,"c":true}\` answers \`["c"]\`, not \`[]\` — caught by re-verifying against \`/usr/bin/jq\` 1.7.1 directly rather than trusting my own prior claim.
- One review pass raised a serious-sounding concern that the CLI's actual evaluation path (\`eval_generic.rs\`) might bypass this fix entirely, based on a test run that showed 4/6 rows failing. I re-ran the exact same test immediately afterward (both via \`cargo test\` and by invoking the release CLI binary directly) and got a clean pass on all rows — the agent's build most likely raced against an in-progress edit of mine (a compile error was open at that moment from an incomplete refactor). Direct, repeated verification just now confirms the fix works correctly through the actual CLI binary for both \`path()\` reads and \`|=\`/\`del()\` writes.
- Also caught and fixed an oversized commit subject (138 chars, over this repo's 100-char lint limit) via a targeted reword + force-with-lease push — this branch has no other claimants.

Full local suite green: tests, clippy (both feature sets), fmt, 100% patch coverage, sweep script re-run clean (\`agree=302 fabricate=0 mismatch=0 refuse-only-NEW=0\`), and the CLI binary directly re-verified against the pinned oracle post-rebase.

PR #3158, opened 2026-09-18
fix(jq): add strftime's missing %c/%x/%X/%r/%k/%l specifiers, fix %P/%E/%O

### LINKED ISSUE #3055: jq: strftime leaves %c %x %X %r %k %l literal and prints %z/%Z as UTC where jq uses the local zone

newhoggy:
## Summary

`format_strftime`'s specifier match had no arms for `%c`, `%x`, `%X`, `%r`, `%k` or `%l`,
so its unrecognized-specifier fallback printed them literally (e.g. `%c` instead of a
formatted timestamp) — and `%P` printed lowercase `am`/`pm` (glibc's own rule) where the
pinned oracle (macOS libc's `strftime`) prints the bare letter `P` unconditionally, not
tied to am/pm at all.

```
$ TZ=UTC jq -n '1700000000 | [strftime("%c"), strftime("%x"), strftime("%P")]'
["Tue Nov 14 22:13:20 2023", "11/14/23", "P"]
$ TZ=UTC succinctly jq -n '1700000000 | [strftime("%c"), strftime("%x"), strftime("%P")]'  # before this PR
["%c", "%x", "pm"]
```

## What changed

- Added `%c`/`%r` as fixed compositions of the existing fields; merged `%x` into `%D` and
  `%X` into `%T`, since the oracle confirms they're byte-identical on this platform; added
  `%k`/`%l` as the blank-padded 24-/12-hour twins of `%H`/`%I`, sharing a single `hour12`
  computation with `%I` and `%r` instead of recomputing it per arm (also removed the
  duplicate weekday/month-abbreviation arrays `%a`/`%b` each hand-rolled, sharing them with
  the new `%c`).
- `%E`/`%O` (POSIX's "alternative representation" modifiers) needed real lookahead, not
  just a bare-letter fallback: confirmed live that every `%E<x>`/`%O<x>` combination for a
  recognized specifier `x` is a pure pass-through to `%x` on this platform (`%EY`, `%Oy`,
  `%Ez`, etc.), that a true end-of-format `%E`/`%O` prints the bare letter, and that
  followed by anything else unrecognized it's silently empty instead (`%E-` is `""`, not
  `"E-"`). Implemented via a one-line recursive delegation for the pass-through case.

## Out of scope (tracked separately)

`%z`/`%Z` for plain `strftime` (not `strflocaltime`) should take the local zone rather than
a fixed UTC one, per this issue's own note — blocked on #3054's still-undecided IANA zone
source, so left as a documented residual (already tracked by #3054) rather than bolted on
here.

## Test plan

- [x] `cargo fmt --check`
- [x] New test (`test_strftime_missing_specifiers_3055`) pins every added/fixed specifier at
      an afternoon timestamp and at midnight/epoch (single-digit-day/blank-padding edge
      cases), the `%E`/`%O` pass-through and end-of-format fallback cases, and confirms
      `%D`=`%x`/`%T`=`%X` stay byte-identical after the merge — plus explicit coverage for
      `%a`/`%b`/`%I` (pre-existing behavior my refactor touched, previously untested by any
      unit test). Every row confirmed live against `/usr/bin/jq` 1.7.1
- [x] `cargo llvm-cov` + `omni-dev coverage diff`: 98.44% patch coverage (63/64), the one
      remaining line a `tolerate`-annotated llvm-cov line-attribution artifact (verified via
      raw lcov hit counts that the call itself executes 3 times under the new test)

Fixes #3055.


newhoggy:
Ran \`/code-review\` (standard effort). Fixed the confirmed findings:

- **A real duplication gap**: \`resolve_node_eager\`'s \`TrackedVar\` arm hand-inlined the #3136 null/bool certification check instead of calling \`marker_identical\`, despite that helper's own doc comment claiming it's shared between this arm and \`resolves_to_register\`'s identical one — the two only agreed because they were textually identical. Now both call \`marker_identical\`.
- **A comment citing memory instead of the oracle**: the \`%k\`/\`%l\` blank-padding comment said "glibc/BSD \`strftime\`" (a from-memory library-convention claim); replaced with the actual live-verified oracle values.
- **A maintenance note** on the \`%E\`/\`%O\` pass-through allow-list, flagging that it's a second, hand-maintained copy of the specifier set the match above implements.

One finding was a **genuinely pre-existing bug**, not introduced by this PR (confirmed via \`git show main:src/jq/eval.rs\` before my changes): \`%a\`/\`%A\`/\`%b\`/\`%B\` (and now \`%c\`) index weekday/month names with an unvalidated \`as usize % N\` cast, silently wrapping an out-of-range hand-constructed value into a wrong-but-valid-looking name instead of jq's own \`"?"\` sentinel. Filed as #3159 rather than fixed here, since it needs its own investigation across every specifier that reads these two fields — this PR's own new \`%c\`/\`%r\`/\`%k\`/\`%l\` arms don't introduce any new instance of it beyond reusing the same pre-existing constants.

Full local suite green: tests, clippy (both feature sets), fmt, 98.46% patch coverage (the one remaining line the same \`tolerate\`-annotated llvm-cov attribution artifact from before, re-verified after the rebase).

